PR #281, opened 2016-10-11
Destroy kinematics plugins before their loader

jonbinney:
In small test programs, I've been getting the following error:
```
[/kinematic_plugin_test] [1476222361.566864946]: SEVERE WARNING!!! Attempting to unload library while objects created by this loader exist in the heap! You should delete your objects before attempting to unload the library or destroying the ClassLoader. The library will NOT be unloaded.
```
I tracked this down to the order of the member variables in RobotModelLoader. Because the KinematicsPluginLoader is last, it gets destroyed before the robot model. The robot model contains shared pointers to the loaded kinematics plugins, so they haven't been destroyed yet. The kinematics plugin loader then gives the above "Severe Warning". Changing the order of variables corrects this.

To reproduce, compile the following program:
```
#include <ros/ros.h>
#include <moveit/robot_model_loader/robot_model_loader.h>
#include <moveit/planning_scene_monitor/planning_scene_monitor.h>
#include <tf/transform_listener.h>

int main(int argc, char **argv) {
    ros::init(argc, argv, "kinematic_plugin_tests");

    ros::AsyncSpinner spinner(1);
    spinner.start();

    boost::shared_ptr<tf::TransformListener> tf(new tf::TransformListener(ros::Duration(10.0)));

    {
        //robot_model_loader::RobotModelLoader rml("robot_description", true);
        planning_scene_monitor::PlanningSceneMonitorPtr planning_scene_monitor(
            new planning_scene_monitor::PlanningSceneMonitor("robot_description", tf));
    }
    return 0;
}
```
Run it after calling `roslaunch moveit_resources test_environment.launch` in another window.

### DIFF
--- a/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h
+++ b/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h
@@ -144,9 +144,9 @@ private:
 
   void configure(const Options &opt);
 
-  robot_model::RobotModelPtr model_;
   rdf_loader::RDFLoaderPtr rdf_loader_;
   kinematics_plugin_loader::KinematicsPluginLoaderPtr kinematics_loader_;
+  robot_model::RobotModelPtr model_;
 
 };
 
--- a/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h
+++ b/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h
@@ -144,9 +144,9 @@ private:
 
   void configure(const Options &opt);
 
+  robot_model::RobotModelPtr model_;
   rdf_loader::RDFLoaderPtr rdf_loader_;
   kinematics_plugin_loader::KinematicsPluginLoaderPtr kinematics_loader_;
-  robot_model::RobotModelPtr model_;
 
 };
 
--- a/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h
+++ b/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h
@@ -100,6 +100,8 @@ public:
 
   RobotModelLoader(const std::string &robot_description, bool load_kinematics_solvers = true);
 
+  ~RobotModelLoader();
+
   /** @brief Get the constructed planning_models::RobotModel */
   const robot_model::RobotModelPtr& getModel() const
   {
--- a/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
+++ b/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp
@@ -51,6 +51,13 @@ robot_model_loader::RobotModelLoader::RobotModelLoader(const Options &opt)
   configure(opt);
 }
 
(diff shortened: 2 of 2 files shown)


mikeferguson:
+1

jonbinney:
Argh - while this PR is an improvement, it won't guarantee that the kinematics plugins are destroyed before the plugin libraries are unloaded. Because RobotModel is a shared_ptr that can be gotten by calling `RobotModelLoader::getModel()` someone else could have a shared pointer to the RobotModel after RobotModelLoader is destroyed.

I don't see any easy way around this. Moveit passes around shared pointers all over the place, and that seems pretty fundamentally incompatible with needing a certain object (in this case a kinematics plugin) to be destroyed before a certain other object (in this case the class loader instance).

Am I understanding this correctly? For example, the following test program generates the original error message, even with my fix:
```
#include <ros/ros.h>
#include <moveit/robot_model_loader/robot_model_loader.h>

int main(int argc, char **argv) {
    ros::init(argc, argv, "kinematic_plugin_tests");

    robot_model::RobotModelPtr rm;
    {
        robot_model_loader::RobotModelLoader rml("robot_description", true);
        rm = rml.getModel();
    }
    return 0;
}
```

jonbinney:
Hmmm... it looks like the KinematicPluginLoader uses the relatively new `ClassLoader::createUniqueInstance()`: https://github.com/ros/pluginlib/blob/86fb2bb84223c7134ef81a1e4cd42ccdc5c96820/include/pluginlib/class_loader.h#L112

/cc @de-vri-es for more info on how that works.

v4hn:
As far as I know @rhaschke analyzed issues with this in the recent past.

rhaschke:
This is a fundamental design flaw of the class_loader lib: https://github.com/ros/class_loader/pull/34#issuecomment-205002900 and needs to be solved there.

However, your PR is definitely an improvement for MoveIt. Instead of changing the declaration order in header (and thus changing ABI), what about explicit destruction in correct order in destructor?

jonbinney:
@rhaschke interesting - will adding an explicit destructor not break ABI compatibility?

v4hn:
On Wed, Oct 12, 2016 at 05:35:52PM -0700, Jonathan Binney wrote:
> @rhaschke interesting - will adding an explicit destructor not break ABI compatibility?

No, it should not.

jonbinney:
Great, @rhaschke I've updated the PR with your suggestion - there is now an explicit destructor that destroys things in the right order.

jonbinney:
Added explanatory comment as @v4hn suggested.

@rhaschke is the underlying pluginlib/classloader problem captured in an issue on one of those repos? I see https://github.com/ros/class_loader/issues/33 but it seems like the description of the problem is in PRs that you made which have either been merged or closed. Would it be worth summarizing some of your points  from https://github.com/ros/pluginlib/pull/37 and https://github.com/ros/class_loader/pull/34 in https://github.com/ros/class_loader/issues/33 (or creating a new issue?)

rhaschke:
The issue in class_loader is filed here: https://github.com/ros/class_loader/issues/33, with references to all the other more detailed issue descriptions.

v4hn:
Does this have to be picked to i/j?

jonbinney:
I think this probably should be cherry-picked to i/j.

v4hn:
Picked to indigo and jade. Thanks again for the analysis!

Turns out this destructor existed in indigo without an explanation but got removed in j/k.
I'm afraid I might be responsible for that...

PR #297, opened 2016-10-18
some cleanup

rhaschke:
These two commits are unquestioned changes from #232, which @davetcoleman  anyway asked to be separate PRs.


v4hn:
I fully agree with these changes. Merging.

v4hn:
I believe this should be cherry-picked to indigo/jade.
If you agree please do that.

rhaschke:
Yes. Cherry-picked to I/J.

PR #310, opened 2016-10-22
change find_package Eigen3 to 'Eigen 3' in chomp_interface

mirsking:
Since there isn't FindEigen3.cmake in cmake_modules of moveit on indigo branch, change this to find Eigen while require version greater than 3.


v4hn:
Thanks for reporting this!
This is a simple mistake @davetcoleman did not spot when he backported the CHOMP interface.

PR #351, opened 2016-11-13
MSA: show disabled collisions as matrix

rhaschke:
This PR allows to switch views between (classical) linear and (new) matrix view when showing disabled collisions in setup assistant. To this end, the `TableWidget` is replaced by a proper Model/View approach, where the matrix model directly accesses the underlying data structure storing the disabled collision pairs.
A proxy model maps the matrix model to a linear model.

Additionally, I added regexp filtering of link names, toggling of multiple pairs by <space>, and reverting of changes.

This is a screenshot of the new matrix view:
screenshot-msa

This PR uses C++11 features. Before backporting this to Indigo, one would need to replace those.


davetcoleman:
Super cool! 

I understand why the middle diagonal line has no checkboxes - you can't disable a link against itself. Perhaps to aid visuals you could make those shaded a dark color? In addition, I do not understand why there are other boxes without checkboxes - could you explain that?

Is there a legend for the colored boxes you are using?

moveit_setup_assistant/src/tools/compute_default_collisions.cpp
```diff
@@ -200,12 +200,14 @@ LinkPairMap computeDefaultCollisions(const planning_scene::PlanningSceneConstPtr
 
   // Create Connection Graph
   computeConnectionGraph(scene->getRobotModel()->getRootLink(), link_graph);
-  *progress = 2;  // Progress bar feedback
+  *progress = 2; // Progress bar feedback
```
davetcoleman: looks like you have a different formatter conflicting with clang-format

moveit_setup_assistant/src/widgets/collision_matrix_model.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+/// Boost mapping of reasons to a background color
+static const boost::unordered_map<moveit_setup_assistant::DisabledReason, QVariant> longReasonsToBrush =
+    boost::assign::map_list_of  // clang-format off
+    ( moveit_setup_assistant::NEVER, QBrush(QColor("lightgreen")) )
+    ( moveit_setup_assistant::DEFAULT, QBrush(QColor("lightpink")) )
+    ( moveit_setup_assistant::ADJACENT, QBrush(QColor("powderblue")) )
+    ( moveit_setup_assistant::ALWAYS, QBrush(QColor("tomato")) )
+    ( moveit_setup_assistant::USER, QBrush(QColor("yellow")) )
+    ( moveit_setup_assistant::NOT_DISABLED, QBrush());  // clang-format on
```
davetcoleman: This would best be documented http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/setup_assistant/setup_assistant_tutorial.html

moveit_setup_assistant/src/widgets/collision_matrix_model.h
```diff
@@ -0,0 +1,80 @@
...
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Robert Haschke */
+
+#ifndef COLLISIONMATRIXMODEL_H
+#define COLLISIONMATRIXMODEL_H
```
davetcoleman: ``MOVEIT_ROS_MOVEIT_SETUP_ASSISTANT_WIDGETS_COLLISION_MATRIX_MODEL_``

moveit_setup_assistant/src/widgets/collision_matrix_model.h
```diff
@@ -0,0 +1,80 @@
+/*********************************************************************
```
davetcoleman: if this is not a widget, i.e. not a view, then I think it belongs in ``src/tools`` next to ``compute_default_collisions.cpp``

moveit_setup_assistant/src/widgets/rotated_header_view.h
```diff
@@ -0,0 +1,54 @@
...
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Robert Haschke */
+
+#ifndef ROTATEDHEADERVIEW_H
+#define ROTATEDHEADERVIEW_H
```
davetcoleman: same header guard naming issue

moveit_setup_assistant/src/widgets/collision_matrix_model.cpp
rhaschke: The color corresponds to the reason for disabling a link pair and the reason is provided as a tooltip when hovering over an item. But you are right, we should document this in the tutorial as well.

moveit_setup_assistant/src/widgets/collision_matrix_model.h
rhaschke: I thought about that during development too. But I decided it more strongly belongs to the associated widget - these model classes are not a tool on its own.

moveit_setup_assistant/src/widgets/collision_matrix_model.h
rhaschke: Fixed those.

rhaschke:
> I understand why the middle diagonal line has no checkboxes - you can't disable a link against itself. Perhaps to aid visuals you could make those shaded a dark color? In addition, I do not understand why there are other boxes without checkboxes - could you explain that?

I think our .srdf file was manually modified and some link pairs were removed. For those, the current (and the old) code simply don't show any checkbox entries. I do consider this a bug that should be tackled soon.

moveit_setup_assistant/src/widgets/collision_matrix_model.h
davetcoleman: I think of ``tools`` as the etc. folder. or how about ``src/models`` then? I don't see why it belongs in a ``widgets`` folder if that is not what it is.

davetcoleman:
> For those, the current (and the old) code simply don't show any checkbox entries. I do consider this a bug that should be tackled soon.

I think you are missing the functionality that used to be available with the checkbox "Show Non-Disabled Link Pairs" in this image:

image

rhaschke:
> I think you are missing the functionality that used to be available with the checkbox "Show Non-Disabled Link Pairs".

No, I didn't missed that. This filter is still there in the linear mode. The actual data structure (`link_pairs_`) simply holds a map of link pairs to `LinkPairData`. It's simply a matter of fact, that our .srdf file is missing some pairs.

davetcoleman:
But as far as I can remember you don't need to explicitly have every link pair listed in the srdf file - only those that have been disabled.

moveit_setup_assistant/src/widgets/default_collisions_widget.cpp
```diff
@@ -581,6 +736,54 @@ void DefaultCollisionsWidget::focusGiven()
...
   disableControls(false);
+  btn_revert_->setEnabled(false);  // no changes to revert
+}
+
+bool DefaultCollisionsWidget::focusLost()
+{
+  if (worker_)
+  {
+    if (QMessageBox::No == QMessageBox::question(this, "Collision Matrix Generation",
+                                                 "Collision Matrix Generation is sill active. Cancel computation?",
```
davetcoleman: "still" misspelled

davetcoleman:
I've just tested this PR locally - thanks again for contributing this big improvement!

**Bugs**
- When I run "Generate Collision Matrix" from the matrix view and wait for the progress bar to finish, I then click "Configuration Files" screen to save my changes to file. But instead I get a popup warning "Collision Matrix Generation is sill active. Cancel computation?" If you say yes to cancel, the program segfaults.
- When I run the "Generate Collision Matrix" from the list view I get the same popup, this time I press Yes (Cancel) and the window freezes.
- On the previously mentioned popup error message, if you click No (do not cancel) the navigation highlight still shows the new window name, even though the actual window still is on "Self-Collisions"
- As previously mentioned, many of the pairs of links are missing a checkbox to allow toggling their state. I checked against the current implementation (mine) and it does not have this same problem - checkboxs are available for all pairs no matter what is listed in the SRDF

**Feature Request**
(feel free to ignore)
- When using the filter search, only limit either the row or column, but show all links for the other so that it is easy to edit a single link against all other links.

rhaschke:
@davecoleman I fixed the gui bugs mentioned and implemented a variant of your feature request: Using right-click context menu on selected rows/columns you can hide selected (and now also all but selected) rows/columns, which should serve your request.

rhaschke:
Also fixed the missing checkboxes problem: The access to the `link_pairs_` map didn't enforce A < B ordering of link pair names on loading and on the new matrix model.

davetcoleman:
The "hide others" menu option is great! I understand the "hide" option but what does the "show" option do? How can you right click on something to show it if its hidden? Also, do you mind making the menu options capitalized to match the rest of the styling of the MSA?

I re-tested this PR and the previously found bugs are fixed - great! 

Do you mind creating a small update to the MSA tutorial now? mentioning this matrix view and perhaps with a better / extra screenshot?

davetcoleman:
@rhaschke this PR is so close to being done, do you have any time to finish it up?

- Fix clang-format
- Remove "show" option on right click menu, or explain to me what it does
- Optional: Capitalize "hide" option on right click menu for better style

davetcoleman:
Fixed up and merged in https://github.com/ros-planning/moveit/pull/394

PR #352, opened 2016-11-14
Fix BFMT default configuration parameters.

rbbg:
Fixing a 2 mistakes in the default planner configuration of BFMT*. Note that the planner still seems to be a little slow on my system, so it doesn't always manage to find a solution with the default `moveit_resources demo.launch` within the alloted 5 seconds.

It should be identical to the default configuration now.


v4hn:
continues #338 
Looks good to me, thanks!

PR #362, opened 2016-11-15
WIP: Add colours to trajectory_visualisation display

bmagyar:
### Description

I needed a way to overlay trajectory displays in an easy to distinguish way so I've added colouring to the trajectory_visualisation display. 

A video can be found here: 
https://www.youtube.com/watch?v=Vj0KNAOtFrw

screenshot_2016-11-15_16-21-48

**Please note that this is a PR to discover interest for this feature.**
Code is probably very dirty and doesn't meet any of the merging requirements, I haven't even rebased it on the current codebase yet, it's a few weeks old. **If I get a few thumbs up, I'd gladly tidy things up to be merged!**

PS: I love this checklist :+1: 

### Checklist

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/include/moveit/rviz_plugin_render_tools/trajectory_visualization.h
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/include/moveit/rviz_plugin_render_tools/trajectory_visualization.h
@@ -122,7 +122,6 @@ protected:
   // Handle colouring of robot
   void setRobotColor(rviz::Robot* robot, const QColor &color);
   void unsetRobotColor(rviz::Robot* robot);
-  std_msgs::ColorRGBA robot_color;
 
   robot_trajectory::RobotTrajectoryPtr displaying_trajectory_message_;
   robot_trajectory::RobotTrajectoryPtr trajectory_message_to_display_;
--- a/moveit_ros/visualization/rviz_plugin_render_tools/include/moveit/rviz_plugin_render_tools/trajectory_visualization.h
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/include/moveit/rviz_plugin_render_tools/trajectory_visualization.h
@@ -105,6 +105,8 @@ private Q_SLOTS:
   void changedShowTrail();
   void changedTrajectoryTopic();
   void changedStateDisplayTime();
+  void changedRobotColor();
+  void enabledRobotColor();
 
 protected:
   /**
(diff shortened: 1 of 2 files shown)


davetcoleman:
+1 to proceeding - I've run into this limitation before and the only current way to differentiate a trajectory is by the transparency level

can you share a screenshot of the updated display properties also?

bmagyar:
screenshot_2016-11-15_16-27-49

I've added a boolean property to disable this functionality, this could be off by default.

bmagyar:
Also I'd only colour the actual moving parts of the robot, I can add that too.

davetcoleman:
I'd suggest the wording "Color Enabled" to match how the RobotState display words "Visual Enabled" and "Collision Enabled"

v4hn:
I like the idea, please proceed!
This is nice to have :-)

bmagyar:
I ran clang-format, but discovered some inconsistencies around pointers and references.

I'd personally prefer the .h version everywhere.

Change it did in .h:
void setRobotColor(rviz::Robot* robot, const QColor& color);

Change it did in .cpp:
void TrajectoryVisualization::setRobotColor(rviz::Robot *robot, const QColor &color)

davetcoleman:
clang-format: I hadn't realized that discrepancy. Feel free to investigate and propose a fix to https://github.com/ros-planning/moveit/blob/kinetic-devel/.clang-format

bmagyar:
Thanks!

I'll change that clang format back by hand first, then investigate.

I'd also like to tweak the behaviour a bit first, now when the .rviz config
is saved with colours on they don't active when loaded. I need to fiddle a
bit there. I'll attempt getting it all done on Friday.

bmagyar:
My last commit fixed the issue with saved rviz configuration of the tickbox & moved those &-s and *-s where they should be.

All green from me :+1:

davetcoleman:
We should not try to override the clang-format default, otherwise all Travis tests will fail in the future. Rather, I have investigated your observation about clang-format and made a upstream fix here: https://github.com/ros-planning/moveit/pull/367

Until then you should do what clang-format suggests so we can merge this

bmagyar:
Right! Actually my hand-fixed stuff is not failing but some of the new
piece of code I touched while adding the final bits. Pushed update.

davetcoleman:
I've confirmed this PR works in my system - having the colored trajectory is great thanks!

davetcoleman:
Please open PRs for Jade and Indigo if you'd like this backported

bmagyar:
Directly to the branches of this repo, right?

davetcoleman:
Yes - create a branch in your fork for i.e. jade and open a pull request to jade-devel in this repo

PR #390, opened 2016-12-28
[indigo] add correct version to all libraries so files

knorth55:
current `indigo-devel` version is `0.7.5`, otherwise versions and so file names are `0.7.3` and `libmoveit_.....so.0.7.3` because there are diff between #383 and #273.
This PR wants to fix this diff.

### DIFF
--- a/moveit_core/background_processing/CMakeLists.txt
+++ b/moveit_core/background_processing/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(MOVEIT_LIB_NAME moveit_background_processing)
 
 add_library(${MOVEIT_LIB_NAME} src/background_processing.cpp)
-set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION 0.7.3)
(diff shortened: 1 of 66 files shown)


v4hn:
The discrepancy you observed is intentional and describes that there has not been ABI breakage in *any* library between version 0.7.3 (that's when the latest ABI changes occurred) and 0.7.5.

v4hn:
Note that maintaining the individual versions for each library is quite a bit of work and we are not doing this for kinetic, but only for indigo. In kinetic we employ the exact scheme you proposed in your request.

knorth55:
I understand.
but it is really complicating situation for users who install moveit! from debian package.
I checked `/opt/ros/indigo/lib` and all SO files tells that you use ver `0.7.3`, however `apt` tells me that you use `0.7.5`.
I thought I got error on installation.
How about adding version like `libmoveit_~~~~.so.0.7`? (I don't know how you use this version number, if there will be no ABI or API changes in `0.7`, it is reasonable)

PR #395, opened 2017-01-02
moveit_core: remove UTF char from v4hn's name in changelog.

gavanderhoorn:
As per subject.

Made DOC job for Indigo fail ([job/Idoc__moveit__ubuntu_trusty_amd64/3](http://build.ros.org/job/Idoc__moveit__ubuntu_trusty_amd64/3/console) fi).


rbbg:
Good catch.

PR #401, opened 2017-01-05
cherry-pick apply* functions in PSI (#381)

v4hn:
This is the jade version of #400.


v4hn:
This was approved last week. Merging.

PR #407, opened 2017-01-08
Issue #406 implementation: Set color of collision objects…

corot:
… via PlanningSceneInterface

I didn't test thoroughly. Specially, I didn't test the Attached methods at all!


v4hn:
Wouldn't it be more convenient to make the color parameters `std_msgs/ColorRGBA` instead of `moveit_msgs/ObjectColor`?
A vector of colors can be readily mapped to the corresponding vector of collision objects by order.
This would remove the need to add valid ids to the `ObjectColor` parameters and in my opinion further simplify the usability here.

The request failed the format test. Could you fix this please?

corot:
Ups...  sorry for the failed tests.  I'll take a look tonight at home.

For  using std_msgs/ColorRGBA... well, is mostly a matter of taste. I choose  ObjectColor because
 a) allows you the flexibility of providing colors just for the objects you want 
 b) allows a good default value (empty id) in case of single object
 c) interface looks more homogeneous, with two moveit_msgs params 
But if I'm alone with my preferences, no problem to change it

v4hn:
On Mon, Jan 09, 2017 at 02:33:24AM -0800, Jorge Santos Simón wrote:
>  a) allows you the flexibility of providing colors just for the objects you want 

good argument

>  b) allows a good default value (empty id) in case of single object

Could you elaborate on this please?
Is that logic implemented in the PlanningSceneMonitor?

You silently skip the color for the single-object functions if the id does not fit.
Wouldn't it make sense to just set the id of the color to the id of the provided object instead?

corot:
>>   b) allows a good default value (empty id) in case of single object
> Could you elaborate on this please?
> Is that logic implemented in the PlanningSceneMonitor?

I mean, I default the ObjectColor vector to an empty vector, and the single ObjectColor to a default constructed ObjectColor, where id is an empty string, so can be cleanly identified as default. But for a std_msgs/ColorRGBA object the default will be a 0, 0, 0, 0 color, what could be a meaningful color and so no well suited as a default value. In the PR I default to ObjectColor() and ignore it if id is empty. That's also why I silently skip the color.... but maybe I should add another case where color's id is not empty but doesn't match the object id

v4hn:
The default-value problem would not exist if you remove the optional parameter and
instead create another version of the relevant functions with two arguments.

This also allows to ignore the id set in the object color and set it to the id of the object instead.

What do you think?

corot:
It sounds reasonable. I'll change the PR.

I just tried to provide similar calls for one and many objects





2017-01-09 14:31 GMT+01:00 Michael Görner <notifications@github.com>:

PR #409, opened 2017-01-09
Issue #406 implementation: Set color of collision objects…

corot:
… via PlanningSceneInterface

2nd try.

I suppose previous PR failed to pass CI due to clang format (he's really picky!!!)
Also, I have tried to accommodate all the comments

### DIFF
--- a/moveit_ros/planning_interface/planning_scene_interface/include/moveit/planning_scene_interface/planning_scene_interface.h
+++ b/moveit_ros/planning_interface/planning_scene_interface/include/moveit/planning_scene_interface/planning_scene_interface.h
@@ -39,6 +39,7 @@
 
 #include <moveit/macros/class_forward.h>
 #include <moveit/robot_state/robot_state.h>
+#include <moveit_msgs/ObjectColor.h>
 #include <moveit_msgs/CollisionObject.h>
 #include <moveit_msgs/AttachedCollisionObject.h>
 #include <moveit_msgs/PlanningScene.h>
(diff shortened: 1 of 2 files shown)


v4hn:
> he's really picky!!!

I hear you... It is a good thing to have a uniform code base, but it does not make it any easier to contribute for users who don't use `clang-format` to fix formatting issues beforehand.

v4hn:
I just tested your patch and the non-attached parts work as expected. Nice, thanks!

I realized this just now (sorry), but it doesn't make much sense to have these color parameters for the attached objects, does it?
At the moment we do not support custom individual colors for attached objects..

corot:
mmm...  to be honest, I never tried to change the color of an attached object! I imagine that someone will want to change the color of an object once it's attached, kind of "highlight"...  but if you say that's not even possible,,, I suppose I should remake the PR

v4hn:
You got that somewhat wrong I suppose.

AttachedCollisionObjects are *not* directly objects in the PlanningScene; they are basically part of the RobotState instead. As such the planning scene (at least at the moment) does not maintain a color for them explicitly. In the RViz visualization, however, the user can choose the color of attached objects in a display property. This does not allow to set the color of *multiple* attached objects independent of each other, but nobody complained about this before and personally I don't see the need to change that.

So while I am in favor of adding backend support in the PlanningScene to set the color of an attached collision object (so that it is shown in that color once you *detached* it again), I would like not to expose this to the front end functions you propose to change. This is, because if I see a function called "applyAttachedCollisionObject" with a parameter "color", I would expect to see the object in that color in RViz after calling the function. This is not the case.

So yes, please remove the attached object functions from the request.
Sorry for making this process a bit tedious to go through. By now I do like the change. :)

btw: you know you can force-push your git-branch instead of closing this request and opening a new one, right?

corot:
Done. Yes, I know I can add commits to the PR, but some groups prefer single-commit PRs.
Btw, clang-formater made me add a imho nonsensical change; I was already respecting the 120 char rule...

v4hn:
> Yes, I know I can add commits to the PR, but some groups prefer single-commit PRs

Please have a look at `git push -f`, i.e. force-push. It is usually encouraged to force-push pull-request branches to cleanup things. So you can *change* your commit there instead of adding new commits.

v4hn:
Thank you for the contribution and your patience! :)
Merging. This will be part of the next release #393 .

corot:
> Please have a look at git push -f, i.e. force-push. It is usually encouraged to force-push pull-request branches to cleanup things. So you can change your commit there instead of adding new commits.

GIT: not a single day without learning something new...  Thanks!

PR #413, opened 2017-01-11
fix Qt4 compatibility for setup assistant

rhaschke:
As pointed out by @v4hn in https://github.com/ros-planning/moveit/pull/394#issuecomment-271185733, recent commit #394 broke Qt4 support for MSA. This is an attempt to fix it.
@v4hn I don't have a Trusty system around anymore. Could you please test this?
Feel free to push any necessary changes directly to this shared branch.


v4hn:
> I don't have a Trusty system around anymore.

Me neither, but I still run ROS indigo (with the exception of MoveIt) on Lunar-Linux :)

I fixed the build and it works great!
Thank you for the cleanup, it would have taken me ages to find my way through this..

PR #421, opened 2017-01-20
Handle multiple shapes in an attached collision object

v4hn:
Objects with multiple collision shapes leave the question what the object's
internal frame is, because a pose is specified *for each shape* individually.

The RobotState::getFrameTransform function handled this by returning the first
registered pose as frame of reference.
This is a perfectly valid approach, so I changed the warning there to a debug message.

RobotState::knowsFrameTransform, however, returned false if asked whether the state knows the
transform for the object. That's inconsistant behavior: if `getFrameTransform` returns a valid result,
then the indicator should also reflect that.

Among other things, the pick-pipeline fails with objects with multiple collision shapes
without this change. I consider this a bug-fix, so I request to merge this into indigo (and forward-port it).

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -983,9 +983,9 @@ const Eigen::Affine3d& moveit::core::RobotState::getFrameTransform(const std::st
     return identity_transform;
   }
   if (tf.size() > 1)
-    logWarn("There are multiple geometries associated to attached body '%s'. Returning the transform for the first "
-            "one.",
-            id.c_str());
+    logDebug("There are multiple geometries associated to attached body '%s'. Returning the transform for the first "
+             "one.",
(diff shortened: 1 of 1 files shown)


130s:
Could any @ros-planning/moveit-developers take a look?

To me, with the tests passing I'm tempted to merge, particularly because time constraint for #393 (may not be a good enough motivation though).

130s:
I merged since the test passed and the code didn't seem breaking at least.

v4hn:
@130s thanks for merging. Please don't forget to cherry-pick the change to j/k :)

130s:
Ok. Forward ported to J https://github.com/ros-planning/moveit/commit/0c325303e75754f512c49105f99fbcf19ac3e8c5, K https://github.com/ros-planning/moveit/commit/44c2f87153d382022dcdc59cb6ba3ae0130db21d

PR #422, opened 2017-01-23
[Indigo] Move metapackages from moveit_metapackage repo.

130s:
Moving the metapackages `moveit_{full, full_pr2, runtime}` into moveit repo for Indigo, as part of https://github.com/ros-planning/moveit_metapackages/issues/11

For Jade onward, I suggest to drop `moveit_full` because `moveit` metapackage has replaced the role. It's still kept in Indigo for backward compatibility only IMO.


130s:
Could anyone including @ros-planning/moveit-developers review? Travis passed.

Issue #425, opened 2017-01-26
Creating moveit::planning_interface::MoveGroup crashes when running in nodelet


gaspers:
### Description

After updating to the current ubuntu packages, we discovered a problem with creating an object instance of `moveit::planning_interface::MoveGroup` in a nodelet. This seems to be an issue with the changes in `moveit_ros/planning_interface/move_group_interface/src/move_group.cpp` in `template <typename T> void waitForAction(..)` and `void waitForExecuteActionOrService(..)`,
where `ros::spinOnce()` was exchanged with a call to the `callbackQueue` (https://github.com/ros-planning/moveit_ros/pull/701). This does not work for nodelet-NodeHandles.

In detail: in commit ros-planning/moveit_ros@9479afcfd6bd05b3580d4bfb87cb773f29a905d4, in lines like
```c++
( ( ros::CallbackQueue * ) node_handle_.getCallbackQueue())->callAvailable();
```

the NodeHandle `node_handle_.getCallbackQueue()` returns a `CallbackQueueInterface *`, which is the base-class of `ros::CallbackQueue *` and (in case of a nodelet) `nodelet::detail::CallbackQueue *`.

The cast seems to break the code, as nodelets have their own implementation of CallbackQueues.

### Your environment
* ROS Distro: Indigo
* OS Version: Ubuntu 14.04
* Source or Binary build? Both
* If binary, which release version? 0.7.6-0trusty-20161230-092613-0800 (bin)
* If source, which git commit or tag? 0.7.6

### Steps to reproduce
Create an object instance of `moveit::planning_interface::MoveGroup` in a _nodelet_:

```c++
using namespace moveit::planning_interface;
MoveGroup::Options options(group_name_);
options.node_handle_ = nh_;
boost::make_shared<MoveGroup>(options, tf_listener_, ros::Duration(0.3));
```

### Expected behaviour
Object instance is created and works properly.

### Actual behaviour
Segmentation fault occurs.

### Backtrace or Console output

```gdb
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffc506f700 (LWP 20788)]
[...]

PR #426, opened 2017-01-26
move_group.cpp: seg fault bug fix

gaspers:
### Description
Fixed a bug with casting CallbackQueueInterface * to ros::CallbackQueue * without a check.
Without this additional check, creating an instance of MoveGroup in a nodelet causes a segmentation fault, because of different CallbackQueue implementations (nodelet::detail::CallbackQueue).

https://github.com/ros-planning/moveit/issues/425

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group.cpp
@@ -188,9 +188,9 @@ public:
         {
           queue->callAvailable();
         }
-        else
+        else // in case of nodelets and specific callback queue implementations
         {
-          ros::spinOnce();
+          ROS_WARN_ONCE_NAMED("move_group_interface", "Non-default CallbackQueue: Waiting for external queue handling.");
         }
       }
     }
@@ -205,9 +205,9 @@ public:
         {
           queue->callAvailable();
         }
-        else
+        else // in case of nodelets and specific callback queue implementations
         {
-          ros::spinOnce();
+          ROS_WARN_ONCE_NAMED("move_group_interface", "Non-default CallbackQueue: Waiting for external queue handling.");
         }
       }
     }
@@ -246,9 +246,9 @@ public:
         {
           queue->callAvailable();
         }
-        else
+        else // in case of nodelets and specific callback queue implementations
         {
-          ros::spinOnce();
+          ROS_WARN_ONCE_NAMED("move_group_interface", "Non-default CallbackQueue: Waiting for external queue handling.");
         }
       }
     }
@@ -264,9 +264,9 @@ public:
         {
           queue->callAvailable();
         }
-        else
+        else // in case of nodelets and specific callback queue implementations
         {
-          ros::spinOnce();
+          ROS_WARN_ONCE_NAMED("move_group_interface", "Non-default CallbackQueue: Waiting for external queue handling.");
         }
       }
     }
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group.cpp
@@ -183,7 +183,15 @@ public:
       {
         ros::WallDuration(0.001).sleep();
         // explicit ros::spinOnce on the callback queue used by NodeHandle that manages the action client
-        ((ros::CallbackQueue*)node_handle_.getCallbackQueue())->callAvailable();
+        ros::CallbackQueue* queue = dynamic_cast<ros::CallbackQueue*>(node_handle_.getCallbackQueue());
+        if (queue)
+        {
+          queue->callAvailable();
+        }
+        else
+        {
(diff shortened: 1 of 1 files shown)


v4hn:
Thank you for the patch! So standard ROS indeed provides another type of callback structure...
I'm sorry, nobody thought of nodelets back then..

I don't really like to rely on RTTI to resolve this, but imo it's roscpp's fault for not exposing a processing-method for the queue via the abstract interface... So I don't see a cleaner way to keep all use-cases working.

I just checked and RTTI is required in *core* MoveIt already, but this patch will force user-code, i.e. everything using the MoveGroupInterface, to compile with it too. Granted, it's not something you switch-off everyday though..

I approve the patch. This is a plain bugfix and easy enough to understand.
Should be forward-ported to j/k.

de-vri-es:
I'm not familiar with nodelets callback queue / threading model, but don't we run the risk of processing the wrong callback queue using `ros::spinOnce()`? If so that could potentially lead to race conditions if handlers are fired that the code doesn't except.

Another option may to to explicitly give a callback queue to the action client on construction and process that queue exactly when desired (of which we will know the exact type, since we created it, so also no RTTI required).

Alternatively, if there is no risk of race conditions, is there a reason not to go for `ros::spinOnce()` in all cases?

gaspers:
ros::spinOnce() was the default before. That was changed by https://github.com/ros-planning/moveit_ros/pull/701 to support specific callbacks, with the problem of dropping nodelet support. That is the reason for not doing ros::spinOnce() in all cases.

de-vri-es:
Right, but then by reverting to `ros::spinOnce()` we don't support specific callback queues any more if the callback queue is not a `ros::CallbackQueue`. Of course it's already broken for nodelets as it is, but we should be sure not to introduce possible race conditions.

v4hn:
On Thu, Jan 26, 2017 at 10:07:50AM -0800, Maarten de Vries wrote:
> Right, but then by reverting to `ros::spinOnce()` we don't support specific callback queues any more
> if the callback queue is not a `ros::CallbackQueue`.

This explicitly was never the case. The requestor of the pull-request for specific callback queues was fine with that restriction.
This patch does not revert anything. It basically adds error handling to the previous request by replacing the static cast with
an RTTI-based dynamic one.

> Of course it's already broken for nodelets as it is, but we should be sure not to introduce possible race conditions.

The patch fixes the recently broken non-`ros::CallbackQueue` support by reverting to the previous behavior in that case.
So this shouldn't break anything that hasn't been broken for years..

de-vri-es:
> The patch fixes the recently broken non-`ros::CallbackQueue` support by reverting to the previous behavior in that case. So this shouldn't break anything that hasn't been broken for years..

Nodelets weren't considered back then. Since this is specifically geared for nodelets it makes sense to look at the implications of that.

However, as far as I can tell from the source, the nodelet manager doesn't even use the default callback queue. It seems to create its own thread pool and manage everything from there. That means the `spinOnce()` may be a nop in the case of nodelets, but that's not a race condition. And since the nodelets are processing their callback queues in a separate thread pool I don't think it matters that the `spinOnce()` may not be doing anything.

More importantly, since all the `waitForAction` calls are in the constructor it should be safe anyway. Any code that is accessing the object before the constructor finished is wrong. So not our problem.

In other words: I think I'm convinced there's no danger of race conditions.

gaspers:
> However, as far as I can tell from the source, the nodelet manager doesn't even use the default callback queue. It seems to create its own thread pool and manage everything from there. That means the spinOnce() may be a nop in the case of nodelets, but that's not a race condition. And since the nodelets are processing their callback queues in a separate thread pool I don't think it matters that the spinOnce() may not be doing anything.

Yes, you are right. Nodelets don't need ros::spinOnce() as the nodelet manager handles threads and processes message queues. Additionally, spinOnce() wouldn't help with different CallbackQueue implementations either, I think. SpinOnce() only calls the global CallbackQueue. So perhaps, I should remove the else-block? That would reduce the fix to preventing the broken cast.
In the case of nodelets, we tested the code without spinOnce successfully. In the default case of a global ros::CallbackQueue, the code should work exactly like spinOnce(). That was stated and tested in https://github.com/ros-planning/moveit_ros/pull/701.

de-vri-es:
> Additionally, spinOnce() wouldn't help with different CallbackQueue implementations either
I think you're right. Maybe we should log a warning instead? Though it is possible that some other thread is already processing the relevant callback queue (like the nodelet queue), in which case the warning would be a bit of a nuisance.

gaspers:
I integrated your suggestion, added a comment (for else branch) and tried to find a helpful warning text that is emitted once instead of calling spinOnce().

v4hn:
Merged and cherry-picked to jade/kinetic. Thank you lots for creating the issue and the patch @gaspers !

gaspers:
Thank you for the fast responses and for including our patch.

PR #433, opened 2017-01-30
Replace const_cast with static_cast where possible

de-vri-es:
This PR replaces `const_cast` with `static_cast` where possible. A `const_cast` is normally an indication of an API problem somewhere. The replaced casts only add `const`, so a `static_cast` will do just fine, with the added benefit that it doesn't look like dangerous things are happening.

Should be fine to pick to I/J. No API or ABI changes.


de-vri-es:
Heh, I did not expect that one letter size difference to matter, but obviously it does :)

I'll wait for CI to pass on all branches and then merge all of them. Probably tomorrow morning.

de-vri-es:
All checks passed, merging.

PR #437, opened 2017-01-31
[Jade] Fix for issue 373 

mmoerdijk:
@davetcoleman :  I just found out that I made an copy paste error in `wrap_python_move_group.cpp` . If you want I could also make a PR for Kinetic


v4hn:
This is a straight-forward fix of a typo and tests passed, so I'll merge it.
Thanks @mmoerdijk !

v4hn:
kinetic version is proposed in #438

PR #438, opened 2017-01-31
Fix typo in setMaxAccelerationScalingFactor mapping (kinetic #437)

v4hn:
Fixed copy paste error when mapping Python to C++ function

Kinetic-version of #437.


jrgnicho:
@v4hn  thanks

Issue #270, opened 2016-10-04
Removing catch(...) instances?


bmagyar:
Hi guys,

I'm working on my own planner and obviously screwed up something in it. I discovered that there are some instances of `catch(...)` basically hiding crashes & possible informative error messages.

Wouldn't it be better to eliminate the use of  `catch(...)` overall in MoveIt? 

Adding some sense to my point:
- Users are given an error message they can use to report their issue, more info for maintainers to address issues, `std::runtime_error`s are handled & printed with some information, `catch(...)` with no information printed only "Planning pipeline threw an exception" is useless.
- Planner issues are raised to the top level, an `std::terminate` after all is pretty harsh, more chance it will get reported
- "Keep going" is probably not a good philosophy when a planner has unhandled exceptions propagate up

https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/move_group/src/default_capabilities/move_action_capability.cpp#L157

https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/move_group/src/default_capabilities/plan_service_capability.cpp#L69

v4hn:
I mostly agree,

these clauses do not really belong there.
`std::runtime_error` should be handled there, but other exceptions should not.
Removing these clauses, however, means that *your entire move_group node* dies if your planner pipeline throws an exception. This is most likely the reason why these clauses are there at all. If your planning request fails due to some internal error, you might still expect to be able to do cartesian path planning (which doesn't use the planner).

I would argue though that all exceptions that are not due to architecture design and misused API, should be `runtime_error`s anyway.

Could you provide a pull-request that removes the cases you noticed in the `kinetic-devel` branch?

davetcoleman:
This seems like a debate between how much MoveIt! should be focused on helping developers and research, versus being "hardened" to continue working even after a crash occurs in the planner, for improved reliability.

As a researcher developing his own planner, I am obviously for removing the ``catch`` statements. +1

bmagyar:
Yeah I've been a user of moveit for a long time, first at PAL Robotics now as a PhD student. Error messages matter in both cases. `catch(...)` hides too much. I'm up for other solutions instead of bare removal, if anyone has a short snippet of a gentle coredump-style or some template "print any exception" solution, that's also good for me. It's more about not letting bugs live under rocks like `catch(...)`.

de-vri-es:
[...]

PR #445, opened 2017-02-14
Remove "catch (...)" instances, catch std::exception instead of std::runtime_error

bmagyar:
### Description
This adds the changes discussed on #270.

I have left only a couple of `catch (...)` instances where I think the changes can be omitted or I wasn't very comfortable with the change.

General logic copied over from the issue:
```
try
{
...
}
catch (std::runtime_error &ex)
{
    logError("Exception caught while processing action '%s': %s", action_name.c_str(), ex.what());
}
catch (...)
{
     logError("Exception caught while processing action '%s'", action_name.c_str()); // here important stuff may remain hidden
}
```

My proposed solution for the pattern above is this:

```
try
{
...
}
catch (std::exception &ex) // since std::runtime_exception < std::exception
{
    logError("Exception caught while processing action '%s': %s", action_name.c_str(), ex.what());
}
```

### Checklist

Thank you!

### DIFF
--- a/moveit_core/background_processing/src/background_processing.cpp
+++ b/moveit_core/background_processing/src/background_processing.cpp
@@ -77,13 +77,9 @@ void moveit::tools::BackgroundProcessing::processingThread()
         fn();
         logDebug("moveit.background: Done executing '%s'", action_name.c_str());
       }
-      catch (std::runtime_error& ex)
+      catch (std::exception& ex)
       {
         logError("Exception caught while processing action '%s': %s", action_name.c_str(), ex.what());
-      }
-      catch (...)
-      {
-        logError("Exception caught while processing action '%s'", action_name.c_str());
       }
       processing_ = false;
       if (queue_change_event_)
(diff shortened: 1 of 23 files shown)


davetcoleman:
Refresher: this is based on your logic from the issue:

> This provides reasonable prints of things that can be printed and fails (hopefully with a stacktrace) on those that cannot.

davetcoleman:
@bmagyar do you mind [documenting](https://github.com/ros-planning/moveit.ros.org) this pattern, and the logic behind it, on the moveit website under [style guidelines](http://moveit.ros.org/documentation/contributing/code/)? Thanks!

v4hn:
Sorry for the delay, thanks for the contribution @bmagyar !

bmagyar:
Thanks for the merge!
I'll add documentation soon.

davetcoleman:
"I'll add documentation soon." almost always means never...

bmagyar:
Yeah, sorry for the cliché line ;) 

https://github.com/ros-planning/moveit.ros.org/pull/117

130s:
I assume this won't get backported to IJ (no api/abi breakage but behavioral change, which is still non trivial)?

v4hn:
We could backport it, but I think this will probably not cause merge conflicts with later backports
and it *might* break one or two setups (which would be pretty broken if they are affected, but anyway).

-- 
Michael Görner, M.Sc. Cognitive Science, PhD Student
Universität Hamburg
Faculty of Mathematics, Informatics and Natural Sciences
Department of Informatics
Group Technical Aspects of Multimodal Systems

Vogt-Kölln-Straße 30
D-22527 Hamburg

Room: F-315
Phone: +49 40 42883-2432

PR #450, opened 2017-02-22
fixed numpy.ndarray indices bug (from #86)

Kukanani:
### Description

A fix for the numpy.ndarray indices bug, the second bug described in #86.

This fix was proposed by @lifangda01 in the same issue.

I was unsure what to do about clang-format, since it seems that it's for C++ only.

### Checklist

Thank you!

### DIFF
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -102,8 +102,8 @@ class PlanningSceneInterface(object):
         mesh = Mesh()
         for face in scene.meshes[0].faces:
             triangle = MeshTriangle()
-            if len(face.indices) == 3:
-                triangle.vertex_indices = [face.indices[0], face.indices[1], face.indices[2]]
+            if len(face) == 3:
+                triangle.vertex_indices = [face[0], face[1], face[2]]
             mesh.triangles.append(triangle)
         for vertex in scene.meshes[0].vertices:
             point = Point()
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -93,26 +93,32 @@ class PlanningSceneInterface(object):
     def __make_mesh(self, name, pose, filename, scale = (1, 1, 1)):
         co = CollisionObject()
         scene = pyassimp.load(filename)
-        if not scene.meshes:
+        if not scene.meshes or len(scene.meshes) == 0:
             raise MoveItCommanderException("There are no meshes in the file")
+        if len(scene.meshes[0].faces) == 0:
+            raise MoveItCommanderException("There are no faces in the mesh")
         co.operation = CollisionObject.ADD
         co.id = name
         co.header = pose.header
         
         mesh = Mesh()
-        for face in scene.meshes[0].faces:
-            triangle = MeshTriangle()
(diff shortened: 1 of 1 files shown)


v4hn:
Nice! Thanks for the pull-request. You're right, strict formatting is only required for C++ (for now...).

I just verified, pyassimp moved from an internal datatype to numpy.ndarray for these fields.

Until now the code is compatible with the pyassimp versions of ROS indigo and kinetic.
This patch breaks compatibility with the old version.

Would you mind to change it into something along the lines of
```
if hasattr(face, '__len__') and len(face) == 3:
    ....
elif hasattr(face, 'indices') and len(face.indices) == 3:
    ....
```
to allow the script to run with the old version too?

Kukanani:
Full disclosure: if accepted, this would be my first successful pull request.

I added an error that is thrown if both approaches fail. Let me know if that's being "too safe." Also, these checks are being run for every face in the mesh, which is extremely slow. I think the better design would be to check only the first face to determine the pyassimp version, then run one of two operations based on the check.

v4hn:
Hm, apparently GitHub decided to drop the mail response I sent yesterday evening. Sorry for that.
Here it is again:

~~~

Hehe, welcome to the happy folks who file pull-requests :)

You're right, it's not very efficient to check this for every face.
Sure, you can also move the conditional outside the loop and run different
loops if you prefer that.

Sadly, the library itself does not seem to export its own version, but in this
case, it might be better to set a global indicator `PYASSIMP_VERSION`(?) at the
very beginning of the file and use that for the check.
There's a conditional there already because pyassimp > 3.0 is included via
`import pyassimp` and not via `from pyassimp import pyassimp` anymore.

I don't think the additional exception is a good idea though.
The old code just ran on when it encountered an invalid face and added an empty
`MeshTriangle`.
Now, we can definitely skip adding the empty entry, but the rest of the mesh
should still be parsed.
There's all kinds of weird mesh files out there, so we should probably extract
whatever we can. :)

Kukanani:
I agree that if we are looping over every face, then it does not make sense to raise exceptions. 

However, in the code I wrote above, the exception at the end is catching the case where the first face specifically is not loadable via either method. It assumes that all the faces in the mesh have the same Python structure. If that's not the case, then I would propose keeping the loop  structure as it currently is, where `hasttr()` is evaluated for every face, and then remove the exception.

So I would say that the two best options are 1. Use the code above, checking only the first face's structure, with the exception, or 2. Check every face for pyassimp structure, throwing no exceptions.

v4hn:
> I agree that if we are looping over every face, then it does not make sense to raise exceptions.

Sorry, that is the current state of the pull-request though. I commented on that...

> 1. Use the code above, checking only the first face's structure, with the exception
> 2. Check every face for pyassimp structure, throwing no exceptions.

Both ways are fine by me. Just choose one and implement it in the request. :)

With (1.), you might want to check for `len(scene.meshes) > 0 and len(scene.meshes[0].faces) > 0` before for completeness sake.

Kukanani:
Alright, I think it should be ready to go.

cschindlbeck:
I encountered this bug in the indigo branch as well, the same changes seem to fix that for me, pull request?

v4hn:
@cschindlbeck I still suspect you broke your local setup somehow, but I don't see any harm in having this
in `indigo-devel` too.

Feel free to file a pull-request for this.

PR #451, opened 2017-02-25
[indigo] cherry-pick #377 (Add set_max_acceleration_scaling_factor to moveit_commander).

130s:
#377 against `jade-devel` was ported to kinetic-devel in #385 but never to indigo, so is this.
#437 seems also relevant, and is included in this PR.

Squashed into a single commit.


130s:
#437 seems also relevant, and is included in this PR.

v4hn:
LGTM. Merging.

PR #452, opened 2017-02-27
correct "simplify widget handling"

Jntzko:
The changes from the pull request #442 leads to an error.

Hiding the docking area when the motion planning panel is part of it disabled the whole display including the motion planning  *scene* that is displayed in the main window.

So instead of relying on the setAssociatedWidget mechanism, we have to implement the wanted behavior ourselves.

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
@@ -39,6 +39,7 @@
 
 #include <rviz/display.h>
 #include <rviz/selection/selection_manager.h>
+#include <rviz/panel_dock_widget.h>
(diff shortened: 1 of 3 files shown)


moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
```diff
@@ -55,6 +56,7 @@
 #endif
 
 #include <moveit_msgs/DisplayTrajectory.h>
+#include <QDockWidget>
```
v4hn: I believe this is not needed, because now you use `rviz::PanelDockWidget` instead, right?
Could you please remove it from the request?

Jntzko: You're right, I removed it.

v4hn:
Thank you.
I believe this should be merged and ported to j/k.

PR #455, opened 2017-03-01
Fix usability bugs in RobotState display

v4hn:
I just tried to make use of this display for the first time
and now I keep wondering how people managed to use it in the first place...

Over here it segfaults when pressing Enter after changing the topic
and it does not connect to the topic when the display becomes enabled.

These patches address both problems.

Should be picked to jade.


v4hn:
ok, I missed that 4e72ac5f96863e7c37a47706645c6634d7672f7b was not backported to indigo.
I'll have another look into this and fixup the commits later.

v4hn:
ok, I failed to reproduce the segfault on a standard ubuntu trusty, so I'll remove it from this request.
I'll add a request for kinetic with both patches instead.

The missing subscribe/unsubscribe is relevant for indigo imo, so I leave the request open for this patch.

PR #471, opened 2017-03-16
address gcc6 build error (indigo-devel)

rojkov:
This is a continuation for https://github.com/ros-planning/moveit/pull/423.

Somehow I overlooked the moveit_ros/planning package which also needs the same treatment.


v4hn:
cherry-picked to jade-devel.

PR #482, opened 2017-04-06
Renaming of display doesn't change panel name

Jntzko:
Renaming the Motion Planning Display doesn't change the name of the Motion Planning Panel
right now.

These changes overrides the function setName of the display class, since we don't have a associated widget
the setName function from display doesn't work.

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
@@ -228,7 +228,7 @@ void MotionPlanningDisplay::onInitialize()
 
   if (window_context)
   {
-    frame_dock_ = window_context->addPane("Motion Planning", frame_);
+    frame_dock_ = window_context->addPane(getName(), frame_);
     connect(frame_dock_, SIGNAL(visibilityChanged(bool)), this, SLOT(motionPanelVisibilityChange(bool)));
     frame_dock_->setIcon(getIcon());
   }
@@ -303,6 +303,13 @@ void MotionPlanningDisplay::reset()
 
   query_robot_start_->setVisible(query_start_state_property_->getBool());
   query_robot_goal_->setVisible(query_goal_state_property_->getBool());
+}
+
+void MotionPlanningDisplay::setName(const QString& name)
+{
+  BoolProperty::setName(name);
+  frame_dock_->setWindowTitle(name);
+  frame_dock_->setObjectName(name);
 }
 
 void MotionPlanningDisplay::backgroundJobUpdate(moveit::tools::BackgroundProcessing::JobEvent, const std::string&)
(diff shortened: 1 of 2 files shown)


jrgnicho:
@Jntzko does this allow changing the Display name at runtime?  I'm not sure about what this change fixes.

Jntzko:
You can change the Display name at runtime with the rename button, this works already. What is missing is the renaming of the Panel. The Panel should get the same name as the Display, for example like the behavior of the Image Display and Panel.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
```diff
@@ -305,6 +305,13 @@ void MotionPlanningDisplay::reset()
   query_robot_goal_->setVisible(query_goal_state_property_->getBool());
 }
 
+void MotionPlanningDisplay::setName(const QString& name)
+{
+  BoolProperty::setName(name);
```
jrgnicho: Why is it necessary to call this static method?

Jntzko: This static method changes the name of the Display.

jrgnicho: This seems redundant given that you are already setting the window title and the underlying QTObject name in the lines that follow.

Jntzko: `BoolProperty::setName(name)` sets the Display name.
`setWindowTitle(name)` sets the Panel name or title.
I'm not sure what `setObjectName(name)` does, but the original code from rviz::display says in comments:
"QMainWindow::saveState() needs objectName to be set."

jrgnicho: @Jntzko  it looks like the reason why ```BoolProperty::setName(name)``` works is because the **MotionPlanningPanel** inherits from **rviz::Display** which itself inherits from **BoolProperty** so this is not a static method but a super class call.  The inherited ```setName()``` method is already implemented by the [rviz::Display](http://docs.ros.org/indigo/api/rviz/html/c++/display_8cpp_source.html#l00410) class and it already does call the same functions that you have added.  Since the **MotionPlanningPanel** already has inherits this method from rviz::Display then there should be no need to override it.

Jntzko: The problem is, that the MotionPlanning Panel is not a "associated widget", so the functions in the rviz::Display class will not change the name of the Panel but only the name of the Display.
There was an issue why the MotionPlanning Panel can't be a "associated widget". See here #442 and here #452.

jrgnicho: I see, makes sense to me.  
Could you cherry pick to jade and kinetic?

v4hn: @jrgnicho can I take that as your approval to merge+cherry-pick this as is?
Then I would directly do so.

jrgnicho: @v4hn I'll merge and do the cherry pick into jade and kinetic then

PR #499, opened 2017-05-08
checks that the joint_state field of the robot state message isn't empty

jrgnicho:
### Description

Addresses #496

### DIFF
--- a/moveit_core/robot_state/src/conversions.cpp
+++ b/moveit_core/robot_state/src/conversions.cpp
@@ -329,10 +329,20 @@ static void _msgToAttachedBody(const Transforms* tf, const moveit_msgs::Attached
 static bool _robotStateMsgToRobotStateHelper(const Transforms* tf, const moveit_msgs::RobotState& robot_state,
                                              RobotState& state, bool copy_attached_bodies)
 {
+  bool valid;
+  const moveit_msgs::RobotState& rs = robot_state;
+
+  if (rs.joint_state.name.empty() && rs.multi_dof_joint_state.joint_names.empty())
+  {
+    logError("Found empty JointState message");
+    return false;
+  }
+
   bool result1 = _jointStateToRobotState(robot_state.joint_state, state);
   bool result2 = _multiDOFJointsToRobotState(robot_state.multi_dof_joint_state, state, tf);
+  valid = result1 || result2;
 
-  if (copy_attached_bodies)
+  if (valid && copy_attached_bodies)
   {
     if (!robot_state.is_diff)
       state.clearAttachedBodies();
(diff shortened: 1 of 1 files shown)


v4hn:
format checking failed.

Returning `false` here breaks the logic in `_robotStateMsgToRobotStateHelper` though, doesn't it?

jrgnicho:
@v4hn The ```_jointStateToRobotState``` function already checked for another condition that can return false so I assume that it wouldn't break the logic in the calling function ```_robotStateMsgToRobotStateHelpe```

v4hn:
At the moment ```_robotStateMsgToRobotStateHelper``` accepts and successfully transforms a ```RobotState`` message if ```joint_state``` is empty, but ```multi_dof_joint_state``` specifies joints.
With your patch this case would return false, right?

How about we change ```_robotStateMsgToRobotStateHelper``` to call both conversion functions only if the size of their corresponding list of joint names is nonzero?

jrgnicho:
@v4hn I don't have a full understanding on the role of the ```multi_dof_joint_state``` field, however I thought that it was used only as a supplement to the ```joint_states``` field .  If a full robot state can be properly initialized from the ```multi_dof_joint_state``` only then I'm OK with implementing your suggested solution.

v4hn:
I'm not sure whether some other place limits the use-case I give above, but I see no reason to limit it in this part of the code..

You know we don't have many invariants or guarantees in the upper layers of MoveIt, so it's always hard to say how something *should* be interpreted...

It would be great if you could retain the current behavior for this case :thumbsup:

jrgnicho:
@v4hn I made the requested changes.

v4hn:
lgtm. Format checking failed though.

jrgnicho:
@v4hn the formatting checks out now.

v4hn:
Thanks a lot and sorry for all the nitpicking!

Keep the good patches coming :)

jrgnicho:
@v4hn thanks for reviewing.

Vonstafenberg:
@v4hn Hi,
I am not sure If I applied the fix in the commit 492e881 properly, since I am still getting the Found joint states empty message. I have Moveit installed in the opt. but for this fix I downloaded the moveit core package and compiled it in my workspace. I am working with pr2 moveit_tutorials planning_scene_ros_api_tutorial

Am I suppose to compile Moveit from source in my _ws and then it would be effective ..or how ?

please elaborate

v4hn:
The error message should *not* affect functionality of MoveIt! and should be ignored until someone had time to fix it.

Vonstafenberg:
@v4hn yes the error message is indeed not affecting the functionality of moveit. I actually have setup a moveit configuration for an arm with the Kinetic Moveit pkg. And the move plans are executed normally. But still these errors messages are displayed. I am actually more concerned about the issue mentioned here #552. the object gets attached to the arm gripper but then does not get detached. Same case for the pr2 tutorial mentioned in #552.

I will report this bug in Moveit tutorial asap.

PR #510, opened 2017-05-22
RViz - Trajectory trail step size

henhenhen:
Adds an integer parameter **Trail Step Size** to the section _Motion Planning_->_Planned Path_ in RViz to adjust the density of visualized trajectory trails.
Since very dense trajectory trails can be difficult to recognize this parameter enables an easy way to make them arbitrarily sparse.
This is especially useful for creating snapshots or for other demonstration purposes. 

By default the step size is 1 so that every trajectory point is displayed.
With higher numbers the intermittent steps are hidden except for the first and last trajectory points.

The following pictures show examples of the default setting and the trail step size set to 4.

dense_trail

sparse_trail

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
@@ -100,10 +100,6 @@ TrajectoryVisualization::TrajectoryVisualization(rviz::Property* widget, rviz::D
   trail_display_property_ =
       new rviz::BoolProperty("Show Trail", false, "Show a path trail", widget, SLOT(changedShowTrail()), this);
 
-  trail_step_size_property_ = new rviz::IntProperty("Trail Step Size", 1, "Specifies the step size of the samples "
-                                                                          "shown in the trajectory trail.",
-  trail_step_size_property_->setMin(1);
-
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
Seems like an ok idea.

Would probably be nice if you could move the new config item to be listed below the *Show Trail* item.

moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
```diff
@@ -100,6 +100,10 @@ TrajectoryVisualization::TrajectoryVisualization(rviz::Property* widget, rviz::D
   trail_display_property_ =
       new rviz::BoolProperty("Show Trail", false, "Show a path trail", widget, SLOT(changedShowTrail()), this);
 
+  trail_step_size_property_ = new rviz::IntProperty("Trail Step Size", 1, "Specifies the step size of the samples "
+                                                                          "shown in the trajectory trail.",
```
v4hn: You introduced a syntax error here with the cleanup commit.

henhenhen: Embarrassing... just fixed it!

v4hn:
I will take @gavanderhoorn's reply as confirmation and merge this now.
Thank you for implementing this @henhenhen !

PR #513, opened 2017-05-29
ikfast_kinematics_plugin: Fix for urdfdom1.x (Lunar)

mintar:
Don't needlessly cast away const.

Without this fix, compilation fails on platforms with urdfdom1.x (ROS Lunar on Ubuntu Yakkety, Ubuntu Zesty and Debian Stretch). This is because on urdfdom1.x, `urdf::LinkSharedPtr` etc. have switched from `boost::shared_ptr` to `std::shared_ptr`, which of course isn't compatible with `boost::const_pointer_cast`.

Unfortunately, all ikfast plugins have to be regenerated (using `update_ikfast_plugin.sh`) before being released into Lunar.

This fix is safe to merge into indigo, jade, kinetic and lunar.

Also see uos/katana_driver#22.

### DIFF
--- a/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
+++ b/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
@@ -363,7 +363,7 @@ bool IKFastKinematicsPlugin::initialize(const std::string& robot_description, co
 
   ROS_DEBUG_STREAM_NAMED("ikfast", "Reading joints and links from URDF");
 
-  urdf::LinkSharedPtr link = boost::const_pointer_cast<urdf::Link>(robot_model.getLink(getTipFrame()));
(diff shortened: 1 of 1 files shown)


v4hn:
Sorry for the delay @mintar I was on vacation.

I suspect the original reason for that cast might have been further below:
`link = link->getParent();` where `getParent` returns a `LinkSharedPtr`.
But implicitly casting that link const works fine, as your patch demonstrates..

Thanks for the patch!

v4hn:
cherry-picked to jade.

In indigo we don't use the `urdf::LinkSharedPtr` yet, so the patch doesn't apply.
I'm currently unsure whether there was a problem with some of these typedefs in indigo-urdf or whether we just didn't update these parts of the code. @de-vri-es any idea?

Unless somebody requests this, I would just leave this patch out for indigo.
We would like to have kinetic-devel compiling on indigo, but the other way around does not matter too much.

PR #515, opened 2017-05-29
ikfast_kinematics_plugin: Add c++11 compile option

mintar:
This is required for Kinetic.

### DIFF
--- a/moveit_kinematics/ikfast_kinematics_plugin/templates/CMakeLists.txt
+++ b/moveit_kinematics/ikfast_kinematics_plugin/templates/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(_PACKAGE_NAME_)
 
+add_compile_options(-std=c++11)
+
 find_package(catkin REQUIRED COMPONENTS
   moveit_core
   pluginlib


gavanderhoorn:
Thanks for the PR @mintar. 🍻

gavanderhoorn:
As this is a Kinetic-only change, no need to cherry-pick.

mintar:
> As this is a Kinetic-only change, no need to cherry-pick.

True. There would still be an advantage in cherry-picking it to indigo: If a user generates their ikfast package on indigo or jade, having this change would make their package "future-proof" when they try to port it to kinetic later on. It might not occur to them that they have to re-generate their ikfast package.

Up to you to decide.

gavanderhoorn:
Yes, I thought about that.

For some of 'my own' plugins I've made enabling C++11 conditional on which MoveIt version is being targetted (ros-industrial/fanuc_experimental@314e1f56191eae0a5c6b2d2081fb575498a208ef). That would do what you suggest I think.

I didn't want to propose that here, as on Kinetic you always need C++11, but hadn't considered people trying to build plugins generated on < Kinetic on Kinetic.

gavanderhoorn:
@ros-planning/moveit-developers: any opinions on whether we want to enable C++11 unconditionally in IKFast plugins on < Kinetic, or perhaps do it conditionally?

@mintar's comment about users not understanding they need to regenerate their plugins -- that 'worked fine' under < Kinetic -- is a valid one.

v4hn:
@mintar you forgot to bump the minimum cmake version to 2.8.12
`add_compile_options` was introduced there. (#521)

Let's cherry-pick this to indigo (together with the cleanup).

mintar:
Hmm, good point. How did you catch this?

v4hn:
> How did you catch this?

Quick arms. And a memory of this being the reason why we didn't have `add_compile_options` in indigo before. 13.10 was/is(???) targeted by indigo and didn't provide the required cmake version.

Meditating on this further: I remember there was some debate on whether 13.10, although deprecated, outdated, obsolete,..., is *still* targeted by indigo. There were some build-farm errors with OSRF that I recall. @130s do you remember this?
If this is still an issue then we can't cherry-pick this as-is.

mintar:
Yes, I remember this. That's why I made [this construction](https://github.com/uos/katana_driver/blob/7a68a7983c20fb25628fbb0a8f23fdd1704b5309/katana_gazebo_plugins/CMakeLists.txt#L35-L47) back then.

But 13.10 is EOL and no longer targeted by indigo. The [katana_moveit_ikfast_plugin with these changes](https://github.com/uos/katana_driver/blob/843b59619e42de37aa4b31cd0505188590cb7fcb/katana_moveit_ikfast_plugin/CMakeLists.txt#L4) produces no errors on the [buildfarm on Indigo](http://build.ros.org/job/Ibin_uT64__katana_moveit_ikfast_plugin__ubuntu_trusty_amd64__binary/).

v4hn:
Good to hear that :)

In this case: +1 to cherry-pick both commits to indigo.

Thanks a lot for your requests @mintar !

v4hn:
As discussed above I cherry-picked this to i/j. Sorry for the delay, this dropped of my list.

PR #521, opened 2017-06-07
adjust cmake_minimum_required for add_compile_options

v4hn:
#515 forgot about this.


v4hn:
this is just a cleanup, merging.

PR #523, opened 2017-06-12
Fix crash if no window_context exists

simonschmeisser:
#491 introduced a regression:

In TrajectoryVisualization::onInitialize the existance of a window_context is checked and the trajectory_slider_panel_ and trajectory_slider_dock_panel_ are only created if it exists.
However later on these two pointers are assumed to be valid (non-NULL) and not checked any further.

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
@@ -153,7 +153,8 @@ void TrajectoryVisualization::onInitialize(Ogre::SceneNode* scene_node, rviz::Di
 
 void TrajectoryVisualization::setName(const QString& name)
 {
-  trajectory_slider_dock_panel_->setWindowTitle(name + " - Slider");
+  if (trajectory_slider_dock_panel_)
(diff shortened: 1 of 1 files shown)


v4hn:
hey @simonschmeisser,

did you encounter this segfault or is this patch preventive?
I believe @Jntzko implemented this similar to other parts of the moveit plugins.
I'm waiting for his response first.

simonschmeisser:
@v4hn we do encounter a crash when it tries to set the title of the non-existent panel. We use several rviz::Display and one rviz::PropertyTreeWithHelp in our application but don't have a central rviz window and thus the window_context == nullptr

v4hn:
@davetcoleman, @130s please review/merge.
@Jntzko also found a similar situation in the motion planning rviz plugin and will probably create a similar pull request for it later today.

Jntzko:
The related pull request is #525.

130s:
I'm not sure how to reproduce the issue. Assuming from the comment https://github.com/ros-planning/moveit/pull/523#issuecomment-308098540 by @simonschmeisser, it can only be replicated from source code (I don't know on RViz GUI how to set a title on a non-existent panel)? So I leave the decision up to others.

v4hn:
merging. We want to have this regression-fix in the next release.

v4hn:
cherry-picked to i/j

PR #525, opened 2017-06-13
Test if frame_dock_ exists in case of no window_context

Jntzko:
This is a similar situation as in the pull request #523.

In the onInitialize function the frame_dock_ panel is only created if window_context exists.
Later on in the setName function the pointer is used without checking for existence.


v4hn:
@Jntzko format check failed. Could you clean this up please?

Jntzko:
Fixed it.

v4hn:
I just merged #523 and this is exactly the same context, just a different variable.
So I'll merge this one too.

v4hn:
picked to j/k

PR #527, opened 2017-06-14
[Future TODO] moveit_ros_planning: switch from TinyXML to TinyXML2

rojkov:
The library TinyXML is considered to be unmaintained and
since all future development is focused on TinyXML2 this
patch updates moveit_ros_planning to use TinyXML2.

depends on https://github.com/ros/robot_model/pull/205

### DIFF
--- a/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
+++ b/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
@@ -40,7 +40,7 @@
 #include <moveit/macros/class_forward.h>
 #include <urdf/model.h>
 #include <srdfdom/model.h>
-#include <tinyxml.h>
+#include <tinyxml2.h>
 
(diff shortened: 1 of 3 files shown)


BryceStevenWilley:
From some discussion on [related](https://github.com/ros/urdfdom/pull/99#issuecomment-375009915) [PRs](https://github.com/ros-planning/srdfdom/pull/32) (which this PR depends on), it seems like this PR should simply deprecate the TinyXml APIs instead of removing them. Quoting that discussion:

> The short of it is that I think it would be too painful to the whole ecosystem to just remove the tinyxml APIs. Thus, in urdf, we ended up closing ros/urdf#4 and instead merging ros/urdf#9, which adds the tinyxml2 APIs and puts a deprecation warning on the tinyxml APIs. I'd suggest that we do the same thing here, for the same reasons.

v4hn:
Related request: #1254

rhaschke:
Resolved by #1254.

PR #529, opened 2017-06-16
fix simple_controller_manager includes

v4hn:
simple_controller_manager: fix include order

The local header files of the packages should always be preferred.
Otherwise the same headers installed to /opt/ros/*/include would be used.

### DIFF
--- a/moveit_plugins/moveit_simple_controller_manager/CMakeLists.txt
+++ b/moveit_plugins/moveit_simple_controller_manager/CMakeLists.txt
@@ -6,8 +6,6 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
 endif()
 
 find_package(Boost REQUIRED thread)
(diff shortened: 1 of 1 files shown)


v4hn:
This just rearranges includes to make sure the local headers are preferred over headers installed in /opt/ros. I'll merge and cherry-pick this now.

moveit_plugins/moveit_simple_controller_manager/CMakeLists.txt
```diff
@@ -26,8 +29,6 @@ catkin_package(
   CATKIN_DEPENDS moveit_core actionlib control_msgs
     )
 
-include_directories(include)
```
130s: Just assuming the key of this PR is to move this line to somewhere before `catkin_package`?

v4hn: No, the key is to move the "include" include directive above the "${catkin_INCLUDE_DIRS}".

The rest was just moving include directives together.

PR #532, opened 2017-06-16
setup assistant: add use_gui param to demo.launch

v4hn:
It bugged me for quite some time now that one has to edit the launch file
just to be able to move the "real" robot around in demo mode.
Thus I want to expose the option as a parameter in the launch file.

### DIFF
--- a/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo.launch
+++ b/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo.launch
@@ -8,6 +8,16 @@
   <!-- By default, we are not in debug mode -->
   <arg name="debug" default="false" />
 
+  <!--
+  By default, hide joint_state_publisher's GUI
+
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
In principle adding the argument would be ok, but would it be an idea to add some comments clarifying why this argument is there?

Reason is, I see a lot of 'moveit launch files' that are basically copy-pastes of `demo.launch` but then with 'real' robot drivers added to them. ROS Answers has quite some questions specifically about how JSP, RSP and robot drivers all relate to each other ("who is publishing what, and why?"), and if we then modify this launch file to accept an argument without explaining that it's basically only useful in this particular demo launch file I think we might run the risk of confusing ppl even more.

v4hn:
@gavanderhoorn I added a longer explanation. Please have a look.

130s:
Ported to indigo-devel https://github.com/ros-planning/moveit/commit/ba0b341a2b38fb6bb81e9ce66957076657c97b13

Issue #511, opened 2017-05-22
MSA: consider using `xacro --in-order` on Indigo as well?


gavanderhoorn:
`xacro` supports Jade+-like parsing of xacros on Indigo, but the Setup Assistant doesn't use this. This leads to the situation where xacros that load perfectly fine when `xacro` is invoked / used stand-alone to not load in the MSA.

Would it be an idea to have Indigo MSA add the `--in-order` argument ([here](https://github.com/ros-planning/moveit/blob/fb2ebe3826e8d2b15dc99188343fc0cf2b3dc939/moveit_setup_assistant/src/widgets/start_screen_widget.cpp#L508)) so that we get Jade+ xacro parsing? Should this be an option (ie: check box)?

v4hn:
On Mon, May 22, 2017 at 12:26:58PM -0700, G.A. vd. Hoorn wrote:
> `xacro` supports Jade+-like parsing of xacros on Indigo, but the Setup Assistant doesn't use this. This leads to the situation where xacros that load perfectly fine when `xacro` is invoked / used stand-alone to not load in the MSA.
> 
> Would it be an idea to have Indigo MSA add the `--in-order` argument ([here](https://github.com/ros-planning/moveit/blob/fb2ebe3826e8d2b15dc99188343fc0cf2b3dc939/moveit_setup_assistant/src/widgets/start_screen_widget.cpp#L508)) so that we get Jade+ xacro parsing? Should this be an option (ie: check box)?

I believe changing the default in indigo in not an option because in-order parsing can also break xacros that work well in current indigo.
That being said, I would support a check-box and/or an item in `.setup_assistant` to address this.

gavanderhoorn:
A checkbox is a good idea. I'll see if I can implement that.

gavanderhoorn:
Although #540 is merged, I'll keep this open as the command line interface to the MSA needs to be checked for support of this.

ipa-fxm:
Though the new checkbox works great when `Create New Moveit Configuration Package` is selected, unfortunately, it fails to load the xacro (which uses Jade+ features) when `Edit Existing Moveit Configuration Package` is selected.

gavanderhoorn:
@ipa-fxm: see the discussion on the #540 for why we didn't add the checkbox to the *Edit existing package* UI.

bmagyar:
+1

gavanderhoorn:
@bmagyar: what are you `+1`-ing specifically? :)

[...]

PR #540, opened 2017-06-27
MSA: support loading xacros that use Jade+ extensions on Indigo

gavanderhoorn:
This adds a checkbox to the Setup Assistant urdf loading widget on the start screen that allows users to enable 'Jade+ extensions' in xacro on Indigo installations. This is needed to support loading xacros that make use of those extensions.

The choice is persisted (`.setup_assistant`) so loading a config pkg that was created with the option enabled will enable them automatically. Bw-compatibility with existing config pkgs is maintained, as absence of the setting will just make the MSA use the default (`false`).

Extensions are also enabled in `planning_context.launch` if needed.

Screenshot of UI change:

msa_enable_jade_xacro_sshot

Ref #511.

As xacro on Jade and newer ROS releases enables those extensions by default, these changes do not need to be cherry-picked to any other MoveIt versions.

### DIFF
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -258,6 +258,7 @@ bool MoveItConfigData::outputSetupAssistantFile(const std::string& file_path)
   emitter << YAML::Value << YAML::BeginMap;
   emitter << YAML::Key << "package" << YAML::Value << urdf_pkg_name_;
   emitter << YAML::Key << "relative_path" << YAML::Value << urdf_pkg_relative_path_;
+  emitter << YAML::Key << "use_jade_xacro" << YAML::Value << urdf_requires_jade_xacro_;
   emitter << YAML::EndMap;
 
   /// SRDF Path Location
@@ -925,8 +926,8 @@ bool MoveItConfigData::inputSetupAssistantYAML(const std::string& file_path)
     YAML::Node doc;
     loadYaml(input_stream, doc);
 
-    yaml_node_t title_node, urdf_node, package_node, srdf_node, relative_node, config_node, timestamp_node,
-        author_name_node, author_email_node;
+    yaml_node_t title_node, urdf_node, package_node, jade_xacro_node, srdf_node, relative_node, config_node,
+        timestamp_node, author_name_node, author_email_node;
 
     // Get title node
     if (title_node = findValue(doc, "moveit_setup_assistant_config"))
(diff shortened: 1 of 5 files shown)


gavanderhoorn:
Exact text used on the checkbox is obviously something we can discuss.

Also: I was unsure whether we would want to add the checkbox to the 'load existing package' UI as well, as technically it's impossible to generate a configuration that needs the extensions without enabling them in the first place -- unless that configuration was modified manually.

If we want to support that, I can add the checkbox to that part of the UI as well.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -258,6 +258,7 @@ bool MoveItConfigData::outputSetupAssistantFile(const std::string& file_path)
   emitter << YAML::Value << YAML::BeginMap;
   emitter << YAML::Key << "package" << YAML::Value << urdf_pkg_name_;
   emitter << YAML::Key << "relative_path" << YAML::Value << urdf_pkg_relative_path_;
+  emitter << YAML::Key << "use_jade_xacro" << YAML::Value << urdf_requires_jade_xacro_;
```
v4hn: does it make sense to add this only if it is set to true?

gavanderhoorn: That occured to me as well.

But I couldn't really make up my mind as I don't see any obvious problems with including it even if it is `false`. Only thing I could come up with was it might 'invite' users to change it to `true` after the package is generated, expecting the rest of the infrastructure to automagically cope with Jade+ xacros as well - which is obviously not going to happen.

tl;dr: I don't really mind either way, so if anyone has a strong preference for not serialising in case the property is `false`, I can change it.

gavanderhoorn:
@v4hn wrote:
> @gavanderhoorn  what is *your* use-case for this?

with the PR you mean? I have a nr of installations still running Indigo, but do use Jade+ xacro on that. See #511. I'd like those MSAs to be able to load those xacros. See also ros-industrial/fanuc#217.

> would you say it is useful to have the checkbox for externally modified setup?

If you were asking whether I could think of a use-case for having the checkbox on the 'load existing package' UI as well: I can think of one, but I'm not sure whether we want to support that. It would basically be the case where an existing package is modified to make use of some Jade+-features including xacro and now has to be updated. Having the checkbox there would make it possible to do that through the UI. Without it, users would have to edit the `.setup_assistant` file.

I don't mind editing a text file, and if you're *this deep* into things anyway, it shouldn't be that much effort.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
v4hn: well, if you include it on false, you might end up with craft after the setups move to kinetic+.
I vote for writing it only on true.

gavanderhoorn: Agreed. Fixed.

gavanderhoorn:
I added a tooltip that gives some info on why and when users would want to use this.

I'll wait for Travis to finish and then merge.

gavanderhoorn:
Thanks for the reviews @v4hn and @davetcoleman.

PR #547, opened 2017-07-13
Fixing segfault due to missing string format parameter.

henhenhen:
This fixes a segfault due to missing string format parameter in logWarn call.

The message occurs when requesting the pose of a collision object that consists of multiple shapes and therefore contains multiple poses. The inserted parameter is now the id descriptor of the object.


v4hn:
This was a stupid typo apparently.
Thanks for the fix!

picked to i/j.

Issue #512, opened 2017-05-23
rviz: displaying attached-object(s) is wrongly triggered by [Show Robot Visual] / [Show Robot Collision]


svenkrause:
### Description
Let
 - V = Show Robot Visual
 - C = Show Robot Collision

be the fields in the rviz-PlanningScenePlugin under "Scene Robot".

I get the following faulty behaviour in rviz. If only one of V & C is "true" and the other is "false", and I set the "true" one to "false" as well, I get:
 - (V, !C) --> (!V, !C): robot is (correctly) not displayed, but attached object (incorrectly) is
 - (!V, C) --> (!V, !C): robot is (correctly) not displayed, but attached object (incorrectly) is

From this state, if I set either V or C to "true" again, I get:
 - (!V, !C) --> (V, !C): robot is (correctly) displayed, but attached object (incorrectly) is not
 - (!V, !C) --> (!V, C): robot is (correctly) displayed, but attached object (incorrectly) is not

In other words, the attached object is visible (invisible) whenever the robot is invisible (visible).

On the other hand, if I then set the other one to "true" as well, the attached object shows up again
 - (V, !C) --> (V, C): robot and attached object are (correctly) both displayed
 - (!V, C) --> (V, C): robot and attached object are (correctly) both displayed

and setting either one of these to "false" again afterwards does not change that:
 - (V, C) --> (V, !C): robot and attached object are (correctly) both displayed
 - (V, C) --> (!V, C): robot and attached object are (correctly) both displayed

### Your environment
* ROS Distro: Kinetic
* OS Version: Ubuntu 16.04

### Steps to reproduce
1) create a PlanningScene with an attached object
2) display the scene in rviz
3) toggle "Show Robot Visual" / "Show Robot Collision"

### Expected behaviour
The attached object should be visible (invisible), whenever the robot is visible (invisible).

### Actual behaviour
Toggling off one of "Visual" / "Collision" while the other is off already, creates the opposite behaviour: The attached object is visible (invisible), whenever the robot is invisible (visible).

v4hn:
sorry for the delay.

This looks like a reasonably simple bug, did you try to find it yourself?

svenkrause:
I had a quick look in the code online, but couldn't find it. I'd have to download the moveit-source and debug it, but I'm short on time at the moment, so I thought, maybe someone who contributed to that part of the code might have an idea, where the bug might lie. If not, I'll investigate further at some point in the (relatively near) future.

[...]

Issue #552, opened 2017-07-18
/attached_collision_object endless loop


zplizzi:
### Description

I am trying to attach an object to a robot arm. I send a message over the `/attached_collision_object` topic. As a result, the object begins being published in the `robot_state: attached_collision_objects:` section of the `/move_group/monitored_planning_scene`, with diff = False. The object shows up in Rviz as expected. However, in the console where I launched my robot launch file (with MoveIt, Rviz, etc), I get an endless stream of messages of the type 

`[ INFO] [1500398010.982460498]: Attached object 'test' to link 'fingertip_link'`

It seems to me that Rviz is trying to apply every message from `/move_group/monitored_planning_scene` as a new attached object, vs recognizing that the object already exists. 

As a result (I think?), collision detection during path planning on that object is broken since the object is deleted faster than the planner can plan a trajectory.

### Your environment
* ROS Distro: Kinetic
* OS Version: e.g. Ubuntu 16.04
* Binary: 1.3.1-0xenial-20170615-105244-0800

zplizzi:
I think I understand the problem a bit better. When adding a normal collision object (unattached), there will be a single message on `/move_group/monitored_planning_scene` containing the new object, and diff=True. All future messages also have diff=True, and don't mention the object. This appears to work fine, as it should.

However, when adding an attached collision object, the object is published in every future `/move_group/monitored_planning_scene` message, with diff=False in the `robot_state` part of that message. Then, apparently whoever is listening to that message (Rviz plugin?) doesn't handle the diff=False properly, and re-creates the object each message.

So, it seems like there's two approaches to a solution: either a) treat attached objects like the unattached, and only publish them once with diff=True, or b) fix the Rviz plugin to handle diff=False messages properly.

v4hn:
First off, thanks for looking into this. 👍 

At first glance I would think publishing the robot state *always* as `is_diff: false` is a bug in this case, but I didn't investigate this thoroughly.
It could be that some other code behaves wrongly if it receives a "diff" RobotState and/or that it is non-trivial to publish the diffs for the RobotState class.

Still, my preferred solution would work towards your proposal "a)".

> As a result (I think?), collision detection during path planning on that object is broken since the object is deleted faster than the planner can plan a trajectory.

The planner operates on a locked planning scene, so this is most likely not the case.

zplizzi:
[...]

PR #565, opened 2017-07-28
remove render_shapes for attached objects when the RobotStateDispla gets disabled

simonschmeisser:
this fixes a crash on the first update after reenabling the plugin:
OGRE EXCEPTION(5:ItemIdentityException): SceneNode 'rviz' not found. in SceneManager::getSceneNode at /build/ogre-1.9-mqY1wq/ogre-1.9-1.9.0+dfsg1/OgreMain/src/OgreSceneManager.cpp (line 920)

### Description

You can trigger the crash in rviz (or any other application loading the RobotState or PlanningScene rviz plugin) by attaching an object and then toggling the "enable" checkbox. It will crash on the first RobotState update after reenabling when it calls render_shapes_.clear(). Apparently some more nodes get deleted/removed once the plugin is disabled and it subsequently does not find them.

Please note that
 * I don't understand the issue fully but this fixes the crash and does not seem to have any negative side effects
 * That region of the code should receive some more love
 * There is a similar (but purely visual) issue that the attached object does not disappear if I disable  both the visual as well as the collision model of the robot (https://github.com/ros-planning/moveit/issues/512)

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/robot_state_visualization.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/robot_state_visualization.cpp
@@ -139,6 +139,8 @@ void RobotStateVisualization::setVisible(bool visible)
 {
   visible_ = visible;
   robot_.setVisible(visible);
+  if (!visible)
+    render_shapes_->clear();
 }
 
 void RobotStateVisualization::setVisualVisible(bool visible)


v4hn:
At first glance, I'm unsure about this request.
Does this require that each received `RobotState` contains all attached objects (i.e. `is_diff == false`)?
I want to get rid of https://github.com/ros-planning/moveit/issues/552 eventually, so we might publish diff RobotState messages in the future.

simonschmeisser:
Well, this handles the case were you disable the whole plugin (the main checkbox). This will also stop the planning scene monitor and generally not receive updates. 
If you don't touch the main checkbox however (do not disable the plugin), this won't have any effects and being more efficient wrt #552 is still possible. I think for planning scenes there is some mechanism to request a "full update", maybe something similar is needed for the robot state/attached objects? or is this included in the planning scene update?

v4hn:
> Well, this handles the case were you disable the whole plugin (the main checkbox). This will also stop the planning scene monitor and generally not receive updates.

fair enough.

> or is this included in the planning scene update?

yes it is.

v4hn:
> Well, this handles the case were you disable the whole plugin (the main checkbox).

This is true for the `planning_scene_rviz_plugin` and the `robot_state_rviz_plugin`, but the start/goal state visualization in the `motion_planning_rviz_plugin` trigger this function too.
I'm not sure this affects them though.

Still, it feels weird to remove additional rendering shapes just because the main object is set to invisible...

simonschmeisser:
Ah, we don't use these. Toggling these does not seem to be affect by the crash in the first place.  After applying the patch they still show the attached object (after the first update ...)

My (limited) understanding is that there is a hierarchy of SceneNodes, the attached objects being some leaf of that tree(/forest?). Now disabling the plugin ("visible_" = false) will remove some node called "rviz". This leads to the attached objects SceneNodes dangling around somewhere and subsequently an exception once they get deleted during the next update. Catching (and ignoring) that exception leads to crashes elsewhere and is not really a solution anyway as then the render shapes will still be around.

v4hn:
@simonschmeisser Now I feel like I stared into the Abyss for too long, but I believe I managed to track down the underlying issue. Please have a look at #572 for a (hopefully better) fix.

simonschmeisser:
@v4hn your version works as well and looks cleaner, will close this one tomorrow

v4hn:
closed in favor of #572

PR #569, opened 2017-08-02
remove references to TinyXML

mikaelarguedas:
### Description

I noticed that several packages were using tinyxml symbols but were not explictely fingind it or linking against it. I modified the package.xml and CMakeLists of the relevant packages and hopefully didnt miss any.

### DIFF
--- a/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
+++ b/moveit_ros/planning/rdf_loader/include/moveit/rdf_loader/rdf_loader.h
@@ -40,7 +40,8 @@
 #include <moveit/macros/class_forward.h>
 #include <urdf/model.h>
 #include <srdfdom/model.h>
-#include <tinyxml.h>
+
+class TiXmlDocument;
 
 namespace rdf_loader
 {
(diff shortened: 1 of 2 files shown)


v4hn:
Thank you for reporting this. Missing dependencies are always a hassle..
I find the TinyXML dependency exposed through the `urdf` package is badly thought out and should have been a forward declaration of `TiXMLElement` in the first place. Let's hope things get better once tinyxml2 is used...

Afaics,

- `moveit_setup_assistant` only uses tinyxml [here](https://github.com/ros/robot_model/pull/213) and there is no reference to `Ti*` structures in that file at all and removing the include does not break the build, so I guess this is obsolete and should be removed.

- `rdf_loader` and `robot_model_loader` only refer to `TiXmlDocument*` and a forward definition of the class should be fine to deal with that.

So instead of declaring it a full dependency, what do you think about removing the tinyxml includes in these packages instead to get rid of the dependency?

related to https://github.com/ros/robot_model/pull/213

mikaelarguedas:
Both very good points, I updated the PR accordingly

v4hn:
+1 I approve. In my opinion we should cherry-pick this to indigo too.

davetcoleman:
I'm confused with Github right now... it appears in your individual commits you made a lot of changes, but in the "Files Changed" tab it looks like only two changes:

screenshot from 2017-08-15 08-41-40

Are these the correct PR changes?

mikaelarguedas:
Oh yeah it's confusing indeed. I first added the tinyxml dependecy in every cmake and oackage.xml that referenced it. Based on @v4hn feedback I actually went the other direction and got rid of all unused references to tinyXML. I can squash those commits together if you want

mikaelarguedas:
But answer is yes, these are the expected changes

mikaelarguedas:
@davetcoleman squashed the commits into a single one

PR #572, opened 2017-08-03
RobotStateVisualization: clear before load to avoid segfault

v4hn:
This addresses the segfault reported by @simonschmeisser in https://github.com/ros-planning/moveit/pull/565 .

rviz::Robot deletes its complete SceneNode structure in the `load()` method.
However, `RobotStateVisualization::render_shapes_` keeps raw pointers
to some of these nodes (the attached objects), so these should be cleared
to avoid broken pointers.

Additionally the order of clearing was bad: the attached objects should
be removed first, and `rviz::Robot` only afterwards to avoid similar problems.

Should be cherry-picked to `indigo-devel`.


simonschmeisser:
@v4hn this fixes the segfault as well and solves the problem at its root so lgtm

thanks!

130s:
Merging since confirmed by @simonschmeisser  that this fixes a targeted issue.

130s:
Ported to indigo-devel https://github.com/ros-planning/moveit/commit/0be0417f602aa1bce69cbff3458aec97e435cb6c

Issue #568, opened 2017-08-02
OMPL with LBKPIECE1 crashes frequently after planning request


christian-rauch:
### Description

MoveIt crashes sporadically during planning using OMPL with LBKPIECE1.
I am using the `move_group.launch` launch file from the ([moveit plugin](https://github.com/christian-rauch/kuka_lwr_sdh2_moveit)) generated by the setup assistant to start the `move_group`. I am then using the rviz plugin to set the end-effector poses and plan the joint trajectory with OMPL.

The crash also occurs when just using the `demo.launch` with the default planner (LBKPIECE1).
A workaround is currently to use e.g. LBKPIECEkConfigDefault.

### Your environment
* Ubuntu 14.04, indigo
* binary, build: 0.7.11

### Steps to reproduce
I can reproduce the same backtrace by launching the `demo.launch` and leaving the planner unspecified. The segfault occurs after hitting `Plan` a couple of times (less than 20) without moving the end-effector.

### Expected behaviour
valid joint trajectory

### Actual behaviour
SIGSEGV

### Backtrace or Console output

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffa37fe700 (LWP 9865)]
0x00007fffc9f82c10 in ompl::Grid<ompl::geometric::Discretization<ompl::geometric::LBKPIECE1::Motion>::CellData*>::neighbors(std::vector<int, std::allocator<int> >&, std::vector<ompl::Grid<ompl::geometric::Discretization<ompl::geometric::LBKPIECE1::Motion>::CellData*>::Cell*, std::allocator<ompl::Grid<ompl::geometric::Discretization<ompl::geometric::LBKPIECE1::Motion>::CellData*>::Cell*> >&) const
    () from /opt/ros/indigo/lib/x86_64-linux-gnu/libompl.so.10

#0  0x00007fffc9f82c10 in ompl::Grid<ompl::geometric::Discretization<ompl::geometric::LBKPIECE1::Motion>::CellData*>::neighbors(std::vector<int, std::allocator<int> >&, std::vector<ompl::Grid<ompl::geometric::Discretization<ompl::geometric::LBKPIECE1::Motion>::CellData*>::Cell*, std::allocator<ompl::Grid<ompl::geometric::Discretization<ompl::geometric::LBKPIECE1::Motion>::CellData*>::Cell*> >&) const () from /opt/ros/indigo/lib/x86_64-linux-gnu/libompl.so.10
[...]

PR #573, opened 2017-08-04
set default planner by default in indigo's MoveGroup

v4hn:
#568 reminded me of this problem, and somewhat increased the pressure to change this...

This patch makes the `MoveGroup` interface behave the same as the RViz plugin
when the user specified a `default_planner_config` in their `ompl_planning.yaml`

Before, the RViz plugin automatically selected the default planner automatically, but the MoveGroup left the planner id unspecified in the MoveGroup action goals. The move_group node received the request with the unspecified field but didn't respect the parameter in indigo and uses `LBKPIECE1` even if the user specified a different default planner for the group.

To circumvent the old default planner `LBKPIECE1` in indigo, the user has to set the default planner **explicitly for each `MoveGroup`** using

```
group.setPlannerId(group.getDefaultPlannerId(group.getName()));
```

If they forget it for one instance, all planning requests from this instance will use the LBKPIECE1 planner config, usually resulting in rather complicated trajectories and apparently in some cases in weird segfaults.

Note that `move_group` respects the selected default in kinetic-devel and chooses that one whenever the planner id is not specified in a request. Thus, this patch is pretty redundant there.
Personally, I don't see the need to pick this to kinetic, but I don't oppose it either. I'll leave the choice to the person who merges this request.

Lastly, let me point out that this could also be addressed in `indigo-devel`'s `move_group` node. I preferred to leave the semantics of an empty "planner_id" field in a MoveGroup untouched in the LTS release, especially because we already broke with it in the `kinetic-devel` branch.

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group.cpp
@@ -131,6 +131,8 @@ public:
       end_effector_link_ = joint_model_group_->getLinkModelNames().back();
     pose_reference_frame_ = getRobotModel()->getModelFrame();
 
+    setPlannerId(getDefaultPlannerId(opt.group_name_));
+
     trajectory_event_publisher_ = node_handle_.advertise<std_msgs::String>(
         trajectory_execution_manager::TrajectoryExecutionManager::EXECUTION_EVENT_TOPIC, 1, false);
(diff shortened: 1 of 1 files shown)


jrgnicho:
+1 from me.  I don't see the need to merge in kinetic if it already behaves as it should.

PR #574, opened 2017-08-07
[indigo-devel] Chomp use PlanningScene (#546) from kinetic-devel

130s:
As per subject. #546


v4hn:
Although I believe this will not be overly useful in indigo, I see no reason why we shouldn't backport. Merging.

Issue #577, opened 2017-08-08
[MoveGroupInterface] Inconsistent usage of end-effector links between setJointValueTarget, setPoseTarget, and computeCartesianPath


hartmanndennis:
### Description

I use IKfast computed from base_link to tool0. The move group is defined in my SRDF like this `<chain base_link="base_link" tip_link="tool0" />`.
If I use setEndEffectorLink of MoveGroupInterface to my end-effector link(different from tool0), setJointValueTarget, setPoseTarget, and computeCartesianPath(all called with a Pose) behave inconsistently.

### Your environment
* ROS Distro: Kinetic
* Ubuntu 16.04
* kinetic-devel

### Steps to reproduce
Use IKFast with kinematic chain to tool0. Set end-effector by calling setEndEffectorLink to something different than tool0. Call setJointValueTarget, setPoseTarget, and computeCartesianPath with pose goals.

### Expected behaviour
The end-effector set by setEndEffectorLink should end up at the specified pose.

### Actual behaviour
setJointValueTarget works as expected.
setPoseTarget, and computeCartesianPath ignore the provided end-effector and instead use tool0.

setJointValueTarget uses RobotState's setFromIK to get the IK and provides the end-effector link as "tip_in" and transforms the pose to solver_tip_frame. This is expected behaviour.

computeCartesianPath calls the MoveGroupCartesianPathService but does not set the field link_name of the request. The service then sets link_name `link_name = jmg->getLinkModelNames().back();` at https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/move_group/src/default_capabilities/cartesian_path_service_capability.cpp#L86
to the last link of the kinematic chain, i.e. tool0. It then uses setFromIK with the wrong link as "tip_in".
If you add `req.link_name = getEndEffectorLink();` at https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L944 the end-effector link will be correctly used.
A change to this could break user code, if they use a fix like suggested in https://groups.google.com/forum/#!topic/moveit-users/VB7VWsgV9-g and set their end-effector to something different than their IK tip frame.

setPoseTarget sets position and orientation goal constraints and correctly adds the end-effector link name to these constraits. I couldn't figure out where the IK is computed when using position and orientation constraints, so I have no idea why this doesn't work as expected.

v4hn:
[...]

PR #580, opened 2017-08-09
MoveGroupInterface: Fixed computeCartesianPath to use selected end-effector.

hartmanndennis:
### Description
Currently computeCartesianPath of MoveGroupInterface ignores the end-effector link set by setEndEffectorLink. It uses the CartesianPathService which has a field link_name to provide this utility.
After this fix the provided poses are correctly transformed to the tip of the ik chain.

Like I wrote in #577, this could potentially break workaround user code, if they manually transform poses to the ik tip before giving them to computeCartesianPath and also calling setEndEffectorLink with their actual intended end-effector link. If their end-effector set by setEndEffectorLink or setEndEffector is the same as the ik chain tip, nothing will change.

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
@@ -941,6 +941,7 @@ public:
     req.jump_threshold = jump_threshold;
     req.path_constraints = path_constraints;
     req.avoid_collisions = avoid_collisions;
+    req.link_name = getEndEffectorLink();
 
     if (cartesian_path_service_.call(req, res))
     {


v4hn:
> Like I wrote in #577, this could potentially break workaround user code, if they manually transform poses to the ik tip before giving them to computeCartesianPath and also calling setEndEffectorLink with their actual intended end-effector link

In my very humble opinion the scenario you describe is a broken workaround and we shouldn't aim to support it. If users noticed the eef does not change, they should either (1.) have fixed it or (2.) not use the broken use-case, but rely on the working default.

The patch is simple enough and was an oversight when the code was originally written.

I approve to merge this into kinetic.
LTS Indigo is a different situation though. Unless multiple users speak up I would not port it there.

hartmanndennis:
I agree and I am in favor for a port to indigo.

bmagyar:
:+1: 

I am also in favor for porting to Indigo as I would even go as far as considering this a bugfix.
Shall I merge it?

v4hn:
>  Shall I merge it?

You are entitled to do so since yesterday and I already approved.
Whether you actually do it is up to you. If something breaks for a demo in some lab on the other side of the globe, you're to blame. :)


As both of you agree to port this back to indigo, I'm ok with that too.
So feel free to cherry-pick to indigo.

PR #585, opened 2017-08-10
IKFastTemplate: searchPositionIK now returns collision-free solution which is nearest to seed state.

### LINKED ISSUE #578: [ikfast_template] searchPositionIK doesn't consider other solutions if the best solution returned by getPositionIK is in collision
### Description

#537 fixed the ikfast plugin so that getPositionIK now returns the nearest solution to the seed state instead of the first solution returned by IKFast. 
Most moveit components call searchPositionIK which has an constraint callback parameter which getPositionIK lacks.
If the kinematics of your robot has no free parameters, searchPositionIK just calls getPositionIK and calls the constraint callback for this single solution. If the constraint fails, the function returns with no solution found even if there are other valid solutions which are not nearest to the seed.
To fix this searchPositionIK could instead call the getPositionIK version which returns every solution, sort them by distance and check the constraints in ascending order until it finds a solution which meet the constraints.

hartmanndennis:
Fixes #578


moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -780,6 +780,18 @@ bool IKFastKinematicsPlugin::searchPositionIK(const geometry_msgs::Pose& ik_pose
...
+{
+  std::vector<double> value;
+  double dist_from_seed;
+
+  bool operator<(const LimitObeyingSol& a) const
+  {
+    return dist_from_seed < a.dist_from_seed;
+  }
+};
+
```
v4hn: Please move this struct to the top of the file where the other auxiliary structures reside.

v4hn:
I very much like this patch. The plugin should definitely prefer a non-colliding solution over the first one it found.

https://github.com/ros-planning/moveit/pull/537#issuecomment-320294696 still stands though, although this request is somewhat independent.
Does the current template leave a way to request "just a solution" without dynamic memory management?

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
hartmanndennis: done

hartmanndennis:
Do you mean std::vector and push_back? IKFast itself uses that.

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -797,31 +809,53 @@ bool IKFastKinematicsPlugin::searchPositionIK(const geometry_msgs::Pose& ik_pose
   {
     ROS_DEBUG_STREAM_NAMED("ikfast", "No need to search since no free params/redundant joints");
 
-    // Find first IK solution, within joint limits
-    if (!getPositionIK(ik_pose, ik_seed_state, solution, error_code))
+    std::vector<geometry_msgs::Pose> ik_poses;
+    ik_poses.push_back(ik_pose);
```
v4hn: nitpick: `std::vector<geometry_msgs::Pose> ik_poses(1, ik_pose)` is shorter and could (in theory) optimize better.

hartmanndennis: done :)

v4hn:
@hartmanndennis Format checking failed due to a whitespace error.

v4hn:
format checking succeeded, awaiting second review.

Do we want this in indigo too?

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -797,31 +809,52 @@ bool IKFastKinematicsPlugin::searchPositionIK(const geometry_msgs::Pose& ik_pose
...
+    std::sort(solutions_obey_limits.begin(), solutions_obey_limits.end());
+
+    // check for collisions if a callback is provided
+    if (!solution_callback.empty())
+    {
+      for (std::size_t i = 0; i < solutions_obey_limits.size(); ++i)
       {
-        ROS_DEBUG_STREAM_NAMED("ikfast", "Solution has error code " << error_code);
-        return false;
+        solution_callback(ik_pose, solutions_obey_limits[i].value, error_code);
```
jrgnicho: Currently you have one for-loop for computing the distance to the seed and another one for collision checks.  Would it not be more efficient to have a single for-loop that checks for collisions first and adds the solution to `solutions_obey_limits` only if it passes the collision test?

v4hn: Wouldn't that add more collision tests than necessary?
Collision tests can be insanely costly, so if we only have to run the check for the closest solution(s) instead of always testing all possible solutions, that should be much more efficient.
Or maybe I misunderstood something here?

jrgnicho: @v4hn  I see, I overlooked that.  This should be good to merge then.

v4hn:
Ok, merging. This leaves the question:

> Do we want this in indigo too?

In my opinion [this comment](https://github.com/ros-planning/moveit/pull/575#issuecomment-322768280) applies here to.

jrgnicho:
I think we should create the cherry-pick PR to indigo and make a note that issue #586 needs to be resolved first.

v4hn:
Sure, feel free to go ahead. :-)

130s:
I can't find a cherrypick PR to indigo for this?

PR #589, opened 2017-08-14
robotStateMsgToRobotState: is_diff==true => not empty

### LINKED ISSUE #579: `Found empty JointState message` ROS_ERROR output with Kinetic
### Description

It appears there is a regression in Kinetic that leads to `Found empty JointState message` ROS_ERROR output getting spammed to the terminal. See also ROS Answers: https://answers.ros.org/question/266274/moveit-regression-found-empty-jointstate-message
It appears for most (all?) people this does not have adverse effects. However, there is at least one regression related to joint states in Kinetic that might be related (#501).

### Your environment
* ROS Distro: Kinetic
* OS Version: Ubuntu 16.04
* Binary build
* 0.9.8-0xenial-20170628-083013-0800

### Steps to reproduce
Start robot in Gazebo simulation, use rviz Motion Planning plugin to plan and execute a motion.

### Expected behaviour
Planning should proceed without errors.

### Actual behaviour
Multiple ROS_ERROR messages are printed out as in the console output below.

> [ INFO] [1502218727.220033766, 86.202000000]: Combined planning and execution request received for MoveGroup action. Forwarding to planning and execution pipeline.
> [ INFO] [1502218727.220192486, 86.202000000]: Planning attempt 1 of at most 1
> [ERROR] [1502218727.220300333, 86.202000000]: Found empty JointState message
> [ERROR] [1502218727.220377697, 86.202000000]: Found empty JointState message
> [ERROR] [1502218727.223660471, 86.204000000]: Found empty JointState message
> [ERROR] [1502218727.229793354, 86.204000000]: Found empty JointState message
> [ INFO] [1502218727.230916230, 86.204000000]: Planner configuration 'arm_with_gripper_group' will use planner 'geometric::RRTConnect'. Additional configuration parameters will be set when the planner is constructed.

v4hn:
In case the incoming RobotStateMsg is a diff, the joint fields are allowed to be empty.
This is used to change attached objects and it's an intended use-pattern in MoveIt to modify the current state with some (possibly) incoming state (see e.g. [the fix_start_state_bounds adapter](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_bounds.cpp#L85))

This needs to be cherry-picked indigo due to #518 .

Fixes #579

### DIFF
--- a/moveit_core/robot_state/src/conversions.cpp
+++ b/moveit_core/robot_state/src/conversions.cpp
@@ -332,7 +332,7 @@ static bool _robotStateMsgToRobotStateHelper(const Transforms* tf, const moveit_
   bool valid;
   const moveit_msgs::RobotState& rs = robot_state;
 
-  if (rs.joint_state.name.empty() && rs.multi_dof_joint_state.joint_names.empty())
+  if (!rs.is_diff && rs.joint_state.name.empty() && rs.multi_dof_joint_state.joint_names.empty())
   {
     logError("Found empty JointState message");
     return false;


jrgnicho:
@v4hn thanks for the contribution.

v4hn:
@jrgnicho you forgot to cherry-pick to indigo. Done

PR #593, opened 2017-08-18
find and link against tinyxml where needed (#569)

v4hn:
Cherry-Pick #569 to indigo

remove tinyxml

spurious cmake_modules addition


bmagyar:
if build passes :+1:

PR #596, opened 2017-08-18
RobotState rviz previewer: First message from e.g. latching publishers is not applied to robot state correctly

### LINKED ISSUE #595: RobotState rviz previewer: First message from e.g. latching publishers is not applied to robot state correctly
I noticed an issue with the robot state display plugin. If you subscribe to a latching publisher, the first update/message is ignored and the robot is displayed in default state. The reason for this behaviour is the delayed loading of the robot model.

RobotStateDisplay::onEnable() subscribes to the configured topic and sets load_robot_model_ = true in order to load the model in the next update() cycle. Unfortunately, the subscriber creates the subscription before and the first message is evaluated and stored in the current robot state. Since loadRobotModel() resets this state, we won't see any incoming updates before executing loadRobotModel().

Proposed solution follows via PR. I am not familiar with the problems that leaded to the delayed loading of the robot model so please review this solution especially regarding thread contexts of the methods..

Thanks!

axelschroth:
Fixes #595

Note: I am not familiar with the problems that leaded to the delayed loading of the robot model so please review this solution especially regarding thread contexts of the methods..

### DIFF
--- a/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
+++ b/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
@@ -393,7 +393,8 @@ void RobotStateDisplay::update(float wall_dt, float ros_dt)
 {
   Display::update(wall_dt, ros_dt);
 
-  if (load_robot_model_) {
+  if (load_robot_model_)
+  {
     loadRobotModel();
     changedRobotStateTopic();
   }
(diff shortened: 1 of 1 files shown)


jrgnicho:
@axelschroth  thanks for submitting this.  Your change seems logical but before accepting I'd like to test it out.  Would you happen to have a repository with a test setup?

axelschroth:
I'm afraid  I don't, sorry.. The test setup I am working with is part of a big software infrastructure. You just have to use a latching publisher, publish a robot state and disable then enable the RobotState display.

Is there any repository already set up using a 6DOF robot which I could use to create a test repository?

v4hn:
I will have a look at and test this later this week.

jrgnicho:
@axelschroth  there are several moveit_config packages for some abb robots here https://github.com/ros-industrial/abb

v4hn:
Ok, I can't reproduce the error you face, but it's a racing condition, so that doesn't matter too much in this case.
I looked through the code and can verify the logic problem you describe in #595 .

I see no problem with changing the threading context of the problematic call to `changedRobotStateTopic`, as you proposed.

Tested and verified.

I believe we should backport this to indigo. I'll leave that to the merging maintainer.

PR #598, opened 2017-08-22
[IKFastTemplate] Expand solutions to full joint range in searchPositionIK

hartmanndennis:
Implementation of #571 for the no free parameter case of searchPositionIK. 
This implementation works without solution explosion and still returns the closest solution to the seed.
It should scale linearly with increased joint limits.
Implementation for the other case and other functions requires more code changes.

### DIFF
--- a/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
+++ b/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
@@ -312,6 +312,12 @@ private:
    */
   void getSolution(const IkSolutionList<IkReal>& solutions, int i, std::vector<double>& solution) const;
 
+  /**
+   * @brief Gets a specific solution from the set with joints rotated 360° to be near seed state where possible
+   */
+  void getSolution(const IkSolutionList<IkReal>& solutions, const std::vector<double>& ik_seed_state, int i,
+                   std::vector<double>& solution) const;
+
   double harmonize(const std::vector<double>& ik_seed_state, std::vector<double>& solution) const;
   // void getOrderedSolutions(const std::vector<double> &ik_seed_state, std::vector<std::vector<double> >& solslist);
   void getClosestSolution(const IkSolutionList<IkReal>& solutions, const std::vector<double>& ik_seed_state,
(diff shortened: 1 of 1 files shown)


moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -820,6 +820,27 @@ bool IKFastKinematicsPlugin::searchPositionIK(const geometry_msgs::Pose& ik_pose
...
+      {
+        double distance = solutions[i][j] - ik_seed_state[j];
+        while (distance > M_PI && (!joint_has_limits_vector_[j] ||
+                                   (solutions[i][j] - 2 * M_PI > (joint_min_vector_[j] - LIMIT_TOLERANCE))))
+        {
+          distance -= 2 * M_PI;
+          solutions[i][j] -= 2 * M_PI;
+        }
+        while (distance < -M_PI && (!joint_has_limits_vector_[j] ||
+                                    (solutions[i][j] + 2 * M_PI < (joint_max_vector_[j] - LIMIT_TOLERANCE))))
```
jrgnicho: Should this be 
`(solutions[i][j] + 2 * M_PI < (joint_max_vector_[j] + LIMIT_TOLERANCE)` with **+** LIMIT_TOLERANCE instead of **-**?

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -820,6 +820,27 @@ bool IKFastKinematicsPlugin::searchPositionIK(const geometry_msgs::Pose& ik_pose
       return false;
     }
 
+    // rotate joints by +/-360° where it is possible and useful
+    for (std::size_t i = 0; i < solutions.size(); i++)
+    {
+      for (std::size_t j = 0; j < num_joints_; ++j)
+      {
+        double distance = solutions[i][j] - ik_seed_state[j];
+        while (distance > M_PI && (!joint_has_limits_vector_[j] ||
```
jrgnicho: It looks like you could check the `joint_has_limits_vector_` variable at the very beginning of the second for loop instead of each while loop

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
hartmanndennis: Yes.

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
hartmanndennis: I don't think so. If a joint doesn't have joint limits I still want to rotate the solution.

v4hn: I'm not sure I interpret your edit right, but yes, if a joint is continuous, the output solution here should always be in [-pi;pi].

hartmanndennis: Changed it so it doesn't change continuous joints.

jrgnicho:
It took me a little while to make sense out of make sense out of this change but I think it should unwind the joint values as expected.

hartmanndennis:
I added a commit that adds this behavior for all cases by adding a new version of getSolution with an extra parameter ik_seed_state and calling this function in the relevant parts.
In this version multi-soluition getPositionIK still return only solutions that differ more than rotation(s) by 360° but each solution returned is closest to the seed state in respect to 360° rotations. I think this is a good compromise.
If you aggree with this version I'll squash rebase the PR. else I'll revert it.

v4hn:
The approach looks good to me.
I can't check the code/ verify runtime at the moment though.

v4hn:
Merging.

I'm unsure about whether this should be back-ported to indigo because it changes the behavior quite a bit.
I'll wait for users to speak up if they need it there.

Issue #608, opened 2017-09-04
Typo?


gerac83:
I found a typo as below:

https://github.com/ros-planning/moveit/blob/4bc43c56ca2a5293f3ef8d55796ec586cfde5f69/moveit_planners/ompl/ompl_interface/src/detail/ompl_console.cpp#L40

it should be ompl_interface. I have no idea if this is relevant at all.

bmagyar:
Thanks,  nice catch!  I'll look further into this today.

bmagyar:
I couldn't find any traces in the current moveit codebase of using this file at all.
As it does nothing but prints, I think it may even be time for this file to be retired...

v4hn:
If it is not referenced +1 to drop it.

bmagyar:
I'll sort that out :+1:

simonschmeisser:
can be closed?

bmagyar:
Yeah, thanks!

PR #609, opened 2017-09-06
Remove OutputHandlerROS from ompl_interface

bmagyar:
### Description

Following up on #608 .
Do you have anything similar you've discovered? 
I imagine the code I removed may have been used in early days of OMPL and ROS integration as it ties in ROS prints with the [OMPL system of console outputs](https://github.com/ompl/ompl/blob/master/src/ompl/util/Console.h#L102).

This is cleaning stuff out, I don't think it's necessary to cherry-pick to Indigo.

### Checklist
- [X] Optional: Decide if this should be cherry-picked to other current ROS branches (Indigo, Jade, Kinetic)


jrgnicho:
Looks good to me.

PR #613, opened 2017-09-19
add moveit_fake_controller_manager to run_depend of moveit_config_pkg_template/package.xml.template

k-okada:
moveit_setup_assistant generate launch file using fake controller, but the package.xml does not incude this package
https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/fake_moveit_controller_manager.launch.xml#L4


v4hn:
This is clearly right. I'll consider this a simple oversight and merge.

k-okada:
thanks.

PR #619, opened 2017-09-22
ikfast: replace ambiguous isnans in generated source

gavanderhoorn:
Alternative fix for #584.

I'd like this to replace #586.


gavanderhoorn:
@v4hn: sorry for taking so long.

v4hn:
Wow, deep Regex magic at work :-)

Thank you for taking care of this and I'm sorry it took me this long to respond after pushing you to do it.
IROS still keeps me pretty much busy.

@hartmanndennis could you please check if this works for you? You were the OP for this issue and I don't have a workspace around that I could use to test.

gavanderhoorn:
re: testing: yes, I've been thinking how to easily / properly test this, but the only thing I could come up with would be to generate an actual IKFast solver and make the test(s) run the `create_ikfast_moveit_plugin.py` script with the appropriate arguments.

I'll see if I can set that up - and how much additional time it adds to the - already long - build / test time of `moveit`.

hartmanndennis:
This solution works. I also checked updating multiple times and it correctly only adds std:: once.
But it seems to me that create_ikfast_moveit_plugin.py has a bug since the change that moved it to the moveit_kinematics package. See #620

gavanderhoorn:
re: multiple times: yes, that is what the negative lookbehind was for.

thanks for testing.

v4hn:
merging. Thanks for the feedback @hartmanndennis .
Thanks for fixing it @gavanderhoorn :)

Issue #618, opened 2017-09-22
ikfast: create plugin script fails to copy template files when built from source


gavanderhoorn:
### Description

`create_ikfast_moveit_plugin.py` in from-source 'builds' of the `moveit_kinematics` package appear to be unable to find the template files it needs when running the script.

### Steps to reproduce

Follow the *Generate IKFast Plugin* tutorial. It will fail when invoking `create_ikfast_moveit_plugin.py`.

### Expected behaviour

it should just copy the template files.

### Actual behaviour

throws an exception as it can't find the templates.

gavanderhoorn:
This would appear to be caused by the fact that `moveit_kinematics` is the package, but the templates are stored in a `ikfast_kinematics_plugin` sub directory.

When installed, the `templates` sub dir is placed in the root of `moveit_kinematics`.

In a from-source workspace, the `templates` dir is in the `ikfast_kinematics_plugin` sub directory, not at the root.

gavanderhoorn:
Fixed in #620.

PR #620, opened 2017-09-26
create_ikfast_moveit_plugin: fixed directory variable for templates that were moved to ikfast_kinematics_plugin

hartmanndennis:
Apparently this was never changed after the move to the moveit_kinematics package.

Needs cherry-picking to indigo.

### DIFF
--- a/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
+++ b/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
@@ -52,6 +52,7 @@ from lxml import etree
 import shutil
 
 plugin_gen_pkg = 'moveit_kinematics'  # package containing this file
+plugin_sub_dir = 'ikfast_kinematics_plugin' # sub directory which contains the template directory
 # Allowed search modes, see SEARCH_MODE enum in template file
 search_modes = ['OPTIMIZE_MAX_JOINT', 'OPTIMIZE_FREE_JOINT' ]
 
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
I think this is actually #618.

The change you propose only works when you build moveit from source in a catkin workspace. When released, the current location of the templates is actually correct.

gavanderhoorn:
To make it work in all cases we could install the `templates` sub dir in a `ikfast_moveit_plugin` sub dir of `moveit_kinematics`, but the change as proposed in this PR will break binary builds of `moveit_kinematics`.

hartmanndennis:
This should do it then.
The install path structure is now the same as the source.

moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
```diff
@@ -181,7 +181,7 @@
 
    # Get template folder location
    try:
-      plugin_gen_dir = roslib.packages.get_pkg_dir(plugin_gen_pkg)
+      plugin_gen_dir = roslib.packages.get_pkg_dir(plugin_gen_pkg) + '/ikfast_kinematics_plugin'
```
gavanderhoorn: Please `os.path.join(..)` for this.

gavanderhoorn: It would also be nice to make the `ikfast_kinematics_plugin` string constant an actual constant (as far as that is possible in Python).

Put it next to the `plugin_gen_pkg` constant at the top of the file.

hartmanndennis: Done.

v4hn:
@gavanderhoorn could you review this again&approve if appropriate?

gavanderhoorn:
This fixes #618.

gavanderhoorn:
And needs a cherry-pick to `indigo-devel`.

gavanderhoorn:
Thanks @hartmanndennis for the PR :100:

PR #621, opened 2017-09-26
Bug fixes in moveit_commander/robot.py

selyunin:
### Description

1. `_joint_owner_groups` had a typo (line 152)

2. `RobotCommander.get_default_owner_group` was incorrectly called (line 116)

3. When executing a motion command (RobotCommander.Joint.move) the target values for **other** joints should be updated only once, at the beginning. Originally line 121 `   gc.set_joint_value_target(gc.get_current_joint_values())` updated target values for all the joints in the group, which is not correct and resulted in a drift of unrelated joints due to forces of gravity, etc. Whenever the group in a robot is used first, joint values are updated (lines 121-123, 153), but for subsequent movements unrelated joints (i.e. **not** a target joint) are left unchanged, only target joint is moving.

### DIFF
--- a/moveit_commander/src/moveit_commander/robot.py
+++ b/moveit_commander/src/moveit_commander/robot.py
@@ -118,9 +118,7 @@ class RobotCommander(object):
                 raise MoveItCommanderException("There is no known group containing joint %s. Cannot move." % self._name)
             gc = self._robot.get_group(group)
             if gc is not None:
-                if not self._robot._is_group_initialized[gc.get_name()]:
-                    gc.set_joint_value_target(gc.get_current_joint_values())
-                    self._robot._is_group_initialized[gc.get_name()] = True
+                gc.set_joint_value_target(gc.get_current_joint_values())
                 gc.set_joint_value_target(self._name, position)
                 return gc.go(wait)
             return False
@@ -150,7 +148,6 @@ class RobotCommander(object):
         self._r = _moveit_robot_interface.RobotInterface(robot_description)
         self._groups = {}
         self._joint_owner_groups = {}
-        self._is_group_initialized = dict(zip(self._r.get_group_names(), [False]*len(self._r.get_group_names())))
 
     def get_planning_frame(self):
         """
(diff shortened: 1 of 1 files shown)


v4hn:
Thank you for the contribution @selyunin .
If nobody else manages to review this before, I will do that next week after I got back from IROS.

v4hn:
sorry for the late reply, (as I should have known beforehand) this dropped of my todo list.

Again I wonder why parts of this Python-API were apparently never tested... 😞 

I fully agree with 1. and 2. Could you please create a separate commit for them?

The third point is not clear to me and sounds like a problem you should fix in a different place.
Nobody guarantees that the other joints have not been moved between different calls to `Joint.move`.
So we *have* to update the other joints to the positions at which we believe them to be, in order to avoid safety hazards in this case.
The only thing I could imagine to do here is to add some magic threshold, below which the previous values are retained. I believe this could be an additional named parameter defaulting to 0, although it is pretty ugly in my opinion.

selyunin:
Thanks for your feedback, I have created separate commit for 1 & 2.

For 3, the problem is more subtle (or my explanation was not clear enough): if I issue a command 
`robor.get_joint('my_joint').move('math.pi/2', True)` i expect only 'my_joint' to move, not other joints.
and if I issue the command above twice:
```
robor.get_joint('my_joint').move('math.pi/2', True)
rospy.sleep(5)
robor.get_joint('my_joint').move('math.pi/2', True)
```
I expect to move only 'my_joint', and not other joints. And since the state of other joints is updated, other joints are moving in my case (even if I issued a command only for 'my_joint', and this error is accumulated over time). If I publish to the topic directly, not from the python API, I get the behavior as expected. IMHO this inconsistency should be resolved.

v4hn:
The current state of the pull-request fixes two straight-forward typos and the previous code did not work because of them.
I'll merge this as is and add a backport for indigo.

v4hn:
Again, thank you so much for contributing the fixes! 👍 

For your third problem:

What do you mean by
> If I publish to the topic directly

Do you have a ros_control topic for each joint independently and thus avoid the whole collision-planning?

Did you try to add auxiliary groups in your srdf that contain only one joint each?
This should avoid the problem in the python code, because it will select these groups as "default_owner_group" and not use the state of any other joint.

I would have to look through `move_group`'s Controller interface to see whether a similar issue could appear there.
Please give it a try and report back - preferably in a new issue.

Issue #636, opened 2017-10-13
optionally wait for controllers indefinitely


v4hn:
During startup, MoveIt's `move_group` node waits for the external controllers (`FollowJointTrajectory` or `GripperCommand` action servers) that have been specified in `config/controllers.yaml`.

However, if they do not come up within some fixed time, they are ignored, `move_group` continues, and they will not be available later on. MoveIt will be unable to execute trajectories for the relevant joints.

I would propose to make this behavior - for backward compatibility - optional via a ros-parameter.

See [here](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h#L76) for the code that handles this.

ipa-bfb:
Me and @ipa-aep can take care of this...

v4hn:
Merged in #695 .Thanks a lot for the contribution!

PR #655, opened 2017-10-18
optionally wait for controllers indefinitely #636

ipa-bfb:
### Description
We have added a node handler to ActionBasedControllerHandleBase class to load a time parameter. This time parameter is going to be used as the waiting time that the controller_action_client_->waitForServer has to wait. If the time is defined as zero the server will wait forever. This parameter is defined in controller_manager.xml in the trajectory_execution namespace.

Done with the contribution of @ipa-aep

### Checklist

### DIFF
--- a/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
+++ b/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
@@ -76,14 +76,19 @@ public:
     double timeout;
     nh_.param("trajectory_execution/controller_connection_timeout",timeout,5.0);
 
-    if(timeout==0)
-      ROS_WARN_STREAM_NAMED("moveit_simple_controller_manager", "Time delay is set to 0. Waiting for the controller connection timeout in the parameter server to be set...");
-
-    while (ros::ok() && !controller_action_client_->waitForServer(ros::Duration(timeout)) && ++attempts < 3){
-      ROS_ERROR_STREAM_NAMED("moveit_simple_controller_manager", "Waiting for " << getActionName() << " to come up");
-      ros::Duration(1).sleep();
+    if(timeout==0.0){
+      ROS_WARN_STREAM_NAMED("moveit_simple_controller_manager", "Time delay is set to 0. Waiting forever for the controller...");
+      while (ros::ok() && !controller_action_client_->waitForServer(ros::Duration(1.0))){
+        ROS_ERROR_STREAM_NAMED("moveit_simple_controller_manager", "Waiting for " << getActionName() << " to come up");
+        ros::Duration(1).sleep();
+      }
     }
-
+    else{
+      while (ros::ok() && !controller_action_client_->waitForServer(ros::Duration(timeout)) && ++attempts < 3){
+        ROS_ERROR_STREAM_NAMED("moveit_simple_controller_manager", "Waiting for " << getActionName() << " to come up");
+        ros::Duration(1).sleep();
+    }
+    }
     if (!controller_action_client_->isServerConnected())
     {
       ROS_ERROR_STREAM_NAMED("moveit_simple_controller_manager", "Action client not connected: " << getActionName());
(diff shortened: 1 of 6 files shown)


jrgnicho:
@ipa-bfb seems like clang formatting test caused the build to fail.

ipa-bfb:
I hope that now it's fixed ;) Sorry for that!

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
```diff
@@ -69,13 +69,30 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
...
     controller_action_client_.reset(new actionlib::SimpleActionClient<T>(getActionName(), true));
     unsigned int attempts = 0;
-    while (ros::ok() && !controller_action_client_->waitForServer(ros::Duration(5.0)) && ++attempts < 3)
-      ROS_INFO_STREAM_NAMED("moveit_simple_controller_manager", "Waiting for " << getActionName() << " to come up");
+    double timeout;
+    nh_.param("trajectory_execution/controller_connection_timeout", timeout, 5.0);
 
+    if (timeout == 0.0)
+    {
+      ROS_WARN_STREAM_NAMED("moveit_simple_controller_manager", "No controller specified. Waiting indefinitely...");
```
v4hn: I believe "No controller specified" is wrong and confusing here.
I would propose to remove the warning here - also because it will be triggered *every time* even if the controller is already connected.

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
```diff
@@ -69,13 +69,30 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
...
+        ros::Duration(1).sleep();
+      }
+    }
+    else
+    {
+      while (ros::ok() && !controller_action_client_->waitForServer(ros::Duration(timeout)) && ++attempts < 3)
+      {
+        ROS_ERROR_STREAM_NAMED("moveit_simple_controller_manager", "Waiting for " << getActionName() << " to come up");
+        ros::Duration(1).sleep();
+      }
```
v4hn: The logic here is a bit weird, isn't it?
If I see this correctly, the new "timeout" parameter means the controller manager will wait up to 3*(timeout+1) seconds in total for the controller because of the loop and the additional sleep.
Could you remove the `sleep`, call `waitForServer` with timeout/3 seconds, and change the default timeout to 15 seconds instead?
That would give the parameter an intuitive meaning and pretty much retain previous behavior.

I'm sorry for not having the time to look at that during WMD.  :(

ipa-bfb:
Thanks for the review @v4hn . I will take care of this asap ;)

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
ipa-bfb: Well it is only printed once and it is just to warn that it is going to wait indefinitely for the controller to connect. Should I just improve the message to "No controller connected. Waiting indefinitely..." or just remove it?

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
ipa-bfb: You are fully right. We just added the sleep to avoid having the warning message being printed all the time. In this way we have it printed only once per second. Is there a better way of doing this?

No problem.. We were already running against time.

ipa-bfb: You are fully right. We just added the sleep to avoid having the warning message being printed all the time. In this way we have it printed only once per second. Is there a better way of doing this?

No problem.. We were already running against time.

v4hn:
I cleaned up the branch (you merged unrelated changes in here)
and removed the additional warning in #695 .

Thank you for the changes again @ipa-bfb @ipa-aep .
I'm sorry I didn't get around to handle this sooner.

Closing in favor of #695

PR #660, opened 2017-10-18
add a helper script to more easily run docker containers with gui elements

tfoote:
This is a simple bash script to add the necessary elements to run gui programs inside docker. It is aware of nvidia-docker and will activate it if nvidia drivers and nvidia-docker is available.

### Description

### Checklist

### DIFF
--- a/.docker/gui-docker
+++ b/.docker/gui-docker
@@ -52,7 +52,5 @@ ${DOCKER_COMMAND} run \
   -v "/etc/localtime:/etc/localtime:ro" \
   -v /tmp/.X11-unix:/tmp/.X11-unix \
   -v "/tmp/.docker.xauth:/tmp/.docker.xauth" \
-  -v /dev/log:/dev/log \
-  --ulimit rtprio=99 \
   ${DOCKER_GPU_PARAMS} \
   $@
--- a/.docker/gui-docker
+++ b/.docker/gui-docker
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+# This script is used to create and run a docker container with settings for gui interactions
+# All arguments to this script will be appended to a docker run command.
+# Example command line:
+# ./gui-docker -ti --rm moveit/moveit /bin/bash
+
+# XAUTH=/tmp/.docker.xauth
+# xauth nlist :0 | sed -e 's/^..../ffff/' | xauth -f $XAUTH nmerge -
+if [ ! -f /tmp/.docker.xauth ]
+then
+  export XAUTH=/tmp/.docker.xauth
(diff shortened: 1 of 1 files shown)


tfoote:
Ok, ready for review now.

.docker/gui-docker
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+
+# This script is used to create and run a docker container with settings for gui interactions
+# All arguments to this script will be appended to a docker run command.
+# Example command line:
+# ./gui-docker -ti --rm moveit/moveit /bin/bash
```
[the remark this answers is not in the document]
davetcoleman: Also this defaults to ``moveit/moveit:latest`` which i don't have setup on dockerhub. i think its pulling indigo and was built a year ago... i'll try to fix that now.

``moveit/moveit:kinetic-source`` would be more explicit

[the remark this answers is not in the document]
davetcoleman: latest now points to kinetic-source...

.docker/gui-docker
```diff
@@ -0,0 +1,58 @@
...
+${DOCKER_COMMAND} run \
+  -e DISPLAY=unix$DISPLAY \
+  -e XAUTHORITY=/tmp/.docker.xauth \
+  -v "/etc/localtime:/etc/localtime:ro" \
+  -v /tmp/.X11-unix:/tmp/.X11-unix \
+  -v "/tmp/.docker.xauth:/tmp/.docker.xauth" \
+  -v /dev/log:/dev/log \
+  --ulimit rtprio=99 \
+  ${DOCKER_GPU_PARAMS} \
+  $@
```
davetcoleman: removing some of this, or adding comments to why its needed, would be preferable

.docker/gui-docker
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
```
davetcoleman: should this file be ``.sh`` to make it clearer its a bash script? I'm not sure what the standard is but personally I always do it this way for clarity from a binary

tfoote: If anything it should be a .bash sh is different than bash. But it's standard for executables to not expose their underlying implementation in their name, only for setup files do you need the explicit extension since they're not cross compatible.  This can be executed from any shell.

.docker/gui-docker
tfoote: I removed the ulimit and log pass through leaving that up to the user. The rest are the setup for the gui. The localtime is not strictly necessary but makes the time display like the host which is quite useful.

mcevoyandy:
Also tested locally. Worked great! Thanks!

PR #664, opened 2017-10-19
Disabled Copy Constructor

Jmeyer1292:
Related to #634 and #657, this PR presents the alternate solution where the `MoveGroupInterface` has its copy constructors disabled all-together. Note this requires changing the Python wrapper as well - not sure about the impacts there.

I also implemented a move constructor for those cases where you do need to move it around.

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -156,6 +156,17 @@ public:
 
   ~MoveGroupInterface();
 
+  /**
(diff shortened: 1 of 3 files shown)


v4hn:
👍 It compiles and works fine.
According to https://wiki.python.org/moin/boost.python/class the `noncopyable` for the python wrapper relates to an internal converter that is not registered. Copying the class in python (using the `copy` packages) is not possible with or without this patch.

This fulfills the description of @davetcoleman in https://github.com/ros-planning/moveit/pull/657#issuecomment-337759104 . I'll take that as a +1 and merge it into kinetic.

C++11, so no cherry-pick to indigo.
Also this changes the API (even though to fix a bug), so no backport either.

v4hn:
Thank you so much @Jmeyer1292 !

Issue #643, opened 2017-10-16
tests for `PlanningScene`


v4hn:
MoveIt's [`PlanningScene`](https://github.com/ros-planning/moveit/tree/kinetic-devel/moveit_core/planning_scene) is at the core of many of the operations the framework performs.

Even so, the [package's tests](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_core/planning_scene/test/test_planning_scene.cpp) contains only **three** tests.

This needs to be extended to account for the many interfaces this class supports!

You don't have to add tests for *everything*.
But if you could contribute reasonable tests for *some* things, this would help a lot.

AndriyPt:
The London team (Shadow + UCL) will be working on this :)

v4hn:
Cool! Looking forward to meaningful tests :-)

felixvd:
Is this done?

AndriyPt:
@felixvd I have managed to add small amount of tests. Maybe it will make sense to keep it open till next MoveIt day :)

PR #667, opened 2017-10-19
Added planning scene tests

AndriyPt:
This is some kind of refactoring which will allow to add more tests easily. Related to https://github.com/ros-planning/moveit/issues/643

### DIFF
--- a/moveit_core/planning_scene/test/test_planning_scene.cpp
+++ b/moveit_core/planning_scene/test/test_planning_scene.cpp
@@ -68,7 +68,11 @@ TEST(PlanningScene, LoadRestore)
   planning_scene::PlanningScene ps(urdf_model, srdf_model);
   moveit_msgs::PlanningScene ps_msg;
   ps.getPlanningSceneMsg(ps_msg);
+  EXPECT_EQ(ps.getName(), ps_msg.name);
+  EXPECT_EQ(ps.getRobotModel()->getName(), ps_msg.robot_model_name);
   ps.setPlanningSceneMsg(ps_msg);
+  EXPECT_EQ(ps.getName(), ps_msg.name);
+  EXPECT_EQ(ps.getRobotModel()->getName(), ps_msg.robot_model_name);
 }
 
 TEST(PlanningScene, LoadRestoreDiff)
--- a/moveit_core/planning_scene/test/test_planning_scene.cpp
+++ b/moveit_core/planning_scene/test/test_planning_scene.cpp
@@ -38,17 +38,19 @@
 #include <moveit/planning_scene/planning_scene.h>
 #include <urdf_parser/urdf_parser.h>
 #include <fstream>
+#include <string>
 #include <boost/filesystem/path.hpp>
 #include <moveit_resources/config.h>
 
+
 // This function needs to return void so the gtest FAIL() macro inside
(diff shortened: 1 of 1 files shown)


moveit_core/planning_scene/test/test_planning_scene.cpp
```diff
@@ -139,6 +155,20 @@ TEST(PlanningScene, MakeAttachedDiff)
...
+{
+  srdf::ModelSharedPtr srdf_model(new srdf::Model());
+  urdf::ModelInterfaceSharedPtr urdf_model;
+  loadRobotModels(urdf_model, srdf_model);
+
+  planning_scene::PlanningScenePtr ps(new planning_scene::PlanningScene(urdf_model, srdf_model));
+  robot_state::RobotState current_state = ps->getCurrentState();
+  if (ps->isStateColliding(current_state, "left_arm"))
+  {
+    EXPECT_FALSE(ps->isStateValid(current_state, "left_arm"));
```
rhaschke: This test isn't very useful, isn't it?
I think tests should verify the correct working of the API. In this particular example, you should set up some robot poses, that are known to collide/not collide and validate the result of `isStateColliding()`.
Or you should set up some constraints that are fulfilled or not and validate `isStateFeasible()`.

AndriyPt: @rhaschke Thank you for feedback. Work was done in the limited amount of time during World MoveIt Day 2017. Hopefully I will find some time to add suggested improvements before next WMD 2018 :)

davetcoleman: This seems useful enough to me... you use two different calls to verify expected behavior of invalid states.

moveit_core/planning_scene/test/test_planning_scene.cpp
```diff
@@ -57,8 +57,22 @@ void loadRobotModel(urdf::ModelInterfaceSharedPtr& robot_model_out)
...
+  std::string xml_string;
+  loadModelFile("pr2_description/urdf/robot.xml", xml_string);
   robot_model_out = urdf::parseURDF(xml_string);
 }
+void loadRobotModels(urdf::ModelInterfaceSharedPtr& robot_model_out, srdf::ModelSharedPtr& srdf_model_out)
+{
+  loadRobotModel(robot_model_out);
+  std::string xml_string;
+  loadModelFile("pr2_description/srdf/robot.xml", xml_string);
+  srdf_model_out->initString(*robot_model_out, xml_string);
```
davetcoleman: Ideally all tests would use the same test utilities for loading URDFs / SRDFs instead of remunerating them in every test. I have the same suggestion for a test @mlautman is working on in fact.

MoveIt! already has common file loading functionality for URDF / SRDF [here](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h#L61) - can that be reused in this file?

moveit_core/planning_scene/test/test_planning_scene.cpp
bmagyar: It's worth picking up these little bits. The more tests the better.

itssomething

PR #668, opened 2017-10-19
replace explicit use of Arial with default font

wjwwood:
### Description

This pull request replaces uses of `"Arial"` with the default application font from Qt with `QGuiApplication::font()`.

### Checklist
- [ ] ~~Extended the tutorials / documentation, if necessary [reference](http://moveit.ros.org/documentation/contributing/)~~
- [ ] ~~Include a screenshot if changing a GUI~~
- [ ] ~~Optional: Created tests, which fail without this PR [reference](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/tests.html)~~
  - I think this should be cherry-picked to indigo as well, but that's up to the maintainers.

### DIFF
--- a/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
@@ -41,6 +41,7 @@
 #include <QApplication>
 #include <QSplitter>
 #include <QRegExp>
+#include <QGuiApplication>
 // ROS
 #include "configuration_files_widget.h"
 #include <srdfdom/model.h>  // use their struct datastructures
@@ -149,7 +150,7 @@ ConfigurationFilesWidget::ConfigurationFilesWidget(QWidget* parent,
 
   // Success label
   success_label_ = new QLabel(this);
-  QFont success_label_font("Arial", 12, QFont::Bold);
+  QFont success_label_font(QGuiApplication::font().family(), 12, QFont::Bold);
   success_label_->setFont(success_label_font);
   success_label_->hide();  // only show once the files have been generated
   success_label_->setText("Configuration package generated successfully!");
--- a/moveit_setup_assistant/src/widgets/double_list_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/double_list_widget.cpp
@@ -43,6 +43,7 @@
 #include <QMessageBox>
 #include <QString>
 #include <QHeaderView>
+#include <QGuiApplication>
 #include "double_list_widget.h"
 
 namespace moveit_setup_assistant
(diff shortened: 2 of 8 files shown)


wjwwood:
Thanks @rhaschke for the fixup. 👍

rhaschke:
@wjwwood Thanks for your search-and-replace work in first place ;-)

rhaschke:
@v4hn Should we cherry-pick this into Indigo too?

v4hn:
I don't believe this is desirable, unless users asks for it.

Thanks for taking care of this both of you!

gavanderhoorn:
Does Indigo RViz not have the same problem? From the comment by @wjwwood I get the impression it does?

v4hn:
> Does Indigo RViz not have the same problem?

O.o I wasn't aware they changed the font *in indigo* too, until I just verified it.
That is a harsh change for a three year old release... Talking about "stable" releases.

In that case it makes sense to backport, yes. I'll add a request.

gavanderhoorn:
@v4hn wrote:
> That is a harsh change for a three year old release... Talking about "stable" releases.

it was a licensing issue. I guess it makes sense to address those, even in stable/old releases.

wjwwood:
> Does Indigo RViz not have the same problem? From the comment by @wjwwood I get the impression it does?

This pr is just "good to do", it wasn't fixing any problem. So it doesn't have to be backported to fix a problem unless you notice one in indigo. I think all the problems people were running into were due to mixing versions of rviz in strange ways in their workspaces.

> > That is a harsh change for a three year old release... Talking about "stable" releases.

> it was a licensing issue. I guess it makes sense to address those, even in stable/old releases.

Exactly. Since you guys are only referring to it here and not distributing it then you _have_ to backport to indigo like rviz did.

wjwwood:
edited the previous comment to add an important "do not"... :)

gavanderhoorn:
@wjwwood wrote:
> edited the previous comment to add an important "do not".. :)

yes, I was getting confused ;)

v4hn:
> Exactly. Since you guys are only referring to it here and not distributing it then you do not *have* to backport to indigo like rviz did.

Well,

1. I didn't dig deep enough but can we dismiss the possibility that the Arial font used on some system was - at least in some usages - the one provided by rviz?
2. If you removed arial usage throughout RViz in indigo, MoveIt might as well follow up for consistancy.

wjwwood:
Yeah, I'm not sure how referring to Arial as the font family on Ubuntu does not break since removing it from rviz but it seems to work. I guessed that Ubuntu or Qt is silently providing a fallback font in those cases for compatibility reasons. Or maybe Arial is redistributed by Ubuntu with a correct license to do so and everything. I'm not sure.

v4hn:
> I'm not sure.

So we might as well follow your change in indigo and not use an explicit font choice in MSA.

PR #669, opened 2017-10-20
replace explicit use of Arial with default application font (#668)

v4hn:
backport of #668 to indigo.


v4hn:
Backporting fails. see #682

PR #671, opened 2017-10-24
In harmonize function: Fix angle in [-2Pi; 2PI]

yoann-solana:
### Description

Hi,
I am reporting a minor bug about the harmonize function for the ik_fast_kinematics_plugin.

A minus sign was missing in order to properly limit the angles in [-2PI; 2PI]
It is fixed with this pull request. 

About the test, the code of the harmonize function was tested quickly in Matlab and returned the expected angles in [-2PI; 2PI].

Regards,
Yoann.

### Checklist

### DIFF
--- a/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
+++ b/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
@@ -586,7 +586,7 @@ double IKFastKinematicsPlugin::harmonize(const std::vector<double>& ik_seed_stat
     {
       ss[i] -= 2 * M_PI;
     }
-    while (ss[i] < 2 * M_PI)
+    while (ss[i] < -2 * M_PI)
     {
       ss[i] += 2 * M_PI;
     }
@@ -594,7 +594,7 @@ double IKFastKinematicsPlugin::harmonize(const std::vector<double>& ik_seed_stat
     {
       solution[i] -= 2 * M_PI;
     }
-    while (solution[i] < 2 * M_PI)
+    while (solution[i] < -2 * M_PI)
     {
       solution[i] += 2 * M_PI;
     }


bmagyar:
How about using the `angles` library?

https://github.com/ros/angles/blob/master/angles/include/angles/angles.h#L81

yoann-solana:
Hi bmagyar,

The normalize_angle function you propose constrained an angle in [-PI; PI]. However, it seems the harmonize function constrained the angle in [-2PI; 2PI]. 
The correction I proposed is minor and ensures the angle to be in [-2PI; 2PI]. 

Regards,
Yoann.

bmagyar:
This is just normalizing the seed state and solution vectors. Any particular reason why ikfast needs that bigger range? To me a normalized angle is within [-PI, PI]. Can you check if that works as well? 

Could you add a test for this please?   I mean a proper integrated gtest,  forget Matlab.

Jmeyer1292:
I believe IKFast typically returns values already normalized to -PI to PI before anything ever hits MoveIt. As some recent pull requests have observed, it is sometimes necessary to explore the kinematically equivalent solutions at joint values 2PI from the current solution. This is most relevant when exploring for the closest position near a given seed. 

I believe the actual point of this function was to perform that operation. This implementation is totally busted though, and it currently does math on the seed state vector that it copies then immediately discards.

The fun bit is that this harmonize function's actual solution has now been 'inlined' into this file in a few other places.

simonschmeisser:
With these changes this code is less obviously broken, however I wondered why this doesn't seem to create any problems for people. Looking closer `harmonize()` is only ever called by `getClosestSolution()` which is not used anywhere. So I think the best would be to completely remove those two functions (at least in melodic) and close this PR

Solutions are "wrapped" towards the seed state in `getSolution()` in melodic-devel (and maybe kinetic, didn't look)

rhaschke:
@yoann-solana Did you experienced an issue without this patch? As @simonschmeisser noted, this function shouldn't be used at all. Could you please comment?

rhaschke:
I will remove those functions in an upcoming PR as suggested in https://github.com/ros-planning/moveit/pull/671#issuecomment-456015384

PR #682, opened 2017-11-08
MSA: use qt4-compatible API for default font

v4hn:
We recently remove the hard-coded "Arial" font family from MSA
and instead detect the default font via Qt.

However, the QGuiApplication class only exists in qt5 and thus
current kinetic-devel is broken with ROS indigo (qt4).

We agreed the current development branch should support
the target platforms of all other active branches of MoveIt
as long as someone is willing to provide the patches.

I propose to move from `QGuiApplication()::font().family()` to `QFont().defaultFamily()`.
The latter is pretty much a stub in Qt4 (it only provides one out of 5-6 different hard-coded strings),
but the API exists in both Qt4 and Qt5.

### DIFF
--- a/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
@@ -41,7 +41,6 @@
 #include <QApplication>
 #include <QSplitter>
 #include <QRegExp>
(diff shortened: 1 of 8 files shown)


v4hn:
CI failed for the lunar test due to unrelated reasons (console output timeout).
Merging

PR #688, opened 2017-11-15
don't crash on empty robot_description in RobotState plugin

simonschmeisser:
when starting robot_state_display with an empty robot_description it will crash
as kstate_ is not checked before attempting to update it. Changing to an empty
robot_description later on seems to be handled more gracefully on a higher level
and does not crash

How to reproduce:
roscore
rviz
add RobotState plugin
observe crash

### Checklist
- [x] Optional: Decide if this should be cherry-picked to other current ROS branches (Indigo, Jade, Kinetic) - Yes, simple fix for crash


simonschmeisser:
@v4hn ping?

bmagyar:
Thanks!

bmagyar:
Also cherry-picked to Indigo.

PR #691, opened 2017-11-17
Avoid segfault when validating a multidof-only trajectory

### LINKED ISSUE #539: Segfault when validating MultiDOF trajectories
### Description
Segfault when executing trajectory with custom action file.  New to moveit so I'm hoping I might get pointed in the right direction or this might flag to someone in the know.

### Your environment
* ROS Distro: [Kinetic]
* OS Version: Ubuntu 16.04
* Source 
* 07448175688087a0216e42181d99b3ed1827e588

### Steps to reproduce
I have been attempting to run the simulator located at https://github.com/wilselby/ROS_quadrotor_simulator.

### Expected behaviour
the custom action file should implement a multi-degree of freedom joint and by all accounts has been working.

### Actual behaviour
When executing a planned trajectory the code segfaults resulting in the following backtrace.

https://gist.github.com/TroyCordie/55de6da4ec32112981d7ba4c38512298

v4hn:
Also warn that MultiDOFJointTrajectory validation is not supported

Fixes #539

Should be backported to indigo.

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
+++ b/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
@@ -951,6 +951,16 @@ bool TrajectoryExecutionManager::validate(const TrajectoryExecutionContext& cont
 
   for (const auto& trajectory : context.trajectory_parts_)
   {
+    if (!trajectory.multi_dof_joint_trajectory.points.empty())
+    {
+      ROS_WARN_NAMED("traj_execution", "Validation of MultiDOFJointTrajectory is not implemented.");
+      // go on to check joint_trajectory component though
+    }
+    if (trajectory.joint_trajectory.points.empty())
+    {
+      // There is nothing to check
+      continue;
+    }
     const std::vector<double>& positions = trajectory.joint_trajectory.points.front().positions;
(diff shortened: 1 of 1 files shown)


bmagyar:
I'd swap the order of those 2 checks as if there's no joint trajectory passed there is no way to get the warning about multidof trajectory.

It may be of personal taste but I'd also revamp this file a bit in order to remove all those `continue`-s. A `goto` is a `goto` even if they give it a new name :D

v4hn:
I forgot about this request...

You are right about the order of the checks. I swapped them.

> It may be of personal taste but I'd also revamp this file a bit in order to remove all those continue-s.

There is only one `continue` in this function and I very much prefer it over a big `if { ...}` block around the whole loop content. I would like to keep it as is.

Please have another look.

bmagyar:
Another question: shouldn't we just error out when a multidof trajectory arrives? This allows for passing in a bunch of stuff that then gets ignored.

v4hn:
In theory we could error out and tell the user to either submit a patch for working validation or disable validation explicitly. But that would entail that everyone planning with a multi-dof joint has to disable start state validation by hand. This use-case "just worked" before we added validation, so it would break peoples workflow.

I believe the warning is enough in this case to tell the user "hey submit a patch or don't complain if something breaks".

bmagyar:
Also, this topic is probably beyond the scope of this PR.

👍

PR #703, opened 2017-12-04
Fixed bugs in computation of AABB in LinkModel and RobotState (kinetic)

peci1:
Port of https://github.com/ros-planning/moveit/pull/699 to kinetic-devel.

### DIFF
--- a/moveit_core/robot_model/src/link_model.cpp
+++ b/moveit_core/robot_model/src/link_model.cpp
@@ -85,15 +85,17 @@ void moveit::core::LinkModel::setGeometry(const std::vector<shapes::ShapeConstPt
          collision_origin_transform_[i].translation().norm() < std::numeric_limits<double>::epsilon()) ?
             1 :
             0;
-    Eigen::Vector3d extents = shapes::computeShapeExtents(shapes_[i].get());
     Eigen::Affine3d transform = collision_origin_transform_[i];
 
     if (shapes_[i]->type != shapes::MESH)
     {
+      Eigen::Vector3d extents = shapes::computeShapeExtents(shapes_[i].get());
       aabb.extendWithTransformedBox(transform, extents);
     }
     else
     {
+      // we cannot use shapes::computeShapeExtents() for meshes, since that method does not provide information about
+      // the offset of the mesh origin
       const shapes::Mesh* mesh = dynamic_cast<const shapes::Mesh*>(shapes_[i].get());
       for (unsigned int j = 0; j < mesh->vertex_count; ++j)
       {
--- a/moveit_core/robot_model/src/link_model.cpp
+++ b/moveit_core/robot_model/src/link_model.cpp
@@ -85,19 +85,19 @@ void moveit::core::LinkModel::setGeometry(const std::vector<shapes::ShapeConstPt
          collision_origin_transform_[i].translation().norm() < std::numeric_limits<double>::epsilon()) ?
             1 :
             0;
-    Eigen::Vector3d ei = shapes::computeShapeExtents(shapes_[i].get());
-    Eigen::Affine3d t = collision_origin_transform_[i];
+    Eigen::Vector3d extents = shapes::computeShapeExtents(shapes_[i].get());
+    Eigen::Affine3d transform = collision_origin_transform_[i];
 
     if (shapes_[i]->type != shapes::MESH)
     {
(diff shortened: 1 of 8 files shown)


moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -2027,67 +2028,42 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup* gro
...
     }
   }
-  if (aabb.empty())
-    aabb.resize(6, 0.0);
+
+  aabb.clear();
+  aabb.resize(6, 0.0);
+  if (!bounding_box.isEmpty())
+  {
+    aabb[0] = bounding_box.min()[0];
```
rhaschke: This assignment is very verbose. What about:

``` c++
Map<VectorXd, Unaligned, InnerStride<2>>(aabb.data(), 3) = bounding_box.min();
Map<VectorXd, Unaligned, InnerStride<2>>(aabb.data()+1, 3) = bounding_box.max();
```

moveit_core/robot_model/src/link_model.cpp
```diff
@@ -85,25 +85,25 @@ void moveit::core::LinkModel::setGeometry(const std::vector<shapes::ShapeConstPt
          collision_origin_transform_[i].translation().norm() < std::numeric_limits<double>::epsilon()) ?
             1 :
             0;
-    Eigen::Vector3d ei = shapes::computeShapeExtents(shapes_[i].get());
-    Eigen::Vector3d p1 = collision_origin_transform_[i] * (-ei / 2.0);
-    Eigen::Vector3d p2 = collision_origin_transform_[i] * (-p1);
+    Eigen::Vector3d extents = shapes::computeShapeExtents(shapes_[i].get());
```
rhaschke: `extends` seems to be used for non-meshes only. Hence computation of shape extends should be moved into the non-meshes `if`-branch below.

moveit_core/robot_model/src/link_model.cpp
```diff
@@ -85,25 +85,25 @@ void moveit::core::LinkModel::setGeometry(const std::vector<shapes::ShapeConstPt
...
-      b = p2;
+      aabb.extendWithTransformedBox(transform, extents);
     }
     else
     {
-      for (int i = 0; i < 3; ++i)
-        a[i] = std::min(a[i], p1[i]);
-      for (int i = 0; i < 3; ++i)
-        b[i] = std::max(b[i], p2[i]);
+      const shapes::Mesh* mesh = dynamic_cast<const shapes::Mesh*>(shapes_[i].get());
```
rhaschke: Can you add a comment, why meshes are handled differently?
I guess, with your approach, you compute the smallest AABB for the mesh given the transformation, while otherwise an AABB of the mesh (in default pose) is computed and transformed afterwards, which usually gives a larger box.

peci1: You're right it deserves a comment, since the reasoning is different -  `shapes::computeShapeExtents` only gives the extents and doesn't provide info about the position of the center of the mesh in these extents.

peci1:
Both indigo and kinetic PRs now pass the tests, and I've addressed the last set of comments from rhascke in them. So from my point of view, the PR is ready.

a-price:
Has this fix been pushed to the Kinetic Debian release? I'm on 0.9.11-0xenial-20180319-133135-0800, but am still seeing this bug.

rhaschke:
No, there wasn't as new release since this PR was merged.

peci1:
Thanks for also releasing for indigo. However, I've noticed that this PR (and its indigo counterpart) isn't referenced in the changelog (to kinetic, it was added in 0.9.12; to indigo in 0.7.14). Should I make PRs against kinetic-devel and indigo-devel adding the mention to the appropriate places, or is it a no-go to edit changelogs of already released versions?

I think mentioning this in the changelog is important because people might find things broken (or fixed :) ) and start searching for the cause.

v4hn:
@130s @rhaschke as they do the releases at the moment.

Personally I believe it is not important to add the notes here anymore.
Also, there is no clear strategy to "fix" it if something is broken because of these patches,
otherwise I would propose to add it to [the MIGRATION.md](https://github.com/ros-planning/moveit/blob/kinetic-devel/MIGRATION.md)

PR #715, opened 2017-12-09
Batch cherry-pick kinetic -> indigo

130s:
- Multi DOF Trajectory only providing translation not velocity (#555)
- benchmarks: always prefer local header over system installations #630
- removed deprecated pluginlib macro #677


v4hn:
replaced by #716

PR #717, opened 2017-12-10
use waitForCurrentState(ros::Time::now(), instead of waitForCurrentSt…

k-okada:
…ate(1.0) in TrajectoryExecutionManager::validate, cherry pick part of #350

### Description

Current indigo moveit has regression compare to hydro, which you could find very easily  at [answers.ros.org](https://www.google.co.jp/search?ei=QUYtWvOXOYfI0gTXvIiYBw&q=%2Bsite%3Aanswers.ros.org+Failed+to+validate+trajectory%3A+couldn%27t+receive+full+current+joint+state+within+1s&oq=%2Bsite%3Aanswers.ros.org+Failed+to+validate+trajectory%3A+couldn%27t+receive+full+current+joint+state+within+1s&gs_l=psy-ab.3...3450.3450.0.3725.1.1.0.0.0.0.0.0..0.0....0...1c.1.64.psy-ab..1.0.0....0.LiRDd4s9FU0), such as

- https://answers.ros.org/question/252114/failed-to-validate-trajectory-couldnt-receive-full-current-joint-state-within-1s-error/
- https://answers.ros.org/question/254927/failed-to-validate-trajectory-after-updating-moveit-feb-2017/
- https://answers.ros.org/question/257874/moveit-cannot-update-robots-state/
- https://answers.ros.org/question/273444/cannot-use-moveit-with-dual_ur5-platform/

and interestingly this problem was fixed in Kinetic

that is because indigo moveit requires update of entire joints from `/joint_states` by using [`haveCompleteState()` function](https://github.com/ros-planning/moveit/blob/0.9.9/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp#L282-L293), whereas kinetic moveit uses `state_update_condition_.wait_for(lock, boost::chrono::nanoseconds((timeout - elapsed).toNSec()));` in https://github.com/ros-planning/moveit/pull/63 which uses https://github.com/ros-planning/moveit/blob/0.9.9/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp#L265-L280 with basicall wait for 

This change has been introduced in https://github.com/ros-planning/moveit/pull/350

To summarize, if `/joint_states` is not complite, 

- Hydro : The robot can move trajectory
- Indigo (before this PR), It shows `Failed to validate trajectory: couldn't receive full current joint state within 1s` and do not move
[...]

PR #721, opened 2017-12-12
Full back port of #350 and #298

k-okada:
see #717 (comment) for discussion

### Description

### Checklist


rhaschke:
Thanks for your effort, @k-okada. I will try to review today. However, first I need to discuss the more fundamental question, how to solve the issue at all (by back-porting #350 and breaking ABI or by some other poor-man's approach).

rhaschke:
We prefer ABI-compatibility and thus slightly modifying #717.

Issue #723, opened 2017-12-14
attached bodies are not shown in trajectory visualization anymore


rhaschke:
Since MoveIt 0.9.10 attached bodies are not shown in trajectory visualization using rviz.
Is anybody aware of changes related to that? I appreciate any hint. Otherwise I will go and bisect...

v4hn:
This issue is a bit underspecified...

Do you mean if you send a `moveit_msgs::DisplayTrajectory` with `AttachedCollisionObjects` in `trajectory_start.attached_collision_objects`, then they are not shown in the visualized trajectory?
I'm not aware of any change there.

jeffreyling:
I ran into a similar issue, and I think I found the problem. In [PR 663](https://github.com/ros-planning/moveit/pull/663/files) the function `attachedBodiesToAttachedCollisionObjectMsgs` in `robot_state/conversions.h` should take a vector reference as argument, not a vector copy. 

This is messing up the part where the attached bodies are supposed to be added to the msg.

rhaschke:
@jeffreyling Thanks for tracking this down. I will have a look.

rhaschke:
Closing via #724

PR #724, opened 2017-12-15
cleanup #663

rhaschke:
This PR provides some cleanup to the previously merged PR #663.
As a major regression, after #663, displaying of attached collision objects didn't work anymore (#723).
Besides fixing this, the present PR provides some other minor improvements.


v4hn:
Tests for the PlanningScene #667 would have helped, but nobody got around to review this yet.
Not to mention adding a test for this case.

I'll merge this to fix the bug and remove the release blocker.

rhaschke:
@velveteenrobot As we had a regression after accepting your original PR #663, may I ask you to provide some unittests that validate the API to interact with the planning scene? This will help to avoid regression issues in future.
For example, you could insert collision objects through various API calls and validate that your new getters will fetch the correct messages. Thanks a lot in advance.

rhaschke:
Thanks. That would be great.

PR #727, opened 2017-12-18
[indigo] Use lookupParam() in kinematics plugins

130s:
Cherry-pick #714 to `indigo-devel` branch.


v4hn:
@130s Thanks for the review!

PR #738, opened 2018-01-09
Fixed bugs in computation of AABB in LinkModel and RobotState.

peci1:
Cherry-picked f6b85c6 from #703 and rebased changes from #699 on top of it.


peci1:
That makes sense, fixed.

PR #748, opened 2018-01-23
Floating Joint Support in CurrentStateMonitor

v4hn:
MoveIt supports Floating and Planar joints throughout the pipeline... almost throughout the pipeline.
Whenever someone actually tried to use it, they encountered problems with how to get the current state information into the MoveIt system. For over at least 4(!!!) years.

See among others [here](https://groups.google.com/forum/#!topic/moveit-users/C07EZmTgSBQ) and [here](https://github.com/ros-planning/moveit_ros/issues/620).

This works (rather hacked) for floating joints as root joint of the robot, but they are also important for online calibration procedures and mobile objects that are inherent to the setup and should be part of the urdf.

Turns out most things that are required to get this to work in general have been done a long time ago.
I simply implement a proposal by @hersh and add a tf callback that reads the state of all multi-dof joints from TF. This still does not allow to get velocities/effort readings into moveit, but it covers most everyday use-cases for multi-dof joints.
Also because it uses TF updates, it "just works" in practice and you don't have to explain the difference between transforms and joint states of floating joints to users. There simply isn't any.

This is mostly a generalization of the previous code. The only behavior that is supposed to change is that joint states for multi-dof joints are updated even if no `JointState` messages are received and are updated whenever a transform is published instead.

Please note that the class already subscribes to TF anyway (it is passed a tf-listener) and that I only lock the state if an actual update happens, so that most TF callbacks don't block the RobotState.

This breaks ABI, so it will not be back-ported to indigo.

Lastly please note the proposed code intentionally supports compiling on indigo.

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
@@ -163,39 +163,17 @@ std::string planning_scene_monitor::CurrentStateMonitor::getMonitoredTopic() con
     return "";
 }
 
-bool planning_scene_monitor::CurrentStateMonitor::isPassiveOrMimicDOF(const std::string& dof) const
-{
-  if (robot_model_->hasJointModel(dof))
-  {
-    if (robot_model_->getJointModel(dof)->isPassive() || robot_model_->getJointModel(dof)->getMimic())
-      return true;
-  }
-  else
-  {
-    // check if this DOF is part of a multi-dof passive joint
-    std::size_t slash = dof.find_last_of("/");
-    if (slash != std::string::npos)
-    {
-      std::string joint_name = dof.substr(0, slash);
-      if (robot_model_->hasJointModel(joint_name))
-        if (robot_model_->getJointModel(joint_name)->isPassive() || robot_model_->getJointModel(joint_name)->getMimic())
-          return true;
-    }
-  }
-  return false;
-}
-
 bool planning_scene_monitor::CurrentStateMonitor::haveCompleteState() const
 {
   bool result = true;
-  const std::vector<std::string>& dof = robot_model_->getVariableNames();
+  const std::vector<const moveit::core::JointModel*>& joints = robot_model_->getJointModels();
   boost::mutex::scoped_lock slock(state_update_lock_);
-  for (std::size_t i = 0; i < dof.size(); ++i)
-    if (joint_time_.find(dof[i]) == joint_time_.end())
+  for (const moveit::core::JointModel* joint : joints)
+    if (joint_time_.find(joint) == joint_time_.end())
     {
-      if (!isPassiveOrMimicDOF(dof[i]))
+      if (!joint->isPassive() && !joint->getMimic())
       {
-        ROS_DEBUG("Joint variable '%s' has never been updated", dof[i].c_str());
+        ROS_DEBUG("Joint '%s' has never been updated", joint->getName());
         result = false;
       }
     }
@@ -205,13 +183,13 @@ bool planning_scene_monitor::CurrentStateMonitor::haveCompleteState() const
 bool planning_scene_monitor::CurrentStateMonitor::haveCompleteState(std::vector<std::string>& missing_states) const
 {
   bool result = true;
-  const std::vector<std::string>& dof = robot_model_->getVariableNames();
+  const std::vector<const moveit::core::JointModel*>& joints = robot_model_->getJointModels();
   boost::mutex::scoped_lock slock(state_update_lock_);
-  for (std::size_t i = 0; i < dof.size(); ++i)
-    if (joint_time_.find(dof[i]) == joint_time_.end())
-      if (!isPassiveOrMimicDOF(dof[i]))
+  for (const moveit::core::JointModel* joint : joints)
+    if (joint_time_.find(joint) == joint_time_.end())
+      if (!joint->isPassive() && !joint->getMimic())
       {
-        missing_states.push_back(dof[i]);
+        missing_states.push_back(joint->getName());
         result = false;
       }
   return result;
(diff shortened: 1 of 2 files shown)


moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -138,7 +145,12 @@ void planning_scene_monitor::CurrentStateMonitor::stopStateMonitor()
   if (state_monitor_started_)
   {
     joint_state_subscriber_.shutdown();
-    ROS_DEBUG("No longer listening o joint states");
+    if (tf_ && tf_connection_)
+    {
+      tf_->removeTransformsChangedListener(*tf_connection_);
+      tf_connection_ = nullptr;
```
rhaschke: Better use `tf_connection_.reset()`?

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -391,53 +403,91 @@ void planning_scene_monitor::CurrentStateMonitor::jointStateCallback(const senso
...
+      const std::string& child_frame = joint->getChildLinkModel()->getName();
+      const std::string& parent_frame =
+          joint->getParentLinkModel() ? joint->getParentLinkModel()->getName() : robot_model_->getModelFrame();
 
       std::string err;
       ros::Time tm;
-      tf::StampedTransform transf;
-      bool ok = false;
-      if (tf_->getLatestCommonTime(parent_frame, child_frame, tm, &err) == tf::NO_ERROR)
+      if (!tf_->getLatestCommonTime(parent_frame, child_frame, tm, &err) == tf::NO_ERROR)
```
rhaschke: Better write as `!= NO_ERROR`

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -391,53 +403,91 @@ void planning_scene_monitor::CurrentStateMonitor::jointStateCallback(const senso
...
       {
+        ROS_DEBUG_STREAM_THROTTLE(1, "Unable to update multi-dof joint '"
+                                         << joint->getName() << "': tf has no common time between '"
+                                         << parent_frame.c_str() << "' and '" << child_frame.c_str() << "': " << err);
+        continue;
+      }
+
+      const std::vector<std::string>& joint_vars = joint->getVariableNames();
+      assert(joint_vars.size() > 0);
+      if (joint_time_[joint_vars[0]] != tm)
```
rhaschke: I think this should use the joint name instead of joint's variable name(s). This also resolves ambiguity, which one to use - why did you choose `joint_vars[0]`?

rhaschke: Is it safe to read-access `joint_time_` without locking? A writing thread may change the map meanwhile...

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -391,53 +403,91 @@ void planning_scene_monitor::CurrentStateMonitor::jointStateCallback(const senso
...
-        ROS_DEBUG_THROTTLE(1, "Unable to lookup transform from %s to %s: no common time.", parent_frame.c_str(),
-                           child_frame.c_str());
-      if (ok && last_tf_update_ != tm)
-      {
-        update = true;
-        last_tf_update_ = tm;
-        const std::vector<std::string>& vars = robot_model_->getRootJoint()->getVariableNames();
-        for (std::size_t j = 0; j < vars.size(); ++j)
-          joint_time_[vars[j]] = tm;
+        for (const std::string& var : joint_vars)
```
rhaschke: Here it suffices to use the joint name as well. If there are several floating joints, their joint vars will be identical and hence, the map will not be unique anymore...

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -391,53 +403,91 @@ void planning_scene_monitor::CurrentStateMonitor::jointStateCallback(const senso
...
+                                         << parent_frame.c_str() << "' and '" << child_frame.c_str() << "': " << err);
+        continue;
+      }
+
+      const std::vector<std::string>& joint_vars = joint->getVariableNames();
+      assert(joint_vars.size() > 0);
+      if (joint_time_[joint_vars[0]] != tm)
+      {
+        // only lock if we actually update something
+        boost::mutex::scoped_lock _(state_update_lock_);
```
rhaschke: Instead of locking the state multiple times, I suggest to accumulate all TF updates in a map and apply them together.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -391,53 +403,91 @@ void planning_scene_monitor::CurrentStateMonitor::jointStateCallback(const senso
...
+    for (std::size_t i = 0; i < update_callbacks_.size(); ++i)
+      update_callbacks_[i](joint_state);
+
+  // notify waitForCurrentState() *after* potential update callbacks
+  state_update_condition_.notify_all();
+}
+
+void planning_scene_monitor::CurrentStateMonitor::tfCallback()
+{
+  bool update = false;
```
rhaschke: I don't see a difference between `update` and `changes`.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
v4hn: nitpick :P

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
v4hn: The implementation of the monitor always triggers the update signal at the end of the JointState callback, because some thread might be waiting for the joint states at the current time.
But it calls the `update_callbacks_` only if something actually *changed* (see [here](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp#L357)).

With the tf callback we can't even be sure the time of any joint updated and we don't want to trigger the notify unless we actually updated *something*.
This is represented in `bool update`.

`bool changes` represents whether something actually changed.
Because Multi-DOF joints tend to be rather noisy, I decided to use a threshold instead of just comparing the values.
Afair the update callbacks in MoveIt mark the planning scene as changed which can trigger safety stops in currently executing plans, re-validation of current plans, and publishing of the monitored_planning_scene topic. So unnecessary update callbacks can cause a lot of overhead.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
v4hn: of course you are right about locking.. fixed.

v4hn: Also I adjusted the map as you proposed. Nice idea 👍 
The previous implementation was correct though.
joint_vars[0] represents the first dof of the joint *including the joint name prefix*, e.g. `joint_name/trans_x`. I chose the first one as a representation for all dof of this joint.
It's much better to just store the time per joint instead and apparently it doesn't break anything else in MoveIt.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
v4hn: makes sense, implemented.

v4hn:
Thanks @rhaschke .
I believe I addressed all your feedback.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
rhaschke: Sorry, still don't get the difference: I guess you update something only if it has changed, don't you?

rhaschke: Got it now from reading the code: If there is only a tiny change, you apply this change, but you don't want to call the callbacks.

v4hn:
Thanks for the review.

> Isn't it overkill to have four similar implementations of haveCompleState()?

I agree, but they are exposed API.
If you feel they should all use the same implementation, go ahead.
Personally I don't mind.

AustinDeric:
@v4hn if i am to understand this correctly, the work flow is:
- Make a floating joint between the robot base and world frame
- update the transform between frames during run time using a TF publisher
- The internal moveit scen should update

If so this, isn't working for us.  I just wanted to make sure there is no other setup necessary besides floating joints and a tf publisher.  We have a solution that updates the robot_description parameter and then forces moveit to  update its internal model but we would like to use floating joints and TF.

v4hn:
> @v4hn if i am to understand this correctly, the work flow is:

That's the workflow, although you can actually have the floating joint anywhere, not just between base and world.
I used this a few weeks ago with a source build for a mobile base and it worked.

If you still encounter problems, please file questions at answers.ros.org

PR #756, opened 2018-02-05
MoveAction capability can drop cancel request if it is sent shortly after goal is sent

khupilz:
### Description

Preempt request could be dropped in `MoveGroupMoveAction` capability if it is sent shortly after goal is sent. 

The problem is that the `preemptMoveCallback` sets the `PlanExecution::preempt_requested_` as `true` but `executeMoveCallback_PlanAndExecute` resets the `PlanExecution::preempt_requested_` as `false`. If `preemptMoveCallback` is called before `plan_execution_->planAndExecute`, the preempt request is  lost. In this case the goal status is set to `PREEMPTING` but the goal is planned and executed. In the end a result with error code `SUCCESS` is received. A test is provided to reproduce the problem.

An easy fix is provided by adding a similar `preempt_requested_` flag in  `MoveGroupMoveAction` and checking this flag before `plan_execution_->planAndExecute` is called. At the end of `executeMoveCallback` the flag is reset to `false`. 

I am not sure if this is the right place to fix since there could be also similar problems in other action capabilities. 

### Checklist

### DIFF
--- a/moveit_ros/move_group/package.xml
+++ b/moveit_ros/move_group/package.xml
@@ -33,7 +33,6 @@
   <run_depend>std_srvs</run_depend>
 
   <test_depend>rostest</test_depend>
-  <test_depend>rosunit</test_depend>
   <test_depend>moveit_resources</test_depend>
 
   <export>
--- a/moveit_ros/move_group/package.xml
+++ b/moveit_ros/move_group/package.xml
@@ -32,6 +32,10 @@
   <run_depend>pluginlib</run_depend>
   <run_depend>std_srvs</run_depend>
 
+  <test_depend>rostest</test_depend>
+  <test_depend>rosunit</test_depend>
+  <test_depend>moveit_resources</test_depend>
+
   <export>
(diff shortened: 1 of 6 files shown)


moveit_ros/move_group/package.xml
```diff
@@ -32,6 +32,10 @@
   <run_depend>pluginlib</run_depend>
   <run_depend>std_srvs</run_depend>
 
+  <test_depend>rostest</test_depend>
+  <test_depend>rosunit</test_depend>
```
rhaschke: Are you sure `rosunit` is required? Looks like you are using only standard python `unittest` and `rostest`.

khupilz: Hi Robert, you are right. `rosunit` is needless. I removed it.

rhaschke:
@khupilz Sometimes, the unittest still fails:
https://travis-ci.org/ros-planning/moveit/jobs/341626697#L1098

Could you have another look?

rhaschke:
@khupilz: Ping.

PR #763, opened 2018-02-13
improve RobotState::updateStateWithLinkAt()

rhaschke:
The method `RobotState::updateStateWithLinkAt()` is used to warp a link (together with its descendants) to an arbitrary position. This is used in moveit_ros_manipulation's pick-place pipeline to (i) check collisions of the end-effector in target pose and (ii) to visualize the end-effector in that pose.

However, so far only the link transforms were updated, not the collision transforms. Consequently, MoveIt complains about dirty collision transforms when trying to access them.

This PR now updates both kind of transforms. However, to suppress MoveIt's warnings about invalid transforms, it's important to not mark the state as dirty (although the state's transforms don't resemble the joint variables)!


rhaschke:
Hm. Obviously I misunderstood the semantics of `RobotState::updateStateWithLinkAt()`. 
I will close this for now and adapt the PR.

Issue #616, opened 2017-09-21
moveit setup show no active joints or end effectors found for group ''


jacknlliu:
I forked ROS packages from  [ROS-I/universal_robot](https://github.com/ros-industrial/universal_robot/). And I follow [Moveit Setup Assistant Tutorial](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/setup_assistant/setup_assistant_tutorial.html). After setup Moveit config, I got a `demo.launch` file, then   
```
roslaunch  ur5_grasp_moveit_config demo.launch
```

I got the following errors, and the `rviz` did not show any `interactive markers`.  

```
[ INFO] [1505957647.779692210]: Fake controller 'fake_ur5manipulator_controller' with joints [ shoulder_pan_joint shoulder_lift_joint elbow_joint wrist_1_joint wrist_2_joint wrist_3_joint ]  
[ INFO] [1505957647.788025775]: Fake controller 'fake_gripper_controller' with joints [ bh_j11_joint bh_j12_joint bh_j22_joint bh_j32_joint ]  
[ INFO] [1505957647.796457498]: Fake controller 'fake_moveit_ee_controller' with joints [ ]  
[ INFO] [1505957647.807434551]: Returned 3 controllers in list  
[ INFO] [1505956629.889136115]: Loading robot model 'ur5'...  
[ WARN] [1505956629.889226914]: Skipping virtual joint 'virtual_joint' because its child frame 'box_fixed_link' does not match the URDF frame 'world'  
[ INFO] [1505956631.143209340]: No active joints or end effectors found for group ''. Make sure you have defined an end effector in your SRDF file and that kinematics.yaml is loaded in this node's namespace.  
[ INFO] [1505956631.143508371]: No active joints or end effectors found for group 'gripper'. Make sure you have defined an end effector in your SRDF file and that kinematics.yaml is loaded in this node's namespace.  
[ INFO] [1505956631.212833955]: No active joints or end effectors found for group 'gripper'. Make sure you have defined an end effector in your SRDF file and that kinematics.yaml is loaded in this node's namespace.  
[ INFO] [1505956631.213958002]: Constructing new MoveGroup connection for group 'gripper' in namespace ''  
[ INFO] [1505956633.034567194]: Ready to take commands for planning group gripper.  
```
It seems that I got a group named '', it's so strange!

I make a  robot using a fixed box as the human-like truck, a UR5 robot arm and a Kinect as human-like eyes.

My URDF model please see [here]()

[...]

PR #769, opened 2018-02-15
improve association of IK solvers to groups

rhaschke:
This is an attempt to slightly improve the association of IK solvers to joint model groups, which is essential to have interactive end-effector markers in rviz.

There are frequently issues reported related to this, e.g. #616.
This PR aims to 
* improve both code and user documentation
* improve on issued warnings

### DIFF
--- a/moveit_core/robot_model/src/robot_model.cpp
+++ b/moveit_core/robot_model/src/robot_model.cpp
@@ -1258,22 +1258,21 @@ void moveit::core::RobotModel::interpolate(const double* from, const double* to,
 void moveit::core::RobotModel::setKinematicsAllocators(const std::map<std::string, SolverAllocatorFn>& allocators)
 {
   // we first set all the "simple" allocators -- where a group has one IK solver
-  for (JointModelGroupMap::const_iterator it = joint_model_group_map_.begin(); it != joint_model_group_map_.end(); ++it)
+  for (JointModelGroup* jmg : joint_model_groups_)
   {
-    std::map<std::string, SolverAllocatorFn>::const_iterator jt = allocators.find(it->second->getName());
+    std::map<std::string, SolverAllocatorFn>::const_iterator jt = allocators.find(jmg->getName());
     if (jt != allocators.end())
     {
       std::pair<SolverAllocatorFn, SolverAllocatorMapFn> result;
       result.first = jt->second;
-      it->second->setSolverAllocators(result);
+      jmg->setSolverAllocators(result);
     }
   }
 
   // now we set compound IK solvers; we do this later because we need the index maps computed by the previous calls to
   // setSolverAllocators()
-  for (JointModelGroupMap::const_iterator it = joint_model_group_map_.begin(); it != joint_model_group_map_.end(); ++it)
+  for (JointModelGroup* jmg : joint_model_groups_)
   {
-    JointModelGroup* jmg = it->second;
     std::pair<SolverAllocatorFn, SolverAllocatorMapFn> result;
     std::map<std::string, SolverAllocatorFn>::const_iterator jt = allocators.find(jmg->getName());
     if (jt == allocators.end())
@@ -1284,13 +1283,12 @@ void moveit::core::RobotModel::setKinematicsAllocators(const std::map<std::strin
 
       std::vector<const JointModelGroup*> subs;
 
-      // go through the groups that we know have IK allocators and see if they are included in the group that does not;
-      // if so, put that group in sub
+      // go through the groups that have IK allocators and see if they are part of jmg; collect them in subs
       for (std::map<std::string, SolverAllocatorFn>::const_iterator kt = allocators.begin(); kt != allocators.end();
            ++kt)
       {
         const JointModelGroup* sub = getJointModelGroup(kt->first);
-        if (!sub)
+        if (!sub)  // this should actually not happen, all groups should be well defined
         {
           subs.clear();
           break;
@@ -1299,10 +1297,15 @@ void moveit::core::RobotModel::setKinematicsAllocators(const std::map<std::strin
         sub_joints.insert(sub->getJointModels().begin(), sub->getJointModels().end());
 
         if (std::includes(joints.begin(), joints.end(), sub_joints.begin(), sub_joints.end()))
-        {
+        {  // sub_joints included in joints: add sub, remove sub_joints from joints set
           std::set<const JointModel*> resultj;
           std::set_difference(joints.begin(), joints.end(), sub_joints.begin(), sub_joints.end(),
                               std::inserter(resultj, resultj.end()));
+          // TODO: instead of maintaining disjoint joint sets here,
+          // should we leave that work to JMG's setSolverAllocators() / computeIKIndexBijection()?
+          // There, a disjoint bijection from joints to solvers is computed anyway.
+          // Underlying question: How do we resolve overlaps? Now the first considered sub group "wins"
+          // But, if the overlap only involves fixed joints, we could consider all sub groups
           subs.push_back(sub);
           joints.swap(resultj);
         }
--- a/moveit_ros/robot_interaction/src/robot_interaction.cpp
+++ b/moveit_ros/robot_interaction/src/robot_interaction.cpp
@@ -93,8 +93,7 @@ void RobotInteraction::decideActiveComponents(const std::string& group, Interact
   decideActiveJoints(group);
   if (!group.empty() && active_eef_.empty() && active_vj_.empty() && active_generic_.empty())
     ROS_INFO_NAMED("robot_interaction", "No active joints or end effectors found for group '%s'. "
-                                        "Make sure you have defined an end effector in your SRDF file and that "
-                                        "kinematics.yaml is loaded in this node's namespace.",
+                                        "Make sure that kinematics.yaml is loaded in this node's namespace.",
                    group.c_str());
 }
 
--- a/moveit_ros/robot_interaction/src/robot_interaction.cpp
+++ b/moveit_ros/robot_interaction/src/robot_interaction.cpp
@@ -91,7 +91,7 @@ void RobotInteraction::decideActiveComponents(const std::string& group, Interact
 {
   decideActiveEndEffectors(group, style);
   decideActiveJoints(group);
-  if (active_eef_.empty() && active_vj_.empty() && active_generic_.empty())
+  if (!group.empty() && active_eef_.empty() && active_vj_.empty() && active_generic_.empty())
     ROS_INFO_NAMED("robot_interaction", "No active joints or end effectors found for group '%s'. "
                                         "Make sure you have defined an end effector in your SRDF file and that "
                                         "kinematics.yaml is loaded in this node's namespace.",
@@ -185,10 +185,10 @@ void RobotInteraction::decideActiveJoints(const std::string& group)
   boost::unique_lock<boost::mutex> ulock(marker_access_lock_);
   active_vj_.clear();
 
-  ROS_DEBUG_NAMED("robot_interaction", "Deciding active joints for group '%s'", group.c_str());
-
   if (group.empty())
     return;
+
+  ROS_DEBUG_NAMED("robot_interaction", "Deciding active joints for group '%s'", group.c_str());
 
   const srdf::ModelConstSharedPtr& srdf = robot_model_->getSRDF();
   const robot_model::JointModelGroup* jmg = robot_model_->getJointModelGroup(group);
@@ -261,10 +261,10 @@ void RobotInteraction::decideActiveEndEffectors(const std::string& group, Intera
   boost::unique_lock<boost::mutex> ulock(marker_access_lock_);
   active_eef_.clear();
 
-  ROS_DEBUG_NAMED("robot_interaction", "Deciding active end-effectors for group '%s'", group.c_str());
-
   if (group.empty())
     return;
+
+  ROS_DEBUG_NAMED("robot_interaction", "Deciding active end-effectors for group '%s'", group.c_str());
 
   const srdf::ModelConstSharedPtr& srdf = robot_model_->getSRDF();
   const robot_model::JointModelGroup* jmg = robot_model_->getJointModelGroup(group);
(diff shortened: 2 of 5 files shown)


moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
```diff
@@ -102,10 +102,12 @@ PlanningGroupsWidget::PlanningGroupsWidget(QWidget* parent, moveit_setup_assista
   QVBoxLayout* layout = new QVBoxLayout();
 
   // Top Label Area ------------------------------------------------
-  HeaderWidget* header =
-      new HeaderWidget("Planning Groups", "Create and edit planning groups for your robot based on joint collections, "
-                                          "link collections, kinematic chains and subgroups.",
-                       this);
+  HeaderWidget* header = new HeaderWidget(
+      "Planning Groups", "Create and edit planning groups for your robot based on joint collections, "
+                         "link collections, kinematic chains and subgroups.\n"
```
v4hn: Please change this `and` to *or*, or make it otherwise clear that not everything is required.
There were complains about this being unclear before.

moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
```diff
@@ -102,10 +102,12 @@ PlanningGroupsWidget::PlanningGroupsWidget(QWidget* parent, moveit_setup_assista
...
+                         "A planning group defines the set of (joint, link) pairs considered for planning. "
```
v4hn: I am not sure what you mean by that statement.
So I'm pretty sure this is not very helpful to end-users...

What about

    Planning groups define the kinematic structures MoveIt can reason about.

or something similar instead?

moveit_ros/robot_interaction/src/robot_interaction.cpp
```diff
@@ -91,10 +91,9 @@ void RobotInteraction::decideActiveComponents(const std::string& group, Interact
 {
   decideActiveEndEffectors(group, style);
   decideActiveJoints(group);
-  if (active_eef_.empty() && active_vj_.empty() && active_generic_.empty())
+  if (!group.empty() && active_eef_.empty() && active_vj_.empty() && active_generic_.empty())
     ROS_INFO_NAMED("robot_interaction", "No active joints or end effectors found for group '%s'. "
-                                        "Make sure you have defined an end effector in your SRDF file and that "
-                                        "kinematics.yaml is loaded in this node's namespace.",
+                                        "Make sure that kinematics.yaml is loaded in this node's namespace.",
```
v4hn: Is this necessary? This code runs in RViz right?
`kinematics.yaml` is only loaded into the `move_group` namespace, isn't it?

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1299,10 +1297,15 @@ void moveit::core::RobotModel::setKinematicsAllocators(const std::map<std::strin
...
-        {
+        {  // sub_joints included in joints: add sub, remove sub_joints from joints set
           std::set<const JointModel*> resultj;
           std::set_difference(joints.begin(), joints.end(), sub_joints.begin(), sub_joints.end(),
                               std::inserter(resultj, resultj.end()));
+          // TODO: instead of maintaining disjoint joint sets here,
+          // should we leave that work to JMG's setSolverAllocators() / computeIKIndexBijection()?
+          // There, a disjoint bijection from joints to solvers is computed anyway.
+          // Underlying question: How do we resolve overlaps? Now the first considered sub group "wins"
+          // But, if the overlap only involves fixed joints, we could consider all sub groups
```
v4hn: Good questions...

MoveIt has been over-engineered in a number of aspects, this being one of them in my opinion...
We don't configure arbitrary new planning groups at runtime and at the moment I believe it is fine to force the user to provide specific ik parameters in ambiguous cases.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1258,22 +1258,21 @@ void moveit::core::RobotModel::interpolate(const double* from, const double* to,
 void moveit::core::RobotModel::setKinematicsAllocators(const std::map<std::string, SolverAllocatorFn>& allocators)
 {
   // we first set all the "simple" allocators -- where a group has one IK solver
-  for (JointModelGroupMap::const_iterator it = joint_model_group_map_.begin(); it != joint_model_group_map_.end(); ++it)
+  for (JointModelGroup* jmg : joint_model_groups_)
   {
-    std::map<std::string, SolverAllocatorFn>::const_iterator jt = allocators.find(it->second->getName());
```
v4hn: Just to clarify.

`it->second->getName()` here is the same as `it->first`, is it?
Who wrote that code. -.-

This could be rewritten to avoid the map in other places in the file too (including the destructor)...

moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
rhaschke: The planning group is not related to a kinematic _structure_ (which is a chain or a tree).

I want to clarify, that a planning group defines the set of joints that are actually used for planning with this group, i.e. joints that are sampled / moved.
I'm not sure, whether the planning group also restricts collision checking.

moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
rhaschke: Good point.

moveit_ros/robot_interaction/src/robot_interaction.cpp
rhaschke: No, rviz has it's a copy of kinematics.yaml loaded into its own (private) namespace. rviz doesn't access the move_group namespace. However, we might want to change _this_. There is no reason to have an independent set of parameters for moveit, indeed.
I would propose to make a new PR for this.

moveit_core/robot_model/src/robot_model.cpp
rhaschke: No, no, no! You misunderstand my point here. I don't want the user to explicitly specify all details of IK for all meta groups. It's excellent that MoveIt can resolve them automatically.
For example, on our bimanual setup, I have planning groups for each arm, fully specifying the IK.
However, the meta group `both_arms`, with subgroups `left_arm` and `right_arm` automatically inherits the IK solvers from their children.

My point here was, that disjointness of joint sets of ensured in two different places.

moveit_core/robot_model/src/robot_model.cpp
rhaschke: Why do you want to get rid of the map? It's useful to have fast name-based access to the JMG.
However, here we need to iterate over all JMGs.

moveit_ros/robot_interaction/src/robot_interaction.cpp
v4hn: It *what*? I am not aware of that.
If you start an anonymous rviz instance from the command line, you can still use the MotionPlanningDisplay.
Is there a fallback in there already to look in `move_group` too?

Either way, the rviz plugin should get the parameters from the move_group namespace.

moveit_core/robot_model/src/robot_model.cpp
v4hn: > However, the meta group both_arms, with subgroups left_arm and right_arm automatically inherits the IK solvers from their children.

Of course. But this is no general "overlap", but clear subset.
I didn't read through this code in detail,
but doesn't the code try to tackle the more general overlap case too?
I simply believe that if there is more than one primitive way to decompose a group, then the user should just specify exactly one on their own.
Do you see problems with this (abstract) requirement?

moveit_core/robot_model/src/robot_model.cpp
v4hn: You misunderstood my comment. I don't want to *get rid* of the map.
We don't want to find the correct names in a vector on each lookup.
However, there is more than one other place in this code, where we can iterate over the jmgs instead of the map...

moveit_ros/robot_interaction/src/robot_interaction.cpp
rhaschke: > If you start an anonymous rviz instance from the command line, you can still use the MotionPlanningDisplay.
> Is there a fallback in there already to look in move_group too?

Interesting question. Can you please open another issue to discuss this...

moveit_core/robot_model/src/robot_model.cpp
rhaschke: The overlap case is handled exactly as you described: the situation is detected and nothing is derived from child groups. However, in that regard, nothing is over-engineered. The existing code is required to detect the situation.

moveit_core/robot_model/src/robot_model.cpp
rhaschke: OK. I see. However, I don't want to search and fix for all these locations. I fix them as I encounter them ;-)

v4hn: That's a reasonable approach, yes...

moveit_core/robot_model/src/robot_model.cpp
rhaschke: OK. To be more precise: The code here, greedily grabs the first sub group which offers an IK solver. A potential conflict with another, later processed group isn't recognized.

The code in `computeIKIndexBijection()` will recognize the overlap and reject IK association (as you requested). However, this is never activated, because the code here only proposed disjoint sets.

As pointed out, we could accept overlaps as long as only fixed joints are overlapping. This is (probably not done yet in `computeIKIndexBijection()`).

To implement the behavior you requested, we need to move disjoint set resolution from `setSolverAllocators()` to `computeIKIndexBijection()` as pointed out in the TODO.
As we are already two of us, agreeing to this behavior, I ask you to file another issue referencing the TODO and this thread - after you merged the present PR ;-)

moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
v4hn: Agreed, "kinematic structure" is not quite correct either (although in most practical cases it is).

JMGs contain a set of joints and a set of links, yes.
What I believe to be confusing in your explanation is the "(joint,link) pairs" bit.

> I'm not sure, whether the planning group also restricts collision checking.

Yes it does.

moveit_core/robot_model/src/robot_model.cpp
v4hn: Ok, I looked into the surrounding code a bit more.
I do not agree with the first part of your TODO at the moment.
In my opinion it does not make sense *not* to select the appropriate subset in this part of the code.

`setSolverAllocators`/`computeIKIndexBijection` currently only initializes the solvers individually.
It does not reason about overlap between solver instances at all.
If you feed multiple overlapping solvers in there, they would all be setup "correctly" without complains.
If we implement your proposal, the `JointModelGroup` would be setup with more `SolverAllocatorFn`s than it actually requires and we would "set" only a subset of the arguments.

The second part of the TODO is fully valid though.
Greedy selection might fail in some (at first glance rather strange...) cases even though there could be a correct assignment.

I would propose to change the comment to:

    // TODO: How do we resolve overlaps? Now the first considered sub group "wins"
    // But, if the overlap only involves fixed joints, we could consider all sub groups

moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
rhaschke: > What I believe to be confusing in your explanation is the "(joint,link) pairs" bit

But that's the crucial point. If you add a link to the group, its parent joint is added too and vice versa. Hence, its all about (joint,link) pairs. If you manage to pack this into a single sentence, you are very welcome to do so.
Probably, we should go for another comment along this line in the joints / links widgets.

moveit_core/robot_model/src/robot_model.cpp
rhaschke: > If you feed multiple overlapping solvers in there, they would all be setup "correctly" without complains.

This is simply wrong. Did you tried it? I did with a pepper robot, defining arm groups such that they shared the common fixed torso joint. Combining both groups, I only get an interactive marker for one arm...

moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
v4hn: > If you add a link to the group, its parent joint is added too and vice versa

This is definitely not described in the assistant at the moment and I was not aware of this behavior. In this case I agree with the formulation.

>  Probably, we should go for another comment along this line in the joints / links widgets.

Yes. Could you add that please?

moveit_core/robot_model/src/robot_model.cpp
v4hn: > with a pepper robot, defining arm groups such that they shared the common fixed torso joint. Combining both groups, I only get an interactive marker for one arm

I was not talking about RViz and I'm not sure what happens there. Should it setup one interactive marker for each entry in `group_kinematics_`?
Just for clarification: You removed the logic from `setKinematicsAllocators` that only adds subsets above, and ran RViz?
With this logic intact, this is the behavior I would expect anyway.
I think we should even print a warning if a meta group has a non-empty subset of joints remaining after looking through all "simple" groups as possible subgroups.

I was talking about calls to [`setSolverAllocators`](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_core/robot_model/src/joint_model_group.cpp#L616) with the Map argument filled in.
The function just loops through all entries, configures them and registers them in `group_kinematics_`.
[`computeIKIndexBijection`](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_core/robot_model/src/joint_model_group.cpp#L575) is called for each solver, but this only sets up the joint_bijection vector *of each solver* and never changes anything in the JMG.
I don't see where an error arises in this part of the code.
This is what I meant by `would all be setup "correctly" without complains`.

rhaschke: I opened a new issue #775 to clarify the wrong behavior and continue the discussion on this there. Just for clarification here: I did not touch `setSolverAllocators` at all yet - except the cosmetic changes of this PR.

moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
rhaschke: Done. However, it wasn't possible to add this phrase in the sub widgets. It's part of the header now as well.

PR #784, opened 2018-03-01
cleanup minimum steps for jump threshold test in computeCartesianPath [fix issue #783]

mlautman:
### Description
issue #783 

With the current implementation, if you use computeCartesianPath with Eigen::Affine3d you can get different results than if you pass in a vector with a single waypoint to computeCartesianPath with EigenSTL::vector_Affine3d.  This fixes that bug. 

### Checklist

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -1773,10 +1773,6 @@ private:
   bool has_acceleration_;
   bool has_effort_;
 
-  /** \brief We at least 5 steps per trajectory for testing jump thresholds with computeCartesianPath. Otherwise
-   * testJointSpaceJump doens't work well. */
-  static const unsigned int min_steps_for_jump_thresh = 5;
-
   const JointModel* dirty_link_transforms_;
   const JointModel* dirty_collision_body_transforms_;
 
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -1773,6 +1773,10 @@ private:
   bool has_acceleration_;
   bool has_effort_;
 
+  /** \brief We at least 5 steps per trajectory for testing jump thresholds with computeCartesianPath. Otherwise
+   * testJointSpaceJump doens't work well. */
+  static const unsigned int min_steps_for_jump_thresh = 5;
+
   const JointModel* dirty_link_transforms_;
   const JointModel* dirty_collision_body_transforms_;
 
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -524,7 +524,7 @@ public:
   {
     if (has_acceleration_)
     {
-      logError("Unable to set joint efforts because array is being used for accelerations");
+      CONSOLE_BRIDGE_logError("Unable to set joint efforts because array is being used for accelerations");
       return;
     }
     has_effort_ = true;
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -166,7 +166,7 @@ bool moveit::core::RobotState::checkJointTransforms(const JointModel* joint) con
 {
   if (dirtyJointTransform(joint))
   {
-    logWarn("Returning dirty joint transforms for joint '%s'", joint->getName().c_str());
+    CONSOLE_BRIDGE_logWarn("Returning dirty joint transforms for joint '%s'", joint->getName().c_str());
     return false;
   }
   return true;
@@ -176,7 +176,7 @@ bool moveit::core::RobotState::checkLinkTransforms() const
 {
   if (dirtyLinkTransforms())
   {
-    logWarn("Returning dirty link transforms");
+    CONSOLE_BRIDGE_logWarn("Returning dirty link transforms");
     return false;
   }
   return true;
@@ -186,7 +186,7 @@ bool moveit::core::RobotState::checkCollisionTransforms() const
 {
   if (dirtyCollisionBodyTransforms())
   {
-    logWarn("Returning dirty collision body transforms");
+    CONSOLE_BRIDGE_logWarn("Returning dirty collision body transforms");
     return false;
   }
   return true;
@@ -860,7 +860,7 @@ const moveit::core::AttachedBody* moveit::core::RobotState::getAttachedBody(cons
   std::map<std::string, AttachedBody*>::const_iterator it = attached_body_map_.find(id);
   if (it == attached_body_map_.end())
   {
-    logError("Attached body '%s' not found", id.c_str());
+    CONSOLE_BRIDGE_logError("Attached body '%s' not found", id.c_str());
     return NULL;
   }
   else
@@ -1003,7 +1003,7 @@ const Eigen::Affine3d& moveit::core::RobotState::getFrameTransform(const std::st
   std::map<std::string, AttachedBody*>::const_iterator jt = attached_body_map_.find(id);
   if (jt == attached_body_map_.end())
   {
-    logError("Transform from frame '%s' to frame '%s' is not known ('%s' should be a link name or an attached body "
+    CONSOLE_BRIDGE_logError("Transform from frame '%s' to frame '%s' is not known ('%s' should be a link name or an attached body "
              "id).",
              id.c_str(), robot_model_->getModelFrame().c_str(), id.c_str());
     return identity_transform;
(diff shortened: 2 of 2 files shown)


rhaschke:
I'm not sure this was a bug. The code clearly states, that a vector of waypoints should be handled differently regarding the joint-space jump test, namely to test the whole trajectory once in the end.
Now, you removed this feature.

mlautman:
> I'm not sure this was a bug. The code clearly states, that a vector of waypoints should be handled differently regarding the joint-space jump test, namely to test the whole trajectory once in the end.

I get that it may not be a bug as it does exactly what you would expect after reading through the code but the behavior is far from desirable. 

> Now, you removed this feature.

The whole trajectory test was removed in favor of testing the individual segments. I just pushed a commit   that restores the global testJointSpaceJump check but throws a warning and returns an empty trajectory if the computed trajecrtory is too short to effectively test for joint space jumps using testJointSpaceJump.

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1773,6 +1773,10 @@ as the new values that correspond to the group */
   bool has_acceleration_;
   bool has_effort_;
 
+  /** \brief We at least 5 steps per trajectory for testing jump thresholds with computeCartesianPath. Otherwise
+   * testJointSpaceJump doens't work well. */
```
davetcoleman: Lets not use the 'we' pronoun... "At least 4 steps are used per trajectory for testing jump thresholds..."

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1773,6 +1773,10 @@ as the new values that correspond to the group */
...
+  static const unsigned int min_steps_for_jump_thresh = 5;
```
davetcoleman: std::size_t
also a static const should be ALL_CAPS
move this to the .cpp file inside the namespace but outside the class

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1922,7 +1923,10 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup* gro
 
   // decide how many steps we will need for this trajectory
   double distance = (rotated_target.translation() - start_pose.translation()).norm();
-  unsigned int steps = (test_joint_space_jump ? 5 : 1) + (unsigned int)floor(distance / max_step);
+
+  // If we are testing using the jump threshold, we always want at least min_steps_for_jump_thresh steps
+  unsigned int steps =
+      std::max(test_joint_space_jump ? min_steps_for_jump_thresh : 0, (unsigned int)floor(distance / max_step) + 1);
```
davetcoleman: break this out into an if statement for readability

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1957,6 +1961,14 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup* gro
 double moveit::core::RobotState::testJointSpaceJump(const JointModelGroup* group, std::vector<RobotStatePtr>& traj,
                                                     double jump_threshold)
 {
+  if (traj.size() < min_steps_for_jump_thresh)
+  {
+    logWarn("Truncating Cartesian path since the computed trajectory was too short to detect jumps in joint-space "
+            "Needed at least %zu steps, only got %zu. Try a lower max_step.",
+            min_steps_for_jump_thresh, traj.size());
+    traj.resize(0);
```
davetcoleman: i don't like that this 'truncates'. seems it should either 1) warn but return best attempt 2) auto fix the step size or 3) throw an error with this message.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1922,7 +1933,13 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup* gro
 
   // decide how many steps we will need for this trajectory
   double distance = (rotated_target.translation() - start_pose.translation()).norm();
-  unsigned int steps = (test_joint_space_jump ? 5 : 1) + (unsigned int)floor(distance / max_step);
+
+  // If we are testing using the jump threshold, we always want at least MIN_STEPS_FOR_JUMP_THRESH steps
+  unsigned int steps;
+  if (test_joint_space_jump)
+    steps = std::max(MIN_STEPS_FOR_JUMP_THRESH, (std::size_t)floor(distance / max_step) + 1);
```
rhaschke: This is redundant now. Could you initialize steps with
`steps = (unsigned int)floor(distance / max_step) + 1;`
and subsequently increase `if (jump_threshold > 0.0 && steps < MIN_STEPS_FOR_JUMP_THRESH)`?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -43,6 +43,17 @@
...
 #include <boost/bind.hpp>
 #include <moveit/robot_model/aabb.h>
+#include <algorithm>
+
+namespace moveit
+{
+namespace core
+{
+/** \brief At least 5 steps per trajectory for testing jump thresholds with computeCartesianPath. Otherwise
+ * testJointSpaceJump doens't work well. */
```
v4hn: typo in `doens't`

mlautman:
I changed the minimum number of steps to 10 based on some data I gathered from 300,000 calls to computeCartesianPath. image

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -44,6 +44,16 @@
...
 
+namespace moveit
+{
+namespace core
+{
+/** \brief At least 10 steps per trajectory for testing jump thresholds with computeCartesianPath. Otherwise
+ * testJointSpaceJump doesn't work well. */
+static const std::size_t MIN_STEPS_FOR_JUMP_THRESH = 10;
+}
+}
```
davetcoleman: Aside: I think I met some resistance in the past to this idea, but I'd rather wrap the whole header with the namespace to improve code readability.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -44,6 +44,16 @@
 #include <boost/bind.hpp>
 #include <moveit/robot_model/aabb.h>
 
+namespace moveit
+{
+namespace core
+{
+/** \brief At least 10 steps per trajectory for testing jump thresholds with computeCartesianPath. Otherwise
+ * testJointSpaceJump doesn't work well. */
```
davetcoleman: expand "doesn't work well."....

"Use at least 10 steps per trajectory for testing jump thresholds with computeCartesianPath(). Otherwise testJointSpaceJump() falsely reports valid path" ?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1922,7 +1932,11 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup* gro
 
   // decide how many steps we will need for this trajectory
   double distance = (rotated_target.translation() - start_pose.translation()).norm();
-  unsigned int steps = (test_joint_space_jump ? 5 : 1) + (unsigned int)floor(distance / max_step);
+
+  // If we are testing using the jump threshold, we always want at least MIN_STEPS_FOR_JUMP_THRESH steps
+  unsigned int steps = (unsigned int)floor(distance / max_step) + 1;
+  if (test_joint_space_jump && steps<MIN_STEPS_FOR_JUMP_THRESH)
```
davetcoleman: space between ``<``

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1922,7 +1932,11 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup* gro
 
   // decide how many steps we will need for this trajectory
   double distance = (rotated_target.translation() - start_pose.translation()).norm();
-  unsigned int steps = (test_joint_space_jump ? 5 : 1) + (unsigned int)floor(distance / max_step);
+
+  // If we are testing using the jump threshold, we always want at least MIN_STEPS_FOR_JUMP_THRESH steps
+  unsigned int steps = (unsigned int)floor(distance / max_step) + 1;
```
davetcoleman: don't use a C-style cast, use static_cast<>() instead

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1922,7 +1934,11 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup* gro
 
   // decide how many steps we will need for this trajectory
   double distance = (rotated_target.translation() - start_pose.translation()).norm();
-  unsigned int steps = (test_joint_space_jump ? 5 : 1) + (unsigned int)floor(distance / max_step);
+
+  // If we are testing using the jump threshold, we always want at least MIN_STEPS_FOR_JUMP_THRESH steps
+  unsigned int steps = static_cast<unsigned int>(floor(distance / max_step) + 1);
```
rhaschke: Do you still need the `static_cast`?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1957,6 +1973,12 @@ double moveit::core::RobotState::computeCartesianPath(const JointModelGroup* gro
 double moveit::core::RobotState::testJointSpaceJump(const JointModelGroup* group, std::vector<RobotStatePtr>& traj,
                                                     double jump_threshold)
 {
+  if (traj.size() < MIN_STEPS_FOR_JUMP_THRESH)
+  {
+    logWarn("The computed trajectory is too short to detect jumps in joint-space "
```
rhaschke: please update to new console_bridge API, see #787.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -991,21 +1003,23 @@ const Eigen::Affine3d& moveit::core::RobotState::getFrameTransform(const std::st
   std::map<std::string, AttachedBody*>::const_iterator jt = attached_body_map_.find(id);
   if (jt == attached_body_map_.end())
   {
-    logError("Transform from frame '%s' to frame '%s' is not known ('%s' should be a link name or an attached body "
-             "id).",
-             id.c_str(), robot_model_->getModelFrame().c_str(), id.c_str());
+    CONSOLE_BRIDGE_logError("Transform from frame '%s' to frame '%s' is not known ('%s' should be a link name or an "
+                            "attached body "
+                            "id).",
```
davetcoleman: please cleanup this string wrapping artifact

PR #789, opened 2018-03-06
Moveit commander planning scene interface fixup

bmagyar:
### Description

This PR is a part of my attempt at reviving the Python interfaces. 
I've fixed the functions adding collision objects for now. 
The refactoring follows PEP recommendations.

We should probably add tests too.

Is anyone aware of a better way of implementing this? 

### Checklist

### DIFF
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -35,11 +35,12 @@
 import rospy
 import conversions
 
-from moveit_msgs.msg import CollisionObject, AttachedCollisionObject
+from moveit_msgs.msg import PlanningScene, CollisionObject, AttachedCollisionObject
 from moveit_ros_planning_interface import _moveit_planning_scene_interface
 from geometry_msgs.msg import Pose, Point
 from shape_msgs.msg import SolidPrimitive, Plane, Mesh, MeshTriangle
 from exception import MoveItCommanderException
+from moveit_msgs.srv import ApplyPlanningScene, ApplyPlanningSceneRequest
 
 try:
     from pyassimp import pyassimp
@@ -61,6 +62,8 @@ class PlanningSceneInterface(object):
 
         self._pub_co = rospy.Publisher('/collision_object', CollisionObject, queue_size=100)
         self._pub_aco = rospy.Publisher('/attached_collision_object', AttachedCollisionObject, queue_size=100)
+        self._apply_planning_scene_diff = rospy.ServiceProxy('apply_planning_scene', ApplyPlanningScene)
+        self._apply_planning_scene_diff.wait_for_service(10.0)
 
     def add_sphere(self, name, pose, radius = 1):
         """
@@ -78,13 +81,19 @@ class PlanningSceneInterface(object):
         """
         Add a mesh to the planning scene
         """
-        self._pub_co.publish(self.__make_mesh(name, pose, filename, size))
+        collision_object = self.__make_mesh(name, pose, filename, size)
+        diff_req = self.__make_planning_scene_diff_req(collision_object)
+        self._apply_planning_scene_diff.call(diff_req)
+        self._pub_co.publish(collision_object)
 
     def add_box(self, name, pose, size = (1, 1, 1)):
         """
         Add a box to the planning scene
         """
-        self._pub_co.publish(self.__make_box(name, pose, size))
+        collision_object = self.__make_box(name, pose, size)
+        diff_req = self.__make_planning_scene_diff_req(collision_object)
+        self._apply_planning_scene_diff.call(diff_req)
+        self._pub_co.publish()
 
     def add_plane(self, name, pose, normal = (0, 0, 1), offset = 0):
         """ Add a plane to the planning scene """
@@ -98,6 +107,8 @@ class PlanningSceneInterface(object):
         co.planes = [p]
         co.plane_poses = [pose.pose]
         self._pub_co.publish(co)
+        diff_req = self.__make_planning_scene_diff_req(co)
+        self._apply_planning_scene_diff.call(diff_req)
 
     def attach_mesh(self, link, name, pose = None, filename = '', size = (1, 1, 1), touch_links = []):
         aco = AttachedCollisionObject()
@@ -284,3 +295,13 @@ class PlanningSceneInterface(object):
         co.primitives = [cylinder]
         co.primitive_poses = [pose.pose]
         return co
+
+    @staticmethod
+    def __make_planning_scene_diff_req(collision_object):
+        scene = PlanningScene()
+        scene.is_diff = True
+        scene.world.collision_objects = [collision_object]
+        planning_scene_diff_req = ApplyPlanningSceneRequest()
+        planning_scene_diff_req.scene = scene
+        return planning_scene_diff_req
+
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -37,7 +37,7 @@ import conversions
 
 from moveit_msgs.msg import CollisionObject, AttachedCollisionObject
 from moveit_ros_planning_interface import _moveit_planning_scene_interface
-from geometry_msgs.msg import PoseStamped, Pose, Point
+from geometry_msgs.msg import Pose, Point
 from shape_msgs.msg import SolidPrimitive, Plane, Mesh, MeshTriangle
 from exception import MoveItCommanderException
 
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -56,20 +56,33 @@ except:
 class PlanningSceneInterface(object):
     """ Simple interface to making updates to a planning scene """
 
-    def __init__(self, ns=''):
+    def __init__(self, ns='', synchronous=False, service_timeout=5.0):
         """ Create a planning scene interface; it uses both C++ wrapped methods and scene manipulation topics. """
         self._psi = _moveit_planning_scene_interface.PlanningSceneInterface(ns)
 
         self._pub_co = rospy.Publisher('/collision_object', CollisionObject, queue_size=100)
         self._pub_aco = rospy.Publisher('/attached_collision_object', AttachedCollisionObject, queue_size=100)
-        self._apply_planning_scene_diff = rospy.ServiceProxy('apply_planning_scene', ApplyPlanningScene)
-        self._apply_planning_scene_diff.wait_for_service(10.0)
+        self.__synchronous = synchronous
+        if self.__synchronous:
+            self._apply_planning_scene_diff = rospy.ServiceProxy('apply_planning_scene', ApplyPlanningScene)
+            self._apply_planning_scene_diff.wait_for_service(service_timeout)
+
+    def __submit(self, collision_object, attach=False):
+        if self.__synchronous:
+            diff_req = self.__make_planning_scene_diff_req(collision_object)
+            self._apply_planning_scene_diff.call(diff_req)
+        else:
+            if attach:
+                self._pub_aco.publish(collision_object)
+            else:
+                self._pub_co.publish(collision_object)
 
     def add_sphere(self, name, pose, radius = 1):
         """
         Add a sphere to the planning scene
         """
-        self._pub_co.publish(self.__make_sphere(name, pose, radius))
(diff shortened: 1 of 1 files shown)


davetcoleman:
@mlautman please review

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -63,98 +62,32 @@ def __init__(self):
...
-        co.id = name
-        co.header = pose.header
-        sphere = SolidPrimitive()
-        sphere.type = SolidPrimitive.SPHERE
-        sphere.dimensions = [radius]
-        co.primitives = [sphere]
-        co.primitive_poses = [pose.pose]
-        return co
+        self._apply_planning_scene_diff = rospy.ServiceProxy('apply_planning_scene', ApplyPlanningScene)
+        self._apply_planning_scene_diff.wait_for_service(10.0)
```
v4hn: Isn't that overly long?
I would expect 3-5 seconds is enough.

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -63,98 +62,32 @@ def __init__(self):
...
-
     def add_mesh(self, name, pose, filename, size = (1, 1, 1)):
         """
         Add a mesh to the planning scene
         """
-        self._pub_co.publish(self.__make_mesh(name, pose, filename, size))
+        collision_object = self.__make_mesh(name, pose, filename, size)
+        diff_req = self.__make_planning_scene_diff_req(collision_object)
+        self._apply_planning_scene_diff.call(diff_req)
+        self._pub_co.publish(collision_object)
```
v4hn: This is redundant and I do not believe it is a good idea to publish *and* apply.
Actually, in my opinion this should even trigger warnings saying "adding object which already exists", but I think we do not do that to make the `move_group` more robust.
I created `apply_planning_scene` as a synchronous alternative to the corresponding topic interfaces, not as an overall replacement.
The shortcoming of the service is that it does not support to update multiple `PlanningSceneMonitor`s simultaneously and I believe some users require this.

In the C++ API, I implemented this via `apply*` functions in contrast to the topic-based `add*` updates.
For the Python API, I would propose to add an optional parameter `synchronous`.
Would this work for you?

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -262,4 +197,95 @@ def get_attached_objects(self, object_ids = []):
...
+    @staticmethod
+    def __make_mesh(name, pose, filename, scale=(1, 1, 1)):
+        co = CollisionObject()
+        if pyassimp is False:
+            raise MoveItCommanderException(
+                "Pyassimp needs patch https://launchpadlibrarian.net/319496602/patchPyassim.txt")
+        try:
+            scene = pyassimp.load(filename)
+        except pyassimp.AssimpError as e:
+            raise MoveItCommanderException(e.message)
```
v4hn: I disagree.
If you don't know how to handle an exception, don't catch it.
Instead, we should document that this can be thrown by the functions which use this.

moveit_commander/src/moveit_commander/planning_scene_interface.py
bmagyar: It depends. Perhaps this could be a param of `__init__`?

moveit_commander/src/moveit_commander/planning_scene_interface.py
bmagyar: Fair point.

moveit_commander/src/moveit_commander/planning_scene_interface.py
v4hn: an option sounds good. I would still propose to change the default to `5.0` here.

moveit_commander/src/moveit_commander/planning_scene_interface.py
bmagyar: You are right, we should not do both at once. This was a change I've made to make sure my setup works when starting everything from the same launch file.

When I add a ton of sleep the topic interface works too. 

Btw looking around the code here I've discovered that `PlanningSceneInterface::addCollisionObjects` is not called at all within MoveIt. The `PlanningSceneMonitor` uses thet methods of `PlanningScene` directly.
```
$ grep -R addCollisionObjects
moveit_ros/planning_interface/planning_scene_interface/include/moveit/planning_scene_interface/planning_scene_interface.h:  void addCollisionObjects(
moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp:  void addCollisionObjects(const std::vector<moveit_msgs::CollisionObject>& collision_objects,
moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp:void PlanningSceneInterface::addCollisionObjects(const std::vector<moveit_msgs::CollisionObject>& collision_objects,
moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp:  impl_->addCollisionObjects(collision_objects, object_colors);
```

bmagyar:
High level question: the Python interface only exposes parts of the C++ interface. Should I add support for the add*, apply* functions too?

moveit_commander/src/moveit_commander/planning_scene_interface.py
bmagyar: Removed this change.

moveit_commander/src/moveit_commander/planning_scene_interface.py
bmagyar: Added 2 parameters to __init__:
 * `asynchronous=False`
 * `service_timeout=5.0`
and a `__submit()` function that decides where to let the message out depending on these params :)

bmagyar:
Ping @v4hn @mlautman

v4hn:
The patch looks good now.

> High level question: the Python interface only exposes parts of the C++ interface. Should I add support for the add*, apply* functions too?

Personally I would prefer a sane rewrite, but I support exposing more useful methods.

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -167,23 +109,23 @@ def add_plane(self, name, pose, normal = (0, 0, 1), offset = 0):
...
         if len(touch_links) > 0:
             aco.touch_links = touch_links
-        self._pub_aco.publish(aco)
+        self.__submit(aco, attach=True)
 
-    def attach_box(self, link, name, pose = None, size = (1, 1, 1), touch_links = []):
+    def attach_box(self, link, name, pose=None, size=(1, 1, 1), touch_links=[]):
         aco = AttachedCollisionObject()
-        if pose!=None:
+        if pose is not None:
```
mlautman: You shouldn't have to double negate. `if pose:` should be enough.

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -167,23 +109,23 @@ def add_plane(self, name, pose, normal = (0, 0, 1), offset = 0):
...
             aco.object = self.__make_existing(name)
         aco.link_name = link
         aco.touch_links = [link]
         if len(touch_links) > 0:
             aco.touch_links = touch_links
-        self._pub_aco.publish(aco)
+        self.__submit(aco, attach=True)
 
-    def attach_box(self, link, name, pose = None, size = (1, 1, 1), touch_links = []):
+    def attach_box(self, link, name, pose=None, size=(1, 1, 1), touch_links=[]):
```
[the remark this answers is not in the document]
mlautman: Also, this function could definitely have a docstring explaining what each parameter does. Some of the behavior is not intuitive. (eg: Why does pose -> make new object rather than move existing object)

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -51,112 +52,53 @@
...
         print("Failed to import pyassimp, see https://github.com/ros-planning/moveit/issues/86 for more info")
 
-# This is going to have more functionality; (feel free to add some!)
-# This class will include simple Python code for publishing messages for a planning scene
 
 class PlanningSceneInterface(object):
     """ Simple interface to making updates to a planning scene """
 
-    def __init__(self):
+    def __init__(self, synchronous=False, service_timeout=5.0):
```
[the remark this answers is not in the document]
davetcoleman: It looks like keeping ``synchronous=False`` is good because of backwards compatibility in ROS Kinetic - we don't want to change behavior. Perhaps add a comment to change this to True in ROS-Melodic?

henningkayser:
This has been on pause for some time but this still is a great improvement.
@v4hn do you give your approval since all your requested changes are addressed?
@bmagyar could you address the open points and rebase so we can get this merged?

bmagyar:
If there's support for merging it I can address the remaining points.

henningkayser:
> If there's support for merging it I can address the remaining points.

Great, thanks! As I see it @v4hn was happy with your changes and @mlautman's requested changes are pretty minor so you'll get it approved.

moveit_commander/src/moveit_commander/planning_scene_interface.py
rhaschke: No. There is a semantic difference between
`if pose:`
and
`if pose is not None:`
The latter compares against the `None` object, while the former considers any truth value of `pose`.

jonbinney: Agree with @rhaschke on this. Here it might not matter, but the Pose class could do something like override `__bool__()` to return `False` for poses where all elements are zero (it doesn't but it could). So it's actually more correct to say `pose is not None`

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -51,112 +52,53 @@
...
+                self._pub_co.publish(collision_object)
 
-    def add_sphere(self, name, pose, radius = 1):
+    def add_sphere(self, name, pose, radius=1):
         """
         Add a sphere to the planning scene 
         """
-        self._pub_co.publish(self.__make_sphere(name, pose, radius))
+        co = self.__make_sphere(name, pose, radius)
+        self.__submit(co, attach=False)
```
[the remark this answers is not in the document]
bmagyar: Should we add these changes to this PR or open a follow-up issue to add these features as well? See: this PR being a year old.

moveit_commander/src/moveit_commander/planning_scene_interface.py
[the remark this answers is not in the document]
bmagyar: I don't know. I've only refactored this part of the code to complain with PEP8, none of it's contents are from me. 

`attach_box()` is used in the moveit commander interpreter (I'm not sure how much use that gets) while `attach_mesh()` remains unused within the moveit codebase.

I would opt for removing both and moving this to a follow-up issue to consolidate functionality. That logic is indeed not intuitive with the current interface, I think the `add_()` interfaces should expose the `attach` flag and we should provide a separate set of methods which operate *only* on existing objects. This way we could rid of the logic around `pose is not None` implying intent.

bmagyar:
@v4hn could you please resolve the outstanding discussions if you think I addressed them?
@mlautman could you check my responses to your points please?

moveit_commander/src/moveit_commander/planning_scene_interface.py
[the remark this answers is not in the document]
mlautman: Fair enough. Maybe just add a TODO for now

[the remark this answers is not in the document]
bmagyar: I've created a todo issue #1390

moveit_commander/src/moveit_commander/planning_scene_interface.py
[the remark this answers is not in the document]
mlautman: sounds good. Maybe add a TODO if you have a moment

mlautman:
Looks good to me. I went ahead and Dismissed @v4hn's review based on his comment:

> The patch looks good now.

This brings up another question.. In general, do we have a policy where the community can dismiss stale reviews? I don't see any policy on moveit.ros.org. It would be nice to create a mechanism to move past a `requested changes` review if the reviewer isn't able to respond within a certain timeframe. Maybe something like 10 days? Maybe only a 3rd maintainer can `Dismiss review`? Perhaps we can discuss this at our next maintainer meeting @davetcoleman?

bmagyar:
@henningkayser this should be ready for merging.

davetcoleman:
i've noted the dismiss review issue for the next meeting

bmagyar:
With two green lights I'm going to merge this. I may need to resolve some Co flicta on the other moveit commander PR afterwards.

PR #793, opened 2018-03-09
MoveitCommander support for TrajectoryConstraints

bmagyar:
### Description

TLDR: 
* `trajectory_constraints` is a member of [`MotionPlanRequest`](http://docs.ros.org/kinetic/api/moveit_msgs/html/msg/MotionPlanRequest.html) yet the MoveGroupInterface doesn't allow for specifying it.
* `trajectory_constraints` can carry entire trajectories while `path_constraints` can't (in a clean way). 

----

The [`MotionPlanRequest`](http://docs.ros.org/kinetic/api/moveit_msgs/html/msg/MotionPlanRequest.html) message has a field `trajectory_constraints` but  `class MoveGroupInterface (C++)` doesn't allow for setting them. I've added support for this both to the C++ class and the Python wrapper. 
What's the motivation for this? Passing in _initial trajectories_, _initial guesses_ or _seed trajectories_ to optimization-based planners such as the case of [STOMP](https://github.com/ros-industrial/industrial_moveit/blob/kinetic-devel/stomp_moveit/src/stomp_planner.cpp#L505).

This was on #790 , now separated.

We should probably add some tests.

### Checklist

### DIFF
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -34,6 +34,7 @@
 
 from geometry_msgs.msg import Pose, PoseStamped
 from moveit_msgs.msg import RobotTrajectory, Grasp, PlaceLocation, Constraints
+from moveit_msgs.msg import MoveItErrorCodes, TrajectoryConstraints
 from sensor_msgs.msg import JointState
 import rospy
 import tf
@@ -388,6 +389,27 @@ class MoveGroupCommander(object):
     def clear_path_constraints(self):
         """ Specify that no path constraints are to be used during motion planning """
         self._g.clear_path_constraints()
+
+    def get_trajectory_constraints(self):
+        """ Get the actual trajectory constraints in form of a moveit_msgs.msgs.Constraints """
+        c = Constraints()
+        c_str = self._g.get_trajectory_constraints()
+        conversions.msg_from_string(c, c_str)
+        return c
+
+    def set_trajectory_constraints(self, value):
+        """ Specify the trajectory constraints to be used (as read from the database) """
+        if value == None:
+            self.clear_path_constraints()
+        else:
+            if type(value) is TrajectoryConstraints:
+                self._g.set_trajectory_constraints_from_msg(conversions.msg_to_string(value))
+            elif not self._g.set_trajectory_constraints(value):
+                raise MoveItCommanderException("Unable to set path constraints " + value)
+
+    def clear_trajectory_constraints(self):
+        """ Specify that no trajectory constraints are to be used during motion planning """
+        self._g.clear_trajectory_constraints()
 
     def set_constraints_database(self, host, port):
         """ Specify which database to connect to for loading possible path constraints """
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -398,14 +398,14 @@ class MoveGroupCommander(object):
         return c
 
     def set_trajectory_constraints(self, value):
-        """ Specify the trajectory constraints to be used (as read from the database) """
+        """ Specify the trajectory constraints to be used """
         if value == None:
-            self.clear_path_constraints()
+            self.clear_trajectory_constraints()
         else:
             if type(value) is TrajectoryConstraints:
                 self._g.set_trajectory_constraints_from_msg(conversions.msg_to_string(value))
             elif not self._g.set_trajectory_constraints(value):
-                raise MoveItCommanderException("Unable to set path constraints " + value)
+                raise MoveItCommanderException("Unable to set trajectory constraints " + value)
 
     def clear_trajectory_constraints(self):
         """ Specify that no trajectory constraints are to be used during motion planning """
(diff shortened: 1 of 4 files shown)


bmagyar:
Thanks for the excellent summary @v4hn ! Indeed, that is my interpretation of this field. Unfortunately even this doesn't really allow for a clean way of defining _multiple trajectories_. 

The definition of `moveit_msgs/Constraints` is as follows:
```
string name
moveit_msgs/JointConstraint[] joint_constraints
moveit_msgs/PositionConstraint[] position_constraints
moveit_msgs/OrientationConstraint[] orientation_constraints
moveit_msgs/VisibilityConstraint[] visibility_constraints
```

and while both `moveit_msgs/PositionConstraint` and `moveit_msgs/OrientationConstraint` allows to fully determine the full-DoF constraint of a given link, `moveit_msgs/JointConstraint` only addresses a single joint. This means that while a vector of Position and Orientaton constraints could determine a trajectory, a vector of PositionConstraints only determine a single joint-space waypoint.

Of course allowing for multiple Position and OrientationConstraints points beyond controlling a single point, which is a nice and general abstraction, we should keep it that way. 

I don't think this issue is critical enough to change message definitions but I'd like to mention that passing in trajectories could also be implemented by passing in a trajectory message but that implies adding a new field to `MotionPlanRequest`

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -389,6 +390,27 @@ def clear_path_constraints(self):
...
 
+    def get_trajectory_constraints(self):
+        """ Get the actual trajectory constraints in form of a moveit_msgs.msgs.Constraints """
+        c = Constraints()
+        c_str = self._g.get_trajectory_constraints()
+        conversions.msg_from_string(c, c_str)
+        return c
+
+    def set_trajectory_constraints(self, value):
+        """ Specify the trajectory constraints to be used (as read from the database) """
```
rhaschke: Which database do you refer to?

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -389,6 +390,27 @@ def clear_path_constraints(self):
...
+        if value == None:
+            self.clear_path_constraints()
```
rhaschke: Why this relays to path constraints, while the function is about trajectory constraints?

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -906,6 +906,11 @@ class MoveGroupInterface
       This removes any path constraints set in previous calls to setPathConstraints(). */
   void clearPathConstraints();
 
+  moveit_msgs::TrajectoryConstraints getTrajectoryConstraints() const;
+  bool setTrajectoryConstraints(const std::string& constraints);
```
rhaschke: This variant doesn't seem to have an implementation. I think, it's superfluous.

moveit_commander/src/moveit_commander/move_group.py
bmagyar: None actually. It's a ctrl+c&ctrl+v leftover

moveit_commander/src/moveit_commander/move_group.py
bmagyar: It's a ctrl+c&ctrl+v leftover, fixed now

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
bmagyar: Nice catch! Removed

bmagyar:
> So far, the field is not used in any planner in MoveIt. 

It is. As mentioned above: [STOMP](https://github.com/ros-industrial/industrial_moveit/blob/kinetic-devel/stomp_moveit/src/stomp_planner.cpp#L505)

>  *  @bmagyar wants to use this field as a trajectory seed for planning, while the waypoints of the finally optimized trajectory might violate the "constraints".
> * @v4hn considered them as real constraints, i.e. the final solution should satisfy them with some waypoint in the given order.

Thank you @rhaschke for summing that up. 
We have precedent for both use-cases at the moment. While the bit of code from the [benchmark](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/benchmarks/benchmarks/src/benchmark_execution.cpp#L450) that suggests `trajectory_constraints` are meant to be `waypoints` while [STOMP](https://github.com/ros-industrial/industrial_moveit/blob/kinetic-devel/stomp_moveit/src/stomp_planner.cpp#L505) uses it for `seed_trajectory`.

Short term: I think this should be merged into kinetic-devel. This is a fix for an interface that wasn't fully transparent with the C++ one before. My motivation for doing it is not relevant.

Long term: We discovered an ambiguity in the interface, we should aim to correct it. This deserves its own issue where we discuss the solution and then implement it. We should touch on this at the maintainers meeting!
My proposed solution:
 * deprecate `trajectory_constraints` from the message,
 * add `waypoints` or `waypoint_constraints`,
 * add `seed_trajectories`, `initial_trajectories`, `trajectory_guesses` -- we can agree on a name.

v4hn:
I would agree that this can be merged without resolving the underlying problem.
We provided that interface before, it was just not wrapped.

Let's talk about the rest shortly next week, but feel free to open an issue already.

I'll leave it to @rhaschke to raise further discussion here or merge.

PR #799, opened 2018-03-14
Support static TFs for multi DOF joints in CurrentStateMonitor

pHaeusler:
### Description
The tfCallback method in CurrentStateMonitor disregards any TF lookups with a common time of zero. This makes it impossible to update a floating joint with a static TF because static TFs are published at time zero.

This fix adds support for static TF updates to floating joints in CurrentStateMonitor by also considering TFs with a common time of zero.

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
@@ -443,15 +443,15 @@ void planning_scene_monitor::CurrentStateMonitor::tfCallback()
       Eigen::Affine3d eigen_transf;
       tf::transformTFToEigen(transf, eigen_transf);
 
-      double new_values[joint->getStateSpaceDimension()]
+      double new_values[joint->getStateSpaceDimension()];
       joint->computeVariablePositions(eigen_transf, new_values);
 
-      if (joint->distance(&new_values[0], robot_state_.getJointPositions(joint)) > 1e-5)
+      if (joint->distance(new_values, robot_state_.getJointPositions(joint)) > 1e-5)
       {
         changes = true;
       }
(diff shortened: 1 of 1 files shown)


v4hn:
I did not look at this yet, but Is this still relevant after https://github.com/ros-planning/moveit/pull/792 ?

pHaeusler:
Yep. Currently any transforms with a common time of zero are not used to update floating joints, which precludes static transforms. This fix allows static transforms to update floating joints.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -402,68 +402,56 @@ void planning_scene_monitor::CurrentStateMonitor::tfCallback()
...
       }
-      joint_time_[joint] = latest_updates[i];
+      joint_time_[joint] = latest_common_time;
 
       Eigen::Affine3d eigen_transf;
       tf::transformTFToEigen(transf, eigen_transf);
 
-      double new_values[joint->getStateSpaceDimension()];
-      joint->computeVariablePositions(eigen_transf, new_values);
+      std::vector<double> new_values(joint->getStateSpaceDimension());
```
rhaschke: Why did you change `double[]` to `vector<double>`? The array is created on stack, while the vector requires heap allocation which is slower.

pHaeusler: When using [cpplint](https://github.com/cpplint/cpplint) I get this error
> Do not use variable-length arrays.  Use an appropriately named ('k' followed by CamelCase) compile-time constant for the size.  [runtime/arrays] [1]

Your comment about speed is valid though, so I will change it back.

PR #801, opened 2018-03-16
MotionPlanningDisplay: check for parent widget

machinekoder:
### Description

The MotionPlanningDisplay does not check if a parent widget exists. This causes the using application to crash if no parent widget was created. This issue happens when building a custom application using the rviz library.

### Checklist
- [x] **Required**: Code is auto formatted using [clang-format]


bmagyar:
Nice catch, thank you!

PR #816, opened 2018-04-03
enable ccache for travis

rhaschke:
https://docs.travis-ci.com/user/caching/#ccache-cache

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ dist: trusty
 services:
   - docker
 language: generic
+cache: ccache
 compiler:
   - gcc
 notifications:


bmagyar:
Does this have a real effect on travis? AFAIK new travis builds done in fresh docker containers and we only build the workspace one time.

rhaschke:
> Does this have a real effect on travis?

Good point. Indeed, we might need to also symlink `$HOME/.ccache` into the docker container in https://github.com/ros-planning/moveit_ci
I filed an appropriate PR: https://github.com/ros-planning/moveit_ci/pull/30

v4hn:
Do we have a way to objectively see whether this actually works once it is merged?
Something like `ccache -s` on the travis system?

PR #823, opened 2018-04-05
MotionPlanningRvizPlugin: Enhanced External ROS Topic API

phy25:
### Description

This PR derived from the idea in [a ROS Answer question](http://answers.ros.org/question/257394/how-to-change-moveit-goal-state-through-code/): Provide external topics setting the start state and the goal state in the MotionPlanningRvizPlugin with a RobotState object. This improves the current API in which the state can only be set with the current robot position.

To achieve this, two new topics are subscribed:

  - `/rviz/moveit/update_custom_start_state`
  - `/rviz/moveit/update_custom_goal_state`

What's more, for the completeness of the external API, stopping command is exposed too:

  - `/rviz/moveit/stop`

Finally, the state of allow_external_program UI checkbox can be kept across sessions now, in order to ease the usage. The slot function adds a init check, to prevent uninitialized error.

As I am new to MoveIt! and C++, any feedback is welcome. The feature has been tested locally, and I will complete the following checklist in a week, sorry about that.

### Checklist
- [ ] Extended the tutorials / documentation, if necessary [reference](http://moveit.ros.org/documentation/contributing/) - **Nothing in the docs for now except [the joystick](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/ros_visualization/joystick.html); I will not add more docs**
- [x] Include a screenshot if changing a GUI - **N/A**

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -285,6 +285,7 @@ private:
 
   ros::Subscriber plan_subscriber_;
   ros::Subscriber execute_subscriber_;
+  ros::Subscriber stop_subscriber_;
   ros::Subscriber update_start_state_subscriber_;
   ros::Subscriber update_goal_state_subscriber_;
   ros::Subscriber update_start_state_RobotState_subscriber_;
@@ -296,6 +297,7 @@ private:
 
   void remotePlanCallback(const std_msgs::EmptyConstPtr& msg);
   void remoteExecuteCallback(const std_msgs::EmptyConstPtr& msg);
+  void remoteStopCallback(const std_msgs::EmptyConstPtr& msg);
   void remoteUpdateStartStateCallback(const std_msgs::EmptyConstPtr& msg);
   void remoteUpdateGoalStateCallback(const std_msgs::EmptyConstPtr& msg);
   void remoteUpdateStartStateRobotStateCallback(const moveit_msgs::RobotStateConstPtr& msg);
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -300,8 +300,8 @@ private:
   void remoteStopCallback(const std_msgs::EmptyConstPtr& msg);
   void remoteUpdateStartStateCallback(const std_msgs::EmptyConstPtr& msg);
   void remoteUpdateGoalStateCallback(const std_msgs::EmptyConstPtr& msg);
-  void remoteUpdateStartStateRobotStateCallback(const moveit_msgs::RobotStateConstPtr& msg);
-  void remoteUpdateGoalStateRobotStateCallback(const moveit_msgs::RobotStateConstPtr& msg);
+  void remoteUpdateCustomStartStateCallback(const moveit_msgs::RobotStateConstPtr& msg);
+  void remoteUpdateCustomGoalStateCallback(const moveit_msgs::RobotStateConstPtr& msg);
 
   /* Selects or unselects a item in a list by the item name */
   void setItemSelectionInList(const std::string& item_name, bool selection, QListWidget* list);
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -288,8 +288,8 @@ private:
   ros::Subscriber stop_subscriber_;
   ros::Subscriber update_start_state_subscriber_;
   ros::Subscriber update_goal_state_subscriber_;
-  ros::Subscriber update_start_state_RobotState_subscriber_;
-  ros::Subscriber update_goal_state_RobotState_subscriber_;
+  ros::Subscriber update_custom_start_state_subscriber_;
+  ros::Subscriber update_custom_goal_state_subscriber_;
   // General
   void changePlanningGroupHelper();
   void importResource(const std::string& path);
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -287,6 +287,8 @@ private:
   ros::Subscriber execute_subscriber_;
   ros::Subscriber update_start_state_subscriber_;
   ros::Subscriber update_goal_state_subscriber_;
+  ros::Subscriber update_start_state_RobotState_subscriber_;
+  ros::Subscriber update_goal_state_RobotState_subscriber_;
   // General
   void changePlanningGroupHelper();
   void importResource(const std::string& path);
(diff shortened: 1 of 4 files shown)


moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
```diff
@@ -285,17 +285,23 @@ private Q_SLOTS:
 
   ros::Subscriber plan_subscriber_;
   ros::Subscriber execute_subscriber_;
+  ros::Subscriber stop_subscriber_;
   ros::Subscriber update_start_state_subscriber_;
   ros::Subscriber update_goal_state_subscriber_;
+  ros::Subscriber update_start_state_RobotState_subscriber_;
+  ros::Subscriber update_goal_state_RobotState_subscriber_;
```
v4hn: Could you please change `update_*_state_RobotState` et al to `update_custom_*_state` throughout the request?
The datatype does not belong into the identifier.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -211,24 +211,36 @@ void MotionPlanningFrame::setItemSelectionInList(const std::string& item_name, b
 
 void MotionPlanningFrame::allowExternalProgramCommunication(bool enable)
 {
+  if (first_time_)
+  {
+    return;
+  }
```
v4hn: why is this needed here?

At first glance this seems unnecessary.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -442,4 +449,36 @@ void MotionPlanningFrame::remoteUpdateGoalStateCallback(const std_msgs::EmptyCon
     }
   }
 }
+
+void MotionPlanningFrame::remoteUpdateStartStateRobotStateCallback(const moveit_msgs::RobotStateConstPtr& msg)
+{
+  moveit_msgs::RobotState msg_no_attached(*msg);
```
v4hn: This is the unchanged message right now and thus redundant.

I think it is a good idea though to restrict these start state changes to forbid changes in the attached collision objects.
We don't have that possibility in the RViz GUI either
and it might cause weird state effects in the GUI.

I would propose to add
```
msg_no_attached.attached_collision_objects.clear();
msg_no_attached.is_diff = true;
```
after the new variable definition here and below.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
phy25: This is required because when resuming saved UI state (before `first_time_ == true`), this function will be executed because of the Qt signal, while something required for this doesn't complete loading (I guess), then it will segfault indefinitely. The quick fix applied here:

1. Don't execute unless `first_time_` load completes;
2. [Manually trigger the function after `first_time_` load](https://github.com/ros-planning/moveit/pull/823/files#diff-0b8bf166073217818c9c79b974152957R346).

phy25:
Renamed stuff, clang-formatted, and added the collision object remover, and local test looks good. A squash merge is preferred as there are too many commits :wrench:.
I will check whether docs update is required later.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -211,24 +211,38 @@ void MotionPlanningFrame::setItemSelectionInList(const std::string& item_name, b
...
     update_start_state_subscriber_ =
         nh.subscribe("/rviz/moveit/update_start_state", 1, &MotionPlanningFrame::remoteUpdateStartStateCallback, this);
     update_goal_state_subscriber_ =
         nh.subscribe("/rviz/moveit/update_goal_state", 1, &MotionPlanningFrame::remoteUpdateGoalStateCallback, this);
+    update_custom_start_state_subscriber_ =
+        nh.subscribe("/rviz/moveit/update_start_state_RobotState", 1,
+                     &MotionPlanningFrame::remoteUpdateCustomStartStateCallback, this);
+    update_custom_goal_state_subscriber_ =
+        nh.subscribe("/rviz/moveit/update_goal_state_RobotState", 1,
+                     &MotionPlanningFrame::remoteUpdateCustomGoalStateCallback, this);
```
v4hn: You forgot to change the topic names (`*_custom_*_state`), please do that too.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
v4hn: Not a neat solution, as you say.
It would be interesting for me to know what goes wrong here. The only thing I can imagine is that `planning_display_->getRobotInteraction()` is still a nullptr.

Please add a proper comment explaining this reasoning above the condition.

phy25: The `marker_access_lock_` in the interactive marker is the issue. I cannot think of a nicer fix for now :disappointed: 

> Thread 1 "rviz" received signal SIGSEGV, Segmentation fault.
__GI___pthread_mutex_lock (mutex=0x158) at ../nptl/pthread_mutex_lock.c:67
> `#0`  __GI___pthread_mutex_lock (mutex=0x158) at ../nptl/pthread_mutex_lock.c:67
> `#1`  0x00007ffff7a8b098 in boost::unique_lock<boost::mutex>::lock() () from /opt/ros/kinetic/lib/librviz.so
> `#2`  0x00007fff90d87a55 in robot_interaction::RobotInteraction::toggleMoveInteractiveMarkerTopic(bool) ()
   from /home/phy/moveit-clone/devel/lib/libmoveit_robot_interaction.so.0.9.11
> `#3`  0x00007fff91ececc6 in moveit_rviz_plugin::MotionPlanningFrame::allowExternalProgramCommunication(bool) ()
   from /home/phy/moveit-clone/devel/lib/libmoveit_motion_planning_rviz_plugin_core.so.0.9.11
> `#4`  0x00007ffff1f80d2a in QMetaObject::activate(QObject*, int, int, void**) () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
> `#5`  0x00007ffff77fe312 in QAbstractButton::toggled(bool) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
> `#6`  0x00007ffff7561645 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
> `#7`  0x00007ffff7562acc in QAbstractButton::setChecked(bool) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
> `#8`  0x00007fff91f4113c in moveit_rviz_plugin::MotionPlanningDisplay::load(rviz::Config const&) ()

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
phy25: Forgot to push that single commit, sorry.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
v4hn: ... which means `getRobotInteraction()` is not a valid object yet, I guess.
Fair enough. Please add a comment that explains this reasoning above the `if` here as well as in the lines below that trigger the lazy load.

phy25:
Got it. Finished.

v4hn:
Awesome, thank you for your endurance :+1: 

I approve to merge this into `kinetic-devel`. Now another maintainer can review and merge.

phy25:
Should I rebase this or someone else is working on this :joy: ?

rhaschke:
@phy25 I'm working on merging this.

rhaschke:
Merged after rebase: 83809de8ef161a60e881edbaa8f1369dc6b222fb, github didn't notice.

PR #824, opened 2018-04-05
fix race condition

rhaschke:
As pointed out in https://github.com/ros-planning/moveit/pull/350#issuecomment-378265191, the race condition in `waitForCurrentState()` was not fully resolved yet.
The reason was that only the first incoming `joint_state` update triggered a `PlanningScene` update, while later updates were throttled. To overcome this issue, we should explicitly trigger a scene update in `waitForCurrentState()`, if there were (further) updates.
As verified by @mmoerdijk, this PR finally resolves the issue. 
The unittest so far only involved fake controllers that didn't revealed the timing issue.


v4hn:
Code simplification, horray!

Merging.

PR #827, opened 2018-04-07
update include statements to use new pluginlib and class_loader headers

mikaelarguedas:
Disclaimer: This in an automated PR, if it is not relevant on this branch, apologies for the inconveniance, feel free close it.

----

`pluginlib` and `class_loader` headers have been refactored and renamed. The previous headers `.h` have been deprecated in favor of their `.hpp` equivalent. The new headers are available on all active ROS distributions and the deprecated ones will be removed in a future version of ROS (likely ROS-N).

This PR migrates the include statements to use the non-deprecated ones and should compile for any active ROS distribution starting with Indigo
The migration was done by running the scripts [pluginlib_headers_migration.py](https://github.com/ros/pluginlib/blob/6ada92c4665a392339c683682de1d1503658c209/scripts/pluginlib_headers_migration.py) and [class_loader_headers_update.py](https://github.com/ros/class_loader/blob/1515546103de4d987daa8f5519ca43fe6cffbca6/scripts/class_loader_headers_update.py) on this repository.
Note: this will not work for Jade

### DIFF
--- a/moveit_experimental/collision_distance_field/src/collision_detector_hybrid_plugin_loader.cpp
+++ b/moveit_experimental/collision_distance_field/src/collision_detector_hybrid_plugin_loader.cpp
@@ -1,5 +1,5 @@
 #include <moveit/collision_distance_field/collision_detector_hybrid_plugin_loader.h>
-#include <pluginlib/class_list_macros.h>
+#include <pluginlib/class_list_macros.hpp>
 
 namespace collision_detection
 {
--- a/moveit_experimental/kinematics_cache/v1/kinematics_cache_ros/include/kinematics_cache_ros/kinematics_cache_ros.h
+++ b/moveit_experimental/kinematics_cache/v1/kinematics_cache_ros/include/kinematics_cache_ros/kinematics_cache_ros.h
@@ -40,7 +40,7 @@
 
 #include <kinematics_cache/kinematics_cache.h>
 #include <planning_models/robot_model.h>
-#include <pluginlib/class_loader.h>
+#include <pluginlib/class_loader.hpp>
 
 namespace kinematics_cache_ros
 {
--- a/moveit_kinematics/cached_ik_kinematics_plugin/src/cached_ik_kinematics_plugin.cpp
+++ b/moveit_kinematics/cached_ik_kinematics_plugin/src/cached_ik_kinematics_plugin.cpp
@@ -40,7 +40,7 @@
 //#include <moveit/lma_kinematics_plugin/lma_kinematics_plugin.h>
 #include <moveit/srv_kinematics_plugin/srv_kinematics_plugin.h>
 
-#include <pluginlib/class_list_macros.h>
+#include <pluginlib/class_list_macros.hpp>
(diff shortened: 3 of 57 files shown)


v4hn:
CI tells us that this is not yet rolled out in indigo and kinetic. So we should not merge it yet.

mikaelarguedas:
These changes have been rolled out in indigo and kinetic during the last sync. CI fails fails because the docker images it uses have not been rebuilt since the last sync. Running an `apt upgrade` in your CI before performing the build should fix it.
More details on the sister discussion [here](https://github.com/ros-planning/warehouse_ros/pull/38#issuecomment-379487450)

rhaschke:
I retriggered Travis. The docker containers should have been rebuild 5 days ago.

mikaelarguedas:
Thanks @rhaschke.
In case it's useful: you can now Squash and Merge on the GitHub UI directly by using the drop down next to the "Merge pull request" button

mikaelarguedas:
:+1:

knorth55:
In `Kinetic`, `pluginlib`change is not released yet :( .
https://github.com/ros/pluginlib/commits/kinetic-devel
https://github.com/ros/pluginlib/commit/84b0c46be4c3f47cf995998db517cfa4700b1514
This causes build error with moveit `0.9.12`.
I will request to release new version of `pluginlib` in `kinetic`.

mikaelarguedas:
This change should not impact downstream packages as both versions of the headers are installed and released.

Moveit 0.9.12 builds successfully on the buildfarm on both Kinetic and Lunar. Can you provide more details about the build failure you are referring to ?

rhaschke:
Our CI server as well as ROS build farm can perfectly build this. Maybe you need to update your pluginlib installation?

v4hn:
@knorth55 This commit only adjusts to the new headers, the old headers are still there.
this should not trigger errors.

v4hn:
@mikaelarguedas I actually hit problems with the header changes though: pal-robotics maintains their own mirror of packages for TIAGo with a more cautious policy than OSRF.
They did not sync pluginlib (yet), so MoveIt devel (and everything else that updated the headers in kinetic) does not build there at the moment.
And no, it's *not* their fault for not updating right away...

rhaschke:
If you rely on a different package repo, you shouldn't update the MoveIt package as well...

knorth55:
Oh, sorry. I misunderstood the problem.
I will check again the build error correctly.

mikaelarguedas:
> pal-robotics maintains their own mirror of packages for TIAGo with a more cautious policy than OSRF.

Good to know, has the problem been reported to them ? If so can you please provide a link here ?

> And no, it's not their fault for not updating right away...

Maybe there's history I'm missing here, but I don't recall suggesting that.


I think the part that was missing in this PR is a version constraint on the pluginlib dependency in the `package.xml`'s of the modified packages.
Please see https://github.com/ros-planning/moveit/pull/927 for a follow-up PR

knorth55:
I realized in our jenkins, and the docker image in the jenkins was not updated.
This causes old version `pluginlib` error, I thank you all!

PR #828, opened 2018-04-07
Provide more informative error message if shape transform cannot be found

v4hn:
Moved error msg from callee (`getShapeTransform`) to caller (`maskContainment`).

Also I don't think it belongs in this lookup function in the first place.
The signature defines a bool return value for success/failure.

This builds on a change in geometric_shapes 0.5.4.
So I think this request can wait until the new version is rolled out.

### DIFF
--- a/moveit_ros/perception/point_containment_filter/src/shape_mask.cpp
+++ b/moveit_ros/perception/point_containment_filter/src/shape_mask.cpp
@@ -128,7 +128,16 @@ void point_containment_filter::ShapeMask::maskContainment(const sensor_msgs::Poi
     std::size_t j = 0;
     for (std::set<SeeShape>::const_iterator it = bodies_.begin(); it != bodies_.end(); ++it)
     {
-      if (transform_callback_(it->handle, tmp))
+      if (!transform_callback_(it->handle, tmp))
+      {
+        if (!it->body)
+          ROS_ERROR_STREAM_NAMED("shape_mask", "Missing transform for shape with handle " << it->handle << " without a "
+                                                                                                           "body");
+        else
+          ROS_ERROR_STREAM_NAMED("shape_mask", "Missing transform for shape " << it->body->getType() << " with handle "
+                                                                              << it->handle);
+      }
+      else
       {
         it->body->setPose(tmp);
         it->body->computeBoundingSphere(bspheres_[j++]);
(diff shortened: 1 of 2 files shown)


moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -151,7 +151,6 @@ bool PointCloudOctomapUpdater::getShapeTransform(ShapeHandle h, Eigen::Affine3d&
   ShapeTransformCache::const_iterator it = transform_cache_.find(h);
   if (it == transform_cache_.end())
   {
-    ROS_ERROR("Internal error. Shape filter handle %u not found", h);
```
rhaschke: Why do you drop this error msg? `getShapeTransform()` could be called in a different context (other than as a callback to `maskContainment()` as well, couldn't it?
I think, you should simply improve the message here, e.g. stating "Missing transform for shape with handle ...".

moveit_ros/perception/point_containment_filter/src/shape_mask.cpp
```diff
@@ -128,7 +128,16 @@ void point_containment_filter::ShapeMask::maskContainment(const sensor_msgs::Poi
     std::size_t j = 0;
     for (std::set<SeeShape>::const_iterator it = bodies_.begin(); it != bodies_.end(); ++it)
     {
-      if (transform_callback_(it->handle, tmp))
+      if (!transform_callback_(it->handle, tmp))
```
rhaschke: Actually, I would propose to put the error msg into the called function...

rhaschke:
> If you are unhappy that users might not see console errors anymore when they call this function, I would propose to add a warn_unused_result attribute to the function.

That's a good idea. I guess, up to now, `getShapeTransform()` is only called in the context of the callback.

v4hn:
> If you are unhappy that users might not see console errors anymore when they call this function, I would propose to add a warn_unused_result attribute to the function.
This ensures they see a compile time warning when they ignore the return value.

Ok, I looked around a bit but I don't think current compilers support attribute checks on function objects.
So I could add the attribute to the function, but it would always be ignored because we only call it via a `boost::function` binding. That's its intended use-case.

Still, I think the additional console output would help to debug problems with this in the future.

PR #831, opened 2018-04-09
computeCartesianPath: only test for jump if there's something to check

v4hn:
If the trajectory consists only of the first (current) robot state,
there is no need to test for joint space jumps and thus we avoid
printing the new warning "too few points to test" in this case.


v4hn:
Fair enough. I just wanted to fixup the unintuitive behavior in current `kinetic-devel` and did not have the time to look at #821 in detail. If #821 will be merged before the next kinetic release, It's ok to add a respective patch there.
In that case please decide which branch is the current pull-request branch @rhaschke @mlautman, so that we can add it there.

rhaschke:
Included in #843, thus closing here.

PR #841, opened 2018-04-14
MotionPlanningRvizPlugin: Save and resume approximate_IK UI setting

phy25:
### Description

In MotionPlanningRvizPlugin, the state of approximate_IK UI checkbox cannot be kept across sessions now. This patch adds the saving and resuming functionality. The slot function adds a init check, to prevent processing before `query_start_state_` and `query_goal_state_` in MotionPlanningDisplay is ready.

### Checklist
- [x] Extended the tutorials / documentation, if necessary [reference](http://moveit.ros.org/documentation/contributing/) - N/A
- [x] Include a screenshot if changing a GUI - N/A
- [x] Optional: Decide if this should be cherry-picked to other current ROS branches (Indigo, Jade, Kinetic) - Nope, Kinetic and later is fine

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -329,6 +329,8 @@ void MotionPlanningFrame::changePlanningGroupHelper()
           planning_display_->setQueryStartState(ps->getCurrentState());
           planning_display_->setQueryGoalState(ps->getCurrentState());
         }
+        // This ensures saved UI settings applied after planning_display_ is ready
+        planning_display_->useApproximateIK(ui_->approximate_ik->isChecked());
       }
     }
   }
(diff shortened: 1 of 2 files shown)


moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -199,6 +199,12 @@ MotionPlanningFrame::~MotionPlanningFrame()
 
 void MotionPlanningFrame::approximateIKChanged(int state)
 {
+  // This is needed to prevent UI event (resuming the options) triggered
+  // before query_start_state_ and query_goal_state_ is ready
+  if (first_time_)
+  {
+    return;
+  }
   planning_display_->useApproximateIK(state == Qt::Checked);
```
rhaschke: What happens, if this is called when `first_time_` is `true`?

phy25: https://github.com/ros-planning/moveit/blob/87b9a3592e16d11e76dc5b041ab611dcdf5a5381/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp#L949-L963

`query_start_state_` and `query_goal_state_` would be null (tested), no side effects.

Signals triggered before these "states" are initialized (in this case, `MotionPlanningDisplay::load()`) would not be actually executed, making the UI state not consistent with the KinematicsQueryOptions.

Because I manually triggered a "sync" in `MotionPlanningFrame::changePlanningGroupHelper()`, this is just for not executing `planning_display_->useApproximateIK()` twice. Should I remove this check?

rhaschke: Yes, I would drop this additional check.

phy25: @rhaschke Got it. Has force-pushed the changes.

PR #607, opened 2017-08-31
Cherry Picking #596  into indigo-devel

jrgnicho:
### Description

Cherry Picking #596  into indigo

PR #846, opened 2018-04-18
fix #786: set robot visible after loading

### LINKED ISSUE #786: Robot model is not displayed when loading a urdf file using the moveit setup assistant
### Description
Hi guys

I tried loading the urdf (xacro format) file from my robot. It shows the loading was successful, but I do not see the rviz model of the robot in the viewing window (right side stays black).

I first tested by installing ros-indigo-moveit, then I remembered that I was able to load that same model in the moveit setup assistant in May 2017. I therefore removed ros-indigo-moveit, and cloned this moveit source repo. Then I tried to checkout to some commits along the way, and I could identify that commit e9e069d [https://github.com/ros-planning/moveit/commit/e9e069d3b98e576edd4f127ead2d558150eb9894](url) was where the issue was introduced. 

Is it possible to revert some changes that were made in this commit? Or else, what's the trick to load the urdf model and get the rviz visualisation working?

Thanks a lot
Martine

Overview of your issue here.

### Your environment
* ROS Distro: Indigo
* OS Version: e.g. Ubuntu 14.04
* Source or Binary build? both source and binary built version tried

rhaschke:
This fixes issue #786, that is due to a missing `robot_->setVisible(true)`, which was lost in #607.

### DIFF
--- a/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
+++ b/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
@@ -396,6 +396,7 @@ void RobotStateDisplay::update(float wall_dt, float ros_dt)
   {
     loadRobotModel();
     changedRobotStateTopic();
+    robot_->setVisible(true);
   }
 
   calculateOffsetPosition();


OmkarKabadagi5823:
can someone tell me how to fix this in ros melodic?

rhaschke:
@OmkarKabadagi5823 This bug doesn't exist in Melodic. If you observe an issue, please file a new issue exactly describing your current problem.

S0UL4:
i have the same problem in Kinetic :'( the robot is not showing in Rviz i tried to export LC_NUMERIC like what they have done but it doesn't work , do this exist in Kinetic ?robot_->setVisible(true)

PR #858, opened 2018-04-25
improve [get|set]JointValueTarget in python wrapper

rhaschke:
As pointed out in #845, the signature and semantics of `getJointValueTarget()` doesn't match the one of the corresponding setter, which renders both functions rather useless. As I pointed out in https://github.com/ros-planning/moveit/pull/845#issuecomment-382644155 a proper solution of this issue, should consider: 

- Add a new method `MoveGroupInterface::getJointValueTarget(std::vector<double>&)` that matches the corresponding `setJointValueTarget(const std::vector<double>&)` method and should use `RobotState::copyJointGroupPositions(const JointModelGroup*, std::vector<double>&)` to do so.
- Depreacate the old method `MoveGroupInterface::getJointValueTarget()`
(currently returning the full `RobotState`) and 
- Instead add a better-named method `MoveGroupInterface::getJointStateTarget()` to correctly reflect the semantics of returning the full robot state.
- To make the difference even clearer, one could think about renaming the getter to `getTargetRobotState()`, both in `MoveGroupInterface` and `MoveGroupInterfaceImpl`.
- Deprecate the dangerous functions `setJointValueTarget(const robot_state::RobotState&)` and `setJointValueTarget(const sensor_msgs::JointState&)`, which allow to set joint values that are not member of the `JointModelGroup`.

This PR considers all of them as well as a sanity check in some other joint setters.

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -385,7 +385,7 @@ public:
 
       If these values are out of bounds then false is returned BUT THE VALUES
       ARE STILL SET AS THE GOAL. */
-  MOVEIT_DEPRECATED bool setJointValueTarget(const robot_state::RobotState& robot_state);
+  bool setJointValueTarget(const robot_state::RobotState& robot_state);
 
   /** \brief Set the JointValueTarget and use it for future planning requests.
 
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -368,7 +368,7 @@ public:
 
       If these values are out of bounds then false is returned BUT THE VALUES
       ARE STILL SET AS THE GOAL. */
-  bool setJointValueTarget(const robot_state::RobotState& robot_state);
+  MOVEIT_DEPRECATED bool setJointValueTarget(const robot_state::RobotState& robot_state);
 
   /** \brief Set the JointValueTarget and use it for future planning requests.
 
@@ -406,7 +406,7 @@ public:
 
       If these values are out of bounds then false is returned BUT THE VALUES
       ARE STILL SET AS THE GOAL. */
-  bool setJointValueTarget(const sensor_msgs::JointState& state);
+  MOVEIT_DEPRECATED bool setJointValueTarget(const sensor_msgs::JointState& state);
 
   /** \brief Set the joint state goal for a particular joint by computing IK.
 
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -330,9 +330,8 @@ public:
 
   /** \brief Set the JointValueTarget and use it for future planning requests.
 
-      \e group_variable_values MUST contain exactly one value per joint
-      variable in the same order as returned by
-      getJointValueTarget().getJointModelGroup(getName())->getVariableNames().
+      \e group_variable_values MUST exactly match the variable order as returned by
+      getJointValueTarget(std::vector<double>&).
 
       This always sets all of the group's joint values.
 
@@ -494,6 +493,9 @@ public:
   /** \brief Set the current joint values to be ones previously remembered by rememberJointValues() or, if not found,
       that are specified in the SRDF under the name \e name as a group state*/
   bool setNamedTarget(const std::string& name);
+
+  /** \brief Get the current joint state goal in a form compatible to setJointValueTarget() */
+  void getJointValueTarget(std::vector<double>& group_variable_values) const;
 
   /// Get the currently set joint state goal
   const robot_state::RobotState& getJointValueTarget() const;
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -497,8 +497,8 @@ public:
   /** \brief Get the current joint state goal in a form compatible to setJointValueTarget() */
   void getJointValueTarget(std::vector<double>& group_variable_values) const;
 
-  /// Get the currently set joint state goal
-  const robot_state::RobotState& getJointValueTarget() const;
+  /// Get the currently set joint state goal, replaced by private getTargetRobotState()
+  MOVEIT_DEPRECATED const robot_state::RobotState& getJointValueTarget() const;
 
   /**@}*/
 
@@ -931,6 +931,10 @@ public:
   void clearTrajectoryConstraints();
 
   /**@}*/
+
+protected:
+  /** return the full RobotState of the joint-space target, only for internal use */
+  const robot_state::RobotState& getTargetRobotState() const;
 
 private:
   std::map<std::string, std::vector<double> > remembered_joint_values_;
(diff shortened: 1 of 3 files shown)


rhaschke:
@willcbaker Please review.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
```diff
@@ -114,24 +114,25 @@ class MoveGroupInterfaceWrapper : protected py_bindings_tools::ROScppInitializer
...
 
   bp::list getJointValueTargetPythonList()
   {
-    const robot_state::RobotState& values = moveit::planning_interface::MoveGroupInterface::getJointValueTarget();
+    std::vector<double> values;
+    MoveGroupInterface::getJointValueTarget(values);
     bp::list l;
-    for (const double *it = values.getVariablePositions(), *end = it + getVariableCount(); it != end; ++it)
-      l.append(*it);
+    for (const double value : values)
```
henningkayser:
```suggestion
    for (const double& value : values)
```

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -400,7 +399,7 @@ class MoveGroupInterface
 
       If these values are out of bounds then false is returned BUT THE VALUES
       ARE STILL SET AS THE GOAL. */
-  bool setJointValueTarget(const sensor_msgs::JointState& state);
+  MOVEIT_DEPRECATED bool setJointValueTarget(const sensor_msgs::JointState& state);
```
henningkayser: Why not just change the implementation to match `setJointValueTarget(const std::map<std::string, double>& variable_values)`? It would probably make sense to use a `setJointValueTarget(const std::vector<std::string>& variable_names, const std::vector<double>& variable_names)` for that.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -362,7 +361,7 @@ class MoveGroupInterface
 
       If these values are out of bounds then false is returned BUT THE VALUES
       ARE STILL SET AS THE GOAL. */
-  bool setJointValueTarget(const robot_state::RobotState& robot_state);
+  MOVEIT_DEPRECATED bool setJointValueTarget(const robot_state::RobotState& robot_state);
```
henningkayser: I agree that these functions should not set any positions other than those inside the corresponding joint model group.
Rather than making these deprecated I would really prefer changing the implementation to just use `RobotState::copyJointGroupPositions()` or `RobotState::getVariablePositions()` selectively.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
rhaschke: It doesn't make sense to use references for plain types. This adds a superfluous dereference and doesn't save any copying effort.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
rhaschke: There is already a `setJointValueTarget(const std::map<std::string, double>& variable_values)`.
I'm not sure about your intention.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
rhaschke: I my opinion, the RobotState shouldn't be exposed by move-group methods.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
henningkayser: My intention is to keep support for setting target states from robot trajectory waypoints directly (from both message and class types). I don't get why `setJointValueTarget(const sensor_msgs::JointState& state)` should be deprecated only because there can be joints that are not member of the group. IMO this function should just ignore non-member joints just like `setJointValueTarget(const std::map<std::string, double>& variable_values)` does.
Alternatively I would suggest adding a `setJointValueTarget(const std::vector<std::string>& variable_names, const std::vector<double>& variable_names)` function that is more convenient to use than the map.

rhaschke: > My intention is to keep support for setting target states from robot trajectory waypoints directly

Makes sense. Please feel free to augment to this PR branch.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
henningkayser: In that case `setStartState(robot_state::RobotState)` should be deprecated as well. But does this really improve usability?

willcbaker:
Iirc I had added robotstate python bindings a while back to expose the robot state methods as at the time it was one of the only ways to get the current full state of the robot and set multidof joint goals.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -915,6 +934,10 @@ class MoveGroupInterface
 
   /**@}*/
 
+protected:
+  /** return the full RobotState of the joint-space target, only for internal use */
```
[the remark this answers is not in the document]
rhaschke: @mlautman, you already questioned the decision to make this function protected. The hint "for internal use only", will remind us that we explicitly have chosen this function to be protected.

henningkayser:
> Iirc I had added robotstate python bindings a while back to expose the robot state methods as at the time it was one of the only ways to get the current full state of the robot and set multidof joint goals.

@willcbaker so you agree with this changes or not?
@rhaschke could you fix CI so we can get this merged?

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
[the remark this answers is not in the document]
mlautman: Only a nit-pick. Feel free to disregard it

willcbaker:
I haven't had much time to play around with this yet, but it from what I can tell by reading:


RE: The get target robot state being protected:
- [ ] The set_start_state is difficult to use (in general). And to @henningkayser 's point https://github.com/ros-planning/moveit/pull/858#discussion_r262228581 this does not increase usability. The public method was used to retrieve potential future robot states and nicely complemented this function.
 I had in the past been using this method in the python wrapper as a way to concatenate plans together to make the robot do complex behaviors, e.g.
```
group.set_joint_value_target(x0,x1,..xn)
plan  = group.plan()
next_start_state = function_to_extract_robot_state_from_trajectory(plan)
group.set_start_state(next_start_state)
group.set_joint_value_target(y0,y1,..yn)
plan2 = group.plan()
complete_plan = function_to_concatenate_plans(plan, plan2)
group.execute(complete_plan)
```
I had also been using the robot state functions to create robot markers from robot state messages, I'm sure there are other ways to do this, but it was a nice to have at this level. (was heading down a route for more interactive planning other than the rviz gui, future work to come!)

RE: Deprecating robot state joint value setter:
- [ ] can we confirm we can set multi dof joints via the setJointValue target?

henningkayser:
@willcbaker Your described behavior would still be possible if the robot state is removed from the interface. You just have to pass the joint states instead of the full `RobotState` message. Also the waypoint from a `RobotTrajectory` message returned from `plan()` would have to be converted either way.
 
> can we confirm we can set multi dof joints via the setJointValue target?

I think this wouldn't be possible anymore. But what is the use case for this? To my knowledge it's not possible to plan for different multi-dof joint poses via the MoveGroupInterface.

rhaschke:
I rebased and targeted this PR against `master`. As `setJointValueTarget(RobotState)` is still used in the rviz plugin (as pointed out by the Travis warning), I suggest to revert the deprecation of this function as well, but not expose it in the Python interface anymore, i.e. remove `set_state_value_target`.

henningkayser:
@rhaschke do you want this to be cherry-picked into other branches as well?

rhaschke:
As this changes API, I wouldn't backport this to any other branch. @henningkayser, thanks for merging.

Issue #842, opened 2018-04-16
Issue while running move_group_python_interface_tutorial.py


sakshri20:
I am new to both moveit and python and I am following this tutorial:
http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/pr2_tutorials/planning/scripts/doc/move_group_python_interface_tutorial.html 

but I am getting an error while executing the following command:
rosrun moveit_tutorials move_group_python_interface_tutorial.py

Here is the console output:

============ Starting tutorial setup
[ INFO] [1523909730.395710578]: Loading robot model 'pr2'...
[ INFO] [1523909730.730948123]: Loading robot model 'pr2'...
[ WARN] [1523909730.747445345]: The root link base_footprint has an inertia specified in the URDF, but KDL does not support a root link with an inertia.  As a workaround, you can add an extra dummy link to your URDF.
[ INFO] [1523909730.773330740]: Loading robot model 'pr2'...
[ WARN] [1523909730.790229082]: The root link base_footprint has an inertia specified in the URDF, but KDL does not support a root link with an inertia.  As a workaround, you can add an extra dummy link to your URDF.
[ INFO] [1523909730.814649078]: Loading robot model 'pr2'...
[ WARN] [1523909730.831617816]: The root link base_footprint has an inertia specified in the URDF, but KDL does not support a root link with an inertia.  As a workaround, you can add an extra dummy link to your URDF.
[ INFO] [1523909730.857104219]: Loading robot model 'pr2'...
[ WARN] [1523909730.874879240]: The root link base_footprint has an inertia specified in the URDF, but KDL does not support a root link with an inertia.  As a workaround, you can add an extra dummy link to your URDF.
Traceback (most recent call last):
  File "/home/sakshi/ws_moveit/src/moveit_tutorials/doc/pr2_tutorials/planning/scripts/move_group_python_interface_tutorial.py", line 243, in <module>
    move_group_python_interface_tutorial()
  File "/home/sakshi/ws_moveit/src/moveit_tutorials/doc/pr2_tutorials/planning/scripts/move_group_python_interface_tutorial.py", line 70, in move_group_python_interface_tutorial
    scene = moveit_commander.PlanningSceneInterface()
  File "/home/sakshi/ws_moveit/src/moveit/moveit_commander/src/moveit_commander/planning_scene_interface.py", line 62, in __init__
    self._psi = _moveit_planning_scene_interface.PlanningSceneInterface()
[...]

PR #861, opened 2018-04-26
Fix and Test for #842

BryceStevenWilley:
### Description

Fixes an interface change on the C++ side of PlanningSceneInterface that was not changed on the python side. Adds a test that fails without this change.

### Checklist


v4hn:
Without further comments.
Thank you and sorry for the trouble.

PR #778, opened 2018-02-19
Fix dependencies for chomp interface test

stwirth:
### Description
Fixes #755.

Trivial change: added missing include and dependencies in `CMakeLists.txt` and `package.xml` of chomp_interface.

Without this, `catkin_make run_tests_moveit_planners_chomp` fails with:
```
[100%] Built target moveit_move_group_interface
Scanning dependencies of target chomp_moveit_test
[100%] Building CXX object moveit/moveit_planners/chomp/chomp_interface/CMakeFiles/chomp_moveit_test.dir/test/chomp_moveit_test.cpp.o
/tmp/moveit_ws/src/moveit/moveit_planners/chomp/chomp_interface/test/chomp_moveit_test.cpp:40:62: fatal error: moveit/move_group_interface/move_group_interface.h: No such file or directory
compilation terminated.
moveit/moveit_planners/chomp/chomp_interface/CMakeFiles/chomp_moveit_test.dir/build.make:62: recipe for target 'moveit/moveit_planners/chomp/chomp_interface/CMakeFiles/chomp_moveit_test.dir/test/chomp_moveit_test.cpp.o' failed
make[3]: *** [moveit/moveit_planners/chomp/chomp_interface/CMakeFiles/chomp_moveit_test.dir/test/chomp_moveit_test.cpp.o] Error 1
CMakeFiles/Makefile2:15547: recipe for target 'moveit/moveit_planners/chomp/chomp_interface/CMakeFiles/chomp_moveit_test.dir/all' failed
make[2]: *** [moveit/moveit_planners/chomp/chomp_interface/CMakeFiles/chomp_moveit_test.dir/all] Error 2
CMakeFiles/Makefile2:15483: recipe for target 'moveit/moveit_planners/chomp/chomp_interface/CMakeFiles/run_tests_moveit_planners_chomp.dir/rule' failed
make[1]: *** [moveit/moveit_planners/chomp/chomp_interface/CMakeFiles/run_tests_moveit_planners_chomp.dir/rule] Error 2
Makefile:3930: recipe for target 'run_tests_moveit_planners_chomp' failed
make: *** [run_tests_moveit_planners_chomp] Error 2
Invoking "make run_tests_moveit_planners_chomp -j4 -l4" failed
```

With this change, running `catkin_make run_tests_moveit_planners_chomp` on a clean workspace containing just moveit runs through the test.

### Checklist
- [ ] **Required**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
[...]

PR #863, opened 2018-04-26
fix various cmake warnings

rhaschke:
This PR fixes various cmake warnings throughout the source code, most notably those introduced in #778.

### DIFF
--- a/moveit_kinematics/CMakeLists.txt
+++ b/moveit_kinematics/CMakeLists.txt
@@ -21,9 +21,6 @@ find_package(catkin REQUIRED COMPONENTS
   tf_conversions
 )
 
-find_package(trac_ik_kinematics_plugin)
-find_package(ur_kinematics QUIET)
-
 find_package(PkgConfig REQUIRED)
 pkg_search_module(EIGEN3 REQUIRED eigen3)
 
--- a/moveit_planners/chomp/chomp_interface/CMakeLists.txt
+++ b/moveit_planners/chomp/chomp_interface/CMakeLists.txt
@@ -30,7 +30,8 @@ catkin_package(
 include_directories(
   include
   ${catkin_INCLUDE_DIRS}
-  ${Boost_INCLUDE_DIRS} ${EIGEN3_INCLUDE_DIRS}
+  ${Boost_INCLUDE_DIRS}
+  ${EIGEN3_INCLUDE_DIRS}
 )
 
 add_library(${PROJECT_NAME} src/chomp_interface.cpp src/chomp_planning_context.cpp)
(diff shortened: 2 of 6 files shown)


moveit_kinematics/CMakeLists.txt
```diff
@@ -21,9 +21,6 @@ find_package(catkin REQUIRED COMPONENTS
   tf_conversions
 )
 
-find_package(trac_ik_kinematics_plugin)
```
rhaschke: These are only required for `cached_ik_kinematics_plugin`.

moveit_planners/chomp/chomp_interface/package.xml
```diff
@@ -20,7 +20,6 @@
   <build_depend>chomp_motion_planner</build_depend>
   <build_depend>moveit_experimental</build_depend>
 
-  <test_depend>moveit_ros_manipulation</test_depend>
```
rhaschke: obviously not required

moveit_planners/chomp/chomp_interface/CMakeLists.txt
```diff
@@ -55,11 +56,11 @@ install(TARGETS ${PROJECT_NAME} chomp_planner_plugin
...
 if(CATKIN_ENABLE_TESTING)
-  find_package(catkin REQUIRED COMPONENTS rostest roscpp moveit_ros_planning_interface)
+  # additional packages needed for testing
+  find_package(catkin REQUIRED COMPONENTS rostest moveit_ros_planning_interface)
   include_directories(${catkin_INCLUDE_DIRS})
   add_rostest_gtest(chomp_moveit_test
     test/chomp_moveit.test
     test/chomp_moveit_test.cpp)
   target_link_libraries(chomp_moveit_test ${catkin_LIBRARIES})
-  add_dependencies(chomp_moveit_test chomp_planner_plugin move_group moveit_move_group_default_capabilities moveit_move_group_pick_place_capability)
```
rhaschke: This is nonsense. All those targets are not known to cmake in isolated builds.

moveit_kinematics/CMakeLists.txt
v4hn: I usually prefer to have all `find_package` calls for a project in a single file even if they are only used in some subfolder(s). This avoids confusion when new users look for options/dependencies in the cmake structure.
In this case I don't have a strong opinion though.

moveit_planners/chomp/chomp_interface/CMakeLists.txt
```diff
@@ -55,11 +56,11 @@ install(TARGETS ${PROJECT_NAME} chomp_planner_plugin
 )
 
 if(CATKIN_ENABLE_TESTING)
-  find_package(catkin REQUIRED COMPONENTS rostest roscpp moveit_ros_planning_interface)
+  # additional packages needed for testing
+  find_package(catkin REQUIRED COMPONENTS rostest moveit_ros_planning_interface)
```
v4hn: shouldn't this be
```
find_package(rostest REQUIRED)
find_package(moveit_ros_planning_interface REQUIRED)
```
?
We already `find_package` catkin further above and one shouldn't lookup the same package twice with different options.

rhaschke: I just moved the corresponding `find_package` calls as they were. Changing this, requires to explicitly refer to package-specific variables instead of using the convenient `catkin_*` vars.

v4hn:
> Changing this, requires to explicitly refer to package-specific variables instead of using the convenient `catkin_*` vars.

Yes. However I don't think it is well-defined what happens if you find_package a module with different components twice.

PR #875, opened 2018-05-01
[Fix] fix wrong logging arguments

jeasinema:
### Description

This PR fixes the issue in [#874 (comment)](https://github.com/ros-planning/moveit/pull/874#issuecomment-385704979). I need to figure out that none of the wrong arguments are introduced by me.

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,23 +3,22 @@ sudo: required
 dist: trusty
 services:
   - docker
-language: generic
+language: cpp
 cache: ccache
-compiler:
-  - gcc
+compiler: gcc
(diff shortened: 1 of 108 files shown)


rhaschke:
@jeasinema Sorry to blame you for the warnings. Indeed, they were just hidden by `console_bridge` before. For this reason, extra thanks to you to fix those issues too.
But: This PR is not based on latest kinetic. I rebased in #876 and changed conversion specifiers to `%zu` which is portable between 32 and 64 bit.
Closing this in favour of #876.

PR #876, opened 2018-05-01
fix printf conversion specifiers for logging 

rhaschke:
#874 revealed some issues in logging, which were hidden previously by `console_bridge`.
For `size_t` types, `%zu` should be used as conversion specifier to be portable between 32 and 64 bit:
https://stackoverflow.com/questions/2524611/how-can-one-print-a-size-t-variable-portably-using-the-printf-family


bmagyar:
Why not go directly with stream operators?

rhaschke:
Because the comment is usually harder to read if interlaced with variables.

v4hn:
I agree with Bence, stream operators are the much nicer representation.
If it would not have been much more work for questionable profit,
I would have asked to convert all of the printf style comments in the previous pull request.

PR #877, opened 2018-05-02
Can now get the jacobian of a child link of a move group.

BryceStevenWilley:
### Description

Originally, if you were to try to get the Jacobian of a link that was at the end of a move group, but not a part of that move group (say a point on a gripper link at the end of a UR robot), then MoveIt would complain that this link was not a part of the move group, even though you should be able to find the jacobian for this point.

This change fixes that, and allows you to get the jacobian of any link that is a child of the move group being queried.

Don't quite have the time to make a test for this, considering there aren't any jacobian tests yet.

### Checklist

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -1151,21 +1151,6 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
   Eigen::Vector3d joint_axis;
   Eigen::Affine3d joint_transform;
 
-  // Travel up the chain to ensure link is a part of group.
-  const LinkModel* given_link = link;
-  while (not group->hasLinkModel(link->getName()))
-  {
-    link = link->getParentLinkModel();
-    if (not link)
-    {
-      ROS_ERROR_NAMED("robot_state",
-                      "Given link '%s' is a child of the chain, but the chain is never reach when going up.",
-                      given_link->getName().c_str());
-      return false;
-    }
-  }
-  // Link is now guaranteed to be a part of group.
-
   while (link)
   {
     /*
@@ -1180,8 +1165,8 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
     {
       if (not group->hasJointModel(pjm->getName()))
       {
-        ROS_ERROR_NAMED("robot_state", "Joint %s is not in the current group.", pjm->getName().c_str());
-        return false;
+        link = pjm->getPalentLinkModel();
+        continue;
       }
       unsigned int joint_index = group->getVariableGroupIndex(pjm->getName());
       if (pjm->getType() == robot_model::JointModel::REVOLUTE)
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -1151,6 +1151,21 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
   Eigen::Vector3d joint_axis;
   Eigen::Affine3d joint_transform;
 
+  // Travel up the chain to ensure link is a part of group.
+  const LinkModel* given_link = link;
+  while (not group->hasLinkModel(link->getName()))
+  {
+    link = link->getParentLinkModel();
+    if (not link)
+    {
+      ROS_ERROR_NAMED("robot_state",
+                      "Given link '%s' is a child of the chain, but the chain is never reach when going up.",
+                      given_link->getName().c_str());
+      return false;
+    }
+  }
+  // Link is now guaranteed to be a part of group.
+
   while (link)
   {
     /*
@@ -1163,6 +1178,11 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
     const JointModel* pjm = link->getParentJointModel();
     if (pjm->getVariableCount() > 0)
     {
+      if (not group->hasJointModel(pjm->getName()))
+      {
+        ROS_ERROR_NAMED("robot_state", "Joint %s is not in the current group.", pjm->getName().c_str());
+        return false;
+      }
       unsigned int joint_index = group->getVariableGroupIndex(pjm->getName());
       if (pjm->getType() == robot_model::JointModel::REVOLUTE)
       {
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -1165,7 +1165,7 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
     {
       if (not group->hasJointModel(pjm->getName()))
       {
-        link = pjm->getPalentLinkModel();
+        link = pjm->getParentLinkModel();
         continue;
       }
       unsigned int joint_index = group->getVariableGroupIndex(pjm->getName());


moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1151,6 +1151,21 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
   Eigen::Vector3d joint_axis;
   Eigen::Affine3d joint_transform;
 
+  // Travel up the chain to ensure link is a part of group.
+  const LinkModel* given_link = link;
+  while (not group->hasLinkModel(link->getName()))
+  {
+    link = link->getParentLinkModel();
+    if (not link)
```
rhaschke: It was already checked in [line 1126](https://github.com/ros-planning/moveit/pull/877/files#diff-15d4dc0e9e1eee143f7576803dfe8ea2L1126) that link is updated by this group.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1163,6 +1178,11 @@ bool RobotState::getJacobian(const JointModelGroup* group, const LinkModel* link
     const JointModel* pjm = link->getParentJointModel();
     if (pjm->getVariableCount() > 0)
     {
+      if (not group->hasJointModel(pjm->getName()))
```
rhaschke: Wouldn't it suffice to continue here with `link = pjm->getParentLinkModel()` to skip all joints that are not part of group (and thus assume they are fixed)?

BryceStevenWilley: That is a better idea, I'll make that change soon.

moveit_core/robot_state/src/robot_state.cpp
BryceStevenWilley: Yes, but it doesn't check that this link/its parent joint are a part of the group, which is an assumption made several lines down at `group->getVariableGroupIndex(pjm->getName());` (see below comment for full explanation.

BryceStevenWilley:
> What does group->getVariableGroupIndex(pjm->getName()) return in this case? An exception?

It prints an error, `[ERROR] [1525283793.492014912, 313.314000000]: Variable 'robotiq_85_right_finger_tip_joint' is not part of group 'manipulator'`, and returns -1. The actual error happens several lines down when it tries to index an Eigen matrix with the returned -1. Here's a partial stack trace. 

```
#0  0x00007ffff5356428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff535802a in __GI_abort () at abort.c:89
#2  0x00007ffff534ebd7 in __assert_fail_base (fmt=<optimized out>, 
    assertion=assertion@entry=0x7ffff44e2598 "startRow >= 0 && BlockRows >= 1 && startRow + BlockRows <= xpr.rows() && startCol >= 0 && BlockCols >= 1 && startCol + BlockCols <= xpr.cols()", 
    file=file@entry=0x7ffff44e2568 "/usr/include/eigen3/Eigen/src/Core/Block.h", line=line@entry=134, 
    function=function@entry=0x7ffff44ee120 <Eigen::Block<Eigen::Matrix<double, -1, -1, 0, -1, -1>, 3, 1, false>::Block(Eigen::Matrix<double, -1, -1, 0, -1, -1>&, long, long)::__PRETTY_FUNCTION__> "Eigen::Block<XprType, BlockRows, BlockCols, InnerPanel>::Block(XprType&, Eigen::Index, Eigen::Index) [with XprType = Eigen::Matrix<double, -1, -1>; int BlockRows = 3; int BlockCols = 1; bool InnerPane"...) at assert.c:92
#3  0x00007ffff534ec82 in __GI___assert_fail (
    assertion=0x7ffff44e2598 "startRow >= 0 && BlockRows >= 1 && startRow + BlockRows <= xpr.rows() && startCol >= 0 && BlockCols >= 1 && startCol + BlockCols <= xpr.cols()", 
    file=0x7ffff44e2568 "/usr/include/eigen3/Eigen/src/Core/Block.h", line=134, 
    function=0x7ffff44ee120 <Eigen::Block<Eigen::Matrix<double, -1, -1, 0, -1, -1>, 3, 1, false>::Block(Eigen::Matrix<double, -1, -1, 0, -1, -1>&, long, long)::__PRETTY_FUNCTION__> "Eigen::Block<XprType, BlockRows, BlockCols, InnerPanel>::Block(XprType&, Eigen::Index, Eigen::Index) [with XprType = Eigen::Matrix<double, -1, -1>; int BlockRows = 3; int BlockCols = 1; bool InnerPane"...) at assert.c:101
#4  0x00007ffff444e7c8 in Eigen::Block<Eigen::Matrix<double, -1, -1, 0, -1, -1>, 3, 1, false>::Block (
    this=0x7fffb1412e90, xpr=..., startRow=0, startCol=4294967295)
    at /usr/include/eigen3/Eigen/src/Core/Block.h:133
#5  0x00007ffff4448249 in Eigen::DenseBase<Eigen::Matrix<double, -1, -1, 0, -1, -1> >::block<3, 1> (
    this=0x7fffb1413260, startRow=0, startCol=4294967295)
    at /usr/include/eigen3/Eigen/src/plugins/BlockMethods.h:725
#6  0x00007ffff443b282 in moveit::core::RobotState::getJacobian (this=0x7fffa000dda0, group=0xb5aa40, 
    link=0xaca5d0, reference_point_position=..., jacobian=..., use_quaternion_representation=false)
    at /home/brycew/Desktop/moveit_ws/src/moveit/moveit_core/robot_state/src/robot_state.cpp:1175
#7  0x00007fffd8464fbd in moveit::core::RobotState::getJacobian (this=0x7fffa000dda0, group=0xb5aa40, 
    link=0xaca5d0, reference_point_position=..., jacobian=..., use_quaternion_representation=false)
    at /home/brycew/Desktop/moveit_ws/src/moveit/moveit_core/robot_state/include/moveit/robot_state/robot_state.h:
```

> I think this PR is way too redundant. It should suffice to skip joints that are not part of the group.

I see your point now, I tend to prefer double checking things when unsure.  I made the suggested changes, and they work, but I won't be able to push them to this PR until later tonight.

v4hn:
The change looks good.

This still misses the case where a static sibling link of a link in the group is requested.
The required information is already collected in `LinkModel::getAssociatedFixedTransforms`.
Would it be too much to ask for this extension here too? :-)

BryceStevenWilley:
@v4hn I'd be happy to add that case, but let me make sure I understand what case you are talking about. Say for 3 links, A, B, and C, where A is a parent to both B and C with fixed joints between A-B and A-C, and B is the start of some joint group.  If the Jacobian of a point in C is requested, we should recognize that C is a static sibling of B and compute the Jacobian from there? 

Given my understanding is correct, I'm not sure what to do, as `getAssociatedFixedTransforms` only contains information between a link and its children (see [computeFixedTransforms](https://github.com/ros-planning/moveit/blob/6bf3d04488a42e4e8e7153c2349c2d35103e7e50/moveit_core/robot_model/src/robot_model.cpp#L1392), which only iterates over children), not information between a link and any of its siblings.

If you are talking about the case where a link is a fixed child of the joint group, the current change should catch that, without needing `getAssociatedFixedTransforms`.

v4hn:
> Say for 3 links, A, B, and C, where A is a parent to both B and C with fixed joints between A-B and A-C, and B is the start of some joint group. If the Jacobian of a point in C is requested, we should recognize that C is a static sibling of B and compute the Jacobian from there?

Yes.

> Given my understanding is correct, I'm not sure what to do, as getAssociatedFixedTransforms only contains information between a link and its children

Thank you for pointing this out! I rewrote this code two months ago, but apparently I forgot to submit a pull-request for it. I just created #912.

Once this is merged it would be great to add the sibling capability to the code here!
I don't see a reason though to hold this request until then, so I will merge this now.

Thank you for the contribution already!

v4hn:
cherry-picked to melodic

rhaschke:
@BryceStevenWilley @v4hn Is there anything else to do here, since #912 is merged now?
See https://github.com/ros-planning/moveit/pull/877#issuecomment-391321954

rhaschke:
Great, thanks!

BryceStevenWilley:
So I was working on finishing up using `getAssociatedTransforms` for static siblings, and realized a few things:

* a static sibling link that is updated by a group must be a descendant of an earlier link in the group
* A link can be a static sibling of a group and not a descendant only in the case described in the discussion above (a to b and to c, both fixed joints, b's parent joint is the first in a group), and in this cane, the Jacobian of that link is trivially the 0 matrix, as moving any joint in the group will not move c. In this case, `JointModelGroup::isLinkUpdated()` returns false for `c`, which is correct: c will not move. 

@v4hn, am I misunderstanding something, or is the only case `getJacobian` missing right now the latter case? If that is the only case, is there any reason someone would want a trivial 0 jacobian for a sibling link like that?

PR #878, opened 2018-05-02
do not always print visible messages when attaching

v4hn:
This should not be a public message.
Especially annoying because the RViz display calls this
function quite often when displaying trajectories with attached objects.


rhaschke:
Yes, I found this message annoying before too!

PR #895, opened 2018-05-15
Replace createUnmanagedInstance with createUniqueInstance

### LINKED ISSUE #866: Melodic: Unmanaged instances to `std::shared_ptr`?
A new method has been added to the api, `std::shared_ptr createSharedInstance`

https://github.com/ros/class_loader/pull/95

It may be a more robust way to handle plugins than the current practice of `createUnmanagedInstance` with raw pointers.

bmagyar:
### Description

Small refactor eliminating use of raw pointers from pluginlib. 
Fixes #866
Although the API mentioned there hasn't been propagated to pluginlib yet, hence the use of `createUniqueInstance()` instead of `createSharedInstance()` but regardless, `std::unique_ptr` converts to `std::shared_ptr`.

### Checklist

### DIFF
--- a/moveit_ros/move_group/src/move_group.cpp
+++ b/moveit_ros/move_group/src/move_group.cpp
@@ -134,10 +134,10 @@ private:
       try
       {
         printf(MOVEIT_CONSOLE_COLOR_CYAN "Loading '%s'...\n" MOVEIT_CONSOLE_COLOR_RESET, plugin->c_str());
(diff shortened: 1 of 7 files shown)


v4hn:
If we really want shared here instead of unique, I would propose to wait for the pluginlib interface instead of merging this.
This is not an important change after all and there's no need to add misleading code if behavior does not change.

rhaschke:
This is superseded by https://github.com/ros-planning/moveit/pull/1104, which also fixes dangling pointers.
Sorry for not merging this for so long...

PR #896, opened 2018-05-15
Remove boost shared ptrs

bmagyar:
### Description

I've removed all actively used occurences of `boost::shared_ptr`.

~~It seems that the issue that kept us using it was the rviz API for fetching the tf listener. I have solved it by adding a conversion function (in separate commit) although it is questionable if we should stick to this rviz API at all. The [updated API](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/frame_manager.h#L243) works with `tf2`. I think we should use that but it implies further refactoring of the `PlanningSceneMonitor` and everything related (which is pretty much everything else in MoveIt).~~

### Checklist

### DIFF
--- a/moveit_experimental/kinematics_cache/v1/kinematics_cache/include/kinematics_cache/kinematics_cache.h
+++ b/moveit_experimental/kinematics_cache/v1/kinematics_cache/include/kinematics_cache/kinematics_cache.h
@@ -42,7 +42,7 @@
 #include <planning_models/robot_model.h>
 #include <planning_models/kinematic_state.h>
 
-#include <boost/shared_ptr.hpp>
+#include <memory>
 
 namespace kinematics_cache
 {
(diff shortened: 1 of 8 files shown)


bmagyar:
I won't be able to adjust this PR in the coming week, feel free to touch it up if needed.

IanTheEngineer:
+1 for removing all instances of `boost::shared_ptr`. However, I see substantial overlap between the this PR and the tf2 migration PR that’s still under review: https://github.com/ros-planning/moveit/pull/830
That PR was the impidus for modifying the TransformListener API :)

bmagyar:
Awesome! Thanks for the heads up, I'll diff against that PR and see if there's anything left ;)

bmagyar:
Rebased on top of #830 !

v4hn:
looks good after #830 has been merged.

rhaschke:
Added to #906. Closing here.

PR #909, opened 2018-05-22
make verbose PS INFO messages DEBUG

v4hn:
Just like https://github.com/ros-planning/moveit/pull/878 ,
these messages should also not be printed every time by default.

At times they can also spam the terminal when a node uses
the PlanningScene internally.


rhaschke:
Cherry-picked to Melodic.

PR #912, opened 2018-05-23
include fixed tfs to siblings in links

v4hn:
If the robot structure contains y-paths made up of fixed transforms, the
previous implementation missed these associations in the associated_transforms.

The structure

- a - b - c
 \
  - d

only added associationed fixed transforms for a and b to c, but missed d.

I missed to submit this request when I wrote it two months ago
and it just popped up again [here](https://github.com/ros-planning/moveit/pull/877#issuecomment-390494029).

### DIFF
--- a/moveit_core/robot_model/src/robot_model.cpp
+++ b/moveit_core/robot_model/src/robot_model.cpp
@@ -294,26 +294,21 @@ void RobotModel::buildJointInfo()
   }
 
   bool link_considered[link_model_vector_.size()] = { false };
-  for (std::size_t i = 0; i < link_model_vector_.size(); ++i)
+  for (const LinkModel* link : link_model_vector_)
   {
-    if (link_considered[i])
+    if (link_considered[link->getLinkIndex()])
       continue;
 
     LinkTransformMap associated_transforms;
-    computeFixedTransforms(link_model_vector_[i], link_model_vector_[i]->getJointOriginTransform().inverse(),
-                           associated_transforms);
+    computeFixedTransforms(link, link->getJointOriginTransform().inverse(), associated_transforms);
     for (auto& tf_base : associated_transforms)
     {
       link_considered[tf_base.first->getLinkIndex()] = true;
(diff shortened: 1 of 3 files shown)


BryceStevenWilley:
Sorry this took so long for me to look at, but LGTM. I added a test that I though I'd be able to add to this PR, but I can't; would you mind cherry-picking commit [76366f126](https://github.com/BryceStevenWilley/moveit/commit/76366f126f5c0e9cb3f77e552995e4f5733c949e) off [my copy of this PR](https://github.com/brycestevenwilley/moveit/tree/pr-kinetic-fix-associated-fixed-transforms)?

Minor comment, but the `Below` in `computeFixedTransformsBelow` is a little vague (depends on whether you think of trees of chains growing up or down). Maybe `computeDescendantFixedTransforms` works better? Up to you though.

rhaschke:
@BryceStevenWilley Thanks for providing an initial unit test for this. I cleaned this up and augmented some more tests.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -293,19 +293,28 @@ void RobotModel::buildJointInfo()
...
-        if (it->first != link_model_vector_[i])
+      link_considered[tf_base.first->getLinkIndex()] = true;
+      for (auto& tf_target : associated_transforms)
+      {
+        if (&tf_base != &tf_target)
         {
-          getLinkModel(it->first->getName())->addAssociatedFixedTransform(link_model_vector_[i], it->second.inverse());
-          link_model_vector_[i]->addAssociatedFixedTransform(it->first, it->second);
+          // this const-cast is equivalent to
+          // associated_transforms[tf_base.first->getLinkIndex()]
```
rhaschke: This doesn't make sense at all, does it?
`associated_transforms` is probably way smaller than `tf_base.first->getLinkIndex()`.
You just want to regain write access to the "base" link, don't you?
I will add a commit to fix this comment.

Issue #913, opened 2018-05-24
Melodic binary release issues


rhaschke:
@clalancette Having released MoveIt for Melodic, we have some strange cmake errors in the binary release builds on the ROS build farm:

http://build.ros.org/job/Mbin_uB64__moveit_setup_assistant__ubuntu_bionic_amd64__binary/3/display/redirect

Essentially cmake complains, that `-lpthreads` cannot be found. 
However, I don't understand, why it doesn't look for `-lpthread` as well... 

Do you have seen this error already in a different context?

clalancette:
@rhaschke Yeah, we've been seeing that elsewhere.  That particular error message is a red herring; the real error is always further up.  If you look at http://build.ros.org/job/Mbin_uB64__moveit_setup_assistant__ubuntu_bionic_amd64__binary/3/console around time 23:15:34, you'll see that the real error is:

```
23:15:34 CMake Error at CMakeLists.txt:36 (find_package):
23:15:34   By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
23:15:34   asked CMake to find a package configuration file provided by "Qt5", but
23:15:34   CMake did not find one.
23:15:34 
23:15:34   Could not find a package configuration file provided by "Qt5" (requested
23:15:34   version 5.9.5) with any of the following names:
23:15:34 
23:15:34     Qt5Config.cmake
23:15:34     qt5-config.cmake
23:15:34 
23:15:34   Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
23:15:34   to a directory containing one of the above files.  If "Qt5" provides a
23:15:34   separate development package or SDK, be sure it has been installed.
```

rhaschke:
Thanks a lot for this hint.

PR #915, opened 2018-05-24
add missing package dependencies

rhaschke:
to fix ROS build farm issues (#913)

### DIFF
--- a/moveit_commander/package.xml
+++ b/moveit_commander/package.xml
@@ -20,6 +20,9 @@
 
   <build_depend>python</build_depend>
 
+  <test_depend>rostest</test_depend>
+  <test_depend>moveit_resources</test_depend>
+
   <run_depend>geometry_msgs</run_depend>
   <run_depend>moveit_msgs</run_depend>
   <run_depend>moveit_ros_planning_interface</run_depend>
--- a/moveit_setup_assistant/package.xml
+++ b/moveit_setup_assistant/package.xml
@@ -18,6 +18,9 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
+  <build_depend>libogre-dev</build_depend>
+  <build_depend>qtbase5-dev</build_depend>
+  <build_depend>libqt5-opengl-dev</build_depend>
   <build_depend>moveit_core</build_depend>
   <build_depend>moveit_ros_planning</build_depend>
   <build_depend>moveit_ros_visualization</build_depend>


v4hn:
out of curiosity: are these dependency back-portable to kinetic?

rhaschke:
> out of curiosity: are these dependency back-portable to kinetic?

I guess you are referring to the Qt packages. If so, yes the package names are the same in Xenial.

rhaschke:
@130s I'm going to re-release for Melodic including these fixes. Any objections?

rhaschke:
@130s I cleaned up the changelogs. Do you mind to re-release for Melodic as version 0.10.0. As the previous release was not complete (`setup_assistant` and `moveit_commander` failed), I suggest to reuse the same version number.

130s:
Pre-release test (I only ran on `Bionic`) went fine.
```
$ mkdir pr_moveit_melodic
$ cd pr_moveit_melodic/
$ generate_prerelease_script.py   https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml melodic default ubuntu bionic amd64  moveit   --level 0  --output-dir ./
$ ./prerelease.sh 
:
moveit_planners_ompl/gtest-test_state_space.xml: 3 tests
moveit_ros_planning_interface/rostest-test_cleanup.xml: 1 tests
moveit_ros_planning_interface/rostest-test_python_move_group.xml: 1 tests
moveit_ros_planning_interface/rostest-test_python_move_group_ns.xml: 2 tests
moveit_ros_planning_interface/rostest-test_robot_state_update.xml: 1 tests
moveit_ros_planning_interface/rosunit-cleanup_tests.xml: 2 tests
moveit_ros_planning_interface/rosunit-python_move_group.xml: 2 tests
moveit_ros_planning_interface/rosunit-python_move_group_ns.xml: 2 tests
moveit_ros_planning_interface/rosunit-robot_state_update.xml: 1 tests
moveit_ros_robot_interaction/gtest-locked_robot_state_test.xml: 16 tests
moveit_ros_visualization/rostest-test_moveit_joy.xml: 1 tests
moveit_ros_visualization/rosunit-moveit_joy.xml: 1 tests
Summary: 119 tests, 0 errors, 0 failures, 0 skipped
```
I might not get to it before weekend, so please go ahead. Changelog is the only thing that takes me forever, which you already took care of, so hope it's quick. 

Just one thing,

> Do you mind to re-release for Melodic as version 0.10.0. As the previous release was not complete (setup_assistant and moveit_commander failed), I suggest to reuse the same version number.

No, with any change, trivial or not, new version should be made. Otherwise we'll lose a way to keep track of "bad" and "good" versions. Micro/patch version number is for fix so in this case it should be totally fine to make a new patch release.

rhaschke:
Why do you want to keep track of a "bad" version at all? Nobody wants to use a bad version...
Just my thoughts.

gavanderhoorn:
@rhaschke: Bloom / the buildfarm will not let you / accept a rerelease of new code with an identical version nr of a previous release. All sort of weird things will happen.

I believe the accepted workflow here would be to add `-1` to the version to indicate that the previous release only had some incorrect packaging / release artefacts (which I believe is the case here, right?).

Can't find a ref for this now on the wiki though :/

mikaelarguedas:
If there is any change in the source repository, a new version number should be used as the debian release process (and the ROS one by extension) assumes that the sources stay unchanged within a debian increment (deb-inc).

So we recommend releasing a new version, (in this case 0.10.1).

HTH,

gavanderhoorn:
Afaik there are no changes in the sources, just packaging (missing dependencies). That would warrant a deb-inc iiuc.

If changes to the changelogs count as 'changes to the sources', then a patch increment would be necessary, I agree.

mikaelarguedas:
Sorry this was a bit unclear: I'll try to give a bit more details.

Changes to any file in the source repository counts as a source change.

We use the [tag on the source repository](https://github.com/ros-planning/moveit/tree/0.10.0) to import the source code into the release repository, and tags in git are assumed immutable. This is then used to create the orig.tar.gz. Per debian rules, the orig.tar.gz should not change for a given version.
The case where you can make a `packaging change` in a deb-inc is when you apply packaging patches to the release repository directly (like modifying the debian rules, cmake flags etc).

In this case the change lies in the source repository, so a new version should be released (even if it was only a change to the package.xml and not modifying the changelogs).

rhaschke:
@mikaelarguedas
The ROS release process in principle allows to re-release a package with the same version. I did that just recently with another small package from my group. Of course you need to force-push to the source repo (the tag) as well as to the release-repo to do so.
The Debian package management system can deal with this as well, as the ROS system adds a complete timestamp to the version number, making it unique.

mikaelarguedas:
I never said it **cannot** work, what I meant is that this is **not a supported use case** and that we **strongly discourage** it. Our guideline being to make a new release for any change happening in the source repository. And a more generic guideline being that force pushing anything to production branches or repositories should be avoided. 

The scenarios where it still works are getting fewer and fewer as we try to move closer to complying to the Debian rules, so I do not recommend assuming that this is a use case that will keep working as chances are high we are going to break it in the future.

Maybe the part I'm missing here is why making a new release is a blocker here. My understanding is that 0.10.0 is broken for some packages, so you release a patched version that fixes it, it makes sense to me that the patched version is named 0.10.1, doesn't it ?

rhaschke:
Personally, I don't like to expose a broken release. I would prefer to overwrite the old one. However, I will comply to the guidelines ;-)

130s:
Since 0.10.0 is already out, there can be users who have it installed. If we overwrote that with a fixed version, there would be 2 types of binaries with exactly the same version number and we would have no way to tell them apart when users filed an issue report. So it's not that we want to keep broken binaries, but we want to keep the way to differentiate the binaries IMO.

I can arrange bloom this weekend if that's not too late but if it is, then please go ahead. Thanks Robert.

rhaschke:
@130s I just bloomed MoveIt! 0.10.1 for Melodic.

PR #916, opened 2018-05-24
OMPL logging that uses rosconsole

### LINKED ISSUE #719: OMPL Info and Debug logs are shown since latest update, even though verbose/debug mode is not specified
### Description
OMPL Info and Debug logs are shown since latest update, even though verbose/debug mode is not specified.

### Your environment
* ROS Distro: Kinetic
* OS Version: Ubuntu 16.04
* Binary build:  0.9.10

### Steps to reproduce
From package created with Setup Assistant, launch move_group.launch with arguments `debug` and `info` set to false.

### Expected behaviour
Not to see OMPL Debug or Info logs in the terminal.

### Actual behaviour
image

https://gist.github.com/hansondon/229df6962bb55389cc72fcb2b2fa7832

zkingston:
### Description

Fixes #719. Adds a custom `ompl::msg::OutputHandler` that uses `rosconsole` to log OMPL messages at the appropriate logging level.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/ompl_planner_manager.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/ompl_planner_manager.cpp
@@ -60,51 +60,53 @@ class OMPLPlannerManager : public planning_interface::PlannerManager
 public:
   OMPLPlannerManager() : planning_interface::PlannerManager(), nh_("~"), display_random_valid_states_(false)
   {
-      class OutputHandler : public ompl::msg::OutputHandler
+    class OutputHandler : public ompl::msg::OutputHandler
+    {
+      const char* format = "[OMPL] %s";
+
+    public:
+      OutputHandler() : ompl::msg::OutputHandler()
       {
-      public:
-          OutputHandler() : ompl::msg::OutputHandler()
-          {
-          }
+      }
 
-          ~OutputHandler() override
-          {
-          }
+      ~OutputHandler() override
+      {
+      }
 
-          void log(const std::string &text, ompl::msg::LogLevel level, const char *filename, int line) override
-          {
-              std::unique_lock<std::mutex> lock(lock_);
-              std::stringstream ss;
-              ss << filename << ":" << line;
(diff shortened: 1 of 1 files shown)


moveit_planners/ompl/ompl_interface/src/ompl_planner_manager.cpp
```diff
@@ -58,6 +60,51 @@ class OMPLPlannerManager : public planning_interface::PlannerManager
...
+          {
+          }
+
+          ~OutputHandler() override
+          {
+          }
+
+          void log(const std::string &text, ompl::msg::LogLevel level, const char *filename, int line) override
+          {
+              std::unique_lock<std::mutex> lock(lock_);
```
rhaschke: Do you really need locking? ROS_CONSOLE / the underlying logging uses locking as well.

moveit_planners/ompl/ompl_interface/src/ompl_planner_manager.cpp
```diff
@@ -47,17 +47,58 @@
...
+
 #include <memory>
 
 namespace ompl_interface
 {
 using namespace moveit_planners_ompl;
 
+#define OMPL_ROS_LOG                                                                                                   \
+  {                                                                                                                    \
+    ROSCONSOLE_DEFINE_LOCATION(true, ::ros::console::levels::Debug, ROSCONSOLE_NAME_PREFIX ".ompl");                   \
```
zkingston: It looks like this should be:
```c++
#define OMPL_ROS_LOG(ros_level) \
 { \
   ROSCONSOLE_DEFINE_LOCATION(true, ros_level, ROSCONSOLE_NAME_PREFIX ".ompl"); \
...
```
Is that correct?

rhaschke: Well spotted. I pushed the fix. If you agree, I will do the final merge.

zkingston:
Looks good to me, thanks for the help.

PR #922, opened 2018-05-28
CSM: copy dynamics to output if enabled

v4hn:
The monitor optionally kept track of velocities, accelerations & effort too,
but never wrote them to the output variable when asked for the current state...

With this change the monitored planning scene of the move_group node contains current dynamics information too, if enabled.


rhaschke:
Hm. Looks like RobotState doesn't provide a function yet, to copy accelerations. 
Hence, I approve as is and will merge next. The Travis failure is due to an unrelated issue: https://github.com/ros-gbp/ompl-release/issues/14

PR #931, opened 2018-06-01
motion_planning_rviz_plugin : add CartesianPath planning check box

k-okada:
### Description

Added "cartesian path" checkbox, so that everyone easily run the result of `computeCartesianPath`example

screenshot from 2018-06-01 12-24-20

screenshot from 2018-06-01 12-24-27

I have asked many times from industrial users about cartesian space interpolation, which they call  `PTP (point-to-point)  Control`, I know they already knows https://rosindustrial.org/news/2014/9/5/cartesian-path-planner-plug-in-for-moveit, but they think trajectory obtained from joint space search space  more advanced feature and always curious that why MoveIt do not have basic featrue (for them) like `PTP`.

### Checklist
- [ ] Optional: Created tests, which fail without this PR [reference]
(http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/tests.html)

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -150,8 +150,8 @@ private Q_SLOTS:
   void approximateIKChanged(int state);
 
   // Planning tab
-  void computeCartesianPlanButtonClicked();
-  void computeJointSpacePlanButtonClicked();
+  void computeCartesianPlan();
+  void computeJointSpacePlan();
   void planButtonClicked();
   void executeButtonClicked();
   void planAndExecuteButtonClicked();
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -150,12 +150,15 @@ private Q_SLOTS:
   void approximateIKChanged(int state);
 
   // Planning tab
+  void computeCartecianPlanButtonClicked();
+  void computeJointSpacePlanButtonClicked();
   void planButtonClicked();
   void executeButtonClicked();
   void planAndExecuteButtonClicked();
   void stopButtonClicked();
   void allowReplanningToggled(bool checked);
   void allowLookingToggled(bool checked);
+  void cartesianPathToggled(bool checked);
   void allowExternalProgramCommunication(bool enable);
   void pathConstraintsIndexChanged(int index);
   void useStartStateButtonClicked();
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -150,7 +150,7 @@ private Q_SLOTS:
   void approximateIKChanged(int state);
 
   // Planning tab
-  void computeCartecianPlanButtonClicked();
+  void computeCartesianPlanButtonClicked();
   void computeJointSpacePlanButtonClicked();
   void planButtonClicked();
   void executeButtonClicked();
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -43,6 +43,8 @@
 
 #include <std_srvs/Empty.h>
 #include <moveit_msgs/RobotState.h>
+#include <eigen_conversions/eigen_msg.h>
+#include <moveit/trajectory_processing/iterative_time_parameterization.h>
 
 #include "ui_motion_planning_rviz_plugin_frame.h"
 
@@ -87,6 +89,10 @@ void MotionPlanningFrame::allowLookingToggled(bool checked)
     move_group_->allowLooking(checked);
 }
 
+void MotionPlanningFrame::cartesianPathToggled(bool checked)
+{
+}
+
 void MotionPlanningFrame::pathConstraintsIndexChanged(int index)
 {
   if (move_group_)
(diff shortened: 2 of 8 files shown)


moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
```diff
@@ -150,12 +150,15 @@ private Q_SLOTS:
   void approximateIKChanged(int state);
 
   // Planning tab
+  void computeCartecianPlanButtonClicked();
```
felixvd: This misspelling of "Cartesian" occurs a few more times in the code and PR title, and a few other ones in the code. I wouldn't bother mentioning this normally, but since it is a function name and inconsistent with the function it calls, I think this is a bit more important than a regular typo.

Regardless, I recommend VS Code's Code Spell Checker extension, which I also use :)

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -135,6 +211,14 @@ void MotionPlanningFrame::computePlanButtonClicked()
   Q_EMIT planningFinished();
 }
 
+void MotionPlanningFrame::computePlanButtonClicked()
+{
+  if (ui_->cartesian_path->checkState())
```
felixvd: I think it would make sense to call the checkbox "use_cartesian_path", to improve readibility in places like this.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -108,7 +114,77 @@ void MotionPlanningFrame::onClearOctomapClicked()
...
+
+    // Second get a RobotTrajectory from trajectory
+    rt.setRobotTrajectoryMsg(*move_group_->getCurrentState(), scaled_trajectory);
+
+    // Thrid create a IterativeParabolicTimeParameterization object
+    trajectory_processing::IterativeParabolicTimeParameterization iptp;
+
+    // Fourth compute computeTimeStamps
+    bool success =
+        iptp.computeTimeStamps(rt, ui_->velocity_scaling_factor->value(), ui_->acceleration_scaling_factor->value());
```
felixvd: So this is the current workaround for velocity scaling cartesian paths? I learned something already.

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
```diff
@@ -150,12 +150,15 @@ private Q_SLOTS:
...
+  void computeJointSpacePlanButtonClicked();
```
felixvd: These two new functions are called ButtonClicked(), implying that they are connected to a button, but they are not. I would remove the "ButtonClicked" from the function name.

On second thought, is that file automatically generated by Qt? Should it be modified like this at all? See the comments [at the bottom of the file](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h#L321).

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
k-okada: Thanks, fixed typos.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
k-okada: renamed

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
k-okada: no in is hand written. so removed  from function name.

felixvd:
Lovely, thank you!

I tried this out locally, and I love it, but I have some issues. 

- The result of the Plan button is "x% achieved" and does not mention the time. For consistency with the other planning mode, I would display the time in a new line.
- The Execute button is not enabled after planning a cartesian path. If you first click Plan without the `Use Cartesian Path` checkbox enabled, the button is available.
- If you Plan, then Execute a non-cartesian path from pose0 to pose1, then Plan a Cartesian path from pose1 to pose2, the displayed trajectory starts from pose0. If you then plan a normal motion from pose1 to pose2, and then a cartesian path again, the displayed trajectory is correct. The `Plan and Execute` button seems to use the correct startState (I see no sudden jumps, although I didn't try on a physical robot), but something is off with the `Plan` button.

Overall I am a big fan, this button is absolutely a must-have. Let's get it working fully. Cheers

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
felixvd: I don't know any, and I actually failed to implement it in my own code, but I can confirm that it works as you wrote it here :)

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -87,6 +89,10 @@ void MotionPlanningFrame::allowLookingToggled(bool checked)
     move_group_->allowLooking(checked);
 }
 
+void MotionPlanningFrame::cartesianPathToggled(bool checked)
```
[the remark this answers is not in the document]
rhaschke: > When using Gazebo, [this bug](https://answers.ros.org/question/253004/moveit-problem-error-trajectory-message-contains-waypoints-that-are-not-strictly-increasing-in-time/) occurs. The simplest solution is to delete the first waypoint (the current robot pose) from the list.

This is also the correct solution. In planCartesianPath, there will be a warning as well: #1085.

rhaschke:
> When using Gazebo, [this bug](https://answers.ros.org/question/253004/moveit-problem-error-trajectory-message-contains-waypoints-that-are-not-strictly-increasing-in-time/) occurs. The simplest solution is to delete the first waypoint (the current robot pose) from the list.
> 
> While I think it should be fixed in planCartesianPath rather than everywhere else, it doesn't hurt to implement the workaround here either.

This is tackled in #1085. See discussion there.

PR #939, opened 2018-06-05
fix PS getDiffMsg

v4hn:
The robot state should be marked as diff, otherwise
PlanningScene::isEmpty(RobotState&) will not acknowledge
it as empty.

@rhaschke

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -710,7 +710,10 @@ void PlanningScene::getPlanningSceneDiffMsg(moveit_msgs::PlanningScene& scene_ms
   if (kstate_)
     robot_state::robotStateToRobotStateMsg(*kstate_, scene_msg.robot_state);
   else
+  {
     scene_msg.robot_state = moveit_msgs::RobotState();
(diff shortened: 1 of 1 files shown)


moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -710,7 +710,10 @@ void PlanningScene::getPlanningSceneDiffMsg(moveit_msgs::PlanningScene& scene_ms
   if (kstate_)
     robot_state::robotStateToRobotStateMsg(*kstate_, scene_msg.robot_state);
   else
+  {
     scene_msg.robot_state = moveit_msgs::RobotState();
+    scene_msg.robot_state.is_diff = true;
```
rhaschke: Why do you want to set `is_diff` only in `else` branch? I think, it should be set always.

v4hn: No. To the best of my knowledge the whole framework expects that in case `kstate_` is defined, it defines a full robot state including all attached objects. kstates of underlying planning scenes are ignored.
This is a rather longstanding shortcoming in MoveIt.

I agree though that it is confusing in this case, because we ask for a diff...

PR #936, opened 2018-06-03
Added gazebo launch file

MohmadAyman:
### Description

Added a gazebo launch file template for the setup assistant.
For the ur10 robot it looks like [that](https://gist.github.com/MohmadAyman/5c4027155abd2bea929b9fa07738600f). 

### Checklist
- [ ] **Required**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] Extended the tutorials / documentation, if necessary [reference](http://moveit.ros.org/documentation/contributing/)
- [ ] Include a screenshot if changing a GUI
- [ ] Optional: Created tests, which fail without this PR [reference](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/tests.html)
- [ ] Optional: Decide if this should be cherry-picked to other current ROS branches (Indigo, Jade, Kinetic)

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #956, opened 2018-06-20
Simulation Screen for the SA

MohmadAyman:
### Description

This PR is a part of improving the integration between gazebo, moveit and ros [issue 894](https://github.com/ros-planning/moveit/issues/894).

For a robot to be rendered in gazebo [it has to has an `inertial` element for each link](http://gazebosim.org/tutorials/?tut=ros_urdf), this PR tells the user if the urdf didn't contain them.
and views the new `urdf` on the simulation window for the user.

Added `elements`  which the user should add to his `urdf` are colored in green.

Once this PR is merged, the `[gazebo.launch]`(https://github.com/ros-planning/moveit/pull/936) will point to this new urdf.

The documentation of the setup assistant are updated in this [PR](https://github.com/ros-planning/moveit_tutorials/pull/184).

### Checklist

### DIFF
--- a/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
+++ b/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
@@ -231,6 +231,9 @@ public:
   /// URDF robot model
   urdf::ModelSharedPtr urdf_model_;
 
+  /// URDF robot model string
+  std::string urdf_string_;
+
   // ******************************************************************************************
   // SRDF Data
   // ******************************************************************************************
@@ -323,6 +326,13 @@ public:
   void outputFollowJointTrajectoryYAML(YAML::Emitter& emitter);
   bool outputROSControllersYAML(const std::string& file_path);
   bool output3DSensorPluginYAML(const std::string& file_path);
+
+  /**
+   * \brief Parses the existing urdf and constructs a string from it with the elements required by gazebo simulator
+   * added
+   * \return gazebo compatible urdf or empty if error encountered
+   */
+  std::string getGazeboCompatibleURDF();
 
   /**
    * \brief Set list of collision link pairs in SRDF; sorted; with optional filter
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -372,6 +372,63 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
   return true;  // file created successfully
 }
 
+// ******************************************************************************************
+// Writes a Gazebo compatible robot URDF to gazebo_compatible_urdf_string_
+// ******************************************************************************************
+std::string MoveItConfigData::getGazeboCompatibleURDF()
+{
+  bool new_urdf_needed = false;
+  TiXmlDocument urdf_document;
+
+  // Used to convert XmlDocument to std string
+  TiXmlPrinter printer;
+  urdf_document.Parse((const char*)urdf_string_.c_str(), 0, TIXML_ENCODING_UTF8);
+  try
+  {
+    for (TiXmlElement* link_element = urdf_document.RootElement()->FirstChildElement(); link_element != NULL;
+         link_element = link_element->NextSiblingElement())
+    {
+      if (((std::string)link_element->Value()).find("link") == std::string::npos)
+        continue;
+      // Before adding inertial elements, make sure there is none and the link has collision element
+      if (link_element->FirstChildElement("inertial") == NULL && link_element->FirstChildElement("collision") != NULL)
+      {
+        new_urdf_needed = true;
+        TiXmlElement inertia_link("inertial");
+        TiXmlElement mass("mass");
+        TiXmlElement inertia_joint("inertia");
+
+        mass.SetAttribute("value", "0.1");
+
+        inertia_joint.SetAttribute("ixx", "0.03");
+        inertia_joint.SetAttribute("iyy", "0.03");
+        inertia_joint.SetAttribute("izz", "0.03");
+        inertia_joint.SetAttribute("ixy", "0.0");
+        inertia_joint.SetAttribute("ixz", "0.0");
+        inertia_joint.SetAttribute("iyz", "0.0");
+
+        inertia_link.InsertEndChild(mass);
+        inertia_link.InsertEndChild(inertia_joint);
+
+        link_element->InsertEndChild(inertia_link);
+      }
+    }
+  }
+  catch (YAML::ParserException& e)  // Catch errors
+  {
+    ROS_ERROR_STREAM_NAMED("moveit_config_data", e.what());
+    return std::string("");
+  }
+
+  if (new_urdf_needed)
+  {
+    urdf_document.Accept(&printer);
+    return std::string(printer.CStr());
+  }
+
+  return std::string("");
+}
+
 bool MoveItConfigData::outputFakeControllersYAML(const std::string& file_path)
 {
   YAML::Emitter emitter;
--- a/moveit_setup_assistant/src/widgets/simulation_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/simulation_widget.cpp
@@ -1,0 +1,202 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2018, Mohamad Ayman.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * The name of Mohamad Ayman may not be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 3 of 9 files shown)


MohmadAyman:
Well once the [ new controllers window](https://github.com/ros-planning/moveit/pull/951) is added, the `gazebo_ros_control` plugin would work without having to do anything manually, that is the plan.

For the inertia values, we tell the user that those default inertial values are defaults and they can edit them if they wanted to. and later we should be able to calculate the inertia of simple shapes like `ignition-math` [does here](https://bitbucket.org/ignitionrobotics/ign-math/src/6863e427572803a15db7ffb48173ad2b97375c0a/include/ignition/math/MassMatrix3.hh#lines-906:1095) .

gavanderhoorn:
Personally I think I'd like a checkbox somewhere to make this configurable. That would at least leave it up to the user to decide whether he wants the MSA changing his URDF or not.

And something else I realised: what if my robot is actually modelled using `xacro`? The MSA converts that to URDF, but that URDF will not be used by the launch files. Saving a new URDF with the added `inertial` elements does not make sense in those cases.

Tbh, I would think that providing a proper URDF to the MSA is the responsibility of the user. The MSA setting up the required *configuration files* for `gazebo_ros_control` to work is something different. We don't add `gazebo_ros_control` elements to the URDF either, do we?

MohmadAyman:
A check box sounds good :+1: 

The newly created `urdf` file will be added to a new directory in the config package, and will be referenced only by the `gazebo.launch` that we are adding in PR#936, so there is no harm done to the user, if he wants to simulate in gazebo, they will have a `urdf` ready for that.

I'm planning to add the `gazebo_ros_control` elements to the `urdf` once the `ros controllers` PR is merged.

MohmadAyman:
@gavanderhoorn 
>forgive me if I misunderstand, but that would mean that the MSA must be rerun for every change made to the xacro backing the urdf then, correct? That seems like quite a big change and rather expensive.

Correct, If the xacro were to be changed often, then yes this would be very expensive, but that doesn't happen often, right?
>As I commented earlier in #894, I'm not aware of what the GSoC assignment / description of work contained specifically, but the direction you're taking this in seems to increase coupling between MoveIt-MSA and Gazebo quite a bit. As Gazebo is just one of the simulators that can be used with MoveIt, is that desirable?

The main objective of this project, as was discussed with @davetcoleman before GSoC, as mentioned [here on the GSoC website](https://summerofcode.withgoogle.com/projects/#5374298588446720), was to improve the integration between gazebo and moveit!.

I do agree with you we shouldn't increase the coupling/dependency between gazebo and moveit!. What we are doing should be something **extra** for the `setup assistant`, the idea is to make simulating a robot in gazebo while using moveit! a more **automated** process through the setup assistant. As **gazebo** maybe the most popular **simulator** used with **ros**, this should benefit a lot of users.

I can't forget to mention your [comment on the main issue](https://github.com/ros-planning/moveit/issues/894#issuecomment-389216214) to make the **SA plugin based**, which sounds very cool and I guess everyone would be thrilled by that. In the future this **extra** **optional** step for **simulation** could be just a **plugin**.

gavanderhoorn:
@MohmadAyman wrote:
> Correct, If the xacro were to be changed often, then yes this would be very expensive, but that doesn't happen often, right?

My workflow while developing can include quite some round-trip editing of xacros, which with this PR would now have to include the MSA as well. But I may not be representative of typical usage.

davetcoleman:
@gavanderhoorn I think you're bringing up some great points that I've been struggling with also. I totally agree we should have a checkbox for generating the suggested Gazebo tags and we should educate the user very carefully about what is happening behind scenes.

I'm uncomfortable with having Gazebo use a different URDF also. Rather, I'd like the MSA to provide to the user suggested XML to add to their URDF/XACO, but that step must be manual. We'd link out to a help page that explains the details of that.

MohmadAyman:
@gavanderhoorn 
Well I've the idea of making this a part of `xacros` package itself, by maybe adding an option called `-gazebo_compatible` or `-add_inertia`, this would then be as easy as running a command.

@davetcoleman 
It is not suggested to either create a new `urdf`, and we can't edit the existing one in most cases. So if we are going to apply the suggestion of promoting the user with the `xml` that they should add to their `urdf`, that does probably mean we would need a new screen, a `simulation screen`, do you suggest another way?

MohmadAyman:
>>that does probably mean we would need a new screen, a simulation screen

>Yes, I think that might be needed.

Then I will rename this PR to `simulation screen`, and instead of writing a new `urdf`, we just show the user what elements they need to add to their `urdf` and where.

I'll start drawing a mock-up to add it to the description.

MohmadAyman:
This is the mock-up


Do we continue the discussion about the Simulation screen here, or do we open an issue for that?
@davetcoleman

davetcoleman:
I'd rather have as few text boxs as necessary for the screen. Can we have just one large one that creates the new URDF, but then the user has to copy into the actual file location?

MohmadAyman:
description updated.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -375,6 +375,64 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
...
+
+          link_element->InsertEndChild(inertial);
+        }
+      }
+    }
+  }
+  catch (YAML::ParserException& e)  // Catch errors
+  {
+    ROS_ERROR_STREAM(e.what());
+    return "";
```
mcevoyandy: I think this should be `return std::string();`

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -375,6 +375,64 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
...
+  }
+
+  if (new_urdf_needed)
+  {
+    urdf_document.Accept(&printer);
+    return printer.CStr();
+  }
+
+  return "";
```
mcevoyandy: `return std::string();`?

moveit_setup_assistant/src/widgets/simulation_widget.cpp
```diff
@@ -0,0 +1,177 @@
...
+SimulationWidget::SimulationWidget(QWidget* parent, moveit_setup_assistant::MoveItConfigDataPtr config_data)
+  : SetupScreenWidget(parent), config_data_(config_data)
+{
+  // Basic widget container
+  QVBoxLayout* layout = new QVBoxLayout();
+  layout->setAlignment(Qt::AlignTop);
+
+  // Top Header Area ------------------------------------------------
+
+  HeaderWidget* header = new HeaderWidget("Simulation", "Shows what changes that should be made for the robot urdf to "
```
mcevoyandy: change `... changes that should ...` to `... changes should ...`

moveit_setup_assistant/src/widgets/simulation_widget.cpp
```diff
@@ -0,0 +1,177 @@
...
+                                                        "be spawned and rendered "
+                                                        "correctly in gazebo simulator. "
+                                                        " Green colored text are the changes.",
+                                          this);
```
mcevoyandy: Could be shorter "The URDF changes needed for Gazebo compatibility are shown in green text."

moveit_setup_assistant/src/tools/moveit_config_data.cpp
MohmadAyman: I agree it seems be better to cast them to `std::string`.

moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
```diff
@@ -280,6 +283,13 @@ class MoveItConfigData
   void outputFollowJointTrajectoryYAML(YAML::Emitter& emitter);
   bool outputROSControllersYAML(const std::string& file_path);
 
+  /**
+   * \brief Parses the existing urdf and constructs a string from it with the elements required by gazebo simulator
+   * added
+   * \ return true if parsed into string with no errors
```
davetcoleman: it does not return true, it returns an empty string...

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -375,6 +375,64 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
...
+
+  // Used to convert XmlDocument to std string
+  TiXmlPrinter printer;
+  urdf_document.Parse((const char*)urdf_string_.c_str(), 0, TIXML_ENCODING_UTF8);
+  try
+  {
+    for (TiXmlElement* link_element = urdf_document.RootElement()->FirstChildElement(); link_element != NULL;
+         link_element = link_element->NextSiblingElement())
+    {
+      if (((std::string)link_element->Value()).find("link") != std::string::npos)
```
davetcoleman: make this if statement be ``==`` and then continue, to prevent deeply nestled if statements

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -375,6 +375,64 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
...
+      {
+        // Before adding inertial elements, make sure there is none and the link has collision element
+        if (link_element->FirstChildElement("inertial") == NULL && link_element->FirstChildElement("collision") != NULL)
+        {
+          new_urdf_needed = true;
+          TiXmlElement inertial("inertial");
```
davetcoleman: inertial vs inertia is very confusing. instead say inertia_link vs inertia_joint

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -375,6 +375,64 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
...
+          mass.SetAttribute("value", "0.1");
+
+          inertia.SetAttribute("ixx", "0.03");
+          inertia.SetAttribute("iyy", "0.03");
+          inertia.SetAttribute("izz", "0.03");
+          inertia.SetAttribute("ixy", "0.0");
+          inertia.SetAttribute("ixz", "0.0");
+          inertia.SetAttribute("iyz", "0.0");
+
+          inertial.InsertEndChild(mass);
```
davetcoleman: for "mass" and "inertial" you can construct those directly in the insert function:

    inertial.InsertEndChild(TiXmlElement("mass"));

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -375,6 +375,64 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
...
+          inertial.InsertEndChild(inertia);
+
+          link_element->InsertEndChild(inertial);
+        }
+      }
+    }
+  }
+  catch (YAML::ParserException& e)  // Catch errors
+  {
+    ROS_ERROR_STREAM(e.what());
```
davetcoleman: add _NAMED

moveit_setup_assistant/src/tools/moveit_config_data.cpp
MohmadAyman: If `mass` as `TiXmlElement` were to be constructed directly in the insert function, I can't think of a simple way to add `attributes` to it afterwards.

MohmadAyman:
@davetcoleman 
Addressed the comments, thanks for the review.

davetcoleman:
@mcevoyandy please review

davetcoleman:
Checks are still failing

MohmadAyman:
@davetcoleman Travis happy.

davetcoleman:
cherry-picked to melodic

PR #957, opened 2018-06-23
collision_detection::World::moveObject()

rhaschke:
Provide a convienency method to move all shapes of a collision object. Before, for all shapes shapes we had to call `moveShapeInObject()`, which created overhead for finding the object over and over.


v4hn:
Thanks, makes sense to add.
Merged and picked to melodic.

May I ask in what context you use these interfaces?
Adding things to the collision world directly is quite low-level and having more than one shape in an object does not behave very nice in other places either, so I'm wondering.

rhaschke:
I used this function to fix initial collisions of objects spawned from our vision pipeline. Due to bad fitting and or bad camera-robot calibration, these objects might collide with the table.
In my case, I have direct access to a PlanningScene and thus the underlying World object.

PR #961, opened 2018-06-26
moveit_ros/perception: address gcc6 build error

bulwahn:
### Description

Since gcc6, compiling fails with `stdlib.h: No such file or directory`, as including '-isystem /usr/include' breaks since gcc6, cf., https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70129.

This commit addresses this issue for this package in the same way it was addressed in various other ROS packages. A list of related commits and pull requests is at https://github.com/ros/rosdistro/issues/12783.

Commit 4417f1e70a68 ("making some includes SYSTEM and re-adding link_directories") on 2012-12-06 added the SYSTEM without providing any further rationale for it.

### Checklist
- [x] **Required**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code); Remark: code was not changed.
- [x] Extended the tutorials / documentation, if necessary [reference](http://moveit.ros.org/documentation/contributing/): Not applicable
- [x] Include a screenshot if changing a GUI: Not applicable

I compile tested my change on Ubuntu 16.04; the change is required for proper use in the OpenEmbedded cross-compile tool chain in meta-ros.

### DIFF
--- a/moveit_ros/perception/CMakeLists.txt
+++ b/moveit_ros/perception/CMakeLists.txt
@@ -84,8 +84,6 @@ include_directories(lazy_free_space_updater/include
                     ${perception_GL_INCLUDE_DIRS}
                     ${Boost_INCLUDE_DIRS}
                     ${catkin_INCLUDE_DIRS}
-                    )
-include_directories(SYSTEM
                     ${OpenCV_INCLUDE_DIRS}
                     ${EIGEN3_INCLUDE_DIRS}
                     ${SYSTEM_GL_INCLUDE_DIR}


v4hn:
One of the main rationales to make includes `SYSTEM` is to avoid warnings from these headers.
With eigen3, this is a valid concern imho.

Nobody noticed problems with this in the last year, personally I work with gcc 7.3 at the moment and some contributors work with gcc 8.1.
How come you have problems with this?

Which package introduces the /usr/include includes?
If I remember correctly, these should not be specified by packages - at least pkg-config strips them.

bulwahn:
I am maintaining the OpenEmbedded/Yocto layer for ROS applications; it is a cross-compile tool chain that allows to compile for various architectures. The reason why the cross-compile tool chain is affected and not other build environment is not completely clear to me, possibly the overall build environment is slightly different; we use a vanilla gcc and pkg-config with some patches, whereas distributions add other additional patches and configurations.
The gcc issue in the issue tracker mentioned above states that warnings should not be made silent with the SYSTEM attribute and this is a widely adopted misuse of the SYSTEM attribute. I personally have no opinion on that, but I currently patch this ROS package with this patch in the pull request before configuring and compiling it.

v4hn:
I don't oppose merging this request,
I'm just wondering why you have problems with it
when nobody else saw that in ages.

I guess I'm the one ROS user who builds on a *very* vanilla linux-based system, running Lunar-Linux, and I don't see this problem.

CMake strips the implicit system directories for ages:
https://gitlab.kitware.com/cmake/cmake/blob/master/Source/cmLocalGenerator.cxx#L847

So maybe one of your toolchains does not have correct IMPLICIT flag set?

bulwahn:
I got side-tracked with other issues; it will take me some time to re-visit this and discuss with maintainers. I would suggest to keep the pull request open until I come back with a conclusion on that topic. There is no need to merge this quickly or do anything further from your side.

v4hn:
@bulwahn your suggestion to keep this open is half a year old by now.
Can we close this?

PR #968, opened 2018-07-02
reset moveit_msgs::RobotState.is_diff to false

simonschmeisser:
Currently once robot_state.is_diff has been set to true it will stay so. This breaks detaching attached objects. This is a fix for a regression introduced in #939

### DIFF
--- a/moveit_core/robot_state/src/conversions.cpp
+++ b/moveit_core/robot_state/src/conversions.cpp
@@ -389,6 +389,7 @@ bool robotStateMsgToRobotState(const Transforms& tf, const moveit_msgs::RobotSta
 
 void robotStateToRobotStateMsg(const RobotState& state, moveit_msgs::RobotState& robot_state, bool copy_attached_bodies)
 {
+  robot_state.is_diff = false;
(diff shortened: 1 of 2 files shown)


v4hn:
I approve this patch with reservations (see below).
Thank you for cleaning up after me @simonschmeisser ... :disappointed: 

On the other hand this wouldn't even be a problem if [robotStateToRobotStateMsg](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_core/robot_state/src/conversions.cpp#L390) used [here](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_core/planning_scene/src/planning_scene.cpp#L711) and [here](https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_core/planning_scene/src/planning_scene.cpp#L923) would *clear* the `is_diff` field, but apparently it does not.
Isn't this the real bug you observe here?

simonschmeisser:
I think we should actually have both fixes. Recycling the message does not give any relevant performance wins but relying on the default value is also hard to follow.

PR #980, opened 2018-07-12
Chomp needs to depend on cmake_modules. (#976)

bmagyar:
Otherwise, the binary builds on the buildfarm fail.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

Bence:
Regardless of buildfarm on Kinetic, `cmake_modules` is `find_package`-d, therefore it should be in the `package.xml` too.

### DIFF
--- a/moveit_planners/chomp/chomp_interface/package.xml
+++ b/moveit_planners/chomp/chomp_interface/package.xml
@@ -12,6 +12,7 @@
   <license>BSD</license>
 
   <buildtool_depend>catkin</buildtool_depend>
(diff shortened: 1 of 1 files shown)


rhaschke:
I think, instead of filing a new PR, we can directly cherry-pick such simple commits.
Of course, that's different, if the cherry-pick requires some branch-specific modifications.
As, I haven't seen this PR, I already cherry-picked this from melodic directly: 3ae1de44a1ce44edf9a0b4759377906c60b27556

bmagyar:
Thank you!

PR #981, opened 2018-07-12
Allow chains with only one active joint

bhomberg:
### Description

I was getting an error using a two-link chain (base link and moving link); this check currently requires more than one active joint, but one active joint still should be a valid (though very short) chain.

### Checklist
- [x ] **Required**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [n/a ] Extended the tutorials / documentation, if necessary [reference](http://moveit.ros.org/documentation/contributing/)
- [ n/a] Include a screenshot if changing a GUI


jonbinney:
Oops - we shouldn't have our branch named kinetic-devel either. I'll fix this now.

jonbinney:
Created new PR without merge commit: https://github.com/ros-planning/moveit/pull/983

PR #982, opened 2018-07-12
fix moveit_ros_perception

rhaschke:
I fixed some more compiler warnings (due to more sensitive compiler flags introduced in 0b4600e2b41e8de61b53ee96272f6eb0539aeb93, but hidden due to #943).

Unfortunately, the unittest in move_ros_perception fails. This seems to be related to a bug in Mesa OpenGL - switching to nvidia OpenGL, the test passes flawlessly.

Hence, I disable this test for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ notifications:
     recipients:
       - 130s@2000.jukuin.keio.ac.jp
 env:
+  global:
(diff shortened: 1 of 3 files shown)


rhaschke:
Hm, although I only want to disable `moveit_ros_perception` for testing, it looks like we need to blacklist all metapackages that eventually `run_depend` on `moveit_ros_perception` as well.

PR #984, opened 2018-07-13
Fix CurrentStateMonitor update callback for floating joints with non-identity joint origin

xaver-k:
### Description

This PR fixes a bug in the CurrentStateMonitor::tfCallback() method. The old implementation does not take the joint's origin into account. This leads to wrong transformations in the PlanningScene of a PlanningSceneMonitor if the transformation to a floating joint's origin is not the identity.

A demo of the bug can be found here: https://github.com/xaver-k/floating_joint_test

#### A more in-depth explanation of the bug and the fix:

We calculate the pose of a child link relative to its parent as the concatenation of the transformations `parent -> joint origin` and `joint origin -> child`, where `joint origin -> child` is dependent on the joint variable(s). With 1D-joint (rotational, prismatic) we can update the joint variable directly from the `/joint_states` topic.

In contrast, we update the joint variables of floating joints from TF. The data in TF is already the transform `child -> parent`. The old implementation saves this transformation as the joint's variables. This means that we practically apply the `child -> joint origin`-transform twice when calculating the link poses internally. This does not show as long as your floating joint's origin is just the identity, but leads to problems otherwise.

The code in the PR corrects this by applying the inverse of the `child -> joint origin`-transformation before storing the transform in the joint variables.

#### Thoughts on the implementation:
* Calculating the inverse during each update cycle should not be too bad, since practically we transpose and multiply (no numeric inverse). I did not experience any performance issues.
* If more performance is required, we could store the inverse of the joint origin in the ` moveit::core::LinkModel` and provide a new method `getJointOriginTransformInverse()`.

### Checklist
 **Note:** There are some changes that clang-format makes but none are related to the code-changes in the PR.
- [X] Extended the tutorials / documentation, if necessary (not applicable)
- [X] Include a screenshot if changing a GUI (not applicable)

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -590,55 +590,32 @@ void RobotState::updateLinkTransforms()
 
 void RobotState::updateLinkTransformsInternal(const JointModel* start)
 {
-  const std::vector<const LinkModel*>& links = start->getDescendantLinkModels();
-  if (!links.empty())
-  {
-    const LinkModel* parent = links[0]->getParentLinkModel();
+  for(const LinkModel* link : start->getDescendantLinkModels()) {
+    const LinkModel* parent = link->getParentLinkModel();
     if (parent)
     {
-      if (links[0]->parentJointIsFixed())
-        global_link_transforms_[links[0]->getLinkIndex()].matrix().noalias() =
-            global_link_transforms_[parent->getLinkIndex()].matrix() * links[0]->getJointOriginTransform().matrix();
+      if (link->parentJointIsFixed())
+        global_link_transforms_[link->getLinkIndex()].matrix().noalias() =
+            global_link_transforms_[parent->getLinkIndex()].matrix() * link->getJointOriginTransform().matrix();
       else
       {
-        if (links[0]->jointOriginTransformIsIdentity())
-          global_link_transforms_[links[0]->getLinkIndex()].matrix().noalias() =
+        if (link->jointOriginTransformIsIdentity())
+          global_link_transforms_[link->getLinkIndex()].matrix().noalias() =
               global_link_transforms_[parent->getLinkIndex()].matrix() *
-              getJointTransform(links[0]->getParentJointModel()).matrix();
+              getJointTransform(link->getParentJointModel()).matrix();
         else
-          global_link_transforms_[links[0]->getLinkIndex()].matrix().noalias() =
-              global_link_transforms_[parent->getLinkIndex()].matrix() * links[0]->getJointOriginTransform().matrix() *
-              getJointTransform(links[0]->getParentJointModel()).matrix();
+          global_link_transforms_[link->getLinkIndex()].matrix().noalias() =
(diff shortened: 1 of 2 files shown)


xaver-k:
Hmm, looks like Travis fails due to failed tests in `moveit_ros/perception/mesh_filter/test/mesh_filter_test.cpp` . As the current state of the `kinetic-devel`-branch has the exact same failures and the `mesh_filter`-package is not dependent on `planning_scene_monitor`, I guess this is not related to the changes in the PR.

rhaschke:
There was a pending PR #982 to fix (disable) the broken test. I know merged this PR and retriggered your Travis build.

xaver-k:
Travis now succeeds after a rebase of the feature branch.

rhaschke:
@xaver-k I added some minor improvement to your code (cdcc89a). Additionally, I took the chance to cleanup code in `RobotState::updateLinkTransformsInternal()` to reduce code duplication (cdcc89a).
@v4hn Please have another look.

Regarding updates of floating joints in rviz, I cannot confirm @xaver-k observations. I tested with his [nicely prepared test package](https://github.com/xaver-k/floating_joint_test), which has two floating joints, but no 1-dof joints at all. On my side, the PlanningScene's robot is shown in a pose assuming Identity transforms for those floating joints as long as they were not yet published. Hence, it doesn't complain about the missing transforms! However, any TF update immediately triggers a change of the rviz display.

In contrast, rviz' internal RobotModel display explicitly complains about the missing transforms and visualizes the corresponding links at the origin as white objects. 
@v4hn Is that what you mean by "offset between the robot model and the planning-scene model"?

xaver-k:
Hi, sorry for the late reply.
@rhaschke the `planning_scene_user`-node in [my test-package](https://github.com/xaver-k/floating_joint_test) does not publish its maintained planning scene, so the rviz display does not actually get any data from the node.

I submitted a new issue for the rviz-related behavior I am experiencing and an example-setup: #1044 . I guess we should continue the discussion there.

I with regards to the subject of this PR, are we just waiting for a new review of @v4hn ? Or is there anything else you need?

rhaschke:
Yes, I'm just waiting for a second approval by e.g. @v4hn.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -590,55 +590,32 @@ void RobotState::updateLinkTransforms()
...
-  {
-    const LinkModel* parent = links[0]->getParentLinkModel();
+  for(const LinkModel* link : start->getDescendantLinkModels()) {
+    const LinkModel* parent = link->getParentLinkModel();
     if (parent)
     {
-      if (links[0]->parentJointIsFixed())
-        global_link_transforms_[links[0]->getLinkIndex()].matrix().noalias() =
-            global_link_transforms_[parent->getLinkIndex()].matrix() * links[0]->getJointOriginTransform().matrix();
+      if (link->parentJointIsFixed())
```
v4hn: could you please add a comment to the `if(parent)` line to explain that only the root link of `start` might not have a parent?

PR #986, opened 2018-07-17
Add missing distance check functions to allValid collision checker

simonschmeisser:
some functions are missing for the allValid dummy collision checker plugin after latest changes


rhaschke:
@Levi-Armstrong Could you please review this (as this is related to your recent changes)?

simonschmeisser:
Yes, simply instantiating it would actually be a sufficient unit test ... will look at it

PR #998, opened 2018-07-24
Added capabilties as arg to move_group.launch

agutenkunst:
Enable the user to pass capabilities that should be loaded / inhibited
as arg to move_group.launch

### Description
I ran into the issue because I wanted to load capabilities into the move_group from another package.
Without this PR I would need to duplicate the move_group.launch file.
Passing it as an arg is the same way the other args are handled.


v4hn:
I do not really see the necessity for this.
You can always set
```
<rosparam name="move_group/capabilities">
my_awesome_capability
</rosparam>
```
The `move_group` namespace is pretty much enforced in MoveIt, so there is not even a problem with node renaming.

Wouldn't you agree?

PR #1008, opened 2018-07-30
added param for timeout, added retries

dg-shadow:
### Description

Added retries param to re-attempt connection to db if timeout occurs.
Also added param for timeout.

### Checklist


v4hn:
incidentally, @dg-shadow could someone from Shadow have a look at Kei's sqlite implementation for the warehouse?
https://github.com/ros-planning/moveit/issues/123#issuecomment-393830086
I believe you would also be relieved to have a lightweight replacement for the mongodb and it should remedy problems like the one you work around in this request.

PR #1010, opened 2018-07-31
publish all joints into fake_controller_joint_states, instad of joint…

k-okada:
…s incldued in current controller

### Description

FakeTrajectoryExecutionController publishes '/joint_states' only when the trajectory is interpolated, this PR changes publish the 'joint_state' all the time and also publishes all joints, instead of joint within the controllers.

### Checklist

### DIFF
--- a/moveit_plugins/moveit_fake_controller_manager/include/moveit_fake_controller_manager/follow_joint_trajectory_server_handle.h
+++ b/moveit_plugins/moveit_fake_controller_manager/include/moveit_fake_controller_manager/follow_joint_trajectory_server_handle.h
@@ -49,13 +49,12 @@ namespace moveit_fake_controller_manager
  */
 class FollowJointTrajectoryServerHandle
 {
-
 public:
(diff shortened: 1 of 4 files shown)


felixvd:
Can you explain why this is necessary or what it helps with?

rhaschke:
Could you elaborate a little bit more?
As far as I remember, for fake controllers, there is a joint_state_publisher running that reads and re-publishes joint states from fake_controller_joint_states. Thus, there you can choose the publishing rate for /joint_states. There is no need to explicitly publish fake_controller_joint_states all the time.
Further, it's common practice for complex robots (e.g. composed of a base, arms, hands) to publish individual joint sets separately - as there are independent low-level controllers doing so.

k-okada:
@rhaschke yes, you're right, we need this code once after "JTA" support patch ( https://github.com/ros-planning/moveit/pull/1001 ) , which starts JointTrajectoryAction servers

I have added "publish all joints" feature to #1001, and close #1010
https://github.com/ros-planning/moveit/pull/1001/commits/6b97db5cd8dd5f5cdd7f1a4582d9d9b0b1d1f94e

PR #1013, opened 2018-08-02
fix bug in perception screen

MohmadAyman:
### Description

1. Fixing a bug that makes the SA crashes when switching from the perception screen.
2. Improve documentation.

### Checklist

### DIFF
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -916,7 +916,7 @@ bool MoveItConfigData::output3DSensorPluginYAML(const std::string& file_path)
   emitter << YAML::BeginMap;
 
   // Make sure sensors_plugin_config_parameter_list_ is not empty
-  if (sensors_plugin_config_parameter_list_.size() == 1)
+  if (!sensors_plugin_config_parameter_list_.empty())
   {
     for (auto& parameter : sensors_plugin_config_parameter_list_[0])
     {
@@ -1506,7 +1506,7 @@ bool MoveItConfigData::input3DSensorsYAML(const std::string& default_file_path, 
   std::ifstream default_input_stream(default_file_path.c_str());
   if (!default_input_stream.good())
   {
-    ROS_ERROR_STREAM_NAMED("sensors_3d.yaml", "Unable to open file for writing " << file_path);
+    ROS_ERROR_STREAM_NAMED("sensors_3d.yaml", "Unable to open file for reading " << default_file_path);
     return false;
   }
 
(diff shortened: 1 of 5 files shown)


moveit_setup_assistant/src/widgets/perception_widget.cpp
```diff
@@ -56,24 +56,18 @@ PerceptionWidget::PerceptionWidget(QWidget* parent, moveit_setup_assistant::Move
 
   // Top Header Area ------------------------------------------------
 
-  HeaderWidget* header =
-      new HeaderWidget("3D Perception Sensor Configuration", "Configure your 3D sensors to work with Moveit! ", this);
+  HeaderWidget* header = new HeaderWidget("3D Perception Sensor Configuration",
+                                          "Configure your 3D sensors to work with Moveit! "
+                                          "Look at <a "
```
davetcoleman: "Look at" --> "Please see"

moveit_setup_assistant/src/widgets/perception_widget.cpp
```diff
@@ -56,24 +56,18 @@ PerceptionWidget::PerceptionWidget(QWidget* parent, moveit_setup_assistant::Move
...
+                                          "href='http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/"
+                                          "perception_configuration/"
+                                          "perception_configuration_tutorial.html'>Perception Documentation</a> "
+                                          "for more details",
```
davetcoleman: Add "." after "details"

davetcoleman:
I still do not see any default values when I choose a sensor plugin:

image

davetcoleman:
Bug: if I go to the "3D Perception" screen and choose "Depth Map" from the drop down list, then switch to the "Passive Joints" screen, the MSA segfaults.

MohmadAyman:
@davetcoleman
I do confirm the other bug about the `segfaults` but I can still not reproduce the no default values. 
Would you tell me what do you see if you run the `catkin test` in the setup assistant? there is actually a test case on reading the default values.

MohmadAyman:
I also cant reproduce the segmentation fault either on this branch.

davetcoleman:
Run on a new moveit_config package, not an existing one.

MohmadAyman:
Fixed the bug the occurs when using on a new package by calling the `input3DSensors` even if the package is new.

MohmadAyman:
@davetcoleman Travis is happy.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -1518,6 +1518,7 @@ bool MoveItConfigData::input3DSensorsYAML(const std::string& default_file_path,
       {
```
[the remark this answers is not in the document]
MohmadAyman: I think can do sth like this
```
    // Get sensors node
    const YAML::Node& sensors_node = doc["sensors"];
  
    // Make sure that the sensors are written as a sequence
    if (sensors_node && sensors_node.IsSequence())
```

davetcoleman:
cherry-picked to melodic

PR #1024, opened 2018-08-18
Add joint state controller config by default

MohmadAyman:
### Description
This PR improves the controllers screen by hiding not useful controller options, and renaming `moveit_sim_controllers` to more appropriate names.

### DIFF
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -725,6 +725,26 @@ void MoveItConfigData::outputFollowJointTrajectoryYAML(YAML::Emitter& emitter,
 }
 
 // ******************************************************************************************
+// Helper function to get the default start state group for moveit_sim_hw_interface
+// ******************************************************************************************
+std::string MoveItConfigData::getDefaultStartStateGroup()
+{
+  if (!srdf_->srdf_model_->getGroups().empty())
+    return srdf_->srdf_model_->getGroups()[0].name_;
+  return "todo_no_group_selected";
+}
+
+// ******************************************************************************************
+// Helper function to get the default start pose for moveit_sim_hw_interface
+// ******************************************************************************************
+std::string MoveItConfigData::getDefaultStartPose()
+{
+  if (!srdf_->group_states_.empty())
+    return srdf_->group_states_[0].name_;
+  return "todo_no_pose_selected";
+}
+
+// ******************************************************************************************
 // Output controllers config files
 // ******************************************************************************************
 bool MoveItConfigData::outputROSControllersYAML(const std::string& file_path)
@@ -771,17 +791,11 @@ bool MoveItConfigData::outputROSControllersYAML(const std::string& file_path)
     {
       // Use the first planning group if initial joint_model_group was not set, else write a default value
       emitter << YAML::Key << "joint_model_group";
-      if (!srdf_->srdf_model_->getGroups().empty())
-        emitter << YAML::Value << srdf_->srdf_model_->getGroups()[0].name_;
-      else
-        emitter << YAML::Value << "todo_no_group_selected";
+      emitter << YAML::Value << getDefaultStartStateGroup();
 
       // Use the first robot pose if initial joint_model_group_pose was not set, else write a default value
       emitter << YAML::Key << "joint_model_group_pose";
-      if (!srdf_->group_states_.empty())
-        emitter << YAML::Value << srdf_->group_states_[0].name_;
-      else
-        emitter << YAML::Value << "todo_no_pose_selected";
+      emitter << YAML::Value << getDefaultStartPose();
 
       emitter << YAML::EndMap;
     }
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -765,19 +765,29 @@ bool MoveItConfigData::outputROSControllersYAML(const std::string& file_path)
   emitter << YAML::Value << YAML::BeginMap;
 
   {
-    emitter << YAML::Comment("MoveIt-specific simulation settings");
+    emitter << YAML::Comment("Simulation settings for using moveit_sim_controllers");
     emitter << YAML::Key << "moveit_sim_hw_interface" << YAML::Value << YAML::BeginMap;
     // Moveit Simulation Controller settings for setting initial pose
     {
+      // Use the first planning group if initial joint_model_group was not set, else write a default value
       emitter << YAML::Key << "joint_model_group";
-      emitter << YAML::Value << "controllers_initial_group_";
+      if (!srdf_->srdf_model_->getGroups().empty())
+        emitter << YAML::Value << srdf_->srdf_model_->getGroups()[0].name_;
+      else
+        emitter << YAML::Value << "todo_no_group_selected";
+
+      // Use the first robot pose if initial joint_model_group_pose was not set, else write a default value
       emitter << YAML::Key << "joint_model_group_pose";
-      emitter << YAML::Value << "controllers_initial_pose_";
+      if (!srdf_->group_states_.empty())
+        emitter << YAML::Value << srdf_->group_states_[0].name_;
+      else
+        emitter << YAML::Value << "todo_no_pose_selected";
+
       emitter << YAML::EndMap;
     }
     // Settings for ros_control control loop
     emitter << YAML::Newline;
-    emitter << YAML::Comment("Settings for ros_control control loop");
+    emitter << YAML::Comment("Settings for ros_control_boilerplate control loop");
     emitter << YAML::Key << "generic_hw_control_loop" << YAML::Value << YAML::BeginMap;
     {
       emitter << YAML::Key << "loop_hz";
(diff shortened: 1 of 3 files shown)


moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -763,19 +763,29 @@ bool MoveItConfigData::outputROSControllersYAML(const std::string& file_path)
...
     emitter << YAML::Key << "moveit_sim_hw_interface" << YAML::Value << YAML::BeginMap;
     // Moveit Simulation Controller settings for setting initial pose
     {
+      // Use the first planning group if initial joint_model_group was not set, else write a default value
       emitter << YAML::Key << "joint_model_group";
-      emitter << YAML::Value << "controllers_initial_group_";
+      if (!srdf_->srdf_model_->getGroups().empty())
+        emitter << YAML::Value << srdf_->srdf_model_->getGroups()[0].name_;
+      else
+        emitter << YAML::Value << "todo_no_group_selected";
```
davetcoleman: make this into a helper function: ``getDefaultStartStateGroup()``

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -763,19 +763,29 @@ bool MoveItConfigData::outputROSControllersYAML(const std::string& file_path)
...
+
+      // Use the first robot pose if initial joint_model_group_pose was not set, else write a default value
       emitter << YAML::Key << "joint_model_group_pose";
-      emitter << YAML::Value << "controllers_initial_pose_";
+      if (!srdf_->group_states_.empty())
+        emitter << YAML::Value << srdf_->group_states_[0].name_;
+      else
+        emitter << YAML::Value << "todo_no_pose_selected";
```
davetcoleman: make this into a helper function: ``getDefaultStartPose()``

moveit_setup_assistant/src/widgets/controller_edit_widget.cpp
```diff
@@ -195,12 +195,11 @@ void ControllerEditWidget::loadControllersTypesComboBox()
...
 
-  const std::array<std::string, 9> default_types = {
+  const std::array<std::string, 8> default_types = {
     "effort_controllers/JointTrajectoryController",   "effort_controllers/JointPositionController",
     "effort_controllers/JointVelocityController",     "effort_controllers/JointEffortController",
-    "joint_state_controller/JointStateController",    "position_controllers/JointPositionController",
-    "position_controllers/JointTrajectoryController", "velocity_controllers/JointTrajectoryController",
-    "velocity_controllers/JointvelocityController"
+    "position_controllers/JointPositionController",   "position_controllers/JointTrajectoryController",
+    "velocity_controllers/JointTrajectoryController", "velocity_controllers/JointvelocityController"
```
davetcoleman: i don't think this is a good work-around for a bug

davetcoleman:
I don't understand why you want to remove the joint state controller - it should still be an option

moveit_setup_assistant/src/widgets/controller_edit_widget.cpp
MohmadAyman: Not a work around, we already add the joint state controllers by default  and it doesnt need the user to add anything to it -it only has a publish rate, right?-. Why would we be adding it to the types of the new conrollers?

MohmadAyman:
@davetcoleman I didn't leave it as an option because the controller fields doesn't need any input from the user, it is just `publish_rate` and all the joints.
I found most controllers configs has this controller, if you believe it should be an option and not added by default, I could  easily modify it to be that way.

MohmadAyman:
I believe this is what the joint state controller usually looks like.
```
joint_state_controller:
   type:         joint_state_controller/JointStateController
   publish_rate: 50
```

moveit_setup_assistant/src/widgets/controller_edit_widget.cpp
davetcoleman: "we already add the joint state controllers by default"

you mean when that button is pushed? i don't like that there is a hidden controller, we should remove that behavior and require that it be manually added

davetcoleman:
Yes that is appears correct.

When I tested this locally it had other files added also - perhaps you have fixed this?

moveit_setup_assistant/src/widgets/controller_edit_widget.cpp
davetcoleman: correct, it should not change

moveit_setup_assistant/test/moveit_config_data_test.cpp
```diff
@@ -100,8 +100,12 @@ TEST_F(MoveItConfigData, ReadingControllers)
   // Test that addDefaultControllers() did accually add a controller for the new_group
   EXPECT_EQ(config_data_->getROSControllers().size(), group_count);
 
+  // Make a temp dir to write ros_controllers file to
+  char file_name[] = "/tmp/ros_controller.XXXXXX.yaml";
+  int fd = mkstemp(file_name);
```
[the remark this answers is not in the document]
MohmadAyman: Locally it works without having to use `mkstemp`, and it writes the file to `/tmp`. Is there a way to test if it would work on the build server or not?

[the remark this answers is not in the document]
rhaschke: Usually /tmp is writable for all users. 
There is no way to check whether it will work on Travis than testing on Travis ;-)
I was asking for deleting the file after you are done. (Always cleanup behind you.)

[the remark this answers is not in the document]
MohmadAyman: Yeah thanks, I re-added the remove file command, I thought `mkstemp` will automatically delete it, but after testing, `mkstemp` didn't actually delete it.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
davetcoleman: ping

moveit_setup_assistant/src/tools/moveit_config_data.cpp
davetcoleman: ping

henningkayser:
@MohmadAyman It looks like you addressed all requested changes, correct?
@davetcoleman @rhaschke would you approve this PR?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -724,6 +724,26 @@ void MoveItConfigData::outputFollowJointTrajectoryYAML(YAML::Emitter& emitter,
...
 }
 
+// ******************************************************************************************
+// Helper function to get the default start state group for moveit_sim_hw_interface
+// ******************************************************************************************
+std::string MoveItConfigData::getDefaultStartStateGroup()
+{
+  if (!srdf_->srdf_model_->getGroups().empty())
+    return srdf_->srdf_model_->getGroups()[0].name_;
+  return "todo_no_group_selected";
```
rhaschke: @MohmadAyman, what do you plan to do with this name?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -724,6 +724,26 @@ void MoveItConfigData::outputFollowJointTrajectoryYAML(YAML::Emitter& emitter,
...
+}
+
+// ******************************************************************************************
+// Helper function to get the default start pose for moveit_sim_hw_interface
+// ******************************************************************************************
+std::string MoveItConfigData::getDefaultStartPose()
+{
+  if (!srdf_->group_states_.empty())
+    return srdf_->group_states_[0].name_;
+  return "todo_no_pose_selected";
```
rhaschke: Same here.

rhaschke:
Looks like the test `moveit_setup_assistant/test/moveit_config_data_test.cpp` disappeared. Why?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
MohmadAyman: Well this name should be added to the `moveit_sim_controllers` if there was no planning groups to avoid writing blank.

MohmadAyman:
I don't think this PR had anything to do with `moveit_setup_assistant/test/moveit_config_data_test.cpp`

moveit_setup_assistant/src/tools/moveit_config_data.cpp
rhaschke: I suggest to issue a warning in MSA when no group was selected. Also, in this case, the sim_controllers shouldn't be written?

henningkayser:
> I don't think this PR had anything to do with `moveit_setup_assistant/test/moveit_config_data_test.cpp`

I can't see that it has been deleted either. @rhaschke?

rhaschke:
I had the impression that a previous version of this PR was comprising the test.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
MohmadAyman: I'm not 100% sure I remember correctly, but I think the reason we went for writing a dummy group name, was to make the output controllers file more of a template for users, so when a user is to use a `sim_controller` will just have to edit the name of the group.
also a warring sounds nice. - but I cant access my dev computer for a few days, maybe someone add it and I'll approve-

rhaschke:
Looks like @davetcoleman's review could be dismissed. His comments are addressed.

rhaschke:
Thanks.

davetcoleman:
+1!

PR #1031, opened 2018-08-22
removed stuff marked to-be-removed in Lunar

rhaschke:


v4hn:
I approve, but could you please add an item to the `MIGRATION.md` with this commit that says which functions got removed and how they should be replaced?
We deprecated them before, but it might still create confusion.

PR #1033, opened 2018-08-23
Fix #868: calling service capability locks PlanningSceneMonitor

### LINKED ISSUE #868: 'Failed to fetch current robot state' when using the 'plan_kinematic_path' service

rhaschke:
As described in https://github.com/ros-planning/moveit/issues/868#issuecomment-385127395, ROS service capabilities will essentially block the global event queue. Thus, key components as the PlanningSceneMonitor (and the associated CurrentStateMonitor) should use their own event queue and spinner to become independent of the global queue.

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
+++ b/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
@@ -38,6 +38,7 @@
 #define MOVEIT_PLANNING_SCENE_MONITOR_PLANNING_SCENE_MONITOR_
 
 #include <ros/ros.h>
+#include <ros/callback_queue.h>
 #include <tf2_ros/message_filter.h>
 #include <tf2_ros/buffer.h>
 #include <message_filters/subscriber.h>
@@ -450,6 +451,8 @@ protected:
 
   ros::NodeHandle nh_;
   ros::NodeHandle root_nh_;
+  ros::CallbackQueue queue_;
+  ros::AsyncSpinner spinner_;
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
 
   std::string robot_description_;
--- a/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
+++ b/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
@@ -144,6 +144,20 @@ public:
                        const std::shared_ptr<tf2_ros::Buffer>& tf_buffer = std::shared_ptr<tf2_ros::Buffer>(),
                        const std::string& name = "");
 
+  /** @brief Constructor
+   *  @param scene The scene instance to maintain up to date with monitored information
+   *  @param rml A pointer to a kinematic model loader
+   *  @param nh external parent NodeHandle
+   *         The monitors will use this NodeHandle's CallbackQueue for updates.
+   *         Usually, this should be a different queue than the global queue, otherwise you might run into timeouts.
+   *  @param tf_buffer A pointer to a tf2_ros::Buffer
+   *  @param name A name identifying this planning scene monitor
+   */
+  PlanningSceneMonitor(const planning_scene::PlanningScenePtr& scene,
+                       const robot_model_loader::RobotModelLoaderPtr& rml, const ros::NodeHandle& nh,
+                       const std::shared_ptr<tf2_ros::Buffer>& tf_buffer = std::shared_ptr<tf2_ros::Buffer>(),
+                       const std::string& name = "");
+
   ~PlanningSceneMonitor();
 
   /** \brief Get the name of this monitor */
@@ -452,7 +466,7 @@ protected:
   ros::NodeHandle nh_;
   ros::NodeHandle root_nh_;
   ros::CallbackQueue queue_;
-  ros::AsyncSpinner spinner_;
+  std::shared_ptr<ros::AsyncSpinner> spinner_;
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
 
   std::string robot_description_;
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -120,28 +120,24 @@ const std::string planning_scene_monitor::PlanningSceneMonitor::MONITORED_PLANNI
 planning_scene_monitor::PlanningSceneMonitor::PlanningSceneMonitor(const std::string& robot_description,
                                                                    const std::shared_ptr<tf2_ros::Buffer>& tf_buffer,
                                                                    const std::string& name)
-  : monitor_name_(name), nh_("~"), tf_buffer_(tf_buffer)
+  : PlanningSceneMonitor(planning_scene::PlanningScenePtr(), robot_description, tf_buffer, name)
 {
-  rm_loader_.reset(new robot_model_loader::RobotModelLoader(robot_description));
-  initialize(planning_scene::PlanningScenePtr());
 }
 
 planning_scene_monitor::PlanningSceneMonitor::PlanningSceneMonitor(const planning_scene::PlanningScenePtr& scene,
                                                                    const std::string& robot_description,
                                                                    const std::shared_ptr<tf2_ros::Buffer>& tf_buffer,
                                                                    const std::string& name)
-  : monitor_name_(name), nh_("~"), tf_buffer_(tf_buffer)
+  : PlanningSceneMonitor(scene, std::make_shared<robot_model_loader::RobotModelLoader>(robot_description), tf_buffer,
+                         name)
 {
-  rm_loader_.reset(new robot_model_loader::RobotModelLoader(robot_description));
-  initialize(scene);
 }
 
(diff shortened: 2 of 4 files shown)


v4hn:
I actually forgot about this issue.. Looks good!

However, there have been complains by users in the past when MoveIt classes start their own threads unconditionally.
I believe we should provide another constructor here to use a user-provided CallbackQueue instead.

rhaschke:
Passing a user-provided CallbackQueue, will not stop creating a thread by the internal spinner.
I think there is no chance to avoid conflicts without another thread (as the main spinner thread blocks on the global callback queue). 
Regarding the complains: All action-server-based capabilities will spawn an own thread as well...

rhaschke:
Would you prefer to pass a CallbackQueue or a NodeHandle (as in other locations of the API)?
NodeHandle might also change the namespace, which is undesirable...

v4hn:
`NodeHandle` is definitely the more common interface for something like this, but then we have `root_nh_` as well as `nh_`.
We don't want to pass two handles (I guess), the alternatives being
(1) taking a root node-handle for the constructor, creating a private handle internally and copy the queue from the root nh we received
(2) pass the queue directly and set it for both nodes
(3) similar to MoveGroupInterface we could have an `Options` struct that can contain both node-handles

I would say (2) is the simplest option.

rhaschke:
There is no problem in using (1). I think, it's not even necessary to copy the the event queue explicitly if we initialize the private handle from the root: `nh_(root_nh, "~")`.
However, passing a root handle with a custom/wrong namespace set, will probably break everything. Hence, option (2) might be safer.

rhaschke:
@v4hn I added a new constructor to accept an external NodeHandle. Can you please review?

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -171,6 +177,7 @@ planning_scene_monitor::PlanningSceneMonitor::~PlanningSceneMonitor()
   parent_scene_.reset();
   robot_model_.reset();
   rm_loader_.reset();
+  spinner_.stop();
```
v4hn: Is it well defined what happens if this was never started?

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -120,34 +120,41 @@ const std::string planning_scene_monitor::PlanningSceneMonitor::MONITORED_PLANNI
...
+  root_nh_.setCallbackQueue(&queue_);
+  nh_.setCallbackQueue(&queue_);
+  spinner_.start();
+  initialize(scene);
+}
+
+planning_scene_monitor::PlanningSceneMonitor::PlanningSceneMonitor(
+    const planning_scene::PlanningScenePtr& scene, const robot_model_loader::RobotModelLoaderPtr& rm_loader,
+    const ros::NodeHandle& nh, const std::shared_ptr<tf2_ros::Buffer>& tf_buffer, const std::string& name)
+  : monitor_name_(name), nh_(nh, "~"), root_nh_(nh), spinner_(1, &queue_), tf_buffer_(tf_buffer), rm_loader_(rm_loader)
```
v4hn: Tested on kinetic:
```
$ cat src/mytest/src/mytest.cpp
#include <ros/ros.h>

int main(int argc, char **argv) {
  ros::init(argc, argv, "mytest");

  ros::NodeHandle nh;
  ros::NodeHandle pnh(nh, "~");
  return 0;
}
$ rosrun mytest mytest
terminate called after throwing an instance of 'ros::InvalidNameException'
  what():  Using ~ names with NodeHandle methods is not allowed.  If you want to use private names with the NodeHandle interface, construct a NodeHandle using a private name as its namespace.  e.g. ros::NodeHandle nh("~");  nh.getParam("my_private_name"); (name = [~])
Aborted (core dumped)
```

I guess you did not add a test for this?
Maybe they changed this in melodic, but I doubt it.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -143,6 +144,20 @@ class PlanningSceneMonitor : private boost::noncopyable
                        const std::shared_ptr<tf2_ros::Buffer>& tf_buffer = std::shared_ptr<tf2_ros::Buffer>(),
                        const std::string& name = "");
 
+  /** @brief Constructor
+   *  @param scene The scene instance to maintain up to date with monitored information
+   *  @param rml A pointer to a kinematic model loader
+   *  @param nh external parent NodeHandle
+   *         Ensure that this NodeHandle's CallbackQueue is different from the global queue and that
+   *         it is served by an external spinner!
```
v4hn: This mechanism can also be used to explicitly run in the global queue, if the user wants it that way (we did it for a long time and it's not necessarily a problem).
Can you change this to just state that the Monitors will be updated with the queue of this node handle?

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
rhaschke: Yes. AsyncSpinner just doesn't do anything:
http://docs.ros.org/melodic/api/roscpp/html/spinner_8cpp_source.html#l00289

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
rhaschke: Fixed.

rhaschke:
@v4hn Is there anything left that keeps you from approving and merging this?

PR #1051, opened 2018-09-06
Added demo_gazebo.launch

MohmadAyman:
### Description
This PR is about connecting the `ROSControl` and `Simulation` widgets, the following GIF sums up what this PR do.
  

[or as video for better quality](https://github.com/MohmadAyman/school/blob/c439b75a5f0f0f5c0daf8afa9b61b69822dbd045/sim-4.ogv)
.
### Checklist


moveit_setup_assistant/templates/moveit_config_pkg_template/launch/moveit_controller_manager.launch.xml
```diff
@@ -1,3 +1,6 @@
 <launch>
```
MohmadAyman: Is modifying this file okay? like would it affect other already running stuff?

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo_gazebo.launch
```diff
@@ -0,0 +1,69 @@
+<launch>
```
MohmadAyman: There is another option of editing the current `demo.launch` and add options to it to be able to launch `gazebo` too.

moveit_setup_assistant/src/widgets/start_screen_widget.cpp
```diff
@@ -437,10 +437,11 @@ bool StartScreenWidget::loadExistingFiles()
   // Load 3d_sensors config file
   load3DSensorsFile();
 
+  // TODO: after removing the robot namespace from ros_controllers.yaml, inputROSControllersYAML needs to be refactored
   // Load ros controllers yaml file if available-----------------------------------------------
-  fs::path ros_controllers_yaml_path = config_data_->config_pkg_path_;
-  ros_controllers_yaml_path /= "config/ros_controllers.yaml";
-  config_data_->inputROSControllersYAML(ros_controllers_yaml_path.make_preferred().native().c_str());
+  // fs::path ros_controllers_yaml_path = config_data_->config_pkg_path_;
```
MohmadAyman: as parsing `ros_controllers.yaml` is not crucial, I think it is better to disable it for now until we get this merged.

moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
```diff
@@ -131,9 +131,9 @@ SetupAssistantWidget::SetupAssistantWidget(QWidget* parent, boost::program_optio
   nav_name_list_ << "Robot Poses";
   nav_name_list_ << "End Effectors";
   nav_name_list_ << "Passive Joints";
```
MohmadAyman: Reorder the nav bar as it makes more sense this was, as the simulation screen uses the controllers.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -760,10 +832,6 @@ bool MoveItConfigData::outputROSControllersYAML(const std::string& file_path)
   YAML::Emitter emitter;
   emitter << YAML::BeginMap;
 
-  // Start with the robot name  ---------------------------------------------------
```
MohmadAyman: According to Gazebo docs it is better to have each Robot controllers under the namespace : `robot_name`, but  `moveit_controller_manager` looks for the controllers under `/controller_list`.
So in favor of the `moveit_controller_manager` the controllers and `/controller_list` will be with absolute namespace e.g. robot name will no longer be the name space of the `controller_list`.

moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
```diff
@@ -327,6 +327,12 @@ class MoveItConfigData
   bool outputROSControllersYAML(const std::string& file_path);
   bool output3DSensorPluginYAML(const std::string& file_path);
 
+  /**
+   * \brief Helper function to get the controller that is controlling the joint
+   * \return controller type
+   */
+  std::string getJointHardwareInterface(std::string joint_name);
```
davetcoleman: strings should always be passed in with const &

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -374,6 +374,28 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
   return true;  // file created successfully
 }
 
+// ******************************************************************************************
+// Helper function to get the controller that is controlling the joint
+// ******************************************************************************************
+std::string MoveItConfigData::getJointHardwareInterface(std::string joint_name)
+{
+  for (size_t i = 0; i < ros_controllers_config_.size(); i++)
```
davetcoleman: ++i is slightly faster so always prefer that :)

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -374,6 +374,28 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
...
+  {
+    std::vector<std::string>::iterator joint_it =
+        std::find(ros_controllers_config_[i].joints_.begin(), ros_controllers_config_[i].joints_.end(), joint_name);
+    if (joint_it != ros_controllers_config_[i].joints_.end())
+    {
+      if (ros_controllers_config_[i].type_.substr(0, 6) == "positi")
+        return "hardware_interface/PositionJointInterface";
+      else if (ros_controllers_config_[i].type_.substr(0, 6) == "veloci")
```
davetcoleman: why only check 6 characters and not 8?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -374,6 +374,28 @@ bool MoveItConfigData::outputKinematicsYAML(const std::string& file_path)
...
+        return "hardware_interface/VelocityJointInterface";
+      else
+        return "hardware_interface/EffortJointInterface";
+    }
+  }
+  return "hardware_interface/EffortJointInterface";
```
davetcoleman: add comment about why effort is default

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -387,34 +409,84 @@ std::string MoveItConfigData::getGazeboCompatibleURDF()
...
-    for (TiXmlElement* link_element = urdf_document.RootElement()->FirstChildElement(); link_element != NULL;
-         link_element = link_element->NextSiblingElement())
+    for (TiXmlElement* doc_element = urdf_document.RootElement()->FirstChildElement(); doc_element != NULL;
+         doc_element = doc_element->NextSiblingElement())
     {
-      if (((std::string)link_element->Value()).find("link") == std::string::npos)
-        continue;
-      // Before adding inertial elements, make sure there is none and the link has collision element
-      if (link_element->FirstChildElement("inertial") == NULL && link_element->FirstChildElement("collision") != NULL)
+      if (((std::string)doc_element->Value()).find("link") != std::string::npos)
```
davetcoleman: use static_cast instead of C-style casting

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -387,34 +409,84 @@ std::string MoveItConfigData::getGazeboCompatibleURDF()
...
 
-        inertia_link.InsertEndChild(mass);
-        inertia_link.InsertEndChild(inertia_joint);
+          inertia_link.InsertEndChild(mass);
+          inertia_link.InsertEndChild(inertia_joint);
 
-        link_element->InsertEndChild(inertia_link);
+          doc_element->InsertEndChild(inertia_link);
+        }
+        // Don't check the next if condition
```
davetcoleman: clarify this comment... why not?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -387,34 +409,84 @@ std::string MoveItConfigData::getGazeboCompatibleURDF()
...
+        continue;
+      }
+      if (((std::string)doc_element->Value()).find("joint") != std::string::npos)
```
davetcoleman: static_cast

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -387,34 +409,84 @@ std::string MoveItConfigData::getGazeboCompatibleURDF()
...
+        }
       }
     }
+
+    // Add gazebo_ros_control plugin which reads the transmission tags
+    TiXmlElement gazebo("gazebo");
+    TiXmlElement plugin("plugin");
+    TiXmlElement robot_namespace("robotNamespace");
+
+    plugin.SetAttribute("name", std::string("gazebo_ros_control"));
```
davetcoleman: is std::string really needed here? its not used in the following line

moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
```diff
@@ -488,6 +488,24 @@ bool ConfigurationFilesWidget::loadGenFiles()
...
+  file.description_ = "Gazebo launch file which also launches ros_controllers and sends robot urdf to param server, "
+                      "then using gazebo_ros pkg the robot is spawned to Gazebo";
+  file.gen_func_ = boost::bind(&ConfigurationFilesWidget::copyTemplate, this, template_path, _1);
+  gen_files_.push_back(file);
+
+  // demo_gazebo.launch ------------------------------------------------------------------
+  file.file_name_ = "demo_gazebo.launch";
+  file.rel_path_ = config_data_->appendPaths(launch_path, file.file_name_);
+  template_path = config_data_->appendPaths(template_launch_path, file.file_name_);
+  file.description_ = "Run a demo of MoveIt. with gazebo and rviz";
```
davetcoleman: "Run a demo of MoveIt! with Gazebo and Rviz"

moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
davetcoleman: love it

moveit_setup_assistant/src/widgets/start_screen_widget.cpp
```diff
@@ -437,10 +437,11 @@ bool StartScreenWidget::loadExistingFiles()
   // Load 3d_sensors config file
   load3DSensorsFile();
 
+  // TODO: after removing the robot namespace from ros_controllers.yaml, inputROSControllersYAML needs to be refactored
```
davetcoleman: add your username to the TODO

TODO(MohmadAyman)

moveit_setup_assistant/src/widgets/start_screen_widget.cpp
davetcoleman: instead of making a Github comment about this, put this comment in the code so that it isn't lost!

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo_gazebo.launch
davetcoleman: I think this second option will make demo.launch easier to maintain, rather than having a fork for gazebo

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/moveit_controller_manager.launch.xml
davetcoleman: add a comment explaining that the contents of this file can be replaced by other controller managers as desired, but that this is a good default for many integrations including Gazebo

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/ros_controllers.launch
```diff
@@ -6,12 +6,6 @@
 
   <!-- Load the controllers -->
   <node name="controller_spawner" pkg="controller_manager" type="spawner" respawn="false"
-    output="screen" ns="[ROBOT_NAME]" args="[ROS_CONTROLLERS]"/>
-
-  <!-- Convert joint states to TF transforms for rviz, etc -->
-  <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher"
-    respawn="false" output="screen">
-    <remap from="/joint_states" to="[ROBOT_NAME]/joint_states" />
-  </node>
```
davetcoleman: why remove this?

moveit_setup_assistant/test/moveit_config_data_test.cpp
```diff
@@ -77,49 +77,52 @@ class MoveItConfigData : public testing::Test
   moveit::core::RobotModelPtr robot_model;
 };
 
-TEST_F(MoveItConfigData, ReadingControllers)
-{
-  boost::filesystem::path res_path(MOVEIT_TEST_RESOURCES_DIR);
+// TODO: after removing the robot namespace from ros_controllers.yaml, inputROSControllersYAML needs to be refactored
+// accordingly
```
davetcoleman: disabling tests is bad practice. can you just refactor inputROSControllersYAML now?

davetcoleman:
Nice video!

moveit_setup_assistant/test/moveit_config_data_test.cpp
MohmadAyman: I tried tho but it wasn't so trivial, so as to get this PR merged as fast as possible I thought of postponed it to a later PR.
I can try to allocate time to do it across this weekend  :thinking:

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo_gazebo.launch
gavanderhoorn: May I suggest to *not* change `demo.launch`? MoveIt packages currently already have quite a few dependencies that make them rather involved to deploy on headless machine (fi). If we start extending `demo.launch` to include Gazebo functionality, that would make Gazebo an `exec_depend` of every MoveIt package.

Gazebo definitely has its usages, but making it a hard dependency of a MoveIt package as it gets generated by the authoratitive tool for such packages seems like a bit of a heavyweight thing to do.

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/gazebo.launch
```diff
@@ -1,17 +1,18 @@
...
-    <arg name="gui" value="$(arg gui)"/>
+    <arg name="gui" value="$(arg gazebo_gui)"/>
   </include>
 
   <!-- send robot urdf to param server -->
-  <param name="robot_description" [URDF_LOAD_ATTRIBUTE] />
+  <param name="robot_description" textfile="$(arg urdf_path)" />
 
   <!-- push robot_description to factory and spawn robot in gazebo at the origin, change x,y,z arguments to spawn in a different position -->
   <node name="spawn_gazebo_model" pkg="gazebo_ros" type="spawn_model" args="-urdf -param robot_description -model robot -x 0 -y 0 -z 0"
```
gavanderhoorn: This file seems to introduce a run dependency on `gazebo_ros`, but the PR does not seem to update the `package.xml` template to include that dependency.

If that was on purpose (fi to keep the nr of dependencies down and make Gazebo support optional), then I believe it would be good to document that somewhere to make it explicit.

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo_gazebo.launch
MohmadAyman: Also answering your comment [here](https://github.com/ros-planning/moveit/pull/1051#discussion_r216137735) .
We are not planning to make gazebo a hard dependency in anyway, it is an optional simulator which is disabled by default. if we ended up modifying the current `demo.launch` , gazebo will only run if the user want that by passing an argument to the launch file.

moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
MohmadAyman: mb :sweat_smile:

moveit_setup_assistant/src/tools/moveit_config_data.cpp
MohmadAyman: yeah you are right, would be cleaner to write the whole word.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
MohmadAyman: It is not in the `else` because it is default, it has to be an effort controller if not velocity or position. Controller type is chosen from a drop down and it has to be one of the three.

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/ros_controllers.launch
MohmadAyman: There is already a `robot_state_publisher` node in `demo.launch`.
removing the `namespace` is for the reason explained [here](https://github.com/ros-planning/moveit/pull/1051/files#diff-b722193e66bd85583ac56b9d68186447L763)

moveit_setup_assistant/test/moveit_config_data_test.cpp
MohmadAyman: Done

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo_gazebo.launch
davetcoleman: I suggest in the package.xml we add the following:

davetcoleman:
I'm ready to approve after https://github.com/ros-planning/moveit/pull/1051#discussion_r216516881 is addressed

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo_gazebo.launch
MohmadAyman: I agree, added.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
davetcoleman: sorry, i meant add a comment to the code. just copy what you just said here

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -1276,12 +1341,13 @@ bool MoveItConfigData::parseROSController(const YAML::Node& controller)
...
 // ******************************************************************************************
-bool MoveItConfigData::processROSControllers(const YAML::Node& controllers)
+bool MoveItConfigData::processROSControllers(std::ifstream& input_stream)
 {
   // Used in parsing ROS controllers
   ROSControlConfig control_setting;
+  YAML::Node controllers = YAML::Load(input_stream);
 
-  // Loop through all controllers
+  // // Loop through all controllers
```
davetcoleman: too many //

davetcoleman:
clang test failed

MohmadAyman:
@davetcoleman ping

MohmadAyman:
Can I ask what does it mean that only `ROS_DISTRO=lunar` and `REPO=ros` fails while all the others passes in Travis?

davetcoleman:
It means on the older ROS Lunar distro there is some bug. Lunar runs on  Ubuntu 17.04, see http://wiki.ros.org/lunar

Looking at the travis logs, it looks like a fluke issue:
> CMakeFiles/test_constraint_samplers.dir/test/pr2_arm_ik.cpp.o: file not recognized: File truncated
collect2: error: ld returned 1 exit status

I've retriggered the build

davetcoleman:
I've just restart the build since it was broken on lunar, which is no longer being tested for.

If it passes Ill merge since @gavanderhoorn reviewed previously but went MIA

MohmadAyman:
It still fails on lunar :thinking:

davetcoleman:
Then I think you need to rebase this PR against kinetic-devel (or move it to melodic).

The Travis has sense been updated: https://github.com/ros-planning/moveit/blob/kinetic-devel/.travis.yml

MohmadAyman:
@davetcoleman  that worked!

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo_gazebo.launch
```diff
@@ -0,0 +1,70 @@
...
+    <arg name="load_robot_description" value="false"/>
+  </include>
+
+  <!-- If needed, broadcast static tf for robot root -->
+  [VIRTUAL_JOINT_BROADCASTER]
+
+  <!-- We do not have a robot connected, so publish fake joint states -->
+  <node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher">
+    <param name="use_gui" value="$(arg use_gui)"/>
+    <rosparam param="source_list">[/joint_states]</rosparam>
```
gavanderhoorn: Late here as well, and I'm not entirely sure I understand this completely, but doesn't this make `joint_state_publisher` subscribe to the `joint_states` topic, to which it also publishes itself?

PR #1054, opened 2018-09-08
towards python3 support

v4hn:
This series of (hopefully backward-compatible) patches enables us to build MoveIt with python 3 if the underlying ROS distribution was compiled with python 3 too.

It also fixes a number of common python 2/3 compatibility issues.

Additionally it supports the new boost-style for the python component.

With this I can successfully build MoveIt in python3 and run the moveit_commander_cmdline.py to get a prompt.
There is still open issues (if you try to "use" a group, the shell dies...), but I will leave those to someone else.


timonegk:
VERSION_GREATER_EQUAL was introduced in cmake 3.7. The minimum required version here is 2.8.12. That is why the CI fails. You can use `if(NOT (Boost_VERSION LESS 106700))` instead. (Or swap if and else and drop the NOT.)

felixvd:
If you write out the open issues and how to compile with python3, maybe someone could pursue this on WMD?

v4hn:
Thanks for looking up the minimum version @timonegk .
I didn't think (1) the VERSION conditionals were that new and (2) ubuntu xenial only provides cmake 3.5.
The problem is not so much the minimum_required version but the one installed on the system...

I changed it as you suggested.

@felixvd That was the plan, feel free to look into it. :)
It's quite a bit of compile time/work though to setup a full ROS python3 ROS workspace,
so as a WMD task this has to be somewhat prepared I guess.

moveit_commander/bin/moveit_commander_cmdline.py
```diff
@@ -10,6 +12,11 @@
 import argparse
 from moveit_commander import MoveGroupCommandInterpreter, MoveGroupInfoLevel, roscpp_initialize, roscpp_shutdown
 
+# compatibility with python3
+# python2's input function is dangerous anyway
+if hasattr(__builtins__, 'raw_input'):
```
mlautman: \_\_builtin\_\_
 https://stackoverflow.com/a/11181616

v4hn:
Sorry for the delay.
I addressed your comment, @mlautman

PR #1056, opened 2018-09-09
Check errors when loading scene geometry from files

jonbinney:
### Description

Before this PR, there was no error checking when loading planning scene geometry from a text file. This meant the caller couldn't tell whether the planning scene had been successfully loaded.

Not sure if this should be cherry-picked. It changes the return type of two PlanningScene member functions, so I think it would require people to re-compile their code.

### Checklist

### DIFF
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -670,7 +670,7 @@ public:
   void saveGeometryToStream(std::ostream& out) const;
 
   /** \brief Load the geometry of the planning scene from a stream */
-  bool loadGeometryFromStream(std::istream &in);
+  bool loadGeometryFromStream(std::istream& in);
 
   /** \brief Load the geometry of the planning scene from a stream at a certain location using offset*/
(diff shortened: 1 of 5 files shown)


felixvd:
I like helpful error messages, so I like this. Is there anything to update in the visualization tab?

jonbinney:
@felixvd do you mean for the rviz plugin? I hadn't looked at that - I was actually using this function directly from a non-gui program.

jonbinney:
@bmagyar good point - I'll add a couple unit tests..

jonbinney:
@bmagyar I've added a couple tests. Not exhaustive by any means, but should do a decent job of sanity checking the code.

rhaschke:
I changed usages of the method in existing MoveIt! code to consider the return value.

jonbinney:
@rhaschke thanks for updating the rviz plugin to use this! Switching to exceptions seems like a bigger change - other functions in `PlanningScene` like `processAttachedCollisionObjectMsg` use a return value (bool) to indicate success/failure and I figured it would be best to keep things similar.

jonbinney:
(if you do want exceptions here though, let me know and I'm happy to update the PR)

PR #1061, opened 2018-09-11
Chomp trajectory seed

thibaultbarbie:
### Description

In the moveit message MotionPlanRequest the member trajectory_constraints is used by the STOMP planner to provide an initial seed trajectory. I implemented the same functionality in the CHOMP planner. 

If one wants to use an initial trajectory for the CHOMP planner, one should only do : 

```
moveit_msgs::TrajectoryConstraints initial_traj;
/*
Filling of initial_traj
*/
my_group.setTrajectoryConstraints(initial_traj);
```
Before the commit, CHOMP did not care about the trajectory constraints, now it automatically detects if there is one and if yes use it.

### DIFF
--- a/moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_trajectory.h
+++ b/moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_trajectory.h
@@ -130,11 +130,11 @@ public:
    */
   void fillInCubicInterpolation();
 
-   /**
-   * \brief Fill the trajectory from a trajectory seed
-   *
-   * Only modifies points from start_index_ to end_index_, inclusive
-   */
+  /**
+  * \brief Fill the trajectory from a trajectory seed
+  *
+  * Only modifies points from start_index_ to end_index_, inclusive
+  */
   void fillInSeed(const moveit_msgs::MotionPlanRequest& req);
 
   /**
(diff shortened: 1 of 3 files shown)


moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_trajectory.h
```diff
@@ -129,6 +130,13 @@ class ChompTrajectory
    */
   void fillInCubicInterpolation();
 
+  /**
+  * \brief Fill the trajectory from a trajectory seed
```
jrgnicho: It'd be great if it described how the trajectory seed data needs to be packed in the request message.

moveit_planners/chomp/chomp_motion_planner/src/chomp_trajectory.cpp
```diff
@@ -252,4 +253,17 @@ void ChompTrajectory::fillInMinJerk()
   }
 }
 
+void ChompTrajectory::fillInSeed(const moveit_msgs::MotionPlanRequest& req)
```
jrgnicho: I'd prefer that this function checked the request message for valid trajectory constraint data and return the corresponding boolean value. This would make it easier to use the ChompTrajectory class outside of the Chomp planner the as the calling code doesn't have to validate the request message; for instance one could write:
```
if(!trajectory.fillInSeed(req) )
{
  // failed to initialize trajectory from plan request so use a default initialization scheme
  trajectory.fillInLinearInterpolation();
}
```

jrgnicho:
@thibaultbarbie ping

simonschmeisser:
@thibaultbarbie I think this PR is mostly if not completely superseeded by https://github.com/ros-planning/moveit/commit/b84b1bf9f6e7454e8c0351bc241420a8df89317c #1012   however that commit could seriously use some quality control/trivial optimization that you might want to do in a new PR?

v4hn:
@simonschmeisser without looking into this further, your comment suggests this request is not relevant anymore.
Does it make sense to keep this open?

simonschmeisser:
@v4hn I think it can be closed @thibaultbarbie  if you object we can reopen it, but as stated before it has mostly been "merged" already. If you have some time however please test and improve the CHOMP code, it needs some love!

PR #1076, opened 2018-10-04
Documenting why setFromIK in computeCartesianPath does not use random restarts

mlautman:
Documenting why the setFromIK call in computeCartesianPath uses a single attempt rather than the default.  

### Description

### Checklist

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -1919,6 +1919,9 @@ double RobotState::computeCartesianPath(const JointModelGroup* group, std::vecto
   traj.clear();
   traj.push_back(RobotStatePtr(new RobotState(*this)));
 
+  // For each waypoint we do not use random restarts since they would very likely result in IK jumps
+  std::size_t ik_attempts = 1;
+
   double last_valid_percentage = 0.0;
   for (std::size_t i = 1; i <= steps; ++i)
   {
@@ -1927,7 +1930,7 @@ double RobotState::computeCartesianPath(const JointModelGroup* group, std::vecto
     Eigen::Affine3d pose(start_quaternion.slerp(percentage, target_quaternion));
     pose.translation() = percentage * rotated_target.translation() + (1 - percentage) * start_pose.translation();
 
-    if (setFromIK(group, pose, link->getName(), 1, 0.0, validCallback, options))
+    if (setFromIK(group, pose, link->getName(), ik_attempts, 0.0, validCallback, options))
       traj.push_back(RobotStatePtr(new RobotState(*this)));
     else
       break;
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -1919,9 +1919,6 @@ double RobotState::computeCartesianPath(const JointModelGroup* group, std::vecto
   traj.clear();
   traj.push_back(RobotStatePtr(new RobotState(*this)));
 
-  // For each waypoint we do not use random restarts since they would very likely result in IK jumps
-  std::size_t ik_attempts = 1;
-
   double last_valid_percentage = 0.0;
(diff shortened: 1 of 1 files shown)


phy25:
CI failed, requiring to fix clang format: https://travis-ci.org/ros-planning/moveit/jobs/436909712

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1827,6 +1827,8 @@ as the new values that correspond to the group */
   /** \brief This function is only called in debug mode */
   bool checkCollisionTransforms() const;
 
+  static std::string name_;
```
rhaschke: This seems to be superfluos. Either use `RobotState::name_` or this `static const name_` everywhere.
Personally, I would prefer the latter actually.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -53,6 +53,7 @@ namespace core
  * it is difficult to choose a jump_threshold parameter that effectively separates
  * valid paths from paths with large joint space jumps. */
 static const std::size_t MIN_STEPS_FOR_JUMP_THRESH = 10;
+std::string RobotState::name_ = "robot_state";
```
rhaschke: If using this one (which I don't prefer), please use `static const`.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1925,7 +1925,7 @@ double RobotState::computeCartesianPath(const JointModelGroup* group, std::vecto
     Eigen::Affine3d pose(start_quaternion.slerp(percentage, target_quaternion));
     pose.translation() = percentage * rotated_target.translation() + (1 - percentage) * start_pose.translation();
 
-    if (setFromIK(group, pose, link->getName(), 1, 0.0, validCallback, options))
+    if (setFromIK(group, pose, link->getName(), 0, 0.0, validCallback, options))
```
rhaschke: I think, a single attempt is used on purpose here: In a Cartesian path, it doesn't make sense to use a random seed for IK. This essentially enforces non-continuous joint-space motions.

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
mlautman: `name_` has to be static to be used by static methods like `testJointSpaceJump` and static member variables have to be initialized outside the class definition. I don't know of a cleaner way to do this. If you have a cleaner syntax please add your commit on top of mine. 

https://stackoverflow.com/questions/1563897/c-static-constant-string-class-member

moveit_core/robot_state/src/robot_state.cpp
mlautman: This refers to the same static member variable. An easy test to demonstrate this is to make one const and not the other.

mlautman: I have made it const

moveit_core/robot_state/src/robot_state.cpp
mlautman: hrm, good point. I will drop this change, and add this as a comment.

mlautman:
@rhaschke changes made

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1919,6 +1919,9 @@ double RobotState::computeCartesianPath(const JointModelGroup* group, std::vecto
   traj.clear();
   traj.push_back(RobotStatePtr(new RobotState(*this)));
 
+  // For each waypoint we do not use random restarts since they would very likely result in IK jumps
+  std::size_t ik_attempts = 1;
```
rhaschke: I don't see a benefit in introducing a new (constant!) variable here. This even more suggests that the number of IK attempts could be changed. I submitted an alternative comment.

PR #1078, opened 2018-10-06
Check errors when loading scene geometry from text file (#1056)

rhaschke:
Backport of #1056 into Kinetic. Do we want to accept the API change, returning bool instead of void in the functions `loadGeometryFromStream()`?

I don't consider this patch essential and personally would prefer API stability over backporting this.
However, I'm open to other opinions.


jonbinney:
I agree with @rhaschke that this change isn't important enough to change the API in kinetic.

v4hn:
The API change here is negligable in my opinion and is tremendously unlikely to cause any issue.

As the two of you agree not to merge it though, I will just close this request.

PR #1038, opened 2018-08-27
fix various compiler warnings

rhaschke:
This PR attempt to fix most compiler warnings in the current melodic-devel branch - there are too many now. I grouped the commits according to types of compiler warnings:
- unused variables / names
- comparison between signed and unsigned
- fix member initialization (typically order)

Fixing those warnings also revealed some errors in the code ;-)

PR #1089, opened 2018-10-16
fix some more compiler warnings

rhaschke:
These are related to testing code, while #1038 targeted main source code only.


rhaschke:
Merged as trivial.

PR #1092, opened 2018-10-16
fixing initial pose behavior when using the fake_controller_manager

mlautman:
This change handles a bug in the fake_controller_manager where the initial pose message was being missed some large percentage of the time. 

The relevant code was initially added by Ioan and updated by @rhaschke in https://github.com/ros-planning/moveit/commit/e1d91fec2587c7f13533ab3c140600685c87ca52

### DIFF
--- a/moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
+++ b/moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
@@ -68,7 +68,9 @@ public:
       return;
     }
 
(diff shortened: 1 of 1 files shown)


rhaschke:
Could you please explain, how/why latching solves the problem?

mlautman:
According to the docs, latch "If true, the last message published on this topic will be saved and sent to new subscribers when they connect"
http://docs.ros.org/kinetic/api/roscpp/html/classros_1_1NodeHandle.html#a6b655c04f32c4c967d49799ff9312ac6

I believe the issue was that the robot_state_publisher was missing the initial joint state message

rhaschke:
Thanks for clarifying.

PR #1099, opened 2018-10-18
Use locale independent conversion from double to string

simonschmeisser:
boost::lexical_cast<std::string>(0.1) returns "0,1" on Dutch, German and probably some other locales
which boost::lexical_cast<double> fails to parse even on Dutch, German etc locales

so this PR uses std::ostringstream which seems to be the only sane way to ignore locale

(this is no new problem, it somehow seemed to "sometimes" work to put strings (ie '0.1') instead off the raw numbers (ie 0.1) in ompl_planning.yaml but not always ...)


moveit_planners/ompl/ompl_interface/src/ompl_interface.cpp
```diff
@@ -202,13 +203,25 @@ bool ompl_interface::OMPLInterface::loadPlannerConfiguration(
...
     if (it->second.getType() == XmlRpc::XmlRpcValue::TypeString)
+    {
       planner_config.config[it->first] = static_cast<std::string>(it->second);
+    }
     else if (it->second.getType() == XmlRpc::XmlRpcValue::TypeDouble)
-      planner_config.config[it->first] = boost::lexical_cast<std::string>(static_cast<double>(it->second));
+    {
+      // convert to string using no locale
+      std::ostringstream oss;
+      oss.imbue(std::locale());
```
rhaschke: I'm not sure that's really neccessary at all.
If so, I guess you need to use [`std::locale::classic()`](https://en.cppreference.com/w/cpp/locale/locale/classic)

simonschmeisser: Yes, using `std::locale::classic()` is indeed better. Omitting it would probably/usually/maybe also be right unless somebody somewhere set the global locale to something else.

As a side note, using `std::stod` for the inverse operation does not work as this uses the c locale internally which does seem to get set automatically from the system.

rhaschke:
@simonschmeisser Please factor out this conversion code into its own library.

simonschmeisser:
sure, @rhaschke , one thing at its time ;) would you prefer

-  a full featured moveit_tools::lexical_cast  (that uses template specialization for double/float and boost::lexical_cast otherwise) 
- or just a simple moveit_tools::toStringLocaleIndependent(float/double) and use std::to_string otherwise? (would require a moveit_tools::toString(bool) as well)

simonschmeisser:
github sometimes forgets about my further commits, I'm confused

gavanderhoorn:
> 
> 
> github sometimes forgets about my further commits, I'm confused

https://status.github.com/messages

moveit_core/utils/include/moveit/utils/lexical_casts.h
```diff
@@ -0,0 +1,61 @@
...
+
+#ifndef MOVEIT_UTILS_
+#define MOVEIT_UTILS_
+
+#include <string>
+namespace moveit
+{
+namespace utils
+{
+/** \brief Convert a double to std::string without using the system locale
```
rhaschke:
```suggestion
/** \brief Convert a double to std::string using the classic C locale
```

moveit_core/utils/include/moveit/utils/lexical_casts.h
```diff
@@ -0,0 +1,61 @@
...
+
+Depending on the system locale, a different decimal seperator might be used
```
rhaschke: I suggest to move this comment to the top as a general comment for this file.

moveit_core/utils/src/lexical_casts.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "moveit/utils/lexical_casts.h"
+
+#include <locale>
+#include <sstream>
+
+namespace moveit
+{
+namespace utils
+{
+std::string toString(double d)
```
rhaschke: The implementation using function overloading is somewhat redundant.
Why didn't you choose templates as discussed?

simonschmeisser: There are only two variable types (float and double) where these functions are required. For other variable types (integer types) std::to_string would be a more obvious solution? This can of course be handled using template specialization if requested but imho just adds compile time and unneeded abstractions.

simonschmeisser:
about string->float/double: from a bug fixing point of view it's not necessary since boost::lexical_cast<double>(string) expects "." independent of locale (there is not a single line about locales in the documentation of boost::lexical_cast so ...)

from a maintenance point of view (as boost::lexical_cast people do not seem to know what a locale is ...) it might be better to indeed add this inverse function

simonschmeisser:
There is only one more places where the conversion double -> string is done incorrectly:
https://github.com/ros-planning/moveit/blob/5a1d5a8f172c43aedee9f47926068e3ac529d312/moveit_ros/benchmarks/benchmarks/src/benchmark_execution.cpp#L812

should I include it in this PR?

rhaschke:
> There is only one more place where the conversion double -> string is done incorrectly. Should I include it in this PR?

Yes, please.

moveit_core/utils/src/lexical_casts.cpp
rhaschke: @simonschmeisser I always try to prevent code duplication as much as possible - even in this simple case. In this particular case, I would suggest to use a template defined in .cpp only, providing the common implementation for both overloaded functions. This doesn't add compile time or any other complexity to the API.

Please, use github's `Reply` button to answer such remarks. This way, the thread is kept together.

moveit_planners/ompl/ompl_interface/src/ompl_interface.cpp
```diff
@@ -202,13 +204,22 @@ bool ompl_interface::OMPLInterface::loadPlannerConfiguration(
   for (XmlRpc::XmlRpcValue::iterator it = xml_config.begin(); it != xml_config.end(); ++it)
   {
     if (it->second.getType() == XmlRpc::XmlRpcValue::TypeString)
+    {
```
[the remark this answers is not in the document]
simonschmeisser: well, we internally use a coding style were they don't cost any extra lines {
} and then it's just a convenient safety thing to always have them

also before refactoring the new entry used to be multi-line so adding parantheses to only one clause seemed "unbalanced" ... can revert it now actually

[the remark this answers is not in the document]
simonschmeisser: Done :)

moveit_core/CMakeLists.txt
```diff
@@ -114,6 +115,7 @@ catkin_package(
     moveit_distance_field
     moveit_kinematics_metrics
     moveit_dynamics_solver
+    moveit_utils
```
davetcoleman: this is not a package, remove

moveit_core/utils/include/moveit/utils/lexical_casts.h
```diff
@@ -0,0 +1,72 @@
...
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Simon Schmeisser */
+
+#ifndef MOVEIT_UTILS_
+#define MOVEIT_UTILS_
```
davetcoleman: add MOVEIT_CORE_ to front

moveit_core/utils/include/moveit/utils/lexical_casts.h
```diff
@@ -0,0 +1,72 @@
...
+ *  Depending on the system locale, a different decimal seperator might be used
+ *  for floating point numbers. This is often not wanted for internal (ie non-user
+ *  facing) purposes. This module provides conversion functions that use std::locale::classic()
+ *  (i.e. the default if no locale is set on the system).
+ */
+
+#include <string>
+namespace moveit
+{
+namespace utils
```
davetcoleman: this should have ``namespace core`` also, or instead of ``utils``, as used in all the other subfolders of moveit_core

moveit_core/utils/include/moveit/utils/lexical_casts.h
```diff
@@ -0,0 +1,72 @@
...
+{
+/** \brief Convert a double to std::string using the classic C locale */
+std::string toString(double d);
+
+/** \brief Convert a float to std::string using the classic C locale */
+std::string toString(float d);
```
davetcoleman:
```suggestion
std::string toString(float f);
```
var name makes more sense

moveit_core/utils/src/lexical_casts.cpp
```diff
@@ -0,0 +1,91 @@
...
+  return toStringImpl(d);
+}
+
+template <class T>
+T toRealImpl(const std::string& s)
+{
+  // convert from string using no locale
+  std::istringstream iss(s);
+  iss.imbue(std::locale::classic());
+  T d;
```
davetcoleman: there are too many one character variable names here, hard to read. add longer variable names

moveit_core/utils/src/lexical_casts.cpp
```diff
@@ -0,0 +1,91 @@
...
+{
+  return toStringImpl(d);
+}
+
+std::string toString(float d)
+{
+  return toStringImpl(d);
+}
+
+template <class T>
```
davetcoleman: I think T would be clearer if called something like ``OutputDataType``

rhaschke: This is nit-picking. T is the standard for a template type. Only if more than one type is involved, there is a need in more verbose type names.

davetcoleman: fair. just pulling in recent experiences in google-style projects, but ignore my comment

moveit_core/CMakeLists.txt
simonschmeisser: this section is about what libraries are exported by the moveit_core package

rhaschke:
LGTM

moveit_core/CMakeLists.txt
davetcoleman: oops sorry. the github code folding hid the relevant section and i made a wrong assumption

davetcoleman:
Melodic: https://github.com/ros-planning/moveit/pull/1155

PR #1104, opened 2018-10-20
WIP: fix memory leaks / ClassLoader SEVERE warnings

rhaschke:
When move_group node finishes, there are always warnings like:
```
[ WARN] ros.rosconsole_bridge.class_loader.ClassLoader: SEVERE WARNING!!!
Attempting to unload library while objects created by this loader exist in the heap!
You should delete your objects before attempting to unload the library or destroying the ClassLoader.
The library will NOT be unloaded.
```
This is a hint that managed plugin instances are not correctly released before the corresponding plugin loader is. Even worse, MoveIt still used a lot of pointers created unmanaged, where we wouldn't even notice such issues. For this reason, use of pluginlib's createUnmanagedInstance() is strongly discouraged:
http://wiki.ros.org/class_loader#Understanding_Loading_and_Unloading

This PR is to be considered as an open branch to contribute to fixing these issues. I started by replacing
`createUnmanagedInstance()` with `createUniqueInstance()`, which is available for C++11.

I tracked down one issue already: The PlanningScenePtr passed to the `ompl_simple_setup_` via `ob::GoalPtr` in [`ModelBasedPlanningContext::setGoalConstraints()`](https://github.com/ros-planning/moveit/blob/melodic-devel/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp#L499) was never released. This is an issue in OMPL.

As a quick fix, I changed the `ConstraintSampler` to hold a raw `PlanningScene*` instead of a `shared_ptr`.
This solved the memory leak: the `PlanningScene` instance is correctly released. 
However, it's not a clean solution: There is no chance to validate the pointer before use. I suggest to use a `weak_ptr` instead, making `scene_` private and only accessible via `getPlanningScene()`, which in turn could validate the weak pointer and issue an error msg (before eventually crashing).


rhaschke:
Next issue for me is that the RobotModelPtr that was passed to the OMPL plugin, is not fully released.

v4hn:
Do you propose to leave this pull-request open?

I would say it makes more sense to merge all the problems we find but have one issue to refer to...

rhaschke:
Yes, I intended this PR as open to all contributors willing to tackle some of the mentioned issues ;-)

rhaschke:
Related to https://github.com/ros-planning/moveit/pull/150#issuecomment-254240560, https://github.com/ros-planning/moveit/pull/150#issuecomment-254103813.

rhaschke:
Fixed two other bugs:
- KinematicsPluginLoader kept all it's cache instances
- Planner plugins might not correctly release the passed RobotModel

moveit_ros/planning/kinematics_plugin_loader/src/kinematics_plugin_loader.cpp
```diff
@@ -239,8 +233,9 @@ class KinematicsPluginLoader::KinematicsLoaderImpl
   std::map<std::string, std::vector<std::string> > iksolver_to_tip_links_;  // a map between each ik solver and a vector
                                                                             // of custom-specified tip link(s)
   std::shared_ptr<pluginlib::ClassLoader<kinematics::KinematicsBase> > kinematics_loader_;
-  std::map<const robot_model::JointModelGroup*, std::vector<kinematics::KinematicsBasePtr> > instances_;
+  std::map<const robot_model::JointModelGroup*, kinematics::KinematicsBasePtr> instances_;
```
simonschmeisser: I had the impression that multiple plugins can be loaded for the same JointModelGroup* (for whatever reason). This seems to change that. Is that intentional?

rhaschke: Sorry, I only explained the reasoning for this in https://github.com/ros-planning/moveit/pull/150#issuecomment-433106660

The KinematicsPluginLoader caches created kinematics solver (KS) instances. But the cache only maintains **unique** instances - they are not shared. The only purpose of the cache is to not allocate a KS twice in a row - once for validating that it is suitable for JMG [here](https://github.com/ros-planning/moveit/blob/a46f0fb654dee077bcf39c1a2c48f4dffe1068f4/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp#L203) and once for actual use [here](https://github.com/ros-planning/moveit/blob/a46f0fb654dee077bcf39c1a2c48f4dffe1068f4/moveit_core/robot_model/src/joint_model_group.cpp#L604). 
Eventually, the JMG owns the created KS instance. KPL only caches it for short term.
As I got the very same impression as @simonschmeisser in the first place, I decided to emphasize the fact that this is only short-term cache by changing the data structure - only the last created KS instance is held per JMG (and passed on to the caller in the 2nd call in a row).
This also resolves a potential issue with dangling shared_ptrs in the cache (which usually should only be weak_ptrs for this reason).

rhaschke: Just as a side note: Indeed each JMG owns its unique solver instance. This helps to avoid race conditions when using them from multiple threads in parallel.

rhaschke:
Tracked down next bug: `MoveGroupInterface` doesn't release `RobotModel` passed from rviz motion planning plugin.

rhaschke:
@v4hn From my point of view, this is now ready to merge. Together with #1166 this fixes all class-loader warnings.

rhaschke:
Rebased to resolve conflicts.

rhaschke:
Ping @v4hn. I would like to merge this into the next release.

v4hn:
gimme a break, I spent all day looking through requests.

I will try to address more open requests tomorrow.

rhaschke:
@v4hn Ping.

rhaschke:
@v4hn I just found another improvement. Working on it later. Please do not yet merge immediately.

rhaschke:
Cleaned up. Ready for final review.

rhaschke:
Turned out, that the main bug for classloader issues was a circular reference in the OMPL wrapper: https://github.com/ros-planning/moveit/pull/1104/commits/ed3f4b4be29856d8705bdf40a4a2f6245234c523

v4hn:
Sorry for not looking through this in time. Post-merge the changes seem valid and I would have merged.

At the same time I really don't think it's a good idea to merge architecture changes in memory management (changing shared for weak ptrs, etc) an hour before a release.

rhaschke:
> I don't think it's a good idea to merge architecture changes in memory management just before a release.

These changes are more than a month old and I'm running this code in production since then. I wasn't afraid at all ;-)

v4hn:
> These changes are more than a month old and I'm running this code in production since then.

Yes, you run it *in your* environment.
The reason for our independent-second-approval-scheme is that environments&setups vary *a lot* between users (and maintainers). Shadow-fixed reaches more users to test changes, but a first independent test/review still makes sense to avoid hotfix commits/releases.

In this case I agree post-merge, so we would have merged (& released) it either way..

rhaschke:
> Yes, you run it in your environment.

Touché.

PR #1107, opened 2018-10-21
fix typo getCollisionBodyTransform(s)

rhaschke:
This method only returns a single transform, hence a singular form name should be used.
Interestingly this method was never used inside MoveIt.


v4hn:
I don't really see the need but fine by me as long as this is not backported to kinetic.

Please add the change to MIGRATION.md before merging.

MIGRATION.md
```diff
@@ -7,6 +7,7 @@ API changes in MoveIt! releases
 - Migration to ``tf2`` API.
 - The move_group capability ``ExecuteTrajectoryServiceCapability`` has been removed in favor of the improved ``ExecuteTrajectoryActionCapability`` capability. Since Indigo, both capabilities were supported. If you still load default capabilities in your ``config/launch/move_group.launch``, you can just remove them from the capabilities parameter. The correct default capabilities will be loaded automatically.
 - Deprecated method ``CurrentStateMonitor::waitForCurrentState(double wait_time)`` was finally removed.
+- Renamed ``RobotState::getCollisionBodyTransforms`` to ``getCollisionBodyTransforms`` as it returns a single transform only.
```
gavanderhoorn:
```suggestion
- Renamed ``RobotState::getCollisionBodyTransforms`` to ``getCollisionBodyTransform`` as it returns a single transform only.
```

PR #1108, opened 2018-10-21
World::moveObject(): only apply changes if there is any non-zero motion

rhaschke:
This is a follow up to #957, which introduced World::moveObject().
When transform is the identity matrix, i.e. there is no movement, the object shouldn't be touched.
This follows the rule, only to create a copy of the parent planning scene, when there are actual modifications.

- [ ] Cherry pick to Kinetic


mlautman:
Why return false? Shouldn't it return true because the object is at the desired location?

PR #1129, opened 2018-10-24
remove dependency on moveit_experimental from chomp packages

rhaschke:
A local test build, removing moveit_experimental, approved that this is actually not needed.


rhaschke:
Hm. My local build found the required includes somewhere else. This doesn't work.

130s:
I also wondered if chomp really depends on `moveit_experimental` or not. Thanks so much @rhaschke for confirming that it is still needed.

PR #1133, opened 2018-10-25
Add coverage analysis for moveit_core

agutenkunst:
Originated from: https://github.com/ros-planning/moveit/issues/1

Provides a CMake target to generate a coverage analysis of moveit_core.

Usage: `catkin_make -DENABLE_COVERAGE_TESTING=ON moveit_core_coverage`

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -138,6 +138,18 @@ catkin_package(
     urdfdom_headers
     )
 
+
+# to run: catkin_make -DENABLE_COVERAGE_TESTING=ON package_name_coverage
+if(ENABLE_COVERAGE_TESTING)
+  find_package(code_coverage REQUIRED)   # catkin package ros-*-code-coverage
+  include(CodeCoverage)
+  APPEND_COVERAGE_COMPILER_FLAGS()
+  set(COVERAGE_EXCLUDES "*/test/*")
+  add_code_coverage(
+    NAME ${PROJECT_NAME}_coverage
+  )
+endif()
+
 include_directories(SYSTEM ${EIGEN3_INCLUDE_DIRS}
                            ${LIBFCL_INCLUDE_DIRS}
                            )


moveit_core/package.xml
```diff
@@ -76,4 +76,5 @@
   <test_depend>orocos_kdl</test_depend>
   <test_depend>rosunit</test_depend>
   <test_depend>tf_conversions</test_depend>
+  <test_depend>code_coverage</test_depend>
```
rhaschke: Coverage testing is not enabled for testing by default, is it?
If so, we can drop this dependency.

agutenkunst:
For context: We are currently at the world moveit day

Talked to @v4hn we agreed that this may not be exactly what was hoped for and misses on the nice to have badges. However maybe even for a short time this would enable to generation of some kind of coverage until some coveralls is made.

Results are stored locally after generation, really not perfect but better than nothing.

mikeferguson:
@davetcoleman the coverage report ends up locally in the workspace (it looks like this is using my code_coverage package -- which is still somewhat new). I'd be happy to accept changes to code_coverage if someone comes up with a good way to get badges/results out easily.

davetcoleman:
>   Could not find a package configuration file provided by "code_coverage" with any of the following names:

Need to add to package.xml

rhaschke:
> > Could not find a package configuration file provided by "code_coverage" with any of the following names:

> Need to add to package.xml

I explicitly suggested not to depend on this by default: https://github.com/ros-planning/moveit/pull/1133#discussion_r228144807. Having this in package.xml only ensures that rosdep will install it. It's not needed for the normal build or test process. I suggest to install this package by default in MoveIt's docker containers.

davetcoleman:
Ok, if we're not going to add it to package.xml please add a new short section at the bottom of this tutorial explaining how to use this experimental feature (including what to install first)

davetcoleman:
Results: 51.2 % of our functions have tests in moveit_core

image

agutenkunst:
@davetcoleman you mean at https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html somewhere?

davetcoleman:
Yes

davetcoleman:
ping re:tutorial :-)

otherwise this feature will likely never be found again as its pretty esoteric

davetcoleman:
Ooops indeed! Humans are faulty like that, hence my recommendation here: https://github.com/ros-planning/moveit/issues/1284

;-)

davetcoleman:
cherry picked to melodic

PR #1152, opened 2018-10-26
Attached object visualization fix

svenkrause:
### Description

When deselecting "Visual Enabled"/"Collision Enabled" in
RobotStateDisplay or "Show Robot Visual"/"Show Robot Collision" in
PlanningSceneDisplay, the attached object was still shown.

Also, there was no way to not display the attached object individually.

--> added new BoolProperty "Attached Body Visible" / "Show Attached
Body" to both Displays and a corresponding field in the
RobotStateVisualization-class, which is then checked in its update-
function.

--> also, now require an update_state_ (RobotStateDisplay) /
planning_scene_needs_render_ (PlanningSceneDisplay), when either of the
three flags (Robot Visual / Robot Collision / Attached Body Visible) is
changed.

### DIFF
--- a/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
+++ b/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
@@ -95,9 +95,9 @@ RobotStateDisplay::RobotStateDisplay() : Display(), update_state_(false), load_r
                                                      "Whether to display the collision representation of the robot.",
                                                      this, SLOT(changedEnableCollisionVisible()), this);
 
-  enable_attached_body_visible_ = new rviz::BoolProperty("Attached Body Visible", true,
-                                                         "Whether to display the attached body.",
-                                                         this, SLOT(changedEnableAttachedBodyVisible()), this);
+  enable_attached_body_visible_ =
+      new rviz::BoolProperty("Attached Body Visible", true, "Whether to display the attached body.", this,
+                             SLOT(changedEnableAttachedBodyVisible()), this);
 
   show_all_links_ = new rviz::BoolProperty("Show All Links", true, "Toggle all links visibility on or off.", this,
                                            SLOT(changedAllLinks()), this);
--- a/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
+++ b/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
@@ -95,6 +95,10 @@ RobotStateDisplay::RobotStateDisplay() : Display(), update_state_(false), load_r
                                                      "Whether to display the collision representation of the robot.",
                                                      this, SLOT(changedEnableCollisionVisible()), this);
 
+  enable_attached_body_visible_ = new rviz::BoolProperty("Attached Body Visible", true,
+                                                         "Whether to display the attached body.",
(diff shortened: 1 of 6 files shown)


simonschmeisser:
This fixes #512

moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
```diff
@@ -95,6 +95,10 @@ RobotStateDisplay::RobotStateDisplay() : Display(), update_state_(false), load_r
                                                      "Whether to display the collision representation of the robot.",
                                                      this, SLOT(changedEnableCollisionVisible()), this);
 
+  enable_attached_body_visible_ =
+      new rviz::BoolProperty("Attached Body Visible", true, "Whether to display the attached body.", this,
```
simonschmeisser: For consistency the label should be "Attached Body Enabled". Or all three renamed to "Show ...". Not sure if renaming them would break config files?

svenkrause: Yes, "Enabled" is more consistent, I just found it less descriptive. But I guess, it would be better to split the re-naming into a separate PR and use "Enabled" in this one.

simonschmeisser:
We discussed this for some while with @v4hn during #wmd. Either way of **a)** hiding attached objects once both visual and collision are hidden or **b)** separating attached objects as a third option are valid and have their pros and cons. The middle way as you propose is not supported by rviz I fear as you can only hide but not disable a `rviz::Property`. Changing the actual value of the checkbox/Property would be complicated/messy as then we would need to track the intended value of `attached_body_visible_` to reset it after re-enabling either visual or collision.

Splitting this into two separate commits would require calling `render_shapes_->clear();` somewhere as well. So the first commit would implement strategy **a)** which shows behavior different from the current state (AOs disappear automatically) while the second commit would change the behavior again (hide AOs manually). Not sure if that is desired.

rhaschke:
> Splitting this into two separate commits would require calling `render_shapes_->clear()`

That's not true. I splitted  the commits as suggested by me, and it perfectly works for the PlanningSceneDisplay. I will close this PR and re-open as a new one with the commits splitted.

> The middle way as you propose is not supported by rviz I fear as you can only hide but not disable a rviz::Property

That's not true either, though it requires some work. First, there is `Property::setReadOnly(true)`. Unfortunately, this doesn't disable the property. If you want to disable as well, you need to inherit and override [`Property::getViewFlags()`](http://docs.ros.org/jade/api/rviz/html/c++/classrviz_1_1Property.html#a16f2af2323d410a9a92691577ba10eb9).

simonschmeisser:
> >  The middle way as you propose is not supported by rviz I fear as you can only hide but not disable a rviz::Property
>
> That's not true either, though it requires some work. First, there is Property::setReadOnly(true). Unfortunately, this doesn't disable the property. If you want to disable as well, you need to inherit and override Property::getViewFlags().

I should have written "within reasonable limits". As I said before, having attached objects separated does make sense as well and since it can be achieved with way less code is preferable IMHO.

rhaschke:
Sorry, but I don't see a benefit 
* in hiding attached bodies when the robot is shown. This puts you at risk to not see/understand why a collision occurred.
* in showing the attached bodies when the robot is not shown. As said before, the attached body is part of the robot and should be hidden with it.

PR #1153, opened 2018-10-26
[WIP] Pass RobotModel to ik plugins

simonschmeisser:
This is a proof of concept for further discussion of a way to pass `RobotModel(Ptr)` to kinematics plugins. It is inspired by #150 and the recent and not so recent discussion therein

The main objective of this (and previous PRs (and previous 4y old PRs)) is to
1) save some parsing and loading time by reusing the `RobotModel`
2) use consistent joint limits in all parts of the code

Passing in `std::shared_ptr<RobotModel>` (aka `RobotModelPtr`) to `kinematics::KinematicsBase` derived plugins and storing it there or in a member variable `RobotState` will however create a cyclic shared pointer loop as
`RobotModel `-> `JointModelGroup` -> `KinematicsSolver`/`KinematicsBase` (plugin) -> `RobotModel`

This was discussed at length yesterday at  https://github.com/ros-planning/moveit/pull/150#issuecomment-433045509 and following

The solution sketched here is to turn the shared pointer from `JointModelGroup` -> `KinematicsSolver` into a weak pointer and have them owned by `RobotModelLoader` in its `KinematicPluginLoader`. `KinematicPluginLoader` already has the `KinematicSolvers` in a cache so changes there would only be needed for clarity. We could also move/copy the `KinematicSolvers` to a more explicit "storage" directly in `RobotModelLoader` (or elsewhere outside of `RobotModel`)

Further points open for discussion:
This PR adds a member variable of type  `RobotModelPtr` to `kinematics::KinematicsBase` for simple access. This breaks ABI. (but not API!) Still it means that e.g. ikFast plugins from ros-I packages need to be recompiled to be able to be loaded. This could be avoided by having a `virtual setRobotModel(RobotModelPtr)` be called right after the constructor and not storing the pointer in `KinematicsBase`. Plugins could then override this function to store the pointer themselves, thus preserving ABI as well.

This PR is made against kinetic-devel, mostly for convenience. Since it can change behavior it should probably only (eventually) be applied against melodic. I'm sure this can be handled by git ninjas

If during the discussion slight hints of approval emerge, I would port further plugins as well during this slightly extended World MoveIt! Week. They will continue working as is however for the time being.

### DIFF
--- a/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -50,7 +50,9 @@ namespace moveit
 namespace core
 {
 class JointModelGroup;
+class RobotModel;
 class RobotState;
+MOVEIT_CLASS_FORWARD(RobotModel);
 }
 }
 
@@ -583,7 +585,19 @@ public:
     supported_methods_.push_back(DiscretizationMethods::NO_DISCRETIZATION);
   }
 
+  void setRobotModel(const moveit::core::RobotModelPtr& rm)
+  {
+    robot_model_ = rm;
+  }
+
+  moveit::core::RobotModelPtr getRobotModel() const
+  {
+    return robot_model_;
+  }
+
+
 protected:
+  moveit::core::RobotModelPtr robot_model_;
   std::string robot_description_;
   std::string group_name_;
   std::string base_frame_;
--- a/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
@@ -78,13 +78,12 @@ public:
     /// Return a flag indicating whether the state of the solver is initialized
     operator bool() const
     {
-      return allocator_ && !bijection_.empty() && solver_instance_;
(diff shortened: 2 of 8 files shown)


simonschmeisser:
@rhaschke Could we go back to closing PRs once discussion is finished?

simonschmeisser:
With this PR ownership of KinematicSolvers (ik plugins) is transfered to RobotModelLoader. So once RobotModelLoader gets released it will release the KinematicsSolvers as well as their use count is now 0. RobotModel only has a weak pointer on the KinematicSolvers which does not increase the use counter. So step 1 is done, ik plugin unloaded/released.

Now during the release of the IK plugin it releases all it's shared pointers to RobotModel (as well as all RobotStates that have shared pointers to RobotModel). If nobody else (outside the kinematics code) owns further shared_pointers of RobotModel it's use count now reaches zero as well and it gets released.

This PR demonstrates a way to achieve the goal of having a shared RobotModel available in plugins and does not add new shared pointer loops. That those exist elsewhere is an unrelated problem.

rhaschke:
@simonschmeisser I filed #1166 as an alternative approach. Please have a look.

rhaschke:
Closing in favour of merged #1166.

PR #1156, opened 2018-10-26
add support for fcl 0.6

2scholz:
The (still unreleased) fcl 0.6 changed paths to relevant headers as well as the name of many types.
This patch adds support for fcl 0.6 without breaking support for fcl 0.5.

This patch adds the header fcl_compat.h to check for the fcl version and include fcl headers accordingly.
The old types are exchanged to the newer ones throughout the package.
Compability to fcl 0.5 is ensured by creating type aliases that map the old types to the new ones.

The other way around might be the better choice for now, but the old type names are still in use in the new versions (as templates), so the alias trick only works this way around.

For this a lot of code from @Levi-Armstrong was used, who already created a branch with support for fcl 0.6 but without support for fcl 0.5: https://github.com/Levi-Armstrong/moveit/tree/newFCL

@v4hn

### DIFF
--- a/moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_common.h
+++ b/moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_common.h
@@ -40,9 +40,18 @@
 #include <moveit/collision_detection/world.h>
 #include <moveit/collision_detection/collision_world.h>
 #include <moveit/macros/class_forward.h>
+#include <moveit/collision_detection_fcl/fcl_compat.h>
+
+#if (MOVEIT_FCL_VERSION >= FCL_VERSION_CHECK(0, 6, 0))
+#include <fcl/broadphase/broadphase_collision_manager.h>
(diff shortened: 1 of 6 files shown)


BryceStevenWilley:
A third option, besides aliasing old types to new or new types to old types, is to using typedefs that are not in the `fcl` namespace. This is the approach taken by OMPLapp (https://github.com/ompl/omplapp/commit/b9868385d5540e8335530dda62dea02acaf01e70#diff-524e09a932cb1946ad510acbcd23bec5R56).  This does make it more difficult to keep up with additions to FCL itself however.  Both require lots of small changes in the codebase as well. 

As long as we agree on how to handle the aliases, this is a great contribution. . :+1:

rhaschke:
I agree to merge this before an official FCL 0.6 release. Backwards-compatibility is ensured ;-)
Please rebase.

2scholz:
I made the requested changes.
`FCL_VERSION` is already defined in `fcl/config.h`, so I named the macro `MOVEIT_FCL_VERSION` instead.

PR #1135, opened 2018-10-25
Fix values in PlanningFrame not beeing stored

agutenkunst:
### Description
Multiple values from the PlanningFrame (see Image) are not stored in the config.
Fixed by this PR. Still missing is the valueChanged notification to rviz. Will open seperate PR.

PR #1160, opened 2018-10-26
cherry picking from Kinetic

rhaschke:
* #1135 
* #1099 

Please use rebase-commit to merge!


moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -45,6 +45,8 @@
 #include <moveit/ompl_interface/constraints_library.h>
 #include <moveit/kinematic_constraints/utils.h>
 #include <moveit/profiler/profiler.h>
+#include <moveit/utils/lexical_casts.h>
+#include <eigen_conversions/eigen_msg.h>
```
IanTheEngineer: This extra include introduces a regression in `melodic-devel`: 
http://build.ros.org/view/Mbin_uB64/job/Mbin_uB64__moveit_planners_ompl__ubuntu_bionic_amd64__binary/17/console
We removed the `eigen_conversions` tf1 dependency in #830

rhaschke: Argh! Can you device a fix?

IanTheEngineer: Yep! This is a spurious include. It's as simple as removing that line.

PR #1166, opened 2018-10-28
Pass RobotModel to kinematics plugins

rhaschke:
This is an alternative implemenation of #150 and #1153 to directly pass a RobotModel (RM) to kinematics plugins, thus
1) saving some loading time by reusing the `RobotModel` (particularly if there are dozens of groups)
2) ensuring a consistent RM (particularly, joint limits) is used throughout the code

#150 eventually got stuck due to circular reference of `RobotModelPtr` as detailed in https://github.com/ros-planning/moveit/pull/150#issuecomment-433106660.
To avoid/break this circular reference #1153 transferred ownership of KinematicSolvers (KS) from JointModelGroup (JMG) to KinematicsPluginLoader (KPL) / RobotModelLoader (RML). While this works, I think this is an evil hack. Ownership definitely belongs to JMG. This already becomes obvious from the fact that a RM with JMGs can (theoretically) exist without a KPL.

In this PR, ownership stays with JMGs. To break the circular reference, the RM is passed as reference to the KS. There, in contrast to #150, a new shared_ptr is created from the raw pointer, thus performing its own use counting. This way we decouple the use of the RM within the external KS from MoveIt. MoveIt just ensures that the RM is alive as long as the KS is. The KinematicsBase destructor could even check that the decoupled shared_ptr was correctly released by KS (and issue a warning if not).

To allow a smooth migration path, KS implementations should be able to choose which API for `initialize()` to implement. This is not (yet) considered in #1153. Here, KPL first tries the new API and - on failure - the old API. KinematicsBase provides default implementations for them, returning false, i.e. indicating failure.

Adaptation of CachedIK was a little bit tricky: This one wraps an arbitrary other KS. To relay `initialize()` calls of the wrapper to the available implementation(s) of the wrapped KS, I use SFINAE magic. This works for KDL IK, which already has the old API removed.

**Please, keep commit history when merging!**

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -9,7 +9,10 @@ API changes in MoveIt! releases
 - Deprecated method ``CurrentStateMonitor::waitForCurrentState(double wait_time)`` was finally removed.
 - Renamed ``RobotState::getCollisionBodyTransforms`` to ``getCollisionBodyTransform`` as it returns a single transform only.
 - Removed deprecated class MoveGroup (was renamed to MoveGroupInterface).
-- KinematicsBase: Deprecate members tip_frame_, search_discretization_. Use tip_frames_ and redundant_joint_discretization_ instead.
+- KinematicsBase: Deprecated members `tip_frame_`, `search_discretization_`.
+  Use `tip_frames_` and `redundant_joint_discretization_` instead.
+- KinematicsBase: Deprecated `initialize(robot_description, ...)` in favour of `initialize(robot_model, ...)`.
+  Adapt your kinematics plugin to directly receive a `RobotModel`. See the [KDL plugin](https://github.com/ros-planning/moveit/tree/melodic-devel/moveit_kinematics/kdl_kinematics_plugin) for an example.
 
 ## ROS Kinetic
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -8,7 +8,8 @@ API changes in MoveIt! releases
 - The move_group capability ``ExecuteTrajectoryServiceCapability`` has been removed in favor of the improved ``ExecuteTrajectoryActionCapability`` capability. Since Indigo, both capabilities were supported. If you still load default capabilities in your ``config/launch/move_group.launch``, you can just remove them from the capabilities parameter. The correct default capabilities will be loaded automatically.
 - Deprecated method ``CurrentStateMonitor::waitForCurrentState(double wait_time)`` was finally removed.
 - Renamed ``RobotState::getCollisionBodyTransforms`` to ``getCollisionBodyTransform`` as it returns a single transform only.
-- Remove deprecated class MoveGroup (was renamed to MoveGroupInterface).
+- Removed deprecated class MoveGroup (was renamed to MoveGroupInterface).
+- KinematicsBase: Deprecate members tip_frame_, search_discretization_. Use tip_frames_ and redundant_joint_discretization_ instead.
 
 ## ROS Kinetic
 
--- a/moveit_core/kinematics_base/CMakeLists.txt
+++ b/moveit_core/kinematics_base/CMakeLists.txt
@@ -2,6 +2,9 @@ set(MOVEIT_LIB_NAME moveit_kinematics_base)
 
 add_library(${MOVEIT_LIB_NAME} src/kinematics_base.cpp)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION ${${PROJECT_NAME}_VERSION})
+# Avoid warnings about deprecated members of KinematicsBase when building KinematicsBase itself.
+# TODO: remove when deperecated variables (tip_frame_, search_discretization_) are finally removed.
+target_compile_options(${MOVEIT_LIB_NAME} PRIVATE -Wno-deprecated-declarations)
 
 # This line is needed to ensure that messages are done being built before this is built
 add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
--- a/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -40,6 +40,7 @@
 #include <geometry_msgs/PoseStamped.h>
 #include <moveit_msgs/MoveItErrorCodes.h>
 #include <moveit/macros/class_forward.h>
+#include <moveit/macros/deprecation.h>
 #include <ros/node_handle.h>
 
 #include <boost/function.hpp>
@@ -49,8 +50,9 @@ namespace moveit
 {
 namespace core
 {
-class JointModelGroup;
-class RobotState;
+MOVEIT_CLASS_FORWARD(JointModelGroup)
+MOVEIT_CLASS_FORWARD(RobotState)
+MOVEIT_CLASS_FORWARD(RobotModel)
 }
 }
 
@@ -168,7 +170,7 @@ public:
                 const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions()) const = 0;
 
   /**
-   * @brief Given a desired pose of the end-effector, compute the set joint angles solutions that are able to reach it.
+   * @brief Given the desired poses of all end-effectors, compute joint angles that are able to reach it.
    *
    * This is a default implementation that returns only one solution and so its result is equivalent to calling
    * 'getPositionIK(...)' with a zero initialized seed.
@@ -366,7 +368,10 @@ public:
    * @param tip_frame The tip of the chain
    * @param search_discretization The discretization of the search when the solver steps through the redundancy
    * @return True if initialization was successful, false otherwise
+   *
+   * Instead of this method, use the method passing in a RobotModel.
    */
+  MOVEIT_DEPRECATED
   virtual bool initialize(const std::string& robot_description, const std::string& group_name,
                           const std::string& base_frame, const std::string& tip_frame,
                           double search_discretization) = 0;
@@ -381,19 +386,30 @@ public:
    * @param tip_frames A vector of tips of the kinematic tree
    * @param search_discretization The discretization of the search when the solver steps through the redundancy
    * @return True if initialization was successful, false otherwise
+   *
+   * Instead of this method, use the method passing in a RobotModel.
    */
+  MOVEIT_DEPRECATED
   virtual bool initialize(const std::string& robot_description, const std::string& group_name,
+                          const std::string& base_frame, const std::vector<std::string>& tip_frames,
+                          double search_discretization);
+
+  /**
+   * @brief  Initialization function for the kinematics, for use with kinematic chain IK solvers
+   * @param robot_model - allow the URDF to be loaded much quicker by passing in a pre-parsed model of the robot
+   * @param group_name The group for which this solver is being configured
+   * @param base_frame The base frame in which all input poses are expected.
+   * This may (or may not) be the root frame of the chain that the solver operates on
+   * @param tip_frames The tip of the chain
+   * @param search_discretization The discretization of the search when the solver steps through the redundancy
+   * @return true if initialization was successful, false otherwise
+   */
+  virtual bool initialize(const moveit::core::RobotModel& robot_model, const std::string& group_name,
                           const std::string& base_frame, const std::vector<std::string>& tip_frames,
                           double search_discretization)
   {
-    // For IK solvers that do not support multiple tip frames, fall back to single pose call
-    if (tip_frames.size() == 1)
-    {
-      return initialize(robot_description, group_name, base_frame, tip_frames[0], search_discretization);
-    }
-
-    ROS_ERROR_NAMED("kinematics_base", "This kinematic solver does not support initialization "
-                                       "with more than one tip frames");
+    // For IK solvers that do not support passing in a RobotModel, return false and have
+    // the kinematics_plugin_loader try the older version of loading
     return false;
   }
 
@@ -446,10 +462,9 @@ public:
    * @brief Set a set of redundant joints for the kinematics solver to use.
    * This can fail, depending on the IK solver and choice of redundant joints!. Also, it sets
    * the discretization values for each redundant joint to a default value.
-   * @param redundant_joint_indices The set of redundant joint indices (corresponding to
-   * the list of joints you get from getJointNames()).
-   * @return False if any of the input joint indices are invalid (exceed number of
-   * joints)
+   * @param redundant_joint_indices The set of redundant joint indices
+   *        (corresponding to the list of joints you get from getJointNames()).
+   * @return False if any of the input joint indices are invalid (exceed number of joints)
    */
   virtual bool setRedundantJoints(const std::vector<unsigned int>& redundant_joint_indices);
 
@@ -457,8 +472,7 @@ public:
    * @brief Set a set of redundant joints for the kinematics solver to use.
    * This function is just a convenience function that calls the previous definition of setRedundantJoints()
    * @param redundant_joint_names The set of redundant joint names.
-   * @return False if any of the input joint indices are invalid (exceed number of
-   * joints)
+   * @return False if any of the input joint indices are invalid (exceed number of joints)
    */
   bool setRedundantJoints(const std::vector<std::string>& redundant_joint_names);
 
--- a/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
+++ b/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
@@ -371,11 +371,12 @@ public:
    * @param search_discretization The discretization of the search when the solver steps through the redundancy
    * @return True if initialization was successful, false otherwise
    *
-   * Instead of this method, use the method passing in a RobotModel.
+   * Instead of this method, use the method passing in a RobotModel!
+   * Default implementation returns false, indicating that this API is not supported.
    */
-  virtual bool initialize(const std::string& robot_description, const std::string& group_name,
-                          const std::string& base_frame, const std::string& tip_frame,
-                          double search_discretization) = 0;
+  MOVEIT_DEPRECATED virtual bool initialize(const std::string& robot_description, const std::string& group_name,
+                                            const std::string& base_frame, const std::string& tip_frame,
+                                            double search_discretization);
 
   /**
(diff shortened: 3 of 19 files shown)


rhaschke:
> A JMG however always needs to be feed with `setSolverAllocators` by someone else, so it is not really self contained in the first place.

That's true.

> This dependency could be made more explicit and then also for your case of having JMG/RM without KPL it would be obvious that you need a replacement class (and storage) for KPL.

However, this dependency can but doesn't need to be satisfied by KPL. In principle you could directly link against a KS plugin and pass `std::make_shared` as an allocator. 
Again, while the JMG _needs_ a KS to be operational (and for this reason should own it), the KPL just _caches_  the shared ptrs. This is a fundamental semantic difference.

> I think the right solution would actually be to have RobotState store (and accept in it's constructor) a raw pointer.

I disagree and I hope you don't consider that as viable option as well. Changing this at such a central spot, would dramatically increase fragility of the framework.

> As said in the inline comment, a shared pointer communicates (potential) transfer of ownership.

Agreed.

> There is however no intention to do transfer the ownership of RobotModel to some random plugin.

Indeed. Which is why I pass a reference, indicating that this is a borrowed pointer.

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -37,42 +37,87 @@
...
+static void no_deleter(const moveit::core::RobotModel*)
 {
-  robot_description_ = robot_description;
+}
+
+void KinematicsBase::storeValues(const moveit::core::RobotModel& robot_model, const std::string& group_name,
+                                 const std::string& base_frame, const std::vector<std::string>& tip_frames,
+                                 double search_discretization)
+{
+  robot_model_ = moveit::core::RobotModelConstPtr(&robot_model, &no_deleter);
```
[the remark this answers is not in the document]
rhaschke: I wouldn't pull in RobotState into the discussion. Passing a reference instead of a shared_ptr into the KS plugin, I indicate that ownership _is not_ transferred. Rather, I guarantee that this reference is alive as long as the KS is. It's the decision of the KS plugin how to proceed with the reference. To comply with the existing RobotState API, I form a new shared_ptr, which can be freely used by all entities of the KS - as long as the KS is alive.
This is safe and clean.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -380,21 +386,31 @@ class KinematicsBase
...
+   * @param group_name The group for which this solver is being configured
+   * @param base_frame The base frame in which all input poses are expected.
+   * This may (or may not) be the root frame of the chain that the solver operates on
+   * @param tip_frames The tip of the chain
+   * @param search_discretization The discretization of the search when the solver steps through the redundancy
+   * @return true if initialization was successful, false otherwise
+   *
+   * When returning false, the KinematicsPlugingLoader will use the old method, passing a robot_description.
+   * Default implementation returns false and issues a warning to implement this new API.
+   * TODO: Make this method virtual after some soaking time, replacing the fallback.
```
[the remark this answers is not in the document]
rhaschke: Yes, that's what I meant ;-)

rhaschke:
```suggestion
   * TODO: Make this method pure virtual after some soaking time, replacing the fallback.
```

moveit_core/kinematics_base/src/kinematics_base.cpp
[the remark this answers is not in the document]
simonschmeisser: I agree with "safe enough" but not with clean. As we are currently defining the API that is going to used for the entire lifetime of melodic I think we need to also consider the API of RobotState and look at the whole picture.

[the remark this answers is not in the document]
rhaschke: Can we continue this discussion on the phone? You can find my phone number from [here](https://ekvv.uni-bielefeld.de/pers_publ/publ/Home.jsp).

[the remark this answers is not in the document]
simonschmeisser: After the phone call I still have my reservations against that construct but since this approach is otherwise minimally invasive I'm fine with having it merged. Let's await reviews from others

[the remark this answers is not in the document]
rhaschke: I believe, this `no-deleter` pointer just feels weird because it's unusual. However it's a clean solution to decouple the usage counting in MoveIt (who actually owns the passed RobotModel) from the one in the external plugin lib (which should only receive and manage a borrowed reference). If the external plugin lib messes up it's usage counting (e.g. introducing circular references), this should not affect MoveIt.

rhaschke:
One idea of this PR was to ensure that the RobotModels used across the system are indeed identical and not out-of-sync. Creating a copy would raise this issue again.
If somebody should ever change the RobotModel on the fly, he should also ensure to reload the solvers, shouldn't he?

jonbinney:
Sounds like there's two somewhat separate issues here:
1. making sure that the initial robot model used by the kinematics matches what the planners are using, including any configuration done by parameters outside of the URDF
2. keeping the model used by the kinematics solver up-to-date if it changes elsewhere in the code

Keeping a copy would satisfy point 1. To satisfy point 2, some amount of reloading may be necessary, but I think between reloads it would make sense for the IK solver's internally stored robot model to match what its internal datastructures were constructed with.

rhaschke:
While the second issue (updating the kinematics solvers or other parts of the framework if the RobotModel has changed) is an interesting and useful feature request, we should not tackle this in the scope of this PR. I think this feature will require more intense discussions on how to handle such changes. Please open a feature request for this.

rhaschke:
@v4hn, @davetcoleman: Would be great to get feedback on the proposed approach. I we agree, it remains to tackle the other plugins (`lma`, `srv`, `ikfast`) in a similar fashion as I already did for `kdl`.

rhaschke:
Rebased to resolve conflicts.

MIGRATION.md
```diff
@@ -8,6 +8,7 @@ API changes in MoveIt! releases
 - The move_group capability ``ExecuteTrajectoryServiceCapability`` has been removed in favor of the improved ``ExecuteTrajectoryActionCapability`` capability. Since Indigo, both capabilities were supported. If you still load default capabilities in your ``config/launch/move_group.launch``, you can just remove them from the capabilities parameter. The correct default capabilities will be loaded automatically.
 - Deprecated method ``CurrentStateMonitor::waitForCurrentState(double wait_time)`` was finally removed.
 - Renamed ``RobotState::getCollisionBodyTransforms`` to ``getCollisionBodyTransform`` as it returns a single transform only.
+- KinematicsBase: Deprecate members tip_frame_, search_discretization_. Use tip_frames_ and redundant_joint_discretization_ instead.
```
davetcoleman: This deprecation change could easily be merged in via a separate PR, making this one easier to review

moveit_core/kinematics_base/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ set(MOVEIT_LIB_NAME moveit_kinematics_base)
 
 add_library(${MOVEIT_LIB_NAME} src/kinematics_base.cpp)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION ${${PROJECT_NAME}_VERSION})
+target_compile_options(${MOVEIT_LIB_NAME} PRIVATE -Wno-deprecated-declarations)
```
davetcoleman: what is the reasoning for this? can we keep this warning on? if not, add comment as to why not

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -569,31 +578,22 @@ class KinematicsBase
...
-                           // previously generated custom ik solvers like IKFast
 
-  double search_discretization_;  // DEPRECATED - this variable only still exists for backwards compatibility
-                                  // with previous implementations.  Discretization values for each joint are
-                                  // now stored in the redundant_joint_discretization_ member
+  // The next two variables still exists for backwards compatibility
+  // with previously generated custom ik solvers like IKFast
+  // Replace tip_frame_ -> tip_frames_[0], search_discretization_ -> redundant_joint_discretization_
+  MOVEIT_DEPRECATED std::string tip_frame_;
+  MOVEIT_DEPRECATED double search_discretization_;
```
davetcoleman: i didn't know you could MOVEIT_DEPRECATED a variable. how does this change the behavior of anything?

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -569,31 +578,22 @@ class KinematicsBase
...
   std::string group_name_;
   std::string base_frame_;
   std::vector<std::string> tip_frames_;
-  std::string tip_frame_;  // DEPRECATED - this variable only still exists for backwards compatibility with
-                           // previously generated custom ik solvers like IKFast
 
-  double search_discretization_;  // DEPRECATED - this variable only still exists for backwards compatibility
-                                  // with previous implementations.  Discretization values for each joint are
-                                  // now stored in the redundant_joint_discretization_ member
+  // The next two variables still exists for backwards compatibility
```
davetcoleman: @mlautman any thoughts on these variables w.r.t. IkFast?

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -647,6 +647,10 @@ class KinematicsBase
     return false;
   }
 
+  void storeValues(const moveit::core::RobotModel& robot_model, const std::string& group_name,
```
davetcoleman: add comment - what does this function do? what is it storing where?

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -37,42 +37,89 @@
...
+
 namespace kinematics
 {
 const double KinematicsBase::DEFAULT_SEARCH_DISCRETIZATION = 0.1;
 const double KinematicsBase::DEFAULT_TIMEOUT = 1.0;
 
-void KinematicsBase::setValues(const std::string& robot_description, const std::string& group_name,
-                               const std::string& base_frame, const std::string& tip_frame,
-                               double search_discretization)
+static void no_deleter(const moveit::core::RobotModel*)
```
davetcoleman: why is this function using underscore format?

moveit_core/kinematics_base/src/kinematics_base.cpp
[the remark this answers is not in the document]
davetcoleman: this is indeed odd and warrants a lot of inline comments as to why this design decision was made. in particular, i'd like to hear your response to @simonschmeisser 's question about raw pointers - why not use those instead?

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -37,42 +37,89 @@
...
+
+  ROS_ERROR_NAMED(LOGNAME, "This solver does not support multiple tip frames");
+  return false;
+}
+
+bool KinematicsBase::initialize(const moveit::core::RobotModel& robot_model, const std::string& group_name,
+                                const std::string& base_frame, const std::vector<std::string>& tip_frames,
+                                double search_discretization)
+{
+  ROS_WARN_NAMED("kinematics_base", "IK plugin for group '%s' relies on deprecated API. "
```
davetcoleman: use LOGNAME

moveit_kinematics/cached_ik_kinematics_plugin/include/moveit/cached_ik_kinematics_plugin/cached_ik_kinematics_plugin.h
```diff
@@ -192,13 +221,23 @@ class CachedIKKinematicsPlugin : public KinematicsPlugin
 
   // virtual methods that need to be wrapped:
 
+  bool initialize(const moveit::core::RobotModel& robot_model, const std::string& group_name,
+                  const std::string& base_frame, const std::vector<std::string>& tip_frames,
+                  double search_discretization) override
+  {
+    return initImpl(robot_model, group_name, base_frame, tip_frames, search_discretization);
```
davetcoleman: rename to: initializeImpl

MIGRATION.md
rhaschke: Usually, when my PRs comprise several individual commits, they all have a distinct purpose and can/should be reviewed one-by-one. I don't think that it makes much sense to split these related commits into several PRs.

moveit_core/kinematics_base/CMakeLists.txt
rhaschke: Need to check.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
rhaschke: You should not use the variable anymore, otherwise you get a deprecation warning.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
rhaschke: Will add a comment. It's a replacement for `setValues()`.

rhaschke:
@davetcoleman I addressed your formal comments. Do you have any content-wise objections?
Otherwise I will proceed and adapt the remaining kinematics solvers.

simonschmeisser:
Do we really need to migrate all IK plugins at once? as @davetcoleman said this is already quite big

rhaschke:
> Do we really need to migrate all IK plugins at once?

I would like to get rid of corresponding deprecation warnings. As I said, the PR is split up into several individual commits to facilitate review and of course, this PR should not be squash-merged!

davetcoleman:
PRs like this are very hard to review, you can totally cherry-pick the simple style stuff into one or more separate PRs (e.g. "use LOGNAME", "cleanup KinematicsPluginLoad"). Those changes require very little cognitive load, whereas the magical pointer / reference stuff is tricky to understand when mixed in with a bunch of other files.

davetcoleman:
>  Do you have any content-wise objections? Otherwise I will proceed and adapt the remaining kinematics solvers.

It all seems reasonable enough

moveit_core/kinematics_base/CMakeLists.txt
rhaschke: Note, that I don't disable deprecation warnings per se, but only for building `kinematics_base.cpp`.
Deprecation of the variables is foremost to warn derived classes, which still use those variables.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
```diff
@@ -570,31 +580,22 @@ class KinematicsBase
...
-       // (if multiple tip frames provided, this variable will be unset)
-    search_discretization_(DEFAULT_SEARCH_DISCRETIZATION)
-    , default_timeout_(DEFAULT_TIMEOUT)
-  {
-    supported_methods_.push_back(DiscretizationMethods::NO_DISCRETIZATION);
-  }
+  KinematicsBase();
 
 protected:
+  moveit::core::RobotModelConstPtr robot_model_;
```
simonschmeisser: Shouldn't we add a warning about the fake nature of this raw pointer in disguise here?

rhaschke: The shared_ptr can be used as usual. There are no restrictions.

jonbinney:
I'm getting segmentation faults using trac_ik from debs and compiling melodic-devel branch of moveit from source with this PR merged. That's expected, right?

rhaschke:
No, that's not expected. I tested with IKFast only. Can you provide more details / a moveit_config to reproduce this?

rhaschke:
I found an TracIK setup and can reproduce the error. Will have a look.

jonbinney:
Thanks! I did just test with trac_ik from source, and no segfault. So that's good at least.

rhaschke:
I'm afraid that a rebuild is required as KinematicsBase has ABI-incompatible changes.

moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h
jonbinney: @rhaschke  does adding this `robot_model_` member variable break ABI compatibility? Woudn't that mean all kinematics plugins will need to be recompiled?

jonbinney: nvm - just saw your comment to this effect.

jonbinney:
Recompiling source code is easy - but this will mean that all IK packages used from debs cannot be used with source version of moveit until moveit is re-released and everything is re-released. @davetcoleman @130s do we have a timeline on next release?

rhaschke:
I'm planning to re-release as soon as #1096 is merged - after the planned sync on this weekend...

jonbinney:
Thanks @rhaschke . I appreciate your work on this!

v4hn:
> but this will mean that all IK packages used from debs cannot be used with source version of moveit until moveit is re-released and everything is re-released.

Do the plugin maintainers have to re-release themselves or is there a way to force rebuilds of non-broken packages on the buildfarm (I would have thought they rebuild everything either way before syncs, but I never looked into this domain before)

Also, the problem will only really be fixed once the next sync is through and this will take a bit anyway.
Thus I don't see the need to release *immediately*.
Given we just merged major changes to the code base I would prefer to hold of the release at least until December 7th (end of next week) to check for potential problems.

jonbinney:
Good point about the sync - and yes, I meant to say that the problem won't be fixed until moveit is re-leased and everything is _rebuilt_ (typo on my part). I do believe that a moveit release will trigger a buildfarm rebuild of all packages which depend on it.

rhaschke:
> I still consider the nodeleter very ugly.

That's a perfectly valid solution. Not breaking API everywhere could be even seen as beautyful ;-)

> Do the plugin maintainers have to re-release themselves?

No, all downstream packages are automatically rebuild and re-released (if they have modeled their dependencies correctly, of course).

v4hn:
> That's a perfectly valid solution.
> Not breaking API everywhere could be even seen as beautyful ;-)

No. It is practical from a maintainance perspective.
Not beautiful from the perspective of good code.

rhaschke:
Ok. I guess, we have different opinions on this...

PR #1167, opened 2018-10-28
WIP: Saving on changes in planning_frame

agutenkunst:
Related to https://github.com/ros-planning/moveit/pull/1135
I want to make the "*" appear in rviz if a change in the MotionPlanningFrame happens.
This is as far as I got. 
Does anyone has a good insight in how the display must "notify" rviz about a saveable state
which does not directly correlate to a rviz::Property?

If I emit changed() this will trigger an onEnable() an therefore perform the same setup steps as if 
the display itself would have been enabled / disabled. This resets things such as the goal_query_ state and is therefore undesired.

`setShouldBeSave(true)` sounds like the right thing todo by seems not to be noticed by Rviz.

Does anyone have a clue?

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
@@ -175,6 +175,7 @@ private Q_SLOTS:
   void changedMetricsSetPayload();
   void changedMetricsTextHeight();
   void changedWorkspace();
+  void changedPlanningFrame();
   void resetInteractiveMarkers();
   void motionPanelVisibilityChange(bool enable);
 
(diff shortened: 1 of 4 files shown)


rhaschke:
Probably, the correct way would be to inherit from rviz::Panel, which offers a signal `configChanged()` for exactly that purpose. I'm not sure, though, why it was decided to inherit from QWidget instead. Maybe @davetcoleman has some insight on this?
There is a related PR on rviz: https://github.com/ros-visualization/rviz/pull/1303

rhaschke:
Sorry for the extreme delay: I had this issue on my todo list for so long, but only today found some time to look into it. The solution was quite easy actually: the PropertyTreeModel has a configChanged() signal, which we need to trigger.
@agutenkunst, could you rebase onto master or melodic, please?
Maybe you want to connect some more widgets as well?

rhaschke:
Closing in favour of #1991.

PR #1173, opened 2018-10-30
MSA: support dark themes

VictorLamoine:
### Description

- GUI change only. This improves the `NavigationWidget` in the Setup Assistant for dark themes, ~~without changing anything when using a white theme~~.
- Can be cherry-picked but it is not necessary.
- Tested with Kubuntu 18.04.

~~The detection of a dark/white theme is done by comparing the luminance value of the widgets background color.~~ QPalette colors are used so that colors always match.

# Before
## White theme
before_white

## Dark theme
before_dark

# After (see updated images in the discussion)
## White theme
[after_white](https://user-images.githubusercontent.com/5566160/47720735-e0b02480-dc4e-11e8-8f8c-eb0739396d1c.png)

## Dark theme
[after_dark](https://user-images.githubusercontent.com/5566160/47720741-e443ab80-dc4e-11e8-919c-d9d35b293e0f.png)

### Checklist
- [x] ~~Extended the tutorials / documentation, if necessary [reference](http://moveit.ros.org/documentation/contributing/)~~
- [x] ~~Document API changes relevant to the user in the moveit/MIGRATION.md notes~~
- [x] ~~Created tests, which fail without this PR [reference](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/tests.html)~~
- [x] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers

### DIFF
--- a/moveit_setup_assistant/src/widgets/navigation_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/navigation_widget.cpp
@@ -150,8 +150,8 @@ void NavDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option, c
     painter->setPen(palette.color(QPalette::Button));
     painter->drawLine(option.rect.topLeft(), option.rect.topRight());
     painter->setPen(palette.color(QPalette::Light));
-    const QPoint offset(0,1);
-    painter->drawLine(option.rect.topLeft()+offset, option.rect.topRight()+offset);
+    const QPoint offset(0, 1);
+    painter->drawLine(option.rect.topLeft() + offset, option.rect.topRight() + offset);
   }
 
   QRect textRect(option.rect.x() + 10, option.rect.y(), option.rect.width() - 10, option.rect.height());
--- a/moveit_setup_assistant/src/widgets/navigation_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/navigation_widget.cpp
@@ -36,7 +36,6 @@
 
 #include "navigation_widget.h"
 #include <QApplication>
-#include <QDebug>
 #include <iostream>
 
 namespace moveit_setup_assistant
--- a/moveit_setup_assistant/src/widgets/navigation_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/navigation_widget.cpp
@@ -35,6 +35,7 @@
 /* Author: Dave Coleman */
 
 #include "navigation_widget.h"
+#include <QApplication>
 #include <QDebug>
 #include <iostream>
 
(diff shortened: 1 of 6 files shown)


gavanderhoorn:
Can we agree that if you've reached the point with an OSS project where things like how your GUI looks in certain themes you deserve a `1.0` release as apparently users have nothing "better" to worry about ;)

v4hn:
> Instead of introducing even more hard-coded colors, I suggest to rely on the theme's palette instead.
The background gradient (if necessary/visible at all) can be achieved by brightening/darkening the palette color.

I would say this is somewhat independent of the requested feature, but nice to have yes.
@rhaschke do you want to hold off the pull-request as long as palette support is not included?

VictorLamoine:
I will improve the merge request, I hope to have some time next week to work on this. (next step is doing the same in RViz!)

Instead of hard coded colors we will have hard coded offsets from colors, which is already much better.

rhaschke:
> Instead of hard coded colors we will have hard coded offsets from colors, which is already much better.

+1

> Do you want to hold off the pull-request as long as palette support is not included?

Yes. I prefer clean code over a quick fix.

v4hn:
> Yes. I prefer clean code over a quick fix.

This is neither about clean code nor about fixing anything.
It's about hardcoding colors or offsets, as @VictorLamoine stated above.

Thanks for your time @VictorLamoine !

VictorLamoine:
I have got a nice result with clean code.
I have ditched some gradients because they would be harder to create/maintain for different themes.

Using http://doc.qt.io/qt-5/qpalette.html roles we can get nice colors.

# After
## White theme
after_white
## Dark theme
after_dark

Tell me what you think!

rhaschke:
Thanks a lot for this update. The solution is much clearer now.
You should avoid merge commits in pull-requests as this will mess up the commit history of our master branch (at least if we don't squash your commits into a single one). Do you mind to re-commit your changes on top of melodic-devel? To do so, just
`git reset --soft origin/melodic-devel`
on this PR branch and commit your changes (again). In future, it's better to directly rebase onto the current head:
`git rebase origin/melodic-devel`

rhaschke:
I pushed another commit restoring the gradients. Please have a look at the results for a dark theme.

VictorLamoine:
It looks good on both themes with the gradients :+1: 

The de-activated menu doesn't look "great" in my opinion but it is fine:
screenshot_20190103_094824
screenshot_20190103_094838

PS: if you have some time you can also review https://github.com/ros-visualization/rviz/pull/1319 which are similar modifications for RViz

rhaschke:
After installing `qt5-style-plugins`, my qt5 applications follow the system style as well...
Using Xubuntu's adwaita-dark theme I observe the same issue as @VictorLamoine with dark texts on dark background. Looks like this is a bug in Qt, because I observe the same behaviour with other Qt applications and standard widgets like `LineEdit` as well.

rhaschke:
I fixed some other small issues with theme switching.

VictorLamoine:
I have tested again and it still looks good on my side. Do you want me to squash or do you use GitHub squash and merge?

rhaschke:
We will squash-commit if this gets approved.

v4hn:
All code changes currently in this request look good to me.
I cannot test this myself at the moment though, so I can't judge the actual appearance.

moveit_setup_assistant/src/widgets/navigation_widget.cpp
```diff
@@ -124,64 +124,48 @@ QSize NavDelegate::sizeHint(const QStyleOptionViewItem& option, const QModelInde
...
+    backgroundGradient.setColorAt(0, palette.color(QPalette::Highlight).lighter(125));
+    backgroundGradient.setColorAt(1, palette.color(QPalette::Highlight));
     painter->fillRect(option.rect, QBrush(backgroundGradient));
   }
   else
   {
-    backgroundGradient.setColorAt(0, QColor(245, 245, 245));
-    backgroundGradient.setColorAt(1, QColor(240, 240, 240));
+    backgroundGradient.setColorAt(0, palette.color(QPalette::Light));
+    backgroundGradient.setColorAt(1, palette.color(QPalette::Light).darker(125));
```
davetcoleman: make this darker amount less?

rhaschke: @davetcoleman I am to blame here, because I tuned those values. 
Of course, that's a matter of taste, but I liked the 3D-look of the gradient. In your original version, the gradient was only barely visible and I suggest to drop the gradient completely if you want to keep it as is (That's what @VictorLamoine did after his rework towards palette colors).
If you like to keep the gradient, however, please play around with the argument to `.darker(xxx)` yourself to find a value the matches your taste. I actually don't care...

VictorLamoine: The point of this code change is to make MSA look nice on most platforms, not just on one particular theme or desktop environment so I agree with you Dave.

 I will try to tweak those values and test the results on multiple environments to get the best compromise :)

davetcoleman:
I tested it again just now and I think 105 looks good:
image

Also:
image

moveit_setup_assistant/src/widgets/navigation_widget.cpp
davetcoleman:
```suggestion
    backgroundGradient.setColorAt(1, palette.color(QPalette::Light).darker(105));
```

rhaschke:
@davetcoleman Do you have any other complaints or is this ready for merging?

davetcoleman:
Thanks!!

VictorLamoine:
Thank you for your support

PR #1176, opened 2018-10-31
Melodic robot model testing utils

BryceStevenWilley:
### Description

Originally discussed in #1126, this PR provides a Builder class for making `RobotModels`. While it's aimed at making it very simple to make a specific `RobotModel` with the details you need for a test, I'm making it able to handle more complex Models as well.  

I'm marking this as a WIP because there are a few tests that I haven't converted to use this yet, and I wanted to ask for feedback: does it look easy to use, etc. It's not going to be useful if we (the developers/community) think the interface is clunky or verbose.

### Checklist
- [ ] Extended the tutorials / documentation: 
       Inline docs are present, but a section under [the tutorials](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html) would be nice too. 
- [x] Decide if this should be cherry-picked to other current ROS branches (should not)


moveit_core/utils/CMakeLists.txt
```diff
@@ -1,9 +1,21 @@
 set(MOVEIT_LIB_NAME moveit_utils)
 
-add_library(${MOVEIT_LIB_NAME} src/lexical_casts.cpp)
-set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION ${${PROJECT_NAME}_VERSION})
+if(CATKIN_ENABLE_TESTING)
+  add_library(${MOVEIT_LIB_NAME}
```
davetcoleman: i don't like how this library changes scope based on whether testing is enabled or not. why not just have a separate library for things that are only for tests, and other utils?

moveit_core/utils/include/moveit/utils/robot_model_builder.h
```diff
@@ -0,0 +1,178 @@
+/*********************************************************************
```
davetcoleman: is this intended to be used anywhere that is not test-specific? if not, rename file to something like ``robot_model_test_utils.h``

if you intend to use it outside of test, i'm concerned it has duplicate functionality with 
https://github.com/ros-planning/moveit/tree/melodic-devel/moveit_ros/planning/rdf_loader
https://github.com/ros-planning/moveit/tree/melodic-devel/moveit_ros/planning/robot_model_loader

moveit_core/utils/include/moveit/utils/robot_model_builder.h
```diff
@@ -0,0 +1,178 @@
...
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of Willow Garage nor the names of its
```
davetcoleman: remove willow here

moveit_core/utils/include/moveit/utils/robot_model_builder.h
```diff
@@ -0,0 +1,178 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Bryce Willey */
+
+#ifndef MOVEIT_CORE_UTILS_TEST_
+#define MOVEIT_CORE_UTILS_TEST_
+
+/** \file test_utils.h
+ *  \brief convenience functions and classes used for making simple robot models for testing.
```
davetcoleman: move brief next to author?

moveit_core/utils/include/moveit/utils/robot_model_builder.h
```diff
@@ -0,0 +1,178 @@
...
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF AD w/ 2VISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Bryce Willey */
+
+#ifndef MOVEIT_CORE_UTILS_TEST_
+#define MOVEIT_CORE_UTILS_TEST_
+
+/** \file test_utils.h
```
davetcoleman: out of date, no need to duplicate

moveit_core/utils/include/moveit/utils/robot_model_builder.h
```diff
@@ -0,0 +1,178 @@
...
+namespace moveit
+{
+namespace core
+{
+/** \brief Loads a robot from moveit_resources.
+ * \param[in] robot_name The name of the robot package in moveit_resources to load.
+ *            For example, "panda_description", or "faunc_description".
+ * \returns a RobotModel constructed from robot_name's URDF and SRDF.
+ */
+moveit::core::RobotModelPtr loadRobot(std::string robot_name);
```
davetcoleman:
```suggestion
moveit::core::RobotModelPtr loadTestingRobotModel(std::string robot_name);
```

moveit_core/utils/include/moveit/utils/robot_model_builder.h
```diff
@@ -0,0 +1,178 @@
...
+
+urdf::ModelInterfaceSharedPtr loadURDF(std::string robot_name);
```
davetcoleman:
```suggestion
urdf::ModelInterfaceSharedPtr loadModelInterface(std::string robot_name);
```

davetcoleman: URDF doesn't seem clear enough to me

moveit_core/utils/include/moveit/utils/robot_model_builder.h
davetcoleman: "Robot" is way too overloaded in this field....

moveit_core/utils/include/moveit/utils/robot_model_builder.h
```diff
@@ -0,0 +1,178 @@
...
+srdf::ModelSharedPtr loadSRDF(std::string robot_name);
+
+/** \brief Easily build different robot models for testing.
+ *  Essentially a programmer-friendly light wrapper around URDF and SRDF.
+ *  Best shown by an example:
+ *  \code{.cpp}
+ *  RobotModelBuilder builder("my_robot", "base_link");
+ *  builder.add("a->b->c", "continuous");
+ *  builder.addGroup({"a", "b"}, {}, "example_group");
+ *  RobotModelPtr model = builder.build();
```
davetcoleman: pretty slick!

moveit_core/utils/include/moveit/utils/robot_model_builder.h
```diff
@@ -0,0 +1,178 @@
...
+  /** \brief Adds a chain of links and joints to the builder.
+   *  The joint names are generated automatically as "<parent>-<child>-joint".
+   * \param[in] section A list of link names separated by "->". The first link should already be added to the build by
+   * the time this function is called
+   * \param[in] type The type of the joints connecting all of the given links, e.g. "revolute" or "continuous". All of
+   * the joints will be given this type. To add multiple types of joints, call this method multiple times
+   * \param[in] joint_origins The "parent to joint" origins for the joints connecting the links. If not used, all
+   * origins will default to the identity transform
+   */
+  void add(std::string section, std::string type, std::vector<geometry_msgs::Pose> joint_origins = {});
```
davetcoleman:
```suggestion
  void addChain(std::string section, std::string type, std::vector<geometry_msgs::Pose> joint_origins = {});
```
more descriptive function names

moveit_core/utils/include/moveit/utils/robot_model_builder.h
```diff
@@ -0,0 +1,178 @@
...
+
+  /** \brief Adds a collision mesh to a specific link.
+   *  \param[in] link_name The name of the link to which the mesh will be added. Must already be in the builder
+   *  \param[in] filename The path to the mesh file, e.g.
+   * "package://moveit_resources/pr2_description/urdf/meshes/base_v0/base_L.stl"
+   *  \param[in] origin The origin pose of this collision mesh relative to the link origin
+   */
+  void addCollMesh(std::string link_name, std::string filename, geometry_msgs::Pose origin);
```
davetcoleman:
```suggestion
  void addCollisionMesh(std::string link_name, std::string filename, geometry_msgs::Pose origin);
```

davetcoleman: don't abbreviate Coll, here and below

moveit_core/utils/include/moveit/utils/robot_model_builder.h
davetcoleman: also it seems this function has significantly different functionality than the rest of this class. does it belong in its own class or as just a helper function utility in another file?

moveit_core/utils/src/robot_model_builder.cpp
```diff
@@ -0,0 +1,335 @@
...
+{
+namespace core
+{
+const std::string LOGNAME = "robot_model_builder";
+moveit::core::RobotModelPtr loadRobot(std::string robot_name)
+{
+  urdf::ModelInterfaceSharedPtr urdf = loadURDF(robot_name);
+  srdf::ModelSharedPtr srdf = loadSRDF(robot_name);
+  moveit::core::RobotModelPtr robot_model;
+  robot_model.reset(new moveit::core::RobotModel(urdf, srdf));
```
davetcoleman: combine lines 52 & 53, no need to reset

moveit_core/utils/src/robot_model_builder.cpp
```diff
@@ -0,0 +1,335 @@
...
+#include <boost/math/constants/constants.hpp>
+#include <geometry_msgs/Pose.h>
+#include "moveit/utils/robot_model_builder.h"
+
+namespace moveit
+{
+namespace core
+{
+const std::string LOGNAME = "robot_model_builder";
+moveit::core::RobotModelPtr loadRobot(std::string robot_name)
```
davetcoleman: everywhere...
```suggestion
moveit::core::RobotModelPtr loadRobot(const std::string& robot_name)
```

moveit_core/utils/src/robot_model_builder.cpp
```diff
@@ -0,0 +1,335 @@
...
+}
+
+void RobotModelBuilder::add(std::string section, std::string type, std::vector<geometry_msgs::Pose> joint_origins)
+{
+  std::vector<std::string> link_names;
+  boost::split_regex(link_names, section, boost::regex("->"));
+  if (link_names.empty())
+  {
+    ROS_ERROR_NAMED(LOGNAME, "No links specified (empty section?)");
+    return;
```
davetcoleman: if this is going to be used for unit tests it seems the function should return false?

moveit_core/utils/src/robot_model_builder.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+  if (not joint_origins.empty() && link_names.size() - 1 != joint_origins.size())
+  {
+    ROS_ERROR_NAMED(LOGNAME, "There should be one more link (%zu) than there are joint origins (%zu)",
+                    link_names.size(), joint_origins.size());
+    return;
+  }
+
+  // Iterate through each link.
+  for (size_t i = 1; i < link_names.size(); i++)
```
davetcoleman: nit: ++i

moveit_core/utils/include/moveit/utils/robot_model_builder.h
BryceStevenWilley: There's not a huge reason this should be it's own class, and it is technically independent of the `RobotModelBuilder` class. I think renaming the file `robot_model_test_utils` will address this.

moveit_core/utils/include/moveit/utils/robot_model_builder.h
davetcoleman: I didn't do a deep dive, but do you think can we get rid of the rdf_loader after this is added? Or are they very different things? I haven't looked in a few years

davetcoleman:
Before we merge this, can you make a short PR to [the tests tutorial](https://github.com/ros-planning/moveit_tutorials/blob/melodic-devel/doc/tests/tests_tutorial.rst) explaining how to use this and linking to an example file or two in the code base?

moveit_core/utils/include/moveit/utils/robot_model_builder.h
BryceStevenWilley: I wouldn't say they're very different, rdf_loader seems to be a bit more general, looks like it handles Xacro files as well. It wouldn't be hard to move rdf_loader into moveit_core/utils and merge with `loadTestingRobotModel()` et al. though.  Should that be another PR, or should it go it here?

davetcoleman: Looking at it further, the rdf_loader has more ROS dependencies than allowed in moveit_core, e.g.

```
#include <ros/ros.h>
#include <ros/package.h>
```

For example it uses the parameter server. While I am conflicted about moveit_core's level of ROS-agnotistism (not very, but still a little) I think that's a discussion for another issue/PR. I'd say let's move with this PR as-is

davetcoleman:
Also please remove WIP when you're ready

BryceStevenWilley:
Should be ready.  Ended up making a conscious choice not to add enough features to converts bits of code like [this](https://github.com/ros-planning/moveit/blob/bce3a38640/moveit_core/robot_state/test/robot_state_test.cpp#L222), since it would make the interface too complicated to be used easily.  Might tackle that in a later PR.

moveit_core/utils/CMakeLists.txt
```diff
@@ -1,11 +1,23 @@
...
 
-add_library(${MOVEIT_LIB_NAME} src/lexical_casts.cpp)
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION ${${PROJECT_NAME}_VERSION})
 
-target_link_libraries(${MOVEIT_LIB_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+set(MOVEIT_TEST_LIB_NAME moveit_test_utils)
+add_library(${MOVEIT_TEST_LIB_NAME}
+  src/robot_model_test_utils.cpp)
+
+find_package(moveit_resources REQUIRED)
```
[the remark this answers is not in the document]
davetcoleman: +1 to using it for testing only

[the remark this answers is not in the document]
BryceStevenWilley: Didn't initially consider the test library could be conditionally built.  Changed

davetcoleman:
Whoo-hoo!

PR #1182, opened 2018-11-01
adapt boost_python dependency discovery for boost 1.65

pholthau:
### Description

I have boost 1.65 on my system which already uses the new python ``find_package`` behaviour that needs the python version as a suffix. ``planning_interface`` uses boost 1.67 as a lower bound for this behaviour instead. If the version string is changed, everything is discovered and ``planning_interface`` compiles successfully.

The commit only changes the version at which the new discovery method is applied. It should probably be cherry-picked to relevant branches as well.


BryceStevenWilley:
> I have boost 1.65 on my system which already uses the new python `find_package` behaviour that needs the python version as a suffix.

Can you explain this further?  The [release notes](https://www.boost.org/doc/libs/1_67_0/libs/python/doc/html/rn.html) of boost_python suggest that only 1.67 has this ability, not 1.65.

pholthau:
Hm, strange. I can't find any documentation or release notes for 1.65. I can only tell that on my system (gentoo) the behaviour is as I described. There are some patches regarding python Jamfiles, though. We might need to check them:
https://github.com/gentoo/gentoo/tree/master/dev-libs/boost

rhaschke:
Hi Patrick,
on my Ubuntu Bionic, I do have both, version and unversioned lib names:
```
$ ls -1 /usr/lib/x86_64-linux-gnu/libboost_python*
/usr/lib/x86_64-linux-gnu/libboost_python3.a
/usr/lib/x86_64-linux-gnu/libboost_python3-py36.a
/usr/lib/x86_64-linux-gnu/libboost_python3-py36.so
/usr/lib/x86_64-linux-gnu/libboost_python3-py36.so.1.65.1
/usr/lib/x86_64-linux-gnu/libboost_python3.so
/usr/lib/x86_64-linux-gnu/libboost_python.a
/usr/lib/x86_64-linux-gnu/libboost_python-py27.a
/usr/lib/x86_64-linux-gnu/libboost_python-py27.so
/usr/lib/x86_64-linux-gnu/libboost_python-py27.so.1.65.1
/usr/lib/x86_64-linux-gnu/libboost_python-py36.a
/usr/lib/x86_64-linux-gnu/libboost_python-py36.so
/usr/lib/x86_64-linux-gnu/libboost_python.so
```
Probably, it depends on how the OS distributor packages. In your case, there is some [symlinking going on in Gentoo](https://github.com/gentoo/gentoo/blob/master/dev-libs/boost/boost-1.65.0.ebuild#L82).

pholthau:
Hi Robert,

Yes, that could be an issue. Probably it's really just distribution-dependent, so please close this PR if you'd like:

```
$ ls -1 /usr/lib/libboost_python*
/usr/lib/libboost_python-2.7-mt.so
/usr/lib/libboost_python-2.7.so
/usr/lib/libboost_python-2.7.so.1.65.0
/usr/lib/libboost_python-3.6-mt.so
/usr/lib/libboost_python-3.6.so
/usr/lib/libboost_python-3.6.so.1.65.0
```

Either way, the find_package macro changed with 1.67, so from that version on everything will work again. I'll just patch the package myself or file an issue with [ros gentoo overlay](https://github.com/ros/ros-overlay).

PR #1199, opened 2018-11-13
fix several issues in rendering of attached bodies in rviz plugins

rhaschke:
Rebase of #1152. Splitted into the basic fix (this PR) and the additional new feature providing a BoolProperty to switch attached bodies on/off (#1200).
This fixes #512.
Also added bug fixes for several other rendering issues I observed during testing the initial commit.
Look into individual commits for details.

PR #1200, opened 2018-11-13
Enable/Disable attached object visualization

rhaschke:
Rebase of #1152, which is splitted into mere bug fixes (#1199) and this PR adding the possibility to enable/disable visualization of attached bodies.
From my point of view this new feature is not neccessary / might be even confusing, because the attached object(s) always belong to the robot and should be displayed along with them. See also discussion in https://github.com/ros-planning/moveit/pull/1152#issuecomment-438200738.

### DIFF
--- a/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
+++ b/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
@@ -124,6 +124,7 @@ private Q_SLOTS:
   void changedSceneEnabled();
   void changedSceneRobotVisualEnabled();
   void changedSceneRobotCollisionEnabled();
+  void changedSceneRobotAttachedBodyEnabled();
   void changedRobotSceneAlpha();
   void changedSceneAlpha();
   void changedSceneColor();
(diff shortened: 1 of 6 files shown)


simonschmeisser:
I totally agree that this is not strictly necessary but on the other hand you can disable all links of the robot individually which has proven quite useful for debugging/introspection. This just gets us one step closer to having the same functionality for attached objects.

v4hn:
> IMHO that attached object belongs to the robot and should be shown/hidden with it.

Opinions differ :)

> Couldn't we even add attached objects to the robot's link tree to individually show and hide them?

Are you volunteering to implement it?
Also I'm not sure it is any better as a solution in practice because attached objects can change a lot based on the context. and is usually very deep in the tree.

I am still very happy with this patch as is.

rhaschke:
I tested this commit locally and observed the following (additional) issues:
- The trajectory path display doesn't provide a separate checkbox to hide/show attached objects.
Fine with me, but inconsistent.
- The `Query Start/Goal State` doesn't show the attached body anymore (and no selection here as well).

**Before merging this, @svenkrause should address these issues.**
Feel free to open a new PR if you can't push against this one.

henningkayser:
> * The trajectory path display doesn't provide a separate checkbox to hide/show attached objects.
> Fine with me, but inconsistent.

I agree that it can stay like that

> * The Query Start/Goal State doesn't show the attached body anymore (and no selection here as well).

@rhaschke @svenkrause could you fix this so we can get this merged?

rhaschke:
I'm not going to work on this as I don't support this new "feature".  For me, it is more confusing than helpful.

rhaschke:
@svenkrause I'm going to close this as you don't seem to be interested anymore in this feature - after all there wasn't any new activity for several weeks. Feel free to open a new PR addressing the open issues pointed out in https://github.com/ros-planning/moveit/pull/1200#issuecomment-441022475.

PR #1203, opened 2018-11-16
Fix build farm error

rhaschke:
This PR fixes errors on the ROS build farm due to missing include folders:
http://build.ros.org/job/Kdev__moveit__ubuntu_xenial_amd64/
http://build.ros.org/job/Ldev__moveit__ubuntu_xenial_amd64/

### DIFF
--- a/moveit_planners/chomp/chomp_interface/CMakeLists.txt
+++ b/moveit_planners/chomp/chomp_interface/CMakeLists.txt
@@ -58,9 +58,7 @@ if(CATKIN_ENABLE_TESTING)
   # additional test dependencies
   find_package(moveit_ros_planning_interface REQUIRED)
   find_package(rostest REQUIRED)
-  # add moveit includes *before* the rest for better overlay compliance
-  include_directories(BEFORE ${moveit_ros_planning_interface})
-  include_directories(AFTER ${rostest_INCLUDE_DIRS})
+  include_directories(${moveit_ros_planning_interface_INCLUDE_DIRS})
   add_rostest_gtest(chomp_moveit_test
     test/chomp_moveit.test
     test/chomp_moveit_test.cpp)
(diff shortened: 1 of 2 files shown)


v4hn:
https://github.com/ros-planning/moveit/commit/2360413ceccb0a33f52877a785a3ad733271e0b7#commitcomment-31328971 for the merged melodic version of this with some comments?

The variable name is broken, yes. I have no idea how that happened...

However the `BEFORE` is necessary here to use the correct version of the MGI headers in the overlay.
With your patch it prefers the *libraries* in the overlay (few lines down) but the *headers* from /opt/ros .
This is broken on two levels.

The best way to avoid such problems altogether is by having a new package for the tests that hard-depends on the correct packages in the first `find_package` already.

rhaschke:
@v4hn This should be merged as a hotfix of the buildfarm. I'm not going to implement your proposed approach. This is an issue only for Xenial (and it's cmake). From Bionic on, we can use BEFORE...

moveit_planners/chomp/chomp_interface/CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ if(CATKIN_ENABLE_TESTING)
     test/chomp_moveit.test
     test/chomp_moveit_test.cpp)
   target_link_libraries(chomp_moveit_test
-    ${moveit_ros_planning_interface_LIBRARIES}
     ${catkin_LIBRARIES}
+    ${moveit_ros_planning_interface_LIBRARIES}
```
rhaschke: Is this semantically different?

v4hn: Yes.

- catkin includes libraries of all dependencies of the package in here (that list is 86 elements long on my system)

- the standard GNU toolchain uses the first appearance of a required function symbol and ignores later appearances

As a result you end up linking a library from an overlay (a dependency of moveit_ros_planning_interface), but included the headers of the overlayed package above.

Changing this does not change the problem though that adding /opt/ros/<distro>/include/ in `catkin_INCLUDE_DIRS` will accidentally provide access to the headers of `moveit_ros_planning_interface` that might be overlayed in a higher workspace. In that case the build will still use system headers but link the libraries added here.

rhaschke:
@v4hn I guess, you plan to apply #1224 here as well. Closing this one.

PR #1205, opened 2018-11-17
MSA: cleanup includes to speedup compiling

rhaschke:
MSA builds surprisingly slow. Cleaning up the includes a little bit, I'm hoping for a speedup.


moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
```diff
@@ -282,10 +280,10 @@ class MoveItConfigData
   // ******************************************************************************************
 
   /// Load a robot model
-  void setRobotModel(robot_model::RobotModelPtr robot_model);
+  void setRobotModel(moveit::core::RobotModelPtr robot_model);
```
v4hn: you're usually more structured with changes like this. I would have expected that in a separate commit.

moveit_setup_assistant/src/tools/collision_linear_model.cpp
```diff
@@ -39,6 +39,7 @@
 
 #include <QItemSelection>
 #include <QPainter>
+#include <cmath>
```
v4hn: same here. I guess clang provides better warnings? :)

rhaschke:
> Did you benchmark the compile times?

Not yet, because locally I'm using ccache. However, after merging, I'm curious to compare compile times on the ROS build farm, which doesn't use ccache. In the past, I have observed reductions in compile time, particularly when limiting template stuff (boost) to cpps.

v4hn:
> a speedup of 4%.

That seems quite a lot indeed. Could you link the webpages please?
I wonder what the deviation in build-time is between different compile jobs of previous moveit releases.

rhaschke:
> Could you link the webpages please?

Not directly. I manually extracted (relative) build times from the absolute time stamps provided by the ROS build farm. There is no web page directly showing the 4% speedup.

Issue #1212, opened 2018-11-21
Duplicate point in concatenated trajectory.


martiniil:
Consider two trajectories traj1 and traj2, where the start state of traj2 matches the end state of traj1. 
The result of appending traj2 to traj1 has a duplicate point. Unfortunately, there is no way of sanitizing a robot_trajectory::RobotTrajectory by e.g. removing duplicate points.

Our goal is to concatenate multiple trajectories given in the form of robot_trajectory::RobotTrajectory objects: https://github.com/PilzDE/pilz_industrial_motion/tree/kinetic-devel/pilz_trajectory_generation

As we do not want to increase the overall duration, we append with time interval dt=0.0. But controllers, so far at least the ros_controllers::JointTrajectoryController require a timewise strictly increasing trajectory.

We implemented a work-around, but it would be nice, if this issue gets solved on the moveit side.
PilzDE/pilz_industrial_motion@5e37860044eb6f467bae51eea4f7f39a81488225

simonschmeisser:
This can now be closed, right?

rhaschke:
At least #1213 implements a clean solution to concatenate these solutions.
However, there is no general method available to remove duplicate points from trajectories.

PR #1213, opened 2018-11-21
Add append function for appending only part of a trajectory

martiniil:
### Description

As a solution to #1212, we propose adding an append function to robot_trajectory::RobotTrajectory, which takes a start- and end-index for appending only a part of the trajectory. By this, we avoid re-assembling or removing points from an existing trajectory.

A cherry-pick to kinetic would be desirable.

### Checklist
- [x] Document API changes relevant to the user in the moveit/MIGRATION.md notes(http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/tests.html)

### DIFF
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -85,12 +85,7 @@ void RobotTrajectory::swap(RobotTrajectory& other)
 
 void RobotTrajectory::append(const RobotTrajectory& source, double dt)
 {
-  waypoints_.insert(waypoints_.end(), source.waypoints_.begin(), source.waypoints_.end());
-  std::size_t index = duration_from_previous_.size();
-  duration_from_previous_.insert(duration_from_previous_.end(), source.duration_from_previous_.begin(),
-                                 source.duration_from_previous_.end());
-  if (duration_from_previous_.size() > index)
-    duration_from_previous_[index] += dt;
+  append(source, dt, 0, source.waypoints_.size());
 }
 
 void RobotTrajectory::append(const RobotTrajectory& source, double dt, size_t start_index, size_t end_index)
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -83,11 +83,6 @@ void RobotTrajectory::swap(RobotTrajectory& other)
   duration_from_previous_.swap(other.duration_from_previous_);
 }
 
-void RobotTrajectory::append(const RobotTrajectory& source, double dt)
-{
-  append(source, dt, 0, source.waypoints_.size());
-}
-
 void RobotTrajectory::append(const RobotTrajectory& source, double dt, size_t start_index, size_t end_index)
 {
   end_index = std::min(end_index, source.waypoints_.size());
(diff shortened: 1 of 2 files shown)


simonschmeisser:
please remove work items (checkboxes) that you don't intent to ever do (ie screenshots etc)

Otherwise this is a really nice PR and I'm looking forward to using it!

moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -93,6 +93,21 @@ void RobotTrajectory::append(const RobotTrajectory& source, double dt)
     duration_from_previous_[index] += dt;
 }
 
+void RobotTrajectory::append(const RobotTrajectory& source, double dt, size_t start_index, size_t end_index)
```
simonschmeisser: Can "end_index" be guessed for convenience? `end_index = 0` leads to `end_index = source.waypoints_.size()`

MIGRATION.md
```diff
@@ -8,6 +8,7 @@ API changes in MoveIt! releases
 - The move_group capability ``ExecuteTrajectoryServiceCapability`` has been removed in favor of the improved ``ExecuteTrajectoryActionCapability`` capability. Since Indigo, both capabilities were supported. If you still load default capabilities in your ``config/launch/move_group.launch``, you can just remove them from the capabilities parameter. The correct default capabilities will be loaded automatically.
 - Deprecated method ``CurrentStateMonitor::waitForCurrentState(double wait_time)`` was finally removed.
 - Renamed ``RobotState::getCollisionBodyTransforms`` to ``getCollisionBodyTransform`` as it returns a single transform only.
+- Added ``RobotTrajectory::append(const RobotTrajectory& source, double dt, size_t start_index, size_t end_index)``.
```
rhaschke: I think this should not be part of migration notes, but rather will become part of the changelog.

rhaschke:
Please, _rebase_ to resolve conflicts.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
martiniil: I am not sure, if I get your point. Are you talking about default arguments? start_index=0 would make sense, but end_index=0 should lead to an empty append trajectory in my opinion. I think that is more intuitive.

MIGRATION.md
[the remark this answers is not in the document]
martiniil: I see...

martiniil: Ok, thanks.

moveit_core/robot_trajectory/src/robot_trajectory.cpp
rhaschke: What about `end_index = std::numeric_limits<std::size_t>::max()`?

rhaschke: As soon as we have default args, for both `start_index` and `end_index`, you can drop the old `append()` method completely.

martiniil: Sounds good.

v4hn: > What about `end_index = std::numeric_limits<std::size_t>::max()`?

0 would have been just fine here as a special case.
It does not make sense to ask to copy no element at all.
But the `max` solution is fine to.

v4hn:
https://github.com/ros-planning/moveit/pull/1162 provides a more general approach and will hopefully allow iterators here later, not just size_t indices.

The approach taken here is valid though and I will merge it.

v4hn:
@martiniil Please feel invited to support the WIP pull-request then :-)
It looks like the original author does not have enough time to work on it after WMD.

PR #1215, opened 2018-11-21
Add getMoveGroupClient() to move_group_interface

mintar:
Use case: I want to execute a trajectory asynchronously while also being able to monitor its execution and perhaps abort it.

* `MoveGroupInterface::asyncMove()` lets me start asynchronous execution
* `MoveGroupInterface::stop()` lets me abort it
* but there is no way to monitor the execution or get the result

This PR simply adds a getter for the internal `move_action_client`. This enables code like shown in the usage example below. I'm not sure that exposing the internal `move_action_client` is the nicest API (comments welcome), but at least it seems more elegant than replicating all of the client's API one-by-one. In the example below, I've only used `getState()`, `getResult()` and `cancelGoal()`; but the other methods are useful as well (for example `waitForResult()`).

### usage example

```cpp
moveit::planning_interface::MoveGroupInterface group;
// [...]
group.setNamedTarget("home");
group.plan(plan);
group.asyncMove();
std::shared_ptr<actionlib::SimpleActionClient<moveit_msgs::MoveGroupAction> >
    move_action_client = group.getMoveActionClient();

ros::Rate rate(10);
while (!move_action_client->getState().isDone() && ros::ok()) {
  ROS_DEBUG("move_action_client: %s", move_action_client->getState().toString().c_str());
  rate.sleep();
  // to abort execution of the trajectory: `move_action_client->cancelGoal();` or `group.stop();`
}

ROS_DEBUG("move_action_client: %s %s",
          move_action_client->getState().toString().c_str(),
          move_action_client->getState().getText().c_str());

auto error_code = move_action_client->getResult()->error_code.val;

if (error_code == moveit::planning_interface::MoveItErrorCode::SUCCESS) {
  ROS_INFO("Moving to home pose SUCCESSFUL");
  return 0;
} else {
  ROS_ERROR("Moving to home pose FAILED");
  return 1;
}
```

### TODOs

- [ ] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers. You can refer to our [ROS code review guide](https://github.com/rosin-project/ros_code_review_guide/blob/master/README.md) for direction

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -688,7 +688,7 @@ public:
   /** \brief Get the move_group action client used by the \e MoveGroupInterface.
       The client can be used for querying the execution state of the trajectory and abort trajectory execution
       during asynchronous execution. */
-  actionlib::SimpleActionClient<moveit_msgs::MoveGroupAction>& getMoveGroupClient();
+  actionlib::SimpleActionClient<moveit_msgs::MoveGroupAction>& getMoveGroupClient() const;
 
   /** \brief Plan and execute a trajectory that takes the group of joints declared in the constructor to the specified
      target.
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -48,7 +48,9 @@
 #include <moveit_msgs/Grasp.h>
 #include <moveit_msgs/PlaceLocation.h>
 #include <moveit_msgs/MotionPlanRequest.h>
+#include <moveit_msgs/MoveGroupAction.h>
 #include <geometry_msgs/PoseStamped.h>
+#include <actionlib/client/simple_action_client.h>
 #include <boost/shared_ptr.hpp>
 #include <tf/tf.h>
 
@@ -682,6 +684,11 @@ public:
      target.
       This call is not blocking (does not wait for the execution of the trajectory to complete). */
   MoveItErrorCode asyncMove();
+
+  /** \brief Get the move action client used by \e move_group.
+      The client can be used for querying the execution state of the trajectory and abort trajectory execution
+      during asynchronous execution. */
+  const std::shared_ptr<actionlib::SimpleActionClient<moveit_msgs::MoveGroupAction> > getMoveActionClient();
 
   /** \brief Plan and execute a trajectory that takes the group of joints declared in the constructor to the specified
      target.
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -688,7 +688,7 @@ public:
   /** \brief Get the move_group action client used by the \e MoveGroupInterface.
       The client can be used for querying the execution state of the trajectory and abort trajectory execution
       during asynchronous execution. */
-  const actionlib::SimpleActionClient<moveit_msgs::MoveGroupAction>& getMoveGroupClient();
+  actionlib::SimpleActionClient<moveit_msgs::MoveGroupAction>& getMoveGroupClient();
 
   /** \brief Plan and execute a trajectory that takes the group of joints declared in the constructor to the specified
      target.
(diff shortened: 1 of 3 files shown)


v4hn:
As discussed in private emails, I support this idea.
The MGI is supposed to make the users life *easier* and should not prevent useful applications,
such as aborting actions, looking at feedback or results in asyncMove.

I don't like the idea of sharing the action client though and would prefer to return a reference from the method.
Why did you decide to use a shared_ptr instead?

For consistency it might also make sense to change the method name to `getMoveGroupClient`.

Do you have to write out the pointer definition, or is there a typedef you can use?

mintar:
Regarding the return type:  I could return a reference to the underlying object directly (so, a `const actionlib::SimpleActionClient<moveit_msgs::MoveGroupAction>&`). Now that I think about it, that sounds even better.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -48,7 +48,9 @@
 #include <moveit_msgs/Grasp.h>
 #include <moveit_msgs/PlaceLocation.h>
 #include <moveit_msgs/MotionPlanRequest.h>
+#include <moveit_msgs/MoveGroupAction.h>
 #include <geometry_msgs/PoseStamped.h>
+#include <actionlib/client/simple_action_client.h>
```
v4hn: This line makes `actionlib` a `CATKIN_DEPENDS` of the `moveit_ros_planning_interface` package,
i.e. packages including the header have to be able to find actionlib headers.

Can you please add that to the `CMakeLists.txt`?

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -683,6 +685,11 @@ class MoveGroupInterface
       This call is not blocking (does not wait for the execution of the trajectory to complete). */
   MoveItErrorCode asyncMove();
 
+  /** \brief Get the move_group action client used by the \e MoveGroupInterface.
+      The client can be used for querying the execution state of the trajectory and abort trajectory execution
+      during asynchronous execution. */
+  const actionlib::SimpleActionClient<moveit_msgs::MoveGroupAction>& getMoveGroupClient();
```
v4hn: can you actually abort trajectories with a `const` client?
I would have expected this has to be non-const.

mintar:
Ok, all requested changes are done.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
mintar: Done.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
mintar: Yep, you're right, very good catch! `cancelGoal()` was commented out in my test code, so the compiler didn't catch that. Resolved.

v4hn:
Looks good to me. Thanks for your time @mintar :+1: 
Let's see what someone else has to say.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -1502,6 +1505,12 @@ moveit::planning_interface::MoveItErrorCode moveit::planning_interface::MoveGrou
   return impl_->move(false);
 }
 
+actionlib::SimpleActionClient<moveit_msgs::MoveGroupAction>&
+moveit::planning_interface::MoveGroupInterface::getMoveGroupClient()
```
rhaschke: This method is missing the const  specifier, while the implementation method has one.
```suggestion
moveit::planning_interface::MoveGroupInterface::getMoveGroupClient() const
```

mintar: Good catch, I obviously meant to make it const everywhere. Fixed!

mintar:
Nice! That went faster than expected.

v4hn:
@mintar Glad you talked to me about it :+1:

davidyanghit:
hi, I discovered there is a bug in current` getMoveGroupClient()`. To get the execution state of `aysncExecute()`, the return client should be `return * execute_action_client_ `instead of `*move_action_client_`. The latter one only works when you use `move()` and `plan(`), while  `execute_action_client_` will always work in `aysncExecute().`

simonschmeisser:
@davidyanghit it would be really nice if you could open a new PR with your proposed changes, it's easier to understand and comment then

PR #1222, opened 2018-11-23
set last_state only for active joints in chomp_planner

knorth55:
### Description

Current `last_state` implementation in `chomp_planner` regards `active joints == all joints`, which does not happen for dual arm robot such as `baxter`.

### Checklist


moveit_planners/chomp/chomp_motion_planner/src/chomp_planner.cpp
```diff
@@ -283,7 +283,9 @@ bool ChompPlanner::solve(const planning_scene::PlanningSceneConstPtr& planning_s
   // check that final state is within goal tolerances
   kinematic_constraints::JointConstraint jc(planning_scene->getRobotModel());
   robot_state::RobotState last_state(planning_scene->getRobotModel());
```
rhaschke: Initialize from `start_state` directly. Not sure, `req.start_state.joint_state` is complete!
```suggestion
  robot_state::RobotState last_state(start_state);
```

v4hn: @rhaschke You are right. Either way it is better to copy the C++ class instead of the message.

Whether your comment would have been a potential bug in the future depends on the outcome of #1165

PR #1227, opened 2018-11-26
kinetic backport of #1213

v4hn:
Add append function for appending only part of a trajectory (#1213)

* Add append function for appending only part of a trajectory

* Replace old append() using default parameters

### DIFF
--- a/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
+++ b/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
@@ -186,11 +186,17 @@ public:
   }
 
   /**
-   * \brief Add a trajectory to the end of the current trajectory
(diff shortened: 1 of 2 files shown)


rhaschke:
@v4hn Do you really want to backport this ABI change?

v4hn:
This question is why I did not merge it myself.

We have package-version-based sonames in kinetic+, so ABI breakage in releases does not matter.
Users are forced to recompile their workspaces on updates anyway.

The only people who will have to recompile *without* explicitly breaking their library references right away are those building moveit from source. But those are used to doing that either way.

So I am in favor of merging ABI changes to "stable" branches based on that reasoning.
API changes will of course not go to stable releases.

rhaschke:
I don't have a strong opinion on this myself. Just wanted to be reassured before merging ;-)

PR #1228, opened 2018-11-26
fix compatibility with KDL 1.3

v4hn:
- The previous VERSION_LESS check is broken,
the orocos_kdl_VERSION is not evaluated in this context

- The definition is not exported to third-party projects
that might include the header. After all these headers are public API.

Verified to compile with ROS kinetic + external kinematics plugin that include the headers (ur_kinematics).


v4hn:
> > The previous VERSION_LESS check is broken, the orocos_kdl_VERSION is not evaluated in this context.

> I think it is. At least in my checks it was.

In this case that might again be a difference between kinetic & melodic.
I did not investigate further, but it believe we checked and it was not evaluated.
The version check returned the wrong result with the right version number at least.

PR #1233, opened 2018-11-27
Selective backport of #1196 (clang fixes)

v4hn:
fix clang issues

- use MOVEIT_STRUCT_FORWARD for structs
- fix initialization of dynamic-size array
- fix implicit conversion for SortableDisabledCollision -> srdf::Model::DisabledCollision
- virtual destructors for classes with virtual methods
- use delete[] for memory allocated with new[]


rhaschke:
@v4hn Shouldn't we add a clang-build for Kinetic too to test this?

v4hn:
> Shouldn't we add a clang-build for Kinetic too to test this?

We could do that, but I don't have a strong opinion on it myself.
We already test melodic-devel and the important point from my perspective was to get the discovered bug-fixes in here too.

Issue #1209, opened 2018-11-20
point_containment_filter::ShapeMask: Find a way to enable OpenMP again


peci1:
This is a continuation of https://github.com/ros-planning/moveit_ros/issues/563 .

As a workaround, openmp pragma was disabled in the referenced bug. However, it seems that having the for loop over scan rays parallelized would make sense.

I did some tests, however, all were unsuccessful.

The commented out line is
`#pragma omp parallel for schedule(dynamic)`

I tried different schedules (static and guided) with various chunk_sizes (1 and 16). Nothing helped.

I also tried to play with the shared/private variables, but that also did not bring any speedup.

Maybe somebody who uderstands OpenMP better could have a look? As I looked around, using OpenMP with pointclouds is no big deal in other ROS nodes. However, this is the only one I found that tries to traverse directly the sensor_msgs::PointCloud2 using the dedicated iterators. All other code works more or less with the pointclouds converted to PCL formats.

Maybe, if somebody solved this issue, the correct way of parallelizing using these iterators could be added to their documentation.

PR #1234, opened 2018-11-27
enforce OpenMP support for perception

v4hn:
The build is broken without it for a long time already
and it is not reasonable to compile the octomap updater
plugins without OpenMP speedup.

### DIFF
--- a/moveit_ros/perception/CMakeLists.txt
+++ b/moveit_ros/perception/CMakeLists.txt
@@ -41,7 +41,7 @@ find_package(catkin REQUIRED COMPONENTS
 )
 
 find_package(Eigen3 REQUIRED)
-find_package(OpenMP)
+find_package(OpenMP REQUIRED)
 find_package(octomap REQUIRED)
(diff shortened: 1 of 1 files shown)


rhaschke:
I didn't understand your comment immediately.
> The build is broken without it for a long time already.

I guess, that the build fails, when OpenMP is missing?

Is / how is this related to https://github.com/ros-planning/moveit/issues/1209?

v4hn:
Sorry, I should have written one more sentence.

> Is / how is this related to #1209?

No, it's not related.

The build scripts of the submodules [add OpenMP_CXX_FLAGS](https://github.com/ros-planning/moveit/blob/melodic-devel/moveit_ros/perception/lazy_free_space_updater/CMakeLists.txt#L5) without checking whether it was found.
When openmp support is not found, the variable contains `NOTFOUND` and the compiler rejects the parameter.

I noticed that by accident compiling with an incomplete build of clang.

PR #1241, opened 2018-12-04
Travis: add clang build

rhaschke:
In addition to #1233, also make Travis to build once with clang.

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -26,8 +26,10 @@ RUN wstool init . && \
         sudo \
         # Required for installing dependencies
         python-rosdep \
-        # Preferred build tool
-        python-catkin-tools && \
+        # Preferred build tools
+        python-catkin-tools \
+        clang \
+        ccache && \
     # Download all dependencies of MoveIt!
     rosdep update && \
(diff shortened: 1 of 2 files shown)


v4hn:
> fatal: Remote branch clang-build not found in upstream origin

what?

rhaschke:
> fatal: Remote branch clang-build not found in upstream origin

I accidentally push my branch to ros-planning, which triggered a Travis build. Then I removed the branch again before Travis actually started to build and this results in the mentioned error. Not a big issue ;-)

v4hn:
>  Could not find compiler set in environment variable CC:
>  clang.

I guess it should have succeeded with your second patch, but this seems to be the output of CI for this commit.

rhaschke:
It can only succeed if the 2nd commit has been merged and dockerhub has rebuild the corresponding docker image (including clang). So, either we reverse the commit order and wait for these steps to be completed or we just trust, that it will work out as in melodic - and fix any issues later if not.

v4hn:
I will leave this choice to @davetcoleman , as I am currently still not familiar enough with our CI system.

v4hn:
@rhaschke reverted this again right after merge in https://github.com/ros-planning/moveit/commit/12129304860190302aea9bfd68b4d0f864e496b7 because of ABI problems with boost, see
https://travis-ci.org/ros-planning/moveit/jobs/464316478#L3205 .

PR #1244, opened 2018-12-05
Add a jog_arm teleop pkg to moveit_experimental

AndyZe:
### Description

This is a teleop/soft real-time control package. It takes up where PR #854 left off 5 months ago. Some of the updates since PR #854:

- Can switch between unitless [-1:1] or unit [m/s, rad/s] commands.

- Predict whether the arm is moving toward or away from singularity. If moving away, full speed is OK.

- Can publish either JointTrajectory msgs (for old UR robots) or ros_control Float64MultiArray messages, as for a JointGroupVelocityController.

A quick way to test this is to clone [this repo](https://github.com/UTNuclearRoboticsPublic/motoman_project). Then:

- roslaunch motoman_gazebo sia5_gazebo_nishida_lab.launch
- roslaunch motoman_moveit sia5_gazebo_nishida_lab_moveit_planning_execution.launch
- roslaunch moveit_experimental jog_with_spacenav.launch  (the yaml file is already set up with the right topic names, etc.)

~~An issue is a circular dependency between **moveit_experimental** and **moveit_ros_planning_interface**.~~ It looks like chomp doesn't actually require a build-dependency on moveit_experimental.

- [x] Update tutorials and write tests if this gets approved.
- [ ] **Merge #1282, and release and sync Kinetic before merging this PR.** #1282 introduces a dependency from moveit_core to moveit_experimental to cleanly uninstall the moveit_experimental package. However this dependency conflicts with required dependencies here.


moveit_experimental/jog_arm/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+add_executable(spacenav_to_twist src/jog_arm/teleop_examples/spacenav_to_twist.cpp)
+add_dependencies(spacenav_to_twist ${catkin_EXPORTED_TARGETS})
+target_link_libraries(spacenav_to_twist ${catkin_LIBRARIES} ${Eigen_LIBRARIES})
+
+add_executable(xbox_to_twist src/jog_arm/teleop_examples/xbox_to_twist.cpp)
+add_dependencies(xbox_to_twist ${catkin_EXPORTED_TARGETS})
+target_link_libraries(xbox_to_twist ${catkin_LIBRARIES} ${Eigen_LIBRARIES})
+
+add_executable(dragonrise_to_twist src/jog_arm/teleop_examples/dragonrise_to_twist.cpp)
```
jonbinney: What is a "dragon rise?"

AndyZe: It's a USB joystick somebody added via PR. I wouldn't mind removing it, though - I don't think they're common.

jonbinney:
The code here looks like it breaks down into two parts - a server that uses a twist to jog the arm, and a number of joystick-specific classes that create a twist message from a `sensor_msgs/Joy` message. Does the Joy to Twist conversion need separate code for every joystick? Would it be possible to use a generic approach that maps joystick axes to twist axes? Or even better, just use the teleop twist joy node for that purpose? https://github.com/ros-teleop/teleop_twist_joy

AndyZe:
The problem with teleop_twist_joy is it's meant for navigation, so there's only x/y and yaw control. I guess I could think about a more generic joystick class, maybe with its own yaml file for configuration.

People might want to use it other ways, like with a keyboard. Press 1 > +x, Press Q > -x, etc. (Also, I want to add a keyboard example...)

jonbinney:
Ah - good point about the 2D limitation of teleop_twist_joy. How about using `teleop_tools`? It has the ability to map to any axis, for example: https://github.com/ros-teleop/teleop_tools/blob/melodic-devel/joy_teleop/config/joy_teleop_example.yaml

(my main thought here is to get as much functionality as possible while adding as little code as possible, since I worry about debugging and maintaining code within moveit for each hardware device)

bmagyar:
We made teleop_tools with commanding higher level functionality in mind,
feel free to hit me up over there with suggested changes if need be :)

AndyZe:
@jonbinney I think we're actually on the same page. I'm not intending to have a different executable for every possible controller. I just wanted a couple small examples, then users will create their own. It was dumb of me to approve the Dragonrise PR.

AndyZe:
@jonbinney my last holdup re. generic joystick teleop. Is the hit to performance worth it? Dave had this comment earlier, so I rewrote the teleop examples as C++:

> having a Python-based bridge between message formats seems like an unnecessary latency for something you want to be very responsive (teleop). is this necessary?

jonbinney:
Hah! Sorry we're bouncing you around in terms of ideas. My sense is that for teleop, human latency (10s to 100s of milliseconds) is way higher than local network latency (1s of milliseconds?) @bmagyar what do you think?

AndyZe:
I think we can have both. One high-performance C++ example and one teleop_tools example. I'm currently hung up because teleop_tools cannot stamp its messages, as far as I can find. Prob need to make a PR there first.

AndyZe:
@jonbinney your change has been implemented  :+1:

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1260 @@
...
+  if (parameter_ns == "")
+  {
+    ROS_ERROR_STREAM_NAMED(NODE_NAME, "A namespace must be specified in the launch file, like:");
+    ROS_ERROR_STREAM_NAMED(NODE_NAME, "<param name=\"parameter_ns\" "
+                                      "type=\"string\" "
+                                      "value=\"left_jog_arm_server\" />");
+    return 0;
+  }
+
+  error += !rosparam_shortcuts::get("", n, parameter_ns + "/publish_period", ros_parameters_.publish_period);
```
jonbinney: Looks like this `rosparam_shortcuts` thing is somethign @davetcoleman made? In this particular case, I think the code would be just as readable (or more so) if you created a namespaced node handle and then used it to read the params rather than calling a utility function. No change required though.

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1260 @@
...
+  if (ros_parameters_.collision_check_rate < 0)
+  {
+    ROS_WARN_NAMED(NODE_NAME, "Parameter 'collision_check_rate' should be "
+                              "greater than zero. Check yaml file.");
+    return 0;
+  }
+
+  return 1;
+}
+}  // namespace jog_arm
```
jonbinney: Nitpick - add endline here

jonbinney:
@AndyZe wow thanks for adding the teleop node support so quickly! I've made some more detailed comments in my review, once those are addressed I think this can be merged.

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,281 @@
...
+#include <std_msgs/Bool.h>
+#include <std_msgs/Float64MultiArray.h>
+#include <tf/transform_listener.h>
+#include <trajectory_msgs/JointTrajectory.h>
+
+namespace jog_arm
+{
+// Variables to share between threads, and their mutexes
+struct jog_arm_shared
+{
```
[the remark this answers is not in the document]
AndyZe: I think for 2 reasons:

1) Simplicity - I only lock the variable(s) I need at any point. I never need to lock more than, like, 2 at a time.

2) If I did lock the whole struct, the threads would likely be waiting longer.

Efficiency wasn't on my mind. But, I'm not a computer scientist, so I don't know too much about mutexes. I could see breaking it into several smaller structs, perhaps (but I would rather not).

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,281 @@
...
+ * Class JogROSInterface - Instantiated in main(). Handles ROS subs & pubs and
+ * creates the worker threads.
+ */
+class JogROSInterface
+{
+public:
+  JogROSInterface();
+
+  // Store the parameters that were read from ROS server
+  static struct jog_arm_parameters ros_parameters_;
```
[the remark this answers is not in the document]
AndyZe: As I understand it, the constructor **JogCalcs** is used like a static function in thread creation, and it takes those parameters as an argument. Thus, they must also be static. The stack trace if those params are not static is:

> error: invalid use of member ‘jog_arm::JogROSInterface::ros_parameters_’ in static member function
   jog_arm::JogCalcs ja(ros_parameters_, shared_variables_, model_loader_ptr_);

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
AndyZe: One nice thing about rosparam_shortcuts is it prints an error message then exits if there's a missing parameter.

AndyZe:
OK, I think I'm caught up on your changes. Let me know if I should squash all of these commits.

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
jonbinney: Is it released in kinetic and melodic? Or is this something we should add to the `moveit.rosinstall` in the root of the repo?

jonbinney:
Changes approved. @davetcoleman it looks like you looked at the previous incarnation of this PR - if you can give a second approval I'll go ahead and merge.

jonbinney:
@AndyZe looks like the cmake file changed - could you resolve the conflicts?

@davetcoleman bump - just need a quick +1 from you if this looks OK. I've already reviewed the code.

moveit_experimental/CMakeLists.txt
```diff
@@ -53,13 +69,19 @@ catkin_package(
...
     moveit_core
     geometric_shapes
     eigen_stl_containers
     eigen_conversions
     moveit_msgs
     kdl_parser
     pluginlib
+    # The following are needed by jog_arm:
+    moveit_ros_planning_interface
+    rosparam_shortcuts
```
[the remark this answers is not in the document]
davetcoleman: I don't think it belongs in ros-planning since its just a core ros utility that wraps rosparam server, but could be moved into some other ROS org like https://github.com/ros... at this point in the life of ROS1 I'm not convinced its worth the effort / OSRF will want the hassle.

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
davetcoleman: Its been released in every ROS distro since Indigo and I release it very quick compared to e.g. moveit ;-)

davetcoleman:
First, the ros-teleop Github org is pretty cool!

moveit_experimental/msg/JogJoint.msg
```diff
@@ -0,0 +1,16 @@
+# This is a message to hold data to jog by specifying joint
```
davetcoleman: I think this new message type should be added to moveit_msgs... there are currently no message types in the /moveit repo which makes since since we have moveit_msgs

moveit_experimental/CMakeLists.txt
```diff
@@ -40,8 +44,20 @@ set(VERSION_FILE_PATH "${CATKIN_DEVEL_PREFIX}/include")
...
 
 set(THIS_PACKAGE_INCLUDE_DIRS
-    ${VERSION_FILE_PATH}
-    collision_distance_field/include
+  ${VERSION_FILE_PATH}
+  collision_distance_field/include
+  jog_arm/include
+)
+
+add_message_files(
```
davetcoleman: move to moveit_msgs (see comment below)

moveit_experimental/CMakeLists.txt
```diff
@@ -40,8 +44,20 @@ set(VERSION_FILE_PATH "${CATKIN_DEVEL_PREFIX}/include")
...
+  FILES
+  JogJoint.msg
+)
+
+generate_messages(
```
davetcoleman: remove in favor of moveit_msgs

moveit_experimental/CMakeLists.txt
```diff
@@ -54,17 +70,23 @@ catkin_package(
...
-    moveit_core
-    geometric_shapes
-    eigen_stl_containers
     eigen_conversions
-    moveit_msgs
+    eigen_stl_containers
+    geometry_msgs
+    geometric_shapes
     kdl_parser
+    message_runtime
```
davetcoleman: remove in favor of moveit_msgs

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+// Server node for arm jogging with MoveIt.
+
+#ifndef JOG_ARM_SERVER_H
+#define JOG_ARM_SERVER_H
```
davetcoleman: ROS style: JOG_ARM_JOG_ARM_SERVER

(package name first, then file name)

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+
+#include <Eigen/Eigenvalues>
+#include <moveit_experimental/JogJoint.h>
```
davetcoleman: moveit_msgs

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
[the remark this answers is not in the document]
davetcoleman: This feels very clunky, as if the individual mutex functionality should be abstracted. Or perhaps there is already a tool for what you are trying to do in ROS, i haven't read through all of this yet

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+
+// ROS params to be read
+struct JogArmParameters
+{
+  std::string move_group_name, joint_topic, cartesian_command_in_topic, command_frame, command_out_topic,
+      planning_frame, warning_topic, joint_command_in_topic, command_in_type, command_out_type;
+  double linear_scale, rotational_scale, joint_scale, lower_singularity_threshold, hard_stop_singularity_threshold,
+      lower_collision_proximity_threshold, hard_stop_collision_proximity_threshold, low_pass_filter_coeff,
+      publish_period, publish_delay, incoming_command_timeout, joint_limit_margin, collision_check_rate;
+  bool gazebo, collision_check, publish_joint_positions, publish_joint_velocities, publish_joint_accelerations;
```
davetcoleman: these variable names could be better - what does ``gazebo`` do? at least add comments... but typically you should first have a verb e.g. ``use_gazebo``

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+  // Jogging calculation thread
+  static void* jogCalcThread(void* thread_id);
+
+  // Collision checking thread
+  static void* collisionCheckThread(void* thread_id);
+
+  // Variables to share between threads
+  static struct jog_arm_shared shared_variables_;
+
+  // static robot_model_loader::RobotModelLoader *model_loader_ptr_;
```
davetcoleman: remove commented code

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+  bool readParameters(ros::NodeHandle& n);
+
+  // Jogging calculation thread
+  static void* jogCalcThread(void* thread_id);
+
+  // Collision checking thread
+  static void* collisionCheckThread(void* thread_id);
+
+  // Variables to share between threads
+  static struct jog_arm_shared shared_variables_;
```
davetcoleman: i don't understand why all of these are ``static``

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+};
+
+/**
+ * Class LowPassFilter - Filter the joint velocities to avoid jerky motion.
+ */
+class LowPassFilter
+{
+public:
+  explicit LowPassFilter(double low_pass_filter_coeff);
+  double filter(double new_measurement_);
```
davetcoleman: function parameters should not have trailing underscore in variable name

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+  void reset(double data);
+  double filter_coeff_ = 10.;
```
davetcoleman: all member variables should be private

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,281 @@
...
+  double filter(double new_msrmt);
+  void reset(double data);
+  double filter_coeff_ = 10.;
+
+private:
+  double prev_msrmts_[3] = { 0., 0., 0. };
+  double prev_filtered_msrmts_[2] = { 0., 0. };
+};
+
+LowPassFilter::LowPassFilter(const double low_pass_filter_coeff)
```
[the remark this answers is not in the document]
davetcoleman: const is not necessary for a POD

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+{
+  previous_measurements_[0] = data;
+  previous_measurements_[1] = data;
+  previous_measurements_[2] = data;
+
+  previous_filtered_measurements_[0] = data;
+  previous_filtered_measurements_[1] = data;
+}
+
+double LowPassFilter::filter(const double new_measurement_)
```
davetcoleman: double new_measurement

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+{
+  // Push in the new measurement
+  previous_measurements_[2] = previous_measurements_[1];
+  previous_measurements_[1] = previous_measurements_[0];
+  previous_measurements_[0] = new_measurement_;
+
+  double new_filtered_msrmt =
```
davetcoleman: can you add a comment for what kind of filter this? or a reference

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+#include <std_msgs/Bool.h>
+#include <std_msgs/Float64MultiArray.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+#include <tf2_ros/transform_listener.h>
+#include <trajectory_msgs/JointTrajectory.h>
+
+namespace jog_arm
+{
+// Variables to share between threads, and their mutexes
+struct jog_arm_shared
```
davetcoleman: ROS style: JogArmShared

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+  Eigen::MatrixXd pseudoInverse(const Eigen::MatrixXd& J) const;
+
+  bool addJointIncrements(sensor_msgs::JointState& output, const Eigen::VectorXd& increments) const;
+
+  // Reset the data stored in low-pass filters so the trajectory won't jump when
+  // jogging is resumed.
+  void resetVelocityFilters();
+
+  // Avoid a singularity or other issue.
+  // Needs to be handled differently for position vs. velocity control
```
davetcoleman: add ``TODO: `` if this is future work

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
```diff
@@ -0,0 +1,284 @@
...
+
+  void lowPassFilterPositions();
+
+  void insertRedundantPointsIntoTrajectory(trajectory_msgs::JointTrajectory& trajectory, int count) const;
+
+  const robot_state::JointModelGroup* joint_model_group_;
+
+  robot_state::RobotStatePtr kinematic_state_;
+
+  sensor_msgs::JointState jt_state_, original_jts_;
```
davetcoleman: variable names are inconsistent, seems the second one should be ``original_jt_state_``

moveit_experimental/jog_arm/launch/spacenav_cpp.launch
```diff
@@ -0,0 +1,12 @@
+<launch>
```
davetcoleman: Clarify with comment what this launch is for: an example implementation?

moveit_experimental/jog_arm/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>jog_arm</name>
+  <version>0.0.3</version>
+
+  <description>Provides real-time manipulator cartesian jogging.</description>
```
davetcoleman: Cartesian

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+// Server node for arm jogging with MoveIt.
+
+#include <jog_arm/jog_arm_server.h>
+#include <memory>
+
+// Initialize these static struct to hold ROS parameters.
+// They must be static because they are used as arguments in thread creation.
+jog_arm::JogArmParameters jog_arm::JogROSInterface::ros_parameters_;
```
davetcoleman: I've never seen a declared variable namespaced in this way and didn't know you could. you should just wrap this whole section in a jog_arm namespace

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+jog_arm::jog_arm_shared jog_arm::JogROSInterface::shared_variables_;
+std::unique_ptr<robot_model_loader::RobotModelLoader> jog_arm::JogROSInterface::model_loader_ptr_ = NULL;
```
davetcoleman: These are global variables, which should be avoided at all costs. Why is this necessary? If necessary, use the global variable format standard. I actually forgot what it is, since i just avoid globals ;-)

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+
+/////////////////////////////////////////////////////////////////////////////////
+// JogROSInterface handles ROS subscriptions and instantiates the worker
+// threads.
+// One worker thread does the jogging calculations.
+// Another worker thread does collision checking.
+/////////////////////////////////////////////////////////////////////////////////
+
+static const char* const NODE_NAME = "jog_arm_server";
```
davetcoleman: nit that you can ignore: C++11 has a better way for this now

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+static const int GAZEBO_REDUNTANT_MESSAGE_COUNT = 30;
+
+// MAIN
+int main(int argc, char** argv)
```
davetcoleman: main function should be at bottom of file

davetcoleman: actually, i've never seen a main class in a .cpp file that also has a .h file... the main should be in a totally different file that includes the JogROSInterface class

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+
+  return 0;
+}
+
+namespace jog_arm
+{
+// Constructor for the main ROS interface node
+JogROSInterface::JogROSInterface()
+{
+  ros::NodeHandle n;
```
davetcoleman: avoid one variable names except in very obvious cases like time t... ``nh`` is common

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+  // Publish freshly-calculated joints to the robot
+  // Put the outgoing msg in the right format (trajectory_msgs/JointTrajectory
+  // or std_msgs/Float64MultiArray).
+  ros::Publisher outgoing_cmd_pub;
+  if (ros_parameters_.command_out_type == "trajectory_msgs/JointTrajectory")
+    outgoing_cmd_pub = n.advertise<trajectory_msgs::JointTrajectory>(ros_parameters_.command_out_topic, 1);
+  else if (ros_parameters_.command_out_type == "std_msgs/Float64MultiArray")
+    outgoing_cmd_pub = n.advertise<std_msgs::Float64MultiArray>(ros_parameters_.command_out_topic, 1);
+
+  // Wait for low pass filters to stabilize
```
davetcoleman: Maybe provide a terminal print out explaining the wait?

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+  ros::Duration(10 * ros_parameters_.publish_period).sleep();
+
+  ros::Rate main_rate(1. / ros_parameters_.publish_period);
+
+  while (ros::ok())
+  {
+    ros::spinOnce();
+
+    pthread_mutex_lock(&shared_variables_.new_traj_mutex);
```
davetcoleman: This large number of mutex locks seems excessive and unnecessary... pro/cons: https://stackoverflow.com/questions/3652056/how-efficient-is-locking-an-unlocked-mutex-what-is-the-cost-of-a-mutex

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+    trajectory_msgs::JointTrajectory new_traj = shared_variables_.new_traj;
+    pthread_mutex_unlock(&shared_variables_.new_traj_mutex);
+
+    // Check for stale cmds
+    pthread_mutex_lock(&shared_variables_.incoming_cmd_stamp_mutex);
+    if ((ros::Time::now() - shared_variables_.incoming_cmd_stamp) <
+        ros::Duration(ros_parameters_.incoming_command_timeout))
+    {
+      // Mark that incoming commands are not stale
+      pthread_mutex_lock(&shared_variables_.command_is_stale_mutex);
```
davetcoleman: this lock can be moved outside the if statement since its used in both the if and else

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+        if (ros_parameters_.publish_joint_positions)
+          joints.data = new_traj.points[0].positions;
+        else if (ros_parameters_.publish_joint_velocities)
+          joints.data = new_traj.points[0].velocities;
+        outgoing_cmd_pub.publish(joints);
+      }
+    }
+    else
+    {
+      ROS_WARN_STREAM_THROTTLE_NAMED(2, NODE_NAME, "Stale or zero command. "
```
davetcoleman: we've standardized to ``LOGNAME`` instead of NODE_NAME for moveit, see http://moveit.ros.org/documentation/contributing/code/

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+
+// A separate thread for the heavy jogging calculations.
+void* JogROSInterface::jogCalcThread(void*)
+{
+  jog_arm::JogCalcs ja(ros_parameters_, shared_variables_, model_loader_ptr_);
+  return nullptr;
+}
+
+// A separate thread for collision checking.
+void* JogROSInterface::collisionCheckThread(void*)
```
davetcoleman:
```suggestion
void* JogROSInterface::collisionCheckThread()
```

passing in void is C-style

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+    {
+      pthread_mutex_lock(&shared_variables.joints_mutex);
+      sensor_msgs::JointState jts = shared_variables.joints;
+      pthread_mutex_unlock(&shared_variables.joints_mutex);
+
+      for (std::size_t i = 0; i < jts.position.size(); ++i)
+        current_state.setJointPositions(jts.name[i], &jts.position[i]);
+
+      // process collision objects in scene
+      std::map<std::string, moveit_msgs::CollisionObject> c_objects_map = planning_scene_interface.getObjects();
```
davetcoleman: i think the planning scene monitor would handle for you all the processing of newly received collision objects. why is this being done so manual here?

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+    }
+  }
+}
+
+// Constructor for the class that handles jogging calculations
+JogCalcs::JogCalcs(const JogArmParameters& parameters, jog_arm_shared& shared_variables,
+                   const std::unique_ptr<robot_model_loader::RobotModelLoader>& model_loader_ptr)
+  : move_group_(parameters.move_group_name), tf_listener_(tf_buffer_)
+{
+  parameters_ = parameters;
```
davetcoleman: move to initializer list, 2 lines above

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+    ros::Duration(0.1).sleep();
+  }
+  const robot_model::RobotModelPtr& kinematic_model = model_loader_ptr->getModel();
+  kinematic_state_ = std::make_shared<robot_state::RobotState>(kinematic_model);
+  kinematic_state_->setToDefaultValues();
+
+  joint_model_group_ = kinematic_model->getJointModelGroup(parameters_.move_group_name);
+
+  std::vector<double> dummy_joint_values;
+  kinematic_state_->copyJointGroupPositions(joint_model_group_, dummy_joint_values);
```
davetcoleman: are you assuming you are initializing the joint group to zero value? what is the point of these two lines?

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+    position_filters_[i].reset(jt_state_.position[i]);
+
+  // Wait for the first jogging cmd.
+  // Store it in a class member for further calcs.
+  // Then free up the shared variable again.
+  geometry_msgs::TwistStamped cartesian_deltas;
+  moveit_experimental::JogJoint joint_deltas;
+  while (ros::ok() && (cartesian_deltas.header.stamp == ros::Time(0.)) && (joint_deltas.header.stamp == ros::Time(0.)))
+  {
+    ros::Duration(0.05).sleep();
```
davetcoleman: sometimes your while loops wait for 0.001 and here its 0.05. this should be a standard constant variable declared above

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
```diff
@@ -0,0 +1,1254 @@
...
+  else
+    publishWarning(false);
+
+  // If using Gazebo simulator, insert redundant points
+  if (parameters_.gazebo)
+  {
+    insertRedundantPointsIntoTrajectory(new_traj_, GAZEBO_REDUNTANT_MESSAGE_COUNT);
+  }
+
+  return 1;
```
davetcoleman: ``return true;`` is more standard

moveit_experimental/jog_arm/src/jog_arm/teleop_examples/spacenav_to_twist.cpp
```diff
@@ -0,0 +1,66 @@
+#include "geometry_msgs/TwistStamped.h"
+#include "moveit_experimental/JogJoint.h"
+#include "ros/ros.h"
+#include "sensor_msgs/Joy.h"
+
+namespace to_twist
```
davetcoleman: namespace should be the package name, typically

moveit_experimental/jog_arm/src/jog_arm/teleop_examples/spacenav_to_twist.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+class spaceNavToTwist
```
davetcoleman: SpaceNavToTwist

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
AndyZe: Yep, I guess that doesn't make sense. I removed the lines and there seems to be no effect.

moveit_experimental/jog_arm/src/jog_arm/jog_arm_server.cpp
AndyZe: From the examples I see online, this is how pthreads are usually used. Anyway, I get compile errors if I try to eliminate that argument.

moveit_experimental/jog_arm/include/jog_arm/jog_arm_server.h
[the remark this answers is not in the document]
AndyZe: Well, I found that a single mutex to lock the whole struct dropped CPU load from ~40% to ~30%, so let's just do it that way.  :+1:

AndyZe:
Alright @davetcoleman, I think those changes are all done

moveit_experimental/CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ COMPONENTS
   std_msgs
   trajectory_msgs
   visualization_msgs
+  # The following are needed by jog_arm:
+  moveit_msgs
```
davetcoleman: duplicate, remove

please alphabetize as the others already are

remove comment

moveit_experimental/CMakeLists.txt
```diff
@@ -54,17 +60,24 @@ catkin_package(
...
-    moveit_core
-    geometric_shapes
-    eigen_stl_containers
     eigen_conversions
-    moveit_msgs
+    eigen_stl_containers
+    geometry_msgs
+    geometric_shapes
     kdl_parser
+    message_runtime
```
davetcoleman: i don't think message_runtime is required any more

moveit_experimental/CMakeLists.txt
```diff
@@ -54,17 +60,24 @@ catkin_package(
...
+    moveit_core
+    moveit_msgs
     pluginlib
     std_msgs
     trajectory_msgs
-    geometry_msgs
     visualization_msgs
+    # The following are needed by jog_arm:
+    joy_teleop
+    moveit_msgs
```
davetcoleman: duplicate
alphabatize the list please

moveit_experimental/CMakeLists.txt
```diff
@@ -54,17 +60,24 @@ catkin_package(
...
+    message_runtime
+    moveit_core
+    moveit_msgs
     pluginlib
     std_msgs
     trajectory_msgs
-    geometry_msgs
     visualization_msgs
+    # The following are needed by jog_arm:
+    joy_teleop
```
davetcoleman: @jonbinney is this a stable package to require moveit to depend on?

moveit_experimental/jog_arm/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+add_executable(jog_arm_server
```
davetcoleman: use a variable to define "jog_arm" and use as a namespace in all executables in this CMakeList, including spacenav_*

see https://github.com/ros-planning/moveit/blob/melodic-devel/moveit_core/robot_state/CMakeLists.txt#L1

moveit_experimental/jog_arm/include/jog_arm/jog_ros_interface.h
```diff
@@ -0,0 +1,252 @@
...
+  // Indicates that we have not received a new command in some time
+  bool command_is_stale = false;
+
+  // The new trajectory which is calculated
+  trajectory_msgs::JointTrajectory new_traj;
+
+  // Timestamp of incoming commands
+  ros::Time incoming_cmd_stamp = ros::Time(0.);
+
+  bool ok_to_publish = false;
```
davetcoleman: add comment indicating why it wouldn't be ok / how this is used

moveit_experimental/jog_arm/include/jog_arm/jog_ros_interface.h
```diff
@@ -0,0 +1,252 @@
...
+  // Jogging calculation thread
+  static bool startJogCalcThread(const JogArmParameters& parameters, JogArmShared& shared_variables,
+                                 const robot_model_loader::RobotModelLoaderPtr model_loader_ptr);
+
+  // Collision checking thread
+  static bool startCollisionCheckThread(const JogArmParameters& parameters, JogArmShared& shared_variables,
+                                        const robot_model_loader::RobotModelLoaderPtr& model_loader_ptr);
+
+  // Variables to share between threads
+  struct JogArmShared shared_variables_;
```
davetcoleman: ``struct`` should not be here  (that's C-style i think)

moveit_experimental/jog_arm/include/jog_arm/jog_ros_interface.h
```diff
@@ -0,0 +1,252 @@
...
+
+  robot_model_loader::RobotModelLoaderPtr model_loader_ptr_;
+
+  // Store the parameters that were read from ROS server
+  struct JogArmParameters ros_parameters_;
```
davetcoleman: ``struct`` should not be here  (that's C-style i think)

moveit_experimental/jog_arm/include/jog_arm/jog_ros_interface.h
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * Class LowPassFilter - Filter the joint velocities to avoid jerky motion.
+ * This is a second-order Butterworth low-pass filter.
+ * See https://ccrma.stanford.edu/~jos/filters/Example_Second_Order_Butterworth_Lowpass.html
+ */
+class LowPassFilter
+{
+public:
+  explicit LowPassFilter(double low_pass_filter_coeff);
+  double filter(double new_measurement_);
```
davetcoleman: remove trailing underscore from new_measurement_

moveit_planners/chomp/chomp_interface/CMakeLists.txt
```diff
@@ -18,7 +18,6 @@ find_package(catkin REQUIRED COMPONENTS
   moveit_core
   pluginlib
   chomp_motion_planner
-  moveit_experimental
```
davetcoleman: these changes would ideally be in a separate PR

moveit_experimental/package.xml
```diff
@@ -59,6 +65,13 @@
   <run_depend>trajectory_msgs</run_depend>
   <run_depend>visualization_msgs</run_depend>
   <run_depend>moveit_core</run_depend>
+  <run_depend>message_runtime</run_depend>
+  <!-- The following are required by jog_arm -->
+  <run_depend>moveit_msgs</run_depend>
```
davetcoleman: duplicate

moveit_experimental/package.xml
```diff
@@ -38,6 +38,12 @@
   <build_depend>rosconsole</build_depend>
   <build_depend>eigen</build_depend>
   <build_depend>moveit_core</build_depend>
+  <build_depend>message_generation</build_depend>
+  <!-- The following are required by jog_arm -->
+  <build_depend>moveit_msgs</build_depend>
```
davetcoleman: duplicate

moveit_experimental/package.xml
```diff
@@ -38,6 +38,12 @@
   <build_depend>rosconsole</build_depend>
   <build_depend>eigen</build_depend>
   <build_depend>moveit_core</build_depend>
+  <build_depend>message_generation</build_depend>
```
davetcoleman: remove

moveit_experimental/package.xml
```diff
@@ -59,6 +65,13 @@
   <run_depend>trajectory_msgs</run_depend>
   <run_depend>visualization_msgs</run_depend>
   <run_depend>moveit_core</run_depend>
+  <run_depend>message_runtime</run_depend>
```
davetcoleman: remove

moveit_experimental/jog_arm/src/jog_arm/teleop_examples/spacenav_to_twist.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    spinner_.start();
+    ros::waitForShutdown();
+  };
+
+private:
+  ros::NodeHandle n_;
+  ros::Subscriber joy_sub_;
+  ros::Publisher twist_pub_, joint_delta_pub_;
+  ros::AsyncSpinner spinner_;
```
davetcoleman: member variables should be listed *after* private functions

moveit_experimental/jog_arm/src/jog_arm/teleop_examples/spacenav_to_twist.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "sensor_msgs/Joy.h"
+
+namespace jog_arm
+{
+class SpaceNavToTwist
+{
+public:
+  SpaceNavToTwist() : spinner_(1)
+  {
+    joy_sub_ = n_.subscribe("spacenav/joy", 1, &SpaceNavToTwist::joyCallback, this);
```
davetcoleman: make the constant ``1`` a constant variable with useful variable name so we know what it represents (buffer size?)

davetcoleman: same on next two lines

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+{
+  previous_measurements_[0] = data;
+  previous_measurements_[1] = data;
+  previous_measurements_[2] = data;
+
+  previous_filtered_measurements_[0] = data;
+  previous_filtered_measurements_[1] = data;
+}
+
+double LowPassFilter::filter(double new_measurement_)
```
davetcoleman: no trailing _ in function parameter

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+      {
+        // Apply user-defined scaling if inputs are unitless [-1:1]
+        if (parameters_.command_in_type == "unitless")
+          result[c] = command.deltas[m] * parameters_.joint_scale;
+        // Otherwise, commands are in m/s and rad/s
+        else if (parameters_.command_in_type == "speed_units")
+          result[c] = command.deltas[m] * parameters_.publish_period;
+        else
+          ROS_ERROR_STREAM_NAMED(LOGNAME, "Unexpected command_in_type");
+        goto NEXT_JOINT;
```
davetcoleman: never use goto's! 

replace with breaks and boolean flags to exit double for loops

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+  {
+    for (std::size_t c = 0; c < jt_state_.name.size(); ++c)
+    {
+      if (incoming_jts_.name[m] == jt_state_.name[c])
+      {
+        jt_state_.position[c] = incoming_jts_.position[m];
+        // Make sure there was at least one nonzero value
+        if (incoming_jts_.position[m] != 0.)
+          all_zeros = false;
+        goto NEXT_JOINT;
```
davetcoleman: remove goto (see below)

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+          halting = true;
+        }
+      }
+    }
+  }
+
+  return !halting;
+}
+
+void JogCalcs::publishWarning(const bool active) const
```
davetcoleman: bools (all PODs) do not need const when passed in

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+  for (std::size_t i = 0, size = static_cast<std::size_t>(delta_theta.size()); i < size; ++i)
+    theta[i] = prev_joints[i] + delta_theta(i);
+
+  kinematic_state_->setVariablePositions(theta);
+  jacobian = kinematic_state_->getJacobian(joint_model_group_);
+  svd = Eigen::JacobiSVD<Eigen::MatrixXd>(jacobian);
+  double new_condition = svd.singularValues()(0) / svd.singularValues()(svd.singularValues().size() - 1);
+  // If new_condition < ini_condition, the singular vector does point towards a
+  // singularity.
+  //  Otherwise, flip its direction.
```
davetcoleman: combine comment lines

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+  Eigen::VectorXd vector_toward_singularity = svd.matrixU().col(5);
+
+  double ini_condition = svd.singularValues()(0) / svd.singularValues()(svd.singularValues().size() - 1);
+
+  // This singular vector tends to flip direction unpredictably. See R. Bro,
+  // "Resolving the Sign Ambiguity
+  // in the Singular Value Decomposition"
+  // Look ahead to see if the Jacobian's condition will decrease in this
+  // direction.
+  // Start with a scaled version of the singular vector
```
davetcoleman: combine comment lines

moveit_experimental/jog_arm/include/jog_arm/jog_ros_interface.h
```diff
@@ -0,0 +1,252 @@
...
+  std::vector<jog_arm::LowPassFilter> position_filters_;
+
+  ros::Publisher warning_pub_;
+
+  JogArmParameters parameters_;
+
+  const int gazebo_redundant_message_count_ = 30;
+};
+
+class collisionCheckThread
```
davetcoleman: classes should be capitalized: ``CollisionCheckThread``

could you actually move this into a separate file? this class is already pretty big

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+// A separate thread for the heavy jogging calculations.
+bool JogROSInterface::startJogCalcThread(const JogArmParameters& parameters, JogArmShared& shared_variables,
+                                         const robot_model_loader::RobotModelLoaderPtr model_loader_ptr)
+{
+  JogCalcs ja(parameters, shared_variables, model_loader_ptr);
+  return true;
+}
+
+// A separate thread for collision checking.
+bool JogROSInterface::startCollisionCheckThread(const JogArmParameters& parameters, JogArmShared& shared_variables,
```
davetcoleman: move into separate file

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+namespace jog_arm
+{
+/////////////////////////////////////////////////////////////////////////////////
+// JogROSInterface handles ROS subscriptions and instantiates the worker
+// threads.
+// One worker thread does the jogging calculations.
+// Another worker thread does collision checking.
+/////////////////////////////////////////////////////////////////////////////////
+
+static const double g_while_loop_wait = 0.001;
```
davetcoleman: this isn't a global but rather a constant, make all caps with no g_

davetcoleman:
Could you make a brief (or extensive) pull request to moveit_tutorials documenting how this works, and maybe adding some images or video?

moveit_experimental/jog_arm/include/jog_arm/jog_ros_interface.h
gavanderhoorn: It's just a `class` but with all members public by default.

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
gavanderhoorn: This confuses me: why don't plain-old-data types need `const`? Do you mean primitive types? That would still confuse me as it would not be universally true.

gavanderhoorn:
@AndyZe: just wanted to say :100: :+1:  for iterating on this PR.

moveit_experimental/jog_arm/include/jog_arm/jog_ros_interface.h
davetcoleman: I know what a struct is! :-)

Here he's doing a C-style declaration of a variable that is only for backwards compatibility 
https://stackoverflow.com/questions/7729646/using-struct-keyword-in-variable-declaration-in-c

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
davetcoleman: a POD variable is copied when passed in, so the ``const`` doesn't do anything and by common convention is left out. 

C++03: 
> Section 3.9.1 lists bool as an integral type, and although not stated explicitly, it is reasonable to deduce that integral types are arithmetic types. Consequently, bool is a POD type.

davetcoleman: @AndyZe overall this really doesn't matter, I'm just giving you a Google-style code review and maybe being too picky

rhaschke:
Attention: this PR conflicts with a pending cleanup of chomp in #1282. In order to "uninstall" the moveit_experimental binary package, we have to install an empty experimental package before all other moveit packages. Thus, #1282 temporarily introduces a dependency from moveit_core to moveit_experimental. **Please merge this PR only after next Kinetic release and sync.**

rhaschke:
My question raised in https://github.com/ros-planning/moveit/pull/854#pullrequestreview-134287967 remains: 
> This contribution is already available as an independent, released ROS package: http://wiki.ros.org/jog_arm. Why do you want to migrate it into `moveit-experimental`?

I like this package very much, but I want to understand your motivation to migrate it.

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+  NEXT_JOINT:;
+  }
+
+  return result;
+}
+
+// Calculate a pseudo-inverse.
+Eigen::MatrixXd JogCalcs::pseudoInverse(const Eigen::MatrixXd& J) const
+{
+  return J.transpose() * (J * J.transpose()).inverse();
```
rhaschke: This computation is rather inefficient and numerically instable (close to singularities).
As you compute the SVD anyway (J = U * S * V^t), consider to use J^# = V * S^-1 * U^t instead.

moveit_experimental/jog_arm/src/jog_arm/jog_ros_interface.cpp
```diff
@@ -0,0 +1,1243 @@
...
+}
+
+// Add the deltas to each joint
+bool JogCalcs::addJointIncrements(sensor_msgs::JointState& output, const Eigen::VectorXd& increments) const
+{
+  for (std::size_t i = 0, size = static_cast<std::size_t>(increments.size()); i < size; ++i)
+  {
+    try
```
rhaschke: Is it really necessary to check this all the time?
The size of those vectors shouldn't change over time, right?

AndyZe:
@rhaschke This isn't limited to `trajectory_msgs/JointTrajectory` anymore. You can specify a `std_msgs/Float64MultiArray` output topic (for ros_control JointGroupVelocityController). I did try using the standard MoveIt! action server about 18 months ago, but it was noticeably jerky.

AndyZe:
> My question raised in [#854 (review)](https://github.com/ros-planning/moveit/pull/854#pullrequestreview-134287967) remains:
> 
> > This contribution is already available as an independent, released ROS package: http://wiki.ros.org/jog_arm. Why do you want to migrate it into `moveit-experimental`?

Well thanks. Some reasons to migrate it are:

- MoveIt! is the de facto standard for arm control in ROS.

- I would like to be able to do more than plan A to B motions with MoveIt!, and I'm sure lots of other people do too. Teleoperation is cool and we will eventually want to start interacting with the world more (which requires force control). So there should be some type of package like this, eventually. (I was highly motivated by Boston Dynamics. I thought, "Why can Boston Dynamics open a door, but I can't do that in ROS?)

- It's really good to get more eyes on the code. For example, your reviews have helped me drop CPU load from ~40% to ~28%, and I think your pseudoinverse suggestion will help even more.

- It will be nice to have some help with maintaining the package, even though I intend to be around to help with maintenance for a long time.

moveit_experimental/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ COMPONENTS
   trajectory_msgs
   visualization_msgs
   # The following are needed by jog_arm:
-  moveit_msgs
+  control_msgs
```
gavanderhoorn: Is [ros-controls/control_msgs](https://github.com/ros-controls/control_msgs) the `control_msgs` package that is referred to here? That [doesn't appear](https://github.com/ros-controls/control_msgs/tree/fbd5cce46383b17504196848ad659f415aa5a2bd/control_msgs/msg) to have a `JointJog.msg`.

jonbinney: It's in a PR - we'll have to make sure it gets merged before this one: https://github.com/ros-controls/control_msgs/pull/26

jonbinney:
@AndyZe what kind of crash? If its reproducible, i can put a little time into helping track it down.

rhaschke:
> There is a weird circular dependency between chomp and some of the other files in moveit_experimental.

We cleaned up the Chomp folders in #1282. However, as pointed out [here](https://github.com/ros-planning/moveit/pull/1244#issue-236263096), we had to introduce a (temporary) dependency from moveit_core to moveit_experimental to cleanly uninstall the moveit_experimental package, which should not be released. This might now pose an issue for your package.

AndyZe:
Actually, going through the code line-by-line, the crash happens almost right away, here in readParameters:

`
  pthread_mutex_lock(&shared_variables_.shared_variables_mutex);
  shared_variables_.command_deltas.header.frame_id = ros_parameters_.command_frame;
  pthread_mutex_unlock(&shared_variables_.shared_variables_mutex);
 `

AndyZe:
An update here... A [PR for a tutorial](https://github.com/ros-planning/moveit_tutorials/pull/283) has been added. I think we are still waiting on the [control_msgs PR](https://github.com/ros-controls/control_msgs/pull/26) to be merged.

bmagyar:
I was preparing a new release of `control_msgs` last night but it turned out there's a hiccup with rosdep at the moment. I'll try again tonight

bmagyar:
Both Kinetic and Melodic releases of `control_msgs` are out.
https://github.com/ros/rosdistro/pull/20346
https://github.com/ros/rosdistro/pull/20347

AndyZe:
@bmagyar thanks for doing that so quickly!

AndyZe:
The following commits are courtesy of @machinekoder  ([PR](https://github.com/UTNuclearRoboticsPublic/jog_arm/pull/84))

AndyZe:
@machinekoder created some tests for this package (very generously!) However it is written in Python. Take a look [here](https://github.com/machinekoder/jog_arm/blob/kinetic/jog_arm/test/integration/test_jog_arm_integration.py) and let me know if you want these tests in MoveIt!

moveit_experimental/CMakeLists.txt
```diff
@@ -24,48 +28,59 @@ find_package(catkin REQUIRED)
...
 #  visualization_msgs
-#)
-#find_package(Eigen3 REQUIRED)
-## Eigen 3.2 (Wily) only provides EIGEN3_INCLUDE_DIR, not EIGEN3_INCLUDE_DIRS
-#if(NOT EIGEN3_INCLUDE_DIRS)
-#  set(EIGEN3_INCLUDE_DIRS ${EIGEN3_INCLUDE_DIR})
-#endif()
+)
+find_package(Eigen3 REQUIRED)
+# Eigen 3.2 (Wily) only provides EIGEN3_INCLUDE_DIR, not EIGEN3_INCLUDE_DIRS
```
davetcoleman: Wily is EOL, please remove this if endif and comment

moveit_experimental/CMakeLists.txt
```diff
@@ -24,48 +28,59 @@ find_package(catkin REQUIRED)
...
+if(NOT EIGEN3_INCLUDE_DIRS)
+  set(EIGEN3_INCLUDE_DIRS ${EIGEN3_INCLUDE_DIR})
+endif()
+
 #find_package(octomap REQUIRED)
 #find_package(urdfdom REQUIRED)
 #find_package(urdfdom_headers REQUIRED)
```
davetcoleman: remove all commented out code please

moveit_experimental/CMakeLists.txt
```diff
@@ -24,48 +28,59 @@ find_package(catkin REQUIRED)
...
+#    eigen_conversions
+#    eigen_stl_containers
+#    geometric_shapes
+#    kdl_parser
+#    moveit_core
+#    moveit_msgs
+#    pluginlib
+#    std_msgs
+#    trajectory_msgs
+#    visualization_msgs
```
davetcoleman: please remove all commented out code

moveit_experimental/package.xml
```diff
@@ -60,7 +66,15 @@
   <run_depend>trajectory_msgs</run_depend>
   <run_depend>visualization_msgs</run_depend>
   <run_depend>moveit_core</run_depend>
+  -->
```
davetcoleman: please remove commented out code

davetcoleman:
Strong yes to integration tests! Although looking at them its not clear if they are auto-run by CMake / rostest...

AndyZe:
I opened a new PR #1360 for Melodic. May this be the last one.

davetcoleman:
For future reference, you can within Github retarget a PR to a different branch:

image

PR #1245, opened 2018-12-07
Mention the fixed AABB computation in migration.md

peci1:
As requested in https://github.com/ros-planning/moveit/pull/703#issuecomment-442811610


rhaschke:
I don't think we should clutter the migration notes with bug fix notes. The former are intended to provide instructions in case of API changes. I can consider this note in the next release changelog.

PR #1249, opened 2018-12-08
cleanup .clang-tidy

rhaschke:
- disable all checks
- to allow selective enabling afterwards (otherwise many undesired checks are enabled too)
- order checks


BryceStevenWilley:
LGTM.  Note this change should not be a functional change, as the only checks that are on by default in clang-tidy are "clang-analyzer-*" checks.  It does make it future proof in case clang decides to add more checks to the default, so I approve.

PR #1252, opened 2018-12-10
chomp: explicitly set variable as empty when unused

v4hn:
Followup to https://github.com/ros-planning/moveit/pull/1251#discussion_r240041450

@rhaschke


rhaschke:
Hm. Looks like a test became flaky... Unrelated to this PR though.

PR #1254, opened 2018-12-10
[proposal] RDFLoader / RobotModelLoader: remove parsing of TinyXML documents

rhaschke:
`urdf::Model::initXml()` is highly inefficient, as it first serializes the string and eventually
parses it again (see https://github.com/ros/urdf/issues/24#issue-389498748). We should discourage and eventually remove the TinyXML API and solely rely on parsing from string.

For now, this is just a proof-of-concept that the MoveIt source repo doesn't need the TinyXML support.
A user-friendly approach would deprecate the TinyXML API first, but, maybe, that's not really necessary?

Opinions?

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -16,7 +16,7 @@ API changes in MoveIt! releases
   Use `tip_frames_` and `redundant_joint_discretization_` instead.
 - KinematicsBase: Deprecated `initialize(robot_description, ...)` in favour of `initialize(robot_model, ...)`.
   Adapt your kinematics plugin to directly receive a `RobotModel`. See the [KDL plugin](https://github.com/ros-planning/moveit/tree/melodic-devel/moveit_kinematics/kdl_kinematics_plugin) for an example.
-- ``RDFLoader`` / ``RobotModelLoader``: removed TinyXML-based API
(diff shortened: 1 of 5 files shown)


jonbinney:
+1! I hadn't realized initXml() was doing that deserialize-reserialize step....

v4hn:
I don't have a strong opinion here, but if we want to remove it in melodic, please add a line to the `MIGRATION.md`

v4hn:
This code triggers [a deprecation warning](http://build.ros.org/job/Mdev__moveit__ubuntu_bionic_amd64/138/warnings24Result/) in the build farm now, because `urdf::Model` deprecated TinyXML.

So I guess this request is more about whether we want to *transition* to tinyxml2 or *only* remove the old API.

Given @rhaschke's argumentation, I agree to only remove the old API in `melodic-devel`.

rhaschke:
Added the requested migration note. @v4hn, do I understand correctly, that you accept the removal without a deprecation phase?

v4hn:
> do I understand correctly, that you accept the removal without a deprecation phase

Yes. I believe it is unnecessay overhead to drag this along as deprecated.
 If you agree, please merge.

PR #1256, opened 2018-12-11
moveit_experimental: commenting out dependencies

rhaschke:
As suggested in https://github.com/ros-planning/moveit/pull/1251#discussion_r240238940, comment-out all stuff in moveit_experimental, as there is nothing built.

### DIFF
--- a/moveit_experimental/package.xml
+++ b/moveit_experimental/package.xml
@@ -1,14 +1,7 @@
 <package>
   <name>moveit_experimental</name>
   <version>0.10.6</version>
-  <description>Experimental packages for moveit.</description>
-  <author email="isucan@google.com">Ioan Sucan</author>
-  <author email="robot.moveit@gmail.com">Sachin Chitta</author>
-  <author email="acorn.pooley@sri.com">Acorn Pooley</author>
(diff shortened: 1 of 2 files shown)


moveit_experimental/package.xml
```diff
@@ -1,21 +1,15 @@
...
   <name>moveit_experimental</name>
   <version>0.10.6</version>
-  <description>Experimental packages for moveit.</description>
-  <author email="isucan@google.com">Ioan Sucan</author>
-  <author email="robot.moveit@gmail.com">Sachin Chitta</author>
-  <author email="acorn.pooley@sri.com">Acorn Pooley</author>
-  <author email="jnicho@swri.org">Jorge Nicho</author>
-
-  <maintainer email="mferguson@fetchrobotics.com">Michael Ferguson</maintainer>
-  <maintainer email="chitt@live.in">Chittaranjan Srinivas Swaminathan</maintainer>
```
gavanderhoorn: I believe that manifests without any `maintainer` tags are invalid and will result in catkin erroring out.

v4hn:
I proposed this change to begin with, so I approve.
Thanks!

Issue #1255, opened 2018-12-10
Bad IK results/issues via setFromIK AND RViz Motion Planning interactive marker (failures and/or joint flips)


BrettHemes:
### Description
Since migrating to Melodic existing code base has started failing with bad trajectories returned from custom interpolators leveraging RobotState::setFromIK calls.  See RosAnswers question [here](https://answers.ros.org/question/310062/incorrectunexpected-behavior-from-trac-ik-with-calls-to-moveitcorerobotstatesetfromik/) when I thought it was a Trac-IK issue.  After playing with multiple solvers ,the issue seems to be deeper within MoveIt.  Running the exact same trajectory results in random failures sometimes succeeding and sometimes failing.  The exact failure mode depends on the solver used and the corresponding parameters (e.g., timeout and number of attempts).

Trac-IK is in general more robust than KDL (making debugging a bit harder) but has issues with both Speed and Distance solve types.  With Speed type, the solver often succeeds but sometimes returns a solution significantly far away from the seeded pose (i.e., joint flip).  The trajectory is pretty dense and thus the seeds are of very "good" quality (e.g., displacement on the order of 1e-2 radians from the true solution) so I don't know why this is happening.  Same problem with Distance solve_type unless the time is set to an overly large number (e.g., > 5ms).  Even then failures still happen...

KDL is not returning joint flips per se but will randomly fail to return a valid solution.  If called multiple times in a row on the **exact same seed** it will succeed, so it seems very much random

**While investigating I found the issue is (I believe) reproducable by just dragging the interactive tool planning marker around in RViz (see Steps to reproduce below).**

### Your environment
* ROS Distro: Melodic
* OS Version: e.g. Ubuntu 18.04
* Binary (newest as of today)

### Steps to reproduce
Calls to RobotState::setFromIK and/or dragging around interactive marker in RViz for any 6-dof industrial arms (e.g., kuka_experimental, fanuc, etc.).

Create a simple MoveIt Config via the setup assistant with the KR120 URDF from kuka_experimental.  Set chain solver from base_link to tool0 with either KDL or Trac-IK as the solver.  Run demo.launch and drag the interactive marker around.

### Expected behaviour
Continuous and good/close IK solutions.

### Actual behaviour
**With Trac-IK** small motions in the workspace of the interactive marker result in numerous and frequent joint flips (as was seen in API calls to RobotState::setFromIK).

[...]

PR #1261, opened 2018-12-11
Fix flaky test

### LINKED ISSUE #1259: Flaky unittests
Since recently, we have some flaky unittests at Travis: 

```
build/moveit_ros_planning_interface/test_results/moveit_ros_planning_interface/rosunit-cleanup_tests.xml
<?xml version="1.0" encoding="utf-8"?>
<testsuite errors="0" failures="1" name="unittest.suite.TestSuite" tests="2" time="3.417"><testcase classname="__main__.CleanupTest" name="test_cpp" time="1.6308"><failure type="AssertionError">False is not true
    testMethod()
  File "/root/ws_moveit/src/moveit/moveit_ros/planning_interface/test/cleanup.py", line 30, in test_cpp
    self.assertTrue(subprocess.call(cmd) == 0)
  File "/usr/lib/python2.7/unittest/case.py", line 422, in assertTrue
    raise self.failureException(msg)
</failure></testcase>
```

I wasn't able to reproduce this locally though. Any feedback is welcome!
Maybe it's related to #1104, using `createUniqueInstance()` for plugins.

rhaschke:
Fixes #1259.
The culprit was the [change in memory management of shared objects](https://github.com/ros-planning/moveit/pull/1104/commits/d0daf644d3d08ca8783d3dc588ced7479ddd9e30), deleting `tf2::Buffer` and it's associated `tf2::TransformListener` together - but unfortunately in the wrong order.

This also cleans up the corresponding unittests.


v4hn:
...
Thanks for the fast fixup.

v4hn:
As we were talking [about hot-fix releases the other day](https://github.com/ros-planning/moveit/pull/1104#issuecomment-445873038), I guess, we should do one for this?
This can trigger segfaults in existing code.

v4hn:
> There is another severe regression, which I want to address: #1255. Any help is welcome!

As I just argued there, I think this might require additional API / a novel concept introduced to MoveIt.

I don't think we should wait for this with the fixup release.

rhaschke:
> As I just argued there, I think this might require additional API / a novel concept introduced to MoveIt.

I don't agree.

> I don't think we should wait for this with the fixup release.

I agree. I can trigger a new release this evening.

Issue #1266, opened 2018-12-14
RobotState RViz crashes when sending joints that are not in the description


christian-rauch:
### Description

Overview of your issue here.

### Your environment
* ROS Distro: melodic
* OS Version: Ubuntu 18.04
* Binary build: 0.10.5

### Steps to reproduce
Activate the `RobotState` RViz plugin and send a `DisplayRobotState` message which contains joints that are not available in the robot description. E.g. "fubar"

Btw, the `Robot State Topic` field gives as hint: `The topic on which the moveit_msgs::RobotState messages are received`, which is wrong since the topic expects a `moveit_msgs::DisplayRobotState`.

### Expected behaviour
The plugin should show the robot articulated with the set of available joints.

### Actual behaviour
The plugin is ignoring the exception and lets RViz crash. The plugin should show a warning once, if it wants to notify the user, but it should definitely not crash the whole RViz session.

### Backtrace or Console output

```
[ERROR] [1544796166.730999500]: Variable 'fubar' is not known to model 'lwr'
Exception thrown.
terminate called after throwing an instance of 'moveit::Exception'
  what():  Variable 'fubar' is not known to model 'lwr'
```

rhaschke:
Closed via #1267. Thanks a lot for reporting and initial fix!

PR #1267, opened 2018-12-14
Handle exception in RobotStateDisplay and print warning to not crash RViz

christian-rauch:
### Description
The RobotStateDisplay plugin crashes RVIz if a robot state with a non-existing joint is provided. See: https://github.com/ros-planning/moveit/issues/1266

This PR simply catches the exception and shows a warning instead of forwarding the exception an have RViz crash.

### DIFF
--- a/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
+++ b/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
@@ -304,7 +304,14 @@ void RobotStateDisplay::newRobotStateCallback(const moveit_msgs::DisplayRobotSta
   if (!robot_state_)
     robot_state_.reset(new robot_state::RobotState(robot_model_));
   // possibly use TF to construct a robot_state::Transforms object to pass in to the conversion function?
-  robot_state::robotStateMsgToRobotState(state_msg->state, *robot_state_);
+  try
+  {
+    robot_state::robotStateMsgToRobotState(state_msg->state, *robot_state_);
+  }
+  catch (const moveit::Exception& e)
+  {
+    ROS_WARN_ONCE("moveit::Exception %s", e.what());
+  }
   setRobotHighlights(state_msg->highlight_links);
   update_state_ = true;
 }
--- a/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
+++ b/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
@@ -307,16 +307,17 @@ void RobotStateDisplay::newRobotStateCallback(const moveit_msgs::DisplayRobotSta
   try
   {
     robot_state::robotStateMsgToRobotState(state_msg->state, *robot_state_);
-    setStatus(rviz::StatusProperty::Ok, "RobotState", "");
     setRobotHighlights(state_msg->highlight_links);
-    update_state_ = true;
+    setStatus(rviz::StatusProperty::Ok, "RobotState", "");
   }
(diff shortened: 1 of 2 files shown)


moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
```diff
@@ -303,8 +303,15 @@ void RobotStateDisplay::newRobotStateCallback(const moveit_msgs::DisplayRobotSta
...
-  // possibly use TF to construct a robot_state::Transforms object to pass in to the conversion functio?
-  robot_state::robotStateMsgToRobotState(state_msg->state, *robot_state_);
+  // possibly use TF to construct a robot_state::Transforms object to pass in to the conversion function?
+  try
+  {
+    robot_state::robotStateMsgToRobotState(state_msg->state, *robot_state_);
+  }
+  catch (const moveit::Exception& e)
+  {
+    ROS_WARN_ONCE("moveit::Exception %s", e.what());
```
v4hn: Please use `setStatus(rviz::StatusProperty::Error, "RobotState", "[...]");`, so the problem shows in RViz too.

moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
```diff
@@ -303,8 +303,15 @@ void RobotStateDisplay::newRobotStateCallback(const moveit_msgs::DisplayRobotSta
...
+  }
   setRobotHighlights(state_msg->highlight_links);
```
v4hn: Does it make sense to run this line when the conversion throws an exception?
I'm don't know whether the underlying code can handle unknown links either.

christian-rauch: It looked like this was ignored by the rest of the code. Anyway, I set `update_state_ = false` and returned immediately, just in case.

christian-rauch:
@v4hn Done.

christian-rauch:
@rhaschke Thanks for pointing to the additional `ROS_ERROR_NAMED` message. I saw it on the terminal but couldn't track it down in the source.
I am fine with the changes.

moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
```diff
@@ -301,18 +301,18 @@ void RobotStateDisplay::newRobotStateCallback(const moveit_msgs::DisplayRobotSta
...
-  if (!robot_state_)
-    robot_state_.reset(new robot_state::RobotState(robot_model_));
+
   // possibly use TF to construct a robot_state::Transforms object to pass in to the conversion function?
   try
   {
-    robot_state::robotStateMsgToRobotState(state_msg->state, *robot_state_);
+    auto robot_state = std::make_shared<robot_state::RobotState>(robot_model_);
+    robot_state::robotStateMsgToRobotState(state_msg->state, *robot_state);
+    robot_state_.swap(robot_state);
```
v4hn: Yes, this makes sense. However, I preferred to skip the allocation because the failure case is very uncommon but the allocation is required on every update.
How about using `setToDefaultValues` to enforce a safe state in case of error?

rhaschke:
@v4hn Can you have a final look, please? If you approve, I would like to squash some of the commits for cleanup before finally merging.

v4hn:
Looks good now.

rhaschke:
Merged in 47ff403aa1717218254ed5f09245ff8ce9b669b6 after some cleanup squashing.

PR #1271, opened 2018-12-16
fix uninitialized RobotState transforms

rhaschke:
With #1096, 6cc3d62c7ae713413d8df7d4cf9ff09c43b5b486, I slightly improved FK in RobotState, exploiting the fact that the last row of the transform matrix is always `[0 0 0 1]`, performing a `3x4` * `4x4` matrix multiplication. However, this improvement also requires an initialization of this row, once.
This PR fixes this and also improves the corresponding timing tests.

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -66,11 +66,7 @@ RobotState::RobotState(const RobotModelConstPtr& robot_model)
   , rng_(nullptr)
 {
   allocMemory();
-
-  // all transforms are dirty initially
-  const int nr_doubles_for_dirty_joint_transforms =
-      1 + robot_model_->getJointModelCount() / (sizeof(double) / sizeof(unsigned char));
-  memset(dirty_joint_transforms_, 1, sizeof(double) * nr_doubles_for_dirty_joint_transforms);
+  initTransforms();
 }
 
 RobotState::RobotState(const RobotState& other) : rng_(nullptr)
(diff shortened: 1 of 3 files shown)


moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -109,6 +109,12 @@ void RobotState::allocMemory()
   velocity_ = position_ + robot_model_->getVariableCount();
   // acceleration and effort share the memory (not both can be specified)
   effort_ = acceleration_ = velocity_ + robot_model_->getVariableCount();
+
+  // initialize last row of transformation matrices, which will not be modified by transform updates anymore
+  for (size_t i = 0, end = robot_model_->getJointModelCount() + robot_model_->getLinkModelCount() +
+                           robot_model_->getLinkGeometryCount();
+       i != end; ++i)
+    variable_joint_transforms_[i].makeAffine();
```
v4hn: By name, this does not belong in this function and it results in unnecessary assignments when copying updated RobotStates.

Could you make that a utility function and call it from [here](https://github.com/ros-planning/moveit/blob/06fcc877ce33db0d430082ae2d99cbec01b26625/moveit_core/robot_state/src/robot_state.cpp#L69) and [here](https://github.com/ros-planning/moveit/blob/06fcc877ce33db0d430082ae2d99cbec01b26625/moveit_core/robot_state/src/robot_state.cpp#L143)?

This can also include the duplicated [`all transforms are dirty initially](https://github.com/ros-planning/moveit/blob/06fcc877ce33db0d430082ae2d99cbec01b26625/moveit_core/robot_state/src/robot_state.cpp#L70) code.

moveit_core/robot_state/test/robot_state_benchmark.cpp
```diff
@@ -37,69 +37,124 @@
...
   {
   }
 
-  robot_model::RobotModelPtr model_;
+public:
+  const Eigen::Isometry3d id = Eigen::Isometry3d::Identity();
+  // put transforms into a vector to avoid compiler optimization on variables
+  EigenSTL::vector_Isometry3d trafos;
+  volatile size_t res = 0;
+  volatile size_t inp = 1;
```
v4hn: How about readable variable names? 
`transforms`, `result_idx` and `input_idx`?

felixvd: Very in favor

v4hn:
Merged, thanks!

I hate to say it, but this is another critical bugfix from my perspective.
Can we push out another fixup release before Christmas?

rhaschke:
> Can we push out another fixup release before Christmas?

It's already scheduled: https://github.com/ros-planning/moveit/issues/1225#issuecomment-447792332

PR #1290, opened 2018-12-31
kinematics: cleanup cmake

rhaschke:
Simple fixes to cmake:
* package dependencies are required
* remove `link_directories`
* remove unnecessary, hard-coded linking of `libmoveit_rdf_loader`

### DIFF
--- a/moveit_kinematics/CMakeLists.txt
+++ b/moveit_kinematics/CMakeLists.txt
@@ -7,8 +7,9 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
-find_package(Boost)
-find_package(orocos_kdl)
+find_package(Boost REQUIRED)
+find_package(Eigen3 REQUIRED)
+find_package(orocos_kdl REQUIRED)
 find_package(catkin REQUIRED COMPONENTS
   moveit_core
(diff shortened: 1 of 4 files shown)


moveit_kinematics/CMakeLists.txt
```diff
@@ -7,8 +7,9 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
-find_package(Boost)
-find_package(orocos_kdl)
+find_package(Boost REQUIRED)
+find_package(Eigen3 REQUIRED)
+find_package(orocos_kdl REQUIRED)
```
IanTheEngineer: by finding these three as system deps, I believe you need to add them after the catkin_package DEPENDS clause, if they are used in any of the public-facing headerfiles:
```
catkin_package(
...
DEPENDS
  Boost
  Eigen3
  orocos_kdl
)
```

rhaschke: Probably you are right. This was out-of-the-scope of this PR.

rhaschke: Actually, it wasn't like this, because the plugins don't expose their headers publicly.

IanTheEngineer: Ah, then that makes sense! Thanks for the clarification.

PR #1291, opened 2019-01-02
RobotInteraction: use a timeout of 0.1s for a more interactive behaviour

rhaschke:
The [default IK timeout is 1.0s](https://github.com/ros-planning/moveit/blob/4d25c3a8e9de614afa28b234dcc6c83648bc47f4/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h#L150), which is too long for an interactive behaviour of RobotInteraction in rviz. This PR reduces the timeout to 100ms.


moveit_ros/robot_interaction/src/kinematic_options.cpp
```diff
@@ -55,7 +55,9 @@ bool robot_interaction::KinematicOptions::setStateFromIK(robot_state::RobotState
     ROS_ERROR("No getJointModelGroup('%s') found", group.c_str());
     return false;
   }
-  bool result = state.setFromIK(jmg, pose, tip, max_attempts_, timeout_seconds_, state_validity_callback_, options_);
+  bool result = state.setFromIK(jmg, pose, tip, max_attempts_,
+                                timeout_seconds_ > 0.0 ? timeout_seconds_ : std::min(0.1, jmg->getDefaultIKTimeout()),
```
IanTheEngineer: It strikes me as odd that there are two minimum values in this statement - even though one is intended to be default. If I am reading it correctly, the `JointModelGroup`'s default IK timeout can't be below `0.1` sec, but a time between `(0.0, 0.1]` is possible if the `timeout_seconds_` variable is set to be inside that range. 

Is that intentional, or should the conditional be checking `timeout_seconds_ > 0.1`, with the constructor defaulting to `0.1`?

```
: timeout_seconds_(0.1) // 0.1 = use default timeout
```

PR #1293, opened 2019-01-03
computeCartesianPath: limit joint-space jumps with IK consistency limits

rhaschke:
To improve quality of Cartesian paths, use joint-space jump limits (if provided) as consistency limits for the IK solver.


rhaschke:
> I like the use of kinematics base here

What do you mean? setFromIK is a method of RobotState.

mlautman:
The `consistency_limit` gets passed through to `solver->searchPositionIK` which then enforces the limit no?

rhaschke:
Yes, that's true.

PR #1298, opened 2019-01-09
Added missing robot state update to iterative spline parameterization

Martin-Oehler:
### Description
The iterative spline parameterization is missing an update to the robot states of the trajectory after iterating over each point. These leads to lots of warnings of the following kind:
```
[ WARN] [1547043223.679662634, 8.816000000] [/move_group]: Returning dirty collision body transforms
```
This PR adds the missing `update()` call (just one line).

I did not auto-format with clang as this changed lines unrelated to my PR. 

### Checklist

### DIFF
--- a/moveit_core/trajectory_processing/src/iterative_spline_parameterization.cpp
+++ b/moveit_core/trajectory_processing/src/iterative_spline_parameterization.cpp
@@ -340,11 +340,19 @@ bool IterativeSplineParameterization::computeTimeStamps(robot_trajectory::RobotT
   {
     for (unsigned int j = 0; j < num_joints; j++)
     {
-      trajectory.getWayPointPtr(i)->setVariablePosition(idx[j], t2[j].positions[i]);
       trajectory.getWayPointPtr(i)->setVariableVelocity(idx[j], t2[j].velocities[i]);
       trajectory.getWayPointPtr(i)->setVariableAcceleration(idx[j], t2[j].accelerations[i]);
     }
-    trajectory.getWayPointPtr(i)->update();
+
+    // Only update position of additional points
+    if (add_points_ && (i == 1 || i == num_points-2))
+    {
+      for (unsigned int j = 0; j < num_joints; j++)
+      {
+        trajectory.getWayPointPtr(i)->setVariablePosition(idx[j], t2[j].positions[i]);
+      }
+      trajectory.getWayPointPtr(i)->update();
+    }
   }
 
   return true;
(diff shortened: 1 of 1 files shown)


Martin-Oehler:
For some reason, the trajectory is first written to an internal data structure before doing the time parameterization. In this final step, the result is written back to the `robot_trajectory::RobotTrajectory` object.

rhaschke:
Yes, I have seen this. The question is why? The joint positions shouldn't be modified, as this might invalidate the trajectory.

Martin-Oehler:
Without looking further into it, I have only two guesses:
1. Performance: The data is restructured, maybe for faster access
2. The code has been copied from somewhere else and this was the fastest way to make it run.

v4hn:
Updating all states can be quite a bit of unnecessary computation here and I would prefer to avoid it where possible.

Looking at the code, it [conditionally includes two new waypoints if `add_points` is set to true](https://github.com/ros-planning/moveit/blob/34570a12c17b6b7fca3d58131a0a07189d0164d9/moveit_core/trajectory_processing/src/iterative_spline_parameterization.cpp#L133) and the code only ever modifies the positions of these two points.

So how about we skip the [`setVariablePosition`](https://github.com/ros-planning/moveit/blob/34570a12c17b6b7fca3d58131a0a07189d0164d9/moveit_core/trajectory_processing/src/iterative_spline_parameterization.cpp#L343) calls and instead do them for the two new points only if `add_points_` is true (and afterwards `update` them too?

rhaschke:
Thanks for looking deeper into this, @v4hn. I would agree to your proposal.

Martin-Oehler:
I agree as well and will update my PR accordingly (or should I open a new one?).

rhaschke:
Updating should be fine. You can also squash commits to cleanup history ;-)

Martin-Oehler:
I implemented the requested changes.

moveit_core/trajectory_processing/src/iterative_spline_parameterization.cpp
```diff
@@ -340,10 +340,19 @@ bool IterativeSplineParameterization::computeTimeStamps(robot_trajectory::RobotT
   {
     for (unsigned int j = 0; j < num_joints; j++)
     {
-      trajectory.getWayPointPtr(i)->setVariablePosition(idx[j], t2[j].positions[i]);
       trajectory.getWayPointPtr(i)->setVariableVelocity(idx[j], t2[j].velocities[i]);
       trajectory.getWayPointPtr(i)->setVariableAcceleration(idx[j], t2[j].accelerations[i]);
     }
+
+    // Only update position of additional points
```
rhaschke:
```suggestion
    // Only update position of additionally inserted points (at second and next-to-last position)
```

PR #1321, opened 2019-01-23
KDL IK solver improvements

rhaschke:
This PR addresses #1255, a regression of KDL-IK performance in Melodic. I didn't find any particular reason for the regression, but I noticed the following general issues:
* The IK solver was often stuck when running into joint limits (wasting most cycles trying to move outside the limits and being reset inside again).
* Singularities were not handled gracefully.

These issues are addressed as follows:
* Decrease step size when error-to-target has increased since last iteration (which is usually also an indicator for a singularity).
* When getting stuck passing a singularity, try to (locally!) wiggle all joints or bail out if target cannot be reached, thus not wasting more cycles to reach an unreachable goal.
* When joint values are clipped back to allowed range, reduce their motion contribution for the next cycle by using a weighted Jacobian. This allows the solver to find an alternative solution in the redundant space (if there is one). If there is no progress due to joint clipping, bail out.
* The outer loop, might reach the goal from another randomly chosen seed.

The solver is now able to pass a singularity without re-seeding (and thus having large joint-space jumps):
singular

### DIFF
--- a/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
+++ b/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
@@ -420,7 +420,14 @@ int KDLKinematicsPlugin::CartToJnt(KDL::ChainIkSolverVel& ik_solver, const KDL::
                    delta_q_norm);
 
     if (delta_q_norm < epsilon_)  // stuck in singularity
-      break;
+    {
+      if (step_size < 0.005)  // cannot reach target
+        break;
+      // wiggle joints
+      last_delta_twist_norm = DBL_MAX;
+      delta_q.data.setRandom();
+      delta_q.data *= std::min(0.1, delta_twist_norm);
+    }
 
     KDL::Add(q_out, delta_q, q_out);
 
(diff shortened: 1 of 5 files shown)


moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
```diff
@@ -367,58 +433,102 @@ bool KDLKinematicsPlugin::searchPositionIK(const geometry_msgs::Pose& ik_pose, c
...
-        break;
+      success = true;
+      break;
+    }
+
+    if (delta_twist_norm >= last_delta_twist_norm)
+    {
+      // if the error increased, we are close to a singularity -> reduce step size
+      double old_step_size = step_size;
+      step_size *= std::min(0.5, last_delta_twist_norm / delta_twist_norm);  // reduce scale;
```
henningkayser: I figured using a lower min value (i.e. 0.2) reduces the time being stuck at singularities.
This can be related to the setup, but is there another reason why this shouldn't be lower?

rhaschke: This is a rather uncritical value. Choosing it too high, simply requires several iterations to actually decrease the error. Choosing it too small, might unnecessarily slow down convergence.
Everything staying in the same order of magnitude is fine. I will try myself on our robot zoo and then merge.

jonbinney:
Do we have any tests that capture the effect of this change quantitatively? (e.g. before we were able to solve for X% of tool poses, but now we can solve for Y%)?

rhaschke:
@jonbinney  It's hard to devise meaningful tests for this. Your `kinematic_tests` approach, using random Cartesian goals is not very expressive, because it's prone to random bias. What we would need is a deterministic sampling of the whole workspace.

PR #1335, opened 2019-02-01
invert waypoint velocities on reverse

v4hn:
This addresses a long-standing issue in the trajectory container.

Without this patch, the trajectory's time profile is broken after calling `reverse()`
and robot controllers that respect velocities in specified trajectories,
e.g. default PR2 and TIAGo, will generate execution trajectories with highly oscillating velocity-profile.

This should be cherry-picked to `melodic-devel` of course.

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -516,6 +516,9 @@ public:
     return effort_[index];
   }
 
+  /** \brief Invert velocity if present. */
+  void invertVelocity();
+
   /** @} */
 
   /** \name Getting and setting joint positions, velocities, accelerations and effort
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -422,6 +422,15 @@ void RobotState::setVariableEffort(const std::vector<std::string>& variable_name
     effort_[robot_model_->getVariableIndex(variable_names[i])] = variable_effort[i];
 }
 
+void RobotState::invertVelocity()
+{
+  if (has_velocity_)
+  {
+    for (size_t i = 0; i < robot_model_->getVariableCount(); ++i)
+      velocity_[i] *= -1;
+  }
+}
+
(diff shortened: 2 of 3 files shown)


v4hn:
> While inversing the velocities is fine, why don't you also inverse the accelerations?
> They also change sign when reversing time.

Accelerations change sign *twice* on reversal because we reverse the trajectory and also the velocity.
Thus, the sign is actually good as is.

[Here](https://gist.github.com/v4hn/d822638d954eff9ea0dcd5e41be95c4c) is a small test that reverses a valid trajectory. With the proposed patch this is the output on our fanuc configuration:

reversed_trajectory

Without the patch the velocity is not inverted for `traj_reversed`.

rhaschke:
@v4hn While I was tempting to believe you on phone, from the figures I conclude that acceleration should change sign as well (and in your figure it actually does).
Turning a bell-shaped velocity profile (+acc -> -acc) into an inverted bell-shaped velocity profile, also means to start with negative acceleration (to reach negative velocity) and finish with positive (to bring velocity back to zero from negative).

v4hn:
> @v4hn While I was tempting to believe you on phone, from the figures I conclude that acceleration should change sign as well (and in your figure it actually does).

The figure was generated for the normal and inverted trajectory with the proposed pull-request *as-is*.
Do you think the trajectory profile in the screenshot looks bad?
The acceleration profile for `traj_reversed` *looks* like the sign was inverted, but actually only the waypoints are reversed (along the time axis).

rhaschke:
You (and your colleague) are right. Let's assume the trajectory is given as `p(t)`, where t runs from 0 to `tmax`. The reversed trajectory then is `q(t)=p(s(t))`, where `s(t)=tmax-t`. The velocity and acceleration profiles are then given by the derivatives using chain rule as follows:
```
v(t)=dq/dt=dp/ds * ds/dt
a(t)=dv/dt=dv/ds * ds/dt
```
The term `ds/dt = -1` creates the required sign change - once for velocity, twice for acceleration, three times for jerk, etc.

v4hn:
> three times for jerk

Yes, if we ever want to extend the representation to jerk, we will have to account for it here.

Thanks for the reviews and thanks for spelling out the formalism @rhaschke :+1:

PR #1338, opened 2019-02-04
Travis: enable clang-tidy-fix

rhaschke:
In https://github.com/ros-planning/moveit/pull/1185 @BryceStevenWilley  introduced some naming conventions, I don't agree with, namely enforcing UPPER_CASE for all constant variables. I think we agreed, that UPPER_CASE is reserved for macros, maybe static const variables as well. However, const members should be lower_case as other variables as well.

This PR also removes the very time-consuming code analysis checks (clang-analyzer-*). They are nice to have locally, but they don't provided automatic fixes and thus are not useful on Travis.

Some remarks on the next steps can be found in https://github.com/ros-planning/moveit/issues/28#issuecomment-451624912.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,6 +1,5 @@
 ---
 Checks:          '-*,
-                  clang-analyzer-*,
                   performance-*,
                   modernize-redundant-void-arg,
                   modernize-use-nullptr,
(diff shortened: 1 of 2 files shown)


.clang-tidy
```diff
@@ -22,30 +21,28 @@ CheckOptions:
...
+  # method names
+  - key:             readability-identifier-naming.MethodCase
+    value:           camelBack
+  # variable names
   - key:             readability-identifier-naming.VariableCase
     value:           lower_case
-  - key:             readability-identifier-naming.ConstantCase
+  - key:             readability-identifier-naming.ClassMemberSuffix
+    value:           '_'
+  # const static or global variables are UPPER_CASE
```
jonbinney: Thanks for adding the comments in this file - it makes it a lot easier to understand what's going on!

jonbinney: Could you move this constant down lower though? It refers to lines 42-47, right?

.travis.yml
```diff
@@ -32,8 +33,10 @@ matrix: # Add a separate config to the matrix, using clang as compiler
       env: ROS_REPO=ros-shadow-fixed
            BEFORE_DOCKER_SCRIPT="source moveit_kinematics/test/test_ikfast_plugins.sh"
 
+  fast_finish: true  # finish, even if allow-failure-tests still running
   allow_failures:
     - env: TEST=catkin_lint  # need to fix catkin_lint errors
+    - env: TEST=clang-tidy-fix  # need to fix clang-tidy issues
```
jonbinney: Having this finishing being optional sounds fine to me. You mentioned that it takes some time to run - what's the order of magnitude on that? Is it seconds? Minutes?

rhaschke: In the [last build](https://travis-ci.org/ros-planning/moveit/builds/488453595?utm_source=github_status&utm_medium=notification), clang-tidy-fix ran for additional 40min. Currently, it produces so many warnings, that Travis refuses to display the whole log.

.clang-tidy
jonbinney: Sorry, typo on my part. Meant to say "this comment" - the comment on line 39 is describing lines 42-47, right?

.travis.yml
jonbinney: Wow.... definitely support making this optional then. Thanks.

.clang-tidy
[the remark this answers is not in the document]
jonbinney: :+1:

rhaschke:
@BryceStevenWilley @davetcoleman @v4hn Would be great to get some feedback from you on this.

.travis.yml
rhaschke: Me too :wink:

PR #1343, opened 2019-02-12
remove catkin_lint suppressions

rhaschke:
https://github.com/fkie/catkin_lint/issues/62 has been resolved and catkin_lint 1.5.6 has been released.
Thus, we can remove some catkin_lint suppressions again.
- [ ] We need to wait until this new release will become available in the standard ROS repositories (https://github.com/ros/rosdistro/issues/20283)

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -1,9 +1,6 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(moveit_core)
 
-#suppress spurious errors: https://github.com/fkie/catkin_lint/issues/62
-#catkin_lint: ignore duplicate_find  (find_package(moveit_resources))
-
 add_compile_options(-std=c++14)
 
 # Warnings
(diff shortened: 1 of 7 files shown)


moveit_core/CMakeLists.txt
```diff
@@ -162,8 +157,7 @@ include_directories(SYSTEM ${EIGEN3_INCLUDE_DIRS}
                            ${LIBFCL_INCLUDE_DIRS}
                            )
 
-#suppress spurious error: https://github.com/fkie/catkin_lint/issues/62
-#catkin_lint: ignore_once external_directory missing_directory  (${VERSION_FILE_PATH})
+#catkin_lint: ignore_once external_directory  (${VERSION_FILE_PATH})
```
jonbinney: It looks like you removed the comment explaining and linking to the github issue, but kept the ignore_once - is that what you intended to do?

moveit_core/CMakeLists.txt
```diff
@@ -25,12 +22,12 @@ endif()
 
 find_package(Boost REQUIRED system filesystem date_time thread iostreams)
 find_package(Eigen3 REQUIRED)
-find_package(PkgConfig REQUIRED)
 
-pkg_check_modules(LIBFCL REQUIRED fcl)
-# replace LIBFCL_LIBRARIES with full path to the library
-find_library(LIBFCL_LIBRARIES_FULL ${LIBFCL_LIBRARIES} ${LIBFCL_LIBRARY_DIRS})
-set(LIBFCL_LIBRARIES "${LIBFCL_LIBRARIES_FULL}")
+include(FindPkgConfig)
```
jonbinney: This change is a bit different than the rest of this PR - the rest is removing suppresions for caktin_lint; this appears to be updating the way we use pkg_config from cmake. Seems innocuous though, so I'm happy to have it in this PR.

moveit_core/CMakeLists.txt
jonbinney: (or was that issue only for the `missing_directory`)?

rhaschke: Yes, the issue was related to `missing_directiory`. The `external_directoring` _warning_ is legitimate but should be suppressed here.

moveit_core/CMakeLists.txt
rhaschke: This is related to this PR, because it successfully suppresses another catkin_lint warning as suggested in https://github.com/fkie/catkin_lint/issues/62#issuecomment-462695235.

roehling: You should keep the original `find_package(PkgConfig REQUIRED)` though. It's the recommended way of using CMake find modules.

rhaschke: Thanks for the hint. I thought `include(FindPkgConfig)` and `find_package(PkgConfig REQUIRED)` are synonymous.

PR #1370, opened 2019-03-02
Fcl as a Plugin

BryceStevenWilley:
### Description

I mentioned at the maintainer's meeting that I had once made the FCL collision detector a plugin, as opposed to its currently hardcoded state. This is the simple change to make that happen, and gets us one step closer to being able to easily use different collision detectors, like [Bullet](https://github.com/bulletphysics/bullet3), in MoveIt.

To actually make this affect anything, you have to set the `collision_detector` param of the move group. In the generated `move_group.launch`, you can add 
```
<param name="collision_detector" value="FCL"/>
``` 
right after `jiggle_fraction` to use it. Then the plugin loader loads it [as seen here](
https://github.com/ros-planning/moveit/blob/822a991e0ea4f713fe672e3e337d7f0702bcc224/moveit_ros/planning/collision_plugin_loader/src/collision_plugin_loader.cpp#L116).
(This doesn't really make a difference, as the planning scene always [loads the FCL plugin by default](https://github.com/BryceStevenWilley/moveit/blob/d1b2fcf4e5622212d620ecde9f2935eed507a7e0/moveit_core/planning_scene/src/planning_scene.cpp#L173), but it's a step.)

Future changes should eventually make the collision detector a parameter in all of the generated `move_group` launch files. If desired, I could do that now, but it would take a while for me to figure out the MSA.

### Checklist

### DIFF
--- a/moveit_core/collision_detection_fcl/CMakeLists.txt
+++ b/moveit_core/collision_detection_fcl/CMakeLists.txt
@@ -11,7 +11,7 @@ target_link_libraries(${MOVEIT_LIB_NAME} moveit_collision_detection ${catkin_LIB
 add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
 
 add_library(collision_detector_fcl_plugin src/collision_detector_fcl_plugin_loader.cpp)
-set_target_properties(collision_detector_fcl_plugin PROPERTIES VERSION ${${PROJECT_NAME}_VERSION})
+set_target_properties(collision_detector_fcl_plugin PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 target_link_libraries(collision_detector_fcl_plugin ${catkin_LIBRARIES} ${MOVEIT_LIB_NAME})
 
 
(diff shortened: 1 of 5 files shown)


moveit_core/collision_detection_fcl/CMakeLists.txt
```diff
@@ -10,7 +10,12 @@ set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_V
 target_link_libraries(${MOVEIT_LIB_NAME} moveit_collision_detection ${catkin_LIBRARIES} ${urdfdom_LIBRARIES} ${urdfdom_headers_LIBRARIES} ${LIBFCL_LIBRARIES} ${Boost_LIBRARIES})
 add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
 
-install(TARGETS ${MOVEIT_LIB_NAME}
+add_library(collision_detector_fcl_plugin src/collision_detector_fcl_plugin_loader.cpp)
+set_target_properties(collision_detector_fcl_plugin PROPERTIES VERSION ${${PROJECT_NAME}_VERSION})
+target_link_libraries(collision_detector_fcl_plugin ${catkin_LIBRARIES} ${MOVEIT_LIB_NAME})
+
+
+install(TARGETS ${MOVEIT_LIB_NAME} collision_detector_fcl_plugin
```
jrgnicho: Your cmakelist is missing the install command for the plugin xml.  See [here](https://github.com/ros-industrial/industrial_moveit/blob/melodic-devel/industrial_collision_detection/CMakeLists.txt#L68) for an example

BryceStevenWilley: Thanks for the tip, added.

rhaschke:
@jrgnicho, thanks for merging. Please consider [our merge policy](https://moveit.ros.org//documentation/contributing/pullrequests) next time: The 3 commits should have been squashed-merged, because the additional commits are just cleanups of the first one.

PR #1376, opened 2019-03-04
more clang-tidy-fix

rhaschke:
Additionally to #1366, clang-tidy 6.0 finds more issues to resolve. 
@RoboticsYY, @BryceStevenWilley please have a look.


moveit_core/collision_distance_field/src/collision_common_distance_field.cpp
```diff
@@ -151,7 +150,7 @@ void getBodySphereVisualizationMarkers(GroupStateRepresentationConstPtr& gsr, st
 
   // creating sphere marker
   visualization_msgs::Marker sphere_marker;
-  sphere_marker.header.frame_id = std::move(reference_frame);
```
jonbinney: Is the `std::move` being removed because it is redundant? Or is there some other reason?

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -1322,7 +1322,7 @@ operator=(MoveGroupInterface&& other)
   if (this != &other)
   {
     delete impl_;
-    impl_ = std::move(other.impl_);
```
jonbinney: I'm still not clear on when `std::move` should be used

rhaschke:
`std::move` is used when you don't need the argument anymore and you agree to reuse it's internal memory. For plain datatypes, there is no benefit in using `std::move`.

There are two possibilities to avoid copying of arguments: Passing (const) references avoids creating copies in the first place, while using `std::move` within the cpp definition allows to reuse the memory and thus avoids a late copy.

PR #1377, opened 2019-03-04
WIP: fix chomp issues

rhaschke:
This is an attempt to fix various chomp issues.
* 1st one is to fix a segfault when the main planner fails. In this case `res.trajectory_` is `NULL`.

Open issues:
* plan + exec behaves differently (fails) than separate plan and execute (when triggered from rviz plugin)
  When running plan+exec, the initial RobotState seems to be undefined, while it is defined otherwise.


bmagyar:
I take this is not yet ready to review, right? The changes so far look good!

rhaschke:
@bmagyar I hope we / you? can fix more of the chomp issues before releasing. That's why I marked this as wip.

PR #1380, opened 2019-03-06
fix more warnings

rhaschke:
Having the warnings check enabled in Travis and motivated by [warnings on gentoo](https://answers.ros.org/question/317202/moveit-segfaults-due-to-eigen/?answer=317607#post-id-317607), I increased the warning level for the whole build (previously we only used extended warnings for moveit_core) and fixed the corresponding warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
     - ROS_REPO=ros
     - UPSTREAM_WORKSPACE=moveit.rosinstall
     - TEST_BLACKLIST=moveit_ros_perception  # mesh_filter_test fails due to broken Mesa OpenGL
+    - CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unused-parameter -Wno-unused-but-set-parameter -Wno-unused-function"
     - WARNINGS_OK=false
   matrix:
     - TEST="clang-format catkin_lint"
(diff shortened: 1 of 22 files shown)


moveit_planners/chomp/chomp_motion_planner/src/chomp_optimizer.cpp
```diff
@@ -126,10 +126,8 @@ void ChompOptimizer::initialize()
   const std::vector<const moveit::core::JointModel*> joint_models = joint_model_group_->getActiveJointModels();
   for (size_t i = 0; i < joint_models.size(); i++)
   {
-    const moveit::core::JointModel* model = joint_models[i];
     double joint_cost = 1.0;
-    const std::string& joint_name = model->getName();
-    // nh.param("joint_costs/" + joint_name, joint_cost, 1.0);
+    // nh.param("joint_costs/" + joint_models[i]->getName(), joint_cost, 1.0);
```
henningkayser: Unrelated, but this could be removed then as well.

.travis.yml
```diff
@@ -18,6 +18,7 @@ env:
     - ROS_REPO=ros
     - UPSTREAM_WORKSPACE=moveit.rosinstall
     - TEST_BLACKLIST=moveit_ros_perception  # mesh_filter_test fails due to broken Mesa OpenGL
+    - CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-unused-parameter -Wno-unused-but-set-parameter -Wno-unused-function"
```
henningkayser: Maybe add a linebreak. And is there a reason why `-Wno-overloaded-virtual` is not set here?

moveit_planners/chomp/chomp_motion_planner/src/chomp_optimizer.cpp
rhaschke: I didn't want to throw away code.

.travis.yml
rhaschke: I'm not sure cmake can handle a linebreak in the CXXFLAGS variable.

PR #1381, opened 2019-03-06
Reduce spurious unittest failures due to randomness

rhaschke:
Kinematics unittests use random sampling, which might (randomly) cause the tests to fail.
This fix reduces randomness by using a fixed seed to sample robot states.

Additionally, the "searchIKWithCallback" test should resample dismissed samples as pointed out
in https://github.com/ros-planning/moveit_kinematics_tests/pull/16.


rhaschke:
> I thought finding rare failures is the point of testing with randomness.

No. We are just lazy and don't want to provide numerous samples manually.
A unit test should validate that code that was working at a certain time, continues to work in the future.

Issue #1386, opened 2019-03-08
Hardcoded file locations in loadTestingRobotModel


mlautman:
### Description

`moveit::core::loadTestingRobotModel("panda_description");` fails silently creating a test log file (eg: `MISSING-gtest-test_cartesian_interpolator.xml`). This is because the method hardcodes the srdf file location to work for `moveit_resources/pr2_description` but not franka or panda.

Overview of your issue here.

### Your environment
* ROS Distro: [melodic]
* OS Version: e.g. Ubuntu 18.04
* Source or Binary build?
* If binary, which release version?
* If source, which git commit or tag?

### Steps to reproduce
Change the `robot_name` in `test_aabb.cpp` to `panda_description`

rhaschke:
@BryceStevenWilley, could you please look into this issue as you initially wrote this code?
I think we need a kind of switch-case for the three robots (pr2, panda, fanuc) to correctly choose paths.

BryceStevenWilley:
Acknowledged, I'll try to get to this this week.  From what I can tell, the issue is the fact that the PR2's SRDF is in the `pr2_description` [here](https://github.com/ros-planning/moveit_resources/tree/master/pr2_description) but Panda's SRDF is in the `panda_moveit_config` [here](https://github.com/ros-planning/moveit_resources/tree/master/panda_moveit_config/config) and not `panda_description`. The best option seems to be keeping PR2 as a hard coded non-standard srdf location, but using Panda's location as the default, so this is future proof too.

BryceStevenWilley:
@mlautman With #1391 merged this should be fixed: changing `pr2` to `panda` no longer crashes when loading the URDF/SRDF. It still segfaults (fails silently) since the tests are hardcoded to use PR2 link names, but I'm guessing that's what you were working on.

rhaschke:
Closes via #1391

PR #1391, opened 2019-03-12
Fixed to handle normal robot urdf paths

BryceStevenWilley:
### Description

Addresses #1386. 

Fixes the fact robot URDF and SRDF paths were hardcoded to the non-standard PR2 locations. Now uses the PR2 as a non-standard if case, and uses the newer normal locations of URDF in `<robot>_description/urdf/<robot>.urdf` and SRDF in `<robot>_moveit_config/config/<robot>.srdf`.
 
### Checklist

### DIFF
--- a/moveit_core/utils/include/moveit/utils/robot_model_test_utils.h
+++ b/moveit_core/utils/include/moveit/utils/robot_model_test_utils.h
@@ -54,23 +54,26 @@ namespace moveit
 namespace core
 {
 /** \brief Loads a robot from moveit_resources.
- * \param[in] robot_name The name of the robot package in moveit_resources to load.
- *            For example, "panda_description", or "fanuc_description".
+ * \param[in] robot_name The name of the robot in moveit_resources to load.
+ *            This should be the prefix to many of the robot packages.
+ *            For example, "pr2", "panda", or "fanuc".
  * \returns a RobotModel constructed from robot_name's URDF and SRDF.
  */
 moveit::core::RobotModelPtr loadTestingRobotModel(const std::string& robot_name);
 
 /** \brief Loads a URDF Model Interface from moveit_resources.
- * \param[in] robot_name The name of the robot package in moveit_resources to load.
- *            For example, "panda_description", or "fanuc_description".
+ * \param[in] robot_name The name of the robot in moveit_resources to load.
+ *            This should be the prefix to many of the robot packages.
+ *            For example, "panda", or "fanuc".
  * \returns a ModelInterface constructed from robot_name's URDF.
  */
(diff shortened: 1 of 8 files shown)


moveit_core/utils/include/moveit/utils/robot_model_test_utils.h
```diff
@@ -54,23 +54,26 @@ namespace moveit
...
  * \returns a RobotModel constructed from robot_name's URDF and SRDF.
  */
 moveit::core::RobotModelPtr loadTestingRobotModel(const std::string& robot_name);
 
 /** \brief Loads a URDF Model Interface from moveit_resources.
- * \param[in] robot_name The name of the robot package in moveit_resources to load.
- *            For example, "panda_description", or "fanuc_description".
+ * \param[in] robot_name The name of the robot in moveit_resources to load.
+ *            This should be the prefix to many of the robot packages.
+ *            For example, "panda", or "fanuc".
```
bmagyar: no "pr2" here?

moveit_core/utils/src/robot_model_test_utils.cpp
```diff
@@ -57,7 +57,16 @@ moveit::core::RobotModelPtr loadTestingRobotModel(const std::string& robot_name)
 urdf::ModelInterfaceSharedPtr loadModelInterface(const std::string& robot_name)
 {
   boost::filesystem::path res_path(MOVEIT_TEST_RESOURCES_DIR);
-  urdf::ModelInterfaceSharedPtr urdf_model = urdf::parseURDFFile((res_path / robot_name / "urdf/robot.xml").string());
+  std::string urdf_path;
+  if (robot_name == "pr2")
+  {
+    urdf_path = (res_path / (robot_name + "_description") / "urdf/robot.xml").string();
```
bmagyar: If you don't intend to add other robots with the PR2-style urdf path, I'd even remove using `robot_name` here and go hardcoded on most of the path. The same below

moveit_core/utils/include/moveit/utils/robot_model_test_utils.h
BryceStevenWilley: Quick mistake, thanks for pointing that out.

moveit_core/utils/src/robot_model_test_utils.cpp
BryceStevenWilley: Good point, nothing else seems to have this exact layout. Changed.

BryceStevenWilley:
PTAL, the comments have been addressed.

rhaschke:
Thanks for addressing the comments so quickly. Merged.

PR #1407, opened 2019-03-19
small fixes to chomp planner

rhaschke:
Tracking down a `Found empty JointState message` from the chomp planner, I stumbled over some smaller issues. This PR
- introduces a (slightly) more verbose variable name to distinguish a MotionPlanResponse (res) from its message counterpart
- removes explicit time parameterization from the chomp planner (time parameterization is configured in the planning pipeline)
- adds a TODO to switch the API of `ChompPlanner::solve` to use real planning_interface datatypes (instead of their moveit_msgs counterparts). This will save some unnecessary serializations. As far as I can see, both usages in chomp_optimizer_adapter and chomp_interface, actually use native types originally. @bmagyar, what do you think of this idea?
Fixing this, will finally resolve the error message, as this one comes from deserializing an empty RobotState msg.

### DIFF
--- a/moveit_planners/chomp/chomp_interface/src/chomp_planning_context.cpp
+++ b/moveit_planners/chomp/chomp_interface/src/chomp_planning_context.cpp
@@ -22,29 +22,29 @@ CHOMPPlanningContext::~CHOMPPlanningContext() = default;
 
 bool CHOMPPlanningContext::solve(planning_interface::MotionPlanDetailedResponse& res)
 {
-  moveit_msgs::MotionPlanDetailedResponse res2;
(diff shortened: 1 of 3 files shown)


rhaschke:
@bmagyar, what do you think about the idea to change the API of `ChompPlanner::solve`?

bmagyar:
While in general I support messages for simple data transfer even within a library, MoveIt is far beyond the use-cases where this is justified. I think the API change is a good idea and we should move forward with it for Melodic.

This raises a point of whether the public API of a planner plugin is considered public API from a MoveIt perspective or we draw a line at the plugin and action interfaces plus config files?

bmagyar:
Anything else to add or may I merge?

rhaschke:
Merged. I will tackle the API change in another PR.

PR #1419, opened 2019-03-27
Clang tidy fix `modernize-loop-convert` to entire code base

RoboticsYY:
### Description

This PR intends to enable `modernize-loop-convert` to the entire code base. It's an effort following up the discussion in [moveit#880](https://github.com/ros-planning/moveit/pull/880). Basicly, clang-tidy [modernize-loop-convert](https://clang.llvm.org/extra/clang-tidy/checks/modernize-loop-convert.html) converts traditional loops `for(...; ...; ...)` into modern range-based loops `for(auto ... : ...)`. Since `auto` is poorly readable, manual fixes are implemented to replace `auto` by explicitly specifying the datatype. 

Please note:
* For the exceptions, seems the iterator and operator[ ] of class `XmlRpcValue` returns different datatype, and modern loop will lookup for the begin() and end() member functions to determine the item datatype, which returns the same type with the iterator. Thus the instance of `XmlRpcValue` can only be used as a range container in a modern loop when it's type is `TypeStruct`. 

### Checklist


moveit_ros/manipulation/pick_place/src/manipulation_pipeline.cpp
```diff
@@ -110,30 +110,30 @@ void ManipulationPipeline::start()
...
 {
   signalStop();
-  for (std::size_t i = 0; i < processing_threads_.size(); ++i)
-    if (processing_threads_[i])
+  for (boost::thread* processing_thread : processing_threads_)
+    if (processing_thread)
     {
-      processing_threads_[i]->join();
-      delete processing_threads_[i];
-      processing_threads_[i] = nullptr;
```
BryceStevenWilley: This is the first time I've seen this, but by iterating over pointers, `processing_thread = nullptr;`doesn't correctly change the entry in the vector, it's only changing the copy of the pointer. Should probably just use `//NOLINT` here. 

I tried to look for other instances of this, but this is the main way where bugs can be introduced I believe.

moveit_ros/perception/lazy_free_space_updater/include/moveit/lazy_free_space_updater/lazy_free_space_updater.h
```diff
@@ -55,6 +56,8 @@ class LazyFreeSpaceUpdater
 private:
 #ifdef __APPLE__
   typedef std::unordered_map<octomap::OcTreeKey, unsigned int, octomap::OcTreeKey::KeyHash> OcTreeKeyCountMap;
+#elif __cplusplus >= 201103L
```
BryceStevenWilley: I don't have anything against this, but what's the reason you decided to make this change in this PR? Does `tr1::unordered_map` not iterate correctly?

moveit_ros/perception/lazy_free_space_updater/src/lazy_free_space_updater.cpp
```diff
@@ -133,40 +133,34 @@ void LazyFreeSpaceUpdater::processThread()
...
       {
         /* compute the free cells along each ray that ends at a model cell */
-        for (octomap::KeySet::iterator it = process_model_cells_set_->begin(), end = process_model_cells_set_->end();
-             it != end; ++it)
-          if (tree_->computeRayKeys(process_sensor_origin_, tree_->keyToCoord(*it), key_ray2))
-            for (octomap::KeyRay::iterator jt = key_ray2.begin(), end = key_ray2.end(); jt != end; ++jt)
-              free_cells2[*jt]++;
+        for (const octomap::OcTreeKey& it : *process_model_cells_set_)
+          if (tree_->computeRayKeys(process_sensor_origin_, tree_->keyToCoord(it), key_ray2))
+            for (auto& jt : key_ray2)
```
BryceStevenWilley: Missed an auto here.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -700,10 +700,11 @@ void PlanningSceneMonitor::includeAttachedBodiesInOctree()
   boost::recursive_mutex::scoped_lock _(shape_handles_lock_);
 
   // clear information about any attached body, without refering to the AttachedBody* ptr (could be invalid)
-  for (AttachedBodyShapeHandles::iterator it = attached_body_shape_handles_.begin();
-       it != attached_body_shape_handles_.end(); ++it)
-    for (std::size_t k = 0; k < it->second.size(); ++k)
-      octomap_monitor_->forgetShape(it->second[k].first);
+  for (std::pair<const robot_state::AttachedBody* const,
```
BryceStevenWilley: It can stay like this, but this is one of the few times that I'd prefer `auto`, when the type itself takes up two lines and the variable only lasts for three lines.

moveit_ros/robot_interaction/src/robot_interaction.cpp
```diff
@@ -329,15 +329,14 @@ void RobotInteraction::decideActiveEndEffectors(const std::string& group, Intera
...
+  for (EndEffectorInteraction& eef : active_eef_)
   {
     // if we have a separate group for the eef, we compute the scale based on it;
     // otherwise, we use the size of the parent_link
-    active_eef_[i].size = active_eef_[i].eef_group == active_eef_[i].parent_group ?
-                              computeLinkMarkerSize(active_eef_[i].parent_link) :
-                              computeGroupMarkerSize(active_eef_[i].eef_group);
-    ROS_DEBUG_NAMED("robot_interaction", "Found active end-effector '%s', of scale %lf",
-                    active_eef_[i].eef_group.c_str(), active_eef_[i].size);
+    eef.size = eef.eef_group == eef.parent_group ? computeLinkMarkerSize(eef.parent_link) :
```
BryceStevenWilley:
```suggestion
    eef.size = (eef.eef_group == eef.parent_group) ? computeLinkMarkerSize(eef.parent_link) :
```
Unrelated, but we might as well make this more readable.

moveit_ros/warehouse/warehouse/src/planning_scene_storage.cpp
```diff
@@ -266,9 +266,9 @@ void moveit_warehouse::PlanningSceneStorage::getPlanningQueriesNames(std::vector
   q->append(PLANNING_SCENE_ID_NAME, scene_name);
   std::vector<MotionPlanRequestWithMetadata> planning_queries = motion_plan_request_collection_->queryList(q, true);
   query_names.clear();
-  for (std::size_t i = 0; i < planning_queries.size(); ++i)
-    if (planning_queries[i]->lookupField(MOTION_PLAN_REQUEST_ID_NAME))
-      query_names.push_back(planning_queries[i]->lookupString(MOTION_PLAN_REQUEST_ID_NAME));
+  for (MotionPlanRequestWithMetadata& planning_querie : planning_queries)
```
BryceStevenWilley:
```suggestion
  for (MotionPlanRequestWithMetadata& planning_query : planning_queries)
```
Another -ies mishap.

moveit_ros/manipulation/pick_place/src/manipulation_pipeline.cpp
RoboticsYY: True, but I think maybe we don't need use `//NOLINT`, it would be enough if changing the range-declaration from pointer to pointer reference for this kind of cases.

moveit_ros/perception/lazy_free_space_updater/include/moveit/lazy_free_space_updater/lazy_free_space_updater.h
RoboticsYY: I came across `tr1::unordered_map` when I was looking for the declaration of the map. `tr1::unordered_map` iterates well. I just try to modernize it. But I'm not quite sure about the difference between `std::unordered_map` and `std::tr1::unordered_map`, so I would like to keep the old one.

moveit_core/kinematics_base/src/kinematics_base.cpp
```diff
@@ -128,9 +128,9 @@ bool KinematicsBase::initialize(const moveit::core::RobotModel& robot_model, con
 
 bool KinematicsBase::setRedundantJoints(const std::vector<unsigned int>& redundant_joint_indices)
 {
-  for (std::size_t i = 0; i < redundant_joint_indices.size(); ++i)
+  for (const unsigned int& redundant_joint_indice : redundant_joint_indices)
```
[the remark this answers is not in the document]
RoboticsYY: Truly, clang-tidy seems just remove `s` for many similar cases. I think I have missed this one.

moveit_ros/manipulation/pick_place/src/manipulation_pipeline.cpp
BryceStevenWilley: That's true, pointers references would work too.

moveit_core/profiler/src/profiler.cpp
```diff
@@ -258,14 +254,14 @@ void Profiler::printThreadInfo(std::ostream& out, const PerThread& data)
     out << "Blocks of time:" << std::endl;
 
   double unaccounted = total;
-  for (unsigned int i = 0; i < time.size(); ++i)
+  for (DataDoubleVal& it : time)
```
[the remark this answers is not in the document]
RoboticsYY: Agree, `time_block` is a better name. Corrected.

moveit_ros/perception/lazy_free_space_updater/src/lazy_free_space_updater.cpp
RoboticsYY: I missed here. Fixed.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
RoboticsYY: I have the same feeling. Maybe it's better to make some additional `typedef`s to simplify this.

moveit_ros/robot_interaction/src/robot_interaction.cpp
RoboticsYY: Agree, corrected.

moveit_ros/warehouse/warehouse/src/planning_scene_storage.cpp
RoboticsYY: I missed this too. Corrected.

PR #1420, opened 2019-03-27
Normalize quaternions when adding new or moving collision objects (#1119)

### LINKED ISSUE #1119: Normalize Quaternions
The PlanningScene class does not always correctly normalize quaternions in new collision objects.

Here, the unnormalized w-component is normalized correctly:
```
$ rostopic pub -1 /collision_object moveit_msgs/CollisionObject "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: 'world'
id: 'box'a
primitives:
- type: 1
  dimensions: [1.8,1.5,1.0]
primitive_poses:
- position: {x: 0.0, y: 0.0, z: 0.15}
  orientation: {x: 0.0, y: 0.0, z: 0.0, w: 2.0}"

$ rosservice call /get_planning_scene "components:
  components: 16" | grep box -A20
        id: "box"
        type:
          key: ''
          db: ''
        primitives:
          -
            type: 1
            dimensions: [1.8, 1.5, 1.0]
        primitive_poses:
          -
            position:
              x: 0.89
              y: 0.74
              z: 0.93
            orientation:
              x: 0.0
              y: 0.0
              z: 0.0
              w: 1.0
        meshes: []
        mesh_poses: []
```

But when another component is set, this suddenly fails:
```
$ rostopic pub -1 /collision_object moveit_msgs/CollisionObject "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: 'world'
id: 'box'
primitives:
- type: 1
  dimensions: [1.8,1.5,1.0]
primitive_poses:
- position: {x: 0.0, y: 0.0, z: 0.15}
  orientation: {x: 0.0, y: 15.0, z: 0.0, w: 2.0}"

$ rosservice call /get_planning_scene "components:
  components: 16" | grep box -A20
        id: "box"
        type:
          key: ''
          db: ''
        primitives:
          -
            type: 1
            dimensions: [1.8, 1.5, 1.0]
        primitive_poses:
          -
            position:
              x: 0.89
              y: 0.74
              z: 0.93
            orientation:
              x: 0.0
              y: 15.0
              z: 0.0
              w: 2.0
        meshes: []
        mesh_poses: []
```

j-petit:
Fixes #1119.

### Description
As proposed in #1119 I implemented quaternion normalization when adding a new collision object and also when providing a new pose (move operation). Instead of using `tf2::fromMsg` as before I wrote a new function which includes orientation normalization and replaced it in the appropriate spots.

### Checklist

### DIFF
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -982,7 +982,9 @@ private:
   bool processCollisionObjectAdd(const moveit_msgs::CollisionObject& object);
   bool processCollisionObjectRemove(const moveit_msgs::CollisionObject& object);
   bool processCollisionObjectMove(const moveit_msgs::CollisionObject& object);
-  void normalizeObjectOrientation(const geometry_msgs::Pose& msg, Eigen::Isometry3d& out);
+
+  /* convert Pose msg to Eigen::Isometry, normalizing the quaternion part if necessary. */
+  static void poseMsgToEigen(const geometry_msgs::Pose& msg, Eigen::Isometry3d& out);
 
   MOVEIT_STRUCT_FORWARD(CollisionDetector);
 
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -982,6 +982,7 @@ private:
   bool processCollisionObjectAdd(const moveit_msgs::CollisionObject& object);
   bool processCollisionObjectRemove(const moveit_msgs::CollisionObject& object);
   bool processCollisionObjectMove(const moveit_msgs::CollisionObject& object);
+  void normalizeObjectOrientation(const geometry_msgs::Pose& msg, Eigen::Isometry3d& out);
 
   MOVEIT_STRUCT_FORWARD(CollisionDetector);
 
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -983,7 +983,7 @@ private:
   bool processCollisionObjectRemove(const moveit_msgs::CollisionObject& object);
   bool processCollisionObjectMove(const moveit_msgs::CollisionObject& object);
 
-  /* convert Pose msg to Eigen::Isometry, normalizing the quaternion part if necessary. */
+  /** convert Pose msg to Eigen::Isometry, normalizing the quaternion part if necessary. */
   static void poseMsgToEigen(const geometry_msgs::Pose& msg, Eigen::Isometry3d& out);
 
   MOVEIT_STRUCT_FORWARD(CollisionDetector);
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1817,7 +1817,6 @@ bool PlanningScene::processCollisionObjectMove(const moveit_msgs::CollisionObjec
     for (std::size_t i = 0; i < object.primitive_poses.size(); ++i)
     {
       Eigen::Isometry3d object_pose;
-      tf2::fromMsg(object.primitive_poses[i], object_pose);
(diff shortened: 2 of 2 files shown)


moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -982,6 +982,7 @@ class PlanningScene : private boost::noncopyable, public std::enable_shared_from
   bool processCollisionObjectAdd(const moveit_msgs::CollisionObject& object);
   bool processCollisionObjectRemove(const moveit_msgs::CollisionObject& object);
   bool processCollisionObjectMove(const moveit_msgs::CollisionObject& object);
+  void normalizeObjectOrientation(const geometry_msgs::Pose& msg, Eigen::Isometry3d& out);
```
rhaschke: I suggest a different name:
```suggestion
  /// convert Pose msg to Eigen::Isometry, normalizing the quaternion part if necessary.
  static void poseMsgToEigen(const geometry_msgs::Pose& msg, Eigen::Isometry3d& out);
```
And always add a doc string to new functions :wink:

j-petit: Thanks for the suggestions, I included them.

moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -983,6 +983,9 @@ class PlanningScene : private boost::noncopyable, public std::enable_shared_from
   bool processCollisionObjectRemove(const moveit_msgs::CollisionObject& object);
   bool processCollisionObjectMove(const moveit_msgs::CollisionObject& object);
 
+  /* convert Pose msg to Eigen::Isometry, normalizing the quaternion part if necessary. */
```
rhaschke: doxygen requires `/** ... */` or `///`
```suggestion
  /** convert Pose msg to Eigen::Isometry, normalizing the quaternion part if necessary. */
```

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1713,6 +1713,14 @@ bool PlanningScene::processCollisionObjectMsg(const moveit_msgs::CollisionObject
   return false;
 }
 
+void PlanningScene::poseMsgToEigen(const geometry_msgs::Pose& msg, Eigen::Isometry3d& out)
+{
+  Eigen::Translation3d translation(msg.position.x, msg.position.y, msg.position.z);
+  Eigen::Quaterniond quaternion(msg.orientation.w, msg.orientation.x, msg.orientation.y, msg.orientation.z);
+  quaternion.normalize();
```
henningkayser: Shouldn't this be added to `tf2_eigen.h` or even all conversion functions with quaternions? `fromMsg()` could have a default parameter `normalize_quaternion=false`.

rhaschke: This would be indeed a nice to have. However, OSRF wasn't very responsive for ROS1 changes in the recent past.

j-petit: I can change the calls to tf2::fromMsg() in other places, however I am not so sure I this is really desired as then quaternion normlization is done more often. I read up on the discussion [here](https://github.com/ros-visualization/rviz/pull/1179), and for me it sounded like we don't want to normalize the quaternions everywhere. But maybe @rhaschke can clarify that?

henningkayser: If that's the case it's fine with me. It just seems odd that collision objects are affected while attached collision objects are not. I approve this patch as is. @rhaschke do you have something to add here? Otherwise lgtm.

rhaschke: I looked into this briefly a week ago and I agree that we should consider normalization in more places. Investigating the extent of required changes, I found some nasty code duplication between `PlanningScene` code and `robot_state/.../conversions.cpp`. I wanted to clean this up but didn't find time yet to do so.

rhaschke: @henningkayser, let's merge for now. Remember to cherry-pick to master. This PR is targeted at melodic-devel currently.

Issue #1067, opened 2018-09-23
get_planning_scene with 0-component should return complete scene


v4hn:
Currently, the `/get_planning_scene` service returns an empty scene message when the request field has `components` set to 0. I believe it would be more useful if the service would return the *full* planning scene in this case.

It is cumbersome and error-prone to specify each and every bit (i.e. bitmask 1023) explicitly for each full call.

It should be validated throughout MoveIt whether we call the service like this in some context (although it does not make much sense...), because the ROS overhead would increase tremendously.

gavanderhoorn:
> It is cumbersome and error-prone to specify each and every bit (i.e. bitmask 1023) explicitly for each full call.

`(~0) & 0x3FF`?

j-petit:
I am taking this up.

rhaschke:
Addressed in #1424

PR #1424, opened 2019-04-02
Complete scene return for component = 0 (#1067)

j-petit:
### Description
When calling the /get_planning_scene service with component = 0, the service will return the full scene instead of an empty one (proposed in #1067)
I identified a single place in the codebase where a call to the service with all components takes place and replaced it with a call with components = 0. I am not to happy about writing it out as it is (that's why I had to add the comment), maybe someone can give me a hint to make it more understandable intuitively.

### Checklist
- [x] Document API changes relevant to the user in the moveit/MIGRATION.md notes


moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -471,12 +471,8 @@ bool PlanningSceneMonitor::requestPlanningSceneState(const std::string& service_
...
   ros::ServiceClient client = ros::NodeHandle().serviceClient<moveit_msgs::GetPlanningScene>(service_name);
   moveit_msgs::GetPlanningScene srv;
-  srv.request.components.components =
-      srv.request.components.SCENE_SETTINGS | srv.request.components.ROBOT_STATE |
-      srv.request.components.ROBOT_STATE_ATTACHED_OBJECTS | srv.request.components.WORLD_OBJECT_NAMES |
-      srv.request.components.WORLD_OBJECT_GEOMETRY | srv.request.components.OCTOMAP |
-      srv.request.components.TRANSFORMS | srv.request.components.ALLOWED_COLLISION_MATRIX |
-      srv.request.components.LINK_PADDING_AND_SCALING | srv.request.components.OBJECT_COLORS;
+  // Requests all components if 0 is specified
+  srv.request.components.components = 0;
```
felixvd: This is already initialized to 0. Would remove this line and reword the comment to "All scene components are returned if none are specified" or similar.

moveit_ros/move_group/src/default_capabilities/get_planning_scene_service_capability.cpp
```diff
@@ -54,7 +54,26 @@ bool move_group::MoveGroupGetPlanningSceneService::getPlanningSceneService(movei
   if (req.components.components & moveit_msgs::PlanningSceneComponents::TRANSFORMS)
     context_->planning_scene_monitor_->updateFrameTransforms();
   planning_scene_monitor::LockedPlanningSceneRO ps(context_->planning_scene_monitor_);
-  ps->getPlanningSceneMsg(res.scene, req.components);
+
+  if (!req.components.components)
```
felixvd: The same comment would be helpful here: "Return all scene components if none are specified"

moveit_ros/move_group/src/default_capabilities/get_planning_scene_service_capability.cpp
```diff
@@ -54,7 +54,26 @@ bool move_group::MoveGroupGetPlanningSceneService::getPlanningSceneService(movei
...
+  {
+    moveit_msgs::PlanningSceneComponents all_components;
+    all_components.components =
```
felixvd: `INT_MAX`?

j-petit:
I searched through the codebase thoroughly and am quite confident that there are no calls to the service where an empty scene response is expected. The service is called only through an intermediate function which always requests the full scene. Nevertheless, I would still like to do the performance check but I don't know what exactly do you mean when referring to performance and how would you measure it. Could you give me a hint how to do this?

felixvd:
The keyword is code benchmarking, and I would simply test if calling the `get_planning_scene` service is slower with `req.components.components = 0` and `req.components.components = UINT_MAX`. If you call the service from a Python script a bunch of times and time the execution, that should be enough. There are much better tools for this, but I don't know them. This might be overkill anyway; someone with more compiler knowledge would know if this will affect performance.

Apart from that, I also had a look through the code and couldn't find any calls to this service except the `PlanningSceneMonitor::requestPlanningSceneState` wrapper which you changed. I think the change itself is fine.

To document it, I would personally add the same comment to the [PlanningSceneRequest msg file](https://github.com/ros-planning/moveit_msgs/blob/melodic-devel/msg/PlanningSceneComponents.msg) and maybe a line like "get_planning_scene service now returns the whole scene if no components are specified" to the migration notes. Maybe some maintainers can give their opinion and request those changes. Apart from that and the first paragraph, I think this is fine.

That said, I haven't tested it.

j-petit:
> The keyword is code benchmarking, and I would simply test if calling the `get_planning_scene` service is slower with `req.components.components = 0` and `req.components.components = UINT_MAX`. If you call the service from a Python script a bunch of times and time the execution, that should be enough. There are much better tools for this, but I don't know them. This might be overkill anyway; someone with more compiler knowledge would know if this will affect performance.
> 
I did some measurements and the difference is tiny. So I think we are good to go from this aspect.

> To document it, I would personally add the same comment to the [PlanningSceneRequest msg file](https://github.com/ros-planning/moveit_msgs/blob/melodic-devel/msg/PlanningSceneComponents.msg) and maybe a line like "get_planning_scene service now returns the whole scene if no components are specified" to the migration notes. Maybe some maintainers can give their opinion and request those changes. Apart from that and the first paragraph, I think this is fine.
>
I think that the comment is better fit in the service description (moveit_msgs/srv/GetPlanningScene.srv) as this is a property of the service and not the message itself?

felixvd:
Yes, the [service description file](https://github.com/ros-planning/moveit_msgs/blob/melodic-devel/srv/GetPlanningScene.srv) is the right place. I got confused copying things around.

MIGRATION.md
```diff
@@ -2,6 +2,12 @@
 
 API changes in MoveIt! releases
 
+## ROS Noetic (upcoming changes in master)
+- Extended the return value of `MoveitCommander.MoveGroup.plan()` from `trajectory` to a tuple of `(success, trajectory, planning_time, error_code)` to better match the C++ MoveGroupInterface ([790](https://github.com/ros-planning/moveit/pull/790/))
+- `moveit_rviz.launch`, generated by MSA, provides an argument `rviz_config` to configure the rviz config to be used. The old boolean config argument was dropped. ([1397](https://github.com/ros-planning/moveit/pull/1397))
+- Moved the example package `moveit_controller_manager_example` into [moveit_tutorials](https://github.com/ros-planning/moveit_tutorials)
+- Requests to `get_planning_scene` service without explicitely setting components returns full scene
```
felixvd: explicitly setting "components" now return the full scene

j-petit:
Sorry, when trying to get a clean PR through incorporating changes from master to my devel-branch, I messed it up. Cleaning up now...

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -999,55 +985,55 @@ bool RobotState::clearAttachedBody(const std::string& id)
     return false;
 }
 
-const Eigen::Isometry3d& RobotState::getFrameTransform(const std::string& id)
+const Eigen::Isometry3d& RobotState::getFrameTransform(const std::string& frame_id)
```
felixvd: These changes should clear up if you rebase your branch on master.

j-petit: Thanks, worked.

MIGRATION.md
```diff
@@ -6,6 +6,7 @@ API changes in MoveIt! releases
 - Extended the return value of `MoveitCommander.MoveGroup.plan()` from `trajectory` to a tuple of `(success, trajectory, planning_time, error_code)` to better match the C++ MoveGroupInterface ([790](https://github.com/ros-planning/moveit/pull/790/))
 - `moveit_rviz.launch`, generated by MSA, provides an argument `rviz_config` to configure the rviz config to be used. The old boolean config argument was dropped. ([1397](https://github.com/ros-planning/moveit/pull/1397))
 - Moved the example package `moveit_controller_manager_example` into [moveit_tutorials](https://github.com/ros-planning/moveit_tutorials)
+- Requests to `get_planning_scene` service without explicitely setting "components" now returns full scene
```
felixvd: -> explicitly

felixvd: -> return

j-petit: Sorry, missed my spelling mistakes.

MIGRATION.md
```diff
@@ -6,6 +6,7 @@ API changes in MoveIt! releases
 - Extended the return value of `MoveitCommander.MoveGroup.plan()` from `trajectory` to a tuple of `(success, trajectory, planning_time, error_code)` to better match the C++ MoveGroupInterface ([790](https://github.com/ros-planning/moveit/pull/790/))
 - `moveit_rviz.launch`, generated by MSA, provides an argument `rviz_config` to configure the rviz config to be used. The old boolean config argument was dropped. ([1397](https://github.com/ros-planning/moveit/pull/1397))
 - Moved the example package `moveit_controller_manager_example` into [moveit_tutorials](https://github.com/ros-planning/moveit_tutorials)
+- Requests to `get_planning_scene` service without explicitly setting "components" now return full scene
```
rhaschke:
```suggestion
- Requests to `get_planning_scene` service without explicitly setting "components" now returns full scene
```

moveit_ros/move_group/src/default_capabilities/get_planning_scene_service_capability.cpp
```diff
@@ -54,7 +54,19 @@ bool move_group::MoveGroupGetPlanningSceneService::getPlanningSceneService(movei
   if (req.components.components & moveit_msgs::PlanningSceneComponents::TRANSFORMS)
     context_->planning_scene_monitor_->updateFrameTransforms();
   planning_scene_monitor::LockedPlanningSceneRO ps(context_->planning_scene_monitor_);
-  ps->getPlanningSceneMsg(res.scene, req.components);
+
+  // all scene components are returned if none are specified
+  if (!req.components.components)
```
rhaschke: I suggested a more compact code snippet.

MIGRATION.md
j-petit: No, I think this is correct, as requests is plural.

moveit_ros/move_group/src/default_capabilities/get_planning_scene_service_capability.cpp
j-petit: thanks!

moveit_ros/move_group/src/default_capabilities/get_planning_scene_service_capability.cpp
```diff
@@ -55,17 +55,8 @@ bool move_group::MoveGroupGetPlanningSceneService::getPlanningSceneService(movei
...
-    moveit_msgs::PlanningSceneComponents all_components;
-    all_components.components = UINT_MAX;
-    ps->getPlanningSceneMsg(res.scene, all_components);
-  }
-  else
-  {
-    ps->getPlanningSceneMsg(res.scene, req.components);
-  }
+  static const moveit_msgs::PlanningSceneComponents all_components{ UINT_MAX };
+  ps->getPlanningSceneMsg(res.scene, req.components.components ? req.components : all_components);
```
felixvd: Pretty, but I'd keep the comment `// Return all scene components if none are specified`.

rhaschke: Fair enough!

MIGRATION.md
rhaschke: You are right.

BryceStevenWilley:
Travis failed because it looks like `rosdep` (or maybe github) was down a few days ago. I'm going to close and re-open this PR to retrigger the build.

BryceStevenWilley:
Attempting to fix flaky CI [here](https://github.com/ros-planning/moveit_ci/pull/63).

BryceStevenWilley:
Trying one more time to get Travis to pass.

PR #1434, opened 2019-04-12
Feature/ik debugging helpers

ct2034:
### Description
This improves

1. The generation of the `update_ikfast_plugin.sh` for it to be more readable and to be debugged better
2. Error output if the initialization of the ikfast plugin fails

### Checklist

### DIFF
--- a/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
+++ b/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
@@ -304,7 +304,7 @@ def update_ikfast_package(args):
             + "eef_link_name=" + args.eef_link_name + "\n"
             + "ikfast_output_path=" + args.ikfast_output_path + "\n\n"
             + "rosrun moveit_kinematics create_ikfast_moveit_plugin.py \\\n"
-            + "--search_mode=$search_mode \\\n"
+            + "  --search_mode=$search_mode \\\n"
             + "--srdf_filename=$srdf_filename \\\n"
             + "--robot_name_in_srdf=$robot_name_in_srdf \\\n"
             + "--moveit_config_pkg=$moveit_config_pkg \\\n"
--- a/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
+++ b/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
@@ -293,27 +293,27 @@ def update_ikfast_package(args):
   # Create a script for easily updating the plugin in the future in case the plugin needs to be updated
   easy_script_file_path = args.ikfast_plugin_pkg_path + "/update_ikfast_plugin.sh"
   with open(easy_script_file_path, 'w') as f:
-    f.write("search_mode="+args.search_mode+
-            "\nsrdf_filename=" + args.srdf_filename +
-            "\nrobot_name_in_srdf=" + args.robot_name_in_srdf +
-            "\nmoveit_config_pkg=" + args.moveit_config_pkg +
-            "\nrobot_name=" + args.robot_name +
-            "\nplanning_group_name=" + args.planning_group_name +
-            "\nikfast_plugin_pkg=" + args.ikfast_plugin_pkg +
-            "\nbase_link_name=" + args.base_link_name +
-            "\neef_link_name=" + args.eef_link_name +
-            "\nikfast_output_path=" + args.ikfast_output_path +
-            "\n\nrosrun moveit_kinematics create_ikfast_moveit_plugin.py \\" +
-            "\n--search_mode=$search_mode \\" +
-            "\n--srdf_filename=$srdf_filename \\" +
-            "\n--robot_name_in_srdf=$robot_name_in_srdf \\" +
-            "\n--moveit_config_pkg=$moveit_config_pkg \\" +
-            "\n$robot_name \\" +
-            "\n$planning_group_name \\" +
-            "\n$ikfast_plugin_pkg \\" +
-            "\n$base_link_name \\" +
-            "\n$eef_link_name \\" +
-            "\n$ikfast_output_path")
+    f.write("search_mode=" + args.search_mode + "\n"
+            + "srdf_filename=" + args.srdf_filename + "\n"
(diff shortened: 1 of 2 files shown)


moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
```diff
@@ -293,17 +293,27 @@ def update_ikfast_package(args):
...
-            " --srdf_filename=" + args.srdf_filename +
-            " --robot_name_in_srdf=" + args.robot_name_in_srdf +
-            " --moveit_config_pkg=" + args.moveit_config_pkg +
-            " " + args.robot_name +
-            " " + args.planning_group_name +
-            " " + args.ikfast_plugin_pkg +
-            " " + args.base_link_name +
-            " " + args.eef_link_name +
-            " " + args.ikfast_output_path)
+    f.write("search_mode="+args.search_mode+
```
rhaschke:
```suggestion
    f.write("search_mode=" + args.search_mode +
```

moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
```diff
@@ -293,17 +293,27 @@ def update_ikfast_package(args):
...
+            "\nsrdf_filename=" + args.srdf_filename +
```
rhaschke: What's the value of listing those parameters before calling `ros_run`? 
Most of them are used in conjuction with a verbosely named parameter, aren't they?

moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
```diff
@@ -293,17 +293,27 @@ def update_ikfast_package(args):
...
+            "\nrobot_name_in_srdf=" + args.robot_name_in_srdf +
```
rhaschke: Please place the newline codes at the end of the lines.

moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
ct2034: this way you can look at the file and know what the values are, before calling it. I just prefer to know what the values are, before calling the script

moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
ct2034: ok

ct2034:
> Usually, there shouldn't be a need for "normal" users to look into this file, right?

What is a _normal_ user in ros?
I think you should always be able to dig deeper into how this works. Especially when you autogenerate code, so many things can go wrong. Also, for example when you rename your move_group or something I think it would be good if you could check it here.

moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
felixvd: I've never used ikfast. Can you give an example of how the generated script would look before and after this change?

moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py
```diff
@@ -293,17 +293,27 @@ def update_ikfast_package(args):
...
+            + "robot_name_in_srdf=" + args.robot_name_in_srdf + "\n"
+            + "moveit_config_pkg=" + args.moveit_config_pkg + "\n"
+            + "robot_name=" + args.robot_name + "\n"
+            + "planning_group_name=" + args.planning_group_name + "\n"
+            + "ikfast_plugin_pkg=" + args.ikfast_plugin_pkg + "\n"
+            + "base_link_name=" + args.base_link_name + "\n"
+            + "eef_link_name=" + args.eef_link_name + "\n"
+            + "ikfast_output_path=" + args.ikfast_output_path + "\n\n"
+            + "rosrun moveit_kinematics create_ikfast_moveit_plugin.py \\\n"
+            + "--search_mode=$search_mode \\\n"
```
rhaschke: Can you please re-add the leading spaces in front of all arguments?
I would go for two spaces per line.
```suggestion
            + "  --search_mode=$search_mode \\\n"
```

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -416,7 +416,13 @@ bool IKFastKinematicsPlugin::initialize(const moveit::core::RobotModel& robot_mo
   // base_frame_ and IKFAST_TIP_FRAME_ the tip_frame_ and transform to the input pose accordingly
   if (!computeRelativeTransform(tip_frames_[0], IKFAST_TIP_FRAME_, group_tip_to_chain_tip_, tip_transform_required_) ||
       !computeRelativeTransform(IKFAST_BASE_FRAME_, base_frame_, chain_base_to_group_base_, base_transform_required_))
+  {
```
rhaschke: This kind of duplicates the error output generated in `computeRelativeTransform`:
https://github.com/ros-planning/moveit/blob/6395358ce41bb6d61f9f29bc0b03054bd1dbb95a/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp#L382-L391

rhaschke:
Ping @ct2034. Could you please address the remaining comments.

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
ct2034: but it relies therefore on computeRelativeTransform generating the output. I think this way it would be clearer why the init fails.

rhaschke:
@ct2034 I was asking to re-add those spaces in all lines. My (single-line) github proposal was just an example.

ct2034:
@rhaschke Yes, sorry.

Issue #1123, opened 2018-10-24
planning_scene_interface::PlanningSceneInterface needs more convenience methods


v4hn:
The `planning_scene_interface::PlanningSceneInterface` class is advertised as the standard interface to interact with the planning scene of the `move_group` node.
The purpose of the class is to simplify most changes to the planning scene to one-liners in C++.

There are a number of possible convenience methods missing though:

- [ ] `getObject` (in contrast to `getObjects`)
- [ ] `getPlanningScene` which forwards the internal `GetPlanningScene` service to the user
- [ ] `getFullPlanningScene`, which requests a complete planning scene and instantiates a `planning_scene::PlanningScene` class instance

There are many other methods that would be appreciated too, e.g., `setColor(std::string, std_msgs::ColorRGBA)`, `allowCollisions`/`forbidCollisions`, ...

Ridhwanluthra:
I am starting to work on this as a part of WMD

felixvd:
I'd be interested to know the progress on this, but I didn't see a branch in your repo. I have some changes that I did over the last month that I could add right away, including Python bindings and `allowCollisions/disallowCollisions` functions.

PR #1440, opened 2019-04-17
Add allowCollisions/disallowCollisions to planning scene interface

felixvd:
### Description

Adds some convenience methods to planning_scene_interface that were mentioned as examples in #1123 : `allowCollisions`, `disallowCollisions` and `setCollisions`. The functions do what you expect them to, and they have comments and python bindings. I fixed typos and clarified some comments on the side, but I am too lazy to open another PR for that.

Cherry-pick to melodic would be useful. I wrote these changes for a project that was running on Kinetic, so they should work there as well.

### Checklist
- [X] Extended the tutorials / documentation, if necessary (should be updated automatically)
- [X] Document API changes relevant to the user in the moveit/MIGRATION.md notes (only new functions are added)
- [ ] Created tests, which fail without this PR [reference](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/tests.html) (I don't know how to write tests :( )
- [X] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers

### DIFF
--- a/moveit_ros/planning_interface/planning_scene_interface/include/moveit/planning_scene_interface/planning_scene_interface.h
+++ b/moveit_ros/planning_interface/planning_scene_interface/include/moveit/planning_scene_interface/planning_scene_interface.h
@@ -138,6 +138,48 @@ public:
       consider using `applyCollisionObjects` instead. */
   void removeCollisionObjects(const std::vector<std::string>& object_ids) const;
 
+  /** \brief Allow or disallow collisions for a set of links or between two sets. */
+  void setCollisions(bool set_to_allow, const std::vector<std::string>& link_group_1,
+                     const std::vector<std::string>& link_group_2 = std::vector<std::string>());
+  void setCollisions(bool set_to_allow, const std::string& link_name_1,
+                     const std::vector<std::string>& link_group_2 = std::vector<std::string>());
+  void setCollisions(bool set_to_allow, const std::string& link_name_1, const std::string& link_name_2 = "");
+
+  /** \brief Allow or disallow collisions for links or groups of links.
+  * This sets the AllowedCollisionMatrix of the PlanningScene.
+  * If the second argument is empty, the collision behavior is set between the link(s) in the first argument
+  * and all other objects in the world. Supplying both arguments only sets collision behavior between the two.
+  * If two link groups are supplied, the collision behavior is set between all the links of group 1 and all the links of
+  * group 2.
+  *
+  * E.g. allowCollisions("gripper_wrist_link", "table_base_link") allows collisions between
+  * the table and the wrist link of your gripper. You will want to add the other gripper links as well.
+ */
+  void allowCollisions(const std::string& link_name_1, const std::string& link_name_2 = "")
+  {
+    setCollisions(true, link_name_1, link_name_2);
+  }
+  void allowCollisions(const std::vector<std::string>& link_group_1, const std::string& link_name_2 = "")
+  {
+    setCollisions(true, link_name_2, link_group_1);
+  }
+  void allowCollisions(const std::vector<std::string>& link_group_1, const std::vector<std::string>& link_group_2)
+  {
+    setCollisions(true, link_group_1, link_group_2);
+  }
+  void disallowCollisions(const std::string& link_name_1, const std::string& link_name_2 = "")
+  {
+    setCollisions(false, link_name_1, link_name_2);
+  }
+  void disallowCollisions(const std::vector<std::string>& link_group_1, const std::string& link_name_2 = "")
+  {
+    setCollisions(false, link_name_2, link_group_1);
+  }
+  void disallowCollisions(const std::vector<std::string>& link_group_1, const std::vector<std::string>& link_group_2)
+  {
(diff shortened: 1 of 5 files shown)


felixvd:
Pinging @v4hn because he wrote the connected issue

moveit_ros/planning_interface/planning_scene_interface/src/wrap_python_planning_scene_interface.cpp
```diff
@@ -110,6 +110,104 @@ class PlanningSceneInterfaceWrapper : protected py_bindings_tools::ROScppInitial
     py_bindings_tools::deserializeMsg(ps_str, ps_msg);
     return applyPlanningScene(ps_msg);
   }
+
+  bool applyCollisionObjectPython1(const std::string& co_str)
```
rhaschke: Where are these new functions used? Shouldn't they be registered with the Python wrapper?

felixvd: Whoops! How embarrassing. I'll push a new set of commits tomorrow.

felixvd:
I don't want to look stupid forgetting committing half my code again, so I tried to write a test for the planning scene interface python wrapper. [This](http://docs.ros.org/melodic/api/moveit_tutorials/html/doc/tests/tests_tutorial.html) is the resource that I found, but the relevant lines don't seem to work for me, even after a complete build in the `master-source` container:

```shell
root@0821e906b791:~/ws_moveit# source devel/setup.bash 
root@0821e906b791:~/ws_moveit# roslaunch moveit_resources demo.launch
RLException: [demo.launch] is neither a launch file in package [moveit_resources] nor is [moveit_resources] a launch file name
The traceback for the exception was written to the log file
root@0821e906b791:~/ws_moveit# rostest moveit_ros_planning_interface python_move_group.test
[python_move_group.test] is neither a launch file in package [moveit_ros_planning_interface] nor is [moveit_ros_planning_interface] a launch file name
```

Do you have an idea what I am overlooking? A little off topic, but I'd appreciate the troubleshooting help.

rhaschke:
Looks like your workspace configuration is broken. You should check that the packages `moveit_resources` and `moveit_ros_planning_interface` are found and contain the corresponding test files.

moveit_ros/planning_interface/planning_scene_interface/include/moveit/planning_scene_interface/planning_scene_interface.h
```diff
@@ -138,6 +138,30 @@ class PlanningSceneInterface
       consider using `applyCollisionObjects` instead. */
   void removeCollisionObjects(const std::vector<std::string>& object_ids) const;
 
+  /** \brief Allow or disallow collisions for links, pairs of links, or groups of links.
+ * This sets the AllowedCollisionMatrix of the PlanningScene accordingly.
+ * If link_name_2 is empty, this sets collision behavior between link_name_1 and all other objects in the world.
+ * Supplying both link names only sets collision behavior between the two link_names.
+ * If both link names are supplied, the collision behavior is set
+ * between all the links of group 1 and all the links of group 2.
```
v4hn: if both names are supplied, the *groups* are set?

moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp
```diff
@@ -247,6 +248,38 @@ class PlanningSceneInterface::PlanningSceneInterfaceImpl
...
+      else
+        for (const auto& b2 : link_group_2)
+          acm.setEntry(b, b2, allowed_or_disallowed);
+    }
+
+    ps.is_diff = true;
+    acm.getMessage(ps.allowed_collision_matrix);
+    return applyPlanningScene(ps);
+  }
+
```
v4hn: This method invites a rather evil kind of racing condition because you pass the whole collision matrix via ROS services between nodes twice.
I am not in favor of having this implementation upstream.
I proposed to change the diff behavior of diff-PlanningScenes a year ago in https://github.com/ros-planning/moveit/issues/964 but nothing came out of it until now.
Do you think you could move your logic to the PlanningScene class and and implement a safe version as proposed in the issue?

felixvd: True. It's safe enough for our purposes, but not generally. I pushed the layer that we have in our system right now and added to the discussion in the other thread.

v4hn:
I will close this for now as the current approach will not get merged upstream.

For a better implementation, it makes sense to have a new request anyway.

felixvd:
> Looks like your workspace configuration is broken. You should check that the packages `moveit_resources` and `moveit_ros_planning_interface` are found and contain the corresponding test files.

Turns out that `source devel/setup.bash` does not work in `ws_moveit`, it has to be `source install/setup.bash`. See [here](https://answers.ros.org/question/346108/why-are-my-packages-linked-to-the-underlay-workspace-rather-than-my-catkin-workspace-problem-with-catkin-install-space/).

Issue #1048, opened 2018-09-04
Catch contradicting goal constraints


felixvd:
I got held up by [this simple bug](https://answers.ros.org/question/299979/moveit-plan-fails-after-attaching-an-object-but-not-from-rviz-or-after-restarting-requesting-node/). It would be nice if this kind of error was caught and the user alerted.

The check could simply be if multiple goal pose and orientation constraints exist although the group consists of a single chain (or has 7 DOF or less, or something of the sort).  Displaying a warning message would be sufficient. A good place might be during the construction of the MotionPlanRequest in the move group interface.

I don't know when I will get around to this, so feel free to pick it up if I haven't updated in a few weeks.

PR #1443, opened 2019-04-19
Add user warning to planning pipeline when multiple constraints are set and planning fails

felixvd:
### Description

Fixes https://github.com/ros-planning/moveit/issues/1048. When I had [this problem](https://answers.ros.org/question/299979/moveit-plan-fails-after-attaching-an-object-but-not-from-rviz-or-after-restarting-requesting-node/), I would have appreciated a more helpful error message. @v4hn requested a PR, so I made this.

The summary of the ROS answers post is that I forgot to call `clearPoseTargets` after changing the end effector link of the `move_group`, so that constraints were active on multiple links of the robot. I added an error message that displays if planning fails and multiple constraints are set.

You can test that the message fires correctly by adding the lines below in [this part](https://github.com/ros-planning/moveit_tutorials/blob/master/doc/move_group_interface/src/move_group_interface_tutorial.cpp#L126) of the tutorial:

```
  geometry_msgs::Pose target_pose2;
  target_pose2.orientation.w = 1.0;
  target_pose2.position.x = 0.28;
  target_pose2.position.y = 0.2;
  target_pose2.position.z = 0.5;
  move_group.setEndEffectorLink("panda_link7");
  move_group.setPoseTarget(target_pose2);
```

This is simple enough to be cherry-picked to both kinetic and melodic.

### Checklist
- [ ] Document API changes relevant to the user in the moveit/MIGRATION.md notes (not needed)
- [ ] Created tests, which fail without this PR [reference](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/tests.html) (I am not sure this is required)

### DIFF
--- a/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
+++ b/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
@@ -339,6 +339,26 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
     display_path_publisher_.publish(disp);
   }
 
+  if (!solved)
+  {
+    // This should alert the user if planning failed because of contradicting constraints.
+    // Could be checked more thoroughly, but it is probably not worth going to that length.
+    bool stacked_constraints = false;
+    if (req.goal_constraints.size() > 1 || req.path_constraints.joint_constraints.size() > 1 ||
+        req.path_constraints.position_constraints.size() > 1 || req.path_constraints.orientation_constraints.size() > 1)
+      stacked_constraints = true;
+    for (auto constraint : req.goal_constraints)
+    {
+      if (constraint.joint_constraints.size() > 1 || constraint.position_constraints.size() > 1 ||
+          constraint.orientation_constraints.size() > 1)
+        stacked_constraints = true;
+    }
+    if (stacked_constraints)
+      ROS_WARN("More than one constraint is set. If your move_group does not have multiple end effectors/arms, this is "
+               "unusual. Are you using a move_group_interface and forgetting to call clearPoseTargets() or "
+               "equivalent?");
+  }
+
   return solved && valid;
 }
 
--- a/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
+++ b/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
@@ -344,13 +344,12 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
     // This should alert the user if planning failed because of contradicting constraints.
     // Could be checked more thoroughly, but it is probably not worth going to that length.
     bool stacked_constraints = false;
-    if (req.goal_constraints.size() > 1 || req.path_constraints.joint_constraints.size() > 1 ||
-        req.path_constraints.position_constraints.size() > 1 || req.path_constraints.orientation_constraints.size() > 1)
+    if (req.goal_constraints.size() > 1 || req.path_constraints.position_constraints.size() > 1 ||
+        req.path_constraints.orientation_constraints.size() > 1)
       stacked_constraints = true;
     for (auto constraint : req.goal_constraints)
     {
-      if (constraint.joint_constraints.size() > 1 || constraint.position_constraints.size() > 1 ||
-          constraint.orientation_constraints.size() > 1)
+      if (constraint.position_constraints.size() > 1 || constraint.orientation_constraints.size() > 1)
         stacked_constraints = true;
     }
     if (stacked_constraints)


moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -339,6 +339,26 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
...
+  {
+    // This should alert the user if planning failed because of contradicting constraints.
+    // Could be checked more thoroughly, but it is probably not worth going to that length.
+    bool stacked_constraints = false;
+    if (req.goal_constraints.size() > 1 || req.path_constraints.joint_constraints.size() > 1 ||
+        req.path_constraints.position_constraints.size() > 1 || req.path_constraints.orientation_constraints.size() > 1)
+      stacked_constraints = true;
+    for (auto constraint : req.goal_constraints)
+    {
+      if (constraint.joint_constraints.size() > 1 || constraint.position_constraints.size() > 1 ||
```
henningkayser: When setting joint state targets it's perfectly normal to have multiple joint constraints (1 for each joint) so the warning is not accurate. I would remove joint constraints from this entirely since they can't really contradict each other (except there are multiple for the same joint). Having multiple position or orientation constraints does not mean that multiple arms are constrained. The links can be and most often are part of the same arm and the goal pose is sampled to meet all such constraints. I think without an actual check if there are unfeasible constraints (i.e. directly conflicting constraints that apply to the same joint/link) this kind of warning doesn't really make sense.

felixvd: I had never used joint constraints and didn't notice each joint would be its own constraint. I agree and changed it.

Multiple position and orientation constraints seems to be an advanced use case to me. I feel like the most common case (and the only one that I remember seeing in the tutorials) is setting a 6D pose target for the end effector link through the move_group_interface, which results in one position and orientation constraint (each of them constraining 3 translations and rotations; I had to [double check](http://docs.ros.org/melodic/api/moveit_msgs/html/msg/Constraints.html)).

Like it says in the comment, a more thorough check seems nice, but nothing trivial comes to mind for me. I mean, you could check if the kinematic chain is overconstrained or multiple joint constraints apply to the same joint, but I don't know if it is worth the effort to avoid the false positives. This is more of a sanity check/nudge after a failed planning attempt to help users troubleshoot, which I assume advanced users would disregard if it doesn't apply to them. I think it's worth displaying this sort of nudge in general, because it can save a lot of user frustration, but I am not sure how the others feel about this.

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -339,6 +339,25 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
     display_path_publisher_.publish(disp);
   }
 
+  if (!solved)
+  {
+    // This should alert the user if planning failed because of contradicting constraints.
+    // Could be checked more thoroughly, but it is probably not worth going to that length.
+    bool stacked_constraints = false;
+    if (req.goal_constraints.size() > 1 || req.path_constraints.position_constraints.size() > 1 ||
```
rhaschke: Several goal_constraints are perfectly fine. In this fashion you can define several independent goal regions.

felixvd: Yes. This warning is displayed upon planning failure although some users may have valid reasons for setting multiple constraints, but that's why the tone of the error message is "Hey maybe this is what made your planning fail, did you mean to do this?"

What is the change you would like to see in this snippet? I think users advanced enough to set multiple goal regions will have the confidence to ignore the message and not be bothered.

felixvd:
If the policy is to cherry-pick bugfixes only, then it's not necessary, but it's a simple enough change to go into melodic, too.

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
rhaschke: The condition should read:
`if (req.path_constraints.position_constraints.size() > 1 || req.path_constraints.orientation_constraints.size() > 1)`
i.e. remove the check `req.goal_constraints.size() > 1`, which is required to define multiple goal regions.
Further below you check each of these goal constraints for multiple position/orientation constraints...

felixvd: I think that might not catch the problem in my ROS Answers post, where I accidentally defined two goals although I only had one robot arm. I would err on the side of displaying too much when things fail. Is there an easy way to make a more precise check? Is it worth the effort to code and execute it?

rhaschke:
According to the [message documentation](http://docs.ros.org/lunar/api/moveit_msgs/html/msg/MotionPlanRequest.html):
> The possible goal states for the model to plan for. Each element of
the array defines a goal region. The goal is achieved
if the constraints for a particular region are satisfied

goal_constraints specifies a _list_ of acceptable goal regions. These goal regions are explicitly allowed to be disjoint.

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
henningkayser: > @henningkayser Why did you merged this without a second review?

@rhaschke Sorry I was a bit too fast here!
It's true that multiple goal constraints should be treated as alternative targets and don't really apply here. Nevertheless, the warning is quite generic and would help in any case where constraints are somehow malformed.

@felixvd did you attempt to fix [your issue](https://answers.ros.org/question/299979/moveit-plan-fails-after-attaching-an-object-but-not-from-rviz-or-after-restarting-requesting-node/) directly?

rhaschke: I agree that the warning is generic. However, we shouldn't issue it, when there is no reason for it.
I don't ask for a more thorough implementation. I'm just asking to remove the unnecessary check `req.goal_constraints.size() > 1`.

felixvd: So all goal_constraints are treated as alternatives, so that only one of them has to be fulfilled for the planning to succeed? In that case it should be edited out, I agree. Please confirm first though.

rhaschke: Yes, that's correct. I'm not sure though, how this relates to your [original issue](https://answers.ros.org/question/299979/moveit-plan-fails-after-attaching-an-object-but-not-from-rviz-or-after-restarting-requesting-node/). Maybe you can try to reproduce your original issue? Maybe the root cause was something else?

felixvd: The cause was surely that a pose constraint was set for two different links on the same robot because I changed the end effector without clearing the targets. Our end effector had multiple tool tips.

I'll put in another PR for the correction.

PR #1449, opened 2019-04-25
Fix IKFastPlugin initialization. Fixes #1448

### LINKED ISSUE #1448: IKFast plugin not initialized correctly
### Description

We have a test failing sporadically after an IKfast update.
https://github.com/PilzDE/pilz_robots/blob/kinetic-devel/prbt_ikfast_manipulator_plugin/test/unittests/tst_prbt_ikfast_manipulator_plugin.cpp

My guess is that `IKFastKinematicsPlugin::computeRelativeTransform` does use an uninitialized RobotState, which can lead to errors.

An issue is also that above function returns true even if the computed transform is not well-defined.

martiniil:
- initialize values of RobotState in computeRelativeTransform()
- add tolerance for comparison to identity

### Description

see #1448

### Checklist

### DIFF
--- a/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
+++ b/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
@@ -375,6 +375,7 @@ bool IKFastKinematicsPlugin::computeRelativeTransform(const std::string& from, c
 {
   RobotStatePtr robot_state;
   robot_state.reset(new RobotState(robot_model_));
+  robot_state->setToDefaultValues();
 
   auto* from_link = robot_state->getLinkModel(from);
   auto* to_link = robot_state->getLinkModel(to);


moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -393,7 +396,7 @@ bool IKFastKinematicsPlugin::computeRelativeTransform(const std::string& from, c
   }
 
   transform = robot_state->getGlobalLinkTransform(from_link).inverse() * robot_state->getGlobalLinkTransform(to_link);
-  differs_from_identity = !transform.matrix().isIdentity();
+  differs_from_identity = !transform.matrix().isIdentity(RELATIVE_TRANSFORM_TOLERANCE);
```
rhaschke: Why do you think, we should change the default (machine) tolerance value here?

martiniil: Because we can avoid unnecessary computations, if we account for numerical errors. However I saw that the default precision is `1e-12`, which should be fine.

PR #1451, opened 2019-04-26
docker: remove CATKIN_WS, define ROS_UNDERLAY

rhaschke:
To enable https://github.com/ros-planning/moveit_ci/pull/62, define `ROS_UNDERLAY` in the source container. The `CATKIN_WS` environment variable is not used anymore by `moveit_ci` and is thus removed.

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -6,9 +6,8 @@ MAINTAINER Dave Coleman dave@picknik.ai
 
 ENV TERM xterm
 
-# Setup catkin workspace
-ENV CATKIN_WS=/root/ws_moveit
-WORKDIR $CATKIN_WS
+# Setup (temporary) ROS workspace
+WORKDIR /root/ws_moveit
 
 # Commands are combined in single RUN statement with "apt/lists" folder removal to reduce image size
 # https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers
--- a/.docker/source/Dockerfile
+++ b/.docker/source/Dockerfile
@@ -4,12 +4,12 @@
 FROM moveit/moveit:master-ci-shadow-fixed
 MAINTAINER Dave Coleman dave@picknik.ai
 
+ENV ROS_UNDERLAY /root/ws_moveit/install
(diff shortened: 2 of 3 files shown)


.docker/ci/Dockerfile
```diff
@@ -6,9 +6,8 @@ MAINTAINER Dave Coleman dave@picknik.ai
 
 ENV TERM xterm
 
-# Setup catkin workspace
-ENV CATKIN_WS=/root/ws_moveit
```
mlautman: I don't like how `/root/ws_moveit` is now spread across multiple files. Wouldn't it be better to set

```dockerfile
ENV CATKIN_WS=/root/ws_moveit
ENV ROS_UNDERLAY=$CATKIN_WS/install
```

.docker/source/Dockerfile
```diff
@@ -4,12 +4,12 @@
 FROM moveit/moveit:master-ci-shadow-fixed
 MAINTAINER Dave Coleman dave@picknik.ai
 
+ENV ROS_UNDERLAY /root/ws_moveit/install
+WORKDIR $ROS_UNDERLAY/../src
```
mlautman: See point above about restoring `CATKIN_WS` as a variable. That way we can set `WORKDIR $CATKIN_WS/src` which I think is quite a bit cleaner

.docker/ci/Dockerfile
rhaschke: The `CATKIN_WS` folder created/used in the `ci` and `ci-shadow-fixed` containers is only temporary: It's deleted after installing dependencies via rosdep. This folder could go anywhere.
Only the `source` container actually creates a workspace. Hence, it is sufficient to create it there.

rhaschke: @mlautman Can we proceed with this PR? I would like to use the `source` docker container for MTC as well.

rhaschke:
Ping @mlautman, @davetcoleman.
Would be nice to get this and https://github.com/ros-planning/moveit_ci/pull/62 merged to progress with Travis testing for
- [ ] https://github.com/ros-planning/moveit_tutorials/pull/317 and
- [ ] https://github.com/ros-planning/moveit_visual_tools/pull/47
- [ ] MTC

PR #1457, opened 2019-05-07
Allow subclassing of point_containment_filter::ShapeMask.

peci1:
### Description

Just made ShapeMask members protected, so that users can subclass it and make use of the existing data members and private classes.

This of course means making the C++ API public and saying it is stable.

### Checklist

### DIFF
--- a/moveit_ros/perception/point_containment_filter/include/moveit/point_containment_filter/shape_mask.h
+++ b/moveit_ros/perception/point_containment_filter/include/moveit/point_containment_filter/shape_mask.h
@@ -69,7 +69,7 @@ public:
   ShapeMask(const TransformCallback& transform_callback = TransformCallback());
 
   /** \brief Destructor to clean up */
-  ~ShapeMask();
+  virtual ~ShapeMask();
 
   ShapeHandle addShape(const shapes::ShapeConstPtr& shape, double scale = 1.0, double padding = 0.0);
   void removeShape(ShapeHandle handle);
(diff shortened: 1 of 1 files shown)


peci1:
Right now I'm writing a class (another variant of self 3D laser filter), and in the subclass I make heavy use of the mutex and mutex-protected members. It's just about knowing which members are mutex-protected and using them correctly. This can be IMO solved by documentation.

In my class, it wouldn't make sense to add mutex-locking accessor methods, because I need to lock the mutex for larger blocks of code at once, and the mutex is not recursive.

I aggree with the `freeMemory` function being private, that makes sense.

v4hn:
@peci1 are you still interested in this pr?

peci1:
Yes, I'm still interested, more even now when [robot_body_filter](https://github.com/peci1/robot_body_filter) is released for melodic. 

I force-pushed to this PR making less member protected and adding a comment to the mutex which explains when it is locked by this class and when it should be locked by subclasses. Does that seem better?

Also, it'd be better to have a virtual destructor here, but it'd change ABI (is that a problem?) and is not really critical. But it might deserve a comment in class definition stating that it cannot be used polymorphically. Would that be ok?

v4hn:
ABI changes are not a problem.
We deliberately decided to bump sonames with each release (as you probably have noticed in the past).
So please feel free to change the destructor to virtual. The change makes sense for your use case.

Incidentally, [some guides](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B) declare a visibility-change from private to protected a potential ABI change, although I don't know of a compiler that actually does break ABI for that.

peci1:
Done.

PR #1468, opened 2019-05-16
Sudden stop for critical issues, filtered deceleration otherwise

AndyZe:
### Description

A suddenHalt() function for critical issues like joint limits. Otherwise, decelerations are low-pass filtered.

### Checklist

### DIFF
--- a/moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
+++ b/moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
@@ -128,10 +128,8 @@ JogCalcs::JogCalcs(const JogArmParameters parameters, JogArmShared& shared_varia
     pthread_mutex_lock(&mutex);
     bool zero_cartesian_cmd_flag = shared_variables.zero_cartesian_cmd_flag;
     bool zero_joint_cmd_flag = shared_variables.zero_joint_cmd_flag;
-    pthread_mutex_unlock(&mutex);
 
     // Pull data from the shared variables.
-    pthread_mutex_lock(&mutex);
     incoming_jts_ = shared_variables.joints;
     pthread_mutex_unlock(&mutex);
 
--- a/moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
+++ b/moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
@@ -165,17 +165,19 @@ JogCalcs::JogCalcs(const JogArmParameters parameters, JogArmShared& shared_varia
         continue;
     }
 
-    // Halt if the command is stale or inputs are all zero
+    // Skip publishing if the command is stale and outgoing velocities are all nearly zero.
+    // (It takes some time for the low-pass velocity filters to settle at zero.)
     pthread_mutex_lock(&mutex);
-    bool stale_command = shared_variables.command_is_stale;
+    bool skip_publication = shared_variables.command_is_stale &&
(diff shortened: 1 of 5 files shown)


AndyZe:
@jschleicher  give it a review, if you please

jschleicher:
What about the check in the main loop, if a zero velocity command is received? https://github.com/ros-planning/moveit/blob/7f142291ccd8c7f203a4427e6353e3a950d09314/moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp#L135

Recording a bag file shows several unexpected behaviours:
1. while filtering looks good for negative joint velocities, it is still hard-reset to zero for positive velocities:
jog_arm-screenshot-2019-05-20 16-17-16
2. during stop sometimes there is an overshoot with movement into the opposite direction:
jog_arm-screenshot-2019-05-20 16-19-33

AndyZe:
Good catch on resetVelocityFilters() being out-of-place.

After moving it to a better spot, I'm not seeing anything unexpected from the filters. There is a tiny bit of overshoot at low_pass_filter_coefficient == 1. You can damp that down with a higher low_pass_filter_coefficient, if you want  (as in the plots with c == 10).

I just discovered PlotJuggler for capturing this data easily. It's fantastic.
position0_c_10
position1_c_10
velocity0_c_1
velocity_c_10

AndyZe:
The filter type used here is known to overshoot. It could be simplified to a first-order filter that apparently doesn't have overshoot.

https://dankelley.github.io/r/2014/01/15/butterworth-filters.html

SansoneG:
Maybe a new ros parameter could be used in the `config.yaml` to select the order of the butterworth filter that should be used.

AndyZe:
I'll make an issue for that.

jschleicher:
Just a pedantic comment: If the `incoming command_timeout` is lower than the time to reach zero velocity, the robot keeps moving now:
stop_missing

But a comment to increase the `incoming_command_timeout` for very high `low_pass_filter_coeff` (we have tested with `low_pass_filter_coeff=30`) in the documentation would be fine for me.

AndyZe:
If I understand right, the last velocity command is sometimes non-zero if `incoming_command_timeout` is really short.

Here's a fix for that.

jschleicher:
> If I understand right, the last velocity command is sometimes non-zero if `incoming_command_timeout` is really short.

Exactly. Your fix works for me.

AndyZe:
@henningkayser @BryceStevenWilley @felixvd Can I get a review please? Not sure where my 'Assign Reviewers' button went

moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
```diff
@@ -124,16 +124,12 @@ JogCalcs::JogCalcs(const JogArmParameters parameters, JogArmShared& shared_varia
   // Now do jogging calcs
   while (ros::ok())
   {
-    // If user commands are all zero, reset the low-pass filters when commands resume
+    // Flag that incoming commands are all zero. May be used to skip calculations/publication
     pthread_mutex_lock(&mutex);
     bool zero_cartesian_cmd_flag = shared_variables.zero_cartesian_cmd_flag;
     bool zero_joint_cmd_flag = shared_variables.zero_joint_cmd_flag;
```
BryceStevenWilley: These variables are unused, should remove.

AndyZe: It looks like they are used (e.g. lines 148,206). But now that I think about it, there are 2 mutex locks here that can be combined.

jschleicher:
@henningkayser @felixvd Ping

I'd appreciate, if this could be merged.

Issue #1470, opened 2019-05-19
CHOMP doesn't work with Floating joint


BryceStevenWilley:
### Description

While attempting to recreate https://github.com/ros-planning/moveit/issues/1452 (tutorial/panda_moveit_config also from source), I ran into another bug that causes CHOMP to fail with `Start state violated joint limits`. 

### Your environment
* ROS Distro: [Melodic]
* OS Version: e.g. Ubuntu 18.04
* Source: moveit/2dfe65e8

### Steps to reproduce
`roslaunch panda_moveit_config demo_chomp.launch`, move arm slightly.

### Expected behaviour
Should be able to plan. 

### Actual behaviour

```
[ INFO] [1558234232.380938293]: Configuring Planning Scene for CHOMP ....
[ERROR] [1558234232.511067939]: Start state violates joint limits
[ WARN] [1558234232.534404999]: Fail: ABORTED: Unknown event
```

### Backtrace or Console output

I printed out the actual joint angles and bounds being checked (code on [this branch](https://github.com/BryceStevenWilley/moveit/tree/floating_joint_broken_debug)), which is being reported as: 
```
[ERROR] [1558245278.740119596]: Joint 0 failed
[ERROR] [1558245278.740181154]: jj: 0: 0.000000: min: -inf, max: inf
[ERROR] [1558245278.740200916]: jj: 1: -0.785000: min: -inf, max: inf
[ERROR] [1558245278.740223415]: jj: 2: 0.000000: min: -inf, max: inf
[ERROR] [1558245278.740237277]: jj: 3: -2.356000: min: -1.000000, max: 1.000000
[ERROR] [1558245278.740251680]: jj: 4: 0.000000: min: -1.000000, max: 1.000000
[ERROR] [1558245278.740263902]: jj: 5: 1.571000: min: -1.000000, max: 1.000000
[ERROR] [1558245278.740279026]: jj: 6: 0.785000: min: -1.000000, max: 1.000000
```
which suggests that for some reason the planning request joint state does not include the floating joint present between the panda base link and the world. This error is present throughout the entire planning pipeline (I've copy pasted the checking code in OMPL's manager and the highest level planning pipeline call), and OMPL somehow handles this fine, but CHOMP will check the start state and fail. Still need to find out either:
* why the planning scene expects the robot state to include the floating joint but doesn't
* how OMPL still plans fine in this case
* what's the correct way to resolve this.

rhaschke:
[...]

PR #1473, opened 2019-05-21
Fix several Chomp issues

### LINKED ISSUE #1452: chomp/OptimizerAdapter: Start state is empty
### Description
Attempted to use `chomp/OptimizerAdapter` after `OMPL`. Followed tutorials and faced the problem of `Start state is empty`. The example print-out can be seen at the bottom,

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Binary build, with `moveit_planners` built from source
* `ros-melodic-moveit` binaries `1.0.1-0bionic.20190320.192509`
* `moveit_planners` from `melodic-devel`

### Steps to reproduce

Start the tutorial through the **Getting Started** portion of [MoveIt! Tutorials](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/getting_started/getting_started.html).

After setting up the workspace, start following the steps of [Planning Adapter Tutorials](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/planning_adapters/planning_adapters_tutorial.html), under **Running OMPL as a pre-processor for CHOMP**.

I noticed that there might be some errors in this section, in order to use this setup with `chomp/OptimizerAdapter`, we would need the `ompl` planner to be called instead of `chomp`, so I believe an error we should be launching
```
roslaunch panda_moveit_config dmeo.launch
```
instead of 
```
roslaunch panda_moveit_config demo_chomp.launch
```

The next error I got was 
```
[ERROR] [1556590982.275248450]: Exception while loading planning adapter plugin 'chomp/OptimizerAdapter': According to the loaded plugin descriptions the class chomp/OptimizerAdapter with base class type planning_request_adapter::PlanningRequestAdapter does not exist. Declared types are  default_planner_request_adapters/AddIterativeSplineParameterization default_planner_request_adapters/AddTimeOptimalParameterization default_planner_request_adapters/AddTimeParameterization default_planner_request_adapters/Empty default_planner_request_adapters/FixStartStateBounds default_planner_request_adapters/FixStartStateCollision default_planner_request_adapters/FixStartStatePathConstraints default_planner_request_adapters/FixWorkspaceBounds industrial_trajectory_filters/AddSmoothingFilter industrial_trajectory_filters/NPointFilter industrial_trajectory_filters/UniformSampleFilter
```

After digging around, I found that the `chomp/OptimizerAdapter` was not built in the binaries, so I did the following
```
cd /tmp
git clone -b melodic-devel https://github.com/ros-planning/moveit
cp -r /tmp/moveit/moveit_planners ~/ws_moveit/src/.
```
Rebuilding them again now,
```
cd ~/ws_moveit
catkin clean -y
catkin build
```
I then got a dependency issue, which was solved with
```
sudo apt-get install ros-melodic-moveit-resources
catkin build
```
The workspace built without any issues, however when I launched the demo via
```
roslaunch panda_moveit_config demo.launch
```
adding the **Motion Planning** panel in `rviz`, under **Planning** tab -> **Planning Group**: panda_arm ->**Goal State**: random_valid, **Plan and Execute**. I got the following logs

### Console output and the files I edited from tutorials
https://gist.github.com/aaronchongth/116e85a65d26e3daef7c47fbed98e209

rhaschke:
This PR comprises several independent commits to improve the Chomp code, namely
- some code cleanup, reducing redundancy and explicit loops
- fix start state handling: 
  - An empty start state should be accepted. In this case, the current state should be used. Fixes #1452
  - Checking start state constraints was plain wrong, resulting in #1470.
- Avoid extra conversion from/to MotionPlanResponse _message_

### DIFF
--- a/moveit_planners/chomp/chomp_interface/test/chomp_moveit_test.cpp
+++ b/moveit_planners/chomp/chomp_interface/test/chomp_moveit_test.cpp
@@ -93,7 +93,7 @@ TEST_F(CHOMPMoveitTest, noStartState)
   move_group_.setJointValueTarget(std::vector<double>({ 0.2, 0.2 }));
 
   moveit::planning_interface::MoveItErrorCode error_code = move_group_.plan(my_plan_);
-  EXPECT_EQ(error_code.val, moveit::planning_interface::MoveItErrorCode::INVALID_ROBOT_STATE);
+  EXPECT_EQ(error_code.val, moveit::planning_interface::MoveItErrorCode::SUCCESS);
 }
 
 TEST_F(CHOMPMoveitTest, collisionAtEndOfPath)
--- a/moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_planner.h
+++ b/moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_planner.h
@@ -38,8 +38,8 @@
 #define _CHOMP_PLANNER_H_
 
 #include <chomp_motion_planner/chomp_parameters.h>
-#include <moveit_msgs/MotionPlanDetailedResponse.h>
-#include <moveit_msgs/MotionPlanRequest.h>
+#include <moveit/planning_interface/planning_request.h>
+#include <moveit/planning_interface/planning_response.h>
 #include <moveit/planning_scene/planning_scene.h>
 
 namespace chomp
@@ -50,9 +50,9 @@ public:
   ChompPlanner() = default;
   virtual ~ChompPlanner() = default;
 
-  // TODO: switch API from moveit_msgs to MotionPlanRequest / MotionPlanDetailedResponse
-  bool solve(const planning_scene::PlanningSceneConstPtr& planning_scene, const moveit_msgs::MotionPlanRequest& req,
-             const ChompParameters& params, moveit_msgs::MotionPlanDetailedResponse& res) const;
+  bool solve(const planning_scene::PlanningSceneConstPtr& planning_scene,
+             const planning_interface::MotionPlanRequest& req, const ChompParameters& params,
+             planning_interface::MotionPlanDetailedResponse& res) const;
 };
 }
 
--- a/moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_planner.h
+++ b/moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_planner.h
@@ -47,8 +47,8 @@ namespace chomp
 class ChompPlanner
 {
 public:
-  ChompPlanner();
-  virtual ~ChompPlanner(){};
+  ChompPlanner() = default;
+  virtual ~ChompPlanner() = default;
 
   // TODO: switch API from moveit_msgs to MotionPlanRequest / MotionPlanDetailedResponse
   bool solve(const planning_scene::PlanningSceneConstPtr& planning_scene, const moveit_msgs::MotionPlanRequest& req,
--- a/moveit_planners/chomp/chomp_motion_planner/src/chomp_planner.cpp
+++ b/moveit_planners/chomp/chomp_motion_planner/src/chomp_planner.cpp
@@ -43,8 +43,6 @@
 
 namespace chomp
 {
-ChompPlanner::ChompPlanner() = default;
-
 bool ChompPlanner::solve(const planning_scene::PlanningSceneConstPtr& planning_scene,
                          const moveit_msgs::MotionPlanRequest& req, const chomp::ChompParameters& params,
                          moveit_msgs::MotionPlanDetailedResponse& res) const
--- a/moveit_planners/chomp/chomp_motion_planner/src/chomp_planner.cpp
+++ b/moveit_planners/chomp/chomp_motion_planner/src/chomp_planner.cpp
@@ -237,7 +237,7 @@ bool ChompPlanner::solve(const planning_scene::PlanningSceneConstPtr& planning_s
   create_time = ros::WallTime::now();
   // assume that the trajectory is now optimized, fill in the output structure:
 
-  ROS_DEBUG_NAMED("chomp_planner", "Output trajectory has %d joints", trajectory.getNumJoints());
+  ROS_DEBUG_NAMED("chomp_planner", "Output trajectory has %zd joints", trajectory.getNumJoints());
 
   res.trajectory.resize(1);
 
--- a/moveit_planners/chomp/chomp_motion_planner/src/chomp_planner.cpp
+++ b/moveit_planners/chomp/chomp_motion_planner/src/chomp_planner.cpp
@@ -47,6 +47,7 @@ bool ChompPlanner::solve(const planning_scene::PlanningSceneConstPtr& planning_s
                          const moveit_msgs::MotionPlanRequest& req, const chomp::ChompParameters& params,
                          moveit_msgs::MotionPlanDetailedResponse& res) const
 {
+  ros::WallTime start_time = ros::WallTime::now();
   if (!planning_scene)
   {
     ROS_ERROR_STREAM_NAMED("chomp_planner", "No planning scene initialized.");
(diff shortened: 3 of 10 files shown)


BryceStevenWilley:
Looks good on a brief scan, I'll have some time to take a look at this PR this weekend.

moveit_planners/chomp/chomp_interface/test/chomp_moveit_test.cpp
```diff
@@ -88,14 +88,6 @@ TEST_F(CHOMPMoveitTest, cartesianGoal)
   EXPECT_EQ(error_code.val, moveit::planning_interface::MoveItErrorCode::INVALID_GOAL_CONSTRAINTS);
 }
 
-TEST_F(CHOMPMoveitTest, noStartState)
```
bmagyar: How are invalid start states handled now?

moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_planner.h
```diff
@@ -38,21 +38,21 @@
...
 #include <moveit/planning_scene/planning_scene.h>
 
 namespace chomp
 {
 class ChompPlanner
 {
 public:
-  ChompPlanner();
-  virtual ~ChompPlanner(){};
+  ChompPlanner() = default;
```
bmagyar: Is this even worth keeping is we don't do anything in it?

moveit_planners/chomp/chomp_motion_planner/src/chomp_planner.cpp
```diff
@@ -43,64 +43,58 @@
...
     ROS_ERROR_STREAM("Only joint-space goals are supported");
-    res.error_code.val = moveit_msgs::MoveItErrorCodes::INVALID_GOAL_CONSTRAINTS;
+    res.error_code_.val = moveit_msgs::MoveItErrorCodes::INVALID_GOAL_CONSTRAINTS;
     return false;
   }
 
-  int goal_index = trajectory.getNumPoints() - 1;
-  trajectory.getTrajectoryPoint(goal_index) = trajectory.getTrajectoryPoint(0);
-  sensor_msgs::JointState js;
+  size_t goal_index = trajectory.getNumPoints() - 1;
```
bmagyar: didn't see any other uses for it
```suggestion
  const size_t goal_index = trajectory.getNumPoints() - 1;
```

moveit_planners/chomp/chomp_motion_planner/src/chomp_trajectory.cpp
```diff
@@ -297,27 +229,30 @@ bool ChompTrajectory::fillInFromTrajectory(moveit_msgs::MotionPlanDetailedRespon
   {
     // perform a decimation sampling in this block if the number of trajectory points in the MotionPlanDetailedResponse
     // res object is more than the number of points in the CHOMP trajectory
-    double decimation_sampling_factor = ((double)num_response_points) / ((double)num_chomp_trajectory_points);
+    double decimation_sampling_factor = ((double)num_input_points) / ((double)num_chomp_trajectory_points);
```
bmagyar:
```suggestion
    const double decimation_sampling_factor = ((double)num_input_points) / ((double)num_chomp_trajectory_points);
```

moveit_planners/chomp/chomp_motion_planner/src/chomp_trajectory.cpp
```diff
@@ -231,64 +169,58 @@ void ChompTrajectory::fillInMinJerk()
...
+  if (num_input_points < 2)
     return false;
 
-  // get the number of joints for each robot state
-  int num_joints_trajectory = trajectory_msg.joint_trajectory.points[0].positions.size();
-
   // variables for populating the CHOMP trajectory with correct number of trajectory points
-  int repeated_factor = num_chomp_trajectory_points / num_response_points;
-  int repeated_balance_factor = num_chomp_trajectory_points % num_response_points;
+  unsigned int repeated_factor = num_chomp_trajectory_points / num_input_points;
```
bmagyar:
```suggestion
  const unsigned int repeated_factor = num_chomp_trajectory_points / num_input_points;
```

moveit_planners/chomp/chomp_motion_planner/src/chomp_trajectory.cpp
```diff
@@ -231,64 +169,58 @@ void ChompTrajectory::fillInMinJerk()
...
+  unsigned int repeated_balance_factor = num_chomp_trajectory_points % num_input_points;
```
bmagyar:
```suggestion
  const unsigned int repeated_balance_factor = num_chomp_trajectory_points % num_input_points;
```

moveit_planners/chomp/chomp_motion_planner/src/chomp_trajectory.cpp
```diff
@@ -231,64 +169,58 @@ void ChompTrajectory::fillInMinJerk()
...
+bool ChompTrajectory::fillInFromTrajectory(const robot_trajectory::RobotTrajectory& trajectory)
 {
-  // create a RobotTrajectory msg to obtain the trajectory from the MotionPlanDetailedResponse object
-  moveit_msgs::RobotTrajectory trajectory_msg = res.trajectory[0];
-
   // get the default number of points in the CHOMP trajectory
-  int num_chomp_trajectory_points = (*this).getNumPoints();
-  // get the number of points in the response trajectory obtained from OMPL
-  int num_response_points = trajectory_msg.joint_trajectory.points.size();
+  size_t num_chomp_trajectory_points = (*this).getNumPoints();
```
bmagyar:
```suggestion
  const size_t num_chomp_trajectory_points = (*this).getNumPoints();
```

moveit_planners/chomp/chomp_motion_planner/src/chomp_trajectory.cpp
```diff
@@ -231,64 +169,58 @@ void ChompTrajectory::fillInMinJerk()
...
+  // get the number of points in the input trajectory
+  size_t num_input_points = trajectory.getWayPointCount();
```
bmagyar:
```suggestion
  const size_t num_input_points = trajectory.getWayPointCount();
```

moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_planner.h
rhaschke: Otherwise, the default constructor wouldn't be created.

rhaschke:
I addressed your comments, @bmagyar. Thanks a lot. Indeed, it's always a good idea to declare const variables.

moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_planner.h
bmagyar: I think it should. When you don't declare it at all, you get the default constructor. When you have one declared, the compiler won't generate the default one for you but the `default` keyword can be used to force generating it. Unfortunately I don't have a setup available ATM where I could double check it but I'm pretty certain.

The story about the destructor is slightly different as if you declare it virtual you are allowing derived classes to do a proper cleanup.

rhaschke: I believe the default constructor is only created by default if there are no other constructors defined. But, I might be wrong. Nevertheless, it's not wrong to explicitly declare that we want to have the default constructor :wink:

bmagyar: Yeah and as far as I see this is the only constructor being defined. This conversation is dragging for too long, let's go to merge :D

moveit_planners/chomp/chomp_interface/test/chomp_moveit_test.cpp
bmagyar: Thanks

bmagyar:
@BryceStevenWilley , next time you are around and happy with the PR, feel free to merge :+1:

Issue #1471, opened 2019-05-21
jog_arm first order low-pass filter


AndyZe:
Per conversation in #1468, the 2nd-order filter that is currently used is known to overshoot. A first-order filter may be a better fit (as well as more efficient).

https://dankelley.github.io/r/2014/01/15/butterworth-filters.html

SansoneG:
Changing
```cpp
double new_filtered_msrmt =
     (1. / (1. + filter_coeff_ * filter_coeff_ + 1.414 * filter_coeff_)) *
     (previous_measurements_[2] + 2. * previous_measurements_[1] + previous_measurements_[0] -
      (filter_coeff_ * filter_coeff_ - 1.414 * filter_coeff_ + 1.) * previous_filtered_measurements_[1] -
      (-2. * filter_coeff_ * filter_coeff_ + 2.) * previous_filtered_measurements_[0]);
```
to
```cpp
double new_filtered_msrmt =
    (1. / (1. + filter_coeff_)) *
    (previous_measurements_[1] + previous_measurements_[0] -
     (-1. * filter_coeff_ + 1.) * previous_filtered_measurements_[0]);
```
in `low_pass_filter.cpp:66-70` seems to solve the issue.

Here is a plot of the new filter:
![Bw_lp_1](https://user-images.githubusercontent.com/50736605/58411744-480f4500-8075-11e9-901b-e83020cf6044.png)

jschleicher:
@SansoneG Please provide a pull-request.

SansoneG:
> @SansoneG Please provide a pull-request.

#1483

rhaschke:
Fixed in #1483.

PR #1483, opened 2019-05-27
Change 2nd order Butterworth low pass filter to 1st order

SansoneG:
### Description

Fixes Issue #1471 about filter overshooting by changing the 2nd order Butterworth low pass filter to 1st order.

### DIFF
--- a/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
+++ b/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
@@ -45,6 +45,8 @@ namespace jog_arm
 /**
  * Class LowPassFilter - Filter a signal to soften jerks.
  * This is a first-order Butterworth low-pass filter.
+ * 
+ * TODO: Use ROS filters package (http://wiki.ros.org/filters, https://github.com/ros/filters)
  */
 class LowPassFilter
 {
--- a/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
+++ b/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
@@ -44,8 +44,7 @@ namespace jog_arm
 {
 /**
  * Class LowPassFilter - Filter a signal to soften jerks.
- * This is a second-order Butterworth low-pass filter.
- * See https://ccrma.stanford.edu/~jos/filters/Example_Second_Order_Butterworth_Lowpass.html
+ * This is a first-order Butterworth low-pass filter.
  */
 class LowPassFilter
 {
--- a/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
+++ b/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
@@ -55,8 +55,8 @@ public:
   void reset(double data);
 
 private:
-  double previous_measurements_[3] = { 0., 0., 0. };
-  double previous_filtered_measurements_[2] = { 0., 0. };
+  double previous_measurements_[2] = { 0., 0. };
+  double previous_filtered_measurement = 0.;
   // Larger filter_coeff-> more smoothing of jog commands, but more lag.
   // Rough plot, with cutoff frequency on the y-axis:
   // https://www.wolframalpha.com/input/?i=plot+arccot(c)
--- a/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
+++ b/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
@@ -45,7 +45,7 @@ namespace jog_arm
 /**
  * Class LowPassFilter - Filter a signal to soften jerks.
  * This is a first-order Butterworth low-pass filter.
- * 
+ *
  * TODO: Use ROS filters package (http://wiki.ros.org/filters, https://github.com/ros/filters)
  */
 class LowPassFilter
--- a/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
+++ b/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
@@ -56,7 +56,7 @@ public:
 
 private:
   double previous_measurements_[2] = { 0., 0. };
-  double previous_filtered_measurement = 0.;
+  double previous_filtered_measurement_ = 0.;
   // Larger filter_coeff-> more smoothing of jog commands, but more lag.
   // Rough plot, with cutoff frequency on the y-axis:
   // https://www.wolframalpha.com/input/?i=plot+arccot(c)
--- a/moveit_experimental/jog_arm/src/jog_arm/low_pass_filter.cpp
+++ b/moveit_experimental/jog_arm/src/jog_arm/low_pass_filter.cpp
@@ -61,9 +61,8 @@ double LowPassFilter::filter(double new_measurement)
   previous_measurements_[0] = new_measurement;
 
   double new_filtered_msrmt =
-      (1. / (1. + filter_coeff_)) *
-      (previous_measurements_[1] + previous_measurements_[0] -
-       (-1. * filter_coeff_ + 1.) * previous_filtered_measurement);
+      (1. / (1. + filter_coeff_)) * (previous_measurements_[1] + previous_measurements_[0] -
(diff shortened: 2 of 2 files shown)


moveit_experimental/jog_arm/src/jog_arm/low_pass_filter.cpp
```diff
@@ -64,10 +64,9 @@ double LowPassFilter::filter(double new_measurement)
   previous_measurements_[0] = new_measurement;
```
jschleicher: You can drop `previous_measurements_[2]` in the buffer and in reset():
https://github.com/ros-planning/moveit/blob/a0c3595f3180bdaf9d3b7101141dc135e15f763f/moveit_experimental/jog_arm/src/jog_arm/low_pass_filter.cpp#L53

and decrease the size in the header file:
https://github.com/ros-planning/moveit/blob/a0c3595f3180bdaf9d3b7101141dc135e15f763f/moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h#L58-L59

moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
```diff
@@ -55,8 +55,8 @@ class LowPassFilter
   void reset(double data);
 
 private:
-  double previous_measurements_[3] = { 0., 0., 0. };
-  double previous_filtered_measurements_[2] = { 0., 0. };
+  double previous_measurements_[2] = { 0., 0. };
+  double previous_filtered_measurement = 0.;
```
AndyZe: Should be previous_filtered_measurement_

SansoneG:
> 
> 
> I'd like to check this equation in Matlab before it gets merged.

Once that is done, I can also change the comment in the header file to say that it is indeed a first order Butterworth filter.

felixvd:
Why was this originally a 2nd order filter? Has anyone looked into the history and can explain?

AndyZe:
It was just re-used code from an older project. I haven't forgotten to test this...

moveit_experimental/jog_arm/src/jog_arm/low_pass_filter.cpp
```diff
@@ -50,28 +50,22 @@ void LowPassFilter::reset(double data)
...
-  previous_measurements_[2] = previous_measurements_[1];
   previous_measurements_[1] = previous_measurements_[0];
   previous_measurements_[0] = new_measurement;
 
   double new_filtered_msrmt =
-      (1. / (1. + filter_coeff_ * filter_coeff_ + 1.414 * filter_coeff_)) *
-      (previous_measurements_[2] + 2. * previous_measurements_[1] + previous_measurements_[0] -
-       (filter_coeff_ * filter_coeff_ - 1.414 * filter_coeff_ + 1.) * previous_filtered_measurements_[1] -
-       (-2. * filter_coeff_ * filter_coeff_ + 2.) * previous_filtered_measurements_[0]);
+      (1. / (1. + filter_coeff_)) * (previous_measurements_[1] + previous_measurements_[0] -
```
AndyZe: Should be (Matlab syntax):

new_filtered_msrmt = (1 + c / (2*pi))^-1 * (previous_measurements_[1] + previous_measurements_[0] - (1 - c/(2*pi)) * previous_filtered_measurement_);

AndyZe:
OK, I tested a 1st-order Butterworth filter with Matlab. Made a suggestion on the equations.

filter_plots

gavanderhoorn:
Would it perhaps be an idea to use the [filters](http://wiki.ros.org/filters) package here?

AndyZe:
> Would it perhaps be an idea to use the [filters](http://wiki.ros.org/filters) package here?

I see that a TransferFunctionFilter is available. That could work. Making it configurable to different cutoff frequencies might be tricky.

AndyZe:
@SansoneG don't forget to do this:

> Once that is done, I can also change the comment in the header file to say that it is indeed a first order Butterworth filter.

moveit_experimental/jog_arm/src/jog_arm/low_pass_filter.cpp
```diff
@@ -50,28 +51,22 @@ void LowPassFilter::reset(double data)
...
   previous_measurements_[1] = previous_measurements_[0];
   previous_measurements_[0] = new_measurement;
 
   double new_filtered_msrmt =
-      (1. / (1. + filter_coeff_ * filter_coeff_ + 1.414 * filter_coeff_)) *
-      (previous_measurements_[2] + 2. * previous_measurements_[1] + previous_measurements_[0] -
-       (filter_coeff_ * filter_coeff_ - 1.414 * filter_coeff_ + 1.) * previous_filtered_measurements_[1] -
-       (-2. * filter_coeff_ * filter_coeff_ + 2.) * previous_filtered_measurements_[0]);
+      (1. / (1. + filter_coeff_ / (2 * M_PI))) * (previous_measurements_[1] + previous_measurements_[0] -
+                                     (-1. * filter_coeff_ / (2 * M_PI) + 1.) * previous_filtered_measurement_);
```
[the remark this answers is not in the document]
AndyZe: I tested carefully with this equation. Did not see any issues:

```
  double new_filtered_msrmt =
      (1. / (1. + filter_coeff_)) * (previous_measurements_[1] + previous_measurements_[0] -
(-filter_coeff_ + 1.) * previous_filtered_measurement_);
```

Top graph is the input to the jogger from my joystick, bottom is the outgoing command to the robot.
jog_plots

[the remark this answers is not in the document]
AndyZe: ^Note, different timescales on the 2 plots

SansoneG:
I was not sure if you meant i should move the constant into the constructor (so the computation is done only once instead of each iteration) or remove it completely (maybe it serves the purpose of giving the filter_coeff_ a certain unit). For now I decided to remove it completely.

AndyZe:
@SansoneG Can you assign a few code reviewers so we can finally get this merged? It's a button in the top right that looks like a gear. Just pick a few MoveIt maintainers at random - I'm not aware of any who have experience with digital filtering. Here's a [list](https://moveit.ros.org/about/) to pick from.

To summarize, the benefits of this PR are:

- Slightly more efficient (in computation and memory)
- No overshoot
- There's a slower "roll off" near the cutoff frequency, which is good because we don't want a drastic change in behavior.

SansoneG:
Unfortunately that button is not present on my end. I think the reason for that is stated in [this StackOverflow answer](https://stackoverflow.com/questions/32262295/adding-a-reviewer-to-a-github-pull-request).
I will just have to mention them here:
@gavanderhoorn, @BryceStevenWilley, @davetcoleman Could one of you review this pull request please?

moveit_experimental/jog_arm/include/jog_arm/low_pass_filter.h
```diff
@@ -44,8 +44,7 @@ namespace jog_arm
 {
 /**
  * Class LowPassFilter - Filter a signal to soften jerks.
- * This is a second-order Butterworth low-pass filter.
- * See https://ccrma.stanford.edu/~jos/filters/Example_Second_Order_Butterworth_Lowpass.html
+ * This is a first-order Butterworth low-pass filter.
```
davetcoleman: Could you add a TODO suggested we consolidate our filter to use 
https://github.com/ros/filters

As @gavanderhoorn suggested. I also opened this related issue: https://github.com/ros/filters/issues/28

jschleicher:
@BryceStevenWilley, @felixvd awaits2ndreview please.

I'd appreciate, if this finally gets merged.

AndyZe:
I still get asked about this from time to time. Here's the MATLAB code for the filter:

```
function output = first_order_filter(c, input)
  % c: filter parameter related to cutoff frequency and sampling rate. Smaller c -->
  % higher cutoff frequency, less filtering
  % input: a vector of input data
  % returns output: a vector of output data

  % Preallocate an array for filter output
  output = zeros(size(input));

  for i=2 : length(input)
      output(i) = (1 + c / (2*pi))^-1 * (input(i) + input(i-1) - (1 - c/(2*pi)) * output(i-1));
  end
```

PR #1484, opened 2019-05-27
Integrating Bullet into MoveIt [WIP] (#1427)

j-petit:
### Description
This PR is work-in-progress for integrating Bullet with continuous collision detection support into MoveIt (see #1427). I started out from a copy of `collision_detection_fcl`. The following high level points are planned:

- [x] Integrating bullet collision manager into `CollisionRobot` for self-collision checks
- [x] Integrating bullet collision manager into `CollisionWorld` for robot world collision checks
- [ ] pass of all tests of `./test_bullet_collision_detection` which is the same test case as for using FCL but using `CollisionRobotBullet` and `CollisionWorldBullet` instead

Building requires a [forked version](https://github.com/j-petit/tesseract/tree/tesseract_reduced) of ros-industrial tesseract to be installed in the same catkin workspace. Tesseract includes a build from source of bullet, it therefore takes a while. Use `--recursive` option to clone the tesseract repo as it contains bullet as a submodule.

**Interesting changes are happening in collision_robot_bt.cpp and collision_world_bt.cpp**


moveit_core/collision_detection_bullet/src/collision_robot_bt.cpp
```diff
@@ -0,0 +1,346 @@
...
+      {
+        if (col_array[i] && col_array[i]->geometry)
+        {
+          shapes::ShapeConstPtr s = tesseract::tesseract_ros::constructShape(col_array[i]->geometry.get());
+          if (s)
+          {
+            shapes.push_back(s);
+            shape_poses.push_back(tesseract::tesseract_ros::urdfPose2Eigen(col_array[i]->origin));
+
+            if (s->type == shapes::MESH)
```
BryceStevenWilley: Add a `TODO(j-petit)` here, as eventually we want to distinguish between convex and non-convex meshes (or at least read if the mesh is convex and spit out a warning if it's not).

moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp
```diff
@@ -286,10 +286,15 @@ TEST_F(FclCollisionDetectionTester, AttachedBodyTester)
   crobot_->checkSelfCollision(req, res, robot_state, *acm_);
   ASSERT_FALSE(res.collision);
 
-  shapes::Shape* shape = new shapes::Box(.1, .1, .1);
+  shapes::Shape* shape = new shapes::Box(.2, .2, .2);
```
BryceStevenWilley: What's the reason for this change?

j-petit: I think Bullet wraps around its objects an additional contact distance threshold, therefore I increased the Box size in the FCL test to get the same resulting collisions. If I don't increase it for FCL, Bullet reports 4 collision whereas FCL only 2.

BryceStevenWilley: Interesting! I found [this bullet forum post](https://pybullet.org/Bullet/phpBB3/viewtopic.php?t=6575#p22856), which says the default "margin" is 0.04m, which would probably cause the same number of contacts to be reported as expanding the box by 0.1m. This margin will be interesting to tinker with in the coming weeks.

j-petit: image

moveit_core/collision_detection_bullet/src/collision_robot_bt.cpp
```diff
@@ -0,0 +1,354 @@
...
+    if (geoms_[i] && geoms_[i]->collision_geometry_)
+    {
+      bt_manager_.setCollisionObjectsTransform(
+          geoms_[i]->collision_geometry_data_->ptr.link->getName(),
+          state.getCollisionBodyTransform(geoms_[i]->collision_geometry_data_->ptr.link,
+                                          geoms_[i]->collision_geometry_data_->shape_index));
+    }
+
+  tesseract::ContactResultMap contact_map;
+  bt_manager_.contactTest(contact_map, tesseract::ContactTestType::FIRST, acm, req);
```
BryceStevenWilley: This line causes an error when compiling: the forked tesseract you have only takes the ACM as the third argument, and not the request. Did you push your latest commits there?

j-petit: Ah sorry, you are right, forgot to push my latest changes!

j-petit: I pushed them now.

moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp
j-petit: distance threshold is 0.05, I am not sure where exactly this is set yet.

moveit_core/CMakeLists.txt
```diff
@@ -55,6 +56,10 @@ COMPONENTS
   urdf
   visualization_msgs
   xmlrpcpp
+  tesseract_core
+  tesseract_collision
+  tesseract_msgs
+  tesseract_ros
```
davetcoleman: @ommmid and I are concerned about this... eventually there will be no dependencies on Tesseract, correct? This is just WIP?

moveit_core/collision_detection_bullet/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+target_link_libraries(collision_detector_bullet_plugin ${catkin_LIBRARIES} ${MOVEIT_LIB_NAME})
+
+
+install(TARGETS ${MOVEIT_LIB_NAME} collision_detector_bullet_plugin
+        LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+        ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION})
+
+install(DIRECTORY include/ DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION})
+
+install(FILES ../collision_detector_fcl_description.xml DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
```
davetcoleman: this should be 'bullet' right?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_common.h
```diff
@@ -0,0 +1,302 @@
...
+
+/* Author: Ioan Sucan */
+
+#ifndef MOVEIT_COLLISION_DETECTION_BT_COLLISION_COMMON_
+#define MOVEIT_COLLISION_DETECTION_BT_COLLISION_COMMON_
+
+#include <moveit/collision_detection/world.h>
+#include <moveit/collision_detection/collision_world.h>
+#include <moveit/macros/class_forward.h>
+#include <moveit/collision_detection_bullet/fcl_compat.h>
```
davetcoleman: all these references to fcl are temporary, right?

moveit_core/CMakeLists.txt
BryceStevenWilley: You're right, this is still a WIP. The plan is to initially depend on tesseract and see how that interface can be wrapped by the MoveIt collision interface. Then, we'll port over each piece that we need.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_common.h
j-petit: Yes

moveit_core/collision_detection_bullet/CMakeLists.txt
j-petit: Yep

moveit_core/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ find_library(LIBFCL_LIBRARIES fcl HINTS ${LIBFCL_PC_LIBRARY_DIRS})
 find_package(octomap REQUIRED)
 find_package(urdfdom REQUIRED)
 find_package(urdfdom_headers REQUIRED)
+find_package(Bullet REQUIRED)
```
henningkayser: Shouldn't we use bullet3_ros?

moveit_core/CMakeLists.txt
henningkayser: I'm also not convinced that wrapping another interface is the best approach since the interfaces seem to make different assumptions about collision world and attached objects.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_robot_bt.h
```diff
@@ -0,0 +1,100 @@
...
+
+#ifndef MOVEIT_COLLISION_DETECTION_BT_COLLISION_ROBOT_
+#define MOVEIT_COLLISION_DETECTION_BT_COLLISION_ROBOT_
+
+#include <moveit/collision_detection_bullet/collision_common.h>
+#include <tesseract_collision/bullet/bullet_discrete_simple_manager.h>
+
+namespace collision_detection
+{
+class CollisionRobotBt : public CollisionRobot
```
henningkayser: I would vote for CollisionRobotBullet (like package name)

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_robot_bt.h
```diff
@@ -0,0 +1,100 @@
...
+  void getAttachedBodyObjects(const robot_state::AttachedBody* ab, std::vector<void*>& geoms) const;
+
+  void checkSelfCollisionHelper(const CollisionRequest& req, CollisionResult& res, const robot_state::RobotState& state,
+                                const AllowedCollisionMatrix* acm) const;
+  void checkOtherCollisionHelper(const CollisionRequest& req, CollisionResult& res,
+                                 const robot_state::RobotState& state, const CollisionRobot& other_robot,
+                                 const robot_state::RobotState& other_state, const AllowedCollisionMatrix* acm) const;
+
+  /** @brief Bullet collision manager taken from tesseract*/
+  mutable tesseract::tesseract_bullet::BulletDiscreteSimpleManager bt_manager_;
```
henningkayser: Why is this mutable?

moveit_core/collision_detection_bullet/src/collision_robot_bt.cpp
```diff
@@ -0,0 +1,251 @@
...
+      }
+      bt_manager_.addCollisionObject(link.second->name, BodyType::ROBOT_LINK, shapes, shape_poses,
+                                     collision_object_types, true);
+    }
+  }
+}
+
+CollisionRobotBt::CollisionRobotBt(const CollisionRobotBt& other) : CollisionRobot(other)
+{
+  const CollisionRobotBt& other_bt = dynamic_cast<const CollisionRobotBt&>(other);
```
henningkayser: Functions like this seem to make it difficult to remove tesseract dependencies later without copying a lot of code.

moveit_core/CMakeLists.txt
BryceStevenWilley: That is the ultimate plan. This uses the same fork that Tesseract uses at the moment, just so Jens can get a checker up and comparable quickly.

moveit_core/collision_detection_bullet/src/collision_robot_bt.cpp
BryceStevenWilley: That's been the plan. Much of what Tesseract does (the contact managers) will have to be included, either by copying/porting code, or by dependencies. Rather than have Jens port every piece individually right now, we decided to keep much of it dependencies while Jens is running benchmarks to see what interface changes we might want to make as well.

BryceStevenWilley:
@henningkayser Using OpenCL is a good idea, but I'm afraid it's too ambitious for this project. From what I can tell, it's [poorly documented](https://pybullet.org/Bullet/phpBB3/viewtopic.php?t=10659) and possibly not tested very well? We'll keep track of the idea [here](https://github.com/orgs/ros-planning/projects/2#card-22441073).

moveit_core/CMakeLists.txt
j-petit: bullet3_ros is actually used, this is still there because initially I thought we can use the debian Bullet version. Will remove this.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_robot_bt.h
j-petit: In CollisionRobot, almost all functions are declared as const. When I now want to update the poses of the collision objects of the manager, I get an error because the functions through which CollisionRobot interacts are declared as `const`. However, I wanted to keep the CollisionRobot interface as unchanged as possible.

j-petit:
I haven't looked into the GPU support yet, but sounds definitely interesting. Thanks @henningkayser for reviewing.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_robot_bt.h
henningkayser: I see. For each collision check in FCL a new manager is initialized and the full robot model is constructed from scratch. I probably wouldn't do it like this, either. Maybe we can keep a const instance of the manager already populated with the robot model and just clone it each time we need it? Cloning shouldn't be too costly since there is not much stored.

moveit_core/CMakeLists.txt
henningkayser: Whatever makes more sense. If the debian version works just fine and is broadly available I don't see an issue with that.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_world_bt.h
```diff
@@ -0,0 +1,94 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2011, Willow Garage, Inc.
```
davetcoleman: All these headers can be renamed to you, with 2019

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_world_bt.h
```diff
@@ -0,0 +1,94 @@
...
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Ioan Sucan */
```
davetcoleman: I understand you are inheriting from Ioan's plugin template, but I think its safe to say this new plugin is your work and not his.

davetcoleman:
> Interesting changes are happening in collision_robot_bt.cpp and collision_world_bt.cpp

I think we might want to split this PR into smaller ones that we can review as you progress. Since you'll be working out of a branch that we'll do a final review of before merging to master, this seems safe. For example, split this PR into:
1) Basic copy of plugin from FCL, but renamed. This can be merged now.
2) Ongoing tweaking and magic as we move away from Tesseract

As it stands, this PR is hard to review due to its size.

j-petit:
Closed because superseded by #1499 and #1500.

PR #1490, opened 2019-06-02
KDL IK solver: fix handling of mimic joints

rhaschke:
This PR fixes an issue introduced in 5bedb7a31db8ab4ac7e474de762e5dfc42ad4d13:
- correctly count mimic joints
- don't resize extra_joint_weights vector

This fixes https://github.com/seed-solutions/aero-ros-pkg/issues/400.

### DIFF
--- a/moveit_kinematics/kdl_kinematics_plugin/src/chainiksolver_vel_mimic_svd.cpp
+++ b/moveit_kinematics/kdl_kinematics_plugin/src/chainiksolver_vel_mimic_svd.cpp
@@ -31,12 +31,10 @@ namespace
 unsigned int countMimicJoints(const std::vector<kdl_kinematics_plugin::JointMimic>& mimic_joints)
 {
   unsigned int num_mimic = 0;
-  unsigned int index = 0;
   for (const auto& item : mimic_joints)
   {
-    if (item.map_index != index)
+    if (!item.active)
       ++num_mimic;
-    ++index;
   }
   return num_mimic;
 }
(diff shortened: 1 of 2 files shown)


moveit_kinematics/kdl_kinematics_plugin/src/chainiksolver_vel_mimic_svd.cpp
```diff
@@ -31,12 +31,10 @@ namespace
 unsigned int countMimicJoints(const std::vector<kdl_kinematics_plugin::JointMimic>& mimic_joints)
 {
   unsigned int num_mimic = 0;
-  unsigned int index = 0;
   for (const auto& item : mimic_joints)
   {
-    if (item.map_index != index)
+    if (!item.active)
```
v4hn: There is a distinction between [active/passive joints](https://github.com/ros-planning/moveit/blob/master/moveit_core/robot_model/include/moveit/robot_model/joint_model.h#L508) and [mimic](https://github.com/ros-planning/moveit/blob/master/moveit_core/robot_model/include/moveit/robot_model/joint_model.h#L486) joints in MoveIt, although passive joints are not really handled anywhere aside from the srdf allowing for them.

Does this distinction not exist in the kdl solver?

rhaschke: @v4hn, sorry for the late reply. `item.active` here refers to [`JointMimic::active`](https://github.com/ros-planning/moveit/blob/master/moveit_kinematics/kdl_kinematics_plugin/include/moveit/kdl_kinematics_plugin/joint_mimic.hpp#L62L63) which is not directly related to [!JointModel::passive_](https://github.com/ros-planning/moveit/blob/master/moveit_core/robot_model/include/moveit/robot_model/joint_model.h#L508) you referenced to. For this reason, the semantics of this function is correct.
However, your question is legitimate as well: I think, the KDL solver doesn't have a notion of passive joints yet!
I think we should tackle this in another PR.

v4hn: True, I'm sorry I only briefly looked at the patch and didn't lookup the definitions before.

PR #1494, opened 2019-06-07
static transform publisher does not take a rate

mlautman:
### Description
The static_transform_publisher node in tf2_ros does not take a rate. The 10th argument was being treated as the child link.

https://github.com/ros/geometry2/blob/c4a64e0c6ea8fb0a8b0c6a032e509ebc449b0519/tf2_ros/src/static_transform_broadcaster_program.cpp#L75

### DIFF
--- a/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
@@ -1085,8 +1085,8 @@ void ConfigurationFilesWidget::loadTemplateStrings()
   {
     const srdf::Model::VirtualJoint& vj = config_data_->srdf_->virtual_joints_[i];
     if (vj.type_ != "fixed")
-      vjb << "  <node pkg=\"tf2_ros\" type=\"static_transform_publisher\" name=\"virtual_joint_broadcaster_" << i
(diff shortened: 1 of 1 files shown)


moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
```diff
@@ -1085,8 +1085,8 @@ void ConfigurationFilesWidget::loadTemplateStrings()
   {
     const srdf::Model::VirtualJoint& vj = config_data_->srdf_->virtual_joints_[i];
     if (vj.type_ != "fixed")
-      vjb << "  <node pkg=\"tf2_ros\" type=\"static_transform_publisher\" name=\"virtual_joint_broadcaster_" << i
-          << "\" args=\"0 0 0 0 0 0 " << vj.parent_frame_ << " " << vj.child_link_ << " 100\" />" << std::endl;
+      vjb << "<node pkg=\"tf2_ros\" type=\"static_transform_publisher\" name=\"virtual_joint_broadcaster_" << i
```
rhaschke: I guess we want to keep the indentation.
```suggestion
      vjb << "  <node pkg=\"tf2_ros\" type=\"static_transform_publisher\" name=\"virtual_joint_broadcaster_" << i
```

mlautman:
@rhaschke merge?

Issue #1368, opened 2019-02-28
nvidia-docker file needs to be updated.


MohmadAyman:
### Description

[nvidia-docker version 1 is now deprecated](https://github.com/NVIDIA/nvidia-docker/wiki/About-version-2.0) , our own [gui-docker](https://github.com/ros-planning/moveit/blob/melodic-devel/.docker/gui-docker) is based on v1.

### Your environment
* ROS Distro: [Melodic]
* OS Version: e.g. Ubuntu 16.04
* Source or Binary build?

felixvd:
I couldn't get the gui-docker script to work on Ubuntu 18.04. This needs fixing.

Installing nvidia-docker2 and following the [ROS.org wiki instructions](http://wiki.ros.org/docker/Tutorials/Hardware%20Acceleration#nvidia-docker2) worked for me(after [this fix](https://answers.ros.org/question/295552/cannot-reproduce-nvidia-docker2-installation-from-the-tutorial/)). Inserting the MoveIt image did not work for me immediately, but I can't reproduce the error and it works now.

I haven't seen a way to do it without building locally. The easiest way to do this might be to replace `gui-docker` with the instructions from the ROS wiki. I will submit a PR for it in a minute where we can discuss further.

Further reading: https://github.com/NVIDIA/nvidia-docker/issues/534

felixvd:
This was fixed in #1532 I believe.

PR #1495, opened 2019-06-08
Update docker instructions

felixvd:
### Description

The Docker instructions are outdated, as noted in #1368. I took the liberty of updating the docker-related scripts and instructions and making the container persistent. This PR is connected to https://github.com/ros-planning/moveit.ros.org/pull/282.

The change to a persistent container is for convenience and because it integrates better with my workflow :)  
Anyone who is competent enough to want anonymous containers will know how to change the script, and I think newcomers will appreciate this change.

### Checklist
- [X] Extended the tutorials / documentation, if necessary [reference](http://moveit.ros.org/documentation/contributing/)
- [] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers


felixvd:
Travis is failing because of #1496.

.docker/run-gui-moveit-docker
```diff
@@ -0,0 +1,44 @@
...
+    --env="QT_X11_NO_MITSHM=1" \
+    --volume="/tmp/.X11-unix:/tmp/.X11-unix:rw" \
+    -env="XAUTHORITY=$XAUTH" \
+    --volume="$XAUTH:$XAUTH" \
+    --volume="/home/parallels/moveit-dev/catkin_ws:/root/linked_ws_moveit" \
+    --runtime=runc \
+    --name local_moveit_container \
+#    -v /home/your_username/moveit_ms:/root/ws_moveit_linked \
+    -w /root/linked_ws_moveit \
+    my_moveit_image
```
BryceStevenWilley: The only way I could get this to run was by changing `my_moveit_image` to `local_moveit_image`. Should that be necessary? I kept getting `docker: Error response from daemon: pull access denied for my_moveit_image, repository does not exist or may require 'docker login'.` before switching.

felixvd: My bad, I didn't coordinate that change between the two PRs. `local_moveit_image` should be correct.

rhaschke:
Thanks @felixvd for taking the initiative on this. However, from my point of view, this PR has several issues:
- It requires to build a docker image and for this reason is tied to a specific base image.
However, the docker image just exports some environment variables, which can be achieved on-the-fly as well.
- It doesn't really employs the nvidia GPU as stated, because you use the default docker runtime `runc`.
- The script is overly complex.

I filed an alternative PR: #1532. Can you, please, give it a try?

felixvd:
>     * It requires to build a docker image and for this reason is tied to a specific base image.
>       However, the docker image just exports some environment variables, which can be achieved on-the-fly as well.

I just ported the OSRF instructions, so I assumed this was a set constraint. Good to know.

>     * It doesn't really employs the nvidia GPU as stated, because you use the default docker runtime `runc`.

Quite the silly oversight :)
 
>     * The script is overly complex.

I find the way the current `gui-docker` has to be used quite cumbersome. One has to remember and type out a lot of parameters that essentially never change, even though most people (or at least developers) who use the instructions just want to start up a sandbox to play around in with minimum effort. I made a script for my own workflow which I found very comfortable, so I wanted to propose to add it here.

If there's no opposition I would file another PR to make it an alternative, e.g. `persistent-gui-docker`.

rhaschke:
> I find the current `gui-docker` quite cumbersome. One has to remember and type out a lot of parameters.

I'm not quite sure about which parameters you are talking about. Except from the ones passed via gui-docker as well, you added:
- `-it`
- `--name local_moveit_container`
- `--volume="/home/parallels/moveit-dev/catkin_ws:/root/linked_ws_moveit"`
- `-w /root/linked_ws_moveit`
- `local_moveit_image`

Except `-it` which is common, but not mandatory, all of them are specific to your use case. I might want to run the image with different containers, mounting different volumes, etc. 
Your volume names don't even exist in other user's setups.

If you use gui-docker always in conjunction with a specific set of params, you are free to use your wrapper around it. But don't impose your settings on other people :wink:

rhaschke:
Closing in favor of #1532, which is more general and supports `nvidia-docker` 1+2, as well as DRI-based graphics cards.

felixvd:
Sheesh, no one is imposing anything :) Both the current scripts and the ones I suggested are light wrappers around docker for convenience. I only proposed another convenience layer, which I've tested in a few labs and with newcomers, because the desire for a persistent state in the container is very common from my experience.

My main gripe with `gui-docker` is that it forces you to remember the image name and the additional parameters. Default parameters would already be an improvement.

Would an additional `persistent-gui-docker` script or default parameters be of interest? I don't think we need a new issue to check that.

rhaschke:
I see a value in persistent containers indeed. However, you didn't argue yet for your specific set of "default" parameters, which IMHO are not at all generic.
I could imagine an option (e.g. `--container` or `-c`) for `gui-docker`, that allows continuing an existing container (with a specific name). To maintain flexibility, I would like
- to avoid a hard-coded container name, allowing for several containers to be used in parallel
  A default could still fall back to `default_container` if nothing was specified.
- to avoid any other default parameters
  If you run a container for the first time, you need to pass the image name and all other options.
  When continuing the container, I think/hope docker remembers all these options (they are tied to the container).

felixvd:
I wouldn't argue necessarily for including all of those default parameters. Especially the `--volume` parameter is hard-coded and not as elegant as I would like, I just haven't come up with a good way to make it easier. 

I would argue for a version of the script that "just works" without having to remember any parameters and commands as the user. You say "when continuing the container", but that's another docker command that isn't mentioned in the docker tutorial and users cannot be expected to know. Bundling all of that in a script that just takes you to the same persistent sandbox makes sense to me.

I am probably biased because that's the way I've set it up for people with no docker experience at all, and they were happy with it from what I could tell. Would be good to know what someone with less experience with docker thinks.

rhaschke:
All of your arguments (except `-it`) are specific to your setup and do not belong into a generic script.
The other features, I have implemented. See https://github.com/ros-planning/moveit/pull/1532#issuecomment-508711693. Let's continue the discussion there.

Issue #631, opened 2017-10-12
MoveGroup::Place ignores goal poses specified in end-effector frame.


mbharatheesha:
### Description
MoveGroup::Place interprets goal poses in the object reference frame despite setting `place_eef` to True in the `PlaceGoal` msg.

[Reference discussion on moveit-users google group](https://groups.google.com/forum/#!topic/moveit-users/MTLEB9diPMs)

[Related reference](https://groups.google.com/forum/#!searchin/moveit-users/place$20pose/moveit-users/_2_MGQ4jy7A/VDMxOABDY60J)

### Environment
* ROS Distro: [Indigo]
* OS Version: e.g. Ubuntu 14.04

### Steps to reproduce
* Specify a place pose in the end-effector frame and set the place_eef flag to True.
* Call the MoveGroup Place API using group.place()

### Expected behaviour
The place call should succeed and the robot (in simulation or reality) should move to the desired location and perform the desired action (unless there are obvious collisions)

### Actual behaviour
Non-obvious collisions are reported and place call fails.

v4hn:
Hey @mbharatheesha , thank you for the report.
I know the structure of the relevant code is not easy, but did you have a look at it?

Do you have an easy-to-setup example for this at hand?

It is definitely possible that this is a mistake in the code, but it might also break for your setup only.

mbharatheesha:
@v4hn:
> I know the structure of the relevant code is not easy, but did you have a look at it?

Yes, I did. And it comes from [here](https://github.com/ros-planning/moveit/blob/069f6c83f75c630b651350b678f45a0f7b6ec3d5/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L681).

Basically, the complete goal information from the user defined PlaceGoal msg is not used.

> Do you have an easy-to-setup example for this at hand?

Unfortunately not. But, this is something that can be produced on any setup by using the MoveGroup `place` API instead of the MoveIt place capability. More details are in this [discussion](https://groups.google.com/forum/#!topic/moveit-users/MTLEB9diPMs)

> It is definitely possible that this is a mistake in the code.

I think it is. I did fix it on my source installation, but I am not very proud of the fix because I basically bulldozed the goal construction by setting the `place_eef` flag. I did this because it was in a rush for another project.

But will prepare a clean PR for this hopefully during WMD17.

v4hn:
In that case I'll mark this as WMD candidate and assign it to you :-)

[...]

PR #1498, opened 2019-06-10
Add functions to set pick and place goal fields

mvieth:
### Description

Add functions to set fields in the pick and place goals, namely `place_eef` and `allow_gripper_support_collision`. Also some minor formatting changes by clang-format.
See also issue #631

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -313,14 +313,6 @@ public:
    * objects are allowed to touch the support surface */
   void setSupportSurfaceName(const std::string& name);
 
-  /** \brief For place operations, either the location of the object can be specified ( \e place_eef = false,
-   * this is the default ) or the location of the end effector ( \e place_eef = true ) */
-  void setPlaceEEF(const bool place_eef);
-
-  /** \brief For pick/place operations, specify whether collisions between the gripper and the support surface should
-   * be acceptable. By default, this is true */
-  void setAllowGripperSupportCollision(const bool allow_gripper_support_collision);
-
   /**
    * \name Setting a joint state target (goal)
    *
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -778,12 +778,12 @@ public:
   void constructMotionPlanRequest(moveit_msgs::MotionPlanRequest& request);
 
   /** \brief Build a PickupGoal for an object named \e object and store it in \e goal_out */
-  moveit_msgs::PickupGoal constructPickupGoal(const std::string& object, std::vector<moveit_msgs::Grasp>&& grasps,
+  moveit_msgs::PickupGoal constructPickupGoal(const std::string& object, std::vector<moveit_msgs::Grasp> grasps,
                                               bool plan_only);
 
   /** \brief Build a PlaceGoal for an object named \e object and store it in \e goal_out */
   moveit_msgs::PlaceGoal constructPlaceGoal(const std::string& object,
-                                            std::vector<moveit_msgs::PlaceLocation>&& locations, bool plan_only);
+                                            std::vector<moveit_msgs::PlaceLocation> locations, bool plan_only);
 
   /** \brief Convert a vector of PoseStamped to a vector of PlaceLocation */
   std::vector<moveit_msgs::PlaceLocation> posesToPlaceLocations(const std::vector<geometry_msgs::PoseStamped>& poses);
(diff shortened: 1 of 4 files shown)


moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -311,6 +311,14 @@ class MoveGroupInterface
    * objects are allowed to touch the support surface */
   void setSupportSurfaceName(const std::string& name);
 
+  /** \brief For place operations, either the location of the object can be specified ( \e place_eef = false,
+   * this is the default ) or the location of the end effector ( \e place_eef = true ) */
+  void setPlaceEEF(const bool place_eef);
```
BryceStevenWilley: No need to declare parameters const like this.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -311,6 +311,14 @@ class MoveGroupInterface
...
+
+  /** \brief For pick/place operations, specify whether collisions between the gripper and the support surface should
+   * be acceptable. By default, this is true */
+  void setAllowGripperSupportCollision(const bool allow_gripper_support_collision);
```
BryceStevenWilley: Same as above.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -524,6 +526,19 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
     support_surface_ = support_surface;
   }
 
+  void setPlaceEEF(const bool place_eef)
+  {
+    place_eef_ = place_eef;
+    ROS_INFO_NAMED("move_group_interface", "Place EEF: %s", place_eef_ ? "yes" : "no");
```
BryceStevenWilley: Not many of the other functions have logs like this. I'm fine with it, but it should be changed to debug, it'll probably just add to the noise most of the time.

rhaschke:
Having thought a little bit more about this PR, I suggest to remove the extra member variables and instead pass those flags as arguments to the corresponding `constructGoal()` and `pick()` / `place()` functions - providing the old hard-coded values as defaults. My reasoning for this is, that those flags usually will change from call to call. What do you think @BryceStevenWilley?

BryceStevenWilley:
I don't think those parameters should change every call. Whatever is providing the goals targets won't change it's transform math, so Place EEF shouldn't change. And, at least for every pick or every place, collision with support should also stay the same. But in general, I agree that less saved state in the class is better, so I'm okay with that implementation as well.

mvieth:
Maybe it would even make sense to expose the `constructGoal()` functions for the Pickup and Place Goals? There are more fields and options a user might want access to (e.g. `attached_object_touch_links`, `minimize_object_distance`, `allowed_touch_objects`), and making setter functions/optional arguments for all of them would not be so elegant. Instead, the goals could be optional arguments for the pick/place functions, so that a user could call the `constructGoal()`function, then change any option in the goal, and pass it on to pick/place.

rhaschke:
> Maybe it would even make sense to expose the `constructGoal()`

I like this idea. Maybe the function names should become more descriptive if exposed, i.e. `constructPickGoal` and `constructPlaceGoal`.

mvieth:
Ok, I pushed the proposed changes. Looking forward to your feedback @rhaschke @BryceStevenWilley . I did not remove the `setPlaceEEF()` and `setAllowGripperTouchSupport()` functions yet because they do not conflict with the new `pick()` and `place()` functions, but could do so.

rhaschke:
> I did not remove the setPlaceEEF() and setAllowGripperTouchSupport()

My original intention was to _get rid off_ the extra state variables :wink:
I pushed some extra commits to your PR branch to reflect my vision of this PR.
You can now decide whether to add those `place_eef`, `allow_touching_support` arguments to `constructPickGoal` and `constructPlaceGoal`.

mvieth:
I would change [this line](https://github.com/ros-planning/moveit/pull/1498/commits/b666b9864cf5ebecd5634777c4ddb398326d4a66#diff-91269c2c54a4fe24e4a4df9d6e8db959L1086) and the previous one to how they originally were, with `goal.allow_gripper_support_collision = true` (same for `constructPlaceGoal`).

> You can now decide whether to add those place_eef, allow_touching_support arguments to constructPickGoal and constructPlaceGoal

I would say that's not necessary any more because a user could just use `constructPickupGoal` and change the `place_eef` and `allow_gripper_support_collision` fields in the goal before calling `pick(PickupGoal)`

rhaschke:
> I would change this line and the previous one to how they originally were, with `goal.allow_gripper_support_collision = true`

I'm fine with this. Didn't noticed the original state.

v4hn:
Thanks Robert, I wasn't sure if you are ok with squashing your commits too.

PR #1501, opened 2019-06-11
Add get_jacobian_matrix to moveit_commander

7675t:
### Description

I'm using `GetPositoinIK` service call from Python and I need to check the manipulability of the robot arm, but it seems no way to get Jacobian matrix in moveit_commander. (http://answers.ros.org/question/324596/is-there-way-to-get-jacobian-or-manipulability-in-python-with-moveit/)

In this PR, I just add `get_jacobian_matrix` function in moveit_commander. I feel its better that the service call `GetPositionIK` contains the corresponding jacobian matrix, but it need to change the message contents, discouraging me to do so.

Please review and give me a hint of desirable direction for my need. I don't even know other people need this feature while definitely I do.

Thanks,

### Checklist

### DIFF
--- a/moveit_commander/test/python_moveit_commander.py
+++ b/moveit_commander/test/python_moveit_commander.py
@@ -94,6 +94,7 @@ class PythonMoveitCommanderTest(unittest.TestCase):
     def test_planning_scene_interface(self):
         planning_scene = PlanningSceneInterface()
 
+
 if __name__ == '__main__':
     PKGNAME = 'moveit_ros_planning_interface'
     NODENAME = 'moveit_test_python_moveit_commander'
--- a/moveit_commander/test/python_moveit_commander.py
+++ b/moveit_commander/test/python_moveit_commander.py
@@ -94,12 +94,6 @@ class PythonMoveitCommanderTest(unittest.TestCase):
     def test_planning_scene_interface(self):
         planning_scene = PlanningSceneInterface()
 
-    def test_get_jacobian_matrix(self):
-        current = self.group.get_current_joint_values()
-        matrix = np.array(self.group.get_jacobian_matrix(current))
-        self.assertEqual(matrix.shape[0], 6)
-        self.assertEqual(matrix.shape[1], 6)
-
 if __name__ == '__main__':
     PKGNAME = 'moveit_ros_planning_interface'
     NODENAME = 'moveit_test_python_moveit_commander'
--- a/moveit_commander/test/python_moveit_commander.py
+++ b/moveit_commander/test/python_moveit_commander.py
@@ -94,6 +94,12 @@ class PythonMoveitCommanderTest(unittest.TestCase):
     def test_planning_scene_interface(self):
         planning_scene = PlanningSceneInterface()
 
+    def test_get_jacobian_matrix(self):
+        current = self.group.get_current_joint_values()
+        matrix = np.array(self.group.get_jacobian_matrix(current))
+        self.assertEqual(matrix.shape[0], 6)
+        self.assertEqual(matrix.shape[1], 6)
+
 if __name__ == '__main__':
     PKGNAME = 'moveit_ros_planning_interface'
     NODENAME = 'moveit_test_python_moveit_commander'
--- a/moveit_commander/test/python_moveit_commander_ns.py
+++ b/moveit_commander/test/python_moveit_commander_ns.py
@@ -95,6 +95,7 @@ class PythonMoveitCommanderNsTest(unittest.TestCase):
         plan3 = self.plan(current)
         self.assertTrue(self.group.execute(plan3))
 
+
 if __name__ == '__main__':
     PKGNAME = 'moveit_ros_planning_interface'
     NODENAME = 'moveit_test_python_moveit_commander_ns'
--- a/moveit_commander/test/python_moveit_commander_ns.py
+++ b/moveit_commander/test/python_moveit_commander_ns.py
@@ -95,12 +95,6 @@ class PythonMoveitCommanderNsTest(unittest.TestCase):
         plan3 = self.plan(current)
         self.assertTrue(self.group.execute(plan3))
 
-    def test_get_jacobian_matrix(self):
-        current = self.group.get_current_joint_values()
-        matrix = np.array(self.group.get_jacobian_matrix(current))
-        self.assertEqual(matrix.shape[0], 6)
-        self.assertEqual(matrix.shape[1], 6)
-
 if __name__ == '__main__':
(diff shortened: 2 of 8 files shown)


BryceStevenWilley:
LGTM, I'll approve when Travis is working again. I don't think the `GetPositionIK` should contain the Jacobian, as that message interface is already very targeted towards IK only, adding another optional Jacobian return would make it too complicated. Which leaves the suggested PR, or adding a whole new `GetJacobian` service, which I feel is overkill.

moveit_commander/test/python_moveit_commander.py
```diff
@@ -94,6 +94,12 @@ def test_validation(self):
     def test_planning_scene_interface(self):
         planning_scene = PlanningSceneInterface()
 
+    def test_get_jacobian_matrix(self):
+        current = self.group.get_current_joint_values()
+        matrix = np.array(self.group.get_jacobian_matrix(current))
+        self.assertEqual(matrix.shape[0], 6)
+        self.assertEqual(matrix.shape[1], 6)
+
```
v4hn: Thank you for adding a simple test for it.
Why did you duplicate the same test code 4 times though?

Testing the dimensions of the Jacobian makes sense, but I would also suggest to test the (known) value of a specific arm configuration. Would that be too much work/ too unstable to add?

7675t:
Thank you @BryceStevenWilley @v4hn for your comments. Now I'm encouraged to go forward.

moveit_commander/test/python_moveit_commander.py
7675t: > Why did you duplicate the same test code 4 times though?

Yes, I felt strange but the other tests have same 4 entry in each file, so I just mimicked them. 
move_group and moveit_commander have same tests and I guess `_ns` tests are for namespace enabled. But if I don't need 4 copies, I want to add only one (I'm not sure which file  is appropriate to add).

> I would also suggest to test the (known) value of a specific arm configuration.

Agreed, it is not difficult to test the known values. Thanks.

v4hn: @mlautman, @rhaschke, @bmagyar I'm not sure who worked on these tests, I guess one of you.
Could anyone of you comment on the current copy&paste situation?

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
```diff
@@ -513,6 +513,27 @@ class MoveGroupInterfaceWrapper : protected py_bindings_tools::ROScppInitializer
...
   }
+
+  bp::list getJacobianMatrixPython(bp::list& joint_values)
+  {
+    std::vector<double> v = py_bindings_tools::doubleFromList(joint_values);
+    robot_state::RobotState state(getRobotModel());
+    state.setToDefaultValues();
+    auto group = state.getJointModelGroup(getName());
+    state.setJointGroupPositions(group, v);
+    Eigen::MatrixXd matrix = state.getJacobian(group);
```
rhaschke: Wouldn't it be possible to return the Jacobian as a `numpy` matrix?
In this case, there will be only one (costly) memory allocation instead of `rows+1`.

moveit_commander/test/python_moveit_commander.py
rhaschke: Indeed this code duplication is horrible. For sure the duplication between `moveit_commander/test` and `moveit_ros/planning_interface/test` is superfluous. Also, the namespace variant could be better handled via an argument to the correponding `.test` file. Looks like @willcbaker introduced these duplicates in #835. @willcbaker, could you please clean this up?

wb-applied: yup, definitely cringe-worthy

willcbaker: yup, definitely cringe-worthy

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
7675t: Thank you, I revised it  to return a numpy matrix. Firstly I thought to use boost.numpy, but libboost seems not support numpy module in Ubuntu 16.04, so I used eigenpy. The returned value is numpy.matrixlib.defmatrix.matrix (not numpy.ndarray), but I believe it is not a problem.

moveit_commander/test/python_moveit_commander.py
7675t: I removed duplicated my tests. Only test in `moveit_ros/planning_interface/test/python_move_group.py` remains.

@v4hn Thanks, I revised to check the jacobian matrix  by known values at initial pose.

moveit_ros/planning_interface/test/python_move_group.py
```diff
@@ -56,6 +56,22 @@ def test_validation(self):
         plan3 = self.plan(current)
         self.assertTrue(self.group.execute(plan3))
 
+    def test_get_jacobian_matrix(self):
+        current = self.group.get_current_joint_values()
+        matrix = self.group.get_jacobian_matrix(current)
```
rhaschke: Please rename `matrix` -> `result` and `answer` -> `expected`.

moveit_ros/planning_interface/test/python_move_group_ns.py
```diff
@@ -94,7 +94,6 @@ def test_validation(self):
         plan3 = self.plan(current)
         self.assertTrue(self.group.execute(plan3))
 
-
```
rhaschke: According to PEP8, this double linebreak was intended/required.

moveit_commander/test/python_moveit_commander_ns.py
```diff
@@ -95,7 +95,6 @@ def test_validation(self):
         plan3 = self.plan(current)
         self.assertTrue(self.group.execute(plan3))
 
-
```
rhaschke: According to PEP8, this double linebreak was intended/required.

moveit_ros/planning_interface/test/python_move_group_ns.py
7675t: Thanks!

moveit_ros/planning_interface/test/python_move_group.py
7675t: Thanks, I changed them.

moveit_commander/test/python_moveit_commander_ns.py
7675t: Thanks!

PR #1502, opened 2019-06-13
docker: rosdep update --include-eol-distros

rhaschke:
Because [ROS Indigo has reached EOL in April 2019](http://wiki.ros.org/Distributions), `rosdep update` doesn't pull any package information for Indigo anymore and our [docker images fail to build](https://cloud.docker.com/u/moveit/repository/docker/moveit/moveit/builds).
To fix this, I added the option `--include-eol-distros` to these calls.
Alternatively, we should disable and remove Indigo docker images. However, I think there might be still many people using these Indigo branches.

### DIFF
--- a/.docker/ci/Dockerfile
+++ b/.docker/ci/Dockerfile
@@ -30,7 +30,7 @@ RUN wstool init . && \
         ros-$ROS_DISTRO-rosbash \
         ros-$ROS_DISTRO-rospack && \
     # Download all dependencies of MoveIt!
-    rosdep update && \
+    rosdep update --include-eol-distros && \
     rosdep install -y --from-paths . --ignore-src --rosdistro ${ROS_DISTRO} --as-root=apt:false && \
     # Remove the source code from this container. TODO: in the future we may want to keep this here for further optimization of later containers
     cd .. && \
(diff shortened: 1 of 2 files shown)


v4hn:
I can't access the docker link you provide (404)? Is this a matter of docker privileges?
If so, could you add me there?

The change is [rather new](https://github.com/ros-infrastructure/rosdep/pull/647).
Did they really update the rosdep version back in their trusty builds or is the current pip version pulled in somewhere in the images?

rhaschke:
> Did they really update the rosdep version back in their trusty builds?

I guess so, because it just works with adding this option - also on Trusty / Indigo.

rhaschke:
My question was primarily: Do we want to continue supporting MoveIt on Indigo?

v4hn:
our pull-request guidelines state:

> Your pull-requests can target any branch of a MoveIt repository that supports an OSRF-supported ROS distribution.

So at least we discourage users from contributing to these old branches.
As far as I know we also restricted releases to ABI-compatible safety fixes a long time ago.
Let's leave it to the next maintainer meeting to decide the state of the branch.
It seems stupid to leave CI broken for such a trivial reason, so I merged your workaround for now.

PR #1503, opened 2019-06-13
ikfast: fix ROS apt-key in OpenRAVE docker image

rhaschke:
Due to new [ROS GPG keys](https://discourse.ros.org/t/new-gpg-keys-deployed-for-packages-ros-org/9454), the IKFast generator tool [fails in Travis](https://travis-ci.org/ros-planning/moveit/jobs/543158027). Hence, we need to update the keys used in the corresponding docker image as well.

Needs to be back-ported to Melodic.


rhaschke:
Merged this tiny change after local testing, because it was a blocker for all Travis builds.

PR #1500, opened 2019-06-10
Adding Bullet and Tesseract into MoveIt [WIP]

j-petit:
### Description
This PR is a follow up to #1499 and fills the empty template with a bullet collision checker using the ported tesseract code. Tesseract is integrated into MoveIt except `tesseract_msgs` packages and `bullet3_ros`.

This is just a current snapshot and not anywhere near a final version. The tesseract code will be further reduced and dismantled.

**Building this requires [my branch of tesseract](https://github.com/j-petit/tesseract/tree/tesseract_reduced) to be in the catkin workspace.**

PR #1504, opened 2019-06-14
Adding Bullet and Tesseract into MoveIt

j-petit:
### Description
This PR is a follow up to #1499 and #1500 which all deal with adding Bullet as collision checker (see #1427) and fills the empty template using the modified tesseract/trajopt code. I used the given `CollisionWorld` and `CollisionRobot` structure.

This is an intermediate state where the continuous collision detection has not been added yet. See PR #1551 for the full bullet integration.

Building requires `libbullet-dev` and `libbullet-extras-dev` which can both be installed through apt.
Bullet passes all [tests which were originally designed for FCL](https://github.com/ros-planning/moveit/pull/1504/files#diff-86244afbfb53a2be7a884f6d519d0873) and I proposed in a separate PR #1543 a typed test suite so we can share the test scenarios across all collision checkers. As soon as this is merged, tests in this PR will be replaced with an instantiation of the typed test suite.

See [this high level flowchart](https://drive.google.com/file/d/1FNt09Ntj_5x3fje5KOykB8YIVo7XK4QX/view?usp=sharing) for the collision checking process using Bullet:
Bullet_collision_moveit

To further understand the collision checking process using Bullet see [this detailed flowchart](https://drive.google.com/file/d/19504a0UODZM2b2VoeVvclaVtB3TxlOpk/view?usp=sharing):
BulletChecking_detailed

Thanks for reviewing!


moveit_core/CMakeLists.txt
```diff
@@ -155,8 +159,7 @@ if(CATKIN_ENABLE_TESTING AND ENABLE_COVERAGE_TESTING)
 endif()
 
 include_directories(SYSTEM ${EIGEN3_INCLUDE_DIRS}
-                           ${LIBFCL_INCLUDE_DIRS}
-                           )
+                           ${LIBFCL_INCLUDE_DIRS})
```
BryceStevenWilley: Unnecessary change.

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
```diff
@@ -95,6 +107,15 @@ struct Contact
 
   /** \brief The type of the second body involved in the contact */
   BodyType body_type_2;
+
+  /** \brief The time until collision */
+  double cc_time;
```
BryceStevenWilley: Time is an overloaded word, especially in terms of trajectories (is it "geometric" time, where the robot moves at the same speed for the entire traj, or trajectory time, where each waypoint is time parameterized? I actually think this is from the range 0 to 1, correct?).  I think the better term would be `cc_percent_distance`, or `collision_percent_completed` (if 0 to 1) or just `cc_distance` if it's actually a distance.

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
```diff
@@ -95,6 +107,15 @@ struct Contact
...
+
+  /** \brief The continous collision type */
+  ContinouseCollisionType cc_type;
+
+  Eigen::Vector3d nearest_points[2];
+  Eigen::Vector3d cc_nearest_points[2];
```
BryceStevenWilley: When are `nearest_points` and `cc_nearest_points` different?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_common.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef MOVEIT_COLLISION_DETECTION_BT_COLLISION_COMMON_
+#define MOVEIT_COLLISION_DETECTION_BT_COLLISION_COMMON_
+
+#include <moveit/collision_detection/world.h>
+#include <moveit/collision_detection/collision_world.h>
+
+namespace collision_detection
+{
+}
```
BryceStevenWilley: Remove this file if it's going to remain empty (or just leave it and start filling it with stuff when refactoring).

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_robot_bt.h
```diff
@@ -0,0 +1,120 @@
...
+                                 const robot_state::RobotState& state, const CollisionRobot& other_robot,
+                                 const robot_state::RobotState& other_state, const AllowedCollisionMatrix* acm) const;
+
+  /** @brief Bullet collision manager taken from tesseract*/
+  mutable tesseract::tesseract_bullet::BulletDiscreteBVHManager bt_manager_;
+  // mutable tesseract::tesseract_bullet::BulletDiscreteSimpleManager bt_manager_;
+
+  mutable tesseract::tesseract_bullet::BulletCastBVHManager bt_manager_CCD_;
+
+  mutable const AllowedCollisionMatrix* acm_;
```
BryceStevenWilley: IMO, `mutable` should be reserved for things like `mutex`es and conditional variables, where the state of the object doesn't really change. Don't think `acm_` actually falling into that category though.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
```diff
@@ -0,0 +1,285 @@
...
+template <typename Key, typename Value>
+using AlignedMap = std::map<Key, Value, std::less<Key>, Eigen::aligned_allocator<std::pair<const Key, Value>>>;
+
+template <typename Key, typename Value>
+using AlignedUnorderedMap = std::unordered_map<Key, Value, std::hash<Key>, std::equal_to<Key>,
+                                               Eigen::aligned_allocator<std::pair<const Key, Value>>>;
+
+using VectorIsometry3d = AlignedVector<Eigen::Isometry3d>;
+using VectorVector4d = AlignedVector<Eigen::Vector4d>;
+using VectorVector3d = std::vector<Eigen::Vector3d>;
```
BryceStevenWilley: I'll give these another more throughout review soon, but I'm not a fan of these sort of typedefs that just redefine a simple and not complicated stl type like `std::vector<Eigen::Vector3d>`. Should remove before merging.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
```diff
@@ -0,0 +1,285 @@
...
+using TransformMap = AlignedMap<std::string, Eigen::Isometry3d>;
+
+typedef Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> TrajArray;
+
+struct AllowedCollisionMatrix
```
BryceStevenWilley: This should be replaced with Moveit's [AllowedCollisionMatrix](https://github.com/ros-planning/moveit/blob/822a991e0ea4f713fe672e3e337d7f0702bcc224/moveit_core/collision_detection/include/moveit/collision_detection/collision_matrix.h#L81).

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
```diff
@@ -0,0 +1,285 @@
...
+{
+enum BodyType
+{
+  ROBOT_LINK = 0,    /**< @brief These are links at the creation of the environment */
+  ROBOT_ATTACHED = 1 /**< @brief These are links that are added after initial creation */
+};
+}
+typedef BodyTypes::BodyType BodyType;
+
+namespace ContinouseCollisionTypes
```
BryceStevenWilley:
```suggestion
namespace ContinousCollisionTypes
```

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
```diff
@@ -0,0 +1,285 @@
...
+{
+  CCType_None,
+  CCType_Time0,
+  CCType_Time1,
+  CCType_Between
+};
+}
+typedef ContinouseCollisionTypes::ContinouseCollisionType ContinouseCollisionType;
+
+namespace ContactTestTypes
```
BryceStevenWilley: Most of these singe enum namespaces don't need to exist, and should be using `enum class` instead. See [this question](https://stackoverflow.com/a/18335862).

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_cast_simple_manager.h
```diff
@@ -0,0 +1,124 @@
...
+
+#include <moveit/collision_detection_bullet/tesseract/bullet_utils.h>
+#include <moveit/collision_detection_bullet/tesseract/continuous_contact_manager_base.h>
+
+namespace tesseract
+{
+namespace tesseract_bullet
+{
+/** @brief A simple implementaiton of a tesseract manager which does not use BHV */
+class BulletCastSimpleManager : public ContinuousContactManagerBase
```
BryceStevenWilley: Once BVH is working sufficiently, we should remove the simple manager classes to simplify things a bit.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/macros.h
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef TESSERACT_CORE_MACROS_H
+#define TESSERACT_CORE_MACROS_H
+#define TESSERACT_IGNORE_WARNINGS_PUSH                                                                                 \
+  _Pragma("GCC diagnostic push") _Pragma("GCC diagnostic ignored \"-Wall\"")                                           \
```
BryceStevenWilley: If possible, just get of this file, it's doing some compiler magic with includes that should be handled in `CMakeLists.txt`/clang tidy configs.

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -0,0 +1,90 @@
+/*********************************************************************
```
BryceStevenWilley: These TrajOpt files shouldn't be in this PR: can you retarget for `bullet_trajopt_feature_branch`?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/macros.h
felixvd: "get rid of", I assume

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_cast_simple_manager.h
j-petit: For the CCD, I didn't even try to port first the BulletCastSimpleManager and instead went directly for the BullectCastBVHManager as I think its implementation is cleaner (for example the looping over all collision objects is left to Bullet instead of doing the loop in tesseract). This means that BVH managers are working for both discrete and continuous cases. Do you think that I can remove the simple managers already?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/macros.h
j-petit: Done

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: Is it not ContinUOUS?

gavanderhoorn: yes

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_common.h
j-petit: it depends a little on how we proceed with the tesseract code as it contains common functionality in `bullet_utils.h` and `ros_tesseract_utils.h` which could have a place in `collision_common.h`. If we want to make the code as closely resemble the FCL version then we would fill `collision_common.h`. However, if we would like to more or less keep the tesseract structure as close to the "original" then it remains empty and could be removed.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: Agree, me neither.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
felixvd: Same with all the other occurrences naturally

BryceStevenWilley: Correct, I too cannot spell. 🙂

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_cast_simple_manager.h
BryceStevenWilley: Cool! Yes, if they're both working, I'd remove the simple managers.

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
j-petit: Yes, it is in the range of 0 to 1, but I have a hard time understanding what it actually is from [this](https://github.com/j-petit/moveit/blob/4797bc41755e4143d77cb289c29a87e42a087476/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h#L494) piece of code. Unfortunately, the tesseract code is lacking documentation [here](https://github.com/ros-industrial-consortium/tesseract/blob/740e6607423e2afbeba04b609e118043c86b7266/tesseract_core/include/tesseract_core/basic_types.h#L208)

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
j-petit: No they are different, for the CCD case the nearest_points are the ones calculated directly by bullet when checking the casted object against the non-casted object and the cc_nearest_points are calculated by some kind of projection which I can't really follow. See [here](https://github.com/j-petit/moveit/blob/4797bc41755e4143d77cb289c29a87e42a087476/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h#L494)

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/continuous_contact_manager_base.h
```diff
@@ -0,0 +1,205 @@
...
+
+#include <geometric_shapes/shapes.h>
+#include <memory>
+
+#include <moveit/collision_detection_bullet/tesseract/basic_types.h>
+#include <moveit/collision_detection/collision_common.h>
+
+namespace tesseract
+{
+class ContinuousContactManagerBase
```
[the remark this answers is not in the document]
j-petit: You're right. But do you think it makes sense to have an abstract parent class for CCD and discrete? They definitely have very similar interfaces when I like at their headers

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_world_bt.h
```diff
@@ -76,6 +79,15 @@ class CollisionWorldBt : public CollisionWorld
   void checkRobotCollisionHelper(const CollisionRequest& req, CollisionResult& res, const CollisionRobot& robot,
                                  const robot_state::RobotState& state, const AllowedCollisionMatrix* acm) const;
 
+  void checkRobotCollisionHelperCCD(const CollisionRequest& req, CollisionResult& res, const CollisionRobot& robot,
+                                    const robot_state::RobotState& state1, const robot_state::RobotState& state2,
+                                    const AllowedCollisionMatrix* acm) const;
+
+  void addToManager(const World::Object* obj) const;
```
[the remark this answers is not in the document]
j-petit: Agree

moveit_core/collision_detection_bullet/src/tesseract/bullet_cast_bvh_manager.cpp
```diff
@@ -0,0 +1,443 @@
...
+            assert(dynamic_cast<CastHullShape*>(compound->getChildShape(i)) != nullptr);
+            const btTransform& local_tf = compound->getChildTransform(i);
+
+            btTransform delta_tf = (tf1 * local_tf).inverseTimes(tf2 * local_tf);
+            static_cast<CastHullShape*>(compound->getChildShape(i))->updateCastTransform(delta_tf);
+            compound->updateChildTransform(i, local_tf, false);  // This is required to update the BVH tree
+          }
+          else if (btBroadphaseProxy::isCompound(compound->getChildShape(i)->getShapeType()))
+          {
+            assert(dynamic_cast<btCompoundShape*>(compound->getChildShape(i)) != nullptr);
```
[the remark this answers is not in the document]
j-petit: Why are those assert unnecessary? I have difficulties differentiating between necessary and unnecessary. I can't really see a difference to [this](https://github.com/ros-planning/moveit/pull/1504#discussion_r298851237).

moveit_core/collision_detection_bullet/src/collision_world_bt.cpp
```diff
@@ -84,22 +98,98 @@ void CollisionWorldBt::checkRobotCollision(const CollisionRequest& req, Collisio
...
 {
-  ROS_ERROR_NAMED("collision_detection.bullet", "Bullet robot world collision checking not yet implemented");
+  const CollisionRobotBt& robot_bt = dynamic_cast<const CollisionRobotBt&>(robot);
+
+  robot_bt.updateTransformsFromState(state);
+
+  tesseract::tesseract_bullet::Link2Cow link2cow;
+  link2cow = robot_bt.bt_manager_.getCollisionObjects();
+
+  for (const auto cow : link2cow)
```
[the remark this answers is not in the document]
j-petit: there are a lot of `auto`s in the FCL code though. But this is the MoveIt style convention right?

moveit_core/collision_detection_bullet/src/collision_robot_bt.cpp
```diff
@@ -35,16 +35,104 @@
...
+  bt_manager_.setIsContactAllowedFn(other_bt.bt_manager_.getIsContactAllowedFn());
+}
+
+void CollisionRobotBt::getAttachedBodyObjects(const robot_state::AttachedBody* ab, std::vector<void*>& geoms) const
+{
+  // TODO: Rewrite for using bullet add alll shapes to a single bullet COW. use geoms as output!
+  const std::vector<shapes::ShapeConstPtr>& shapes = ab->getShapes();
+  for (std::size_t i = 0; i < shapes.size(); ++i)
+  {
+  }
```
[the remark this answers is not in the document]
j-petit: Removed this function in latest commit

moveit_core/collision_detection_bullet/src/collision_robot_bt.cpp
```diff
@@ -63,21 +151,44 @@ void CollisionRobotBt::checkSelfCollision(const CollisionRequest& req, Collision
                                           const robot_state::RobotState& state1,
                                           const robot_state::RobotState& state2) const
 {
-  ROS_ERROR_NAMED("collision_detection.bullet", "Bullet continuous collision checking not yet implemented");
+  ROS_ERROR_NAMED("collision_detection.bullet", "Bullet self continuous collision checking not yet implemented");
```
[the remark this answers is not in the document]
j-petit: Added as ToDo in the Kanban

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
j-petit: I think then we don't need `nearest_points` and `cc_nearest_points` at all and keep using only `pos` of the contact? See the documentation [here](https://pybullet.org/Bullet/BulletFull/classbtManifoldPoint.html#ac4fb334242b89daa97eb635ca69cb01f)

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
```diff
@@ -0,0 +1,1077 @@
...
+  void getAabb(const btTransform& t_w0, btVector3& aabbMin, btVector3& aabbMax) const override
+  {
+    m_shape->getAabb(t_w0, aabbMin, aabbMax);
+    btVector3 min1, max1;
+    m_shape->getAabb(t_w0 * m_t01, min1, max1);
+    aabbMin.setMin(min1);
+    aabbMax.setMax(max1);
+  }
+
+  void getAabbSlow(const btTransform& /*t*/, btVector3& /*aabbMin*/, btVector3& /*aabbMax*/) const override
```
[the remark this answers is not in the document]
j-petit: The function is declared in the parent class as `= 0` and therefore has to be implemented right?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
```diff
@@ -0,0 +1,1077 @@
...
+  btVector3 normalWorldFromCast = -(castShapeIsFirst ? 1 : -1) * cp.m_normalWorldOnB;
+  const btCollisionObjectWrapper* firstColObjWrap = (castShapeIsFirst ? colObj0Wrap : colObj1Wrap);
+
+  if (castShapeIsFirst)
+  {
+    std::swap(col->nearest_points[0], col->nearest_points[1]);
+    contact.pos = convertBtToEigen(cp.m_positionWorldOnB);
+    std::swap(col->body_name_1, col->body_name_2);
+    std::swap(col->body_type_1, col->body_type_2);
+    col->normal *= -1;
```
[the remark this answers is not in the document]
j-petit: Yes, that is how I understood it.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
```diff
@@ -0,0 +1,1077 @@
...
+typedef std::shared_ptr<const CollisionObjectWrapper> COWConstPtr;
+typedef std::map<std::string, COWPtr> Link2Cow;
+typedef std::map<std::string, COWConstPtr> Link2ConstCow;
+
+/** @brief This is a casted collision shape used for checking if an object is collision free between two transforms */
+struct CastHullShape : public btConvexShape
+{
+public:
+  btConvexShape* m_shape;
+  btTransform m_t01;
```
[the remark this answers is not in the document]
j-petit: As far as I understood this is the local transformation to the shape coordinate system. So from COW into shape coordinates. So something like `m_shape_transform`?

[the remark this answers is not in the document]
BryceStevenWilley: That would be good.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/continuous_contact_manager_base.h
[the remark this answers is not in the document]
BryceStevenWilley: That's not a bad idea, but I don't know how to handle adding the second collision obj position in the parent class. Unless there's duplicated code, I don't think having a shared abstract helps much.

moveit_core/collision_detection_bullet/src/tesseract/bullet_cast_bvh_manager.cpp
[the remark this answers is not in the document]
BryceStevenWilley: The linked comment/asserts are at the begining of a function, which could be called with wrong arguments.  These asserts on the other hand come directly after an if statement that is checking that the above if statement is true in a different way, so it seems a little redundant.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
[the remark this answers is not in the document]
BryceStevenWilley: Ah, sorry, I thought I saw that getAABBSlow wasn't pure, my mistake

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/continuous_contact_manager_base.h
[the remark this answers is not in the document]
j-petit: Hm yes agree. Then I'll remove the abstract parent classes for now and we might discuss that at a later point again.

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
BryceStevenWilley: Not sure about that, I think TrajOpt might need both points, esp. in the case of a self collision.

j-petit: @ommmid What kind of information do you need of a contact point for TrajOpt?

ommmid: Sorry, not sure at this point but maybe it is good to have both points available in case they are needed.

j-petit: but we can calculate `m_positionWorldOnA = m_positionWorldOnB + m_normalWorldOnB * m_distance1;` so having two points is redundant?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/contact_checker_common.h
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * @brief Determine if contact is allowed between two objects.
+ * @param name1 The name of the first object
+ * @param name2 The name of the second object
+ * @param acm The contact allowed function
+ * @param verbose If true print debug informaton
+ * @return True if contact is allowed between the two object, otherwise false.
+ */
+inline bool isContactAllowed(const std::string& name1, const std::string& name2, const IsContactAllowedFn acm_fn,
```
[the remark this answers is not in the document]
j-petit: I agree but this comes from the fact that the goal was to have it as generic as possible and therefore make it possible to add other custom ACM check functions. through the function pointer `IsContactAllowed fn_` of the managers. I can remove this and hardcode it into the collision checking, would that be ok? I don't fully understand what you mean with a local lambda function.

moveit_core/collision_detection_bullet/test/test_bt_collision_detection.cpp
```diff
@@ -0,0 +1,558 @@
...
+  {
+    cworld_->getWorld()->removeObject("shape");
+    shapes.clear();
+    poses.clear();
+    shapes.push_back(shapes::ShapeConstPtr(new shapes::Box(1 + i * .0001, 1 + i * .0001, 1 + i * .0001)));
+    poses.push_back(Eigen::Isometry3d::Identity());
+    cworld_->getWorld()->addToObject("shape", shapes, poses);
+    collision_detection::CollisionRequest req;
+    collision_detection::CollisionResult res;
+    cworld_->checkCollision(req, res, *crobot_, robot_state1, *acm_);
```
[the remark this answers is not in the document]
j-petit: This is an exact copy of [this FCL test](https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp) which I use to see if the Bullet functionality matches FCL. I think it would make sense to have the same test scenario (and only a single file) for both of them somewhere (neither in the bullet folder, nor in the fcl folder).

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
BryceStevenWilley: That's true, but you might as well keep it if it's already calculated and you're just going to use it again.

moveit_core/collision_detection_bullet/test/test_bt_collision_detection.cpp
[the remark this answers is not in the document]
BryceStevenWilley: I agree in principle, Gtest has some pretty neat tricks for running exact same tests on multiple classes. But I don't know of a good place to put that without breaking the modularity of the packages. Maybe someone else has some good ideas.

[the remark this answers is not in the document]
j-petit: But both are part of `moveit_core` so this shouldn't be a problem, right?

BryceStevenWilley:
This is close to being ready, the last big thing that I'd suggest is to `git rebase -i` and cleanup the commits so they're easier to read through, i.e. making format commits fixups of previous commits, combining related commits, etc.

moveit_core/collision_detection_bullet/test/test_bt_collision_detection.cpp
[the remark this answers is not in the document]
BryceStevenWilley: Yeah, but each of the `moveit_core` subfolders kinda act like their own packages. 

Not sure how GTest [typed tests](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#typed-tests) handle being included from other packages, but we could define the base checks in `collision_detection`, and then just include and instantiate the test from each FCL and Bullet folder with 

```
using MyTypes = ::testing::Types<CollisionAllocatorFCL>; // or Bt  
TYPED_TEST_SUITE(FooTest, MyTypes);
```

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/contact_checker_common.h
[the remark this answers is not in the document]
j-petit: Decided in sync call that we keep it as it is for now.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_common.h
j-petit: Removed the collision common files

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_cast_bvh_manager.h
```diff
@@ -0,0 +1,218 @@
...
+  void setCollisionObjectsTransform(const std::vector<std::string>& names,
+                                    const AlignedVector<Eigen::Isometry3d>& poses);
+
+  /**@brief Set a series of static collision objects' tranforms
+   * @param transforms A transform map <name, pose> */
+  void setCollisionObjectsTransform(const AlignedMap<std::string, Eigen::Isometry3d>& transforms);
+
+  /**@brief Set a single cast (moving) collision object's tansforms
+   *
+   * This should only be used for moving objects. Use the base class methods for static objects.
```
BryceStevenWilley: There is no more base class, just say single pose methods instead.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_robot_bt.h
```diff
@@ -37,7 +37,10 @@
 #ifndef MOVEIT_COLLISION_DETECTION_BT_COLLISION_ROBOT_
 #define MOVEIT_COLLISION_DETECTION_BT_COLLISION_ROBOT_
 
-#include <moveit/collision_detection_bullet/collision_common.h>
+#include <moveit/collision_detection_bullet/tesseract/bullet_discrete_bvh_manager.h>
+#include <moveit/collision_detection_bullet/tesseract/bullet_cast_bvh_manager.h>
+#include <moveit/collision_detection/collision_robot.h>
+#include <urdf/model.h>
```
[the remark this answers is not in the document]
j-petit: No, you are right.

moveit_ros/planning/planning_components_tools/src/compare_collision_speed_checking_fcl_bullet.cpp
```diff
@@ -0,0 +1,351 @@
...
+  std::uniform_real_distribution<> distr_position_z(0.0, 1);
+  std::uniform_real_distribution<> distr_size;
+
+  if (type == CollisionObjectType::MESH)
+  {
+    distr_size = std::uniform_real_distribution<>(0.3, 1);
+  }
+  else
+  {
+    distr_size = std::uniform_real_distribution<>(0.05, 0.2);
```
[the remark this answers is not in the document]
j-petit: Thanks for the hint, did not know that

moveit_core/collision_detection_bullet/test/test_bt_collision_detection.cpp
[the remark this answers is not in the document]
j-petit: I opened a separate PR for this #1543 and kept the test in this PR as it is. As soon as it is merged, I rebase.

BryceStevenWilley:
@davetcoleman @v4hn @rhaschke, this PR is ready for review.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
```diff
@@ -0,0 +1,109 @@
+/**
+ * @file basic_types.h
+ * @brief The tesseract_core package types.
+ *
+ * @author Levi Armstrong
+ * @date April 15, 2018
+ * @version TODO
+ * @bug No known bugs
```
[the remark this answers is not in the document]
davetcoleman: lol yea, there's always bugs so i think we should remove everything at the top here, except the author which we could move below the license per moveit standard

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
```diff
@@ -0,0 +1,109 @@
...
+    , fn(fn)
+    , acm(acm)
+    , res(res)
+    , req(req)
+    , done(false)
+    , pair_done(false)
+  {
+  }
+
+  const std::vector<std::string>& active;
```
davetcoleman: adapt variables for mvoeit style 
```suggestion
  const std::vector<std::string>& active_;
```

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_discrete_bvh_manager.h
```diff
@@ -0,0 +1,198 @@
...
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef TESSERACT_COLLISION_BULLET_DISCRETE_BVH_MANAGERS_H
+#define TESSERACT_COLLISION_BULLET_DISCRETE_BVH_MANAGERS_H
+
+#include <moveit/collision_detection_bullet/tesseract/bullet_utils.h>
+#include <moveit/macros/class_forward.h>
+
+namespace tesseract
```
davetcoleman: could we change namespace to collision_detection_bullet ?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
```diff
@@ -0,0 +1,1155 @@
...
+  }
+  else
+  {
+    outpt = shape->localGetSupportingVertexWithoutMargin(localNormal);
+    outsupport = localNormal.dot(outpt);
+  }
+}
+
+/** @brief Check if a collision check is required between the provided collision objects
+ *  @param cow1 The first collision object
```
davetcoleman: why is it called cow?

moveit_core/collision_detection_bullet/src/collision_world_bt.cpp
```diff
@@ -84,22 +105,71 @@ void CollisionWorldBt::checkRobotCollision(const CollisionRequest& req, Collisio
...
+}
+
+void CollisionWorldBt::checkRobotCollisionHelperCCD(const CollisionRequest& req, CollisionResult& res,
+                                                    const CollisionRobot& robot, const robot_state::RobotState& state1,
+                                                    const robot_state::RobotState& state2,
+                                                    const AllowedCollisionMatrix* acm) const
+{
+  const CollisionRobotBt& robot_bt = dynamic_cast<const CollisionRobotBt&>(robot);
+
+  tesseract::BulletCastBVHManagerPtr temp_clone_manager = bt_manager_CCD_->clone();
```
davetcoleman:
```suggestion
  tesseract::BulletCastBVHManagerPtr bvh_clone_manager = bt_manager_CCD_->clone();
```
what does temp mean here?

moveit_core/collision_detection_bullet/src/collision_world_bt.cpp
```diff
@@ -84,22 +105,71 @@ void CollisionWorldBt::checkRobotCollision(const CollisionRequest& req, Collisio
...
 }
 
 void CollisionWorldBt::checkRobotCollisionHelper(const CollisionRequest& req, CollisionResult& res,
                                                  const CollisionRobot& robot, const robot_state::RobotState& state,
                                                  const AllowedCollisionMatrix* acm) const
 {
-  ROS_ERROR_NAMED("collision_detection.bullet", "Bullet robot world collision checking not yet implemented");
+  const CollisionRobotBt& robot_bt = dynamic_cast<const CollisionRobotBt&>(robot);
+
+  tesseract::BulletDiscreteBVHManagerPtr temp_clone_manager = bt_manager_->clone();
```
davetcoleman: remove temp

moveit_core/collision_detection_bullet/test/test_bt_collision_detection_panda.cpp
```diff
@@ -0,0 +1,354 @@
...
+  pos1.translation().z() = 0.3;
+  cworld_->getWorld()->addToObject("box", shape_ptr, pos1);
+  cworld_->checkRobotCollision(req, res, *crobot_, *robot_state_, *acm_);
+  ASSERT_TRUE(res.collision);
+  ASSERT_GE(res.contact_count, 3u);
+  res.clear();
+}
+
+/** \brief Continuous self collision checks are not supported yet by tesseract */
+TEST_F(BulletCollisionDetectionTester, DISABLED_ContinuousCollisionSelf)
```
davetcoleman: Why is this disbled?

moveit_core/collision_detection_bullet/test/test_bt_collision_detection_panda.cpp
```diff
@@ -0,0 +1,354 @@
...
+  res.clear();
+
+  crobot_->setLinkPadding("panda_hand", 0.0);
+  cworld_->checkRobotCollision(req, res, *crobot_, *robot_state_, *acm_);
+  ASSERT_FALSE(res.collision);
+}
+
+int main(int argc, char** argv)
+{
+  testing::InitGoogleTest(&argc, argv);
```
davetcoleman: great tests!

moveit_ros/planning/planning_components_tools/src/compare_collision_speed_checking_fcl_bullet.cpp
```diff
@@ -0,0 +1,354 @@
...
+      planning_scene->getRobotModel()->getLinkModelNames(), true) };
+
+  // set the robot state to home position
+  robot_state::RobotState& current_state{ planning_scene->getCurrentStateNonConst() };
+  collision_detection::CollisionRequest req;
+  current_state.setToDefaultValues(current_state.getJointModelGroup("panda_arm"), "home");
+  current_state.update();
+
+  // load panda link5 as world collision object
+  std::string name{ "" };
```
davetcoleman: no need to initialize a string

https://stackoverflow.com/questions/17738439/value-and-size-of-an-uninitialized-stdstring-variable-in-c

davetcoleman:
Great work!!

j-petit:
> This PR is huge and difficult to review because Chrome / Github are super slugging on my computer at this size. Can this PR be split at all?
> 
I don't see an easy and nice way to split it without breaking the functionality. It just needs a lot of glue code to bridge the gap between Bullet and MoveIt. [The benchmark script](https://github.com/ros-planning/moveit/pull/1504/files#diff-3e2a424896ae0403cadb57341df860a6) can be ignored at first / removed from this PR?

About Github / Chrome issue: I use Firefox and don't have this problem, maybe give it a try?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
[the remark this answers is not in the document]
j-petit: Should I format it in a MoveIt way so it looks exactly like the MoveIt header but with the Apache license?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
j-petit: Collision Object Wrapper (COW) as it wraps around a bullet collision object `btCollisionObject`

moveit_core/collision_detection_bullet/test/test_bt_collision_detection_panda.cpp
j-petit: because continuous to continuous checks are not implemented yet, this feature was / is missing in tesseract and I haven't added it yet. However, I already wrote a test for it.

moveit_core/collision_detection_bullet/src/collision_world_bt.cpp
j-petit: The clone() mostly takes place because all the collision checking function are `const`. Therefore, the cloned one is only used in single collision check, that is why I called it temp. But I agree, all local variables are "temp", so I'll rename it.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: For all member variables? There are also other parts of the code where a prefixed `m_` is used (the code which directly inherits from bullet as in bullet all members are prefixed with `m_`). Should I also change this?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_discrete_bvh_manager.h
j-petit: sure

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common.h
```diff
@@ -469,7 +469,7 @@ TYPED_TEST_P(CollisionDetectorTest, TestCollisionMapAdditionSpeed)
   ros::WallTime start = ros::WallTime::now();
   this->cworld_->getWorld()->addToObject("map", shapes, poses);
   double t = (ros::WallTime::now() - start).toSec();
-  EXPECT_GE(1.0, t);
+  EXPECT_GE(5.0, t);
```
rhaschke: Why do you need to change this? Adding objects to the world takes more than a second now?

moveit_core/collision_detection_bullet/CMakeLists.txt
```diff
@@ -1,29 +1,39 @@
...
   src/collision_robot_bt.cpp
   src/collision_world_bt.cpp
+  src/tesseract/bullet_utils.cpp
+  src/tesseract/bullet_discrete_bvh_manager.cpp
 )
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 
-target_link_libraries(${MOVEIT_LIB_NAME} moveit_collision_detection ${catkin_LIBRARIES} ${urdfdom_LIBRARIES} ${urdfdom_headers_LIBRARIES} ${LIBFCL_LIBRARIES} ${Boost_LIBRARIES})
+target_link_libraries(${MOVEIT_LIB_NAME} moveit_collision_detection
+    moveit_collision_detection_fcl ${catkin_LIBRARIES} ${urdfdom_LIBRARIES}
```
rhaschke: FCL shouldn't be needed here.
```suggestion
    ${catkin_LIBRARIES} ${urdfdom_LIBRARIES}
```

moveit_core/collision_detection_bullet/CMakeLists.txt
```diff
@@ -1,29 +1,39 @@
...
+    ${urdfdom_headers_LIBRARIES} ${LIBFCL_LIBRARIES} ${Boost_LIBRARIES}
```
rhaschke:
```suggestion
    ${urdfdom_headers_LIBRARIES} ${Boost_LIBRARIES}
```

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
```diff
@@ -0,0 +1,109 @@
...
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * @par
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef TESSERACT_CORE_BASIC_TYPES_H
```
rhaschke: This should be renamed to match the file name.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
rhaschke: `clang-tidy --fix` does this automatically. And yes, you should adapt all MoveIt code, otherwise Travis will complain anyway.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
```diff
@@ -0,0 +1,109 @@
...
+    , req(req)
+    , done(false)
+    , pair_done(false)
+  {
+  }
+
+  const std::vector<std::string>& active;
+
+  /** \brief If after a positive broadphase check the distance is below this threshold, a contact is added. */
+  const double& contact_distance;
```
rhaschke: Shouldn't this be called `broadphase_threshold` then (according to the comment)?
A contact should only be added if there is a real contact...

moveit_core/collision_detection_bullet/test/test_bt_collision_detection_panda.cpp
rhaschke: Fully agree. Can we make these tests generic as well, i.e. to be used for both FCL and Bullet?

moveit_core/collision_detection_bullet/CMakeLists.txt
j-petit: right

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common.h
j-petit: Yes, I haven't looked closer into this but with bullet I need around 3 seconds.

rhaschke: This needs to be investigated! Even a second is _really_ long...

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
felixvd: MoveIt style is verbose variable names, so I would call it coll_obj_wrapper or even collision_object_wrapper.

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common.h
j-petit: But these are **10000** shapes, not a single object? So a single object is less than 1 ms...

rhaschke: Ok, this is calming me down. Nevertheless, an increase from <1s to >3s is strange and should be understood. Please investigate at some point in future and leave a TODO here for now.

simonschmeisser: is it 10,000 times the same shape/mesh? Then maybe the bullet part creates copies of the mesh while the fcl one has a cache (it just stores the shape_ptr I think and should be even faster ...). Looking at the memory usage should answer this question easily (if not increase number/mesh size)

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: but for struct members there are no fixed rules?

moveit_core/CMakeLists.txt
```diff
@@ -55,6 +54,7 @@ COMPONENTS
   urdf
   visualization_msgs
   xmlrpcpp
+  eigen_conversions
```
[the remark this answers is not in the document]
j-petit: removed it

moveit_core/CMakeLists.txt
```diff
@@ -13,15 +13,14 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-
 if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   message("${PROJECT_NAME}: You did not request a specific build type: Choosing 'Release' for maximum performance")
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
 find_package(Boost REQUIRED system filesystem date_time thread iostreams)
 find_package(Eigen3 REQUIRED)
+find_package(Bullet REQUIRED)
```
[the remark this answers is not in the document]
j-petit: added ToDo

moveit_core/collision_detection_bullet/CMakeLists.txt
j-petit: removed

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: done in all headers

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
```diff
@@ -95,6 +103,14 @@ struct Contact
 
   /** \brief The type of the second body involved in the contact */
   BodyType body_type_2;
+
+  /** \brief The distance percentage between casted poses until collision */
+  double percent_interpolation;
+
+  /** \brief The continous collision type */
+  ContinuousCollisionType cc_type;
```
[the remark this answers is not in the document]
j-petit: yes, you're right. I'll add a comment explaining that to the `percent_interpolation` and remove the enum.

moveit_core/collision_detection_bullet/test/test_bt_collision_detection_panda.cpp
j-petit: Yes, done.

moveit_ros/planning/planning_components_tools/src/compare_collision_speed_checking_fcl_bullet.cpp
j-petit: changed in PR #1551

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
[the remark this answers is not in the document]
j-petit: > Should I format it in a MoveIt way so it looks exactly like the MoveIt header but with the Apache license?

I changed it as suggested

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
j-petit: if i could vote for a single abbreviation in the bullet code, it would be `cow` - it is used all over the code and writing `collision_object_wrapper` for each instance would lead to many ugly line breaks which in my opinion decrease readability.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: No, this is not the broadphase threshold, this is another parameter specific to each collision object. For a broadphase candidate pair, their distance is calculated and if its below the `contact_distance` the distance and all other information is added as a result. In Bullet this is regarded as a contact. Note that contact does not necessarily refer to a real contact (distance < 0 is possible).

Contrary to FCL, Bullet does not make a difference between distance and collision requests which in my opinion is the cleaner solution with less duplicate code (the FCL MoveIt implementation has plenty of duplicate code because of this distinction). In Bullet, distance requests are simply treated as contacts with the only difference that the contact depth/distance is positive. So a contact is not necessarily a real collision. This wording is used throughout the Bullet Library.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
rhaschke: I fully understand you. IMHO, going for cow is fine. Elsewhere we used co for collision_object as well.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
rhaschke: If I get you correct, Bullet performs the detailed collision/distance check only if the broadphase check passed. So, if two objects fail the broadphase check (e.g. because they are nicely aligned and their axes-aligned bounding boxes don't overlap), their distance will not be reported?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
```diff
@@ -0,0 +1,106 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *********************************************************************/
+
+/* Author: Levi Armstrong */
+
+#ifndef MOVEIT_COLLISION_DETECTION_BULLET_TESSERACT_BASIC_TYPES_H_
```
rhaschke: There are still some files left with tesseract in their file path and header guard.
I'm fine with this, if you want to keep those as a reminder to migrate later.
Otherwise, these shoult be renamed as well, shouldn't they?

moveit_core/package.xml
```diff
@@ -47,6 +48,7 @@
   <depend>trajectory_msgs</depend>
   <depend>visualization_msgs</depend>
   <depend>xmlrpcpp</depend>
+  <depend>eigen_conversions</depend>
```
rhaschke: `eigen_conversions` still mentioned here.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: No, they won't. But you can easily expand the AABB of each object by a distance (this is a parameter) and then they will be reported. So I could expand each AABB by 5 cm and then I will get for sure all objects which are closer than 5 cm.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
felixvd: I don't have a strong position, but in my last PR there were wishes to spell it out, which I did and preferred. I'd probably vote for `coll_obj_wrapper`, but maybe you can link a few spots where `cow` would be significantly better?

j-petit: We have ~ 200 occurrences of cow in the code. Replacing leads in many lines to line breaks like
image

What I could do is replace the type specifier (not variable names) with the longer `CollisionObjectWrapper`. This is not in too many places and then one can simply peek at the function header to see the type of `cow` if this abbreviation is used as a variable name.

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common.h
j-petit: Yes it is the same shape/mesh. FCL definitely is much faster, but the memory usage of both programs seems similar (see below). The reason must be something else, I added the ToDo and will investigate at a later point.

FCL:
image

Bullet:
image

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: Yes, we discussed the name `bullet_integration` as a subfolder for the old tesseract code. Any objections to the name?

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
BryceStevenWilley: I think the longer type specifier is a better choice.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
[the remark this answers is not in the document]
davetcoleman: Thanks!

moveit_core/collision_detection_bullet/test/test_bt_collision_detection_panda.cpp
davetcoleman: Please add TODO comment with this explanation :-)

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
davetcoleman: If its derived from a bullet class then no, we can't really change it. You see this predicament with Qt-code usage, too. We may have to add ``ignore`` rules around that code, so clang-tidy doesn't complain.

Using ``clang-tidy --fix`` as @rhaschke is the best way to go, its nice and automated :+1:

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
davetcoleman: I wouldn't be too concerned about line breaks. I've gotten used to Google-style code which is limited to 80 char width, *and* they use verbose variable names. I think readability of full names is better than worrying about line count.

Otherwise I keep thinking about milk and farms :cow:

davetcoleman:
Excited to seeing this close to ready! Travis is still unhappy...

moveit_core/collision_detection_bullet/test/test_bt_collision_detection_panda.cpp
j-petit: Added it, not in this PR but in #1551 as CCD got removed from this PR (this test included)

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/bullet_utils.h
j-petit: I changed all type specifiers to `CollisionObjectWrapper`

moveit_core/package.xml
j-petit: Missed this one. Removed.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: I already ran clang-tidy (super nice tool!) but this is in a struct, and for struct there are no rules specified in the clang-tidy config. Is it common practice to use `_` postfix for struct members? Then this should be added to the clang-tidy config I think. However I see them without the `_` postfix all over the collision checking code, see [here](https://github.com/ros-planning/moveit/blob/b853c6bc48c41af7475db8cbb4d11078bb8725eb/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h#L137), [here](https://github.com/ros-planning/moveit/blob/b853c6bc48c41af7475db8cbb4d11078bb8725eb/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h#L74), [here](https://github.com/ros-planning/moveit/blob/b853c6bc48c41af7475db8cbb4d11078bb8725eb/moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_common.h#L62).

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/tesseract/basic_types.h
j-petit: Changed it to `bullet_integration` including in all header guards

j-petit:
> Excited to seeing this close to ready! Travis is still unhappy...

Only kinetic is now failing (I was told it does not need to pass), the three other builds pass :tada:

rhaschke:
> Only kinetic is now failing.

Looks like Kinetic/Xenial ships with an older Bullet version. I suggest dropping Kinetic support for now.

davetcoleman:
Moving this convo here: https://github.com/ros-planning/moveit/issues/1570
;-)

j-petit:
**Ping**
Can we get this merged / further discussed / reviewed ?

j-petit:
Thanks for reviewing, next up is #1551 for the continuous collision checking.

PR #1518, opened 2019-06-23
Replace eigenpy with boost::python::numpy

rhaschke:
#1501 used `eigenpy` to directly return the Jacobian matrix from C++ to python as an Eigen::MatrixXd.
Unfortunately, using `eigenpy`, I experienced strange behaviour of `numpy`: `flatten()` didn't work on arrays anymore (https://github.com/stack-of-tasks/eigenpy/issues/63). Removing `eigenpy` solved the issue immediately.
As a work-around (and to get rid of this extra dependency), I replaced `eigenpy` with plain `numpy`.
@7675t, please review.

### DIFF
--- a/moveit_ros/planning_interface/CMakeLists.txt
+++ b/moveit_ros/planning_interface/CMakeLists.txt
@@ -18,7 +18,6 @@ find_package(catkin REQUIRED COMPONENTS
   tf2_eigen
   tf2_geometry_msgs
   tf2_ros
-  eigenpy
   roscpp
   actionlib
   rospy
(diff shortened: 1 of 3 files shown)


jcarpent:
boost::numpy has been integrated in Boost since version 1.63.0

rhaschke:
> boost::numpy has been integrated in Boost since version 1.63.0

Arggh! This constrains my solution to ROS Melodic / Ubuntu Bionic.

7675t:
@jcarpent Right, and it seems troublesome if we mix the old and new numpy version as discussed [here](http://answers.ros.org/question/274016/ros-dependency-boost-158-additional-dependency-boost-164/).

@rhaschke How about keeping this solution for only Melodic? And for kinetic, we can just return a list of list (not eigenpy or numpy array) and cast it in python, as in my original PR.

rhaschke:
Let's see whether the fix https://github.com/stack-of-tasks/eigenpy/pull/64 resolves the issue.
I don't have time until tomorrow to validate this...

rhaschke:
I have found a work-around for my eigenpy issue. No need for this PR anymore.

Issue #1526, opened 2019-06-28
TrajectoryMonitor leaks memory


jschleicher:
### Description

Monitoring the memory usage of move_group via `htop`, multiple "Plan and execute" requests consume additional memory.

### Your environment
* ROS Distro: melodic
* Ubuntu 18.04
* Source build 898cdc8d8df5689319cd462c2c674954c6134604

### Steps to reproduce
1. `roslaunch panda_moveit_config demo.launch`
2. In RViz choose a random goal and press "Plan and Execute" (continue this step and watch memory consumption of the move_group process).

### Expected behaviour
Memory consumption stays constant after some initial runs.

### Actual behaviour
Memory footprint increases. Digging deeper:
https://github.com/ros-planning/moveit/blob/cfb46fe3ba6b0d708f22c8692c337a47d6cde6da/moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp#L90
is never called. Adding a debug output to line 75 shows, that the trajectory gets more and more point from request to request.
```
ROS_INFO_STREAM("trajectory size is " << trajectory_.getWayPointCount());
```

Note:
* Memory consumption stays constant, if pressing "Plan" and "Execute" buttons as individual steps
* Only some kilobytes are allocated after some minutes running the robot (tiny increase), but over the long run this is still a problem for long-running applications, since it is never freed.

Question:
* What is the use case of `TrajectoryMonitor`? I cannot find, where the `trajectory_` is actually read / returned to the user.

rhaschke:
Thanks for point out this issue. Looks like TrajectoryMonitor is only used in [`PlanExecution`](https://github.com/ros-planning/moveit/blob/91f55a58007df4e1ed2e8cff8393337f4f9a3596/moveit_ros/planning/plan_execution/include/moveit/plan_execution/plan_execution.h#L155), which calls [`startTrajectoryMonitor`](https://github.com/ros-planning/moveit/blob/91f55a58007df4e1ed2e8cff8393337f4f9a3596/moveit_ros/planning/plan_execution/src/plan_execution.cpp#L399) and [`stopTrajectoryMonitor`](https://github.com/ros-planning/moveit/blob/91f55a58007df4e1ed2e8cff8393337f4f9a3596/moveit_ros/planning/plan_execution/src/plan_execution.cpp#L444). However, aside from these calls, the recorded trajectory is never used. So, I don't have an answer to your question as well. @davetcoleman, @v4hn Do you have an idea?
[...]

PR #1528, opened 2019-06-28
remove (unused) TrajectoryMonitor from PlanExecution

rhaschke:
As pointed out in #1526, TrajectoryMonitor in PlanExecution leaks memory and is not used at all.
That's why I suggest to remove it.


rhaschke:
> This also seems to be the only place TrajectoryMonitor is used. But I'd rather leave it in the codebase.

I completely agree, the class itself should be kept. I didn't explicitly mentioned that...

jschleicher:
+1 for cherry-picking this into melodic as in https://github.com/ros-planning/moveit/issues/1526#issuecomment-506742669

rhaschke:
Looks like the `getTrajectory()` call was lost [here](https://github.com/ros-planning/moveit/commit/5f172e2a0e9875f62e74eea784085deae233b82b#diff-5fb405ea950d332e97a35b53bda6b6eaL499). I will prepare another PR.

PR #1530, opened 2019-07-01
Allow ROS namespaces for planning request adapters: pass a NodeHandle for their initialization

henningkayser:
This PR adds support for running multiple planning pipelines with different configurations at the same time. The planning including planning request adapters from separate namespaces. While the configuration of the planner plugin can already be scoped by passing a child namespace to the planning pipeline, the parameters of the planning request adapters are always queried from the private node handle from within the constructor. The parameter lookup is now moved to a new virtual `initialize()` function (empty by default) which expects the node handle to use. The planning pipeline calls this function on all constructed planning request adapters.

### DIFF
--- a/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -64,8 +64,12 @@ public:
   }
 
   /// Initialize parameters using the passed NodeHandle
+  // TODO - Make initialize() a pure virtual function
   virtual void initialize(const ros::NodeHandle& node_handle)
   {
+    ROS_WARN_NAMED("planning_request_adapter",
+                   "Function initialize() is not implemented. All parameters should be"
+                   "loaded from the passed NodeHandle. Using a private NodeHandle for this is deprecated.");
   }
 
   /// Get a short string that identifies the planning request adapter
--- a/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -64,7 +64,7 @@ public:
   }
 
   /// Initialize parameters using the passed NodeHandle
-  // TODO - Make initialize() a pure virtual function
+  // TODO - Make initialize() a pure virtual function in O-turtle
   virtual void initialize(const ros::NodeHandle& node_handle)
   {
     ROS_WARN_NAMED("planning_request_adapter",
--- a/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -41,6 +41,7 @@
 #include <moveit/planning_interface/planning_interface.h>
 #include <moveit/planning_scene/planning_scene.h>
 #include <boost/function.hpp>
+#include <cxxabi.h>
 
 /** \brief Generic interface to adapting motion planning requests */
 namespace planning_request_adapter
@@ -67,9 +68,16 @@ public:
   // TODO - Make initialize() a pure virtual function in O-turtle
   virtual void initialize(const ros::NodeHandle& node_handle)
   {
-    ROS_WARN_NAMED("planning_request_adapter",
-                   "Function initialize() is not implemented. All parameters should be"
-                   "loaded from the passed NodeHandle. Using a private NodeHandle for this is deprecated.");
+    // Get name of derived adapter
+    std::string adapter_name = typeid(*this).name();
+    // Try to demangle the name
+    int status;
+    std::string demangled_name = abi::__cxa_demangle(adapter_name.c_str(), NULL, NULL, &status);
+    if (status == 0)
+      adapter_name = demangled_name;
+    ROS_WARN_NAMED("planning_request_adapter", "Implementation of function initialize() is missing from '%s'."
+                                               "Any parameters should be loaded from the passed NodeHandle.",
+                   adapter_name.c_str());
   }
 
   /// Get a short string that identifies the planning request adapter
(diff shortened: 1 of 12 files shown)


moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -142,18 +142,21 @@ void planning_pipeline::PlanningPipeline::configure()
     if (adapter_plugin_loader_)
       for (const std::string& adapter_plugin_name : adapter_plugin_names_)
       {
-        planning_request_adapter::PlanningRequestAdapterConstPtr ad;
+        planning_request_adapter::PlanningRequestAdapter* ad;
         try
         {
-          ad = adapter_plugin_loader_->createUniqueInstance(adapter_plugin_name);
+          ad = adapter_plugin_loader_->createUnmanagedInstance(adapter_plugin_name);
```
rhaschke: Creating unmanaged plugins is very bad practice and needs to be reverted.
As `initialize()` requires a modifyable pointer, just use a non-const `PlanningRequestAdapterPtr` or even better a `UniquePtr`.

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
```diff
@@ -63,6 +63,11 @@ class PlanningRequestAdapter
   {
   }
 
+  /// Initialize parameters using the passed NodeHandle
+  virtual void initialize(const ros::NodeHandle& node_handle)
+  {
```
rhaschke: This default implementation should issue a warning indicating that the old way of initializing from a private `NodeHandle` is deprecated. In future this method should become abstract. Please leave a corresponding TODO.

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
henningkayser: I think it's fine in this case since either the instance fails to be constructed and the exception is caught or the valid pointer is immediately managed after constructing a PlanningRequestAdapterConstPtr with it. I can change it to `createUniqueInstance()` thought.

rhaschke: No, it's not fine at all! When creating the pointer as unmanaged, the _class loader_ cannot keep track of the instance and thus cannot _unload_ the library. This will issue a warning message, which I just recently fixed (#1104).

henningkayser: Got it, please review 84e2223

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
```diff
@@ -63,6 +63,15 @@ class PlanningRequestAdapter
   {
   }
 
+  /// Initialize parameters using the passed NodeHandle
+  // TODO - Make initialize() a pure virtual function
```
rhaschke: A deadline is missing:
```suggestion
  // TODO - Make initialize() a pure virtual function in O-turtle
```

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
```diff
@@ -63,6 +63,15 @@ class PlanningRequestAdapter
...
+  virtual void initialize(const ros::NodeHandle& node_handle)
+  {
+    ROS_WARN_NAMED("planning_request_adapter",
```
rhaschke: What about using `typeid(*this).name()` to provide a useful hint which adapter is effected.

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
```diff
@@ -63,6 +63,15 @@ class PlanningRequestAdapter
...
+                   "Function initialize() is not implemented. All parameters should be"
```
rhaschke:
```suggestion
                   "Implementation of function initialize() is missing, thus preventing to load parameters from the passed NodeHandle / namespace."
```

henningkayser:
> From the title, I was expecting a much larger contribution, namely supporting multiple planning pipelines (and planners) in the `move_group` node. To clarify the contribution of this PR, I changed the title.

I'm actually looking into this, but I think it's a first good step to be able to actually run different planning pipelines without using the MoveGroup node.

> There are more adapters in folder `moveit_ros/planning/planning_request_adapter_plugins/src` that need to be adapted to the new API.

You mean by adding an empty initialize() implementation?

moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
henningkayser: Isn't this a bit confusing since not all adapters load parameters?

rhaschke:
> > There are more adapters in folder moveit_ros/planning/planning_request_adapter_plugins/src that need to be adapted to the new API.

> You mean by adding an empty initialize() implementation?

If these adapters don't use parameters: yes. The best way to validate completeness of your API change within the code base is to make `initialize()` abstract.

gavanderhoorn:
> > From the title, I was expecting a much larger contribution, namely supporting multiple planning pipelines (and planners) in the `move_group` node. To clarify the contribution of this PR, I changed the title.
> 
> I'm actually looking into this, but I think it's a first good step to be able to actually run different planning pipelines without using the MoveGroup node.

@henningkayser: can you say anything about the approach you're taking / looking into?

henningkayser:
@gavanderhoorn I didn't implement anything yet but I think the most straightforward approach would be to scope optional (or even all) planning pipeline configurations in sub-namespaces and pass the names as parameters to the MoveGroup node. Inside the MoveGroupContext we load all specified pipelines and instead of the member `planning_pipeline_` we use a getter function to access the pipeline by name. The name itself should be an optional parameter inside the MotionPlanRequest to support both MoveGroupPlanService and MoveGroupAction capabilities.

PR #1532, opened 2019-07-02
gui-docker: Update to nvidia-docker version 2

rhaschke:
This is a replacement for #1495, updating the gui-docker script to work with [`nvidia-docker 2`](https://github.com/NVIDIA/nvidia-docker/wiki) as well. Both versions, 1 and 2, of nvidia-docker are supported now.
I also added support for DRI-based cards.

### DIFF
--- a/.docker/gui-docker
+++ b/.docker/gui-docker
@@ -59,6 +59,24 @@ function transfer_x11_permissions() {
     xauth nlist $DISPLAY | sed -e 's/^..../ffff/' | xauth -f $XAUTH nmerge -
 }
 
+function count_positional_args() {
+    while true ; do
+       case "${1:-}" in
+          # Skip common options with a subsequent positional argument
+          # This list is not exhaustive! Augment as you see fit.
+          -v|--volume) shift ;;
+          -w) shift ;;
+          -e) shift ;;
+          # Skip all other options
+          -*) ;;
+          *) break ;;
+       esac
+       shift
+    done
+    # Return remaining number of arguments
+    echo $#
+}
+
 # Check for option -c or --container in first position
 case "${1:-}" in
    -c|--container)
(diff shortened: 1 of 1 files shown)


rhaschke:
As pointed out by @felixvd in https://github.com/ros-planning/moveit/pull/1495#issuecomment-508632232, it would be comfortable for unexperienced docker users to just run gui-docker without the need to remember options to pass. Also, support for persistent containers would be great, i.e. the ability to re-enter a previously ran container.
I added support for both features now: `gui-docker -c <container>` (or just `gui-docker -c`) allows to continue an existing container with the given name (or default_container).
If the container doesn't yet exist and no (positional) options are provided to specify the docker image and command to execute, a meaningful default is used.

Note: Continuing an existing container doesn't stop the container in the end. I like this feature, because it allows to run an additional process within the same container. But there might be other opinions on that.

BryceStevenWilley:
I'm interested in trying this out, I'll look at it tonight.

.docker/gui-docker
```diff
@@ -85,6 +103,10 @@ if [ -n "${CONTAINER_NAME:-}" ] ; then
    if [ -z "$($DOCKER_EXECUTABLE ps -aq --filter name=^$CONTAINER_NAME\$)" ] ; then
       # container not yet existing, add an option to name the container when running docker below
       NAME_OPTION="--name=$CONTAINER_NAME"
+      if [ $(count_positional_args) -eq 0 ] ; then
+         # If no further (positional) arguments were provided, start a bash in the default image (for dummy users)
```
felixvd: > dummy users

Harsh, harsh.

felixvd:
> I added support for both features now: `gui-docker -c <container>` (or just `gui-docker -c`) allows to continue an existing container with the given name (or default_container).

Great. I would go even further and make that the default behavior when `gui-docker` is called without arguments.

> Note: Continuing an existing container doesn't stop the container in the end. I like this feature, because it allows to run an additional process within the same container. But there might be other opinions on that.

I think this is fine and will test soon.

rhaschke:
> I would go even further and make that the default behavior when gui-docker is called without arguments.

This cannot become the default, as it requires to specify a container name.

felixvd:
I assumed it would start up "default_container" as you suggested, unless I misunderstood what you meant.

rhaschke:
Indeed, if there are no arguments at all, we could go for `default_container`.

rhaschke:
Done

.docker/gui-docker
```diff
@@ -1,56 +1,139 @@
...
+function check_dri() {
+    # If there is no /dev/dri, bail out
+    test -d /dev/dri || return 1
+    DOCKER_PARAMS="--device=/dev/dri --group-add video"
+}
+
+function transfer_x11_permissions() {
+    # store X11 access rights in temp file to be passed into docker container
+    XAUTH=/tmp/.docker.xauth
+    touch $XAUTH
```
BryceStevenWilley:
```suggestion
    touch "$XAUTH"
```
I'll default to you, but my understanding of bash scripts is that all variables should be in quotes.

rhaschke: Indeed, it's good practice to quote variables. In this case, it's not necessary, because the variable doesn't contain spaces.

PR #1533, opened 2019-07-03
move occupancy_map_monitor into its own package

jschleicher:
### Description
Suggestion for #1463:

To break the dependency `moveit_ros_planning` -> `moveit_ros_perception`, this PR factors out  `libmoveit_occupancy_map_monitor` (which is required by `moveit_ros_planning`) into a new package `moveit_ros_occupancy_map`. 
This allows to relax the dependency of `moveit_ros_planning` on `moveit_ros_perception`, replacing it with a dependency on `moveit_ros_occupancy_map` only. In this fashion, `moveit_ros_planning` can be installed w/o `moveit_ros_perception`, which pulls in a whole bunch of additional 3rd-party dependencies like e.g. OpenCV.

### Checklist
- [ ] tutorials / documentation: Any changes necessary?

Could someone test a binary build in a private buildfarm to check, what other dependencies are necessary?


v4hn:
> However, I would like to question the (nested) folder structure: occupancy_map/occupancy_map_monitor/(src|include).
If we restructure anyways, couldn't we go for a simpler structure: occupancy_map/(src|include)
and name the lib moveit_ros_occupancy_map?

Leaving out the upper folder makes sense to me. :+1: 
I would prefer to name the library either `moveit_occupancy_map_monitor` (as-is) or (for consistency) `moveit_ros_occupancy_map_monitor`.

rhaschke:
@v4hn, is there a special reason that you prefer the more specific name `moveit_ros_occupancy_map`?
My suggestion was to use the more general name  `moveit_ros_occupancy_map` for both package and library. As the library/package comprises 3 classes (`occupancy_map.h, occupancy_map_monitor.h, occupancy_map_updater.h`), I think this can be justified.

v4hn:
> is there a special reason that you prefer the more specific name `moveit_ros_occupancy_map_monitor`?

The package is currently named `occupancy_map_monitor` and adding the `moveit_ros` prefix does not change the semantics, in contrast to your preference.
90% of the package content are directly related to updates and monitoring of the map, which is probably where the original name comes from.
The `occupancy_map.h` you refer to creates a header-only, thin wrapper around `octomap::OcTree` which only exists to make access thread-safe. It's not difficult to come up with a layout where this locking is done by the monitor itself and the class is not needed.
By your reasoning we should actually pick a name like `moveit_ros_octomap` (and I believe this is too generic too).

Dropping the `_monitor` suffix might make people expect another package with the monitor class because the name implies only the base data structure and would change terms more than necessary to achieve consistency.

At the end of the day it's a minor detail either way and the main point is to get rid of the additional folder.

rhaschke:
Thanks for this motivation.

jschleicher:
Any comments regarding the tasks in the description?
I've added `moveit_ros_perception` to the setup assistant template and a note to MIGRATION.md now as suggestion.

rhaschke:
I think the dependency on ros_perception of moveit_config packages is only required when sensing is configured. It wasn't there in the first place.

rhaschke:
Regarding back-porting: I think we could backport this to Melodic. Other opinions?

jonbinney:
Backporting to melodic sounds good to me. If someone was `<depend>ing` from their own package on some package that wasn't the moveit_ros metapackage, then rosdep will no longer auto-install perception stuff for them. But that's an easy fix for them.

tylerjw:
This created a circular dependency that was only hidden because we have such poor test coverage.  I think we should revert this change and @jschleicher can update it to remove the circular dependency.

For information on the circular dependency this created see https://github.com/ros-planning/moveit/issues/2042

In the future changes like this should include tests that follow all paths if the code being moved isn't covered by an existing test.

Thoughts, @rhaschke @v4hn @henningkayser ?

PR #1537, opened 2019-07-04
Cherry-pick #1508 to master

BryceStevenWilley:
### Description

* Python wrapper: add acceleration_scaling_factor argument to retime_trajectory.

* wrap_python_move_group.cpp, move_group.py: add algorithm option to retime_trajectory in pyhon interface of move_group.

* moveit_commander: add python_time_parameterization test

I realized shortly after squashing that #1508 was target for melodic, not master, so the cherry-pick process is happening a little backwards, sorry.


rhaschke:
Cherry-picked directly on cmdline to keep author, time, etc of original commit, which facilitates branch comparison.

PR #1538, opened 2019-07-05
PlanExecution: Return executed trajectory

rhaschke:
Fixes memory leak reported in #1526. Alternative to #1528, which removed TrajectoryMonitor completely.
@jschleicher, could you please validate again?

### DIFF
--- a/moveit_ros/planning/plan_execution/include/moveit/plan_execution/plan_representation.h
+++ b/moveit_ros/planning/plan_execution/include/moveit/plan_execution/plan_representation.h
@@ -73,7 +73,7 @@ struct ExecutableMotionPlan
 
   std::vector<ExecutableTrajectory> plan_components_;
 
-  // The trace of the trajectory recorded during execution
+  /// The trace of the trajectory recorded during execution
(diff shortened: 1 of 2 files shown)


rhaschke:
There is not more memory overhead than before (in the past we were recording the trajectory, but not it). But, you are right, the action result message is bloated.
I don't like the idea to add yet another ROS parameter. As you said, nobody was actually using this feature in the past (and nobody missed it), which is why I originally proposed to drop the trajectory recording completely :wink:
If this feature should be enabled on demand, I suggest having a parameter in the action request message.

jschleicher:
:+1: Thank you!
I can confirm, that the memory footprint no longer increases (at least for some minutes).

The frequency of the trajectory_monitor seems to be adjustable using dynamic_reconfigure (though it doesn't seem to work in a quick test). Maybe setting that parameter to `0` could turn off monitoring, if anyone is hurt by the message overhead. But that's future enhancement...

rhaschke:
That's indeed a great idea! Will have a look later...

PR #1542, opened 2019-07-06
TrajectoryMonitor: zero sampling frequency disables recording

rhaschke:
Implements https://github.com/ros-planning/moveit/pull/1538#issuecomment-508770275.
@v4hn, should we disable the monitoring by default in `PlanExecution`?

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
@@ -40,6 +40,8 @@
 #include <limits>
 #include <memory>
 
+static const std::string LOGNAME = "TrajectoryMonitor";
+
 planning_scene_monitor::TrajectoryMonitor::TrajectoryMonitor(const CurrentStateMonitorConstPtr& state_monitor,
                                                              double sampling_frequency)
   : current_state_monitor_(state_monitor)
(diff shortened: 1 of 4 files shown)


moveit_ros/planning/planning_scene_monitor/src/trajectory_monitor.cpp
```diff
@@ -57,7 +59,7 @@ planning_scene_monitor::TrajectoryMonitor::~TrajectoryMonitor()
 void planning_scene_monitor::TrajectoryMonitor::setSamplingFrequency(double sampling_frequency)
 {
   if (sampling_frequency <= std::numeric_limits<double>::epsilon())
-    ROS_ERROR("The sampling frequency for trajectory states should be positive");
+    ROS_INFO_NAMED(LOGNAME, "Disabling trajectory recording");
   else
     sampling_frequency_ = sampling_frequency;
```
jschleicher: I didn't test it yet, but it looks like you would want to set the member variable in the if-branch as well?

rhaschke: You are absolutely right. I pushed this in a rush w/o prior testing...

rhaschke:
Thanks a lot, @jschleicher, for carefully testing this and revealing those issues. This changed looked so simple, but the devil is in the details (as usual). 
I fixed both issues now, disabled trajectory recording by default, and tested locally :wink:
Could you, please, give it another try?

PR #1360, opened 2019-02-22
A jogging PR for Melodic.

AndyZe:
Updating PR #1244 for Melodic.

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [x] Extended the tutorials / documentation, if necessary [reference](http://moveit.ros.org/documentation/contributing/)
- [ ] Include a screenshot if changing a GUI
- [ ] Document API changes relevant to the user in the moveit/MIGRATION.md notes
- [x] Created tests, which fail without this PR [reference](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/tests.html)
- [ ] Decide if this should be cherry-picked to other current ROS branches
- [ ] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers

[//]: # "You can expect a response from a maintainer within 7 days. If you haven't heard anything by then, feel free to ping the thread. Thank you!"

PR #1549, opened 2019-07-11
Separate experimental pkgs

mlautman:
### Description

This PR separates out moveit_experimental into separate packages for jog arm and the existing moveit_experimental package. This is intended to pave the way for users to selectively build one or more experimental packages

This reverts some of the changes made in https://github.com/ros-planning/moveit/pull/1360
Specifically, it reverts the CMakeLists.txt and package.xml changes

### Checklist


moveit_experimental/jog_arm/config/ur5_simulated_config.yaml
```diff
@@ -0,0 +1,38 @@
...
+# Currently supported are std_msgs/Float64MultiArray (for ros_control JointGroupVelocityController)
+# or trajectory_msgs/JointTrajectory (for Universal Robots and other non-ros_control robots)
+command_out_type: std_msgs/Float64MultiArray
+# Can save some bandwidth as most robots only require positions or velocities
+publish_joint_positions: true
+publish_joint_velocities: false
+publish_joint_accelerations: false
+
+
+
```
AndyZe: Too many blank lines at the end here. Remove 2

v4hn:
> But instead of just moving moveit_experimental/* into moveit_experimental/moveit_experimental/*, i.e. just introducing another folder layer, I suggest to create packages for all entities instead.

:+1: 

>  Another option is to move jog_arm into moveit_ros as a separate package.

I strongly prefer to keep this in `experimental` until we have a number of independent developers using this package (preferably with demo videos). If this is already the case, we can discuss it of course.

AndyZe:
> until we have a number of independent developers using this package (preferably with demo videos)

Well, here's a video of Jeff Bezos using it. I don't mind keeping it in moveit_experimental but I would like to make it easy to copy/paste the package into a catkin ws.

https://www.youtube.com/watch?v=7XLcgEJVWZU

https://github.com/shadow-robot/jog_arm

v4hn:
> Well, here's a video of Jeff Bezos using it.

Ah, that's the same code base. Show-off. Cool though. ;)

> I don't mind keeping it in moveit_experimental but I would like to make it easy to copy/paste the package into a catkin ws.

Of course, that's why we talk about the current pull-request.

mlautman:
> But instead of just moving moveit_experimental/* into moveit_experimental/moveit_experimental/*, i.e. just introducing another folder layer, I suggest to create packages for all entities instead.

My goal here wasn't to restore all of the old projects that live in moveit_experimental but to clear the way for new standalone packages to live in moveit_experimental so Jeff Bazos can beta test them :P

Joking aside, I see moveit_experimental as a staging ground for packages before we move them into the core MoveIt project. The current structure doesn't allow for that and I think that should be changed. I can rename the moveit_experimental package to something like `miscellaneous` so it doesn't live at `moveit_experimental/moveit_experimental` but restoring all of the code in there is out of scope. I think this PR is valuable as is and we should reserve "request changes" for PR's where specific issues exist that would be problematic if merged. Since the request here was for a rename and or follow on work perhaps a comment would be more appropriate.

v4hn:
You make it sound like splitting up the package would fill your whole day.

By introducing a `moveit_experimental` folder that contains packages, you effectively turn it into a meta-package.
So I would propose to actually *have* a meta-package by that name then.

So the name of the old package should be changed then, yes.
`moveit_experimental_misc` sounds viable, although properly splitting up into subpackages would be cleaner.

rhaschke:
As @mlautman, didn't take further action on this, I filed a replacement PR (#1606) that avoids an extra subfolder. Closing the PR here.

Issue #1536, opened 2019-07-04
remove dependency on eigenpy in master


v4hn:
Currently, [moveit_ros_planning_interface depends on eigenpy](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp#L50).
I would like to see this dependency removed again soon, although I do not provide a solution here.

https://github.com/ros-planning/moveit/pull/1518 already provided a replacement for it, but `boost_numpy` did not exist back in xenial, so this is no solution as long as we want to stay compatible with Ubuntu 16.04.

To be precise, we depend on @wxmerkt's [eigen_catkin](https://github.com/ipab-slmc/eigenpy_catkin-release/tree/release/melodic/eigenpy) package which wraps eigenpy as an `ExternalProject` for catkin, although that is [not the correct/preferred way to handle third-party packages](http://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty). In the process it adds a cmake config which [the upstream package](https://github.com/stack-of-tasks/eigenpy) does not provide. As we [`find_package` eigenpy as a catkin COMPONENT](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/CMakeLists.txt#L21), that makes it impossible to use eigenpy installed from upstream (or even use upstream's non-standard PPA builds).
Instead upstream relies on pkg-config, but incidentally their pkgconfig file is broken too - it uses `-isystem` to include numpy which `pkg-config --cflags-only-I eigenpy` fails to extract.
So the build fails to work if we switch to `pkg_check_modules`.

Importantly, the catkin package [does not support a python3 build](https://github.com/ipab-slmc/eigenpy_catkin-release/blob/release/melodic/eigenpy/CMakeLists.txt#L21) and upstream also needs a patch to enable python3 with a newer boost version.
I managed to hack the package [to build with both](https://github.com/v4hn/eigenpy_catkin/tree/python3-boost1.69-fix-numpy-include), but the patches do not generalize beyond my setup and the whole build process of the library involving `ExternalProject` and a [git submodule with a whole cmake subsystem](https://github.com/stack-of-tasks/eigenpy/blob/master/CMakeLists.txt) to detect standard dependencies is rather fragile.

wxmerkt:
> although that is not the correct/preferred way to handle third-party packages

[...]

PR #1550, opened 2019-07-11
Adapt eigenpy to upcoming third-party re-release

wxmerkt:
This is a test to trigger CI to see whether a fixed 3rd-party-released package with a fixed pkg-config would resolve the issues raised in #1536.

### DIFF
--- a/moveit.rosinstall
+++ b/moveit.rosinstall
@@ -27,5 +27,5 @@
 # Temporary for running CI to test new EigenPy
 - git:
     local-name: eigenpy
-    uri: https://github.com/wxmerkt/eigenpy.git
-    version: wxm-update-eigenpy-for-3rd-party-release
+    uri: https://github.com/stack-of-tasks/eigenpy.git
+    version: devel
--- a/moveit.rosinstall
+++ b/moveit.rosinstall
@@ -24,3 +24,8 @@
     local-name: moveit_tutorials
     uri: https://github.com/ros-planning/moveit_tutorials.git
     version: master
+# Temporary for running CI to test new EigenPy
+- git:
+    local-name: eigenpy
+    uri: https://github.com/wxmerkt/eigenpy.git
+    version: wxm-update-eigenpy-for-3rd-party-release
--- a/moveit_ros/planning_interface/CMakeLists.txt
+++ b/moveit_ros/planning_interface/CMakeLists.txt
@@ -75,7 +75,6 @@ catkin_package(
     tf2_geometry_msgs
   DEPENDS
     EIGEN3
-    eigenpy
 )
 
 include_directories(${THIS_PACKAGE_INCLUDE_DIRS}
--- a/moveit_ros/planning_interface/CMakeLists.txt
+++ b/moveit_ros/planning_interface/CMakeLists.txt
@@ -19,12 +19,14 @@ find_package(catkin REQUIRED COMPONENTS
   tf2_eigen
   tf2_geometry_msgs
   tf2_ros
-  eigenpy
   roscpp
   actionlib
   rospy
   rosconsole
 )
+
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(eigenpy eigenpy REQUIRED)
 
 find_package(PythonInterp REQUIRED)
 find_package(PythonLibs "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" REQUIRED)
@@ -73,10 +75,12 @@ catkin_package(
     tf2_geometry_msgs
   DEPENDS
     EIGEN3
+    eigenpy
 )
 
 include_directories(${THIS_PACKAGE_INCLUDE_DIRS}
                     ${Boost_INCLUDE_DIRS}
+                    ${eigenpy_INCLUDE_DIRS}
                     ${catkin_INCLUDE_DIRS})
 
 include_directories(SYSTEM
--- a/moveit_ros/planning_interface/package.xml
+++ b/moveit_ros/planning_interface/package.xml
@@ -33,8 +33,8 @@
   <depend>tf2_geometry_msgs</depend>
   <depend>tf2_ros</depend>
   <depend>python</depend>
-  <depend>eigenpy</depend>
 
+  <build_depend>eigenpy</build_depend>
   <build_depend>eigen</build_depend>
 
   <test_depend>moveit_resources</test_depend>


rhaschke:
Travis reveals two issues:
- `eigenpy` isn't a proper ROS package (anymore?) and it is not listed as a system package for `rosdep`
  Thus, `catkin_lint` complains that you cannot `build_depend` on it w/o configuring it in cmake as a catkin/ROS dependency.
- As a ROS package `eigenpy` should provide a test target.

I think, you should keep your ROS release wrapper around eigenpy, but use pkg-config to find all path settings within the ROS packages' cmake file: As you release eigenpy as a ROS package, it should truly behave like a ROS package.

v4hn:
> eigenpy isn't a proper ROS package

This is not true. It's no *catkin* package.
The infrastructure supports 3rd-party packages not being included via catkin.

> Thus, catkin_lint complains that you cannot build_depend on it w/o configuring it in cmake as a catkin/ROS dependency.

This seems to be a shortcoming of `catkin_lint` then, as it is perfectly viable to include dependencies via `pkg_check_modules` when they do not provide cmake configurations.

> As a ROS package eigenpy should provide a test target.

I believe this is only required by our current CI configuration and generally not enforced for third-party packages?

> I think, you should keep your ROS release wrapper around eigenpy, but use pkg-config to find all path settings within the ROS packages' cmake file

:confused: Playing around with package-config files in 3rd-party releases should be supported, but is not ideal, that's true.
The cleanest solution would be to add a proper cmake configuration to upstream or your fork @wxmerkt .

gavanderhoorn:
> This seems to be a shortcoming of `catkin_lint` then, as it is perfectly viable to include dependencies via `pkg_check_modules` when they do not provide cmake configurations.

No, @rhaschke is correct: the issue is not that `eigenpy` is not a Catkin package, the problem is that any dependency that is not a Catkin package *must* be a system dependency, and system dependencies can only be recognised as such if they have a rule in the `rosdep` database.

As there is no such rule, `catkin_lint` (but really `rospack`/`rosdep`) will assume it must be an unknown/uninstalled Catkin/ROS package.

And as that unknown `eigenpy` ROS package is listed as a `build_depend`, but is *not* listed as a `COMPONENT` of the `find_package(catkin ..)` call nor the `catkin_package(CATKIN_DEPENDS ..`) list, `catkin_lint` complains that it is an "unconfigured build depend".

v4hn:
> the problem is that any dependency that is not a Catkin package *must* be a system dependency,
> and system dependencies can only be recognised as such if they have a rule in the `rosdep` database.

Please give a reference for this rule. I do not believe this is true.
The only difference between this case and [our use of OMPL](https://github.com/ros-planning/moveit/blob/master/moveit_planners/ompl/CMakeLists.txt#L22)
is that OMPL already provides a cmake configuration whereas eigenpy only provides package-config up to now.

The problem will probably not be relevant anymore in a bit because [the author promised to add cmake support](https://github.com/stack-of-tasks/eigenpy/pull/69#issuecomment-510859070).
But my point stands.

gavanderhoorn:
> Please give a reference for this rule. I do not believe this is true.

there is no "rule" for this. It's simply how `rospack`/`rosdep` works afaik.

And `catkin_lint` just makes use of this.

v4hn:
Then I wonder why OMPL does not trigger the same error ;)

wxmerkt:
I think one of the issues could be that `rosdep` currently finds the catkinised wrapper, whereas the locally added (`moveit.rosinstall`) will turn eigenpy from a _catkin package_ into a normal third party released library with pkg-config and in the future cmake config files. It will ship with a `package.xml` manifest.

rhaschke:
I'm afraid @v4hn is right here: ompl [isn't listed as a system package](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml). I think the only difference to eigenpy is that the latter uses pkgconfig instead of cmake to be found.

gavanderhoorn:
I tested this locally with just a `build_depend` on `eigenpy`, nothing in `CMakeLists.txt` -> `catkin_lint` complaints.

Dependency on `ipython`, nothing in `CMakeLists.txt` -> no complaint.

What could have interfered here is that `ipython` is a Python package.

In any case, `catkin_lint` does indeed not consider pkgs other than those occuring in `find_package(..)` and `find_package(catkin ..)` ([here](https://github.com/fkie/catkin_lint/blob/5b94f7c7c7df268ce3e7a67d7cdedc579ab3931e/src/catkin_lint/checks/build.py#L280-L282
)).

It does however also have the logic I described above. But that is not what complains here.

wxmerkt:
Dear All,
Thank you all for your comments and feedback. 

We have now upstreamed the `package.xml` and fixed the previous buildsystem issues. I am close to re-releasing `eigenpy` as a third-party library replacing the current catkin wrapper (background: the buildsystem previously was incompatible with doing a third-party release - as this is now fixed, I want to go the third-party release route and deprecate the catkin-wrapper-workaround).

With eigenpy re-released as a third-party library, are the changes in this PR suitable or are you looking for something else? For testing, I added eigenpy to rosinstall.

Many thanks,
Wolfgang

v4hn:
Thanks for your efforts @wxmerkt !

CI still fails for various structural reasons. Please have a look

wxmerkt:
Thank you @v4hn. The build is failing due to (a) unrelated catkin_lint issues not touched in this PR and (b) the underlying third-party CMake library not having the `tests` target. The latter appears to be an issue with catkin tools - do you know of a workaround?

v4hn:
(a) No, the unrelated warnings should of course also be fixed, but the command fails because of
> moveit_ros_planning_interface: CMakeLists.txt: error: unconfigured build_depend on 'eigenpy'

This seems to be a problem with catkin_lint not recognizing the `pkg_check_module` call as build configuration. This is a shortcoming in the linter that should be resolved there (and possibly there is some magic cmake comment to disable the error until then).

> (b) the underlying third-party CMake library not having the tests target. The latter appears to be an issue with catkin tools - do you know of a workaround?

Well, the easiest workaround would probably require you to add a `tests` target upstream.
Otherwise I'm open to local workarounds - I didn't look into it - but this should really be resolved in catkin_tools.
Alternatively it might be possible to exclude the external project from testing explicitly.

wxmerkt:
> Well, the easiest workaround would probably require you to add a tests target upstream.

I'd prefer not to go this way for a flaw in catkin_tools.

> Otherwise I'm open to local workarounds - I didn't look into it - but this should really be resolved in catkin_tools.

The upstream issue is https://github.com/catkin/catkin_tools/issues/159 - no current activity.

> Alternatively it might be possible to exclude the external project from testing explicitly.

Do you have a pointer where I would set this? I'd like to push ahead with the eigenpy re-release but this will break MoveIt CI as long as it requires the current `make tests` workflow. 

`industrial_ci` for instance first builds with `catkin build` and then runs tests with `catkin run_tests` and checks output with `catkin_test_results`. Would this amendment to `moveit_ci` be a workaround? [We have CIs with `industrial_ci` passing with eigenpy]

wxmerkt:
1. The catkin_lint error won't be fixed until the new eigenpy is merged to rosdistro, cf. https://github.com/fkie/catkin_lint/issues/71
2. The make targets can be worked around by adding ``--continue-on-failure`` is that an acceptable solution? You can check failure status with `catkin_test_results build_directory`

rhaschke:
I had a direct conversation to @wxmerkt. We discussed essentially two options:
1. Restrict the eigenpy release (https://github.com/ros/rosdistro/pull/22798 / https://github.com/ros/rosdistro/pull/22797) and the present PR to Kinetic only.
For Melodic we could remove eigenpy from MoveIt completely, using `boost::numpy` as discussed in https://github.com/ros-planning/moveit/issues/1536. (This option is not feasible for Kinetic, as boost::numpy is not available for Kinetic). However, this solution means that MoveIt Melodic will not compile out-of-the-box on Kinetic anymore. @v4hn would that be ok?
2. eigenpy will provide a cmake config file that provides `_INCLUDES` and `_LIBRARIES` variables (found via the fixed pkgconfig), such that we can continue to catkin-find-package it. Including this into the eigenpy repo might require difficult discussions with those maintainers. Alternatively, we could provide a corresponding cmake config file just via the ROS distribution. Maybe that's the easiest solution for now.

rhaschke:
Note that the new eigenpy is already released into Kinetic (https://github.com/ros/rosdistro/pull/22798) and now [breaks the MoveIt build on the ROS build farm](http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__moveit_ros_planning_interface__ubuntu_xenial_amd64__binary).

v4hn:
> eigenpy will provide a cmake config file that provides _INCLUDES and _LIBRARIES variables (found via the fixed pkgconfig), such that we can continue to catkin-find-package it. Including this into the eigenpy repo might require difficult discussions with those maintainers. Alternatively, we could provide a corresponding cmake config file just via the ROS distribution. Maybe that's the easiest solution for now.

Wow, breaking build API for kinetic seems harsh. :confused: 
As `kinetic` is currently broken, a re-release of eigenpy with a (adapted copy of an older...?) cmake file added through bloom seems the best way to fix it.
If @wxmerkt goes that way anyway for kinetic, it would make sense to me to do the same for melodic.

---

> For Melodic we could remove eigenpy from MoveIt completely, using boost::numpy as discussed in #1536. (This option is not feasible for Kinetic, as boost::numpy is not available for Kinetic).  @v4hn would that be ok?

I'm still not happy with it, but would not block it anymore.
Ultimately, `boost::numpy` should be the way forward, I guess.
For myself, this implies I will have to migrate at least two setups over here to Ubuntu 18.04. (~1 week of work which I don't really have available in the foreseeable future), and at least one colleague who will complain quite a bit.

> However, this solution means that MoveIt Melodic will not compile out-of-the-box on Kinetic anymore.

Just to clarify: Nobody tried yet whether it's actually feasible to have `boost::numpy` available with Ubuntu 16.04 / kinetic.
- It definitely requires compiling boost_numpy, possibly entire boost, by hand
  - This possibly entails required updates of system dependencies (e.g., cmake)
- I'm not sure whether Ubuntu 16.04's python/numpy are compatible with `boost_numpy`
    - It possibly requires an entire ROS system to be setup with a different version of Python
      which takes ... a lot of enthusiasm to setup on Ubuntu 16.04/kinetic

@davetcoleman previously argued to support `master` (which is implied by your proposal for `melodic` I guess) for ROS kinetic because of some of their clients.
@henningkayser told me about rather recent experience with setting up MoveIt `master` on ROS kinetic.

Let's hear about their current opinion.

wxmerkt:
There are two options:

a) You try the eigenpy from https://github.com/stack-of-tasks/eigenpy/pull/101. This is still subject to getting merged upstream in the first place - and possibly may not.
b) A hotfix release for kinetic-devel accepting this PR here.

rhaschke:
> There are two options:
> a) You try the eigenpy from [stack-of-tasks/eigenpy#101](https://github.com/stack-of-tasks/eigenpy/pull/101). This is still subject to getting merged upstream in the first place - and possibly may not.

This didn't fully fix the problem for MoveIt yet. I filed another fix (https://github.com/stack-of-tasks/eigenpy/pull/109) that makes it work for MoveIt.
I prefer this solution over a patch in MoveIt. Hence, closing the PR here.

PR #1551, opened 2019-07-12
Adding continuous collision detection to Bullet

j-petit:
### Description
This PR is a follow up to #1504 which deals with adding Bullet as collision checker (see #1427) and adds continuous collision detection capabilities. ~~It will be considerably smaller when #1504 gets merged.~~

Building requires `libbullet-dev` and `libbullet-extras-dev` which can both be installed through apt.
Bullet passes all [tests which were originally designed for FCL](https://github.com/ros-planning/moveit/pull/1504/files#diff-86244afbfb53a2be7a884f6d519d0873).

See [this high level flowchart](https://drive.google.com/file/d/1FNt09Ntj_5x3fje5KOykB8YIVo7XK4QX/view?usp=sharing) for the collision checking process using Bullet:
Bullet_collision_moveit

To further understand the collision checking process using Bullet see [this detailed flowchart](https://drive.google.com/file/d/19504a0UODZM2b2VoeVvclaVtB3TxlOpk/view?usp=sharing):
BulletChecking_detailed

Thanks for reviewing!

PR #1552, opened 2019-07-12
Speed benchmark script between Bullet and FCL

j-petit:
### Description
The PR provides an easy to use script for benchmarking collision detection libraries. It can clutter a scene with random meshes and then run collision trials. It can also find random robot state which can be in collision or not.

It will be considerable smaller as soon as #1504 gets merged.


j-petit:
Closing this in favor of just single PR #1551 .

PR #1553, opened 2019-07-12
fix World::getTransform

rhaschke:
Looks like World::getTransform is broken since 3e8feeb637 (#1439).


moveit_core/collision_detection/src/world.cpp
```diff
@@ -168,6 +168,7 @@ const Eigen::Isometry3d& World::getTransform(const std::string& name) const
 
 const Eigen::Isometry3d& World::getTransform(const std::string& name, bool& frame_found) const
 {
+  frame_found = true;
   std::map<std::string, ObjectPtr>::const_iterator it = objects_.find(name);
   if (it != objects_.end())
     return it->second->shape_poses_[0];
```
rhaschke: This first return was missing `frame_found` to be set to true beforehand.

PR #1554, opened 2019-07-12
add RobotTrajectory::getDuration()

v4hn:
came up in https://github.com/ros-planning/moveit_task_constructor/pull/97#discussion_r302957574

I would propose to pick this to melodic.

### DIFF
--- a/moveit_core/robot_trajectory/src/robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/src/robot_trajectory.cpp
@@ -66,13 +66,17 @@ const std::string& RobotTrajectory::getGroupName() const
   return EMPTY;
 }
 
+double RobotTrajectory::getDuration() const
+{
+  return std::accumulate(duration_from_previous_.begin(), duration_from_previous_.end(), 0.0);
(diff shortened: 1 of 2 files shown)


moveit_core/robot_trajectory/src/robot_trajectory.cpp
```diff
@@ -66,13 +66,17 @@ const std::string& RobotTrajectory::getGroupName() const
...
+}
+
 double RobotTrajectory::getAverageSegmentDuration() const
 {
   if (duration_from_previous_.empty())
     return 0.0;
   else
-    return std::accumulate(duration_from_previous_.begin(), duration_from_previous_.end(), 0.0) /
-           (double)duration_from_previous_.size();
+    return getDuration() / static_cast<double>(duration_from_previous_.size());
```
henningkayser: Why did you switch to static_cast? just curious

v4hn: It's C++ code. This is preferred over C-style casting.

PR #1562, opened 2019-07-15
Return failed planning result if time parameterization fails

bhomberg:
### Description

We switched to a new time parameterization and noticed that it sometimes failed on our input -- but instead of returning that planning failed, planning succeeded and the resulting trajectory had all times set to 0. This PR makes the time parameterization plugins adjust the return boolean appropriately.

### Checklist
- [ y ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)

### DIFF
--- a/moveit_ros/planning/planning_request_adapter_plugins/src/add_iterative_spline_parameterization.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/src/add_iterative_spline_parameterization.cpp
@@ -68,7 +68,10 @@ public:
       ROS_DEBUG("Running '%s'", getDescription().c_str());
       if (!time_param_.computeTimeStamps(*res.trajectory_, req.max_velocity_scaling_factor,
                                          req.max_acceleration_scaling_factor))
-        ROS_WARN("Time parametrization for the solution path failed.");
+      {
+        ROS_ERROR("Time parametrization for the solution path failed.");
+        result = false;
+      }
     }
 
     return result;
--- a/moveit_ros/planning/planning_request_adapter_plugins/src/add_time_optimal_parameterization.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/src/add_time_optimal_parameterization.cpp
@@ -71,7 +71,10 @@ public:
       TimeOptimalTrajectoryGeneration totg;
       if (!totg.computeTimeStamps(*res.trajectory_, req.max_velocity_scaling_factor,
                                   req.max_acceleration_scaling_factor))
-        ROS_WARN("Time parametrization for the solution path failed.");
+      {
+        ROS_ERROR("Time parametrization for the solution path failed.");
+        result = false;
+      }
     }
 
     return result;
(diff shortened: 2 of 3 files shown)


moveit_ros/planning/planning_request_adapter_plugins/src/add_iterative_spline_parameterization.cpp
```diff
@@ -68,7 +68,10 @@ class AddIterativeSplineParameterization : public planning_request_adapter::Plan
       ROS_DEBUG("Running '%s'", getDescription().c_str());
       if (!time_param_.computeTimeStamps(*res.trajectory_, req.max_velocity_scaling_factor,
                                          req.max_acceleration_scaling_factor))
+      {
         ROS_WARN("Time parametrization for the solution path failed.");
```
gavanderhoorn:
```suggestion
         ROS_ERROR("Time parametrization for the solution path failed.");
```

If we're going to make failure of time parameterisation a negative result (ie: by returning `false`) it becomes an error, so we should notify the user about that.

A warning would imply that it's something to take note of, but is not fatal.

moveit_ros/planning/planning_request_adapter_plugins/src/add_time_optimal_parameterization.cpp
```diff
@@ -71,7 +71,10 @@ class AddTimeOptimalParameterization : public planning_request_adapter::Planning
       TimeOptimalTrajectoryGeneration totg;
       if (!totg.computeTimeStamps(*res.trajectory_, req.max_velocity_scaling_factor,
                                   req.max_acceleration_scaling_factor))
+      {
         ROS_WARN("Time parametrization for the solution path failed.");
```
gavanderhoorn:
```suggestion
         ROS_ERROR("Time parametrization for the solution path failed.");
```

moveit_ros/planning/planning_request_adapter_plugins/src/add_time_parameterization.cpp
```diff
@@ -67,7 +67,10 @@ class AddTimeParameterization : public planning_request_adapter::PlanningRequest
       ROS_DEBUG("Running '%s'", getDescription().c_str());
       if (!time_param_.computeTimeStamps(*res.trajectory_, req.max_velocity_scaling_factor,
                                          req.max_acceleration_scaling_factor))
+      {
         ROS_WARN("Time parametrization for the solution path failed.");
```
gavanderhoorn:
```suggestion
         ROS_ERROR("Time parametrization for the solution path failed.");
```

moveit_ros/planning/planning_request_adapter_plugins/src/add_iterative_spline_parameterization.cpp
bhomberg: Done -- though personally I wouldn't consider a failed plan/time parameterization to be fatal; I often attempt to make plans that aren't necessarily expected to succeed when considering several plan options.  (Though this PR isn't the place to get into a larger discussion of log levels and I'll be consistent with the rest of MoveIt).

v4hn:
Thank you for the contribution @bhomberg !
I fully agree with @gavanderhoorn , if you return a negative planning result, an error is most appropriate here.

PR #1575, opened 2019-07-19
[Windows][master] Fix binary artifact install locations

seanyen:
This pull request is to edit the `install` usage in `moveit` and follow the `catkin` guidance to make them portable.
* https://docs.ros.org/api/catkin/html/howto/format1/building_libraries.html
* https://docs.ros.org/api/catkin/html/howto/format1/building_executables.html
* http://docs.ros.org/melodic/api/catkin/html/howto/format2/installing_python.html


dirk-thomas:
The separation of `CATKIN_GLOBAL_BIN_DESTINATION` and `CATKIN_PACKAGE_BIN_DESTINATION` only exists so that package specific executable which are not *soo* important don't pollute the global namespace.

The first one is already on the `PATH` while the second one isn't. Therefore on Windows `.dll` files should be installed to `CATKIN_GLOBAL_BIN_DESTINATION` in order to not require additional entries for `PATH`.

We won't use option 2 since that would imply diverging from FHS which we use consistently across all platforms. So please follow option 1 as this PR does.

seanyen:
@rhaschke I attended the moveit maintainer meeting today and one takeaway is that it is more preferred to use `master` directly for Windows change, so I redirect my PRs to `master` now. And because there is nothing being merged yet, so feel free to nuke `feature-windows-port` branch. Thanks!

rhaschke:
Unforunately I couldn't make it for the maintainer meeting and thus couldn't argue for my preference of having feature branches. But, it looks like other maintainers do not agree on this point. Thus, feel free to redirect everything to master.

Issue #1565, opened 2019-07-16
No collision checks within its own world


j-petit:
As I wrote a [test-case](https://github.com/j-petit/moveit/blob/444a7bfd4e286482882a09eadaf15af939e80738/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h#L152) for checking collision within a single `CollisionWorld` environment, I noticed that there does not exist the possibility to check collisions within the `CollisionWorld` itself - only to [other `CollisionWorld`s](https://github.com/ros-planning/moveit/blob/56b7c61a88287bf027415d96ce038de668d2e05a/moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_world_fcl.h#L70) (using the FCL collision checking implementation).

Personally, I always assumed this is a very basic functionality and should be included, but maybe I am missing something why this is not necessary (or maybe it is implemented and I just don't see where?).

BryceStevenWilley:
Sorry, I missed this issue. If you have a some spare time while working on documentation you could add that functionality. I don't see any problems with the method, but I also don't quite see the benefits. Often, items in the world will be defined very closely together and sometimes overlapping, so I don't think that colliding objects in a single world print a warning or something like that. If it's something that might be useful for user defining collision worlds, just as an available query, just having it in C++ might not be as useful, and it might need to go in the python interface as well.

j-petit:
You're right does not make too much sense. Closing this.

PR #1584, opened 2019-07-25
Unified Collision Environment Integration

j-petit:
### Description
This PRs purpose is to show what changes are necessary for a unified collision environment instead of keeping separate `collision_robot` and `collision_world`. Main changes are combining `collision_robot_fcl` and `collision_world_fcl`, `collision_world_distance_field` and `collision_robot_distance_field`, `collision_world_hybrid` and `collision_robot_hybrid` into a single environment each derived from the new base class `collision_env`. Furthermore, the planning scene and the collision detector allocator for all detectors had to be modified.

This is independent of integrating Bullet into MoveIt. The current state compiles and passes all tests in in Moveit.

### Related PRs in other MoveIt repos
- [moveit_tutorials](https://github.com/ros-planning/moveit_tutorials/pull/374)
- [moveit_visual_tools](https://github.com/ros-planning/moveit_visual_tools/pull/54)

#### The case for a separate `collision_world` and `collision_robot`
Currently, the collision detection is split into two separate parts. `collision_robot` is responsible for self-collision checks. `collision_world` for collision detection of the robot with the world environment.
For these, the collision function calls contain a `collision_robot` as an argument from which then the link collision geometries can be accessed.

The split was made for multi-threaded collision detection where you have one `collision_robot` per thread and only a single `collision_world` as in motion planning you need to check a large amount of robot poses against the same current world environment. Keeping a single `collision_world` has memory advantages especially in the case of complicated point cloud world environments.

About the multi-threading advantages I am not 100% sure and would appreciate any additional thoughts.

#### FCL implementation of `collision_robot` and `collision_world`
On `collision_robot_fcl`: The robot link geometries are saved as member variables. When self-collision checks are executed a new FCL manager is created and all links are added to it. As the link geometries are only calculated once and stored persistently (including their local AABB), adding them to the new collision manager is cheap. Then, the broadphase collision check with updated global AABB is executed before narrowphase calculations take place for overlapping pairs.

On `collision_world_fcl`: It contains a member collision manager where all world objects are added / removed. This manager has a persistent tree structure for fast overlap calculations. When a robot-world collision check is executed, all robot links are tested individually against the world as we do not need to do self-collision checks here. For this the pre-calculated collision geometries contained in `collision_robot_fcl` are used and their global pose updated with the current `robot_state`.

#### The unified `collision_env`
It combines `collision_robot` and `collision_world` into a single `collision_env` from which specific collision detection environment like `collision_env_fcl` can be derived. 

Additional changes which are not yet reflected in this PR (to keep it simple first) are dropping  `const` from all collision checking functions. This is not necessary for the `collision_env_fcl` as it simply combines `collision_robot_fcl` and `collision_world_fcl` as it is (and there all are `const`), but brings significant advantages for a Bullet unified environment in the future.

The following paragraph is not part of this PR, just general thoughts for improving the collision checking: I would advocate for including the allowed collision matrix (ACM) (which specifies which objects are allowed to collide) into the unified environment instead of providing it as a pointer to each single collision query. I think in most use cases the ACM stays the same between two checks and therefore it makes sense to keep this information persistent in the collision manager. After constructing the `collison_env` with an ACM, the ACM should only be modified through the `collision_env` so the internal collision manager is able to update its broadphase structure with disabled/enabled collisions and can keep those changes persistent. This gives performance improvements as disabled collision are culled even before broadphase checking and keeps the `collision_env` consistent as this is the place to modify any collision checking behaviour. This proposal is not yet reflected in this PR. I will create a separate issue for this.

**Pro's**
The biggest advantage is maintainability. The current split setup is difficult to understand. Changing anything is hard because the interaction and dependencies between `collision_robot` and `collision_world` are sometimes not clear.

Another point is that handling of attached objects which change from being part of the world to being part of the robot (and back) currently is inefficient and could be made very easy in a unified environment as they are naturally part of the same single environment and only some collision flags would need to change.

Collision detection's main application is in computer games and physics simulation where you naturally have a single collision environment. Through mirroring this in MoveIt we can more easily leverage specific capabilities of collision detection libraries which give better performance like very early culling based on groups and masks even before broadphase checks are performed. It reduces the complexity in MoveIt and instead relies more on the intelligence in the collision library which I think is advantageous. Examplary, if one is interested in both collision with itself and the world, a single pass through unified collision environment is possible instead of two separate checks.

The unified environment brings speed improvements for using Bullet (see [this](https://github.com/ros-planning/moveit/issues/1427#issuecomment-514541239)) and also possibly for using FCL. However, compared to the general improvements for using Bullet (in separate `collision_world_bullet` and `collision_robot_bullet`, the gain in performance is not so big. 

**Con's**
- For multi-threaded collision checks, it is not directly clear for me how we can avoid copying the whole `collision_env`. But as maybe memory got cheaper than it was some years ago, these advantages can be forsaken in turn for the above mentioned advantages. Or, other smart copying techniques could be applied (I am no expert here).
- Major changes throughout the codebase are necessary
   - all current collision environments (`collision_world/robot_fcl`, `collision_robot/world_distance_field`, `collision_robot/world_hybrid`) have to be unified
   - dropping `const` goes all up to the planning scene and further


moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -0,0 +1,307 @@
+#include <gtest/gtest.h>
```
davetcoleman: please add BSD license

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -0,0 +1,307 @@
...
+  void TearDown() override
+  {
+  }
+
+protected:
+  bool robot_model_ok_;
+
+  robot_model::RobotModelPtr robot_model_;
+
+  // std::shared_ptr<collision_detection::CollisionEnvBullet> c_env_;
```
davetcoleman: remove?

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -0,0 +1,307 @@
...
+  collision_detection::CollisionEnvPtr c_env_;
+
+  collision_detection::AllowedCollisionMatrixPtr acm_;
+
+  robot_state::RobotStatePtr robot_state_;
+};
+
+/** \brief Correct setup testing. */
```
davetcoleman: love it!

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -0,0 +1,307 @@
...
+  robot_state_->setToDefaultValues();
+  robot_state_->update();
+
+  collision_detection::CollisionRequest req;
+  collision_detection::CollisionResult res;
+  c_env_->checkSelfCollision(req, res, *robot_state_, *acm_);
+  ASSERT_TRUE(res.collision);
+}
+
+// TODO: Add collision check capability within world itself and then enable test
```
davetcoleman: add your usename to the TODO, or a github issue id

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+  c_env_->setLinkPadding("panda_hand", 0.08);
+  c_env_->checkRobotCollision(req, res, *robot_state_, *acm_);
+  ASSERT_TRUE(res.collision);
+  res.clear();
+
+  c_env_->setLinkPadding("panda_hand", 0.0);
+  c_env_->checkRobotCollision(req, res, *robot_state_, *acm_);
+  ASSERT_FALSE(res.collision);
+}
```
davetcoleman: line break

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+  c_env_->checkRobotCollision(req, res, state2, *acm_);
+  ASSERT_FALSE(res.collision);
+  res.clear();
+
+  c_env_->checkRobotCollision(req, res, state1, state2, *acm_);
+  ASSERT_TRUE(res.collision);
+  ASSERT_EQ(res.contact_count, 4u);
+  res.clear();
+}
```
davetcoleman: @ommmid great example of moveit unit tests ;-)

moveit_core/CMakeLists.txt
```diff
@@ -202,6 +202,6 @@ add_subdirectory(planning_interface)
 add_subdirectory(planning_request_adapter)
 add_subdirectory(trajectory_processing)
 add_subdirectory(distance_field)
-add_subdirectory(collision_distance_field)
+#add_subdirectory(collision_distance_field)
```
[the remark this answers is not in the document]
felixvd: Because `collision_distance_field_robot` and `collision_distance_field_world` also need to be merged in the same way as the other files in this PR, but it's not done yet. @j-petit mentioned it in the call although it's not written out here.

moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -678,7 +679,7 @@ class PlanningScene : private boost::noncopyable, public std::enable_shared_from
   double distanceToCollisionUnpadded(const robot_state::RobotState& robot_state,
                                      const collision_detection::AllowedCollisionMatrix& acm) const
   {
-    return getCollisionWorld()->distanceRobot(*getCollisionRobotUnpadded(), robot_state, acm);
+    return getCollisionRobotUnpadded()->distanceRobot(robot_state, acm);
```
BryceStevenWilley:
```suggestion
    return getCollisionEnvUnpadded()->distanceRobot(robot_state, acm);
```

moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -640,7 +641,7 @@ class PlanningScene : private boost::noncopyable, public std::enable_shared_from
    * self-collisions), if the robot has no padding */
   double distanceToCollisionUnpadded(const robot_state::RobotState& robot_state) const
   {
-    return getCollisionWorld()->distanceRobot(*getCollisionRobotUnpadded(), robot_state, getAllowedCollisionMatrix());
+    return getCollisionRobotUnpadded()->distanceRobot(robot_state, getAllowedCollisionMatrix());
```
BryceStevenWilley:
```suggestion
    return getCollisionEnvUnpadded()->distanceRobot(robot_state, getAllowedCollisionMatrix());
```

moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```diff
@@ -625,7 +626,7 @@ class PlanningScene : private boost::noncopyable, public std::enable_shared_from
    */
   double distanceToCollision(const robot_state::RobotState& robot_state) const
   {
-    return getCollisionWorld()->distanceRobot(*getCollisionRobot(), robot_state, getAllowedCollisionMatrix());
+    return getCollisionWorld()->distanceRobot(robot_state, getAllowedCollisionMatrix());
```
BryceStevenWilley:
```suggestion
    return getCollisionEnv()->distanceRobot(robot_state, getAllowedCollisionMatrix());
```

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -292,25 +281,22 @@ void PlanningScene::addCollisionDetector(const collision_detection::CollisionDet
...
 
-  detector->cworld_ = detector->alloc_->allocateWorld(world_);
-  detector->cworld_const_ = detector->cworld_;
+  detector->cenv_ = detector->alloc_->allocateEnv(world_, getRobotModel());
+  detector->cenv_const_ = detector->cenv_;
 
   // Allocate CollisionRobot unless we can use the parent's crobot_.
   // If active_collision_->crobot_ is non-NULL there is local padding and we cannot use the parent's crobot_.
-  if (!detector->parent_ || active_collision_->crobot_)
+  if (!detector->parent_)
```
BryceStevenWilley: According to the comment here, this happens when there is local padding. Since the equilavent, `active_collision_->cenv_` is always non-null, do we have to assume there is always local padding and this becomes `if (true)`?

moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_env_fcl.h
```diff
@@ -0,0 +1,162 @@
...
+
+  /** \brief Updates the specified object in \m fcl_objs_ and in the manager from new data available in the World.
+   *
+   *  If it does not exist in world, it is deleted. If it's not existing in \m fcl_objs_ yet, it's added there. */
+  void updateFCLObject(const std::string& id);
+
+  /// FCL collision manager which handles the collision checking process
+  std::unique_ptr<fcl::BroadPhaseCollisionManagerd> manager_;
+
+  std::map<std::string, FCLObject> fcl_objs_;
```
BryceStevenWilley: Order all protected members below all protected methods.

moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_env_fcl.h
```diff
@@ -0,0 +1,162 @@
...
+
+  /** \brief Bundles the different checkSelfCollision functions into a single function */
+  void checkSelfCollisionHelper(const CollisionRequest& req, CollisionResult& res, const robot_state::RobotState& state,
+                                const AllowedCollisionMatrix* acm) const;
+
+  void checkRobotCollisionHelper(const CollisionRequest& req, CollisionResult& res,
+                                 const robot_state::RobotState& state, const AllowedCollisionMatrix* acm) const;
+
+  /** \brief Construct an FCL collision object from MoveIt's World::Object. */
+  void constructFCLObject(const World::Object* obj, FCLObject& fcl_obj) const;
```
BryceStevenWilley: There are some method name collisions here, e.g., 2 `constructFCLObjects`. Be sure to differentiate which one is World and which one is Robot.

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -0,0 +1,307 @@
...
+  panda_state.setToDefaultValues();
+  double joint2 = -0.785;
+  double joint4 = -2.356;
+  double joint6 = 1.571;
+  double joint7 = 0.785;
+  panda_state.setJointPositions("panda_joint2", &joint2);
+  panda_state.setJointPositions("panda_joint4", &joint4);
+  panda_state.setJointPositions("panda_joint6", &joint6);
+  panda_state.setJointPositions("panda_joint7", &joint7);
+  panda_state.update();
```
BryceStevenWilley: Doesn't this need to set all other joints to 0 explicitly?

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -0,0 +1,307 @@
...
+  double joint_4{ -1.6 };
+  state2.setJointPositions("panda_joint2", &joint_2);
+  state2.setJointPositions("panda_joint4", &joint_4);
+  state2.update();
+
+  c_env_->checkRobotCollision(req, res, state1, state2, *acm_);
+  ASSERT_FALSE(res.collision);
+  res.clear();
+
+  // Adding the box which is not in collision with the individual states but with the casted one.
```
BryceStevenWilley: nit: FCL at the moment isn't casting, it's doing discrete edge checks.

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
```diff
@@ -0,0 +1,307 @@
...
+  state2.setJointPositions("panda_joint5", &joint_5_other);
+  state2.setJointPositions("panda_joint7", &joint7);
+  state2.update();
+
+  c_env_->checkSelfCollision(req, res, state2, *acm_);
+  ASSERT_FALSE(res.collision);
+  res.clear();
+
+  // c_env_->checkSelfCollision(req, res, state1, state2, *acm_);
+  ROS_INFO_STREAM("Continous to continous collisions are not supported yet, therefore fail here.");
```
BryceStevenWilley: FCL should support this: if you aren't planning on templatizing this test on both Bullet and FCL, change this.

moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp
```diff
@@ -0,0 +1,441 @@
...
+  std::size_t index;
+  robot_geoms_.resize(robot_model_->getLinkGeometryCount());
+  robot_fcl_objs_.resize(robot_model_->getLinkGeometryCount());
+  // we keep the same order of objects as what RobotState *::getLinkState() returns
+  for (auto link : links)
+    for (std::size_t j = 0; j < link->getShapes().size(); ++j)
+    {
+      FCLGeometryConstPtr g = createCollisionGeometry(link->getShapes()[j], getLinkScale(link->getName()),
+                                                      getLinkPadding(link->getName()), link, j);
+      if (g)
```
BryceStevenWilley: Avoid single letter variable names.

moveit_core/collision_detection/src/collision_env.cpp
```diff
@@ -0,0 +1,298 @@
...
+    link_scale_[link->getName()] = scale;
+  }
+  if (!u.empty())
+    updatedPaddingOrScaling(u);
+}
+
+void CollisionEnv::setLinkPadding(const std::string& link_name, double padding)
+{
+  bool update = getLinkPadding(link_name) != padding;
+  link_padding_[link_name] = padding;
```
BryceStevenWilley: Can we add checks to `validatePadding` here?

moveit_core/collision_detection/src/collision_env.cpp
```diff
@@ -0,0 +1,298 @@
...
+    return 0.0;
+}
+
+void CollisionEnv::setLinkPadding(const std::map<std::string, double>& padding)
+{
+  std::vector<std::string> u;
+  for (const auto& link_pad_pair : padding)
+  {
+    bool update = getLinkPadding(link_pad_pair.first) != link_pad_pair.second;
+    link_padding_[link_pad_pair.first] = link_pad_pair.second;
```
BryceStevenWilley: Same as above: should add `validatePadding` here.

moveit_core/collision_detection/src/collision_env.cpp
```diff
@@ -0,0 +1,298 @@
...
+}
+
+const std::map<std::string, double>& CollisionEnv::getLinkPadding() const
+{
+  return link_padding_;
+}
+
+void CollisionEnv::setLinkScale(const std::string& link_name, double scale)
+{
+  bool update = getLinkScale(link_name) != scale;
```
BryceStevenWilley: Should add `validateScale` here.

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
j-petit: yes

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
j-petit: I removed this test, see #1565 for reference.

moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
j-petit: I have kept the getCollisionRobot[Unpadded] / getCollisionWorld functions for now and made them return the same respective collision environment. As these function might be used also in user code, I would add an ROS_ERROR_MESSAGE stating that these functions will be eventually deprecated and getCollisionEnv[Unpadded] should be used. Or what is the common way to handle functions which will be deprecated in the future?

But throughout the moveit codebase, I should change all getCollisionWorld / getCollisionRobot to a new getCollisionEnv function right?

j-petit:
How should we handle the `all_valid` collision detector? I don't really see a sense in keeping something which just always returns no collision detected.

moveit_core/planning_scene/src/planning_scene.cpp
j-petit: I am really unsure about all of this parent planning scene. I think the behavior before was:

`cworld_` is always allocated new.  `crobot_` is only allocated if we don't have a parent planning scene or it is the active detector. The unpadded version is allocated if we don't have a parent (although I thought it should only be allocated when the padding is changed...).

With `cenv_` it should be: `cenv_` is always allocated and the unpadded version only when the padding changes?

moveit_core/collision_detection/src/collision_env.cpp
j-petit: added it.

moveit_core/collision_detection/src/collision_env.cpp
j-petit: added it.

moveit_core/collision_detection/src/collision_env.cpp
j-petit: yes added it.

moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp
j-petit: changed it to `link_geometry`

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
j-petit: No, this happens through setToDefaultValues()

moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_env_fcl.h
j-petit: I changed them to `constructFCLObjectRobot` and `constructFCLObjectWorld` respectively

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
j-petit: right

moveit_core/collision_detection_fcl/test/test_fcl_env.cpp
j-petit: I don't think I have enough time to get CCD for FCL working during GSoC so this test will not be templated and only be checked for bullet. So remove here, right?

j-petit:
I was successful in merging `collision_distance_field` and `collision_hybrid` into single collision environments. MoveIt as a whole compiles now (except MoveIt tutorials which needs to include the new correct header but I cannot change this here as it is not part of this repo) and all existing tests pass. However, I am not too sure if everything is really working as before. Ideas how to best ensure this?

felixvd:
> Ideas how to best ensure this?

Write more, many more tests and improve coverage? I can't think of much else, except to ask for manual beta testing.

moveit_core/planning_scene/src/planning_scene.cpp
felixvd: I want to defer to someone else on this, because I never felt like I really figured out the intricacies of these copies and pointers. But I note that reducing this complexity is another big reason why this PR has value, in my opinion.

simonschmeisser:
> How should we handle the `all_valid` collision detector? I don't really see how something which just always returns no collision is useful as it only inflates the codebase.

AFAIK MoveIt's PlanningSceneDisplay in rviz uses a replicated `PlanningSceneMonitor` which includes everything necessary for collision checking. If you don't need those collision checking features it saves quite some resources to use `all_valid` as it does not build up any collision data structures on every world change. Also it's about three files right?

simonschmeisser:
I'm currently on vacation so I cannot test your code for regression wrt multi-threaded collision checking. Running some OMPL planning should give you a simple test case however as it uses several threads in most algorithms.

j-petit:
> AFAIK MoveIt's PlanningSceneDisplay in rviz uses a replicated `PlanningSceneMonitor` which includes everything necessary for collision checking. If you don't need those collision checking features it saves quite some resources to use `all_valid` as it does not build up any collision data structures on every world change. Also it's about three files right?

Yes, it is just three files.

> Note: this feature is not very accessible as of now ... I hope to do something about it at some point

Which feature do you mean? The `all_valid` collision checker or this PR?

simonschmeisser:
> > Note: this feature is not very accessible as of now ... I hope to do something about it at some point
> 
> Which feature do you mean? The `all_valid` collision checker or this PR?

I was referring to my memory that switching the collision checking plugin of the display is complicated or maybe only possible from code and not by configuration. Not related to this PR at all :) 
> 
> -- 
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/ros-planning/moveit/pull/1584#issuecomment-51811887

-- 
Gesendet von meinem Jolla

davetcoleman:
> I would advocate for including the allowed collision matrix (ACM) (which specifies which objects are allowed to collide) into the unified environment instead of providing it as a pointer to each single collision query. I think in most use cases the ACM stays the same between two checks and therefore it makes sense to keep this information persistent in the collision manager

This concerns me, as I've seen use cases where you disable collision checking between fingers / grippers and objects / surfaces when doing fine grasping approaches.

moveit_core/collision_detection/include/moveit/collision_detection/collision_env.h
```diff
@@ -34,42 +34,51 @@
 
 /* Author: Ioan Sucan */
```
davetcoleman: Please add yourself as an author also

moveit_core/collision_detection/src/collision_env.cpp
```diff
@@ -34,7 +34,7 @@
 
 /* Author: Ioan Sucan */
```
davetcoleman: Please add yourself as author

moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_env_hybrid.h
```diff
@@ -0,0 +1,154 @@
...
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: E. Gil Jones */
```
davetcoleman: Please addyourself as an author also, and add a descriptin of what this file does as documentation

moveit_core/collision_distance_field/src/collision_env_distance_field.cpp
```diff
@@ -35,54 +35,61 @@
...
 #include <ros/console.h>
 #include <ros/assert.h>
 #include <tf2_eigen/tf2_eigen.h>
+#include <moveit/collision_distance_field/collision_env_distance_field.h>
+#include <moveit/collision_distance_field/collision_common_distance_field.h>
+#include <moveit/distance_field/propagation_distance_field.h>
+#include <moveit/collision_distance_field/collision_detector_allocator_distance_field.h>
+#include <boost/bind.hpp>
+#include <memory>
+#include <utility>
```
davetcoleman: Are there any unit tests for this collision checker? Maybe add some to make sure there are no breakages, making sure they pass before and after this PR?

j-petit:
> > I would advocate for including the allowed collision matrix (ACM) (which specifies which objects are allowed to collide) into the unified environment instead of providing it as a pointer to each single collision query. I think in most use cases the ACM stays the same between two checks and therefore it makes sense to keep this information persistent in the collision manager
> 
> This concerns me, as I've seen use cases where you disable collision checking between fingers / grippers and objects / surfaces when doing fine grasping approaches.

But these changes can be done through the collision environment instead of directly changing the ACM. I think in most collision cases the ACM stays the same, right? So it makes sense to store this information persistently in the broadphase collision checking and register any changes to the ACM direclty in the broadphase.

moveit_core/collision_detection/include/moveit/collision_detection/collision_env.h
j-petit: done

moveit_core/collision_detection/src/collision_env.cpp
j-petit: done

moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_env_hybrid.h
j-petit: done

moveit_core/collision_distance_field/src/collision_env_distance_field.cpp
j-petit: Yes there is in [moveit_core/collision_distance_field/test/test_collision_distance_field.cpp](https://github.com/ros-planning/moveit/pull/1584/files?file-filters%5B%5D=.cpp#diff-9a16e0b8917fb63dce1d48d83243106b) and all pass.

j-petit:
I added [the PR](https://github.com/ros-planning/moveit_tutorials/pull/374) for `moveit_tutorials` which would allow Travis to build and run tests.

moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
felixvd: IIRC the currently correct way is to mark it deprecated using the moveit macros for one release or year (?), and then remove it. See the discussion and link to the macros [here](https://github.com/ros-planning/moveit/issues/1229)

BryceStevenWilley: That's fair. Not sure if it's considered marking deprecated if we're changing the return type, but I guess this would still work okay.

j-petit: You're right Bryce we are changing the return type so users will get errors anyway...

j-petit:
I added a PR in moveit_visual_tools which adapts for the unified collision environment changes [here](https://github.com/ros-planning/moveit_visual_tools/pull/54).

davetcoleman:
> > > I would advocate for including the allowed collision matrix (ACM) (which specifies which objects are allowed to collide) into the unified environment instead of providing it as a pointer to each single collision query. I think in most use cases the ACM stays the same between two checks and therefore it makes sense to keep this information persistent in the collision manager
> > 
> > 
> > This concerns me, as I've seen use cases where you disable collision checking between fingers / grippers and objects / surfaces when doing fine grasping approaches.
> 
> But these changes can be done through the collision environment instead of directly changing the ACM. I think in most collision cases the ACM stays the same, right? So it makes sense to store this information persistently in the broadphase collision checking and register any changes to the ACM direclty in the broadphase.

If its possible to do the above scenario, then this sounds good to me. I don't exactly follow your approach, but that's just my ignorance... you're certainly more expert at this part of the code base now :-)

davetcoleman:
We switched to ``pragma once``... i bet that is what is causing these merge conflicts

moveit_core/collision_detection/src/allvalid/collision_env_allvalid.cpp
```diff
@@ -32,118 +32,104 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  *********************************************************************/
 
-/* Author: Ioan Sucan */
+/* Author: Ioan Sucan, Jia Pan, Jens Petit */
```
BryceStevenWilley: Why are you adding Jia Pan as an author here?

moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp
```diff
@@ -318,39 +314,39 @@ TEST_F(FclCollisionDetectionTester, AttachedBodyTester)
...
 
   acm_->setEntry("coll", "r_gripper_palm_link", true);
   res = collision_detection::CollisionResult();
-  cworld_->checkRobotCollision(req, res, *crobot_, robot_state, *acm_);
+  c_env_->checkRobotCollision(req, res, robot_state, *acm_);
   ASSERT_TRUE(res.collision);
 }
 
-TEST_F(FclCollisionDetectionTester, DiffSceneTester)
+TEST_F(FclCollisionDetectionTester, DISABELD_DiffSceneTester)
```
BryceStevenWilley: What's the reason for this being disabled?

felixvd: Also typo

moveit_core/collision_detection/src/allvalid/collision_env_allvalid.cpp
j-petit: Because he is mentioned in `collision_robot_allvalid.cpp` see [here](https://github.com/ros-planning/moveit/blob/4ca54cd0cda1bc0ceef072d769357ce3802e62da/moveit_core/collision_detection/src/allvalid/collision_robot_allvalid.cpp#L35)

j-petit:
> We switched to `pragma once`... i bet that is what is causing these merge conflicts

Yes, rebased on master and changed new headers to also use `pragma once`.

moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp
j-petit: Good catch Bryce! I disabled this when I was developing the `unified_env` and had not yet implemented / combined all of the constructors. I enabled the test and it passes.

j-petit:
> If its possible to do the above scenario, then this sounds good to me. I don't exactly follow your approach, but that's just my ignorance... you're certainly more expert at this part of the code base now :-)

Anyway, this is not something for this PR. I will create issue(s) with thoughts about improvements to the collision checking process during the next days as part of wrapping up the GSoC.

davetcoleman:
Let's merge this! Is tomorrow (Tue) a good time for you @j-petit to be on standby to fix any issues we missed across multiple repos?

fyi @rhaschke

j-petit:
> Let's merge this! Is tomorrow (Tue) a good time for you @j-petit to be on standby to fix any issues we missed across multiple repos?
> 
> fyi @rhaschke

Exciting! Today (Tuesday) I can't be available too long (only until 5 pm UTC + 2), but on Wednesday I could be available till late. So maybe postpone to Wednesday?

davetcoleman:
@BryceStevenWilley can you do the honors of merging this and the related PRs in other repos? 

@j-petit what time will you be available Wed?

j-petit:
I am available from now (8 am UTC+2) until 8 pm UTC+2.

BryceStevenWilley:
Lets get this rolling: squash merging now.

BryceStevenWilley:
Squash completed locally, building locally to before pushing.

BryceStevenWilley:
Everything (except moveit_tutorials on master) build locally right now, but unfortunately, I cannot push the changes with Travis failing, even with `--force`. 
```
$ git push --force upstream master
Counting objects: 97, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (50/50), done.
Writing objects: 100% (97/97), 44.98 KiB | 14.99 MiB/s, done.
Total 97 (delta 53), reused 70 (delta 33)
remote: Resolving deltas: 100% (53/53), completed with 39 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: Required status check "continuous-integration/travis-ci" is expected. At least 1 approving review is required by reviewers with write access.
To https://github.com/ros-planning/moveit
 ! [remote rejected]     master -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/ros-planning/moveit'
```
According to [this Github article](https://help.github.com/en/articles/about-required-status-checks), only administrators can ignore failing statuses (@davetcoleman @rhaschke).  

In the meantime, I've pushed the squashed commit to master [on my fork](https://github.com/BryceStevenWilley/moveit/commit/b060dd7cb66b5c4595b77cc440c62ec3893a98f2) if you just want to grab and push that to master.

davetcoleman:
image

v4hn:
@j-petit @gavanderhoorn The API change broke compatibility with [stomp_ros](https://github.com/ros-industrial/stomp_ros).

@j-petit Could you briefly look at it and maybe provide a patch?

j-petit:
Yes, I'll look into it.

j-petit:
Made the patch, see [here](https://github.com/ros-industrial/stomp_ros/pull/9)

PR #1586, opened 2019-07-29
clang-tidy fixup

v4hn:
Should fix the CI fails currently observed


v4hn:
due to access restriction which hopefully have been agreed by a group of maintainers at some point,
I'm not allowed to fix CI for the master branch.

PR #1593, opened 2019-08-01
MVP TrajOpt Planner Plugin

ommmid:
TrajOpt working as the motion planner plugin for MoveIt, collision cost is created using tesseract and needs to be replaced by information from MoveIt.
 
- The tesseract env and kin are not used at all. They are replaced with what we have in planning_scene, for that I copied and adjusted the needed files from trajopt_ros, i.e., problem_description and kinematic_terms. This is only for the current implemented joint constraint. More work is needed to replace tesseract env and kin with planning scene information for collision cost function.
- The constraint from planning interface request is converted to the constraint we need in trajopt (JointPoseTermInfo) and that is only for joint state constraint for now.
- The output solution from trajopt is fed to PlanningInterface::MotionPlanResponse.
- A yaml file is containing the parameters needed for trajopt algorithm. Still, initial trajectory is hard-coded but needs to be defined by the user.
- JointVelTermInfo is needed to have velocity constraint but it needs to be defined by the user, it is hard coded now.

- There are still some dependencies to trajopt but gradually by converting the constraints to MoveIt compatible counterpart, we can remove those dependencies too.

- The last commit used from trajopt_ros:
ba56be9a0ed71f2f9ad17a4b135e466965a335be

- There is no dependency to tesseract directly but this package is needed for trajopt_ros. The last commit used from tesseract is:
740e6607423e2afbeba04b609e118043c86b7266

@Levi-Armstrong @mpowelson Can you please review this PR?


moveit_planners/trajopt/src/kinematic_terms.cpp
```diff
@@ -0,0 +1,67 @@
...
+  // Note that for equality terms tols are 0, so error is effectively doubled
+  VectorXd result(vel.rows() * 2);
+  result.topRows(vel.rows()) = -(upper_tol_ - (vel.array() - target_));
+  result.bottomRows(vel.rows()) = lower_tol_ - (vel.array() - target_);
+  return result;
+}
+
+MatrixXd JointVelJacobianCalculator::operator()(const VectorXd& var_vals) const
+{
+  // var_vals = (theta_t1, theta_t2, theta_t3 ... 1/dt_1, 1/dt_2, 1/dt_3 ...)
```
BryceStevenWilley: These two comments are describing the same thing in very different ways. IMO I like the latter better, can we stick to that?

moveit_planners/trajopt/src/kinematic_terms.cpp
```diff
@@ -0,0 +1,67 @@
...
+  int half = num_vals / 2;
+  int num_vels = half - 1;
+  MatrixXd jac = MatrixXd::Zero(num_vels * 2, num_vals);
+
+  for (int i = 0; i < num_vels; i++)
+  {
+    // v = (j_i+1 - j_i)*(1/dt)
+    // We calculate v with the dt from the second pt
+    int time_index = i + half + 1;
+    // dv_i/dj_i = -(1/dt)
```
BryceStevenWilley: In this case, I don't think the comments are very helpful.

moveit_planners/trajopt/src/trajopt_planner_manager.cpp
```diff
@@ -33,26 +33,41 @@
...
 /* Author: Omid Heidari
-   Desc:   Main TrajOpt planning plugin interface to MoveIt
+   Desc:   TrajOpt planning plugin
 */
 
 #include <moveit/planning_interface/planning_interface.h>
-#include <moveit/planning_scene/planning_scene.h>
-#include <moveit/robot_state/conversions.h>
+
+#include <moveit/collision_detection_fcl/collision_detector_allocator_fcl.h>
```
BryceStevenWilley: Now that you've rebased onto the feature branch, you should be able to use Bullet as a checker. Change this to `collision_dection` (no FCL) and get the loaded Plugin instead (see how OMPL does this)

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include <moveit/robot_model_loader/robot_model_loader.h>
+#include <moveit/planning_scene/planning_scene.h>
+#include <moveit/robot_state/robot_state.h>
+
+#include <trajopt/common.hpp>
+
+#include "trajopt_interface/problem_description.h"
+
+namespace trackpose_cpp
```
BryceStevenWilley: What's the reason for this namespace?

davetcoleman:
> collision detection is done from tesseract and needs to be replaced by bullet in MoveIt.

I think this is a type, you mean FCL, correct? There's no dependency on Tesseract. Could you edit your top post?

davetcoleman:
> More work is needed to replace tesseract env and kin with planning scene information for collision cost function.

Are all the TODOs, such as this one, documented inline in the code for future reference?

davetcoleman:
> Still, initial trajectory is hard-coded but needs to be defined by the user.

What do you mean by this? The initial guessed trajectory? Would the straight-line interpolation from start to goal be a good initial trajectory?

davetcoleman:
> There is no dependency to tesseract directly but this package is needed for trajopt_ros. The last commit used from tesseract is:

trajopt_ros depends on Tesseract? That's not ideal... can we work with the trajopt_ros maintainers to fix this?

moveit_planners/trajopt/CMakeLists.txt
```diff
@@ -7,30 +7,122 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
...
+
+      trajopt
 )
 
 catkin_package(
-  CATKIN_DEPENDS
-    moveit_core
+    INCLUDE_DIRS include
+    LIBRARIES
+      interactivity_utils
```
davetcoleman: interactive markers are not used, remove

moveit_planners/trajopt/package.xml
```diff
@@ -16,10 +17,21 @@
...
 
-  <depend>roscpp</depend>
-  <depend>moveit_core</depend>
   <depend>pluginlib</depend>
 
+  <depend>moveit_core</depend>
+  <depend>moveit_ros_planning</depend>
+  <depend>moveit_ros_planning_interface</depend>
+
+  <depend>interactive_markers</depend>
```
davetcoleman: interactive markers are not used, remove

moveit_planners/trajopt/CMakeLists.txt
```diff
@@ -7,30 +7,122 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
...
+  DIRECTORY
+    launch
+  DESTINATION
+    ${CATKIN_PACKAGE_SHARE_DESTINATION}
+)
 
 install(FILES trajopt_interface_plugin_description.xml
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
+
+
```
davetcoleman: remove excessive line breaks here and everywhere

moveit_planners/trajopt/config/setup.yaml
```diff
@@ -0,0 +1,50 @@
+trajopt_param:
```
davetcoleman: beautiful documentation!

moveit_planners/trajopt/include/trajopt_interface/kinematic_terms.h
```diff
@@ -0,0 +1,88 @@
+#pragma once
```
davetcoleman: This is not allowed, but I think it should be now: https://github.com/ros-planning/moveit/issues/1597

moveit_planners/trajopt/include/trajopt_interface/kinematic_terms.h
```diff
@@ -0,0 +1,88 @@
...
+
+namespace trajopt_interface
+{
+Vector3d rotVec(const Matrix3d& m)
+{
+  Quaterniond q;
+  q = m;
+  return Vector3d(q.x(), q.y(), q.z());
+}
+
```
davetcoleman: add doxygen on what these functions do, there function names are already too terse to be helpful

moveit_planners/trajopt/include/trajopt_interface/trajopt_interface.h
```diff
@@ -0,0 +1,81 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019, PickNik, Inc.
```
davetcoleman: LLC, not Inc ;-)

moveit_planners/trajopt/include/trajopt_interface/trajopt_interface.h
```diff
@@ -0,0 +1,81 @@
...
+  void setProblemInfoParam(ProblemInfo& problem_info);
+  void setJointPoseTermInfoParams(JointPoseTermInfoPtr& jp, std::string name);
+
+  ros::NodeHandle nh_;  /// The ROS node handle
+  sco::BasicTrustRegionSQPParameters params_;
+  std::vector<sco::Optimizer::Callback> callbacks_;
+  TrajOptProblemPtr prob_;
+};
+
+void callBackFunc(sco::OptProb* opt_prob, sco::OptResults& opt_res);
```
davetcoleman: remove?

.gitignore
```diff
@@ -58,3 +58,7 @@ CATKIN_IGNORE
 .moveit_ci
 
 *.pyc
+
+#gdb
+*.gdb
+/moveit_planners/trajopt/launch/gdb_settings.gdb
```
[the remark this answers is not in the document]
ommmid: right, I think the best place would be MoveIt tutorial. I will move the launch file there.

moveit_planners/trajopt/include/trajopt_interface/trajopt_interface.h
ommmid: Left it for later development but we can remove it or I can add a TODO

moveit_planners/trajopt/test/trajectory_test.cpp
ommmid: mistake, removed

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -0,0 +1,434 @@
...
+  joint_vel->term_type = trajopt_interface::TT_COST;
+  problem_info.cost_infos.push_back(joint_vel);
+
+  ROS_INFO(" ======================================= Visibility Constraints");
+  if (!req.goal_constraints[0].visibility_constraints.empty())
+  {
+    // TODO: Add visibility constraint
+  }
+
+  ROS_INFO(" ======================================= DEBUG");
```
BryceStevenWilley: If these are debug prints, shouldn't they be `ROS_DEBUG`?

moveit_planners/trajopt/config/setup.yaml
```diff
@@ -0,0 +1,50 @@
...
+  improve_ratio_threshold: 0.25   # minimum ratio true_improve/approx_improve to accept step. [Research Paper] TrueImprove/ModelImprove > c
+  min_trust_box_size: 1e-4        # if trust region gets any smaller, exit and report convergence. [Research Paper] xtol
+  min_approx_improve: 1e-4        # if model improves less than this, exit and report convergence
+  min_approx_improve_frac: -.Inf  # if model improves less than this, exit and report convergence
```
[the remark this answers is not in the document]
ommmid: https://yaml.org/refcard.html

moveit_planners/trajopt/include/trajopt_interface/problem_description.h
```diff
@@ -0,0 +1,360 @@
...
+
+Note: Velocity is calculated numerically using forward finite difference
+
+\f{align*}{
+  cost = \sum_{t=0}^{T-2} \sum_j c_j (x_{t+1,j} - x_{t,j})^2
+\f}
+where j indexes over DOF, and \f$c_j\f$ are the coeffs.
+*/
+
+struct TermInfo;
```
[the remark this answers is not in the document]
ommmid: They are being used, in problem_description.cpp for example

moveit_planners/trajopt/include/trajopt_interface/problem_description.h
```diff
@@ -0,0 +1,360 @@
...
+  /** @brief Default value the final column of the optimization is initialized too if time is being used */
+  double dt = 1.0;
+};
+
+/**
+When cost or constraint element of JSON doc is read, one of these guys gets
+constructed to hold the parameters.
+Then it later gets converted to a Cost object by the hatch method
+*/
+struct TermInfo
```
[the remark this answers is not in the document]
ommmid: Agree, it is confusing. Will keep work on it to find a better pattern.

moveit_planners/trajopt/package.xml
```diff
@@ -1,7 +1,8 @@
 <package format="2">
   <name>moveit_planners_trajopt</name>
   <version>1.0.1</version>
-  <description>TrajOpt planning plugin</description>
+
+  <description>TrajOpt planning plugin, an optimization based motion planner </description>
 
   <author email="omid@picknik.ai">Omid Heidari</author>
```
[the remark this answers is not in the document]
ommmid: Correct, I change it to another email address.

moveit_planners/trajopt/src/kinematic_terms.cpp
ommmid: sorry, I did not get it. which two comments?

moveit_planners/trajopt/config/setup.yaml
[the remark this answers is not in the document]
BryceStevenWilley: Oh, neat. :+1:

moveit_planners/trajopt/src/trajopt_interface.cpp
ommmid: yes. I change them to ROS_DEBUG

moveit_planners/trajopt/config/setup.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+  init_info:
+    type: 3                    # 1 = STATIONARY, 2 = JOINT_INTERPOLATED, 3 = GIVEN_TRAJ
+    # request.start_state should be provided for JOINT_INTERPOLATED and GIVEN_TRAJ
+    dt: 0.5
+
+joint_pos_term_info:
+  start_pos:   # from req.start_state
+    name: start_pos
+    first_timestep: 10                # First time step to which the term is applied.
```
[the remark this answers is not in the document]
ommmid: start state is not necessarily the same as current state.

ommmid:
> > More work is needed to replace tesseract env and kin with planning scene information for collision cost function.
> 
> Are all the TODOs, such as this one, documented inline in the code for future reference?

Yes, I added TODOs. I am still working on the collision cost though

ommmid:
> > collision detection is done from tesseract and needs to be replaced by bullet in MoveIt.
> 
> I think this is a type, you mean FCL, correct? There's no dependency on Tesseract. Could you edit your top post?

FCL is used for planning scene (changed it to bullet). But to make collision cost in trajopt, it still uses tesseract env and its own collision.

ommmid:
> > Still, initial trajectory is hard-coded but needs to be defined by the user.
> 
> What do you mean by this? The initial guessed trajectory? Would the straight-line interpolation from start to goal be a good initial trajectory?

Yes, the initial guessed trajectory and an interpolation would be a good option but the user has the option to define it however they want

moveit_planners/trajopt/config/setup.yaml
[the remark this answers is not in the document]
BryceStevenWilley: I understand the distinction that Levi has explained, but the way MoveIt uses it, the start state is usually the current state (or it's the first state in the trajectory).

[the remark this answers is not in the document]
ommmid: In this example, I considered the start state to be different from current state so we can see the difference between setting start_fixed to true and false. I think MotionPlanning Display in rviz sends the current state in request.start_state, in this case, yes, we only need to change this timestep to 1 and set start_fixed to false as we do not want two constraints (which are the same) at one timestep. start_fixed = true adds a constraint at the current joint state.

moveit_planners/trajopt/include/trajopt_interface/trajopt_interface.h
davetcoleman: add TODO on why its empty

moveit_planners/trajopt/include/trajopt_interface/kinematic_terms.h
```diff
@@ -0,0 +1,94 @@
...
+#include <moveit/planning_scene/planning_scene.h>
+
+#include <Eigen/Geometry>
+
+namespace trajopt_interface
+{
+/**
+ * @brief Extracts the vector part of quaternion
+ */
+inline Eigen::Vector3d rotVec(const Eigen::Matrix3d& m)
```
davetcoleman: rename all these functions to explain what they do, similar to the comments

moveit_planners/trajopt/src/problem_description.cpp
```diff
@@ -0,0 +1,567 @@
+
```
davetcoleman: add license (copy from tesseract)

davetcoleman: if there isn't a license, we can copy it from the root of tesseract

moveit_planners/trajopt/src/problem_description.cpp
```diff
@@ -0,0 +1,567 @@
...
+#include <trajopt/trajectory_costs.hpp>
+#include <trajopt_sco/expr_op_overloads.hpp>
+#include <trajopt_sco/expr_ops.hpp>
+#include <trajopt_utils/eigen_conversions.hpp>
+#include <trajopt_utils/eigen_slicing.hpp>
+#include <trajopt_utils/logging.hpp>
+#include <trajopt_utils/vector_ops.hpp>
+
+#include "trajopt_interface/problem_description.h"
+#include "trajopt_interface/kinematic_terms.h"
```
davetcoleman: Is this file copied in temporarily? If so, please explain in a TODO

moveit_planners/trajopt/src/problem_description.cpp
```diff
@@ -0,0 +1,567 @@
...
+      prob.getConstraints().back()->setName(name);
+    }
+  }
+  else
+  {
+    ROS_WARN("JointPosTermInfo does not have a valid term_type defined. No cost/constraint applied");
+  }
+}
+
+void JointVelTermInfo::hatch(TrajOptProblem& prob)
```
davetcoleman: can we use a better name than hatch in all of these?

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -0,0 +1,70 @@
+/*********************************************************************
+ * Software License Agreement
+ *
+ *  Copyright (c) 2019, PickNik Consulting
+ *  All rights reserved.
+ *
+ *********************************************************************/
+
+/* Author: Omid Heidari, Dave Coleman
```
davetcoleman: how am i an author? you can remove me. the template you used doesn't need attribution

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+class TrajectoryTest : public ::testing::Test
+{
+public:
+  TrajectoryTest()
+  {
+  }
+
+protected:
+protected:
```
davetcoleman: remove double

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+  trajopt_interface::ProblemInfo pci(planning_scene, "panda_arm");
+  pci.basic_info.n_steps = 30;
+  std::vector<double> joint_values = { 0.4, 0.3, 0.5, -0.55, 0.88, 1.0, -0.075 };
+  trajopt::TrajArray init_traj;
+
+  generateInitialTrajectory(pci, joint_values, init_traj);
+
+  EXPECT_EQ(init_traj.cols(), joint_values.size());
+  EXPECT_EQ(init_traj.rows(), pci.basic_info.n_steps);
+}
```
davetcoleman: please add significantly more tests

moveit_planners/trajopt/src/problem_description.cpp
ommmid: No, it is not temporary

moveit_planners/trajopt/test/trajectory_test.cpp
ommmid: I will do this when we want to merge the next PR with no tesseract dependency, just because I need more time for creating tests

moveit_planners/trajopt/src/problem_description.cpp
ommmid: `addObjectiveTerms` instead? @BryceStevenWilley ?

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -0,0 +1,452 @@
...
+  }
+
+  if (req.start_state.joint_state.name.empty())
+  {
+    ROS_ERROR_STREAM_NAMED("trajopt_planner", "Start_state name is empty");
+    res.error_code.val = moveit_msgs::MoveItErrorCodes::INVALID_ROBOT_STATE;
+    return false;
+  }
+
+  if (not planning_scene->getRobotModel()->satisfiesPositionBounds(req.start_state.joint_state.position.data()))
```
henningkayser: This check must use the `JointModelGroup`  specified in the `MotionPlanRequest` (`req.group_name`). Otherwise subgroups can't be used for planning.

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -0,0 +1,452 @@
...
+  }
+  joint_start_pos->targets = joint_start_constraints;
+  setJointPoseTermInfoParams(joint_start_pos, "start_pos");
+  problem_info.cnt_infos.push_back(joint_start_pos);
+
+  ROS_INFO(" ======================================= Velocity Constraints, hard-coded");
+  // TODO: should be defined by user, its parametes should be added to setup.yaml
+  JointVelTermInfoPtr joint_vel(new JointVelTermInfo);
+
+  joint_vel->coeffs = std::vector<double>(7, 5.0);
```
henningkayser: The number of coefficients should match the DOF of the specified joint model group.

moveit_planners/trajopt/CMakeLists.txt
```diff
@@ -7,30 +7,117 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
...
+  DIRECTORY
+    include/trajopt_interface/
+  DESTINATION
+    ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+)
+
+# Mark roslaunch files for installation
+install(
+  DIRECTORY
+    launch
```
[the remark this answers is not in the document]
ommmid: removed it. I had moved the launch file to moveit_tutorials.

moveit_planners/trajopt/src/trajopt_interface.cpp
ommmid: right, corrected it

moveit_planners/trajopt/src/trajopt_interface.cpp
ommmid: right. replaced it with the joint positions related to the joint_model_group

jonbinney:
@dmillard could you take a look at this since you've been using trajopt?

jonbinney:
@ommmid it looks like this needs a rebase since a lot of the changes are things from other PRs that are already in master, and there are merge conflicts.

ommmid:
> @ommmid it looks like this needs a rebase since a lot of the changes are things from other PRs that are already in master, and there are merge conflicts.

We are planning to separate this PR from @j-petit work (bullet collision), it is actually done. I will rebase and retarget to master.

dmillard:
@jonbinney From a brief look, this looks pretty reasonable. I'm assuming the SCO logic is ported over in another PR, which should be pretty MoveIt agnostic anyway.

As it stands, I don't see that this PR includes cost/constraint terms for collision checking. Is that still left to be ported?

Nice work, excited to have this in MoveIt!

ommmid:
> @jonbinney From a brief look, this looks pretty reasonable. I'm assuming the SCO logic is ported over in another PR, which should be pretty MoveIt agnostic anyway.
> 
> As it stands, I don't see that this PR includes cost/constraint terms for collision checking. Is that still left to be ported?
> 
> Nice work, excited to have this in MoveIt!

@dmillard thanks. Yes, sco is going to be ported in the next PR along with other core classes that trajopt needs, all of them are MoveIt agnostic. For collision checking, yes it is still in progress, I see if I can add it for the next PR if not I will do it in the near future.

moveit_planners/trajopt/include/trajopt_interface/trajopt_planning_context.h
```diff
@@ -0,0 +1,35 @@
+#ifndef TrajOpt_PLANNING_CONTEXT_H
+#define TrajOpt_PLANNING_CONTEXT_H
+
+#include <moveit/planning_interface/planning_interface.h>
+
+#include "trajopt_interface/problem_description.h"
```
henningkayser: These includes should use `< >` brackets

moveit_planners/trajopt/src/trajopt_planning_context.cpp
```diff
@@ -0,0 +1,74 @@
...
+    res.trajectory_ = res_detailed.trajectory_[0];
+    res.planning_time_ = res_detailed.processing_time_[0];
+  }
+
+  return planning_success;
+}
+
+bool TrajOptPlanningContext::terminate()
+{
+  return true;
```
henningkayser: As long as this function is not implemented it should return false and print an error.

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -0,0 +1,458 @@
...
+  setTrajOptParams(params_);
+
+  if (!planning_scene)
+  {
+    ROS_ERROR_STREAM_NAMED("trajopt_planner", "No planning scene initialized.");
+    res.error_code.val = moveit_msgs::MoveItErrorCodes::FAILURE;
+    return false;
+  }
+
+  if (req.start_state.joint_state.position.empty())
```
henningkayser: I think empty start states should be handled by using the current state in the planning scene. However, we should check if the start state is valid for the specified planning group (number of joints and joint names match).

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -0,0 +1,458 @@
...
+    res.error_code.val = moveit_msgs::MoveItErrorCodes::INVALID_ROBOT_STATE;
+    return false;
+  }
+
+  ROS_INFO(" ======================================= Extract current state information");
+  ros::WallTime start_time = ros::WallTime::now();
+  robot_model::RobotModelConstPtr robot_model = planning_scene->getRobotModel();
+  robot_state::RobotStatePtr current_state(new robot_state::RobotState(robot_model));
+  *current_state = planning_scene->getCurrentState();
+  const robot_state::JointModelGroup* joint_model_group = current_state->getJointModelGroup(req.group_name);
```
henningkayser: Please check for success of `getJointModelGroup()`

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -0,0 +1,458 @@
...
+  std::vector<std::string> joint_names = joint_model_group->getVariableNames();
```
henningkayser: I think this should use `getActiveJointModelNames()`, or are mimic joints used somewhere?

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -0,0 +1,458 @@
...
+  ROS_INFO(" ======================================= Extract start stae infromation");
+  std::unordered_map<std::string, double> all_joints;
+  trajopt::DblVec joint_start_constraints;
+
+  for (int joint_index = 0; joint_index < req.start_state.joint_state.position.size(); ++joint_index)
+  {
+    all_joints[req.start_state.joint_state.name[joint_index]] = req.start_state.joint_state.position[joint_index];
+  }
+
+  for (auto joint_name : joint_names)
```
henningkayser: For robots with mimic joints this probably fails since `all_joints` (from `req.start_state`) usually only contains active joints. I would suggest to test if names and positions of the start state have equal size and then just iterate over those. Inside the loop we could check if the joint name actually refers to an active joint.

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -0,0 +1,458 @@
...
+  {
+    ROS_INFO(" joint position from start state, name: %s, value: %f", joint_name.c_str(), all_joints[joint_name]);
+    joint_start_constraints.push_back(all_joints[joint_name]);
+  }
+
+  // check the joint limit violiation for star state
+  if (not planning_scene->getRobotModel()->satisfiesPositionBounds(joint_start_constraints.data()))
```
henningkayser: The position bounds should be checked using the joint model group or otherwise the number and order of joint values could be invalid. Could you change this to use `JointModelGroup::satisfiesPositionBounds()` and maybe verify that the order of the joint values matches the order specified in the group?

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -0,0 +1,458 @@
...
+  {
+    problem_info.init_info.data = current_joint_values_eigen;
+  }
+  else if (problem_info.init_info.type == InitInfo::GIVEN_TRAJ)
+  {
+    problem_info.init_info.data = current_joint_values_eigen.transpose().replicate(problem_info.basic_info.n_steps, 1);
+  }
+
+  ROS_INFO(" ======================================= Create Constraints");
+  if (req.goal_constraints.empty())
```
henningkayser: This would be more readable if you would move some code into separate functions. For instance this part could just go into a function`extractGoalConstraints()`.

moveit_planners/trajopt/include/trajopt_interface/kinematic_terms.h
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/**
+ * @brief Appends b to a of type T
+ */
+template <typename T>
+std::vector<T> concatVector(const std::vector<T>& a, const std::vector<T>& b)
+{
+  std::vector<T> out;
+  std::vector<int> x(a.size() + b.size());
```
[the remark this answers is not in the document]
ommmid: right, x is excessive. removed it.

moveit_planners/trajopt/include/trajopt_interface/trajopt_interface.h
```diff
@@ -0,0 +1,79 @@
...
+  /** @brief Configure everything using the param server */
+  void setTrajOptParams(sco::BasicTrustRegionSQPParameters& param);
+  void setDefaultTrajOPtParams();
+  void setProblemInfoParam(ProblemInfo& problem_info);
+  void setJointPoseTermInfoParams(JointPoseTermInfoPtr& jp, std::string name);
+
+  ros::NodeHandle nh_;  /// The ROS node handle
+  sco::BasicTrustRegionSQPParameters params_;
+  std::vector<sco::Optimizer::Callback> callbacks_;
+  TrajOptProblemPtr prob_;
```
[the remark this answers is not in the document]
ommmid: done

moveit_planners/trajopt/src/trajopt_interface.cpp
ommmid: there is a distinction between the start state from request and the current state in trajopt. We have the option to set them the same or they can be different. But in anyways, it should be checking the specified group. I will fix that.

moveit_planners/trajopt/src/trajopt_interface.cpp
ommmid: if it returns a nullptr or not?

moveit_planners/trajopt/src/trajopt_interface.cpp
ommmid: If we want to check for joints to be active in the loop why do we even want to include the mimic ones in the first place (we are ignoring them with this check)?

moveit_planners/trajopt/src/trajopt_interface.cpp
ommmid: used joint_model_group. 
joint values in joint_start_constraints correspond to the ones in the joint model group.

moveit_planners/trajopt/src/trajopt_interface.cpp
ommmid: added extractStartJointValues(req)

moveit_planners/trajopt/test/trajectory_test.cpp
davetcoleman: This is not an acceptable reason, unit tests are EXTREMELY easy to skip and never come back to (look at all of moveit).

They are easiest to write when the code is fresh in your head / not yet merged. Please add tests before merging this.

moveit_planners/trajopt/src/problem_description.cpp
davetcoleman: love it

moveit_planners/trajopt/include/trajopt_interface/kinematic_terms.h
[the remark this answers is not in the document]
davetcoleman: please make the variables better names, also

moveit_planners/trajopt/src/trajopt_interface.cpp
davetcoleman: Could this be an inline TODO for future work? To keep this PR moving?

moveit_planners/trajopt/src/trajopt_planning_context.cpp
davetcoleman: Add a TODO also

moveit_planners/trajopt/include/trajopt_interface/kinematic_terms.h
davetcoleman: Now its allowed :-)

moveit_planners/trajopt/include/trajopt_interface/trajopt_interface.h
```diff
@@ -0,0 +1,79 @@
...
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Omid Heidari */
+
+#ifndef TRAJOPT_INTERFACE_TRAJOPT_INTERFACE_H
+#define TRAJOPT_INTERFACE_TRAJOPT_INTERFACE_H
```
davetcoleman: replace with pragma once

moveit_planners/trajopt/include/trajopt_interface/trajopt_planning_context.h
```diff
@@ -0,0 +1,35 @@
+#ifndef TrajOpt_PLANNING_CONTEXT_H
+#define TrajOpt_PLANNING_CONTEXT_H
```
davetcoleman: replace with pragma once

.gitignore
```diff
@@ -52,9 +52,12 @@ qtcreator-*
 *.swp
 
 # Catkin custom files
-CATKIN_IGNORE
+# CATKIN_IGNORE
```
jonbinney: Something very meta about ignoring (commenting) the git ignoring of the catkin ignore file here.

Maybe just delete this line?

ommmid: agree, I remove it.

moveit_planners/trajopt/src/trajopt_interface.cpp
ommmid: It is already implemented, we can have that by setting the start_fixed in the yaml file. No need for TODO.

moveit_planners/trajopt/include/trajopt_interface/kinematic_terms.h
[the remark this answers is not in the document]
ommmid: done

ommmid:
> Add more tests, then looks good to merge!

Added another small unittest.

moveit_planners/trajopt/src/trajopt_interface.cpp
henningkayser: Yes, since the planning group could be wrong

moveit_planners/trajopt/test/trajectory_test.cpp
ommmid: added a unit test for goal tolerance

moveit_planners/trajopt/src/trajopt_interface.cpp
```diff
@@ -84,6 +84,9 @@ bool TrajOptInterface::solve(const planning_scene::PlanningSceneConstPtr& planni
   ROS_INFO(" ======================================= Extract current state information");
   ros::WallTime start_time = ros::WallTime::now();
   robot_model::RobotModelConstPtr robot_model = planning_scene->getRobotModel();
+  bool robot_model_ok_ = static_cast<bool>(robot_model);
```
davetcoleman: no _

moveit_planners/trajopt/src/trajopt_interface.cpp
davetcoleman: still TODO

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -37,49 +46,169 @@ class TrajectoryTest : public ::testing::Test
...
+    robot_model_ = moveit::core::loadTestingRobotModel("panda");
+    bool robot_model_ok_ = static_cast<bool>(robot_model_);
+    if (!robot_model_ok_)
+      ROS_ERROR_STREAM_NAMED("trajectory_test", "robot model is not loaded correctly");
+  }
 
+protected:
+  robot_model::RobotModelPtr robot_model_;
+  std::vector<std::string> group_joint_names_;
+  const std::string PLANNING_GROUP_ = "panda_arm";
```
davetcoleman: no _ for const

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -37,49 +46,169 @@ class TrajectoryTest : public ::testing::Test
...
-  std::vector<double> vec_b = {6, 7, 8, 9, 10};
+  std::vector<double> vec_a = { 1, 2, 3, 4, 5 };
+  std::vector<double> vec_b = { 6, 7, 8, 9, 10 };
   std::vector<double> vec_c = trajopt_interface::concatVector(vec_a, vec_b);
   EXPECT_EQ(vec_c.size(), vec_a.size() + vec_b.size());
 
-  size_t length_ab = vec_a.size() + vec_b.size();
-  for(size_t index = 0; index < length_ab ; ++index)
+  std::size_t length_ab = vec_a.size() + vec_b.size();
+  for (std::size_t index = 0; index < length_ab; ++index)
```
davetcoleman: add some comments explaning

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -37,49 +46,169 @@ class TrajectoryTest : public ::testing::Test
...
+  const std::string NODE_NAME = "trajectory_test";
+
+  const std::string PLANNING_GROUP = "panda_arm";
+  const std::string ROBOT_DESCRIPTION = "robot_description";
+
+  // Create a RobotState and JointModelGroup to keep track of the current robot pose and planning group
+  robot_state::RobotStatePtr current_state(new robot_state::RobotState(robot_model_));
+  current_state->setToDefaultValues();
+
+  const robot_state::JointModelGroup* joint_model_group = current_state->getJointModelGroup(PLANNING_GROUP);
```
davetcoleman: EXPECT_NE(joint_model_group, nullptr)

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -37,49 +46,169 @@ class TrajectoryTest : public ::testing::Test
...
+  const std::vector<std::string>& joint_names = joint_model_group->getActiveJointModelNames();
```
davetcoleman: test count of joint names EXPECT_EQ

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -37,49 +46,169 @@ class TrajectoryTest : public ::testing::Test
...
+  goal_state->update();
+  moveit_msgs::Constraints joint_goal = kinematic_constraints::constructGoalConstraints(*goal_state, joint_model_group);
+  req.goal_constraints.push_back(joint_goal);
+  req.goal_constraints[0].name = "goal_pos";
+  // Set joint tolerance
+  std::vector<moveit_msgs::JointConstraint> goal_joint_constraint = req.goal_constraints[0].joint_constraints;
+  for (std::size_t x = 0; x < goal_joint_constraint.size(); ++x)
+  {
+    ROS_INFO_STREAM_NAMED(NODE_NAME, " ======================================= joint position at goal: "
+                                         << goal_joint_constraint[x].position);
```
davetcoleman: remove debug output

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -37,49 +46,169 @@ class TrajectoryTest : public ::testing::Test
...
+  // We will now construct a loader to load a planner, by name.
+  // Note that we are using the ROS pluginlib library here.
+  boost::scoped_ptr<pluginlib::ClassLoader<planning_interface::PlannerManager>> planner_plugin_loader;
+  planning_interface::PlannerManagerPtr planner_instance;
+
+  std::string planner_plugin_name = "trajopt_interface/TrajOptPlanner";
+  node_handle_.setParam("planning_plugin", planner_plugin_name);
+
+  // Make sure the planner plugin is loaded
+  if (!node_handle_.getParam("planning_plugin", planner_plugin_name))
```
davetcoleman: EXPECT_TRUE(

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -37,49 +46,169 @@ class TrajectoryTest : public ::testing::Test
...
+                                                                          << "Available plugins: " << ss.str());
+  }
+
+  // Creat planning context
+  // ========================================================================================
+  planning_interface::PlanningContextPtr context =
+      planner_instance->getPlanningContext(planning_scene, req, res.error_code_);
+
+  context->solve(res);
+  if (res.error_code_.val != res.error_code_.SUCCESS)
```
davetcoleman: EXPECT

moveit_planners/trajopt/test/trajectory_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+    else
+    {
+      EXPECT_EQ(vec_c[index], vec_b[index - vec_a.size()]);
+    }
+  }
+}
+
+TEST_F(TrajectoryTest, goalTolerance)
+{
+  SetUp();
```
BryceStevenWilley: SetUp should be called already before this method is entered by gtest, should remove.

moveit_planners/trajopt/src/trajopt_planning_context.cpp
```diff
@@ -0,0 +1,75 @@
...
+  {
+    res.trajectory_.resize(1);
+    res.trajectory_[0] =
+        robot_trajectory::RobotTrajectoryPtr(new robot_trajectory::RobotTrajectory(robot_model_, getGroupName()));
+
+    moveit::core::RobotState start_state(robot_model_);
+    robot_state::robotStateMsgToRobotState(res_msg.trajectory_start, start_state);
+    res.trajectory_[0]->setRobotTrajectoryMsg(start_state, res_msg.trajectory[0]);
+
+    res.description_.push_back("plan");
```
BryceStevenWilley: For the detailed response, should trajopt keep around the initial trajectory as well for comparison?  If we think so, I'd be fine with a TODO in this PR.

moveit_planners/trajopt/src/trajopt_planning_context.cpp
```diff
@@ -0,0 +1,75 @@
...
+    res.trajectory_ = res_detailed.trajectory_[0];
+    res.planning_time_ = res_detailed.processing_time_[0];
+  }
+
+  return planning_success;
+}
+
+bool TrajOptPlanningContext::terminate()
+{
+  ROS_ERROR_STREAM_NAMED("trajopt_planning_context", "terminate() is not implemented");
```
BryceStevenWilley: Better error message: `"TrajOpt is not interruptable yet"`, also add a TODO.

moveit_planners/trajopt/README.md
```diff
@@ -0,0 +1 @@
+As of June 2019, this is a work in progress towards adding trajopt motion planning algorithm to MoveIt as a planner plugin.
```
BryceStevenWilley:
```suggestion
As of August 2019, this is a work in progress towards adding trajopt motion planning algorithm to MoveIt as a planner plugin.
```
:smile:

moveit_planners/trajopt/CMakeLists.txt
BryceStevenWilley: This hasn't been removed yet, and I can't find it being used in this PR still.

moveit_planners/trajopt/include/trajopt_interface/problem_description.h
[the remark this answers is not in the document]
BryceStevenWilley: Oh, so they're just not used in this file.  Not the biggest fan of that, but it's okay.

moveit_planners/trajopt/include/trajopt_interface/problem_description.h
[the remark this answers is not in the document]
BryceStevenWilley: I suppose we can keep the same design pattern, changing the name from `hatch` was about 50% of the problem that I had with it.

moveit_planners/trajopt/src/trajopt_planning_context.cpp
ommmid: I add a TODO

davetcoleman:
Merging after many rounds of reviews from multiple maintainers. This is a new feature that is currently disabled with CATKIN_IGNORE, so even though we could keep optimizing it, we need to move forward with it as an 'experimental planner' so that @ommmid can submit his code before the end of his internship Friday.

PR #1595, opened 2019-08-02
Add parameter to get_jacobian_matrix

felixvd:
### Description

I needed to expose the `reference_point_position` parameter of the `getJacobian` function in Python, so I added it as an optional parameter. Tested on a UR5, worked as expected.

Can be cherry-picked to kinetic and melodic just like #1501, where the `get_jacobian_matrix` was added that this PR modifies.

### Checklist
- [X] While waiting for someone to review your request, please consider reviewing [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers

### DIFF
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -596,7 +596,7 @@ class MoveGroupCommander(object):
         traj_out.deserialize(ser_traj_out)
         return traj_out
 
-    def get_jacobian_matrix(self, joint_values):
+    def get_jacobian_matrix(self, joint_values, reference_point_position=[0.0, 0.0, 0.0]):
         """ Get the jacobian matrix of the group as a list"""
-        return self._g.get_jacobian_matrix(joint_values)
+        return self._g.get_jacobian_matrix(joint_values, reference_point_position)
 
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -596,7 +596,7 @@ class MoveGroupCommander(object):
         traj_out.deserialize(ser_traj_out)
         return traj_out
 
-    def get_jacobian_matrix(self, joint_values, reference_point_position=[0.0, 0.0, 0.0]):
+    def get_jacobian_matrix(self, joint_values, reference_point_position=None):
         """ Get the jacobian matrix of the group as a list"""
-        return self._g.get_jacobian_matrix(joint_values, reference_point_position)
+        return self._g.get_jacobian_matrix(joint_values, [0.0, 0.0, 0.0] if reference_point_position is None else reference_point_position)
 
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -596,7 +596,7 @@ class MoveGroupCommander(object):
         traj_out.deserialize(ser_traj_out)
         return traj_out
 
-    def get_jacobian_matrix(self, joint_values, reference_point_position=None):
+    def get_jacobian_matrix(self, joint_values, reference_point=None):
         """ Get the jacobian matrix of the group as a list"""
-        return self._g.get_jacobian_matrix(joint_values, [0.0, 0.0, 0.0] if reference_point_position is None else reference_point_position)
+        return self._g.get_jacobian_matrix(joint_values, [0.0, 0.0, 0.0] if reference_point is None else reference_point)
 
--- a/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
@@ -536,14 +536,16 @@ public:
     }
   }
 
-  Eigen::MatrixXd getJacobianMatrixPython(bp::list& joint_values)
+  Eigen::MatrixXd getJacobianMatrixPython(bp::list& joint_values, bp::list& reference_point_position)
   {
-    std::vector<double> v = py_bindings_tools::doubleFromList(joint_values);
+    std::vector<double> joint_vals = py_bindings_tools::doubleFromList(joint_values);
+    std::vector<double> ref_point_pos_vector = py_bindings_tools::doubleFromList(reference_point_position);
+    Eigen::Vector3d ref_point_pos(ref_point_pos_vector.data());
     robot_state::RobotState state(getRobotModel());
     state.setToDefaultValues();
     auto group = state.getJointModelGroup(getName());
-    state.setJointGroupPositions(group, v);
-    return state.getJacobian(group);
+    state.setJointGroupPositions(group, joint_vals);
+    return state.getJacobian(group, ref_point_pos);
   }
 };
 
--- a/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
@@ -536,13 +536,20 @@ public:
     }
   }
 
-  Eigen::MatrixXd getJacobianMatrixPython(const bp::list& joint_values,
-                                          const bp::list& reference_point_position = bp::list())
+  Eigen::MatrixXd getJacobianMatrixPython1(const bp::list& joint_values)
+  {
+    const std::vector<double> joint_vals = py_bindings_tools::doubleFromList(joint_values);
+    robot_state::RobotState state(getRobotModel());
+    state.setToDefaultValues();
+    auto group = state.getJointModelGroup(getName());
+    state.setJointGroupPositions(group, joint_vals);
+    return state.getJacobian(group);
+  }
+
+  Eigen::MatrixXd getJacobianMatrixPython2(const bp::list& joint_values, const bp::list& reference_point_position)
   {
     const std::vector<double> joint_vals = py_bindings_tools::doubleFromList(joint_values);
     std::vector<double> ref_point_pos_vector = py_bindings_tools::doubleFromList(reference_point_position);
-    if (ref_point_pos_vector.size() == 0)
-      ref_point_pos_vector = { 0.0, 0.0, 0.0 };
     const Eigen::Vector3d ref_point_pos(ref_point_pos_vector.data());
 
     robot_state::RobotState state(getRobotModel());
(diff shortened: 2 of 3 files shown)


felixvd:
Travis is returning this error, even after restarting the build:

```
<testsuite errors="1" failures="0" name="unittest.suite.TestSuite" tests="3" time="5.233"><testcase classname="__main__.PythonMoveGroupTest" name="test_get_jacobian_matrix" time="0.2506"><error type="ArgumentError">Python argument types in
    MoveGroupInterface.get_jacobian_matrix(MoveGroupInterface, list)
did not match C++ signature:
    get_jacobian_matrix(moveit::planning_interface::MoveGroupInterfaceWrapper {lvalue}, boost::python::list {lvalue}, boost::python::list {lvalue})
    testMethod()
  File "/root/ros_ws/src/moveit/moveit_ros/planning_interface/test/python_move_group.py", line 73, in test_get_jacobian_matrix
    result = self.group.get_jacobian_matrix(current)
```

But when I run `catkin run_tests --no-deps --this` in `moveit/moveit_ros/planning_interface`, I don't get any errors. The code is also running on our machine without problems. Anyone have an idea what's going on?

rhaschke:
It's written in the error: The call to get_jacobian_matrix in https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/test/python_move_group.py#L73 is missing an argument.
This test was recently added in master. Probably you need to rebase your PR branch onto master to see this error locally as well.

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -596,7 +596,7 @@ def retime_trajectory(self, ref_state_in, traj_in, velocity_scaling_factor=1.0,
         traj_out.deserialize(ser_traj_out)
         return traj_out
 
-    def get_jacobian_matrix(self, joint_values):
+    def get_jacobian_matrix(self, joint_values, reference_point_position=[0.0, 0.0, 0.0]):
```
rhaschke: It's bad practice to use mutable default arguments, see https://docs.python-guide.org/writing/gotchas.
Alternatively, use `None` as the default arg but forward the zero list if `None` was given.
```suggestion
    def get_jacobian_matrix(self, joint_values, reference_point_position=None):
```

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -596,7 +596,7 @@ def retime_trajectory(self, ref_state_in, traj_in, velocity_scaling_factor=1.0,
...
         """ Get the jacobian matrix of the group as a list"""
-        return self._g.get_jacobian_matrix(joint_values)
+        return self._g.get_jacobian_matrix(joint_values, reference_point_position)
```
rhaschke:
```suggestion
        return self._g.get_jacobian_matrix(joint_values, [0,0,0] if reference_point_position is None else reference_point_position)
```

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
```diff
@@ -536,14 +536,16 @@ class MoveGroupInterfaceWrapper : protected py_bindings_tools::ROScppInitializer
     }
   }
 
-  Eigen::MatrixXd getJacobianMatrixPython(bp::list& joint_values)
+  Eigen::MatrixXd getJacobianMatrixPython(bp::list& joint_values, bp::list& reference_point_position)
```
rhaschke: To correctly handle the issue in https://github.com/ros-planning/moveit/pull/1595#issuecomment-519434052, reference_point_position needs to be optional here as well.
That's because this method can be called directly as well.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
```diff
@@ -536,14 +536,16 @@ class MoveGroupInterfaceWrapper : protected py_bindings_tools::ROScppInitializer
...
   {
-    std::vector<double> v = py_bindings_tools::doubleFromList(joint_values);
+    std::vector<double> joint_vals = py_bindings_tools::doubleFromList(joint_values);
+    std::vector<double> ref_point_pos_vector = py_bindings_tools::doubleFromList(reference_point_position);
+    Eigen::Vector3d ref_point_pos(ref_point_pos_vector.data());
```
rhaschke: Use a `const Eigen::Map<Eigen::Vector3d>` wrapper here to avoid copying the data.

moveit_commander/src/moveit_commander/move_group.py
felixvd: Nice, thanks!

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
felixvd: Can you explain more? `getJacobian` takes an `Eigen::Vector3d` as input, and I'm not sure how to avoid copying the data via `py_bindings_tools::doubleFromList`

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
felixvd: I don't know how to initialize a list to `[0.0, 0.0, 0.0]` for a default argument, but the new commit should work too. Thanks for taking another look.

felixvd:
> It's written in the error: The call to get_jacobian_matrix in https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/test/python_move_group.py#L73 is missing an argument.
> This test was recently added in master. Probably you need to rebase your PR branch onto master to see this error locally as well.

I saw the error message, but I don't understand it. The second argument is optional, so why would the call with one argument give an error?

I rebased on master and added your changes, but the tests succeeded locally for me at every point. We'll see if this commit changes anything.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
rhaschke: You copy twice: First `doubleFromList` is used to create a `std::vector`. Then, this std::vector is copied over to an `Eigen::Vector3d`. This second copy can be avoided using an Eigen::Map, which is essentially a wrapper around already allocated memory.

rhaschke:
> The second argument is optional, so why would the call with one argument give an error?

We do have a two-layer wrapper here: There is the C++ level wrapper, and on top of this the thin python-level wrapper, which you adapted. Only the latter has the optional argument so far, not the former one. That's why my suggestion in https://github.com/ros-planning/moveit/pull/1595#discussion_r311939262.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
```diff
@@ -536,11 +536,15 @@ class MoveGroupInterfaceWrapper : protected py_bindings_tools::ROScppInitializer
     }
   }
 
-  Eigen::MatrixXd getJacobianMatrixPython(bp::list& joint_values, bp::list& reference_point_position)
+  Eigen::MatrixXd getJacobianMatrixPython(const bp::list& joint_values,
+                                          const bp::list& reference_point_position = bp::list())
```
rhaschke: I suggest to use the default bp::object() here as well (if possible), which would be equivalent to None.
By default, boost::python only exposes the full signature. To also expose other variants (keeping the defaults), you need to perform [overload resolution](https://www.boost.org/doc/libs/1_66_0/libs/python/doc/html/tutorial/tutorial/functions.html#tutorial.functions.overloading).

felixvd:
I pushed a version with overload resolution because I didn't know how to check if `bp::object()` is `None`, and an additional test to confirm that the new parameter works.

> We do have a two-layer wrapper here: There is the C++ level wrapper, and on top of this the thin python-level wrapper, which you adapted. Only the latter has the optional argument so far, not the former one. That's why my suggestion in [#1595 (comment)](https://github.com/ros-planning/moveit/pull/1595#discussion_r311939262).

I understand that it's two layers, and when I first wrote this I put the default argument into both layers, but then I figured that the C++ level wrapper would not be called by anything but the Python wrapper. I also don't see where the test would be causing this error, and I don't understand why the test does not fail locally. Oh well. If this one passes I am happy.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
felixvd: I missed this comment and am not at my PC anymore, but if simply replacing `Eigen::Vector3d` with `Eigen::Map<Eigen::Vector3d>` works, then I'll put it in on Monday.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
rhaschke: I implemented this.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
rhaschke: Done.

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -596,7 +596,7 @@ def retime_trajectory(self, ref_state_in, traj_in, velocity_scaling_factor=1.0,
         traj_out.deserialize(ser_traj_out)
         return traj_out
 
-    def get_jacobian_matrix(self, joint_values, reference_point_position=None):
+    def get_jacobian_matrix(self, joint_values, reference_point=None):
```
felixvd: For the record, I took the parameter name from the original function.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
```diff
@@ -536,30 +536,28 @@ class MoveGroupInterfaceWrapper : protected py_bindings_tools::ROScppInitializer
...
+    if (ref.size() != 3)
+      throw std::invalid_argument("reference point needs to have 3 elements, got " + std::to_string(ref.size()));
 
     robot_state::RobotState state(getRobotModel());
     state.setToDefaultValues();
     auto group = state.getJointModelGroup(getName());
-    state.setJointGroupPositions(group, joint_vals);
-    return state.getJacobian(group, ref_point_pos);
+    state.setJointGroupPositions(group, v);
+    return state.getJacobian(group, Eigen::Map<Eigen::Vector3d>(&ref[0]));
```
felixvd: Thanks!

Issue #1567, opened 2019-07-16
Ikfast template no longer allows robot prefix


jschleicher:
The ikfast plugin template now hard-codes the base and tip link inside its cpp:

https://github.com/ros-planning/moveit/blob/b853c6bc48c41af7475db8cbb4d11078bb8725eb/moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp#L166-L167

The version before #1014 allowed to use a single plugin for multirobot (e.g. left_* and right_*). The new version is no longer compatible to such setups.

As described on the [ROS-Industrial wiki page](http://wiki.ros.org/Industrial/Tutorials/WorkingWithRosIndustrialRobotSupportPackages) a prefix allows to instantiate a xacro macro twice for multi-robot setups. It should be possible, to pass this prefix to the kinematics solver as well.

jschleicher:
Fixed in #1599.

PR #1599, opened 2019-08-05
optional prefix for link names in ikfast plugin

jschleicher:
If you pass a `linkprefix` parameter in your `kinematics.yaml`
this string is prepended to the base and tip link. It allows
multi-robot setups (e.g. dual-arm) and still instantiate the same
solver for both manipulators.

Implements #1567

### Checklist
- [X] Suggesting to cherry-pick this to melodic (since it is backwards-compatible).


moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -409,18 +412,30 @@ bool IKFastKinematicsPlugin::initialize(const moveit::core::RobotModel& robot_mo
   }
 
   storeValues(robot_model, group_name, base_frame, tip_frames, search_discretization);
+  if (!lookupParam("linkprefix", linkframeprefix_, std::string("")))
+  {
+    ROS_INFO_NAMED(name_, "linkprefix parameter not set. Assuming empty prefix.");
```
v4hn: If the `else` case here is a `DEBUG` log, I would expect this to be the same.

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -409,18 +412,30 @@ bool IKFastKinematicsPlugin::initialize(const moveit::core::RobotModel& robot_mo
   }
 
   storeValues(robot_model, group_name, base_frame, tip_frames, search_discretization);
+  if (!lookupParam("linkprefix", linkframeprefix_, std::string("")))
```
v4hn: `link_prefix`?

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -166,6 +166,9 @@ class IKFastKinematicsPlugin : public kinematics::KinematicsBase
   const std::string IKFAST_TIP_FRAME_ = "_EEF_LINK_";
   const std::string IKFAST_BASE_FRAME_ = "_BASE_LINK_";
 
+  // prefix added to tip- and baseframe to allow different namespaces or multi-robot setups
+  std::string linkframeprefix_;
```
jrgnicho: I would prefer if this member name was [under_scored](http://wiki.ros.org/CppStyleGuide#Naming)

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
jrgnicho: This should be a ROS_WARN

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -409,18 +412,30 @@ bool IKFastKinematicsPlugin::initialize(const moveit::core::RobotModel& robot_mo
...
+  }
 
   // This IKFast solution was generated with IKFAST_TIP_FRAME_ and IKFAST_BASE_FRAME_.
   // It is often the case that fixed joints are added to these links to model things like
   // a robot mounted on a table or a robot with an end effector attached to the last link.
   // To support these use cases, we store the transform from the IKFAST_BASE_FRAME_ to the
   // base_frame_ and IKFAST_TIP_FRAME_ the tip_frame_ and transform to the input pose accordingly
-  if (!computeRelativeTransform(tip_frames_[0], IKFAST_TIP_FRAME_, group_tip_to_chain_tip_, tip_transform_required_) ||
-      !computeRelativeTransform(IKFAST_BASE_FRAME_, base_frame_, chain_base_to_group_base_, base_transform_required_))
+  if (!computeRelativeTransform(tip_frames_[0], linkframeprefix_ + IKFAST_TIP_FRAME_, group_tip_to_chain_tip_,
```
jrgnicho: Will the function  `computeRelativeTransform` print an error when the combined link prefix and hardcoded IKFAST frame names do not exists?  If not then we should check if the resulting link name is valid

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -409,18 +412,30 @@ bool IKFastKinematicsPlugin::initialize(const moveit::core::RobotModel& robot_mo
...
+                                base_transform_required_))
   {
-    if (!computeRelativeTransform(tip_frames_[0], IKFAST_TIP_FRAME_, group_tip_to_chain_tip_, tip_transform_required_))
+    if (!computeRelativeTransform(tip_frames_[0], linkframeprefix_ + IKFAST_TIP_FRAME_, group_tip_to_chain_tip_,
+                                  tip_transform_required_))
       ROS_ERROR_NAMED(name_, "Failed to compute transform between IKFAST_TIP_FRAME_ and tip_frames_[0]");
-    if (!computeRelativeTransform(IKFAST_BASE_FRAME_, base_frame_, chain_base_to_group_base_, base_transform_required_))
+    if (!computeRelativeTransform(linkframeprefix_ + IKFAST_BASE_FRAME_, base_frame_, chain_base_to_group_base_,
+                                  base_transform_required_))
       ROS_ERROR_NAMED(name_, "Failed to compute transform between base_frame_ and IKFAST_BASE_FRAME_");
```
jrgnicho: This error message should print the full link name = link_prefix_ + IKFAST_FRAME

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
v4hn: I would not expect *any* output here, although an info could be justified.
It's perfectly valid to run the plugin without a prefix. Why change that?

v4hn: I do understand your motivation to have a warning, but as a consequence this would add new warnings to all setups that use the plugins (and regenerate them).
If you think the additional warning (before the fatal error later on you will surely get when it tries to access nonexisting joints) is worth this breakage, I do not mind too much.
Personally, I would prefer to reduce output for current users over changing interfaces.

v4hn:
@jschleicher Could you please address the feedback?
After all this is a rather straight-forward change to merge, in theory...

jschleicher:
@v4hn Thanks for the feedback. Yep, I'm just back from vacation, sorry for the delay.

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
jschleicher: For single-arm-setups, it's valid to leave the prefix empty. Thus it shouldn't be a warning. I'm increasing verbosity in the "computeRelativeTransform" part to give good Errors, if the expected frame is missing.

jschleicher: For conistiency making both ROS_INFO, see below https://github.com/ros-planning/moveit/pull/1599#discussion_r317199820

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
jschleicher: absolutely. :+1:

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
jschleicher: inside getLinkModel() the link names are output already, but for easier debugging which-is-which, I'll add `hasLink()` checks beforehand for all of them.

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
jschleicher: computeRelativeTransform printed a duplicated warning with the frame name; `getLinkModel()` spits an error as well. But to draw the user's attention to the `link_prefix` parameter I've added explicit checks with appropriate error messages.

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
v4hn:
```suggestion
    ROS_INFO_NAMED(name_, "link_prefix parameter not set. Assuming empty prefix.");
```

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -377,12 +380,8 @@ bool IKFastKinematicsPlugin::computeRelativeTransform(const std::string& from, c
   robot_state.reset(new RobotState(robot_model_));
   robot_state->setToDefaultValues();
 
-  auto* from_link = robot_state->getLinkModel(from);
+  auto* from_link = robot_state->getLinkModel(from);  // prints ROS_ERRORS for inexistent frames
```
mlautman:
```suggestion
  auto* from_link = robot_state->getLinkModel(from);  // prints ROS_ERRORS for non-existent frames
```

mlautman: nit

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
```diff
@@ -409,19 +408,39 @@ bool IKFastKinematicsPlugin::initialize(const moveit::core::RobotModel& robot_mo
...
+  if (!lookupParam("link_prefix", link_prefix_, std::string("")))
+  {
+    ROS_INFO_NAMED(name_, "link_prefix parameter not set. Assuming empty prefix.");
+  }
+  else
+  {
+    ROS_INFO_STREAM_NAMED(name_, "using link_prefix " << link_prefix_);
+  }
+
+  // verbose error output. subsequent checks in computeRelativeTransform return false then
```
rhaschke: If you add these error messages here, the ones from `computeRelativeTransform()` should be suppressed.
I have implemented this in https://github.com/ubi-agni/moveit/tree/ikfast-prefix

gavanderhoorn:
> @jschleicher, I have implemented all my suggestions already in https://github.com/ubi-agni/moveit/tree/ikfast-prefix. Unfortunately, I was neither able to push those changes to the `PilzDE` fork directly, nor to create a PR against the fork. I'm not sure, how this fork was setup. So, simply pull from above fork+branch.

@rhaschke: this should let you create a PR: https://github.com/PilzDE/moveit/compare/ikfast-prefix...ubi-agni:ikfast-prefix

rhaschke:
> @rhaschke: this should let you create a PR: [PilzDE/moveit@ikfast-prefix...ubi-agni:ikfast-prefix](https://github.com/PilzDE/moveit/compare/ikfast-prefix...ubi-agni:ikfast-prefix)

@gavanderhoorn, thanks for providing this link. When I tried to create the PR via github's web interface, PilzDE's fork was not offered as a base. I created the PR: https://github.com/PilzDE/moveit/pull/1

gavanderhoorn:
> @gavanderhoorn, thanks for providing this link. When I tried to create the PR via github's web interface, PilzDE's fork was not offered as a base. 

Yes, I noticed that too.

Not sure what is going on there.

moveit_kinematics/ikfast_kinematics_plugin/templates/ikfast61_moveit_plugin_template.cpp
jschleicher: Thanks! I've merged your branch in here.

PR #1603, opened 2019-08-07
Do not wait for command msg to start spinning in jog_arm

PaulRuvu:
### Description
- Only wait for the message we try to smooth with the low-pass filter (JointState). This also allows the initial jog command to be of the JointJog type.
- Do not assume the command to be cartesian when the zero_joint_cmd_flag is set, could be a JointJog command with zero velocity
- valid_nonzero_command flag is set when either zero_cartesian_cmd_flag **or** zero_joint_cmd_flag is not set
- Remove unused publish_delay parameter

### DIFF
--- a/moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
+++ b/moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
@@ -66,10 +66,6 @@ JogCalcs::JogCalcs(const JogArmParameters& parameters, JogArmShared& shared_vari
   ros::topic::waitForMessage<sensor_msgs::JointState>(parameters_.joint_topic);
   ROS_INFO_NAMED(LOGNAME, "Received first joint msg.");
 
-  ROS_INFO_NAMED(LOGNAME, "Waiting for first command msg.");
-  ros::topic::waitForMessage<geometry_msgs::TwistStamped>(parameters_.cartesian_command_in_topic);
-  ROS_INFO_NAMED(LOGNAME, "Received first command msg.");
-
   resetVelocityFilters();
 
   jt_state_.name = joint_model_group_->getVariableNames();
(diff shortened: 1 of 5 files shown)


moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
```diff
@@ -635,7 +631,9 @@ void JogCalcs::halt(trajectory_msgs::JointTrajectory& jt_traj)
   {
     // For position-controlled robots, can reset the joints to a known, good state
     if (parameters_.publish_joint_positions)
+      {
```
PaperFanz: Indentation looks wrong here and the next couple lines. See the moveit contributors guide about clang-format -- it will fix this for you.

https://moveit.ros.org//documentation/contributing/code/

AndyZe: Indentation looks wrong here and the next couple lines. See the moveit contributors guide about clang-format -- it will fix this for you.

https://moveit.ros.org//documentation/contributing/code/

moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
```diff
@@ -181,7 +181,7 @@ JogCalcs::JogCalcs(const JogArmParameters& parameters, JogArmShared& shared_vari
       zero_joint_cmd_flag = true;
     }
 
-    bool valid_nonzero_command = !zero_cartesian_cmd_flag && !zero_joint_cmd_flag;
+    bool valid_nonzero_command = !zero_cartesian_cmd_flag || !zero_joint_cmd_flag;
```
AndyZe: This looks like an important bug to have caught. I appreciate that.

moveit_experimental/jog_arm/src/jog_arm/jog_calcs.cpp
PaulRuvu: My bad, fixed it

PaulRuvu:
> The original reason to wait for a message before spinning was to reduce CPU load. With this PR, my jog_arm CPU load is about 16% as soon as the node launches. I would like to keep it under 1% until messages start coming in.
> 
> I often load the jogger and have it running in the background until I switch to jog mode.
> 
> If you can keep the other changes you've made AND wait to spin until all necessary messages arrive, I would definitely approve it.
> 
> Besides that, I tested functionality and it seems fine.

I can understand the desire to not spin when the jogging functionality is not active. But in that case it should also exit the spinning loop when input commands get stale. 
Furthermore it should not only be triggered by the configured _cartesian_command_in_topic_ but also by the _joint_command_in_topic_

rhaschke:
@PaulRuvu, please rebase to resolve conflicts.

PaulRuvu:
> @PaulRuvu, please rebase to resolve conflicts.

Done.

PR #1607, opened 2019-08-08
[Windows][master] Use CMAKE_CXX_STANDARD to enforce c++14 for portability

seanyen:
* Use `CMAKE_CXX_STANDARD` to replace native compiler flags for `c++14` compiler feature enforcement.
* Add `set(CMAKE_CXX_EXTENSIONS OFF)` to follow the same convention used in [`moveit_tutorials`](https://github.com/ros-planning/moveit_tutorials/pull/357).

### DIFF
--- a/moveit/CMakeLists.txt
+++ b/moveit/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.1.3)
 project(moveit)
 find_package(catkin REQUIRED)
 catkin_metapackage()
--- a/moveit_commander/CMakeLists.txt
+++ b/moveit_commander/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.1.3)
 project(moveit_commander)
 
 find_package(catkin REQUIRED)
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -1,7 +1,8 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.1.3)
 project(moveit_core)
 
-add_compile_options(-std=c++14)
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 # Warnings
 add_compile_options(-Wall -Wextra
--- a/moveit_experimental/CMakeLists.txt
+++ b/moveit_experimental/CMakeLists.txt
@@ -1,7 +1,8 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.1.3)
 project(moveit_experimental)
 
-add_definitions(-std=c++14)
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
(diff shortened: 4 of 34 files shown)


bmagyar:
If we update the minimum required cmake version and set this flag, I think it should be set throughout the repo.

I've found a few occurences where it's outdated:

```
moveit_planners/moveit_planners/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.8.3)

moveit_planners/sbpl/core/sbpl_interface/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.4.6)

moveit_planners/sbpl/ros/sbpl_interface_ros/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.4.6)

moveit_runtime/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.8.3)

moveit_commander/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.8.3)

moveit/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.8.3)

moveit_plugins/moveit_plugins/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.8.3)

moveit_experimental/collision_distance_field_ros/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.4.6)

moveit_experimental/kinematics_cache/v1/kinematics_cache/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.4.6)

moveit_experimental/kinematics_cache/v1/kinematics_cache_ros/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.4.6)

moveit_setup_assistant/templates/moveit_config_pkg_template/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.8.3)

moveit_ros/moveit_ros/CMakeLists.txt
1:cmake_minimum_required(VERSION 2.8.3)
```

jonbinney:
What's the highest version of cmake available on 16.04? People still build master on indigo, right?

seanyen:
> What's the highest version of cmake available on 16.04? People still build master on indigo, right?

That's a good question. To everyone in review, what's the target platforms to support for `master` branch? Looking into [`REP 3 - Target Platforms`](http://www.ros.org/reps/rep-0003.html), it seems ROS distro before `lunar` is not compatible with this `CMAKE_CXX_STANDARD` new feature.

BryceStevenWilley:
We had a relevant conversation about dropping master support for kinetic for different reasons here: https://github.com/ros-planning/moveit/issues/1570. 

I'd also argue that if you are compiling master from source on kinetic still, you can probably install a newer version of cmake yourself.

IanTheEngineer:
> What's the highest version of cmake available on 16.04? People still build master on indigo, right?

`indigo` (and ubuntu 14.04 for that matter) are [EoL at this point](https://discourse.ros.org/t/indigo-igloo-officially-eol/9211). For that matter, `lunar` has been EoL for a while. I believe `kinetic` and 16.04 is where MoveIt is currently drawing the line for `master` - other maintainers, please correct me if I'm wrong here.

seanyen:
> If we update the minimum required cmake version and set this flag, I think it should be set throughout the repo.

I pushed a new commit to update the rest `cmake_minimum_required` usage.

rhaschke:
The target platforms for master are ROS Kinetic and Melodic. Ubuntu 16.04 uses cmake 3.5, which supports the new flag. So, all is fine.

rhaschke:
Hm. Looks like clang++ throws a lot of warnings now that we use `CMAKE_CXX_STANDARD=14`. However, I'm not sure how this differs from `-std=c++14`. It shouldn't, but it actually does for clang!?!

seanyen:
> Hm. Looks like [clang++ throws a lot of warnings](https://travis-ci.org/ros-planning/moveit/jobs/569222507) now that we use `CMAKE_CXX_STANDARD=14`. However, I'm not sure how this differs from `-std=c++14`. It shouldn't, but it actually does for clang!?!

It seems that the same warnings also manifested on other pull requests for `master` (for example, https://travis-ci.org/ros-planning/moveit/jobs/569489967). Maybe it is not related to `CMAKE_CXX_STANDARD` change?

rhaschke:
You are right. Seems to be related to recent changes in `geometric_shapes`. Looking into this now.

v4hn:
Thank you so much for the follow-up patch @seanyen !

PR #1610, opened 2019-08-09
fix clang-tidy warnings

rhaschke:
clang-tidy issues [several warnings on Travis](https://travis-ci.org/ros-planning/moveit/jobs/569519848#L1360), e.g.:

```
/root/ros_ws/src/moveit/moveit_core/robot_state/src/robot_state.cpp:1052:8: warning: the type of the loop variable 'body' is different from the one returned by the iterator and generates an implicit conversion; you can either change the type to the matching one ('const std::pair<const std::__cxx11::basic_string<char>, moveit::core::AttachedBody *> &' but 'const auto&' is always a valid option) or remove the reference to make it explicit that you are creating a new value [performance-implicit-conversion-in-loop]
for (const std::pair<std::string, AttachedBody*>& body : attached_body_map_)
    ^
```

Although gcc and clang themselves don't complain, I would like to clean those warnings as well.
Looks like clang-tidy cannot match `std::__cxx11::basic_string<char>` to `std::string`.
Hence, I propose to replace the explicit type with `auto`.

### DIFF
--- a/moveit_core/collision_detection/src/world.cpp
+++ b/moveit_core/collision_detection/src/world.cpp
@@ -146,7 +146,7 @@ bool World::knowsTransform(const std::string& name) const
     return it->second->shape_poses_.size() == 1;
   else  // Then objects' subframes
   {
-    for (const std::pair<std::string, ObjectPtr>& object : objects_)
+    for (const auto& object : objects_)
     {
       // if "object name/" matches start of object_id, we found the matching object
       if (boost::starts_with(name, object.first) && name[object.first.length()] == '/')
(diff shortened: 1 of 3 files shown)


v4hn:
Is there actually an implicit conversion taking place in this case(and clang-tidy is right) or is `std::string` a syntactic alias for `std::__cxx11::basic_string<char>`?

I am not a fan of either way to get rid of the warning.
(1) The expanded type is unnecessarily verbose and should not be used directly and
(2) auto hides the tuple types, which I consider essential for future readers of the code (type inference in some IDEs is no excuse for using `auto` in this way).

rhaschke:
> I am not a fan of either way to get rid of the warning.

I was afraid of this ;-)
I'm not sure, an actual conversion is taking place. Looks like it's an issue of clang-tidy only. I just had these changes lying around and pushed them as a PR now... Feel free to close as won't fix.

v4hn:
> Looks like it's an issue of clang-tidy only

So where's your patch to clang-tidy? :laughing:

Thanks for pushing your patch all the same.

Closing as not-our-bug, unless someone proofs otherwise.

PR #1611, opened 2019-08-09
psi: wait for services

v4hn:
Turns out, the `PlanningSceneInterface` never waited for its services to be available.
However, all but two functions of the class utilize this class (and fail if the service is not setup yet).

As the original goal of the `apply*` function family was to get rid of random `sleep` statements in the code base, I consider it important that there is no need to wait when using the class methods.

### DIFF
--- a/moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp
+++ b/moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp
@@ -53,11 +53,14 @@ public:
   explicit PlanningSceneInterfaceImpl(const std::string& ns = "")
   {
     node_handle_ = ros::NodeHandle(ns);
+    planning_scene_diff_publisher_ = node_handle_.advertise<moveit_msgs::PlanningScene>("planning_scene", 1);
     planning_scene_service_ =
         node_handle_.serviceClient<moveit_msgs::GetPlanningScene>(move_group::GET_PLANNING_SCENE_SERVICE_NAME);
     apply_planning_scene_service_ =
         node_handle_.serviceClient<moveit_msgs::ApplyPlanningScene>(move_group::APPLY_PLANNING_SCENE_SERVICE_NAME);
-    planning_scene_diff_publisher_ = node_handle_.advertise<moveit_msgs::PlanningScene>("planning_scene", 1);
+
+    waitForService(planning_scene_service_);
+    waitForService(apply_planning_scene_service_);
   }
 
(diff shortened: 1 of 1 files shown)


rhaschke:
Hi @v4hn,
this change introduced a regression: If there is no move_group (with a PlanningScene service) available, rviz' MotionPlanning plugin, instantiating a PSI, cannot correctly initialize and rviz doesn't even open the main window.
In rviz, the PlanningScene service client is only needed for the manipulation panel: retrieving objects in ROI and picking an object. The key goal of this PR was to get rid of nasty sleeps to ensure that the PSI is really usable.
I see the following options to resolve the issue:
1. Provide an explicit function PSI::waitForServices(timeout) to replace those sleeps
2. Pass a timeout parameter from the constructor(s) all the way down through the interfaces: 
python -> PSI -> PSIimpl -> waitForService(timeout)

Actually, I'm leaning towards a combination of these. What do you think?

v4hn:
Conceptually, I believe anyone who tries to instantiate a PSI without running the appropriate services of a `move_group` does something wrong.
The pull-request here does already output appropriate warnings and I think the code can be kept as-is.

On the other hand, my personal opinion of `motion_planning_frame_manipulation` is very negative and I would be happy to throw away the whole tab until someone rewrites the whole of `motion_planning_frame_manipulation.cpp`.
Incidentally, this would get rid of the PSI instance in the display...

Alternatively, as a less drastic measure, I propose to 
(1) initialize the PSI lazy if needed
and (2) create a `static bool PSI::areServicesAvailable()` that can check all mandatory ROS interfaces for existence independent of an instance.

That way the tab can check and will never try to create an instance of the class if it's not available.
If you agree I can write (either) patch.

rhaschke:
I agree on your opinion about the manipulation tab... Removing (or disabling) the corresponding code would also clean up the messy panel...
However, this might be a very drastic method indeed. The other two options, I don't really like:
(2) might cause a race condition where you positively check for existence of the services, but then fail to create the PSI (hanging forever), because meanwhile a service died.
(1) Same issue: you might lock up rviz just by triggering the lazy initialization.

From my point of view, the only option is to provide an optional timeout parameter to the PSI constructor. For the python interface (and maybe also C++), this might be infinite. However, rviz could choose to limit the waiting time...
However, I'm also questioning the whole construct now: Usually (for non-persistent connections),
the service is contacted for each service call, which means that it can handle a crash and restart of the service. So why wait at all in the constructor? Shouldn't we have a timeout for each call instead?

PR #1621, opened 2019-08-14
pure virtual `planning_request_adapter::initialize()`

v4hn:
The method was introduced in #1530, together with [a TODO](https://github.com/ros-planning/moveit/pull/1530/files#diff-9247374d69dc34d55ae1a420105cd85dR68) to make it pure virtual "in O-turtle".

[A Windows-compatibility pr](https://github.com/ros-planning/moveit/pull/1583#discussion_r308193285) noticed that the current functionality is not available in Windows, as it uses `cxxabi.h`.
I would propose to either go for the pure virtual method directly or just implement an empty default implementation (apparently this was dismissed in #1530).

### DIFF
--- a/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
+++ b/moveit_core/planning_request_adapter/include/moveit/planning_request_adapter/planning_request_adapter.h
@@ -40,9 +40,6 @@
 #include <moveit/planning_interface/planning_interface.h>
 #include <moveit/planning_scene/planning_scene.h>
 #include <boost/function.hpp>
-#if !defined(_MSC_VER)
-#include <cxxabi.h>
-#endif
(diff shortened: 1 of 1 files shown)


rhaschke:
We dismissed this approach in #1530, because this will leave existing plugins in a broken stage.
If we are willing to go this route now, I'm fine with it.

v4hn:
> We dismissed this approach in #1530, because this will leave existing plugins in a broken stage.

Well, it is a breaking API change, yes. But I believe we agreed to allow for them in `master`.
The situation will not be much different for the `master` branch once we split of `noetic-devel` (i.e. for O-turtle).
Also, the current code breaks API just as well... only at runtime, but that's worse here in my opinion.

As I said, personally I'm open to have an empty default implementation too.

PR #1624, opened 2019-08-15
(fix) moveit_commander python planning_scene_interface for attached obj

norips:
### Description

Moveit commander python planning_scene_interface crashed when we remove attached object, this was caused by the request function. Request was adding AttachedObjectCollision to the world collision object instead of robot state collision object.

### DIFF
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -309,8 +309,8 @@ class PlanningSceneInterface(object):
     def __make_planning_scene_diff_req(collision_object, attach=False):
         scene = PlanningScene()
         scene.is_diff = True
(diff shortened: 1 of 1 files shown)


v4hn:
We see quite a lot of these errors lately on CI:
```
Updating system packages

apt-get -qq update

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
```

I have no idea where this comes from at the moment.

@norips Thank you for the contribution! I approve this request.

I added a small cleanup that sets `robot_state.is_diff = True` unconditionally.
Otherwise an empty list of attached objects would imply all attached objects in the scene should be deleted.

PR #1627, opened 2019-08-16
moveit::core::isEmpty

v4hn:
Builds on top of #1625 (although it could be isolated) and only consists of one commit.
I will rebase here once #1625 is merged.

[As pointed out by @davetcoleman](https://github.com/ros-planning/moveit/pull/1625#issuecomment-521980815) it seems weird to keep simple test functions like `isEmpty` within the full `PlanningScene` class.

This patch proposes to collect them in a new header in `moveit/utils/message_checks.h`.
Please review.


v4hn:
I rebased the request.
Merging with @rhaschke's approval.

PR #1634, opened 2019-08-21
[windows][master] Use ros::Duration::sleep for portability.

seanyen:
In some code paths, it uses `POSIX` [`sleep`](http://pubs.opengroup.org/onlinepubs/009695399/functions/sleep.html). However, for `non-POSIX` systems, the `sleep` may not exist. This change is to use the abstract `ros::Duration::sleep` for portability.


v4hn:
Merging. This is a trivial cleanup.

Actually, a number of these could be rewritten in a way to avoid the sleep, but it's mostly minor demo code where timing does not really matter.

PR #1637, opened 2019-08-21
[windows][master] Use `.pyd` as the output suffix for Python module on Windows.

seanyen:
The [`.pyd`](https://docs.python.org/3/faq/windows.html#is-a-pyd-file-the-same-as-a-dll) is the file extension for Python extension modules on Windows. This change is to fix the file extension on Windows.


v4hn:
> The macro should be defined by ros_core and thus would be automatically included by all packages that depend on ros_core.

Sounds like a good idea, I'm not sure you will find someone to review such an addition for `ros_comm` for ROS1, but at least for ROS2 this seems like a great idea.

I would propose to merge this request nonetheless until such a feature becomes available somewhere.

rhaschke:
Sorry, I meant `moveit_core` instead of `ros_core`. I think, for us, it will be sufficient to have such a macro in MoveIt. I agree with you, @v4hn, that probably nobody will agree to augment ros_comm or even catkin with such a cmake macro.

v4hn:
Merging with two approvals.

@rhaschke I do not really like the idea of adding this macro only within MoveIt.
Custom project macros always make it much harder to understand the build structure of big code bases.
Until now, MoveIt did a great job at sticking to "standard" cmake/catkin.

PR #1638, opened 2019-08-21
Fix clang-tidy for unified collision environment

j-petit:
### Description
Travis fails due to `performance-unnecessary-value-param` which has been introduced through the new unified environment in #1584. This fixes the issue.

### Checklist


BryceStevenWilley:
Merging as this fixes a failing build.

PR #1640, opened 2019-08-22
Handle the updated plan() function of MoveGroupCommander

JafarAbdi:
### Description

After this [PR](https://github.com/ros-planning/moveit/pull/790) the MoveGroupCommander.plan() now returns [success, trajectory, planning_time, error_code] which leads to a run-time error when executing this [line](https://github.com/ros-planning/moveit/blob/master/moveit_commander/src/moveit_commander/interpreter.py#L349)

### Checklist

### DIFF
--- a/moveit_commander/src/moveit_commander/interpreter.py
+++ b/moveit_commander/src/moveit_commander/interpreter.py
@@ -336,11 +336,11 @@ class MoveGroupCommandInterpreter(object):
                 if clist[1] == "rand" or clist[1] == "random":
                     vals = g.get_random_joint_values()
                     g.set_joint_value_target(vals)
-                    self._last_plan = g.plan()
+                    self._last_plan = g.plan()[1] # The trajectory msg
                 else:
                     try:
                         g.set_named_target(clist[1])
-                        self._last_plan = g.plan()
+                        self._last_plan = g.plan()[1] # The trajectory msg
                     except MoveItCommanderException as e:
                         return (MoveGroupInfoLevel.WARN, "Planning to " + clist[1] + ": " + str(e))
                     except:


moveit_commander/src/moveit_commander/interpreter.py
```diff
@@ -336,11 +336,11 @@ def execute_group_command(self, g, cmdorig):
                 if clist[1] == "rand" or clist[1] == "random":
                     vals = g.get_random_joint_values()
                     g.set_joint_value_target(vals)
-                    self._last_plan = g.plan()
+                    self._last_plan = g.plan()[1] # The trajectory msg
```
rhaschke: Probably you should also evaluate the `success` return value and set `_last_plan` to None if planning failed.

JafarAbdi: (Sorry for the late reply, I had to travel to another city) 
From what I understand from [move_group_interface.cpp](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L716) and [wrap_python_move_group.cpp](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp#L423) the `g.plan()` function will return an empty RobotTrajectory msg if the plan didn't succeed which will cause `self._last_plan` to be set to `None` by [this line](https://github.com/ros-planning/moveit/blob/master/moveit_commander/src/moveit_commander/interpreter.py#L349) if this happens -- do you think that I should evaluate `success` anyway .?

rhaschke: You are right. I had the actual C++ `plan()` in mind, which might return a preliminary, but failed trajectory.

v4hn:
urgh, this would be the ugly syntax required by the change to return a tuple...

The patch is sound though and I approve.

PR #1647, opened 2019-08-26
Migration notes for unified collision environment

j-petit:
One extra sentence for the migration notes added which states the fact that collision checks don't take a `CollisionRobot` as an argument anymore as it is implicitly contained in the new collision environment.


v4hn:
> Calling of world-robot collisions

What is that supposed to mean? If you refer to specific API, please name the API instead.

j-petit:
Added more details which functions are meant

j-petit:
Its not the FCL collision test that is failing but:
> build/moveit_ros_planning_interface/test_results/moveit_ros_planning_interface/rosunit-robot_state_update.xml: 1 tests, 0 errors, 1 failures, 0 skipped

I have no clue...

PR #1649, opened 2019-08-29
Cleanup OMPL dynamic reconfigure config

v4hn:


v4hn:
Hm, my `hub` failed with the title of the issue for some reason. Doesn't really matter though.

The patch cleans up the configuration a bit and proposes to change the default for `minimum_waypoint_count` to address #19.

PR #1652, opened 2019-08-30
depth_image_octomap_updater: correctly set properties of debug images

MatthiasNieuwenhuisen:
### Description

The step width of debug images was copied from the input depth image, but the encoding might be different resulting in a wrong step width. With this PR the step width is calculated correctly.

### DIFF
--- a/moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp
+++ b/moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp
@@ -429,7 +429,6 @@ void DepthImageOctomapUpdater::depthImageCallback(const sensor_msgs::ImageConstP
 
   if (!filtered_cloud_topic_.empty())
   {
-    static std::vector<float> filtered_data;
     sensor_msgs::Image filtered_msg;
     filtered_msg.header = depth_msg->header;
     filtered_msg.height = h;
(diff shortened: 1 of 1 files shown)


v4hn:
Thank you for the fix! This clearly makes sense. I still don't get why `step` needs to be part of the Image message in the first place...

Looking at your patch, I felt the need to adapt the surrounding code a bit to, setting `is_bigendian` from the host system and improving readability a bit.
Could you please try out the current branch on your system, just to validate behavior did not change?

moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp
```diff
@@ -394,33 +394,32 @@ void DepthImageOctomapUpdater::depthImageCallback(const sensor_msgs::ImageConstP
   {
     sensor_msgs::Image debug_msg;
     debug_msg.header = depth_msg->header;
-    debug_msg.height = depth_msg->height;
-    debug_msg.width = depth_msg->width;
+    debug_msg.height = h;
+    debug_msg.width = w;
+    debug_msg.is_bigendian = HOST_IS_BIG_ENDIAN;
```
gavanderhoorn: Is this correct? I'm not a `sensor_msgs/Image` expert, but the comment in [the documentation](http://docs.ros.org/melodic/api/sensor_msgs/html/msg/Image.html) says:

> is this data bigendian?

That does not necessarily mean that the field should reflect whether the originating or processing machine is big endian, only whether the data is.

There are image formats which store data MSB first, and those that store data LSB first. Would that not be what this field is supposed to encode?

v4hn: You are right, in theory a big-endian system can of course generate and send a little-endian image message and the flag refers to the data, not the system.
But all data fields here are filled in by reinterpret_cast from local arrays of `float`, `unsigned short`, and `unsigned int` which where generated through computation on the host.
Thus, the data in these messages has host-endianness.

Incidentally, the plugin will also [generate ROS_ERRORs when the received data does not match host endianness](https://github.com/ros-planning/moveit/blob/master/moveit_ros/perception/depth_image_octomap_updater/src/depth_image_octomap_updater.cpp#L316) (instead of silently converting it, which would be quite a bit of additional computation).

MatthiasNieuwenhuisen: Your changes look good to me and did not change the behavior on my system.

v4hn:
Thank you for the quick verification @MatthiasNieuwenhuisen .

I'll merge this as straight-forward fixup + cosmetic changes.
I'll also file a foreword-cherry-pick for `master`

PR #1653, opened 2019-08-30
depth_image_octomap_updater: correctly set properties of debug images

v4hn:
Cherry-pick of #1652 to `master`.

* calculate step width independent from input depth image

* Also set endianness based on the host system, as the data stems from local computation.

* depth updater: streamline Image properties

* depth updater: enhance readability for filtered image generation


v4hn:
Indeed, thank you.

I asked @davetcoleman to put permissions and safety restrictions on the agenda for the next maintainer meeting.

PR #1572, opened 2019-07-18
Unified Collision Environment Bullet

j-petit:
### Description
This PR unifies the Bullet collision environment and gets the feature branch to build again.

I reduced / improved the Bullet glue code in many places through creating a common parent class for both discrete and cast collision checkers and removed a lot of unnecessary code in `bullet_utils.h`.

PR #1657, opened 2019-09-03
Multi threaded collision checking test

j-petit:
### Description
This adds a unit test for multi-threaded collision checking using FCL.

### Checklist

### DIFF
--- a/moveit_core/planning_scene/test/test_multi_threaded.cpp
+++ b/moveit_core/planning_scene/test/test_multi_threaded.cpp
@@ -38,7 +38,6 @@
 #include <moveit/robot_state/robot_state.h>
 #include <moveit/utils/robot_model_test_utils.h>
 #include <moveit/planning_scene/planning_scene.h>
-#include <urdf_parser/urdf_parser.h>
 #include <gtest/gtest.h>
 #include <thread>
 
@@ -46,20 +45,26 @@
 #include <moveit/collision_detection/collision_env.h>
 #include <moveit/collision_detection/collision_detector_allocator.h>
 
-#include <urdf_parser/urdf_parser.h>
-
 const int TRIALS = 1000;
 const int THREADS = 2;
 
-void runCollisionDetection(unsigned int id, unsigned int trials, const planning_scene::PlanningScene* scene,
+bool runCollisionDetection(unsigned int id, unsigned int trials, const planning_scene::PlanningScene* scene,
                            const robot_state::RobotState* state)
 {
   collision_detection::CollisionRequest req;
+  collision_detection::CollisionResult res;
   for (unsigned int i = 0; i < trials; ++i)
   {
-    collision_detection::CollisionResult res;
+    res.clear();
     scene->checkCollision(req, res, *state);
   }
+  return res.collision;
+}
+
+void runCollisionDetectionAssert(unsigned int id, unsigned int trials, const planning_scene::PlanningScene* scene,
+                                 const robot_state::RobotState* state, bool expected_result)
+{
+  ASSERT_EQ(expected_result, runCollisionDetection(id, trials, scene, state));
 }
 
 class CollisionDetectorThreadedTest : public testing::Test
@@ -100,6 +105,7 @@ TEST_F(CollisionDetectorThreadedTest, FCLThreaded)
 {
   std::vector<robot_state::RobotStatePtr> states;
   std::vector<std::thread*> threads;
+  std::vector<bool> collisions;
 
   for (unsigned int i = 0; i < THREADS; ++i)
   {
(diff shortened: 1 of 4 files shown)


moveit_core/test/test_multi_threaded.cpp
```diff
@@ -0,0 +1,129 @@
...
+  for (unsigned int i = 0; i < THREADS; ++i)
+    threads.push_back(
+        new std::thread(std::bind(&runCollisionDetection, i, TRIALS, planning_scene_.get(), states[i].get())));
+
+  for (unsigned int i = 0; i < states.size(); ++i)
+  {
+    threads[i]->join();
+    delete threads[i];
+  }
+  ASSERT_TRUE(true);
```
BryceStevenWilley: This is a good start to make sure nothing crashes, but I think we should also check that the collision detection calls are correct. Could we check each state when adding it to `states` and make sure each result in the thread is correct?

j-petit: Added this in the latest commit

j-petit:
I moved the test into the `planning_scene`. Thanks for reviewing Bryce!

mamoll:
Cool, this is one area where FCL and Bullet differ, no? I seem to remember that Bullet has some mutable members / persistent state that would make it not thread-safe. This PR looks fine, but making explicit the limitations of Bullet in the documentation somewhere is helpful and doesn't take away from all the work you have done.

j-petit:
Yes, the current Bullet collision checking can not be used multi-threaded in the same way as FCL because of the mutable collision managers. I added a comment in the bullet readme, see [this commit](https://github.com/ros-planning/moveit/pull/1572/commits/f2d9a934a71e1a4843b8ee11dd4a13ce74de94ff).

moveit_core/planning_scene/test/test_multi_threaded.cpp
```diff
@@ -0,0 +1,139 @@
...
+  for (unsigned int i = 0; i < THREADS; ++i)
+    threads.push_back(new std::thread(
+        std::bind(&runCollisionDetectionAssert, i, TRIALS, planning_scene_.get(), states[i].get(), collisions[i])));
+
+  for (unsigned int i = 0; i < states.size(); ++i)
+  {
+    threads[i]->join();
+    delete threads[i];
+  }
+  ASSERT_TRUE(true);
```
henningkayser: Is this a critical test case? ;)

moveit_core/planning_scene/test/test_multi_threaded.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <moveit/planning_scene/planning_scene.h>
+#include <urdf_parser/urdf_parser.h>
+#include <gtest/gtest.h>
+#include <thread>
+
+#include <moveit/collision_detection/collision_common.h>
+#include <moveit/collision_detection/collision_env.h>
+#include <moveit/collision_detection/collision_detector_allocator.h>
+
+#include <urdf_parser/urdf_parser.h>
```
henningkayser: included twice, is this even needed?

henningkayser:
@j-petit looks good, +1 after comments are resolved

moveit_core/planning_scene/test/test_multi_threaded.cpp
j-petit: wups, leftover from the commit before. Removed now.

moveit_core/planning_scene/test/test_multi_threaded.cpp
j-petit: no, it's not, thanks for catching this!

j-petit:
Ping, can be merged?

moveit_core/planning_scene/test/test_multi_threaded.cpp
henningkayser: @j-petit it's still there?

j-petit: Sorry, I forgot to push I think.. Now it's definitely fixed, I double checked :)

moveit_core/planning_scene/test/test_multi_threaded.cpp
```diff
@@ -0,0 +1,135 @@
...
+  collision_detection::CollisionEnvPtr cenv_;
+
+  collision_detection::AllowedCollisionMatrixPtr acm_;
+
+  robot_state::RobotStatePtr robot_state_;
+
+  planning_scene::PlanningScenePtr planning_scene_;
+};
+
+TEST_F(CollisionDetectorThreadedTest, InitOk)
```
rhaschke: To ensure that the robot_model ist valid, it doesn't make sense to define a separate test like this, as each test runs from its own test fixture instance. From the [official docs](https://github.com/google/googletest/blob/master/googletest/docs/primer.md):

> For each test defined with TEST_F(), googletest will create a fresh test fixture at runtime, immediately initialize it via SetUp(), run the test, clean up by calling TearDown(), and then delete the test fixture. Note that different tests in the same test suite have different test fixture objects, and googletest always deletes a test fixture before it creates the next one. googletest does not reuse the same test fixture for multiple tests. Any changes one test makes to the fixture do not affect other tests.

j-petit: But to ensure that in general this works? I assumed this to be good practice as this is done in all collision test files which existed like [here](https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection_fcl/test/test_fcl_collision_detection.cpp)

rhaschke: It doesn't make sense there as well. Better it would be to `ASSERT(robot_model_)` in `Setup()`.

j-petit: I made a PR #1777 to fix this. For now, just for this test as the other's will be changed to the templated version when `feature-bullet` is merged.

PR #1658, opened 2019-09-03
Fix plot details, correcting xlabels positions and cleaning the graph

KaikeWesleyReis:
### Description
Correcting minor issues in plot results from *moveit_benchmark_statistics.py*. This changes was made to increase the graphical potencial from this tool to papers plots (This solution first came for a publication that I came with, using moveit benchmarking).

### Checklist
- Change xtick label rotation from 30 to 45, fixing the issue that the labels was not syncronyzed with the boxplot/barplot.
- Solve the labels names for OMPL planners that by default appears with 'kConfigDefault* at the end (some websites recommend to put this string in the yaml file definition from moveit!). This solutions replace this part of the string by ''
- Take out nanSolutions count from the ax.text, making the chart cleaner

#### Before
a1
#### After
a2

**OBS** This is my first contribution to a project in github, any issue or anything please respond this request.

Best regards,

### DIFF
--- a/moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
+++ b/moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
@@ -306,7 +306,7 @@ def plotAttribute(cur, planners, attribute, typename):
                 color=matplotlib.cm.hot(int(floor(i*256/numValues))),
                 label=descriptions[i])
             heights = heights + measurements[i]
-        xtickNames = plt.xticks([x+width/2. for x in ind], labels, rotation=30)
+        xtickNames = plt.xticks([x + width / 2. for x in ind], labels, rotation=45, fontsize=8)
         ax.set_ylabel(attribute.replace('_',' ') + ' (%)')
         box = ax.get_position()
         ax.set_position([box.x0, box.y0, box.width * 0.8, box.height])
@@ -318,7 +318,10 @@ def plotAttribute(cur, planners, attribute, typename):
         measurementsPercentage = [sum(m) * 100. / len(m) for m in measurements]
         ind = range(len(measurements))
         plt.bar(ind, measurementsPercentage, width)
-        xtickNames = plt.xticks([x + width / 2. for x in ind], labels, rotation=30, fontsize=8)
+	### Insert take out of this term, a practical solution for OMPL solvers
+	labels = [l.replace('kConfigDefault', '') for l in labels]
+
+        xtickNames = plt.xticks([x + width / 2. for x in ind], labels, rotation=45, fontsize=8)
         ax.set_ylabel(attribute.replace('_',' ') + ' (%)')
         plt.subplots_adjust(bottom=0.3) # Squish the plot into the upper 2/3 of the page.  Leave room for labels
     else:
@@ -330,9 +333,12 @@ def plotAttribute(cur, planners, attribute, typename):
 
         #xtickNames = plt.xticks(labels, rotation=30, fontsize=10)
         #plt.subplots_adjust(bottom=0.3) # Squish the plot into the upper 2/3 of the page.  Leave room for labels
-
-        xtickNames = plt.setp(ax,xticklabels=labels)
-        plt.setp(xtickNames, rotation=30)
+	
+	### Insert take out of this term, a practical solution for OMPL solvers
+	labels = [l.replace('kConfigDefault', '') for l in labels]
+        
+	xtickNames = plt.setp(ax,xticklabels=labels)
+        plt.setp(xtickNames, rotation=45)
         for tick in ax.xaxis.get_major_ticks(): # shrink the font size of the x tick labels
             tick.label.set_fontsize(8)
         plt.subplots_adjust(bottom=0.3) # Squish the plot into the upper 2/3 of the page.  Leave room for labels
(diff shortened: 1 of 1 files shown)


moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
```diff
@@ -318,7 +318,10 @@ def plotAttribute(cur, planners, attribute, typename):
         measurementsPercentage = [sum(m) * 100. / len(m) for m in measurements]
         ind = range(len(measurements))
         plt.bar(ind, measurementsPercentage, width)
-        xtickNames = plt.xticks([x + width / 2. for x in ind], labels, rotation=30, fontsize=8)
+	### Insert take out of this term, a practical solution for OMPL solvers
+	labels = [l.replace('kConfigDefault', '') for l in labels]
```
v4hn: Sorry, but although I agree this might look cleaner, I believe the name needs to keep it's full length, as it reflects the configuration specified in `config/ompl_planning.yaml`. For most planners, this involves a number of parameters and you can specify arbitrary names in the config referring to the same planner with different parameterization.
Please remove this again for upstream.

moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
```diff
@@ -330,9 +333,12 @@ def plotAttribute(cur, planners, attribute, typename):
 
         #xtickNames = plt.xticks(labels, rotation=30, fontsize=10)
         #plt.subplots_adjust(bottom=0.3) # Squish the plot into the upper 2/3 of the page.  Leave room for labels
-
-        xtickNames = plt.setp(ax,xticklabels=labels)
-        plt.setp(xtickNames, rotation=30)
+	
+	### Insert take out of this term, a practical solution for OMPL solvers
+	labels = [l.replace('kConfigDefault', '') for l in labels]
```
v4hn: see above

moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
```diff
@@ -342,7 +348,7 @@ def plotAttribute(cur, planners, attribute, typename):
         maxy = max([max(y) for y in measurements])
         for i in range(len(labels)):
             x = i+width/2 if typename=='BOOLEAN' else i+1
-            ax.text(x, .95*maxy, str(nanCounts[i]), horizontalalignment='center', size='small')
+            ### ax.text(x, .95*maxy, str(nanCounts[i]), horizontalalignment='center', size='small')
```
v4hn: Could you please add a comment above the commented line that explains what this would add if uncommented?

moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
KaikeWesleyReis: I will comment this line and specify the reason above, is this ok to you? Or I simply remove?

moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
KaikeWesleyReis: See above.

moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
KaikeWesleyReis: Will be done.

moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
KaikeWesleyReis: I made your changes in the second commit, please check.
Thanks,

moveit_ros/benchmarks/scripts/moveit_benchmark_statistics.py
KaikeWesleyReis: I made your changes in the second commit, please check.
Thanks,

KaikeWesleyReis:
@rhaschke 
> @KaikeWesleyReis, looks like your fixup commit messed up with indentation. I pushed another fixup commit - please have a look.

Well, sorry for this part. I reviewed my code and it's indentation in my remote, I will do another commit and if continues this problem, please say how can I get your commit to my fork.

> I also re-enabled displaying the number of failed attempts. While you might not want to see those figures in your plot, they might be important for others.

This number appears in all graphs, but there is a specific graph that is exactly this info (solved):

image

So besides the "clean view" as I said before, plotting this value in all graphs would be a over information that I don't think that is necessary.

> I don't quite understand, why your _after_ image has so much less bars than your _before_ image?
> If you changed the benchmark config, the plots become not comparable. Please update the plots.

Sorry, because I used other db for the second plot. Here is a complete plot (but is not the same data, but with the same planners). PS: You will notice a 90 degrees rotation.
plot

> Finally, I don't really like the change of rotation from 30 to 45 degrees. Less rotation is usually easier to read. However, I definitely would appreciate if you could _right-align_ the labels to the bars. Now, they are centered on the bars, which is confusing when there are many bars and labels.

This situation is a bit complex. For example, with a 90 degree rotation you can see that particular bar belongs to that planner. However at 30 degrees it is more complicated to figure out who belongs to whom, because the plot labels shifts to the right (as you can see in the first image from this PR). 
I noticed that at 45 degrees, this question persists. My solution would be 90 degrees, but it gets complicated to read (as you could see previously). This could be changed by **decreasing the planners name or fontsize** to maintain legibility and positioning.

I will wait for further instructions to change my code.

Best regards,

rhaschke:
> I reviewed my code and it's indentation in my remote, I will do another commit and if continues this problem, please say how can I get your commit to my fork.

I already pushed a fixing commit to your fork. You just need to pull it.

> > I also re-enabled displaying the number of failed attempts. While you might not want to see those figures in your plot, they might be important for others.

> This number appears in all graphs, but there is a specific graph that is exactly this info (solved).

I see. In this case, I agree with you. Please feel free to revert to your initial version.

> Well, I followed what you said and this would be the results (30 degrees, ha='right'). Is this ok to you?

This looks great! Please push this.

KaikeWesleyReis:
@rhaschke I got your commit and made the changes in this new commit. Please, feel free to test and give a feedback.

PR #1661, opened 2019-09-09
depth_image_octomap_updater: reset depth transfer function to standard values

### LINKED ISSUE #1109: Correctly fix moveit_ros_perception's mesh filter
Since #982, unit tests for `moveit_ros_perception` are disabled, [as the mesh_filter_test fails](https://travis-ci.org/ros-planning/moveit/jobs/403081276).
As the test passes on my side with an nvidia OpenGL setup, the issues seems to be related to Mesa OpenGL. 
* Either there is a (newly introduced) bug in Mesa OpenGL - this test used to pass and we didn't change it for ages.
* Or the way to use the API has changed somewhere in between.

Somebody with experience in offscreen OpenGL rendering should have a look at this.
https://github.com/ros-planning/moveit/tree/kinetic-devel/moveit_ros/perception/mesh_filter

Unfortunately, due to a typo in cmake files (fixed in #954), the mesh filter wasn't built for quite a while.

MatthiasNieuwenhuisen:
### Description

Currently, the pixel transfer function is set accordingly to the encoding of the input sensor data in the filtering step. This seems to be correct for rendering the sensor data only, but causes errors when rendering the robot model and when visualizing the filtered data and robot model as depth images.

This PR resets the pixel transfer function after sensor data has been rendered.

This also partially fixes #1109 and #374 for me (using Mesa OpenGL). Some tests still fail because of numerical inaccuracies.

As the mesh filter does depth calculations distributed over multiple places and, hence, is hard to debug, a review of this patch by someone who is more familiar with the mesh filter would be highly appreciated.

Examples:

Visualization of depth images without the PR (white: input depth image, yellow: filtered_cloud, red: model_depth) for floating point images. Z-scaling of the filtered cloud and the model_depth is wrong. The OctoMap (cubes) is scaled correctly as it does not employ the z-values but only the labels:

float-wo-modifications-ps

Visualization of the same situation for an unsigned depth image (from a RealSense). The rendered model and the filtered cloud disappear due to the scaling:

uint-wo-modifications

Unsigned with my modifications:

uint-w-modifications

When rendering the robot model robot model too close to the camera (cut by near clipping plane), the model's z-values are distributed over the depth buffer (unsigned image):

uint-too-close

With the modification, model points behind the near clipping plane are rendered as expected:

uint-too-close-modified

Kind regards,
Matthias

### DIFF
--- a/moveit_ros/perception/mesh_filter/src/gl_renderer.cpp
+++ b/moveit_ros/perception/mesh_filter/src/gl_renderer.cpp
@@ -182,7 +182,7 @@ void mesh_filter::GLRenderer::deleteFrameBuffers()
 
 void mesh_filter::GLRenderer::begin() const
 {
-  glPushAttrib(GL_VIEWPORT_BIT | GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_POLYGON_BIT);
+  glPushAttrib(GL_VIEWPORT_BIT | GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_POLYGON_BIT | GL_PIXEL_MODE_BIT);
   glBindFramebuffer(GL_FRAMEBUFFER, fbo_id_);
   glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
   glViewport(0, 0, width_, height_);
--- a/moveit_ros/perception/mesh_filter/src/mesh_filter_base.cpp
+++ b/moveit_ros/perception/mesh_filter/src/mesh_filter_base.cpp
@@ -382,8 +382,6 @@ void mesh_filter::MeshFilterBase::doFilter(const void* sensor_data, const int en
   glActiveTexture(GL_TEXTURE4);
   glBindTexture(GL_TEXTURE_2D, color_texture);
   glCallList(canvas_);
-  glPixelTransferf(GL_DEPTH_SCALE, 1.);
-  glPixelTransferf(GL_DEPTH_BIAS, 0.);
   depth_filter_->end();
 }
 
--- a/moveit_ros/perception/mesh_filter/src/mesh_filter_base.cpp
+++ b/moveit_ros/perception/mesh_filter/src/mesh_filter_base.cpp
@@ -383,6 +383,8 @@ void mesh_filter::MeshFilterBase::doFilter(const void* sensor_data, const int en
   glBindTexture(GL_TEXTURE_2D, color_texture);
   glCallList(canvas_);
   depth_filter_->end();
+  glPixelTransferf(GL_DEPTH_SCALE, 1.);
+  glPixelTransferf(GL_DEPTH_BIAS, 0.);
 }
 
 void mesh_filter::MeshFilterBase::setPaddingOffset(float offset)
--- a/moveit_ros/perception/mesh_filter/src/mesh_filter_base.cpp
+++ b/moveit_ros/perception/mesh_filter/src/mesh_filter_base.cpp
@@ -382,9 +382,9 @@ void mesh_filter::MeshFilterBase::doFilter(const void* sensor_data, const int en
   glActiveTexture(GL_TEXTURE4);
   glBindTexture(GL_TEXTURE_2D, color_texture);
   glCallList(canvas_);
-  depth_filter_->end();
   glPixelTransferf(GL_DEPTH_SCALE, 1.);
   glPixelTransferf(GL_DEPTH_BIAS, 0.);
+  depth_filter_->end();
 }
 
 void mesh_filter::MeshFilterBase::setPaddingOffset(float offset)


rhaschke:
Thanks a lot for attempting to fix #1109! Unfortunately, I'm not an expert of the mesh filter either and thus cannot judge this change. Unfortunately, I cannot recognize anything useful in the example images either -- I would expect to see the robot somehow in the raw depth images, but I can't. That's probably my fault, I'm not used to interpret depth images.

What puzzles me, is that the mesh filter (at least the unit tests) used to work in the past (back in Indigo), but we havent' changed them. Also, they still work on nvidia. So, my key question is: why is your proposed change required for Mesa OpenGL and how does it impact the mesh filter on nvidia OpenGL?

v4hn:
> Unfortunately, I cannot recognize anything useful in the example images either

Pictures are simply not a good way to illustrate changes in 3d...
Short screencasts with varying camera angle are so much easier to make sense of.

I'll elaborate a bit on my understanding of this request. Please correct me @MatthiasNieuwenhuisen if I misunderstood something.

> I would expect to see the robot somehow in the raw depth images, but I can't.

- Part of it shows up in the white cloud. The robot is mostly outside the field of view of the camera.
- The pictures show a clear mismatch between white (unfiltered) and yellow (filtered) without the patch, that is resolved by the patch. As the mismatch appears to be a scaling factor in the first image, it makes sense that resetting a stateful scaling variable after applying the operations it was assigned for fixes it.

- The red robot model cloud, which looks good in the first images, suddenly appears all over the place for the robot outside the clipping range. After resetting a stateful variable (as above) for the min range, represented by a bias value in transfer, this is fixed too, which makes sense to me.

> why is your proposed change required for Mesa OpenGL and how does it impact the mesh filter on nvidia OpenGL

I did not validate this at all, but I could imagine the lifetime and scope of changes to `GL_DEPTH_SCALE` and `GL_DEPTH_BIAS` is not aligned between implementations.
Assuming the values transfer between framebuffers only in current OpenGL, this would make sense.

Which brings me to feedback to this patch I missed last time:
@MatthiasNieuwenhuisen, did you intentionally add the resets *after* `depth_filter_->end();`?
I would expect them to happen before releasing the framebuffer.
Could you please verify whether this changes anything for your use case?

MatthiasNieuwenhuisen:
> > Unfortunately, I cannot recognize anything useful in the example images either
> 
> Pictures are simply not a good way to illustrate changes in 3d...
> Short screencasts with varying camera angle are so much easier to make sense of.

I will create one and upload it.

> I'll elaborate a bit on my understanding of this request. Please correct me @MatthiasNieuwenhuisen if I misunderstood something.
> 
> > I would expect to see the robot somehow in the raw depth images, but I can't.
> 
>     * Part of it shows up in the white cloud. The robot is mostly outside the field of view of the camera.

That's correct, part of the elbow are visible in the white cloud.

>     * The pictures show a clear mismatch between white (unfiltered) and yellow (filtered) without the patch, that is resolved by the patch. As the mismatch appears to be a scaling factor in the first image, it makes sense that resetting a stateful scaling variable after applying the operations it was assigned for fixes it.

The first picture shows that the filtered cloud (yellow) and the rendered robot model (red) are scaled wrong. 

>     * The red robot model cloud, which looks good in the first images, suddenly appears all over the place for the robot outside the clipping range. After resetting a stateful variable (as above) for the min range, represented by a bias value in transfer, this is fixed too, which makes sense to me.

The fourth picture shows that the rendered robot model is broken and the resulting filtered cloud is wrong as a result.

> > why is your proposed change required for Mesa OpenGL and how does it impact the mesh filter on nvidia OpenGL
> 
> I did not validate this at all, but I could imagine the lifetime and scope of changes to `GL_DEPTH_SCALE` and `GL_DEPTH_BIAS` is not aligned between implementations.
> Assuming the values transfer between framebuffers only in current OpenGL, this would make sense.

Unfortunately I'm not an expert in OpenGL myself, thus a review from an expert would be great, if there is one... 

> Which brings me to feedback to this patch I missed last time:
> @MatthiasNieuwenhuisen, did you intentionally add the resets _after_ `depth_filter_->end();`?
> I would expect them to happen before releasing the framebuffer.
> Could you please verify whether this changes anything for your use case?

This should not make any difference in my understanding, as begin() and end() push and pop state variables, not including the ones that I change. I can also not observe a difference when testing this. I did not have the chance to check the visual output on a system with Nvidia, I will test it as soon as possible.

> @rhaschke It's not clear to me whether you can specify some criteria to get this merged from your perspective or whether you want to wait for another maintainer to approve instead.

rhaschke:
@MatthiasNieuwenhuisen, thanks for the clarification of the images. As I understand it, the yellow (filtered) points should match to the white (raw) point cloud, which wasn't the case without your modifications.

> @rhaschke, can you specify some criteria to get this merged from your perspective?

Of course, a review from an OpenGL expert would be great. But I'm afraid there is none in the MoveIt maintainer team. @wjwwood or @iche033, can you have a look or suggest somebody?

Apart from this, an explanation why it does work on nvidia, but not in OpenGL, and why the behaviour of OpenGL changed at some point, would be great. However, I'm fully aware, that this might be a too hard hurdle for this PR to be accepted.

MatthiasNieuwenhuisen:
> @MatthiasNieuwenhuisen, thanks for the clarification of the images. As I understand it, the yellow (filtered) points should match to the white (raw) point cloud, which wasn't the case without your modifications.
> 

The yellow cloud should match the white cloud with the exception of points filtered out because they are in the shadow of the padded robot model. E.g., the point-free circle in the left bottom of the third picture is larger in the filtered cloud because of the padding around the robot model.

v4hn:
> > specify some criteria to get this merged
> 
> Of course, a review from an OpenGL expert would be great. But I'm afraid there is none in the MoveIt maintainer team. However, I'm fully aware, that this might be a too hard hurdle for this PR to be accepted.

I very much agree with the whole statement. :smile: 
I don't expect anyone with a strong background in visualization and OpenGL wants to spend their time debugging this OpenGL-backend issue that's only relevant when working with live sensor data, so I was looking for other ways to get this merged...

@MatthiasNieuwenhuisen 
> This should not make any difference in my understanding, as begin() and end() push and pop state variables, not including the ones that I change.
> I can also not observe a difference when testing this

From a code design perspective this should still be inside the `begin()/end()` block though, as the variables are also assigned inside the block. If it makes no difference in behavior, please change it still.

wjwwood:
> @wjwwood or @iche033, can you have a look or suggest somebody?

Sorry, I'm by far not an OpenGL expert. Maybe @iche033 can comment if he has time (but he's very busy), otherwise I sadly don't have anyone to recommend.

iche033:
I have to say I have not touched these OpenGL call in a while and have mainly been doing operations like these directly in the shaders. The code changes make sense though. 

I did wonder why the reset is not handled by `GLRenderer`'s `begin()` and `end()` functions. From reading [glPushAttrib](https://opengl.developpez.com/docs/man/man2/glPushAttrib) documentation, `GL_DEPTH_SCALE` and `GL_DEPTH_BIAS` are part of the `GL_PIXEL_MODE_BIT` mask. So one thing to try is add `GL_PIXEL_MODE_BIT` to `glPushAttrib` in the [begin](https://github.com/ros-planning/moveit/blob/master/moveit_ros/perception/mesh_filter/src/gl_renderer.cpp#L183) function and see if that alone resolves the issue.

rhaschke:
@iche033, thanks for reviewing and your suggestion. @MatthiasNieuwenhuisen, could you give this a try?

MatthiasNieuwenhuisen:
I tested the modification suggested by @iche033 and it resolves the problem, too.

PR #1672, opened 2019-09-19
Fix double mutex unlock in JogArm

henningkayser:
The double unlock caused spurious errors. This PR should be cherry-picked to melodic.


rhaschke:
@henningkayser, the jogging package isn't part of melodic-devel (and probably shouldn't be), so why cherry-picking?

henningkayser:
@rhaschke thanks for merging. You're right, I just thought it was part of melodic

PR #1683, opened 2019-10-02
fixed REALTIME planned path trajectory playback

MarqRazz:
### Description
When using the REALTIME mode to display the planned path the program would stop at the last or second to last waypoint and is unable to re-play the trajectory when using the trajectory slider plugin.

This pull request fixes the REALTIME playback mode to always finish playing the full trajectory and has the ability to properly pause and continue playback using the trajectory slider plugin.

Please let me know if you would like me to add any tests, features or changes.
Thanks,
-Marq

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
@@ -43,17 +43,17 @@
 #include <moveit/rviz_plugin_render_tools/robot_state_visualization.h>
 #include <rviz/robot/robot.h>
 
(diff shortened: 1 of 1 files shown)


MarqRazz:
@henningkayser  what problems are you seeing with loop animation enabled?

I just ran a few simple tests and don't see any issues. The tool was able to loop the entire animation and was also able to pickup looping when starting from any point in the trajectory. If I know what to look for I might have some time to fix it.

henningkayser:
@MarqRazz actually, I expected the TrajectorySlider to have the same behavior as the MotionPlanning slider (syncing to the loop), which obviously is not the case. So nothing to fix here ;)

Issue #1682, opened 2019-10-02
Missing large voxels in octomap visualization in RViz


RMonica:
### Description

Most large voxels in an octree are not displayed in the planning scene.

The issue is in `moveit_ros/visualization/rviz_plugin_render_tools/src/octomap_render.cpp`, method `octreeDecoding`.
There is an optimization to remove voxels which are completely surrounded by other voxels and can't be seen.
Something is wrong, here: for example, it seems that the `key` variable contains a key for depth `octree_depth_`, but it is used in a full-depth search later: `octree->search(key)`. Changing to `octree->search(key, octree_depth_)` is not enough to fix the issue, however.

If I disable the optimization, like this:
```
diff --git a/moveit_ros/visualization/rviz_plugin_render_tools/src/octomap_render.cpp b/moveit_ros/visualization/rviz_plugin_render_tools/src/octomap_render.cpp
index f427a7315..672b4738b 100644
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/octomap_render.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/octomap_render.cpp
@@ -172,7 +172,8 @@ void OcTreeRender::octreeDecoding(const std::shared_ptr<const octomap::OcTree>&
     // traverse all leafs in the tree:
     for (octomap::OcTree::iterator it = octree->begin(octree_depth_), end = octree->end(); it != end; ++it)
     {
-      bool display_voxel = false;
+      //bool display_voxel = false;
+      bool display_voxel = ((int)octree->isNodeOccupied(*it) + 1) & render_mode_mask;
 
       // the left part evaluates to 1 for free voxels and 2 for occupied voxels
       if (((int)octree->isNodeOccupied(*it) + 1) & render_mode_mask)
```
the large voxels are displayed.

Example:
With the optimization:
![orig_ok](https://user-images.githubusercontent.com/1881566/66028277-0bf56080-e4fd-11e9-8af7-7e6377d17f0b.png)
Optimization disabled:
![no_optim_ok](https://user-images.githubusercontent.com/1881566/66028311-22032100-e4fd-11e9-8d07-642d7e2e3ac1.png)

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Source build: commit 2a3d8f8d4b729b3c7aaf1ed9175df478ff52b5ea

### Steps to reproduce

- Launch RViz.
- Add a planning scene display.
[...]

PR #1685, opened 2019-10-06
Fix pruning of enclosed nodes when rendering octomap in RViz

RMonica:
### Description

Fixes issue
https://github.com/ros-planning/moveit/issues/1682
by porting
https://github.com/OctoMap/octomap_rviz_plugins/pull/30

I took the liberty of using `getIndexKey` instead of `getKey`. AFAIK, `getIndexKey` returns the key of the first corner of the voxel. Then, all those ternaries in the original patch are not necessary.

### Checklist

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/octomap_render.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/octomap_render.cpp
@@ -169,7 +169,7 @@ void OcTreeRender::octreeDecoding(const std::shared_ptr<const octomap::OcTree>& 
 
   size_t point_count = 0;
   {
-    int stepSize = 1 << (octree->getTreeDepth() - octree_depth_);  // for pruning of occluded voxels
+    int step_size = 1 << (octree->getTreeDepth() - octree_depth_);  // for pruning of occluded voxels
 
     // traverse all leafs in the tree:
     for (octomap::OcTree::iterator it = octree->begin(octree_depth_), end = octree->end(); it != end; ++it)
@@ -200,13 +200,13 @@ void OcTreeRender::octreeDecoding(const std::shared_ptr<const octomap::OcTree>& 
           for (int i = 0; all_neighbors_found && i < 2; ++i)
           {
             key[idx_0] = n_key[idx_0] + diff[i];
-            // if rendering is restricted to treeDepth < maximum tree depth inner nodes with distance stepSize can
+            // if rendering is restricted to treeDepth < maximum tree depth inner nodes with distance step_size can
             // already occlude a voxel
             for (key[idx_1] = n_key[idx_1] + diff[0] + 1; all_neighbors_found && key[idx_1] < n_key[idx_1] + diff[1];
-                 key[idx_1] += stepSize)
+                 key[idx_1] += step_size)
             {
               for (key[idx_2] = n_key[idx_2] + diff[0] + 1; all_neighbors_found && key[idx_2] < n_key[idx_2] + diff[1];
-                   key[idx_2] += stepSize)
+                   key[idx_2] += step_size)
               {
                 octomap::OcTreeNode* node = octree->search(key, octree_depth_);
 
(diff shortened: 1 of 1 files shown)


RMonica:
This is a visualization issue in a RViz plugin. I tried, but it seems that the OcTreeRender class cannot be easily allocated without RViz.

Nonetheless, I can explain what the code does in more detail. Actually, I mostly wrote the explanation a few weeks ago, but apparently I forgot to finish it for some reason.

In the `octreeDecoding` function (`octomap_render.cpp`), an OctoMap octree is converted into cubes to be displayed by RViz.
In particular, the `for` at line 173 iterates over every leaf node of the octree and, if the node is occupied (line 178), a cube is generated (lines 210-238).

For performance optimization (lines 180-207), the code checks if the cube is completely surrounded by other occupied cubes. If so, the cube is not displayed, because it would not be visible anyway.
This optimization is currently broken, and less voxels than needed are displayed.

**What happens without this patch**

At line 184, the current node key is obtained. The key is a data structure which contains 3 integers, corresponding to the 3D coordinates of the node divided by the octree resolution and rounded. Then, at lines 186-190, three nested `for`s generate the 26 keys of the voxels in the 26-neighborhood by incrementing and decrementing each key coordinate by one. The nodes at those keys are checked (line 197) and, if all of them are occupied, the voxel is not displayed (line 207).

The issue occurs because some leaf nodes are not at the maximum octree depth, i.e., they are larger than the octree resolution. Since the three `for`s increment and decrement the key coordinates by only one, the new key may be still in the same node. If this happens, the node is considered an occupied neighbor of itself, and it is not displayed.

**What this patch does**

With this patch, the code scans the whole surface of the node, to check for neighbors. Since OctoMap leaf nodes may have multiple sizes, it is not enough to check if there is a neighbor adjacent to each of the six faces. Instead, a node should not be displayed only if the whole surface is covered by neighbors.

My code use the `getIndexKey` method (line 186) to obtain the key `n_key` at one corner of the node, in particular of the corner with lowest coordinates. From this key, neighboring voxels can be be at (-1,*a*,*b*), (*b*,-1,*a*) and (*a*,*b*,-1), and at (`diff_base`,*a*,*b*), (*b*,`diff_base`,*a*) and (*a*,*b*,`diff_base`), where `diff_base` is the current node side in key coordinates, and *a* and *b* vary across the node face. `diff_base` is computed as 2^(number of levels above maximum level in the octree) (line 190).

In practice, the code iterates over the six cube faces (lines 194-200), by selecting one of the three coordinates (`idx_case` in {0, 1, 2}, that is, {x, y, z}) and two directions (negative or positive, `i` in {0, 1}). The array `diff` contains the translation to exit the current node: for the positive direction `diff` is `diff_base`, for the negative direction `diff` is -1 (line 191).
At lines 202-209, the keys of all possible voxels next to the face are generated with two nested `for`s, which iterate over the two coordinates `idx_1` (that is, variable *a*) and `idx_2` (that is, variable *b*) which are different from `idx_case` (initialized at lines 196-198).
If the OctoMap node at at-least-one key is not occupied (line 214), then the face is not fully covered by neighbors, and the node is displayed.

rhaschke:
To "proof" that the code is correct, could you please post rviz screenshots illustrating the behaviour before and after this fix. 
As you said, there is a corresponding rviz plugin as well. Wouldn't it make sense to propose your simplification there as well and then build here upon the code in the original plugin?

RMonica:
Okay, I'll post some screenshots.

Here is the planning scene message: [planning_scene_moveit_test.yaml.zip](https://github.com/ros-planning/moveit/files/4158274/planning_scene_moveit_test.yaml.zip) (reference frame is `world`).

Current MoveIt:
f1_crop

Pruning disabled:
f2_crop

With my patch:
f3_crop

Current MoveIt is incorrect: the front surface is missing, and some holes are visible in the top surface.
Without pruning and my patch give the same result.

This is the same octree viewed from the inside.
Current MoveIt:
d1_crop

Pruning disabled:
d2_crop

With my patch:
d3_crop

Current MoveIt is incorrect: the enclosed voxels have been pruned, but two outer surfaces are also missing.
Without pruning, we can't see anything, as the octree is filled with voxels. This is not wrong, but it is inefficient.
With my patch the octree is hollow, because the enclosed voxels have been pruned.

I have no time to propose the same fix to the rviz plugin. Besides, their plugin is not broken, yours is.
If you don't like `getIndexKey`, I can just open another pull request using `getKey` (and the ternary operators) as in the original patch.

PR #1686, opened 2019-10-08
Add OMPL planner 'AnytimePathShortening', take #2

mamoll:
### Description

This builds on @henningkayser's PR #1629. Due to changes in OMPL, only that first commit of that other PR is now necessary. This PR requires OMPL to be built from source at the moment. I am planning to release a new version of OMPL soon, though. 

This PR allows you configure AnytimePathShortening like so in your ompl_planning.yaml:
```
planner_configs:
  AnytimePathShortening:
    type: geometric::AnytimePathShortening
    shortcut: true
    hybridize: true
    planners: PRM,PRM,PRM[max_nearest_neighbors=5],RRTConnect[goal_bias=.2],RRTConnect,BiEST,BiEST,BiEST
```

Note how several planning instances can be created, even multiple of the same type with different parameters.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
@@ -136,27 +136,27 @@ void ompl_interface::PlanningContextManager::registerDefaultPlanners()
 {
   registerPlannerAllocator("geometric::AnytimePathShortening", allocatePlanner<og::AnytimePathShortening>);
   registerPlannerAllocator("geometric::BFMT", allocatePlanner<og::BFMT>);
-  registerPlannerAllocator("geometric::BiEST",  allocatePlanner<og::BiEST>);
+  registerPlannerAllocator("geometric::BiEST", allocatePlanner<og::BiEST>);
   registerPlannerAllocator("geometric::BiTRRT", allocatePlanner<og::BiTRRT>);
   registerPlannerAllocator("geometric::BKPIECE", allocatePlanner<og::BKPIECE1>);
   registerPlannerAllocator("geometric::EST", allocatePlanner<og::EST>);
   registerPlannerAllocator("geometric::FMT", allocatePlanner<og::FMT>);
   registerPlannerAllocator("geometric::KPIECE", allocatePlanner<og::KPIECE1>);
-  registerPlannerAllocator("geometric::LazyPRM",allocatePlanner<og::LazyPRM>);
+  registerPlannerAllocator("geometric::LazyPRM", allocatePlanner<og::LazyPRM>);
   registerPlannerAllocator("geometric::LazyPRMstar", allocatePlanner<og::LazyPRMstar>);
   registerPlannerAllocator("geometric::LazyRRT", allocatePlanner<og::LazyRRT>);
   registerPlannerAllocator("geometric::LBKPIECE", allocatePlanner<og::LBKPIECE1>);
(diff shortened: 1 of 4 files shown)


v4hn:
Before merging this, we need to talk about our policy for dependencies (esp. OMPL) in the `master` branch. De-facto this requires OMPL in our build instructions.

@mamoll do you plan a release in kinetic/melodic or noetic?
Or just a release in general leaving the ROS deployment for later?

mamoll:
It's probably best that this PR gets merged only after a new version of OMPL is released.

I *can* create a release for kinetic and melodic, but the API/ABI has changed slightly since the last release. If that's acceptable, the release needs to be coordinated with a release of MoveIt. I will do some testing to make sure that things work with released versions of MoveIt.

v4hn:
> I can create a release for kinetic and melodic, but the API/ABI has changed slightly since the last release

This is exactly why I asked and I would recommend **not** to release into anything before noetic.
But this leaves OMPL to be built from source for each and every system that uses MoveIt's `master` branch.

You should definitely do a release first either way.

rhaschke:
Relying on a delicate synchronization of OMPL and MoveIt releases isn't very appealing. Can't we modify this PR to work with both, current and future OMPL releases, e.g. using compile-time switches? In a future release (e.g. for Noetic), these switches can be removed again.
Alternatively, if OMPL can be released without breaking current MoveIt's Kinetic / Melodic releases, that's fine too. Remember, that the ROS build farm rebuilds downstream packages anyway. So if OMPL is API compatible to those releases, an ABI incompatibility shouldn't pose a problem.

v4hn:
> So if OMPL is API compatible to those releases, an ABI incompatibility shouldn't pose a problem

Assuming the SONAME(s?) of libompl is updated. Otherwise, this would of course lead to undetected broken binaries in user codebases.

mamoll:
I added an OMPL version check to make things backwards compatible with older releases of OMPL.

moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
```diff
@@ -134,79 +134,31 @@ ompl_interface::PlanningContextManager::plannerSelector(const std::string& plann
...
-      "geometric::LazyPRMstar",  //
-      std::bind(&allocatePlanner<og::LazyPRMstar>, std::placeholders::_1, std::placeholders::_2,
-                std::placeholders::_3));
-  registerPlannerAllocator(  //
-      "geometric::SPARS",    //
-      std::bind(&allocatePlanner<og::SPARS>, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
-  registerPlannerAllocator(   //
-      "geometric::SPARStwo",  //
-      std::bind(&allocatePlanner<og::SPARStwo>, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
+  registerPlannerAllocator("geometric::AnytimePathShortening", allocatePlanner<og::AnytimePathShortening>);
```
rhaschke: I guess, because all these functions take 3 arguments, it's not necessary to use `std::bind`?

rhaschke:
Please address the clang-format issues.

mamoll:
Will do the reformatting and testing.

moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
mamoll: Yes. I think this revision makes the code more readable. (In general, I'd prefer lambda expressions over std::bind, but in this case neither is necessary.)

PR #1687, opened 2019-10-09
Add docker documentation

felixvd:
### Description

While updating the Docker documentation ([this PR](https://github.com/ros-planning/moveit.ros.org/pull/282)) to include the changes from #1532, I noticed that that page should be referenced and made some minor changes that I believe are helpful when using this script.

The change from `default_container` to `default_moveit_container` is already part of the other PR, so it would be good if at least that bit was included. The name change is for clarity when listing containers via `docker ps -a`.

### Checklist

### DIFF
--- a/.docker/gui-docker
+++ b/.docker/gui-docker
@@ -1,17 +1,14 @@
 #!/bin/bash -u
 
 # This script is used to run a docker container with graphics support.
-# Documentation: https://moveit.ros.org/install/docker/
-# http://wiki.ros.org/docker/Tutorials/Hardware%20Acceleration
-
 # All arguments to this script except "-c <container_name>" will be appended to a docker run command.
 # If a container name is specified, and this container already exists, the container is re-entered,
 # which easily allows entering the same persistent container from multiple terminals.
+# See documentation for detailed examples: https://moveit.ros.org/install/docker/
 
 # Example commands:
 # ./gui-docker --rm -it moveit/moveit:master-source /bin/bash     # Run a (randomly named) container that is removed on exit
 # ./gui-docker -v ~/ros_ws:/root/ros_ws --rm -it moveit/moveit:master-source /bin/bash   # Same, but also link host volume ~/ros_ws to /root/ros_ws in the container
-# ./gui-docker -c container_name -v ~/ros_ws:/root/ros_ws moveit/moveit:master-source echo  # Create a persistent container that links host volume ~/ros_ws to /root/ros_ws
 # ./gui-docker -c container_name                                  # Start (or continue) an interactive bash in a moveit/moveit:master-source container
 # ./gui-docker                                                    # Same, but use the default container name "default_moveit_container"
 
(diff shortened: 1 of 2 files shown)


.docker/gui-docker
```diff
@@ -1,11 +1,17 @@
...
-# All arguments to this script will be appended to a docker run command.
-# Example command line:
-# ./gui-docker -it --rm moveit/moveit:master-source /bin/bash
+# Runs and enters a persistent docker container with moveit installed by default.
+# All arguments to this script except "-c" will be appended to a docker run command.
+
+# Example commands:
+# ./gui-docker
+# ./gui-docker -c different_container_name                              # To create a different persistent container
+# ./gui-docker --rm -it moveit/moveit:master-source /bin/bash           # To remove and recreate the container
```
rhaschke: This command creates a new, randomly named container, which is removed on exit.

felixvd: True, I misread. Will update the other PR as well.

felixvd:
Thanks for the additions and corrections. I suggested another example line to combine persistent container + linked volume, which I think it will be helpful to be new users as `./gui-docker -c container_name -v ~/ros_ws:/root/ros_ws` exits in a way that might be confusing (the `-it` option has to be specified when `-v` is used).

rhaschke:
I suggest to have these different examples in the tutorial only.
I don't see a reason why `-it` has to be specified for `-v`.

.docker/gui-docker
```diff
@@ -6,13 +6,15 @@
...
+# which easily allows entering the same persistent container from multiple terminals.
 
 # Example commands:
 # ./gui-docker --rm -it moveit/moveit:master-source /bin/bash     # Run a (randomly named) container that is removed on exit
-# ./gui-docker -v ~/ros_ws:/root/ros_ws --rm -it moveit/moveit:master-source /bin/bash   # Same, but also link host volume ~/ros_ws to /root/ros_ws
-# ./gui-docker -c different_container_name                        # Start (or continue) an interactive bash in a moveit/moveit:master-source container
+# ./gui-docker -v ~/ros_ws:/root/ros_ws --rm -it moveit/moveit:master-source /bin/bash   # Same, but also link host volume ~/ros_ws to /root/ros_ws in the container
+# ./gui-docker -c container_name                                  # Start (or continue) an interactive bash in a moveit/moveit:master-source container
 # ./gui-docker                                                    # Same, but use the default container name "default_moveit_container"
+# ./gui-docker -c container_name -v ~/ros_ws:/root/ros_ws -it moveit/moveit:master-source /bin/bash  # Same, but also link host volume ~/ros_ws to /root/ros_ws in the container
```
rhaschke: I explicitly dropped this example, because you cannot run this twice in a row:
The second call doesn't allow any options, but just will re-enter the already running container.

I can image though, to add an example like this:
`gui-docker -c container_name <command>`
which runs the command (instead of bash) in non-interactive fashion within an existing container.

rhaschke: Because it's too complicated to explain the interplay of the initial and subsequent gui-docker commands, I suggest dropping the example here. However, in the documentation (https://github.com/ros-planning/moveit.ros.org/pull/282) we can provide the required explanations and the example.

felixvd: Good with me, I pushed the change.

PR #1693, opened 2019-10-16
Update version of occupancy_map_monitor

simonschmeisser:
it is still at 1.0.1 while the old version from `moveit_ros_perception` is at 1.0.2 . This creates confusion


v4hn:
These files should not be edited without a release, although it might not be a problem in this particular case.
Looking into it, the `moveit_ros_occupancy_map_monitor` package has never been released yet,
as it was created only a few days after the previous release.

PR #1705, opened 2019-10-30
plan_execution: refine logging for invalid paths

v4hn:
This differentiates between checks caused by scene updates
and checks caused by changing trajectory components.

### DIFF
--- a/moveit_ros/planning/plan_execution/src/plan_execution.cpp
+++ b/moveit_ros/planning/plan_execution/src/plan_execution.cpp
@@ -262,13 +262,6 @@ void plan_execution::PlanExecution::planAndExecuteHelper(ExecutableMotionPlan& p
                     getErrorCodeString(plan.error_code_).c_str());
 }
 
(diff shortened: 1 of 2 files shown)


moveit_ros/planning/plan_execution/src/plan_execution.cpp
```diff
@@ -266,7 +266,16 @@ bool plan_execution::PlanExecution::isRemainingPathValid(const ExecutableMotionP
 {
   // check the validity of the currently executed path segment only, since there could be
```
henningkayser: I don't see this function being used anymore and it is private. Do we still need to keep it as it's just a wrapper?

v4hn: You're right, I just removed it.

I'm so used to MoveIt exposing every single function, it surprised me to see this one (which might actually be useful to an external user) was actually private. :smile:

PR #1706, opened 2019-10-31
Specify retime_trajectory parameters as arguments.

mmurooka:
### Description

This PR makes it possible to specify retime_trajectory parameters as arguments in Python interface of move_group.

### Checklist

### DIFF
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -591,9 +591,42 @@ class MoveGroupCommander(object):
         self._g.set_support_surface_name(value)
 
     def retime_trajectory(self, ref_state_in, traj_in, velocity_scaling_factor=1.0, acceleration_scaling_factor=1.0, algorithm="iterative_time_parameterization"):
+        """ Retime the trajectory by the time parameterization algorithm """
+        if algorithm == "iterative_time_parameterization":
+            return self.retime_trajectory_itp(ref_state_in, traj_in, velocity_scaling_factor, acceleration_scaling_factor)
+        elif algorithm == "iterative_spline_parameterization":
+            return self.retime_trajectory_isp(ref_state_in, traj_in, velocity_scaling_factor, acceleration_scaling_factor)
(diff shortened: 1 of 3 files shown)


mmurooka:
Thank you for comments, and sorry for late reply.
I updated as the following comment:
> keep the parameters of retime_trajectory unchanged, call retime_trajectory_totg, retime_trajectory_itpt etc

(Original version is copied here: https://github.com/ros-planning/moveit/compare/master...mmurooka:retime-param-all-argument-orig )

I don't know if this is the best, so please let me know if there is a better way.
As an implementation choice, C++ function can also be separated by algorithms although one common function is used at present:
https://github.com/ros-planning/moveit/blob/b8c4c2e2f09db8f1c6794a643c264e1e7c06f874/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp#L487-L491
This makes passing dummy arguments of other algorithms unnecessary:
https://github.com/ros-planning/moveit/blob/b8c4c2e2f09db8f1c6794a643c264e1e7c06f874/moveit_commander/src/moveit_commander/move_group.py#L606-L607

felixvd:
More like sorry for *our* late response. This has been lying around for a while.

I wasn't aware that the C++ implementation had all the parameters in the top-level function. It's not very pretty. Feel free to change it if you like, but it's not needed to merge this PR.

I don't see a problem with this at first sight, but I haven't tested it. You can restart Travis by closing and re-opening this PR. It looks like CI might have failed for unrelated reasons.

v4hn:
By now there is also a file conflict that needs to be resolved.

mmurooka:
rebased origin/master and resolved conflicts. but it's a bit hard to test this PR in my own environment now.

DLu (changes requested):
Sorry, this one looks like it sat unreviewed for long enough in its rebased state that it has gathered more merge conflicts. Should we try to resolve them or close this PR?

v4hn:
@felixvd @mmurooka  Are you interested enough in this PR to clean it up and get it merged? Sorry for the timeouts, we are low on active maintainers who look at stuff like this.

PR #1712, opened 2019-11-06
TrajectoryExecutionManager: protect access to execution_thread_

### LINKED ISSUE #1481: Race condition causes trajectory_execution_manager
### Description
When a trajectory is preempted, trajectory_execution_manager sometimes crashes with a boost mutex error. 

### Environment
* ROS Distro: Melodic
* OS Version: e.g. Ubuntu 18.04
* Moveit binary build
* ros-industrial/universal_robot source build

### Steps to reproduce
The script below should cause the crash within a few minutes.

<script src="https://gist.github.com/henryponti/21d14e59804d294fc3617e21ccdbd91e.js"></script>

### Expected behaviour
move_group should allow the trajectory to be preempted without crashing.

### Actual behaviour
move_group crashes.

### Backtrace

<script src="https://gist.github.com/henryponti/5ed3bafd6f66fb4aedbadea2df82f336.js"></script>

rhaschke:
As outlined in https://github.com/ros-planning/moveit/issues/1481#issuecomment-498731966
stopExecution() is accessed in parallel from several threads, all trying to join and reset
the execution_thread_. As the access was not correctly protected, this resulted in race
conditions and segfaults accessing a just released `execution_thread_`.

Fixes #1481.
Alternative to #1709: Instead of introducing yet another mutex, just reuse `execution_state_mutex_`, which protected the `execution_complete_` flag. As this flag is kind of synonymous to `execution_thread_ == nullptr`, reusing the same mutex makes sense to me.

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
+++ b/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
@@ -1154,7 +1154,7 @@ void TrajectoryExecutionManager::stopExecution(bool auto_clear)
 
   if (!execution_complete_)
   {
-    execution_state_mutex_.lock();
+    boost::mutex::scoped_lock lock(execution_state_mutex_);
     if (!execution_complete_)
     {
       // we call cancel for all active handles; we know these are not being modified as we loop through them because of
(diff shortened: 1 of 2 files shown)


livanov93:
@rhaschke This doesn't work for me. It is blocked after first preemption. I could try to make the test when I catch the time so we can compare fixes easily.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1166,24 +1166,26 @@ void TrajectoryExecutionManager::stopExecution(bool auto_clear)
...
       // we set the status here; executePart() will not set status when execution_complete_ is true ahead of time
       last_execution_status_ = moveit_controller_manager::ExecutionStatus::PREEMPTED;
-      execution_state_mutex_.unlock();
-      ROS_INFO_NAMED(name_, "Stopped trajectory execution.");
-
       // wait for the execution thread to finish
       execution_thread_->join();
       execution_thread_.reset();
 
+      lock.unlock();
```
rhaschke: Misusing `execution_state_mutex_` here to protect `execution_thread_` as well doesn't work, because the execution thread tries to lock the mutex as well to indicate that it's done via the `execution_complete_` flag.

rhaschke:
Closing again as this is not a valid solution to the problem. I'm afraid we need to introduce a new mutex as proposed in #1709. 
Alternatively, I was asking myself to get rid of the boolean flag, which should be identical to `execution_thread_ != nullptr`. @v4hn, did you made progress understanding/restructuring this code?

v4hn:
> @v4hn, did you made progress understanding/restructuring this code?

[Here's my current state](https://github.com/v4hn/moveit/tree/pr-master-execution-thread).
I spent a bit of time by now, generally cleaning up and documenting the class. Still going on as time permits.
As this effort will probably change API and ABI, I'm perfectly happy to use #1709 for `kinetic-devel` and `melodic-devel`.
Assuming we want a quick fix in `master`, we could also have it there for the moment.

PR #1723, opened 2019-11-11
remove opertunity for use after free

tylerjw:
A reference to a shared_ptr in a function call presents an opportunity for use after free bugs.  If we are ok with the overhead cost of shared_ptr for memory safety we should never make a reference to one as it does not increase the reference count and therefore presents opportunity for memory bugs and security issues.

This also applies to functions that return a reference to the object pointed to by a shared pointer.  I did not address those in this PR.  If the goal of using shared_ptr is to get memory safety we should avoid ever having a reference to the object it points to, let alone making that part of the API.

For an example of why this is a bad practice:
```
#include <memory>
#include <thread>
#include <iostream>
#include <functional>

// moveit API function
void foo(std::shared_ptr<int>& data) {
    while( true ) {
        std::cout << data.use_count() << std::endl;
        std::this_thread::sleep_for (std::chrono::milliseconds(100));
    }
}

// async user code
std::shared_ptr<std::thread> my_thread_;
void bar() {
    std::shared_ptr<int> data = std::make_shared<int>(42);
    my_thread_.reset( new std::thread(std::bind(foo, std::ref(data))));
}

int main() {
    bar();
    while(true);
    return 0;
}
```

This will brake any interface that expects the api function to reset the shared_ptr it previously had a reference to.

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/allvalid/collision_env_allvalid.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/allvalid/collision_env_allvalid.h
@@ -46,10 +46,10 @@ namespace collision_detection
 class CollisionEnvAllValid : public CollisionEnv
 {
 public:
-  CollisionEnvAllValid(const robot_model::RobotModelConstPtr& robot_model, double padding = 0.0, double scale = 1.0);
-  CollisionEnvAllValid(const robot_model::RobotModelConstPtr& robot_model, const WorldPtr& world, double padding = 0.0,
+  CollisionEnvAllValid(const robot_model::RobotModelConstPtr robot_model, double padding = 0.0, double scale = 1.0);
+  CollisionEnvAllValid(const robot_model::RobotModelConstPtr robot_model, const WorldPtr world, double padding = 0.0,
                        double scale = 1.0);
-  CollisionEnvAllValid(const CollisionEnv& other, const WorldPtr& world);
+  CollisionEnvAllValid(const CollisionEnv& other, const WorldPtr world);
 
   void checkRobotCollision(const CollisionRequest& req, CollisionResult& res,
                            const robot_state::RobotState& state) const override;
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
@@ -247,7 +247,7 @@ struct DistanceRequest
   }
 
   /// Compute \e active_components_only_ based on \e req_
-  void enableGroup(const robot_model::RobotModelConstPtr& robot_model)
+  void enableGroup(const robot_model::RobotModelConstPtr robot_model)
   {
     if (robot_model->hasJointModelGroup(group_name))
       active_components_only = &robot_model->getJointModelGroup(group_name)->getUpdatedLinkModelsSet();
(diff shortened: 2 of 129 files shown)


jonbinney:
@tylerjw it looks like there are a lot of formatting changes that are adding noise to the PR. Can you re-format using clang as described here: https://moveit.ros.org/documentation/contributing/code/

rhaschke:
I don't really agree with this PR proposal. Usually, when passing a const reference to an API function, it's clear that the pointer will survive the call of the API function and will NOT used beyond (as in your thread). Of course, I agree that it is always possible to abuse such an API. However, I don't agree that this justifies so many additional copies. In the same vein you could argue that references should be avoided per se.
In your particular example, I would expect that the caller (`bar()`) ensures that a copy of the shared_ptr is passed into `foo`.

tylerjw:
I'm content that in general this PR is not worth it based on the feedback.  I've tried several ways to create examples of why this is bad and in many of the simpler cases I could think of the compiler does actually stop you.  You have to maintain access to the reference after the return of the API function call either by storing it as state or by starting a thread.  In both of these situations you are explicitly saving a copy of a reference.  

On the other hand, if a section of code is in a critical path where the run-time cost of copying a shared_ptr is bad we shouldn't use shared_ptr at all.  The copy of a shared_ptr is the cost of making a copy of a pointer on the stack and incrementing a counter.  While that cost is not zero it is deterministic.  On the other hand the act of creating a shared pointer involves a heap allocation and that is not deterministic and in the worst case can be very slow.  This means that if you expect these interfaces to be used in a deterministic control loop we should abandon shared_ptr all together.  If they are not expected to be used in a real time control loop caring about the runtime cost of a copy of a shared_ptr seems silly to me.

rhaschke:
> The copy of a shared_ptr is the cost of making a copy of a pointer on the stack and incrementing a counter. 

That's not the full truth: Incrementing the counter involves locking a mutex to guarantee atomic access. This is not even deterministic. For this reason, copying a `shared_ptr` is rather costly.

tylerjw:
In most implementations of the std lib it uses atomic operations instead of a mutex.  Only on really old library implementations and architectures without support for atomic operations does it use a mutex.

v4hn:
In general I would argue that we might want to consider actually returning references *to objects* instead of shared pointers, if the pointer is not needed for other APIs down the call hierarchy.
But this would be such a big API change that we probably do not care either...

rhaschke:
> In general I would argue that we might want to consider actually returning references *to objects* instead of shared pointers.

This doesn't make sense to me at all, because then you cannot get shared ownership onto this object anymore. Having a reference to the shared_ptr, you are free to create your local copy and thus increase the use count.

v4hn:
> because then you cannot get shared ownership onto this object anymore

Of course. This change only makes sense to declare that you *do not* claim ownership. Especially for the robot model, I believe there are a lot of places in the API where a reference-to-object would be the better choice.
Passing pointers for use should always require nullptr-checks in the called function, and we don't do that in all public APIs either.

tylerjw:
> In general I would argue that we might want to consider actually passing references *to objects* instead of shared pointers, if the pointer is not needed for other APIs down the call hierarchy. But this would be such a big API change that we probably do not care either...

I am generally taken back at how many shared_ptr objects there are in moveit.  It makes so there is so many places where objects can easily retain state and extend the lifetime of any object and unless you fully understand the code inside the parts of the library you are using it is not practical to use in threaded application.  

After reading more I have come around to thinking that cost references to shared pointers are a fine interface decision.  I'm not sure that shared_ptr for every object is though.

PR #1734, opened 2019-11-15
Fix travis: suppress catkin_lint error for eigenpy

rhaschke:
See https://github.com/fkie/catkin_lint/issues/73

### DIFF
--- a/moveit_ros/planning_interface/CMakeLists.txt
+++ b/moveit_ros/planning_interface/CMakeLists.txt
@@ -8,6 +8,8 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
+#avoid https://github.com/fkie/catkin_lint/issues/73
+#catkin_lint: ignore_once no_catkin_component
 find_package(catkin REQUIRED COMPONENTS
   moveit_msgs
   moveit_ros_planning


rhaschke:
This is not related to Noetic. We can remove it if catkin_lint relaxes the check.
If @roehling decides not to relax the check, we need to find_package(eigenpy) on its own and consider the corresponding `_INCLUDE_DIRS` and `_LIBRARIES` variables explicitly. Currently, they just become part of the corresponding `catkin_*` variables.
I primarily filed this to fix our Travis, particularly in view of WMD next week.

v4hn:
> we need to find_package(eigenpy) on its own and consider the corresponding _INCLUDE_DIRS and _LIBRARIES variables explicitly

And that's the way I would like to handle this in the future, as I agree that this should not be a catkin component if it's not built through catkin. Thus the request for the TODO.

Anyway, as we plan to remove the dependency in the future in favor of boost_python anyway, I'm also fine with this as-is until then.

rhaschke:
If you agree that eigenpy shouldn't be a catkin component if it's not building via catkin, we should directly prefer a cleaner (but more laborious) solution as suggested in https://github.com/ros-planning/moveit/pull/1734#issuecomment-554328047.
IMHO, if a package provides a package.xml it should be accepted as a proper ROS package that can be imported via find_package(catkin ...). Let's wait for feedback from @roehling.

v4hn:
> if a package provides a package.xml it should be accepted as a proper ROS package that can be imported via find_package(catkin ...).

I did not look at the actual implementation of the catkin-compatible cmake config.
If it is actually merged (and released) in upstream sources, I'm fine with your solution.
The use-case I have in mind is other linux distributions which would package eigenpy as a ROS-agnostic library from upstream sources.

PR #1737, opened 2019-11-17
Fix Travis: treat eigenpy as a system dependency

rhaschke:
Alternative to #1734.

### DIFF
--- a/moveit_ros/planning_interface/CMakeLists.txt
+++ b/moveit_ros/planning_interface/CMakeLists.txt
@@ -19,7 +19,6 @@ find_package(catkin REQUIRED COMPONENTS
   tf2_eigen
   tf2_geometry_msgs
   tf2_ros
-  eigenpy
   roscpp
   actionlib
   rospy
@@ -45,6 +44,7 @@ find_package(Boost REQUIRED COMPONENTS
   thread
 )
 find_package(Eigen3 REQUIRED)
+find_package(eigenpy REQUIRED)
 
 set(THIS_PACKAGE_INCLUDE_DIRS
   py_bindings_tools/include
@@ -81,6 +81,7 @@ include_directories(${THIS_PACKAGE_INCLUDE_DIRS}
 
 include_directories(SYSTEM
                     ${EIGEN3_INCLUDE_DIRS}
+                    ${eigenpy_INCLUDE_DIRS}
                     ${PYTHON_INCLUDE_DIRS})
 
 add_subdirectory(py_bindings_tools)
--- a/moveit_ros/planning_interface/move_group_interface/CMakeLists.txt
+++ b/moveit_ros/planning_interface/move_group_interface/CMakeLists.txt
@@ -6,7 +6,7 @@ target_link_libraries(${MOVEIT_LIB_NAME} moveit_common_planning_interface_object
 add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
 
 add_library(${MOVEIT_LIB_NAME}_python src/wrap_python_move_group.cpp)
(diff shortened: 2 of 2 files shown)


lepalom:
I'm sorry but I don't understand this push and all the eigenpy in moveit. I'm trying to build moveit in debian buster and it fails because it doesn't find eigenpy:

find_package(eigenpy REQUIRED)

AFAIK this CMake line implies that or exists a "Findeigenpy.cmake" or eigenpy has installed  eigenpyConfig.cmake or eigenpy-config.cmake files.

Checking eigenpy files and my installation eigenpy doesn't install that files. So, how do you solve it?

rhaschke:
@lepalom, please indicate next time that you are using Noetic. Did you installed the ROS eigenpy package: `ros-noetic-eigenpy`? This one adds the cmake config files.

lepalom:
Nops! I'm not using Noetic. I have a list of packages (melodic version) that I want to use in my debian buster, so I prepare to build them.

However, I have detected that I carried some old version 1.3.2 of eigenpy so the errors come from there. Updating it, I found eigenpy 2.3.1-1 for melodic that install the cmake files.

Sorry for the noise. But building from source a complete ros distro is not an easy task.

gavanderhoorn:
>  But building from source a complete ros distro is not an easy task.

That's why we typically try to avoid it ;)

lepalom:
Better if I don't explain to do it also with Python 3 _only_  :-(

v4hn:
> building from source a complete ros distro ... Python 3 _only_  :-(

I do that on a more or less biweekly basis for a non-Debian system
and it works quite well with the occational patch.

Hang in there, it's quite straightforward once you figured out a few things. :)

lepalom:
> it works quite well with the occational patch.

changing all the shebangs?

> Hang in there, it's quite straightforward once you figured out a few things. :)

Please, could you give me that few things ... I'm using a Debian system

v4hn:
I don't know about debian, but my `/usr/bin/python` points to 3.8.2. 
There shouldn't be the need to change any shebang (unless you find one hardcoded to python2) if you build the core system (including the catkin packages) with `export ROS_PYTHON_VERSION=3`

lepalom:
Well, that was one solution. But I was surfing a bit on the web and looking [https://www.python.org/dev/peps/pep-0394/] I understood that that was not a good idea.

In any case, my main problems come from old packages that had problems with print " and except with ,

Another thing is strange behaviours.

We have in debian the core system in python3 and not python2.

v4hn:
> and looking [https://www.python.org/dev/peps/pep-0394/] I understood that that was not a good idea.

I don't follow.

> In any case, my main problems come from old packages that had problems with print " and except with ,

Yes, there is still a lot of migration work left to do...
Please file pull-request for as many of these issues as possible because for noetic we *need* them.

> We have in debian the core system in python3 and not python2.

Now that's a debian-specific issue. ;)

lepalom:
addendum 
[https://blogs.gnome.org/mcatanzaro/2018/02/16/on-python-shebangs/]

lepalom:
addedum https://blogs.gnome.org/mcatanzaro/2018/02/16/on-python-shebangs/

v4hn:
At least for scripts [catkin_install_python](http://docs.ros.org/melodic/api/catkin/html/howto/format2/installing_python.html) takes care of adding the correct version in the shebang line.

Aside from that I do very much hope that there is no need for an API-incompatible python4 ever and python2 dies in the near future. So I do not see a reason to mark `python3` by hand of a programmer.

> ROS is too much Ubuntu centric.

Tell me about it, I've been fighting that fight for 6 years now and believe me, it's a stroll through the park these days to get things to compile on non-ubuntu...

Issue #1736, opened 2019-11-15
Fix setup assistant


v4hn:
With a number of recent changes to the code base, new bugs were introduced that make the setup assistant hard to use in some cases.

This is a meta-issue to collect individual issues.

To contribute to this issue, go through the motions to create a new moveit_config package for a urdf/xacro you have around or look at your existing package through `roslaunch <robot>_moveit_config setup_assistant.launch` and try to perform reasonable adjustments.
@henningkayser and I encountered a number of issues with the custom [panda_env](https://github.com/henningkayser/ROS-Industrial_EU_Fall19_MoveIt) robot we used in a workshop recently).

- [ ] When loading the urdf / the package the robot does not appear on the right side until one tries to add a Robot Pose. This break visualization of every step before that.
- [ ] I seem to recall adding two named robot poses reset some of the joints of the first one (???)
- [ ] The `ros_controllers.yaml` file deals with `ros_control` controller configuration, but somehow got mixed up with the `controllers.yaml`, which describes static ROS interfaces for sending trajectories
- I seem to recall some complains that some configuration files were overwritten unconditionally.
- ... 

No guarantee on any of these issue, I listed those off the top of my head. :man_shrugging:

gavanderhoorn:
> I seem to recall some complains that some configuration files were overwritten unconditionally.

Perhaps #1680 and #1679?

rhaschke:
> When loading the urdf / the package the robot does not appear on the right side

I performed a `git bisect` for this issue and figured out that the breaking commit is b0f01b2dc79f99f5650933834c8e50905bdbe168, which was part of #1625.

PR #1738, opened 2019-11-18
Fixup #1625: in MSA explicitly call setVisible(true)

rhaschke:
... to actually show the robot of the RobotStateDisplay (which is hidden initially since #1625).
Fixes b0f01b2dc79f99f5650933834c8e50905bdbe168.
Addresses https://github.com/ros-planning/moveit/issues/1736#issuecomment-554953490.

### DIFF
--- a/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
+++ b/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
@@ -341,6 +341,11 @@ void RobotStateDisplay::unsetLinkColor(const std::string& link_name)
   unsetLinkColor(&robot_->getRobot(), link_name);
 }
 
+void RobotStateDisplay::setVisible(bool visible)
+{
+  robot_->setVisible(visible);
+}
+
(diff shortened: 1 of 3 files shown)


moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
```diff
@@ -416,7 +421,10 @@ void RobotStateDisplay::update(float wall_dt, float ros_dt)
   if (load_robot_model_)
   {
     loadRobotModel();
+    // The following call to changedRobotStateTopic() should not change visibility
+    bool visible = robot_->isVisible();
     changedRobotStateTopic();
+    robot_->setVisible(visible);
```
v4hn: Why is that needed? What might change the visibility when changing the topic?

rhaschke: `changedRobotStateTopic()` resets the robot to invisible to indicate that we didn't receive a RobotState message on the (new) topic yet. However, here we want to keep the current visibility state.

PR #1741, opened 2019-11-20
Wait for done-callback

### LINKED ISSUE #1493: Possible Threading error during trajectory execution
### Description

During Execution of one Trajectory part, my controller handle is sometime still reporting status running after the the `handle->waitForExecution();` Already returned.

From looking at the code, waitForExecution should only return after getting the result and setting the status. But i am getting the following logging output:

```
Stopped trajectory monitor
Completed trajectory execution with status RUNNING ...
Controller torso_controller is done with state SUCCEEDED:
Controller torso_controller successfully finished
Controller gripper_controller is done with state SUCCEEDED:
Controller handle gripper_controller reports status RUNNING
Controller gripper_controller successfully finished
Controller arm_controller is done with state SUCCEEDED:
Controller arm_controller successfully finished
```

My bags only record one result message per goal (success) so i am not really sure what is causing the state to be set to RUNNING.

```
topics:      /torso_controller/follow_joint_trajectory/feedback    74 msgs    : control_msgs/FollowJointTrajectoryActionFeedback
             /torso_controller/follow_joint_trajectory/goal         3 msgs    : control_msgs/FollowJointTrajectoryActionGoal    
             /torso_controller/follow_joint_trajectory/result       3 msgs    : control_msgs/FollowJointTrajectoryActionResult  
             /torso_controller/follow_joint_trajectory/status     303 msgs    : actionlib_msgs/GoalStatusArray
```

This "_fix_" lets me work around this completing the whole trajectory https://github.com/LeroyR/moveit/commit/8a99875fab981a73459ababe91911f17ee246d5b 

### Your environment
* ROS Distro: Kinetic
* OS Version: Ubuntu 16.04
* Source build, Master

jschleicher:
Fixes #1493

### Description

After `waitForResult` returned, we cannot assume that the done callback has been called already. This PR still uses the `waitForResult` function, that has sophisticated boost-wait-conditions inside, but adds a while-loop afterwards waiting for the done-flag to be set within our callback routine.

Please cherry-pick to melodic.

### DIFF
--- a/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
+++ b/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
@@ -128,7 +128,7 @@ public:
     if (controller_action_client_ && !done_)
     {
       bool result = controller_action_client_->waitForResult(timeout);
-      while(ros::ok() && !done_ && (timeout.isZero() || ros::Time::now() < end_time))
+      while (ros::ok() && !done_ && (timeout.isZero() || ros::Time::now() < end_time))
         ros::Duration(0.0001).sleep();  // wait for done callback (race-condition)
       return result && (timeout.isZero() || ros::Time::now() < end_time);
     }
--- a/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
+++ b/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
@@ -128,9 +128,13 @@ public:
     if (controller_action_client_ && !done_)
     {
       bool result = controller_action_client_->waitForResult(timeout);
+#if 1  // TODO: remove when https://github.com/ros/actionlib/issues/155 is fixed
+      // workaround for actionlib issue: waitForResult() might return before our doneCB finished
       while (ros::ok() && !done_ && (timeout.isZero() || ros::Time::now() < end_time))
-        ros::Duration(0.0001).sleep();  // wait for done callback (race-condition)
-      return result && (timeout.isZero() || ros::Time::now() < end_time);
+        ros::Duration(0.0001).sleep();  // thus check the done_ flag explicitly
+      result = result && (timeout.isZero() || ros::Time::now() < end_time);
+#endif
+      return result;
     }
     return true;
   }
--- a/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
+++ b/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
@@ -124,8 +124,14 @@ public:
 
   bool waitForExecution(const ros::Duration& timeout = ros::Duration(0)) override
   {
+    ros::Time end_time = ros::Time::now() + timeout;
     if (controller_action_client_ && !done_)
-      return controller_action_client_->waitForResult(timeout);
+    {
+      bool result = controller_action_client_->waitForResult(timeout);
+      while(ros::ok() && !done_ && (timeout.isZero() || ros::Time::now() < end_time))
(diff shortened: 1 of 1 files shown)


moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
```diff
@@ -124,8 +124,14 @@ class ActionBasedControllerHandle : public ActionBasedControllerHandleBase
...
   bool waitForExecution(const ros::Duration& timeout = ros::Duration(0)) override
   {
+    ros::Time end_time = ros::Time::now() + timeout;
     if (controller_action_client_ && !done_)
-      return controller_action_client_->waitForResult(timeout);
+    {
+      bool result = controller_action_client_->waitForResult(timeout);
+      while (ros::ok() && !done_ && (timeout.isZero() || ros::Time::now() < end_time))
+        ros::Duration(0.0001).sleep();  // wait for done callback (race-condition)
+      return result && (timeout.isZero() || ros::Time::now() < end_time);
```
mikeferguson: What's the reasoning behind the logic here? why include timeout.isZero()?

v4hn:
Could you please rework the logic to use a signal instead of busy waiting?

rhaschke:
I don't think actionlib has an issue here: https://github.com/ros/actionlib/pull/156#pullrequestreview-320843673.
I will try to have a look into this later this evening.

moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
jschleicher: If timeout isn't set, the thread should wait forever.

jschleicher:
> Could you please rework the logic to use a signal instead of busy waiting?

Since the done-callback is really short, I didn't build any singal in here: https://github.com/ros-planning/moveit/blob/fcf35c803c4380bc62ef1c3c087eb282d876c773/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h#L163

Imho it should be enough, to let the scheduler do a single context switch (sleep() calls `nanosleep` which suspends the thread) to finish the callback and set `done_`. The better fix would be inside actionlib, which already has boost::conditions to wait for but wake them up which allows this race condition...

To sum up: I'd suggest this (a bit hacky) fix for melodic and do a proper rework in actionlib for noetic.

v4hn:
> To sum up: I'd suggest this (a bit hacky) fix for melodic and do a proper rework in actionlib for noetic.

Sounds acceptable to me, as long as a patch is provided (and merged...) for actionlib that resolves the issue there.

rhaschke:
As this should act as a temporary fix only, and since the done-callback is really short, I suggest dropping the timeout handling as well - as proposed in the alternative PR #1783 (I was not able to push to the present PR branch, which is on an organizational fork).
As this work attributes to @jschleicher, I kept his name as the author of the commit.

v4hn:
@rhaschke I disagree with your proposal to drop the timeout handling.
I've debugged more pleasant issues in my life. If your motivation is to force people to fix the upstream issue, it's a rather bad idea still :stuck_out_tongue: 

@jschleicher please add [the TODO comment](https://github.com/ros-planning/moveit/pull/1783/files#diff-225e7027e7c90abab783d52ab34d01acR129) proposed by @rhaschke to this request.

It appears we have multiple maintainers who agree to merge this now. :)

rhaschke:
@v4hn My motivation is to keep this quick fix as simple as possible. In a year, when the upstream issue in actionlib might be fixed (:wink:), nobody remembers what to exactly remove again. With my proposed PR this becomes immediately clear. Functionally, waiting a few milliseconds more or less doesn't matter.

jschleicher:
:+1: For keeping the fix obvious by introducing more comments.

Let's see if travis is fine...

rhaschke:
Fixed via #1783. Thanks again for your effort @jschleicher!

Issue #1735, opened 2019-11-15
MoveGroupCapability to publish frames to TF


v4hn:
As @rhaschke stated [in a related issue](https://github.com/ros-planning/moveit/issues/1122#issuecomment-433713923)

> However, it's not the task of move_group to publish [the PlanningScene frames]

In general, I strongly agree that the `move_group` node should not by default publish it's current frame information to TF. This would be partially redundant and would would clutter the TF system for no reason.

Still, I would like to see an optional `MoveGroupCapability` plugin, that *can* publish the information in that way, i.e., loop and publish a TF for each frame in the current `PlanningScene`.

- To avoid (hopefully) redundant information, It might make sense to prefix all published frames with something like `mg/`.
- To avoid unnecessary traffic on TF, it might make sense to enable/disable it by a ROS service call. Enabling/Disabling it by adding it to the ROS parameter to load on startup is always possible, but maybe not flexible enough here.

A first version for such a capability can be implemented in one day.

v4hn:
Implemented in #1761 and #1792

PR #1761, opened 2019-11-21
move_group capability for publishing planning scene frames to the tf system

JonasTietz:
### Description

This PR points to issue #1735 . I have implemented a move_group capability, which publishes frames of planning scene objects to the tf system.

### DIFF
--- a/moveit_ros/move_group/default_capabilities_plugin_description.xml
+++ b/moveit_ros/move_group/default_capabilities_plugin_description.xml
@@ -66,7 +66,7 @@
     </description>
   </class>
   
-  <class name="move_group/TfPublisher" type="move_group::MoveGroupTfPublisher" base_class_type="move_group::MoveGroupCapability">
+  <class name="move_group/TfPublisher" type="move_group::TfPublisher" base_class_type="move_group::MoveGroupCapability">
     <description>
       Provide a capability that publishes PlanningScene frames to the tf system
     </description>
--- a/moveit_ros/move_group/default_capabilities_plugin_description.xml
+++ b/moveit_ros/move_group/default_capabilities_plugin_description.xml
@@ -65,5 +65,11 @@
       Provide a ROS service that allows for clearing the octomap
     </description>
   </class>
+  
+  <class name="move_group/MoveGroupTfPublisher" type="move_group::MoveGroupTfPublisher" base_class_type="move_group::MoveGroupCapability">
+    <description>
+      Provide a capability that publishes PlanningScene frames to the tf system
+    </description>
+  </class>
 
 </library>
--- a/moveit_ros/move_group/default_capabilities_plugin_description.xml
+++ b/moveit_ros/move_group/default_capabilities_plugin_description.xml
@@ -66,7 +66,7 @@
     </description>
   </class>
   
-  <class name="move_group/MoveGroupTfPublisher" type="move_group::MoveGroupTfPublisher" base_class_type="move_group::MoveGroupCapability">
+  <class name="move_group/TfPublisher" type="move_group::MoveGroupTfPublisher" base_class_type="move_group::MoveGroupCapability">
     <description>
       Provide a capability that publishes PlanningScene frames to the tf system
     </description>
--- a/moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
@@ -84,8 +84,9 @@ void MoveGroupTfPublisher::initialize()
 {
   ros::NodeHandle nh = ros::NodeHandle("~");
 
-  prefix_ = nh.getNamespace() + "/";
+  std::string prefix = nh.getNamespace() + "/";
   nh.param("planning_scene_frame_publishing_rate", rate_, 10);
+  nh.param("planning_scene_tf_prefix", prefix_, prefix);
 
   ROS_INFO("Initializing MoveGroupTfPublisher with a frame publishing rate of %d", rate_);
   std::thread publisher_thread(&MoveGroupTfPublisher::publishPlanningSceneFrames, this);
--- a/moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
@@ -1,0 +1,83 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2012, Willow Garage, Inc.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of Willow Garage nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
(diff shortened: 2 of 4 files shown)


moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -0,0 +1,83 @@
...
+  ros::Rate rate(10);
+  std::string prefix = "move_group/";
+
+  while (ros::ok())
+  {
+    planning_scene_monitor::LockedPlanningSceneRO locked_planning_scene(context_->planning_scene_monitor_);
+    collision_detection::WorldConstPtr world = locked_planning_scene->getWorld();
+    std::string planning_frame = locked_planning_scene->getPlanningFrame();
+
+    for (auto obj = world->begin(); obj != world->end(); ++obj)
```
rhaschke: Instead of busy publishing at a fixed rate, we might want to publish changes only.

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -0,0 +1,83 @@
...
+    {
+      tf::poseEigenToTF(obj->second->shape_poses_[0], transform);
```
rhaschke: It's a pity that all transforms are published w.r.t. planning_frame only. I guess the original information is lost.
What about sub frames?

v4hn:
i guess this is still work-in-progress, although it's not marked as such.
thanks for the review, I did not actually look at it before.

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -0,0 +1,83 @@
...
+void MoveGroupTfPublisher::publishPlanningSceneFrames()
+{
+  tf::TransformBroadcaster broadcaster;
+  tf::Transform transform;
+  ros::Rate rate(10);
+  std::string prefix = "move_group/";
+
+  while (ros::ok())
+  {
+    planning_scene_monitor::LockedPlanningSceneRO locked_planning_scene(context_->planning_scene_monitor_);
```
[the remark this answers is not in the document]
v4hn: it's RO, so planning should still work and the object gets reconstructed between iterations, right?
but the lock should definitely not be held during the sleep, yes.

moveit_ros/move_group/default_capabilities_plugin_description.xml
```diff
@@ -65,5 +65,11 @@
       Provide a ROS service that allows for clearing the octomap
     </description>
   </class>
+  
+  <class name="move_group/MoveGroupTfPublisher" type="move_group::MoveGroupTfPublisher" base_class_type="move_group::MoveGroupCapability">
```
v4hn: no need for the "MoveGroup" prefix anywhere. we stopped using the prefix with newly added capabilities.

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -0,0 +1,83 @@
...
+MoveGroupTfPublisher::MoveGroupTfPublisher() : MoveGroupCapability("TfPublisher")
+{
+}
+
+void MoveGroupTfPublisher::publishPlanningSceneFrames()
+{
+  tf::TransformBroadcaster broadcaster;
+  tf::Transform transform;
+  ros::Rate rate(10);
+  std::string prefix = "move_group/";
```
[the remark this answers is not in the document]
v4hn: I'd prefer to keep the prefix configurable, but the namespace seems a reasonable default

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
v4hn: transforms are stamped, and become outdated in tf, so republishing is important. but the rate could be a ros param

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+    rate.sleep();
+  }
+}
+
+void MoveGroupTfPublisher::initialize()
+{
+  ROS_INFO("Initializing MoveGroupTfPublisher");
+  std::thread publisher_thread(&MoveGroupTfPublisher::publishPlanningSceneFrames, this);
+  publisher_thread.detach();
```
v4hn: I'm not perfectly happy with detaching, but it seems feasible here

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
felixvd: Also, what about CollisionObjects that are attached to the robot? I don't see them published in this file and I don't believe the robot state publishes them either.

Seconding the question about subframes, and agree that parent link information would be interesting. Nothing we can do about it for now I suppose.

felixvd:
Thanks for tackling this!

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -51,20 +51,21 @@ void MoveGroupTfPublisher::publishPlanningSceneFrames()
...
-          tf::StampedTransform(transform, ros::Time::now(), planning_frame, prefix + obj->second->id_));
+      planning_scene_monitor::LockedPlanningSceneRO locked_planning_scene(context_->planning_scene_monitor_);
+      collision_detection::WorldConstPtr world = locked_planning_scene->getWorld();
+      std::string planning_frame = locked_planning_scene->getPlanningFrame();
+
+      for (auto obj = world->begin(); obj != world->end(); ++obj)
+      {
+        tf::poseEigenToTF(obj->second->shape_poses_[0], transform);
+        broadcaster.sendTransform(
+            tf::StampedTransform(transform, ros::Time::now(), planning_frame, prefix_ + obj->second->id_));
```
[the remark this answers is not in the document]
rhaschke: I agree. The clean approach would be to keep the thread instance as a member variable of this capability and
stop+join the thread in the destructor of the class.

moveit_ros/move_group/default_capabilities_plugin_description.xml
```diff
@@ -66,7 +66,7 @@
     </description>
   </class>
   
-  <class name="move_group/MoveGroupTfPublisher" type="move_group::MoveGroupTfPublisher" base_class_type="move_group::MoveGroupCapability">
+  <class name="move_group/TfPublisher" type="move_group::MoveGroupTfPublisher" base_class_type="move_group::MoveGroupCapability">
```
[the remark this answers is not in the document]
JonasTietz: I could definitly do that. But I think renaming the other capabilities should go in another PR.

v4hn:
> I could definitly do that. But I think renaming the other capabilities should go in another PR.

I did not ask you to rename the *other* capabilities. I asked you to remove the prefix from the C++ class.

JonasTietz:
> I could definitly do that. But I think renaming the other capabilities should go in another PR.
> I did not ask you to rename the *other* capabilities. I asked you to remove the prefix from the C++ class.

Ahh. Ok I misunderstood that.

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+namespace move_group
+{
+TfPublisher::TfPublisher() : MoveGroupCapability("TfPublisher")
+{
+}
+
+TfPublisher::~TfPublisher()
+{
+  thread_.join();
```
rhaschke: You rely on `ros::ok()` to return false to finish the thread. Please add a stop_ flag to explicitly stop on destruction, which might be/become unrelated to ROS shutdown.

v4hn:
> Please re-add `ros::ok()`: If ROS dies for any reason, the thread needs to be finished as well.

I'm not sure I get your reasoning here.
I thought you explicitly want to get rid of this exit condition, as it adds an outside flag for an internally-managed thread.
If ROS dies, the `move_group` node should clean up and delete the capability, thus setting the flag and joining the thread.

I do not see why you would want to add the check back.

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
rhaschke: Attached objects are not yet addressed, are they?

felixvd: Not that I can see

v4hn:
Congrats @JonasTietz and thank you for this nice contribution!

v4hn:
> Attached objects are not yet addressed, are they?

Very true!

Sorry, I merged after I got CI to accept it and did not see your comment.

@JonasTietz Could you have a look at attached objects in a follow-up request? :)
Attached objects are stored in the `RobotState` of the `PlanningScene` and can be accessed via `RobotState::getAttachedBodies`.

felixvd:
Please also add subframes while you're at it. They are part of the CollisionObject.

rhaschke:
subframes are already published for CollisionObjects, but I think attached collision objects also have subframes.

JonasTietz:
> @JonasTietz Could you have a look at attached objects in a follow-up request? :) Attached objects are stored in the `RobotState` of the `PlanningScene` and can be accessed via `RobotState::getAttachedBodies`.

I will take a look at it. :)

PR #1763, opened 2019-11-21
jog_arm: add a C++ API

AndyZe:
Adds a C++ API. There's an example in src/cpp_interface_example


henningkayser:
@AndyZe Thanks for preparing this! I'll give this a test but so far the implementation seems very reasonable to me. `clang-tidy` doesn't like capitalized function names, could you fix this?

henningkayser:
@AndyZe thanks for fixing this up, CI fails unrelated to this PR!

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_interface_base.h
```diff
@@ -0,0 +1,81 @@
+/*
```
henningkayser: I'd prefer the license headers to match the format of the other sources in the repo.

henningkayser:
Looks great! lgtm

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_interface_base.h
AndyZe: Can I do this in a follow-up PR? We just merged an update to the license formatting that's going to conflict -- https://github.com/ros-planning/moveit/pull/1764

AndyZe: done!

moveit_experimental/moveit_jog_arm/CMakeLists.txt
```diff
@@ -3,6 +3,8 @@ project(moveit_jog_arm)
 
 add_compile_options(-std=c++14)
```
mamoll: Replace with this to be compiler agnostic:
```
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
```

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_interface_base.h
```diff
@@ -0,0 +1,80 @@
...
+  bool readParameters(ros::NodeHandle& n);
+
+  robot_model_loader::RobotModelLoaderPtr model_loader_ptr_;
+
+  // Store the parameters that were read from ROS server
+  JogArmParameters ros_parameters_;
+
+  // Share data between threads
+  JogArmShared shared_variables_;
+  pthread_mutex_t shared_variables_mutex_;
```
mamoll: Is it really necessary to use pthread? Using mutexes from the standard C++ library would be preferable (https://en.cppreference.com/w/cpp/thread). This will matter when running on non-POSIX systems.

AndyZe: It looks like the std::mutex api is a little easier to use, too.

PR #1765, opened 2019-11-22
WIP: Enable codecov.io

rhaschke:
Alternative to #1719, not requiring to touch cmake files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
     - WARNINGS_OK=false
   matrix:
     - TEST="clang-format catkin_lint"
+    - TEST=code-coverage
     - ROS_DISTRO=melodic
     - ROS_DISTRO=kinetic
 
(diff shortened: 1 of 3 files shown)


rhaschke:
Hi @tylerjw,

based on your great work in https://github.com/ros-planning/moveit_ci/pull/91 and #1719, I experimented a little bit with an alternative solution. I came up with: https://github.com/rhaschke/moveit_ci/tree/codecov, which doesn't require changes to the cmake files. I already tested this on [rviz](https://travis-ci.com/rhaschke/rviz/jobs/259278374).
This PR aims to test it for MoveIt.

As suggested in https://github.com/ros-planning/moveit/pull/1719#issuecomment-556073142, I tried to keep optimization on to avoid the (still unexplained?) slowdown due to the debug build.
The main advantage, in my point of view, is that we don't need to adapt the cmake files.

tylerjw:
Looking at the codecov.io report it looks like it is reporting from the catkin workspace directory (there are files from `devel` directory in there).  Also, in src there is only `geometric_shapes`.  Why doesn't this give a coverage report of the other directories.

rhaschke:
> Looking at the codecov.io report it looks like it is reporting from the catkin workspace directory.

I'm afraid this is some artifact from early tests (codecov.io not corretly updating?).
Looking at the [output of `lcov --list coverage.info`](https://travis-ci.org/ros-planning/moveit/jobs/615560610#L1611) you see that only relevant files of the MoveIt repository are reported.

rhaschke:
I will close this PR and reopen to trigger a fresh codecov report.

PR #1766, opened 2019-11-22
Show attached bodies trail in RViz

Jntzko:
### Description
Attached bodies were not included in the trail visualization of a trajectory, changing the type from rviz::Robot to RobotStateVisualization solves this issue (#190).

Anyway, the bodies color is always green, I didn't find a proper way to get the color of the attached object and set it in the update method.

### Checklist

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
@@ -202,7 +202,7 @@ void TrajectoryVisualization::reset()
 void TrajectoryVisualization::clearTrajectoryTrail()
 {
   for (std::size_t i = 0; i < trajectory_trail_.size(); ++i)
-    delete trajectory_trail_[i];
+    trajectory_trail_[i].reset();
   trajectory_trail_.clear();
 }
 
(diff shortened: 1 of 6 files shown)


moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
```diff
@@ -202,7 +202,7 @@ void TrajectoryVisualization::reset()
 void TrajectoryVisualization::clearTrajectoryTrail()
 {
   for (std::size_t i = 0; i < trajectory_trail_.size(); ++i)
-    delete trajectory_trail_[i];
+    trajectory_trail_[i].reset();
```
v4hn: There's no need for this loop anymore. shared_ptr are cleared implicitly in the next line.

rhaschke:
> Objects only have colors in the PlanningScene due to the implementation and changing this is outside the scope of this patch.

That's not perfectly true. There is a `default_attached_body_color_`, both in the PlanningSceneDisplay and the TrajectoryVisualization (the former forwards the color value to the latter). This color should be used for the trail as well. I will push corresponding changes immediately.

v4hn:
> That's not perfectly true.

true! thanks for adding patches for that!

PR #1768, opened 2019-11-23
fixing CADB loading bug

mlautman:
This fixes a small bug introduced in https://github.com/ros-planning/moveit/pull/1428 where loadConstraintDatabase would load the parameter from rosparam and then try to recursively call itself.


rhaschke:
Travis error is due to an unrelated, flaky test. Thus merged nonetheless.

BryceStevenWilley:
Sorry about that, I'll have to more thoroughly test next time. Thanks for the fix @mlautman

PR #1769, opened 2019-11-23
Fix python PlanningSceneInterface's attach_*()

rhaschke:
The attach_*() functions didn't handle the synchronous service communication yet,
triggering an exception when serializing a CollisionObject instead of an AttachedCollisionObject.


v4hn:
looks like you merged this yourself after 5 hours on a Saturday.

the patch exploits python type checks without checking for AttachedCoisionObject too. 
even though the submit function, one level higher maintains the information in a boolean flag.

v4hn:
also, it seems you did not port this forward to master yet?

rhaschke:
This was a trivial patch. So I merged.
I tried to forward port to master, but noticed (only then) that master already had a fix for this - utilizing the `attach` flag :smile:

PR #1772, opened 2019-11-24
rviz panel: pass <current> start state as empty to move_group

rhaschke:
... to ensure that the most recent current state is actually used for planning

When testing #1766 I observed several times that planning, triggered from, rviz wasn't using the most recent RobotState (where I just added an attached object).
The reason was the following:
- The rviz plugin queried the current state on startup.
- Then I added an attached object, which was correctly reflected by the PlanningSceneDisplay.
However, the start state was still referring to the original state at startup.
- Planning didn't use the current robot state, but the one saved at some point in the past.
- I was wondering, why the attached object wasn't considered during planning.

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
@@ -172,6 +172,9 @@ void MotionPlanningFrame::computePlanButtonClicked()
   ui_->result_label->setText("Planning...");
 
   configureForPlanning();
+  // move_group node uses an empty start state to refer to the most recent current state
+  if (ui_->start_state_combo_box->currentText() == "<current>")
+    move_group_->setStartStateToCurrentState();
   planning_display_->rememberPreviousStartState();
   bool success = (ui_->use_cartesian_path->isEnabled() && ui_->use_cartesian_path->checkState()) ?
                      computeCartesianPlan() :


felixvd:
Just for clarification, I don't see where a start state that was passed before is now empty. I just see an explicit call to the move_group to update the start state. I think this will work fine, but the comment and the name of the commit confuse me. It sounds like the name of the branch or "Force update before planning when start state set to <current> in Rviz" might be a more appropriate name for the change?

As an aside, I remember a ROSCon talk about a package to do GUI tests as the procedure you describe.

rhaschke:
The call to `setStartStateToCurrentState()` actually [resets the start state](https://github.com/ros-planning/moveit/blob/11c82ece67302e9190bfa45418ee5e162d8950a2/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L363-L366) passed on to the move_group node for planning.

v4hn:
I can't merge this because CI failed for unrelated reasons and I do not have admin privileges to bypass CI.

@henningkayser @davetcoleman Please look into the test for moveit_cpp, it broke multiple times by now in unrelated pull-requests. If you can't find the problem, just disable it for now.

> Test log of build/moveit_ros_planning_interface/test_results/moveit_ros_planning_interface/rostest-test_moveit_cpp_test.xml
> 
> <?xml version="1.0" encoding="utf-8"?>
> 
> <testsuite errors="0" failures="1" name="unittest.suite.TestSuite" tests="1" time="7.350"><testcase classname="rostest.runner.RosTest" name="testmoveit_cpp_test" time="7.3493"><failure type="AssertionError">test [moveit_cpp_test] did not generate test results

PR #1773, opened 2019-11-24
fix compiler warning thoughout the code base

rhaschke:
The NodeHandle variables `nh` are never used in this function.

### DIFF
--- a/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_env_distance_field.h
+++ b/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_env_distance_field.h
@@ -105,19 +105,19 @@ public:
   void createCollisionModelMarker(const moveit::core::RobotState& state,
                                   visualization_msgs::MarkerArray& model_markers) const;
 
-  virtual double distanceSelf(const moveit::core::RobotState& state) const
+  virtual double distanceSelf(const moveit::core::RobotState& /* state */) const
   {
     return 0.0;
-  };
+  }
 
-  virtual double distanceSelf(const moveit::core::RobotState& state,
-                              const collision_detection::AllowedCollisionMatrix& acm) const
+  virtual double distanceSelf(const moveit::core::RobotState& /* state */,
+                              const collision_detection::AllowedCollisionMatrix& /* acm */) const
   {
     return 0.0;
-  };
+  }
 
-  void distanceSelf(const DistanceRequest& req, DistanceResult& res,
-                    const robot_state::RobotState& state) const override
+  void distanceSelf(const DistanceRequest& /* req */, DistanceResult& /* res */,
+                    const robot_state::RobotState& /* state */) const override
   {
     ROS_ERROR_NAMED("collision_distance_field", "Not implemented");
   }
@@ -179,17 +179,22 @@ public:
 
   virtual double distanceRobot(const robot_state::RobotState& state, bool verbose = false) const
   {
+    (void)state;
+    (void)verbose;
     return 0.0;
   }
 
   virtual double distanceRobot(const robot_state::RobotState& state, const AllowedCollisionMatrix& acm,
                                bool verbose = false) const
   {
+    (void)state;
+    (void)acm;
+    (void)verbose;
     return 0.0;
   }
 
-  void distanceRobot(const DistanceRequest& req, DistanceResult& res,
-                     const robot_state::RobotState& state) const override
+  void distanceRobot(const DistanceRequest& /* req */, DistanceResult& /* res */,
+                     const robot_state::RobotState& /* state */) const override
   {
     ROS_ERROR_NAMED("collision_distance_field", "Not implemented");
   }
@@ -263,7 +268,7 @@ protected:
   bool compareCacheEntryToAllowedCollisionMatrix(const DistanceFieldCacheEntryConstPtr& dfce,
                                                  const collision_detection::AllowedCollisionMatrix& acm) const;
 
-  void updatedPaddingOrScaling(const std::vector<std::string>& links) override{};
+  void updatedPaddingOrScaling(const std::vector<std::string>& /*links*/) override{};
 
   DistanceFieldCacheEntryWorldPtr generateDistanceFieldCacheEntryWorld();
 
--- a/moveit_core/macros/include/moveit/macros/diagnostics.h
+++ b/moveit_core/macros/include/moveit/macros/diagnostics.h
@@ -1,7 +1,7 @@
 /*********************************************************************
  * Software License Agreement (BSD License)
  *
- *  Copyright (c) 2012, Willow Garage, Inc.
+ *  Copyright (c) 2019, Bielefeld University
  *  All rights reserved.
  *
  *  Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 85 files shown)


rhaschke:
> I agree, those are unused, but in that case why not delete [this one](https://github.com/ros-planning/moveit/blob/029f239d2a3ff0a9fa0e2e5a01174f17b7267041/moveit_ros/robot_interaction/src/robot_interaction.cpp#L555) as well?

Didn't see this ;-)
I took the chance to clean up the whole codebase more thoroughly and fix all warnings suppressed so far, mostly `-Wunused-parameter`. Some warnings are due to external headers. For those, I implemented a mechanism to temporarily disable (and eventually restore) warnings.
@v4hn, please can you re-review?

moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_env_distance_field.h
```diff
@@ -179,17 +179,22 @@ class CollisionEnvDistanceField : public CollisionEnv
 
   virtual double distanceRobot(const robot_state::RobotState& state, bool verbose = false) const
   {
+    (void)state;
```
felixvd: In the other files, you commented out the parameters in the function definition (e.g. `const robot_state::RobotState& /*state*/, bool verbose = false`), only in this one it's done differently. Is there a reason?

moveit_core/macros/include/moveit/macros/diagnostics.h
```diff
@@ -0,0 +1,55 @@
...
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+#pragma once
+
+#if defined(_MSC_VER)
```
felixvd: Could you add a summary of what this file does?

moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_env_distance_field.h
rhaschke: Yes. I use three variantes:
1. Commenting
2. `(void)variable;`  (when commenting wasn't feasible due to default arguments defined)
3. Removing the variable (in cases where the type name already suggests the variable meaning, e.g. `NodeHandle`).

rhaschke:
> The name "diagnostics" is easy to mistake for a debugging or performance measuring tool or something.

This just reflects the name gcc/clang use for the pragma as well, e.g.:
`#pragma GCC diagnostic push`

moveit_core/macros/include/moveit/macros/diagnostics.h
rhaschke: Sure.

moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_env_distance_field.h
BryceStevenWilley: I'm not sure of the exact rules, but it looks like you are able to comment with default arguments defined in other files, such as https://github.com/ros-planning/moveit/pull/1773/files?file-filters[]=.cpp&file-filters[]=.h&file-filters[]=.yml#diff-22ed9d5e496d7469a03e0b414299ebdeR919.

moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
```diff
@@ -50,11 +51,14 @@
 #include <QDir>
 #include <pluginlib/class_loader.hpp>  // for loading all avail kinematic planners
 // Rviz
+DIAGNOSTIC_PUSH
+SILENT_UNUSED_PARAM
 #include <rviz/render_panel.h>
 #include <rviz/visualization_manager.h>
 #include <rviz/view_manager.h>
 #include <rviz/default_plugin/view_controllers/orbit_view_controller.h>
 #include <moveit/robot_state_rviz_plugin/robot_state_display.h>
```
BryceStevenWilley: How is the unused param calculated? We should be able to fix the warning if it's in our own file, right?

moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_env_distance_field.h
rhaschke: Of course, when the default argument is in the header, I can comment (or omit) the variable name in the .cpp.

moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
rhaschke: What do you mean by _calculated_? This specific macro disables unused parameter warnings.
For other warnings, one should add another specific macro.

BryceStevenWilley: Sorry, calculated was a poor choice of words. I mostly wanted to know if we can fix the warning for `robot_state_display.h`, or if it needs be be silenced because how it's used (or not used) in this particular file.

moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_env_distance_field.h
mamoll: I think `bool = false` or `bool /* verbose */ = false` for a default argument would also work. The latter seems clearer than the `(void)variable;` variant.

moveit_core/macros/include/moveit/macros/diagnostics.h
```diff
@@ -0,0 +1,73 @@
...
+#define DIAGNOSTIC_POP __pragma(warning(pop))
+#define SILENT_UNUSED_PARAM __pragma(warning(disable : 4100))
+
+#elif defined(__GNUC__) || defined(__clang__)
+#define DO_PRAGMA(X) _Pragma(#X)
+#define DIAGNOSTIC_PUSH DO_PRAGMA(GCC diagnostic push)
+#define DIAGNOSTIC_POP DO_PRAGMA(GCC diagnostic pop)
+
+#if defined(__clang__)
+#define SILENT_UNUSED_PARAM DO_PRAGMA(GCC diagnostic ignored "-Wunused-parameter")
```
mamoll: *clang* diagnostic ignored?

moveit_core/macros/include/moveit/macros/diagnostics.h
```diff
@@ -0,0 +1,73 @@
...
+#else
+#define SILENT_UNUSED_PARAM                                                                                            \
+  DO_PRAGMA(GCC diagnostic ignored "-Wunused-parameter")                                                               \
+  DO_PRAGMA(GCC diagnostic ignored "-Wunused-but-set-parameter")
```
mamoll: same here

PR #1777, opened 2019-11-25
Move ASSERT() into test setup function

j-petit:
### Description
As discussed in [this PR review](https://github.com/ros-planning/moveit/pull/1657#discussion_r348925925), instead of an extra test case the robot setup should be tested in each run.

### DIFF
--- a/moveit_core/planning_scene/test/test_multi_threaded.cpp
+++ b/moveit_core/planning_scene/test/test_multi_threaded.cpp
@@ -73,8 +73,7 @@ protected:
   void SetUp() override
   {
     robot_model_ = moveit::core::loadTestingRobotModel("panda");
-
-    ASSERT_TRUE(robot_model_);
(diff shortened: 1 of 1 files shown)


moveit_core/planning_scene/test/test_multi_threaded.cpp
```diff
@@ -76,6 +76,8 @@ class CollisionDetectorThreadedTest : public testing::Test
     robot_model_ok_ = static_cast<bool>(robot_model_);
```
rhaschke: Please replace this line with your assertion.
The following statements actually already rely on a valid robot_model_.

j-petit: You're right, changed it.

rhaschke:
@j-petit Could you address Travis issues, please.

j-petit:
Why does this fail for kinetic but not for melodic?

j-petit:
Thanks @rhaschke

PR #1656, opened 2019-09-02
Implement low-level interface moveit_cpp

henningkayser:
As already discussed in the last maintainer meeting we at PickNik have a need for a set of low-level interface classes that allow us to use, modify and exchange core components without the overhead 
 to adjust message types. Also, we are implementing performance-critical functions that are not throttled by the ROS ecosystem. The first MVP is a MoveGroup-like set of features that allows basic planning and execution for different planning groups. I will follow-up with more detailed explanation of the goals, requirements, roadmap and design decisions.

PR #1780, opened 2019-11-26
Fix moveitcpp unit test bug

JafarAbdi:
### Description

This PR fix ~~3~~ 2 bugs

1- As @henningkayser explained in his [comment](https://github.com/ros-planning/moveit/pull/1656#issuecomment-558271580) 

2- Seem like there's a race condition happen sometime which is fixed by defaulting both destructors
```bash
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
```
~~3- When I get a warning about dirty transform the tests segfaults~~

This will always publish an initial joint state msg even if no initial pose specified for any joint model group, by calling `robot_model->getVariableDefaultPositions(joints);`

### Checklist

### DIFF
--- a/moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
+++ b/moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
@@ -133,7 +133,7 @@ public:
     if (!node_handle_.getParam("initial", initial))
     {
       ROS_INFO_NAMED("loadInitialJointValues",
-                     "No initial pose specifid for any joint model group, using the default values for the joints");
+                     "No initial pose specified for any joint model group, using the default values for the joints");
     }
     else
     {
--- a/moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
+++ b/moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
@@ -73,13 +73,9 @@ public:
     pub_ = node_handle_.advertise<sensor_msgs::JointState>("fake_controller_joint_states", 100, latch);
 
     /* publish initial pose */
-    XmlRpc::XmlRpcValue initial;
-    if (node_handle_.getParam("initial", initial))
-    {
-      sensor_msgs::JointState js = loadInitialJointValues(initial);
-      js.header.stamp = ros::Time::now();
-      pub_.publish(js);
-    }
+    sensor_msgs::JointState js = loadInitialJointValues();
+    js.header.stamp = ros::Time::now();
+    pub_.publish(js);
 
     /* actually create each controller */
     for (int i = 0; i < controller_list.size(); ++i)  // NOLINT(modernize-loop-convert)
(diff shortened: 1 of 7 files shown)


JafarAbdi:
I should remove the `start_state->update();` it doesn't happen again after fixing the fake controller manager

JafarAbdi:
@rhaschke This's the trace when the `MoveItFakeControllerManager` publish random big values for `panda_finger_joint1` and `panda_finger_joint2`

Screenshot from 2019-11-25 16-48-53

henningkayser:
> Thanks for looking into this. I have several remarks though:
> 
> 1. It's _not_ the task of the MoveItFakeControllerManager to publish a full joint state. Real-world controllers won't do this as well. Rather, MoveItCpp should explicitly handle the situation that some joints may not have been published yet and thus are still _uninitialized_. What exactly happens to MoveItCpp (and your tests) in this case?

The problem is that the MoveItFakeControllerManager publishes a full joint state even if some joint values are undefined. In this case there is an initial state defined for the whole group (arm+hand) which only contains joint values for the arm. The fake controller initializes the specified values but then publishes values for all joints including the fingers which at this point are undefined. IMO this is a bug and 0 should be the default value.

> 2. I don't see how the default constructors avoid any race condition. The default constructor releases all the variables, much like clearContents() was doing before - maybe in another order. If the order should matter, it should be adapted in clearContents as well. In any case you should provide more details about this race conditions to allow understanding whether your modification is indeed a fix.
> 
> As fixing these tests might take some more time, I suggest to disable the test for now and thus fix Travis.

It's actually not a race condition since all joint states are available. The test fails because of random joint values being published by the fake controller. The fix just resolves this undefined behavior.


The only question for me is if we additionally want the fake controller to publish 0 values by default even if `initial` is not specified.

moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
```diff
@@ -123,67 +119,77 @@ class MoveItFakeControllerManager : public moveit_controller_manager::MoveItCont
...
     JointPoseMap joints;
 
-    for (int i = 0, end = param.size(); i != end; ++i)
+    robot_model->getVariableDefaultPositions(joints);
+
+    if (!node_handle_.getParam("initial", initial))
     {
-      try
+      ROS_INFO_NAMED("loadInitialJointValues",
+                     "No initial pose specifid for any joint model group, using the default values for the joints");
```
henningkayser:
```suggestion
                     "No initial pose specified for any joint model group, using the default values for the joints");
```

moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
```diff
@@ -123,67 +119,77 @@ class MoveItFakeControllerManager : public moveit_controller_manager::MoveItCont
...
-    }
-
+    XmlRpc::XmlRpcValue initial;
     robot_model_loader::RobotModelLoader robot_model_loader(ROBOT_DESCRIPTION);
     const robot_model::RobotModelPtr& robot_model = robot_model_loader.getModel();
     typedef std::map<std::string, double> JointPoseMap;
     JointPoseMap joints;
 
-    for (int i = 0, end = param.size(); i != end; ++i)
+    robot_model->getVariableDefaultPositions(joints);
```
henningkayser: I think I agree with @rhaschke that we should not publish the full joint state of the robot model. I would only publish values for joints that the fake controller is actually managing.

v4hn:
> The only question for me is if we additionally want the fake controller to publish 0 values by default even if initial is not specified.

This makes sense to me. Wasn't the default behavior to publish 0 for all joints if no initial state is specified?

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -124,7 +124,7 @@ class MoveItCpp
   MoveItCpp& operator=(MoveItCpp&& other);
 
   /** \brief Destructor */
-  ~MoveItCpp();
+  ~MoveItCpp() = default;
```
henningkayser: How do these changes affect the test?

henningkayser:
> > The only question for me is if we additionally want the fake controller to publish 0 values by default even if initial is not specified.
> 
> This makes sense to me. Wasn't the default behavior to publish 0 for all joints if no initial state is specified?

I thought so too, but apparently it doesn't (just tested this again). The fake controller first starts publishing when a trajectory is executed. This is also the reason why visualized arms can have weird poses initially when running the fake controller.

Also, I'm a bit confused that this issue doesn't happen on my system, but can be reproduced reliably with Travis and on @JafarAbdi's system.

JafarAbdi:
> > > The only question for me is if we additionally want the fake controller to publish 0 values by default even if initial is not specified.
> > 
> > 
> > This makes sense to me. Wasn't the default behavior to publish 0 for all joints if no initial state is specified?
> 
> I thought so too, but apparently it doesn't (just tested this again). The fake controller first starts publishing when a trajectory is executed. This is also the reason why visualized arms can have weird poses initially when running the fake controller.

I added these arguments to the test so it runs forever and I can detect when it fault `--gtest_repeat=-1 --gtest_break_on_failure`, but running `rostopic echo /moveit_cpp_test/fake_controller_joint_states`  gives weird values for the fingers even if the test doesn't fail ( when the values are big the test fails"

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
JafarAbdi: It fixes the boost throwing bug

JafarAbdi: I didn't check it again after updating the fake controller manager, I'm testing it right now locally, if there's no error I'll revert my changes

moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp
JafarAbdi: So we should move `loadInitialJointValues` after creating the controllers, right .?

rhaschke:
IMHO, the fake controllers are _not_ responsible for publishing the full robot state!
Controllers are defined _per joint group_, thus they should always publish joints from this group only.
So, the question is why does it publish those finger states?

Looking into [`fake_controllers.yaml`](https://github.com/ros-planning/panda_moveit_config/blob/4a16a4fcdbf5cc7975937faa1b0ebcf8d41b365d/config/fake_controllers.yaml), we see that an initial pose (`ready`) is defined for the JMG `panda_arm_hand`. This pose is defined in [panda_arm_hand.srdf.xacro](https://github.com/ros-planning/panda_moveit_config/blob/4a16a4fcdbf5cc7975937faa1b0ebcf8d41b365d/config/panda_arm_hand.srdf.xacro#L21-L29), but doesn't include the finger joints! **This is the real issue.**

But we can identify several further shortcomings:
1. MoveIt should complain about this incomplete group state definition.
2. Indeed the fake controllers could initialize undefined joints to their defaults.

I am preparing a corresponding PR right now.

> Also, I'm a bit confused that this issue doesn't happen on my system, but can be reproduced reliably with Travis and on Jafar's system.

This is about (random) variable initialization in release mode. If you don't use `-O3` optimization, you might not observe this bug, because the compiler initializes all variables to zero in your case. This is particularly true when compiling in debug mode.

rhaschke:
I filed a new PR: https://github.com/ros-planning/moveit/pull/1781.
You didn't yet fully answer the question in https://github.com/ros-planning/moveit/pull/1780#discussion_r350651761. If this was a separate issue, it will need a separate fix. But, as said initially, I don't think that changing to the default destructor resolves any race condition.

henningkayser:
> IMHO, the fake controllers are _not_ responsible for publishing the full robot state!
> Controllers are defined _per joint group_, thus they should always publish joints from this group only.

They are defined for arbitrary joints (doesn't need to be a group) and I think the fake controllers are responsible for publishing values for all of their joints. I agree that no other joint values should be published. To me it just seems very reasonable to always publish default values even if the initial state is not set.

> So, the question is why does it publish those finger states?
> 
> Looking into [`fake_controllers.yaml`](https://github.com/ros-planning/panda_moveit_config/blob/4a16a4fcdbf5cc7975937faa1b0ebcf8d41b365d/config/fake_controllers.yaml), we see that an initial pose (`ready`) is defined for the JMG `panda_arm_hand`. This pose is defined in [panda_arm_hand.srdf.xacro](https://github.com/ros-planning/panda_moveit_config/blob/4a16a4fcdbf5cc7975937faa1b0ebcf8d41b365d/config/panda_arm_hand.srdf.xacro#L21-L29), but doesn't include the finger joints! **This is the real issue.**
> 

Sure, the target definition is incomplete and should be fixed. I would still say that publishing undefined joint values is a bug, either in the fake controller or in the robot state initialization as you pointed out in #1781.

> But we can identify several further shortcomings:
> 
> 1. MoveIt should complain about this incomplete group state definition.
> 2. Indeed the fake controllers could initialize undefined joints to their defaults.
> 
> I am preparing a corresponding PR right now.

Thanks!

rhaschke:
> They are defined for arbitrary joints (doesn't need to be a group)

You are absolutely right! My bad.

JafarAbdi:
@rhaschke I run Travis with verbose test and this's what's failing it

```
[ WARN] [1574780475.854329208]: Returning dirty link transforms
moveit_cpp_test: /root/ros_ws/src/moveit/moveit_core/robot_state/src/robot_state.cpp:1022: const Eigen::Isometry3d &moveit::core::RobotState::getFrameInfo(const std::string &, const moveit::core::LinkModel *&, bool &) const: Assertion `checkLinkTransforms()' failed.
```
I'll debug it more when I've time today, Thanks!

rhaschke:
This refers to an not-updated RobotState. Add an update() at the _appropriate_ place(s) :wink:

henningkayser:
> @rhaschke I run Travis with verbose test and this's what's failing it
> 
> ```
> [ WARN] [1574780475.854329208]: Returning dirty link transforms
> moveit_cpp_test: /root/ros_ws/src/moveit/moveit_core/robot_state/src/robot_state.cpp:1022: const Eigen::Isometry3d &moveit::core::RobotState::getFrameInfo(const std::string &, const moveit::core::LinkModel *&, bool &) const: Assertion `checkLinkTransforms()' failed.
> ```
> 
> I'll debug it more when I've time today, Thanks!

@JafarAbdi It looks like the internal robot state of the planning scene we pass it to the planning pipeline is not updated. I don't see any other part of the code that might lead to this.

rhaschke:
Closing in favor of #1781. The present PR introduces a lot of unrelated noise.

PR #1781, opened 2019-11-26
Fix group states

rhaschke:
As pointed out in https://github.com/ros-planning/moveit/pull/1780#issuecomment-558587911,

1. The robot model loader doesn't complain about malformed group states (having missing joints).
2. FakeControllers used those uninitialized joint values for publishing.

Both issues are fixed with this PR. This should fix the flaky moveit-cpp unit test.

### DIFF
--- a/moveit_core/robot_model/src/robot_model.cpp
+++ b/moveit_core/robot_model/src/robot_model.cpp
@@ -327,6 +327,7 @@ void RobotModel::buildGroupStates(const srdf::Model& srdf_model)
     if (hasJointModelGroup(group_state.group_))
     {
       JointModelGroup* jmg = getJointModelGroup(group_state.group_);
+      std::vector<const JointModel*> remaining_joints = jmg->getActiveJointModels();
       std::map<std::string, double> state;
       for (std::map<std::string, std::vector<double>>::const_iterator jt = group_state.joint_values_.begin();
            jt != group_state.joint_values_.end(); ++jt)
@@ -335,6 +336,10 @@ void RobotModel::buildGroupStates(const srdf::Model& srdf_model)
         {
           const JointModel* jm = jmg->getJointModel(jt->first);
           const std::vector<std::string>& vn = jm->getVariableNames();
+          // Remove current joint name from remaining list.
+          auto it_found = std::find(remaining_joints.begin(), remaining_joints.end(), jm);
+          if (it_found != remaining_joints.end())
+            remaining_joints.erase(it_found);
           if (vn.size() == jt->second.size())
             for (std::size_t j = 0; j < vn.size(); ++j)
               state[vn[j]] = jt->second[j];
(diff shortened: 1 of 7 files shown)


moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -349,6 +354,8 @@ void RobotModel::buildGroupStates(const srdf::Model& srdf_model)
                                    "but that joint is not part of group '%s'",
                           group_state.name_.c_str(), jt->first.c_str(), jmg->getName().c_str());
       }
+      if (!remaining_joints.empty())
```
henningkayser: Additionally to throwing a warning, default values should be 0 or the state might be invalid. Alternatively, we could also make this an error and don't add the state at all.

rhaschke: The group state is essentially a map: joint name -> value. Hence, it is fine to leave some values "undefined" (not listed in the map) as long as `setToDefaultState` starts from a fully-defined state.
This should be the case anyway as the there might be joints outside the JMG, which will definitely not be modified by  `setToDefaultState`. The fake controller manager takes care of this (now).

rhaschke:
@henningkayser, @JafarAbdi: This fix doesn't resolve the failing unit test. Same for #1780.

JafarAbdi:
@rhaschke @henningkayser I debugged this exception

```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
```

I think the reason is a dangling reference, the exception happens in this [line](https://github.com/ros/geometry2/blob/melodic-devel/tf2/src/buffer_core.cpp#L265)

In the MoveItCpp constructor, we [pass](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp#L66) `tf_buffer_` to the `tf_listener` which takes a [reference](https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/src/transform_listener.cpp#L38) and store it as [reference](https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/include/tf2_ros/transform_listener.h#L74), now when the `MoveItCpp` destructor is called it will [reset](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp#L322) `tf_buffer_` which will cause the pointee to be destructed while there's another thread that uses the object when we default the destructor the `tf_listener_` will be destructed before `tf_buffer_` and the exception won't occur

This's the stack call for a reference

Screenshot from 2019-12-02 16-59-41

rhaschke:
Just fix the order of resetting variables in the custom destructor:
1. `tf_listener_`
2. `tf_buffer_`

PR #1782, opened 2019-11-26
disable flaky moveit_cpp_test

rhaschke:
As we don't yet have a solution for the flaky test (#1780 and #1781 still fail), I suggest to disable the unit test for now. @JafarAbdi, please test #1781 locally and provide a stack trace if possible.


rhaschke:
Merging w/o approval to eventually fix Travis.

Issue #156, opened 2016-08-25
rviz display should default to `move_group/monitored_planning_scene`


v4hn:
Another rather common problem of new users to MoveIt is that they do not realize that they should
change the topic of the display from `planning_scene` to `move_group/monitored_planning_scene` to see the internal state of the `move_group` (including a maintained octomap).
This definitely confused me a lot two years ago and I just noticed that this issue still exists (https://github.com/ros-planning/moveit/pull/128#issue-172791297).

I propose to change the default to `move_group/monitored_planning_scene`

Note that for example [the pr2 default rviz config already sets it to the monitored topic on startup](https://github.com/ros-planning/moveit_pr2/blob/indigo-devel/pr2_moveit_config/launch/moveit.rviz#L520)

mikeferguson:
+1

PR #1783, opened 2019-11-26
ControllerManager: wait for done-callback

rhaschke:
This is an alternative for #1741. Thanks to the very detailed analysis of @jschleicher, the race condition reported in #1493 was tracked down to an issue in actionlib: https://github.com/ros/actionlib/issues/155.
"As the done callback is really short" (https://github.com/ros-planning/moveit/pull/1741#issuecomment-557279556), the busy-waiting solution offered in #1741 is temporarily acceptable. Due to the same fact, I suggest to drop the remaining boilerplate (considering the timeout) of #1741 as well and keep the quick fix as simple as possible.

### DIFF
--- a/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
+++ b/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h
@@ -126,6 +126,11 @@ public:
   {
     if (controller_action_client_ && !done_)
       return controller_action_client_->waitForResult(timeout);
+#if 1  // TODO: remove when https://github.com/ros/actionlib/issues/155 is fixed
+    // workaround for actionlib issue: waitForResult() might return before our doneCB finished
+    while (!done_ && ros::ok())       // Thus, check the done_ flag explicitly,
+      ros::Duration(0.0001).sleep();  // which is eventually set in finishControllerExecution
+#endif
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


v4hn:
I don't agree.
It's not a good idea to introduce one more potential issue while working around one.

rhaschke:
Where do you see the potential issue?

v4hn:
The function clearly gets passed a timeout parameter.
If you introduce a (syntactic) endless-loop, that can violate the call specification.
I agree that this will *usually* not be an issue, but if it becomes one, I really don't want to debug it...

rhaschke:
In general, you are right and I would agree, but here the function we are waiting for is a no-brainer, setting a few variables only. I don't think this will change in the very next future. And as far as I have seen, users cannot pass an arbitrary function there. In this case we wouldn't like to busy-wait anyway.
So, from my point of view, we need to balance code maintainability (being able to remove this quick fix in the near future) vs. some potential change of the done callback, which then might take ages to finish. If you want to avoid the (potentially) infinite loop, I'm fine to limit the time to, say 0.1s, given that the corresponding code snippet fits into the `#ifdef` block.

v4hn:
> I'm fine to limit the time to, say 0.1s, given that the corresponding code snippet fits into the `#ifdef` block.

Sounds good to me.

rhaschke:
@v4hn: Done

rhaschke:
I will cherry-pick this to melodic as well. [Travis is currently validating it](https://travis-ci.com/ubi-agni/moveit/builds/138402019).

cpt-yoshi:
For your information. The `race condition` still occurs when testing with URSim in docker. I didin't test with other limit time.

The pull request from [ros/actionlib #156](https://github.com/ros/actionlib/pull/156) fixes the problem.

Issue #1743, opened 2019-11-20
Python3 move_commander_cmdline error in Melodic with _moveit_roscpp_initializer


Briancbn:
### Description
https://github.com/ros-planning/moveit/issues/1722

### Your environment
* ROS Distro: Melodic
* OS Version: e.g. Ubuntu 18.04.3 LTS
* Source build
* master branch commit 5b44334bcc5666ebf615b63ebd6f80ebb800b905

### Steps to reproduce
I install the python3 components through running
```
sudo apt-get install python3-pip python3-yaml
sudo pip3 install rospkg catkin_pkg

```
I tried to compile the whole moveit using python3 by running by setting the catkin config:
```
catkin config -DPYTHON_EXECUTABLE=/usr/bin/python3 -DPYTHON_INCLUDE_DIR=/usr/include/python3.6m -DPYTHON_LIBRARY=/usr/lib/x86_64-linux-gnu/libpython3.6m.so -DCMAKE_BUILD_TYPE=Release
```
Compile succeeded
Then I also manually add in the python3 directories into the PYTHONPATH
```
export PYTHONPATH=/home/rosi/Documents/moveit_ws/devel/lib/python3/dist-packages:$PYTHONPATH
```
I edit the first line of the `moveit_commander_cmdline.py` to `#!/usr/bin/env python3`

run the `rosrun moveit_commander moveit_commander_cmdline.py` command. It gives me an error, it comes from importing the packages from `moveit_ros_planning_interface`, it seems that it cannot find the py_binding_tools _roscpp_initializer 
```
Traceback (most recent call last):
  File "/home/rosi/Documents/moveit_ws/src/moveit/moveit_commander/bin/moveit_commander_cmdline.py", line 16, in <module>
    from moveit_commander import MoveGroupCommandInterpreter, MoveGroupInfoLevel, roscpp_initialize, roscpp_shutdown
  File "/home/rosi/Documents/moveit_ws/devel/lib/python3/dist-packages/moveit_commander/__init__.py", line 35, in <module>
    exec(__fh.read())
  File "<string>", line 2, in <module>
  File "/home/rosi/Documents/moveit_ws/src/moveit/moveit_commander/src/moveit_commander/roscpp_initializer.py", line 35, in <module>
    from moveit_ros_planning_interface import _moveit_roscpp_initializer
[...]

PR #1788, opened 2019-11-28
[moveit_commander] python3 import fixes

v4hn:
* [moveit_commander] python3 import fixes

- force relative import in conversions.py and planning_scene_interface.py
- try to import StringIO from StringIO module first, then from io module

CP of #1786

### DIFF
--- a/moveit_commander/src/moveit_commander/conversions.py
+++ b/moveit_commander/src/moveit_commander/conversions.py
@@ -32,14 +32,20 @@
 #
 # Author: Ioan Sucan
 
-import StringIO
+try:
+    # Try Python 2.7 behaviour first
+    from StringIO import StringIO
+except ImportError:
+    # Use Python 3.x behaviour as fallback
+    from io import StringIO
+
 from moveit_commander import MoveItCommanderException
 from geometry_msgs.msg import Pose, PoseStamped, Transform
 import rospy
 import tf
 
 def msg_to_string(msg):
-    buf = StringIO.StringIO()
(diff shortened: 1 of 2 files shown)


rhaschke:
Thanks for cherry-picking. I directly committed from commandline.

YijiongLin:
Hello Sir @v4hn ,

I'd like to use Moveit with python 3 in Melodic. And I just installed the Moveit with this [guide](http://docs.ros.org/en/melodic/api/moveit_tutorials/html/doc/getting_started/getting_started.html#install-moveit). And I also met this [problem (/issues/1743)](https://github.com/ros-planning/moveit/issues/1743).

Could I know if this commit can solve this problem?

rhaschke:
The MoveIt master branch, in principle, runs with python3. But this requires the whole ROS environment, i.e. all upstream packages, to be built with python3 as well. This is not recommended. If you want to go for python3, I suggest switching to Noetic.

YijiongLin:
HI @rhaschke ,

Thanks for your insight and kind reminder!

Though Noetic is python3-friendly,  some ROS packages don't catch up with it, e.g. [franka_ros is still on its way](https://github.com/frankaemika/franka_ros/issues/123), and many machine learning packages are built with python3. So running them all together might indeed need the Melodic with python3.

Anyway, thanks for your reply!!

Cheers,

rhaschke:
It might be easier to build not-yet migrated packages like franka_ros from source instead of building everything from scratch...

PR #1789, opened 2019-11-28
codecov yml setting master as default branch

tylerjw:
### Description

This change sets the default branch for code coverage to be `master`.  This is the branch that new PRs will be compared against for relative changes in code coverage.


tylerjw:
Looking at the report this file isn't being picked up by codecov website and your change effectively fixes this.

PR #1792, opened 2019-11-29
Publish attached collsion objects with the move_group TfPublisher capability

JonasTietz:
### Description

This PR adds functionality of publishing attached collision objects and their subframes to the TfPublisher capability.
It also fixes the timestamps for the transform msgs.

### DIFF
--- a/moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
@@ -87,15 +87,15 @@ void TfPublisher::publishPlanningSceneFrames()
 
       for (const auto& obj : *world)
       {
-        std::string parent_frame = prefix_ + obj.second->id_;
+        std::string object_frame = prefix_ + obj.second->id_;
         transform = tf2::eigenToTransform(obj.second->shape_poses_[0]);
-        transform.child_frame_id = parent_frame;
+        transform.child_frame_id = object_frame;
         transform.header.stamp = stamp;
         transform.header.frame_id = planning_frame;
         broadcaster.sendTransform(transform);
 
         const moveit::core::FixedTransformsMap& subframes = obj.second->subframe_poses_;
-        publishSubframes(broadcaster, subframes, parent_frame, stamp);
+        publishSubframes(broadcaster, subframes, object_frame, stamp);
       }
 
       const robot_state::RobotState& rs = locked_planning_scene->getCurrentState();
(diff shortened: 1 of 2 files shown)


moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -52,6 +53,20 @@ TfPublisher::~TfPublisher()
   thread_.join();
 }
 
+void TfPublisher::publishSubframes(tf2_ros::TransformBroadcaster& broadcaster,
+                                   moveit::core::FixedTransformsMap& subframes, std::string parent_frame)
```
v4hn: CI complains that `parent_frame` should be `const std::string&` instead of `std::string`

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -70,6 +72,7 @@ void TfPublisher::publishPlanningSceneFrames()
         std::string parent_frame = prefix_ + obj.second->id_;
         transform = tf2::eigenToTransform(obj.second->shape_poses_[0]);
         transform.child_frame_id = parent_frame;
+        transform.header.stamp = ros::Time::now();
```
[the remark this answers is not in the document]
v4hn: Yes and no.

Not providing a timestamp is an actual bug @JonasTietz discovered, as frames will not be updated.

But, mostly for aesthetic reasons, could you rework this to get the time only once and use it for all frames?
It is nicer (and actually less overhead for the TF system if it has to chain transforms) if all frames have the same stamp.

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
```diff
@@ -70,17 +85,28 @@ void TfPublisher::publishPlanningSceneFrames()
...
+      for (auto& attached_body : attached_collision_objects)
+      {
+        std::string parent_frame = prefix_ + attached_body->getName();
+        transform = tf2::eigenToTransform(attached_body->getFixedTransforms()[0]);
+        transform.child_frame_id = parent_frame;
+        transform.header.stamp = ros::Time::now();
+        transform.header.frame_id = attached_body->getAttachedLinkName();
+        broadcaster.sendTransform(transform);
+
+        moveit::core::FixedTransformsMap subframes = attached_body->getSubframeTransforms();
```
v4hn: You missed a reference here. This line copies the subframes for no reason.

moveit_ros/move_group/src/default_capabilities/tf_publisher_capability.cpp
v4hn: Also, there is no reason to have this as a member function at all, as it does not use any class variables.
I'd propose to turn this into a local function in anonymous namespace (or at least make the function definition static if you prefer).

v4hn:
Thanks for the patch @JonasTietz ! Thanks for the cleanup @rhaschke.

This is a neat addition to the project!

Issue #1785, opened 2019-11-27
[ros_planning_interface] Undefined symbol on python module import with geometry2 and moveit installed from source


gleichdick:
### Description

According to the [Move Group Tutorial](https://ros-planning.github.io/moveit_tutorials/doc/move_group_python_interface/move_group_python_interface_tutorial.html), I tried to import the `moveit_commander.conversions` python module on an openSUSE source setup.
It failed because the symbol `void tf2::fromMsg(geometry_msgs::Quaternion const&, Eigen::Quaterniond &)` (demangled and simplified) could not be found.
You can reproduce it in a container with geometry2 and moveit installed from source, see below.

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04, ros:melodic-ros-base image
* source installed melodic-devel branch, [rosinstall file](https://gist.github.com/gleichdick/152c0b3a7932ee772d4d6342b83b138f)

### Steps to reproduce

Build and run docker container (takes a while tho)

```Dockerfile
FROM ros:melodic-ros-base

RUN apt-get update && apt-get upgrade -y && apt-get install -y python3-pip

# force installation of geometry2 from source
RUN apt-mark hold ros-melodic-tf2-eigen

# install catkin tools from source because .deb package is broken
RUN git clone --depth=1 https://github.com/catkin/catkin_tools /catkin_tools && \
   cd /catkin_tools && pip3 install -e .

WORKDIR /ws_moveit

RUN wstool init src https://gist.github.com/gleichdick/152c0b3a7932ee772d4d6342b83b138f/raw/b63b050d1362cdea3de05573de62f838bfa11241/moveit.rosinstall && \
    wstool update -t src && \
    catkin config --extend /opt/ros/melodic --install --install-space /ws_moveit/install --cmake-args -DCMAKE_BUILD_TYPE=Release

RUN rosdep install --from-paths src --ignore-src --rosdistro melodic -y

RUN catkin build
```

Inside the container:

```sh
$ source /ws_moveit/install/setup.sh
$ python
Python 2.7.15+ (default, Oct  7 2019, 17:39:04) 
[GCC 7.4.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import moveit_commander
```

### Expected behaviour

Module should be imported

### Actual behaviour

Module could not be imported because symbol

```c++
[...]

PR #1794, opened 2019-12-01
Workaround for #1785

gleichdick:
### Description

Reordered tf2 header includes, workaround for #1785. Needs ros/geometry2#431 applied to geometry2 package

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
@@ -34,6 +34,9 @@
 
 /* Author: Ioan Sucan */
 
+#include <tf2_eigen/tf2_eigen.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+#include <tf2/convert.h>
 #include <moveit/move_group_interface/move_group_interface.h>
 #include <moveit/py_bindings_tools/roscpp_initializer.h>
 #include <moveit/py_bindings_tools/py_conversions.h>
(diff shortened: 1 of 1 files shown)


v4hn:
Does this patch *require* the geometry2 pull-request or is it a workaround for it?

I'd support only the latter one.

gleichdick:
To compile moveit with latest geometry2 melodic-devel branch, both packages need to reorder their headers (as I did in both PRs). The reason is a templated declaration (not definition) of `fromMsg()` / `toMsg()`, so `tf2::convert` picks the templated declaration instead of the correct overload (see ros/geometry2#340 for details). By reordering the overload is defined before the template declaration.
This PR should not affect compiling with the latest **packaged** geometry2 version, where ros/geometry2#367 is not applied.
This is oviously not bullet-proof, but the tf2 api has to be changed for a proper solution.

kejxu:
> We should not rely on a specific ordering of include headers.
> If there is a bug in tf2, it should be fixed there. Hopefully @kejxu, the original author of [ros/geometry2#367](https://github.com/ros/geometry2/pull/367), can help.

thanks much for the redirection; however, would appreciate really appreciate it if there could be more info. What exactly is the bug? https://github.com/ros/geometry2/pull/367 itself did not cause any build break in tf2 right?

kejxu:
read through https://github.com/ros/geometry2/issues/430, https://github.com/ros/geometry2/pull/431, and https://github.com/ros-planning/moveit/issues/1785. @gleichdick, thanks a lot for the amount of time and effort invested in this issue!

Agree on the conclusion that the root cause is around the use of C++ templates (https://github.com/ros-planning/moveit/issues/1785#issuecomment-559271015), and a proper fix (instead of workarounds) will be to correct those usages altogether. The objective of https://github.com/ros/geometry2/pull/367 was to fix a MSVC build error in a small yet precise change, apologies for introducing build breaks in its downstream project (moveit).

gleichdick:
So, I proposed a solution in ros/geometry2#433, any comments are welcome :smile:

rhaschke:
Closing here. Should be fixed via https://github.com/ros/geometry2/issues/430 or https://github.com/ros/geometry2/pull/433.

Issue #1115, opened 2018-10-24
RViz broken Interactive Marker in "Scene Objects" Tab


v4hn:
The MotionPlanning panel provides a tab called "Scene Objects" that can be used to move objects in the PlanningScene maintained by `move_group`.
In order to do so, the user chooses one of the objects in the scene from the list on the left and an interactive marker appears that can be used to move the object.

![scene_objects](https://user-images.githubusercontent.com/680358/47417029-5875df00-d777-11e8-940f-1b5be9adc515.png)

However, this marker has a few issues:

- [ ] If the user disables and reenables the MotionPlanning display (the panel disappears and reappears too), suddenly a (fixed-size?) coordinate system and some black lines, that jump around when the camera is moved, appear. Neither of them should be there:

![re_enable](https://user-images.githubusercontent.com/680358/47417354-18fbc280-d778-11e8-9dd2-1b6a00523e40.png)

- [ ] The scale of the marker is independent of the size of the CollisionObject. It appears to big for small objects and can be hidden by big objects:

![hidden_marker](https://user-images.githubusercontent.com/680358/47418671-e30c0d80-d77a-11e8-9b6f-6e3971ed714a.png)

- [ ] The coupling between the Marker and the MotionPlanning display seems to be insufficient. When RViz additionally contains an PlanningScene this can result in glitches where the Marker becomes active even when the MotionPlanning display is disabled:

![planning_scene_im_glitch](https://user-images.githubusercontent.com/680358/47417993-7cd2bb00-d779-11e8-9a6f-0c8310934ba2.png)

To reproduce (does not necessarily happen every time):
- add a MotionPlanning and a PlanningScene display to RViz
- disable the PlanningScene display
- select the "Scene Objects" tab in the Motion Planning Panel and select an object from the list
  -> an interactive marker appears
- disable the MotionPlanning display
  -> the marker disappears
- enable the PlanningScene display
  -> sometimes the marker reappears

These are partially overlapping issues, but probably warrant different pull-requests.

### Your environment
* ROS Distro: Kinetic
* OS Version: Lunar-Linux / similar on Ubuntu
* Source or Binary build? Both

RyodoTanaka:
@v4hn 
I'm trying to reproduce the same error.
Especially for the first error, the situation is a little bit different.
[...]

PR #1796, opened 2019-12-01
Resize scene marker with collision object

RyodoTanaka:
### Description
As mentioned at #1115 , the scene objects marker was broken at scaling.
This PR fix that problem.
And the same PR for master branch on #1795 .
This PR is for melodic-devel branch.

### The change on GUI
With this PR, the marker on the scene objects will be able to change their scales as showing on following.
This PR also fix the default marker size to fit the size of the objects.
scale1
scale2

### Checklist

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
@@ -105,44 +105,7 @@ void MotionPlanningFrame::sceneScaleChanged(int value)
           s->scale((double)value / 100.0);
           ps->getWorldNonConst()->addToObject(scaled_object_->id_, shapes::ShapeConstPtr(s),
                                               scaled_object_->shape_poses_[i]);
-
-          // Resize the marker scale
-          // 1. Get obj shape bound size
-          Eigen::Vector3d center;
-          double radius;
-          shapes::computeShapeBoundingSphere(s, center, radius);
-          double bound = (center.cwiseAbs().maxCoeff() + radius) * 2.0;
-          // bound + padding (20%) size
-          bound *= 1.2;
-          // 2. Change marker size
-          for (std::size_t i = 0; i < viz_scene_marker_->controls.size(); i++)
-          {
-            for (std::size_t j = 0; j < viz_scene_marker_->controls[i].markers.size(); j++)
-            {
-              switch (viz_scene_marker_->controls[i].markers[j].type)
-              {
-                case visualization_msgs::Marker::ARROW:
-                {
-                  viz_scene_marker_->controls[i].markers[j].points[0].x = pow(-1.0, j % 2) * (bound * 0.5);
-                  viz_scene_marker_->controls[i].markers[j].points[1].x = pow(-1.0, j % 2) * (bound * 0.9);
-                  viz_scene_marker_->controls[i].markers[j].scale.x = bound * 0.15;
-                  viz_scene_marker_->controls[i].markers[j].scale.y = bound * 0.25;
-                  viz_scene_marker_->controls[i].markers[j].scale.z = bound * 0.2;
-                  break;
-                }
-                case visualization_msgs::Marker::TRIANGLE_LIST:
-                {
-                  viz_scene_marker_->controls[i].markers[j].scale.x = bound;
-                  viz_scene_marker_->controls[i].markers[j].scale.y = bound;
-                  viz_scene_marker_->controls[i].markers[j].scale.z = bound;
-                  break;
-                }
-                default:
-                  break;
-              }
-            }
-          }
-          scene_marker_->processMessage(*viz_scene_marker_);
+          resizeInteractiveMarker(s);
         }
         planning_display_->queueRenderSceneGeometry();
       }
@@ -764,48 +727,11 @@ void MotionPlanningFrame::createSceneInteractiveMarker()
 
     rviz::InteractiveMarker* imarker = new rviz::InteractiveMarker(planning_display_->getSceneNode(), context_);
     interactive_markers::autoComplete(int_marker);
-
-    // Resize the marker scale
-    // 1. Get obj shape bound size
-    Eigen::Vector3d center;
-    double radius;
-    shapes::computeShapeBoundingSphere(obj->shapes_[0]->clone(), center, radius);
-    double bound = (center.cwiseAbs().maxCoeff() + radius) * 2.0;
-    // bound + padding (20%) size
-    bound *= 1.2;
-    // 2. Change marker size
-    for (std::size_t i = 0; i < int_marker.controls.size(); i++)
-    {
-      for (std::size_t j = 0; j < int_marker.controls[i].markers.size(); j++)
-      {
-        switch (int_marker.controls[i].markers[j].type)
-        {
-          case visualization_msgs::Marker::ARROW:
-          {
-            int_marker.controls[i].markers[j].points[0].x = pow(-1.0, j % 2) * (bound * 0.5);
-            int_marker.controls[i].markers[j].points[1].x = pow(-1.0, j % 2) * (bound * 0.9);
-            int_marker.controls[i].markers[j].scale.x = bound * 0.15;
-            int_marker.controls[i].markers[j].scale.y = bound * 0.25;
-            int_marker.controls[i].markers[j].scale.z = bound * 0.2;
-            break;
-          }
-          case visualization_msgs::Marker::TRIANGLE_LIST:
-          {
-            int_marker.controls[i].markers[j].scale.x = bound;
-            int_marker.controls[i].markers[j].scale.y = bound;
-            int_marker.controls[i].markers[j].scale.z = bound;
-            break;
-          }
-          default:
-            break;
-        }
-      }
-    }
-    imarker->processMessage(int_marker);
     imarker->setShowAxes(false);
     scene_marker_.reset(imarker);
     viz_scene_marker_.reset(new visualization_msgs::InteractiveMarker);
     *viz_scene_marker_ = int_marker;
+    resizeInteractiveMarker(obj->shapes_[0]->clone());
 
     // Connect signals
     connect(imarker, SIGNAL(userFeedback(visualization_msgs::InteractiveMarkerFeedback&)), this,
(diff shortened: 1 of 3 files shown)


moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -105,6 +105,70 @@ void MotionPlanningFrame::sceneScaleChanged(int value)
...
+          double half_bound = -std::numeric_limits<double>::max();
+          switch(s->type){
+            case shapes::ShapeType::SPHERE:{
+              shapes::Sphere* sphere = static_cast<shapes::Sphere*>(s);
+              half_bound = sphere->radius;
+              break;
+            }
+            case shapes::ShapeType::CYLINDER:{
+              shapes::Cylinder* cylinder = static_cast<shapes::Cylinder*>(s);
+              half_bound = (cylinder->radius > cylinder->length/2.0) ? cylinder->radius : cylinder->length/2.0;
```
mamoll: why not simply use `std::max(cylinder->radius, cylinder->length/2.0)`

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -105,6 +105,70 @@ void MotionPlanningFrame::sceneScaleChanged(int value)
...
+              break;
+            }
+            case shapes::ShapeType::CONE:{
+              shapes::Cone* cone = static_cast<shapes::Cone*>(s);
+              half_bound = (cone->radius > cone->length/2.0) ? cone->radius : cone->length/2.0;
```
mamoll: same here

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -105,6 +105,70 @@ void MotionPlanningFrame::sceneScaleChanged(int value)
...
+              break;
+            }
+            case shapes::ShapeType::BOX:{
+              shapes::Box* box = static_cast<shapes::Box*>(s);
+              for(std::size_t i = 0; i < 3; i++)
+                half_bound = (half_bound < box->size[i]/2.0) ? box->size[i]/2.0 : half_bound;
```
mamoll: replace for loop with `half_bound = std::max_element(box->size.begin(), box->size.end())`?

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -105,6 +105,70 @@ void MotionPlanningFrame::sceneScaleChanged(int value)
...
+              break;
+            }
+            case shapes::ShapeType::MESH:{
+              shapes::Mesh* mesh = static_cast<shapes::Mesh*>(s);
+              for(std::size_t i = 0; i < mesh->vertex_count*3+3; i++)
+                half_bound = (half_bound < abs(mesh->vertices[i])) ? abs(mesh->vertices[i]) : half_bound;
```
mamoll: same here.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -105,6 +105,70 @@ void MotionPlanningFrame::sceneScaleChanged(int value)
...
+            {
+              switch(viz_scene_marker_->controls[i].markers[j].type){
+                case visualization_msgs::Marker::ARROW:{
+                  // // bound + padding (40%) size
+                  double bound_size = half_bound*2.0*1.4;
+                  viz_scene_marker_->controls[i].markers[j].points[0].x = pow(-1.0,j%2)*(bound_size*0.5);
+                  viz_scene_marker_->controls[i].markers[j].points[1].x = pow(-1.0,j%2)*(bound_size*0.9);
+                  viz_scene_marker_->controls[i].markers[j].scale.x = bound_size*0.15;
+                  viz_scene_marker_->controls[i].markers[j].scale.y = bound_size*0.25;
+                  viz_scene_marker_->controls[i].markers[j].scale.z = bound_size*0.2;
```
mamoll: why different scaling factors for x, y, and z?

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -105,6 +105,70 @@ void MotionPlanningFrame::sceneScaleChanged(int value)
...
+              break;
+          }
+          // Change marker size
+          for(std::size_t i = 0; i < viz_scene_marker_->controls.size(); i++)
+          {
+            for(std::size_t j = 0; j < viz_scene_marker_->controls[i].markers.size(); j++)
+            {
+              switch(viz_scene_marker_->controls[i].markers[j].type){
+                case visualization_msgs::Marker::ARROW:{
+                  // // bound + padding (40%) size
```
mamoll: delete one `//`

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -709,12 +786,78 @@ void MotionPlanningFrame::createSceneInteractiveMarker()
         robot_interaction::make6DOFMarker(std::string("marker_") + sel[0]->text().toStdString(), shape_pose, 1.0);
     int_marker.header.frame_id = planning_display_->getRobotModel()->getModelFrame();
     int_marker.description = sel[0]->text().toStdString();
-
+    
     rviz::InteractiveMarker* imarker = new rviz::InteractiveMarker(planning_display_->getSceneNode(), context_);
     interactive_markers::autoComplete(int_marker);
+
+    // Resize the marker scale
```
mamoll: the code below looks the same as above. maybe refactor to avoid the duplicate code?

rhaschke:
This PR is a duplicate of #1795. Let's focus on one of them first and subsequently tackle cherry-picking.
Thus, closing here for now.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
RyodoTanaka: I agree with utilizing `std::max()` function.
But I will replace this function into `computeShapeBoundingSphere()`  as @rhaschke mentioned.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
RyodoTanaka: Same, I will replace here into `computeShapeBoundingSphere()` function.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
RyodoTanaka: Same, I will replace here into computeShapeBoundingSphere() function.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
RyodoTanaka: Same, I will replace here into computeShapeBoundingSphere() function.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
RyodoTanaka: I will delete duplicated `//`.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
RyodoTanaka: For the `visualization_msgs/Marker` message, the `scale` is used different way for each marker types.
For this part, I am using `Arrow` marker. And on this type, 
- scale.x is the shaft diameter.
- scale.y is the head diameter
- scale.z the head length.
Thus, to keep same size ratio, I need to set different numbers for them.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
RyodoTanaka: I agree with this, and as @rhaschke mentioned, I will replace this part to the `computeShapeBoundingSphere()` function.

RyodoTanaka:
@mamoll @rhaschke 
Thank you for reviewing.
1.  About computeShapeExtents() or computeShapeBoundingSphere()
I did not know the function is. And the `computeShapeBoundingSphere()` function is what I actually needed.
Thus, I will replace them.

2. > Is it really necessary to maintain the variable viz_scene_marker_?
On this point, I think we need to keep the information of the interactive marker.
In the class, we have the variable `scene_marker_` .
But with this variable, we could not modify the visualization size.

If we do not use the variable `viz_scene_marker_`, we need to create local `visualization_msgs::InteractiveMarker` variables.
I thought, that is not efficiency. Thus I am using `viz_scene_marker_` variable.

So, I want to ask you which will be a best ?

Best Regards,

rhaschke:
As far as I can see, the only reason to keep `viz_scene_marker_` is to easily adjust the scale.
The need for this, is actually a limitation of rviz::InteractiveMarker's API not providing a setScale() method. I will add this function in rviz and see whether this is sufficient.
If you could work on the other comments that would be of great help.

RyodoTanaka:
@rhaschke 
Yes, exactly, the `rviz::interactiveMarker` API is not follow the actual scaling method.
They have kind of scaling. But they just modify the `visualization_msgs::InteractiveMarker`'s scale, not `visualization_msgs::Marker's scale.
To do actual modify of the marker's scale, we need to modify  `visualization_msgs::Marker`'s scale, from `visualization_msgs::InteractiveMarkerControl`.
Thus, if there will be some function to access `visualization_msgs::Interactivemarker` from `rviz::interactiveMarker` class, that will help that I think.

In anyway, I will fix the calculation of bounding part to use `computeShapeBoundingSphere()` function.

Best Regards,

RyodoTanaka:
Oh sorry, I will replace the bounding part to ` computeShapeExtents()` function.

RyodoTanaka:
@rhaschke 
I updated the code to utilize `computeShapeBoundingSphere()` function.

RyodoTanaka:
The duplicate of the code to change the scale of `visualization_msgs::InteractiveMarker` is still remain.
But this point is come from the `rviz::InteractiveMarker` class does not support actual scaling that @rhaschke mentioned.

RyodoTanaka:
@rhaschke @mamoll 
Sorry for bother you.
But may I ask you please review or merge this ?
Best Regards,

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
rhaschke: Additionally to using `computeShapeBoundingSphere()`, you should factor the marker adjustment code into its own function.

rhaschke:
@RyodoTanaka, please see https://github.com/ros-planning/moveit/pull/1796#discussion_r355325246 for an open comment.

RyodoTanaka:
@rhaschke 
Oh, I am sorry. I did not recognized.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
RyodoTanaka: @rhaschke 
The "own function" you mean, I should create original function in this class to utilize `computeShapeBoundingSphere()` function and to adjust the marker. Is that right ?

rhaschke: Exactly.

RyodoTanaka: I created the `resizeInteractiveMarker()` function, and to utilize it.

rhaschke:
I included this PR in #1808, which tackles all the remaining issues mentioned in #1115 as well.
Thus closing here.

RyodoTanaka:
@rhaschke 
Thank you for your review and fixing.
I did not know how to treat `rviz::InteractiveMarker` class nicely as you did.
This helped me to understand it !!

PR #1802, opened 2019-12-06
clear roadmap validity for LazyPRM/LazyPRMstar when using multi-query planning is enabled

mamoll:
### Description

When multi-query planning is enabled (a feature add in PR #1799), the roadmap is kept around between calls to solve. When the planning scene doesn't change between calls to solve(), this is not a problem. In pick-and-place tasks it often does, so it makes sense to recheck validity of nodes/edges. With LazyPRM / LazyPRMstar this is easy to do. For PRM / PRMstar it does not make sense.

### Checklist


mamoll:
See also https://github.com/ompl/ompl/commit/31c3689d12ca797306e444580e8c7f3705cc086f

PR #1813, opened 2019-12-19
RobotTrajectory copy() method

mlautman:
A follow on PR to https://github.com/ros-planning/moveit/pull/1760 with the following modifications: 

1. Adds tests
2. Removes the deprecation of RobotTrajectory copy constructor as it makes it frustrating to pass  RobotTrajectory as the return value of a method.
3. Makes shallow copy the default behavior for the copy constructor to preserve expected functionality but allows users to call a method copy() directly with a parameter to `shallow_copy_waypoints` which causes the copy() method to copy pointers as opposed to making copies of RobotState.

@shivaang12 @v4hn and @rhaschke

### DIFF
--- a/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
@@ -1,0 +1,218 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2013, Willow Garage, Inc.
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the Willow Garage nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+*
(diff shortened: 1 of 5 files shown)


shivaang12:
@mlautman Couldn't have done better!

moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+  void initTestTrajectory(robot_trajectory::RobotTrajectoryPtr& trajectory)
+  {
+    // Init a traj
+    ASSERT_TRUE(robot_model_->hasJointModelGroup(arm_jmg_name_)) << "Robot model does not have group: "
+                                                                 << arm_jmg_name_;
+
+    trajectory = std::make_shared<robot_trajectory::RobotTrajectory>(robot_model_, arm_jmg_name_);
+
+    ASSERT_EQ(trajectory->getGroupName(), arm_jmg_name_) << "Generated trajectory group name does not match";
```
rhaschke: `ASSERT` should only be used if the function should be left on failure.

mlautman:
> In future, you should try to augment an existing PR instead of opening a new one - if ever possible,
because that maintains the discussion history.

Agreed but I didn't want to push to the PR directly in case @shivaang12 didn't approve of my changes. I also didn't want to open a PR to his fork as it wouldn't get as much visibility. Should I have just pushed to the PR then?

> I pushed some more cleanup changes to the original #1760 and will close this one here again.
Great

rhaschke:
Usually it's fine to push to PR branches. If contributors don't like this, they can tick the corresponding checkbox when creating the PR.

felixvd:
And they will still have their local copy, so they can incorporate your
commits and force-push if they wish.

Maybe that helps assuage your guilt.

PR #1814, opened 2019-12-19
[AIST-QA]Fix potential null-pointer dereferencing issues in moveit_setup_assistant.

max-krichenbauer:
### Description

`model()` may return `nullptr` and `qobject_cast<>`, `findPoseByName()` and `findVJointByName` may return zero on failure.
Furthermore, in `default_collisions_widget.cpp`, the `header` variable is not initialized. 
These variables are used without checking, potentially causing null pointer de-referencing.
This fix adds safeguards to mitigate this risk.

This contribution is made by AIST ( https://www.aist.go.jp ) based on static code analysis with klocwork (Perforce Software).

### Checklist

### DIFF
--- a/moveit_setup_assistant/src/tools/collision_linear_model.cpp
+++ b/moveit_setup_assistant/src/tools/collision_linear_model.cpp
@@ -132,8 +132,8 @@ DisabledReason CollisionLinearModel::reason(int row) const
   CollisionMatrixModel* m = qobject_cast<CollisionMatrixModel*>(sourceModel());
   if (!m)
   {
-    Q_ASSERT(m);
-    return moveit_setup_assistant::NOT_DISABLED;
+    ROS_FATAL_NAMED("collision_linear_model", "sourceModel() is null");
+    ROS_BREAK();
   }
   QModelIndex srcIndex = this->mapToSource(index(row, 0));
 
@@ -247,7 +247,10 @@ bool SortFilterProxyModel::filterAcceptsRow(int source_row, const QModelIndex& s
 {
   CollisionLinearModel* m = qobject_cast<CollisionLinearModel*>(sourceModel());
   if (!m)
-    return false;
+  {
+    ROS_FATAL_NAMED("collision_linear_model", "sourceModel() is null");
+    ROS_BREAK();
+  }
 
   if (!(show_all_ || m->reason(source_row) <= moveit_setup_assistant::ALWAYS ||
         m->data(m->index(source_row, 2), Qt::CheckStateRole) == Qt::Checked))
--- a/moveit_setup_assistant/src/tools/collision_linear_model.cpp
+++ b/moveit_setup_assistant/src/tools/collision_linear_model.cpp
@@ -129,8 +129,15 @@ QVariant CollisionLinearModel::data(const QModelIndex& index, int role) const
 
 DisabledReason CollisionLinearModel::reason(int row) const
 {
+  CollisionMatrixModel* m = qobject_cast<CollisionMatrixModel*>(sourceModel());
+  if (!m)
+  {
+    Q_ASSERT(m);
+    return moveit_setup_assistant::NOT_DISABLED;
+  }
   QModelIndex srcIndex = this->mapToSource(index(row, 0));
-  return qobject_cast<CollisionMatrixModel*>(sourceModel())->reason(srcIndex);
+
+  return m->reason(srcIndex);
 }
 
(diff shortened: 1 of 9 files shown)


rhaschke:
Thanks for looking into this, but actually I don't see a benefit in replacing (potential) segfaults due to null pointers with an abort due to an assert. Even worse: in release mode assertions are disabled and the functions just silently return "something" in case of these fatal errors. Actually, if the pointers _are_ zero, the functions were wrongly used and thus should fail as early as possible.

max-krichenbauer:
Thank you for your review!
The idea was that, as you say, execution should fail as early as possible, but should give an indication where and why it failed. A segfault doesn't help finding the root of the problem.
I wasn't aware that in release mode, asserts will result in silently returning "something". Would an error message via `ROS_ERROR_NAMED` be more appropriate?

rhaschke:
In all these cases, an invalid pointer is a fatal error condition. I agree, that the segfault doesn't immediately name the offending line, but I don't agree that this error condition should be handled and indicated by explicit code. As I said, the code is/should be designed such that these pointers _are_ always valid. Obviously the static code analyzer doesn't analyze so deep and thus complains.
However, I expect issues only during development phase and in this phase, running a debugger, you will find the culprit immediately. Hence, I don't see a benefit of cluttering the source code with error handling code for errors that should not occur. If they occur, this is a bug that needs to be fixed, which usually requires deeper analysis anyway.

max-krichenbauer:
Sorry, I'm not sure if I agree that "crashing without checks or error reporting" is the best way to handle exceptional cases.

During debugging, just because you're using a debug build doesn't mean you always have a debugger attached. Then an assert would be quite helpful.

During deployment, you'd still like to have some indication where the crash originated. Else the user can't properly report the issue other than "it crashed, don't know why".

I understand that you want to avoid clutter in the code, but I believe the few lines are well worth the benefits.

rhaschke:
I'm fine with adding an `assert(pointer)` where appropriate.

> During deployment, you'd still like to have some indication where the crash originated. Else the user can't properly report the issue other than "it crashed, don't know why".

However, asserts will only be considered for non-release builds, i.e. only during development.
Thus, it will not help the user to provide a more detailed error report. My key message is that these pointers should not be undefined after proper testing. If there are anyway, there is some serious bug in the code, which cannot/should not be just ignored by returning some default value.

max-krichenbauer:
> However, asserts will only be considered for non-release builds

About this topic again, I did some research but could not confirm it.
Releases are build with `O3` and `NDEBUG` which disables C++ `assert()` and `ROS_ASSERT`,
but `Q_ASSERT` is dependent on `QT_NO_DEBUG`, which apparently is **not** set.
Please let me know if I'm missing something.

I do understand that your main point is that the code should be so well designed that null pointers can never happen, but since it's almost impossible to guarantee that code is bug-free, error checking is usually considered *good practice* and may help improve code quality by making it easier to find bugs.

moveit_setup_assistant/src/tools/collision_linear_model.cpp
```diff
@@ -129,8 +129,15 @@ QVariant CollisionLinearModel::data(const QModelIndex& index, int role) const
 
 DisabledReason CollisionLinearModel::reason(int row) const
 {
+  CollisionMatrixModel* m = qobject_cast<CollisionMatrixModel*>(sourceModel());
+  if (!m)
+  {
+    Q_ASSERT(m);
```
rhaschke: In a release build, `Q_ASSERT` should be a no-op. Just silently continuing and returning an arbitrary value is actually hiding the error and will make it even harder to track the wrong behavior.

moveit_setup_assistant/src/tools/collision_linear_model.cpp
```diff
@@ -239,6 +246,9 @@ void SortFilterProxyModel::setShowAll(bool show_all)
 bool SortFilterProxyModel::filterAcceptsRow(int source_row, const QModelIndex& source_parent) const
 {
   CollisionLinearModel* m = qobject_cast<CollisionLinearModel*>(sourceModel());
+  if (!m)
+    return false;
```
rhaschke: Same here: You catch the invalid pointer, but you return an arbitrary value that changes the behavior of the program without indicating the error (by a segfault).

max-krichenbauer:
Thank you for the review.
I changed all checks to issue a fatal error and terminate execution.
This should make identifying the root cause in both debugging and deployment easier.

PR #1816, opened 2019-12-23
Improve JogArm thread handling

henningkayser:
This moves thread handling to the JogBaseInterface class and adds atomic bool flags for stopping all main loops from outside their threads.

Also, this fixes an issue where the `CollisionCheckThread` would indefinitely wait for command messages and won't get to run any collision checks using the C++ API.

### DIFF
--- a/moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
@@ -49,7 +49,7 @@ int main(int argc, char** argv)
 
   // Run the jogging C++ interface in a new thread to ensure a constant outgoing message rate.
   moveit_jog_arm::JogCppApi jog_interface;
-  std::thread jogging_thread(&moveit_jog_arm::JogCppApi::mainLoop, &jog_interface);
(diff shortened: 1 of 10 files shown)


AndyZe:
The integration tests still pass, the tutorial still works, the CPU load looks like it hasn't changed much. Looking good so far. Now I'll review the code.

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -49,7 +49,7 @@ int main(int argc, char** argv)
 
   // Run the jogging C++ interface in a new thread to ensure a constant outgoing message rate.
   moveit_jog_arm::JogCppApi jog_interface;
-  std::thread jogging_thread(&moveit_jog_arm::JogCppApi::mainLoop, &jog_interface);
+  std::thread jogging_thread([&]() { jog_interface.startMainLoop(); });
```
AndyZe: This syntax is weird to me. What's going on here?

Is this similar to `this`?

gavanderhoorn: This is a [lambda](https://en.cppreference.com/w/cpp/language/lambda).

PR #1821, opened 2019-12-31
delete attached body before adding a new one with the same id

mlautman:
### Description

Fixes a memory leak described here: https://github.com/ros-planning/moveit/issues/1820


moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -880,6 +880,9 @@ const AttachedBody* RobotState::getAttachedBody(const std::string& id) const
 
 void RobotState::attachBody(AttachedBody* attached_body)
 {
+  // Remove the old attached body if it still exists
```
mlautman:
```suggestion
  // Remove the old attached body if it exists
```

PR #1830, opened 2020-01-09
Follow up on #1828 and #1829

rhaschke:
Better fix for #1828: Avoid explicit use of new/delete when we use shared pointers anyway.
Cleanup #1829: Use range-based for-loop


rhaschke:

PR #1853, opened 2020-01-22
getFrameInfo(): avoid double search for link name

rhaschke:
`hasLinkModel()` and `getLinkModel()` both search for the link name...

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -1019,11 +1019,9 @@ const Eigen::Isometry3d& RobotState::getFrameInfo(const std::string& frame_id, c
     frame_found = true;
     return IDENTITY_TRANSFORM;
   }
-  if (robot_model_->hasLinkModel(frame_id))
+  if ((robot_link = robot_model_->getLinkModel(frame_id, &frame_found)))
   {
-    robot_link = robot_model_->getLinkModel(frame_id);
     BOOST_VERIFY(checkLinkTransforms());
-    frame_found = true;
     return global_link_transforms_[robot_link->getLinkIndex()];
   }
   robot_link = nullptr;


moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1019,11 +1019,9 @@ const Eigen::Isometry3d& RobotState::getFrameInfo(const std::string& frame_id, c
     frame_found = true;
     return IDENTITY_TRANSFORM;
   }
-  if (robot_model_->hasLinkModel(frame_id))
+  if ((robot_link = robot_model_->getLinkModel(frame_id, &frame_found)))
   {
-    robot_link = robot_model_->getLinkModel(frame_id);
     BOOST_VERIFY(checkLinkTransforms());
-    frame_found = true;
```
felixvd: This boolean is used outside of the function and shouldn't be deleted. Apart from that this all looks good to me.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1019,11 +1019,9 @@ const Eigen::Isometry3d& RobotState::getFrameInfo(const std::string& frame_id, c
     frame_found = true;
     return IDENTITY_TRANSFORM;
   }
-  if (robot_model_->hasLinkModel(frame_id))
+  if ((robot_link = robot_model_->getLinkModel(frame_id, &frame_found)))
```
felixvd: nit: Remove the outermost bracket?

rhaschke: The extra pair of brackets is required to interpret the assignment result as a boolean (and silent compiler warnings, because this might be wrongly typed `==` comparison as well.)

moveit_core/robot_state/src/robot_state.cpp
rhaschke: Both, robot_link and frame_found are set in the above call to getLinkModel.

felixvd: I overlooked it, my bad. Good call.

moveit_core/robot_state/src/robot_state.cpp
felixvd: Interesting. Thanks.

PR #1861, opened 2020-01-23
Fix segfault in totg

### LINKED ISSUE #1805: Segfault in totg
### Description

Segfault in totg, possibly Eigen related

### Your environment
* ROS Distro: [Melodic]
* OS Version: e.g. Ubuntu 18.04
* Source or Binary build?
Source, master

### Steps to reproduce
Difficult at the moment to reproduce without sharing my code, but likely has to do with having a portion of the plan "static" (for example, where the first 3 steps don't really go anywhere because they come from trying to plan to where you already are)

### Expected behaviour
No segfault

### Actual behaviour
Segfault

```
Thread 16 "program_name" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff9a930700 (LWP 25171)]
0x00007ffff29662a4 in trajectory_processing::Trajectory::getVelocityMaxPathVelocityDeriv(double)
    ()
   from /some_directory/install/lib/libmoveit_trajectory_processing.so.1.0.1
(gdb) bt
#0  0x00007ffff29662a4 in trajectory_processing::Trajectory::getVelocityMaxPathVelocityDeriv(double) ()
   from /some_directory/install/lib/libmoveit_trajectory_processing.so.1.0.1
#1  0x00007ffff29669ac in trajectory_processing::Trajectory::integrateForward(std::__cxx11::list<trajectory_processing::Trajectory::TrajectoryStep, std::allocator<trajectory_processing::Trajectory::TrajectoryStep> >&, double) ()
   from /some_directory/install/lib/libmoveit_trajectory_processing.so.1.0.1
#2  0x00007ffff2967dfa in trajectory_processing::Trajectory::Trajectory(trajectory_processing::Path const&, Eigen::Matrix<double, -1, 1, 0, -1, 1> const&, Eigen::Matrix<double, -1, 1, 0, -1, 1> const&, double) ()
   from /some_directory/install/lib/libmoveit_trajectory_processing.so.1.0.1
#3  0x00007ffff296a0f4 in trajectory_processing::TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotTrajectory&, double, double) const ()
```
[With debug info](https://gist.github.com/Dale-Koenig/823d98b3fd033682fe65862c91e8555d)
[Another possibly related error](https://gist.github.com/Dale-Koenig/4fe5d4e3230bebe690bc07b726f6ea6d)

Dale-Koenig:
### Description

(At least partially) fixes  #1805.  Calls to `acos` can result in nan values if not careful due to numerical floating point errors.  If there is a possibility of calling `acos` on values that are approximately `-1`, then it will result in nan, which leads to a segfault later in totg. Also removes an incorrect comment that implies a section would avoid this error, but it actually handles a different case.

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -110,7 +110,6 @@ public:
     const Eigen::VectorXd start_direction = (intersection - start).normalized();
     const Eigen::VectorXd end_direction = (end - intersection).normalized();
 
-    // check if directions are divergent
     if ((start_direction - end_direction).norm() < 0.000001)
     {
       length_ = 0.0;
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -121,7 +121,7 @@ public:
     }
 
     // directions must be different at this point so angle is always non-zero
-    const double angle = acos(start_direction.dot(end_direction));
+    const double angle = acos(std::max(-1.0, start_direction.dot(end_direction)));
     const double start_distance = (start - intersection).norm();
     const double end_distance = (end - intersection).norm();
 


Dale-Koenig:
> Thanks for fixing this! Shouldn't we clip the upper limit as well? Also, did you rule out that the dot product produces nan values?

The check `if ((start_direction - end_direction).norm() < 0.000001)` does special  handling in the case where the dot product would be 1, so there should be no need to clip.  As for the dot product I don't think it can produce nan values unless the inputs are already nan, since it is just multiplication and addition.  For the errors I encountered the `acos` was at fault.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -110,7 +110,6 @@ class CircularPathSegment : public PathSegment
     const Eigen::VectorXd start_direction = (intersection - start).normalized();
```
rhaschke: If these vectors are close to zero, normalization will result in NaNs or Infs. Not sure though, that this can happen.

Dale-Koenig: @rhaschke True, I suppose there is division in there.  I believe the following loop, which is done before the constructor for `Path`, ensures that the vectors are large enough that we do not have to worry about failure when taking the dot product.  On the other hand, I think once it gets into the heart of the algorithm there are various parts that are not robust to very small values, where for example a 180 degree turn, which this prevents from throwing an exception, might still result in very small time steps between switching points, causing failure during one of the integration methods. 
```
for (size_t j = 0; j < num_joints; j++)
{
  if (p > 0 && std::abs(new_point[j] - points.back()[j]) > 0.001)
    diverse_point = true;
}
```

mlautman:
@rhaschke and @Dale-Koenig Any reason to not merge this?

PR #1862, opened 2020-01-23
Use incoming joint state for computing command frame transform

henningkayser:
The previous implementation allowed the returned command transforms to be altered by the current jogging target. The reason is that joint_state_ represents JogArm's internal state representation (which is the future target state) and not the actual current state. This makes it difficult to produce accurate motions in closed-loop controlled applications.

@AndyZe do you think it makes sense to still provide access to JogArm's internal state?


AndyZe:
Theoretically the difference should be small but I can see how updating with an inaccurate joint_state_ can cause instabilities. I think this is a good idea. Testing now.

moveit_experimental/moveit_jog_arm/src/jog_cpp_interface.cpp
```diff
@@ -215,10 +215,8 @@ sensor_msgs::JointState JogCppApi::getJointState()
 
 Eigen::Isometry3d JogCppApi::getCommandFrameTransform()
 {
-  shared_variables_mutex_.lock();
-  Eigen::Isometry3d tf_moveit_to_cmd_frame = shared_variables_.tf_moveit_to_cmd_frame;
-  shared_variables_mutex_.unlock();
+  const std::lock_guard<std::mutex> lock(shared_variables_mutex_);
```
AndyZe: Nit: remove this empty line

henningkayser:
Fixed by https://github.com/ros-planning/moveit/pull/1871

PR #1863, opened 2020-01-23
[jog_arm] SRDF velocity and acceleration limit enforcement

AndyZe:
Read velocity and acceleration limits from the SRDF and enforce them

### DIFF
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -91,6 +91,7 @@ protected:
   // Suddenly halt for a joint limit or other critical issue.
   // Is handled differently for position vs. velocity control.
   void suddenHalt(trajectory_msgs::JointTrajectory& joint_traj);
+  void suddenHalt(Eigen::ArrayXd& delta_theta);
 
   void publishWarning(bool active) const;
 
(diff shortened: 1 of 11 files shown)


AndyZe:
The problem with the tests (I think) is that it's a fairly large integration test, and it includes several sub-tests that exceed the 60s timeout. There are some small improvements I can make to get is passing for now.

henningkayser:
@AndyZe Please rebase

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -142,7 +143,7 @@ class JogCalcs
 
   robot_state::RobotStatePtr kinematic_state_;
 
-  sensor_msgs::JointState joint_state_, original_joint_state_;
+  sensor_msgs::JointState incoming_joint_state_, internal_joint_state_, original_joint_state_;
```
felixvd: Does the `internal_joint_state_` correspond to a "current" joint state?

AndyZe: What I mean is that it's used in calculations, it shouldn't be relied on to be accurate. Would there be a better way to say that?

I'm gonna try to get this list down to 2. It seems like incoming_joint_state_ and original_joint_state_ should be the same thing.

AndyZe:
The test that was added (test_vel_accel_limits.py) will make more sense when I can check for specific error codes. For now, it's just a placeholder.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
AndyZe: It turns out that `original_joint_state_` is a little bit different from `incoming_joint_state_` so I left a comment to explain it. `incoming_joint_state_` has every joint, but passive joints might get dropped from `original_joint_state_`

PR #1757, opened 2019-11-20
Integration test to defend subframe tutorial (and bugfix for subframes of attached collision objects)

tylerjw:
### Description

Issue: #1733
This PR is an integration test for the subframe tutorial that tests that the tutorial continues to work.

Future work:
* test orientation

Thank you @jaas14 for helping me write this.

PR #1866, opened 2020-01-24
Fix subframes disappearing when object is detached/scaled/renamed

felixvd:
### Description

Subframes were being forgotten when objects were detached/removed from the robot, and when they were moved. They also would have overwritten the old subframes during an APPEND call. This PR should fix all of that, and adds a line to the Rviz plugin that says "The object has subframes X, Y, Z."

Subframes are not updated if the object is moved from the Rviz plugin. I am filing another PR about this in a minute.

### Checklist

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1607,14 +1607,14 @@ bool PlanningScene::processAttachedCollisionObjectMsg(const moveit_msgs::Attache
       const moveit::core::AttachedBody* body = robot_state_->getAttachedBody(object.object.id);
       if (body)
       {
-        attached_bodies.push_back(body);
-        if ((body->getAttachedLinkName() != object.link_name) && (!object.link_name.empty()))
+        if (!object.link_name.empty() && (body->getAttachedLinkName() != object.link_name))
         {
-          ROS_ERROR_NAMED(LOGNAME, "The AttachedCollisionObject message stated that the object is attached to ",
-                          object.link_name, ", but it was actually attached to ", body->getAttachedLinkName(),
-                          ". Did something go wrong? Aborting and leaving the body attached.");
+          ROS_ERROR_NAMED(LOGNAME, "The AttachedCollisionObject message states the object is attached to ",
+                          object.link_name, ", but it is actually attached to ", body->getAttachedLinkName(),
+                          ". Leave the link_name empty or specify the correct link.");
           return false;
         }
+        attached_bodies.push_back(body);
       }
     }
 
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1608,10 +1608,13 @@ bool PlanningScene::processAttachedCollisionObjectMsg(const moveit_msgs::Attache
       if (body)
       {
         attached_bodies.push_back(body);
-        if (body->getAttachedLinkName() != object.link_name)
-          ROS_WARN_NAMED(LOGNAME, "The AttachedCollisionObject message stated that the object is attached to ",
-                         object.link_name, ", but it was actually attached to ", body->getAttachedLinkName(),
-                         ". Did something go wrong?");
+        if ((body->getAttachedLinkName() != object.link_name) && (!object.link_name.empty()))
+        {
+          ROS_ERROR_NAMED(LOGNAME, "The AttachedCollisionObject message stated that the object is attached to ",
+                          object.link_name, ", but it was actually attached to ", body->getAttachedLinkName(),
+                          ". Did something go wrong? Aborting and leaving the body attached.");
+          return false;
+        }
       }
     }
 
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1609,9 +1609,10 @@ bool PlanningScene::processAttachedCollisionObjectMsg(const moveit_msgs::Attache
       {
         if (!object.link_name.empty() && (body->getAttachedLinkName() != object.link_name))
         {
-          ROS_ERROR_NAMED(LOGNAME, "The AttachedCollisionObject message states the object is attached to ",
-                          object.link_name, ", but it is actually attached to ", body->getAttachedLinkName(),
-                          ". Leave the link_name empty or specify the correct link.");
+          ROS_ERROR_STREAM_NAMED(LOGNAME, "The AttachedCollisionObject message states the object is attached to "
+                                              << object.link_name << ", but it is actually attached to "
+                                              << body->getAttachedLinkName()
+                                              << ". Leave the link_name empty or specify the correct link.");
           return false;
         }
         attached_bodies.push_back(body);
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1593,44 +1593,26 @@ bool PlanningScene::processAttachedCollisionObjectMsg(const moveit_msgs::Attache
   {
     // STEP 1: Get info about the object from the RobotState
     std::vector<const moveit::core::AttachedBody*> attached_bodies;
-    if (object.link_name.empty())
+    if (object.object.id.empty())
     {
-      if (object.object.id.empty())
-        robot_state_->getAttachedBodies(attached_bodies);
+      const moveit::core::LinkModel* link_model =
+          object.link_name.empty() ? nullptr : getRobotModel()->getLinkModel(object.link_name);
+      if (link_model)  // if we have a link model specified, only fetch bodies attached to this link
+        robot_state_->getAttachedBodies(attached_bodies, link_model);
       else
-      {
-        const moveit::core::AttachedBody* body = robot_state_->getAttachedBody(object.object.id);
-        if (body)
-          attached_bodies.push_back(body);
-      }
+        robot_state_->getAttachedBodies(attached_bodies);
     }
-    else
+    else  // A specific object id will be removed.
     {
-      const moveit::core::LinkModel* link_model = getRobotModel()->getLinkModel(object.link_name);
-      if (link_model)
-      {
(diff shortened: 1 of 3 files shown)


felixvd:
The tests on this PR succeed, only the code coverage is complaining. The tests for the functions affected by this should be implemented in https://github.com/ros-planning/moveit/pull/1757.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1835,12 +1838,32 @@ bool PlanningScene::processCollisionObjectMove(const moveit_msgs::CollisionObjec
...
+        std::string name = object.subframe_names[i];
+        subframe_poses[name] = t * frame_pose;
+      }
+    }
+    else if (!obj->subframe_poses_.empty())  // If there were subframes but no new poses are given
+    {
+      ROS_WARN_NAMED(LOGNAME, "Object '%s' had subframes, but no new ones were given for Move operation. ",
+                     object.id.c_str());
+    }
+
```
v4hn: i am not happy with the need to have the names vector transmitted here, but there's no way around that.
at least, it should be a normal use case of this to specify no subframes, even if the world object includes some.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1835,12 +1838,32 @@ bool PlanningScene::processCollisionObjectMove(const moveit_msgs::CollisionObjec
...
     if (obj->shapes_.size() == new_poses.size())
     {
       std::vector<shapes::ShapeConstPtr> shapes = obj->shapes_;
       obj.reset();
       world_->removeObject(object.id);
       world_->addToObject(object.id, shapes, new_poses);
+      world_->setSubframesOfObject(object.id, subframe_poses);
```
v4hn: isn't updating the subframes logically independent of specifying poses for everything else?

i can't help but think the use case you want to allow here could be much nicer included in APPEND.
maybe we could just add subframes to the warning in the beginning that says they will be ignored?

rhaschke: My first thought was the very same as Michael's here: Why should we couple an update of shapes to the updates of subframes.
On the other hand, moving shapes, typically also invalidates subframes and thus it makes sense to update the latter in the same call. The present code essentially replicates the corresponding code of the `APPEND` mode. A warning is issued when shapes have moved, but no updates (at all) were specified for subframes.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1822,12 +1807,32 @@ bool PlanningScene::processCollisionObjectMove(const moveit_msgs::CollisionObjec
     }
 
     collision_detection::World::ObjectConstPtr obj = world_->getObject(object.id);
+
+    // Update subframe poses
+    moveit::core::FixedTransformsMap subframe_poses = obj->subframe_poses_;
```
rhaschke: This whole code snipped on subframe updates is only relevant if a shape update is actually performed below.
Hence, I think this code should be moved into the if condition below and a warning should be issued if a user just tries to move subframes (but no shapes).

moveit_core/planning_scene/src/planning_scene.cpp
v4hn: > The present code essentially replicates the corresponding code of the APPEND mode.

I would argue the code should sanitize the subframes list and forbid adding new frames here.
After all, if you want to add new frames, you can use `APPEND`.

moveit_core/planning_scene/src/planning_scene.cpp
v4hn: The data structures actually allow to handle the additional use-case to specify *only* subframe poses and no shapes.
It's not really an important use-case (maybe for tcp online estimation?), but I think we could include it here?
Either way, the current patch need updating as it mixes both interpretations, doing the computations independent and then update them only if new shape poses exist...

rhaschke: If you just want to update subframes, you better use APPEND mode.

felixvd:
I split this as requested without any other changes for now.

v4hn:
@felixvd is (part of) this still unmerged? could you resolve the conflict?

moveit_core/planning_scene/src/planning_scene.cpp
felixvd: Yes, the MOVE operation shouldn't add anything to the object. I removed the whole if block and just updated the world object.

moveit_core/planning_scene/src/planning_scene.cpp
felixvd: I removed this because the logic will change in #2037

felixvd:
I updated and rebased this. The PR now only fixes the problem that subframes disappear when objects are detached objects from the robot, or scaled/renamed in Rviz. The changes to the MOVE operation will be in #2037 instead.

felixvd:
Ping

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -122,6 +123,7 @@ void MotionPlanningFrame::sceneScaleChanged(int value)
           ps->getWorldNonConst()->addToObject(scaled_object_->id_, shapes::ShapeConstPtr(s),
                                               scaled_object_->shape_poses_[i]);
         }
+        ps->getWorldNonConst()->setSubframesOfObject(scaled_object_->id_, subframes);
```
v4hn, line 126: This might be a matter of personal preference, but shouldn't the subframes be scaled as well?
I guess there is no reasonable way to do that until your main pose request is merged...
Could you add a TODO in the code here though?

moveit_core/robot_state/include/moveit/robot_state/attached_body.h
```diff
@@ -120,6 +120,12 @@ class AttachedBody
     return subframe_poses_;
   }
 
+  /** \brief Get subframes of this object (relative to the link) */
```
v4hn, line 123: This is probably just another issue with your ongoing pose work, but why are *global* frames documented as *relative to the link*? Maybe extend/change the comment?

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1593,51 +1593,35 @@ bool PlanningScene::processAttachedCollisionObjectMsg(const moveit_msgs::Attache
...
-        else  // A specific object id will be removed.
-        {
-          const moveit::core::AttachedBody* body = robot_state_->getAttachedBody(object.object.id);
-          if (body)
-            attached_bodies.push_back(body);
-        }
-      }
+      const moveit::core::AttachedBody* body = robot_state_->getAttachedBody(object.object.id);
+      if (body)
+        attached_bodies.push_back(body);
```
v4hn, line 1609: Could you please check for whether the id is actually attached to `object.link_name` and trigger a warning if that's not the case?
The previous version would simply ignore the object if a wrong link is specified, but a warning is of course preferred. :)

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1773,7 +1757,7 @@ bool PlanningScene::processCollisionObjectAdd(const moveit_msgs::CollisionObject
     setObjectType(object.id, object.type);
 
   // Add subframes
```
v4hn, line 1759:
```suggestion
  // Add subframes to the newly created or (possibly) modified object
```
It took me a moment to understand why this `world_->getObject(id)` call always returns a valid object.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
felixvd: Good idea

moveit_core/planning_scene/src/planning_scene.cpp
felixvd: Done. Yes, the world object structure is a little awkward.

moveit_core/robot_state/include/moveit/robot_state/attached_body.h
felixvd: I think I just forgot to update this comment, my bad.

moveit_core/planning_scene/src/planning_scene.cpp
felixvd: Sure, seems like a rare error though. Please have a look and confirm that this is what you meant.

v4hn: The condition is good, but I would say we either treat it as a proper error the user has to fix before they can continue (so the method should return), or we specify in the warning that the object attached to the other link is used instead.

felixvd: I don't know which way will be better, but I chose to make it an error and fixed the condition (it would have fired for the default value otherwise).

v4hn:
> Perfect

Not really. I just had to fix one misuse of `ROS_ERROR_NAMED`.

The second review might take some time because Robert is on vacation.
That is to say, unless some other maintainer is willing to look at this. :-)

felixvd:
I see. I figured it would also concatenate. The more you know

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -159,6 +159,7 @@ void MotionPlanningFrame::sceneScaleChanged(int value)
     {
       if (ps->getWorld()->hasObject(scaled_object_->id_))
       {
+        const moveit::core::FixedTransformsMap subframes = scaled_object_->subframe_poses_;  // Keep subframes
```
tylerjw, line 162: [nit] Do you need to copy this to a local if it is const?  Could this be a `const&`?

Or maybe it shouldn't be const if you are planning on modifying it when you scale it?

tylerjw: :+1:

PR #1867, opened 2020-01-24
WIP: Moving objects with multiple shapes and subframes in Rviz

felixvd:
### Description

I noticed that objects with multiple shapes cannot be moved from the Rviz plugin interface, and that subframes are also not moved when moving objects through it. I was playing around with some solutions and would like to hear opinions about this one. Instead of moving only the shape of the object via `moveShapeInObject`, the plugin calls a new function `moveObjectAbsolute` that applies the same relative transformation to all of the shapes and subframes.

I am not a particularly big fan of the fact that the relative poses of the shapes and subframes will accumulate rounding errors with this approach, but without a full scene graph/transformation tree, this might be a reasonable workaround. 

I also noticed that the `MOVE` operation for `CollisionObject`s is not very useful as is, since the positions of all the components have to be updated manually. That sort of defeats the purpose, in my opinion. I was thinking of a `MOVE_REL` or `MOVE_ABS` command, in the absence of a scene graph. This discussion might be suited for a different issue.

This code also currently has a bug: the position of the object changes after a few moves, as if errors accumulated, but I don't see how the "main" object could, since the target position is always set relative to its frame of reference.

I use [this modified version of the subframes tutorial](https://gist.github.com/felixvd/857263192f62526c82db68f2a4d57218) to test.

### Checklist

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/world.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/world.h
@@ -178,8 +178,15 @@ public:
   bool moveShapeInObject(const std::string& object_id, const shapes::ShapeConstPtr& shape,
                          const Eigen::Isometry3d& pose);
 
-  /** \brief Move all shapes in an object according to the given transform specified in world frame */
+  /** \brief Move all shapes and subframes in an object according to the given transform specified in world frame.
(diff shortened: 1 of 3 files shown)


felixvd:
Superceded by https://github.com/ros-planning/moveit/pull/2037

Issue #1722, opened 2019-11-11
Python 3 support


v4hn:
The MoveIt Python API is not yet fully functional with Python 3 and scarcely tested.
As one example, `rosrun moveit_commander moveit_commander_cmdline.py` does not provide a usable shell.

It only makes sense to look into this issue, if you have previous experience with ROS on Python 3.
This is not the default configuration of ROS and you will need to build many standard packages manually to attempt it.

Briancbn:
`moveit_ros_planning_interface` package needs editing to generate the python3 APIs

I tried to compile the whole moveit using python3 by running by setting the catkin config:
```
catkin config -DPYTHON_EXECUTABLE=/usr/bin/python3 -DPYTHON_INCLUDE_DIR=/usr/include/python3.6m -DPYTHON_LIBRARY=/usr/lib/x86_64-linux-gnu/libpython3.6m.so -DCMAKE_BUILD_TYPE=Release
```
Then I also manually add in the python3 directories into the PYTHONPATH
```
export PYTHONPATH=/home/rosi/Documents/moveit_ws/devel/lib/python3/dist-packages:$PYTHONPATH
```
I edit the first line of the `moveit_commander_cmdline.py` to `#!/usr/bin/env python3`run the `rosrun moveit_commander moveit_commander_cmdline.py` command. It gives me a segfault, it comes from importing the packages from `moveit_ros_planning_interface`

130s:
“I’ll work on this for WMD 2020”. However I can't guarantee full-day commit so I'd appreciate company.
I've done some Python3 conversion from Python2.

lorepieri8:
Any progress on this?

v4hn:
Please try to use the APIs in ROS noetic and report errors if you should find any.
The state of the python3 migration is unclear to me, but with ROS noetic releases it is supposed to work now.

lorepieri8:
Where I can find the python API? 

Besides https://ros-planning.github.io/moveit_tutorials/doc/move_group_python_interface/move_group_python_interface_tutorial.html where I can I look at? The source code?

felixvd:
There is also the auto-generated code API documentation:  
[MoveGroupInterface](http://docs.ros.org/en/api/moveit_python/html/classmoveit__python_1_1planning__scene__interface_1_1PlanningSceneInterface.html)  
http://docs.ros.org/en/api/moveit_python/html/classmoveit__python_1_1move__group__interface_1_1MoveGroupInterface.html
[...]

PR #1869, opened 2020-01-29
Python 3 friendly message serialization

gleichdick:
### Description

Messages are passed between C++ and Python via wrapper classes written in C++ and boost::python. These messages are (de-) serialized on both sides. With Python 3, `std::string` are implicitly converted to `Unicode` objects by boost which does not work with arbitrary binary data.
When compiled with Python 3, a new C++ class (`moveit::py_bindings_tools::ByteString`) builds a bridge between `std::string` and Python 3 `Bytes` objects. Nothing is changed under Python 2.

Related: #1722 

### Checklist

### DIFF
--- a/moveit_commander/bin/moveit_commander_cmdline.py
+++ b/moveit_commander/bin/moveit_commander_cmdline.py
@@ -42,7 +42,7 @@ class SimpleCompleter(object):
         prefix = ""
         if len(cmds) > 0:
             prefix = cmds[0]
-            if not self.options.has_key(prefix):
+            if not prefix in self.options:
                 prefix = ""
 
         if state == 0:
--- a/moveit_commander/src/moveit_commander/conversions.py
+++ b/moveit_commander/src/moveit_commander/conversions.py
@@ -36,8 +36,8 @@ try:
     # Try Python 2.7 behaviour first
     from StringIO import StringIO
 except ImportError:
-    # Use Python 3.x behaviour as fallback
-    from io import StringIO
+    # Use Python 3.x behaviour as fallback and choose the non-unicode version
+    from io import BytesIO as StringIO
 
 from moveit_commander import MoveItCommanderException
 from geometry_msgs.msg import Pose, PoseStamped, Transform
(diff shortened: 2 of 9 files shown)


gleichdick:
Thank you for your fast reply. The reason why I renamed the `serializeMsg` method is that the return type of _templated_ C++ functions is not used for choosing the correct overload. But renaming the old method to prevent this issue would break the public API.
Another proposal: Python 3 proof (de)serialization will take place in member functions of `ByteString` and the old (de)serializeMsg functions will be deprecated in noetic.
```c++
class ByteString : public bp::object {
  template <class T>
  explicit ByteString(const T& t) {
     /* convert PyBytes from C++ message */
  }
  template <class T>
  void deserialize(T& t) {
    /* convert C++ message from PyBytes */
  }
};
```
Btw, I just found out that all `PyBytes` methods are also available with Python 2.7, so we can get rid of the compiler switches.

rhaschke:
> Renaming the old method to prevent this issue would break the public API.

As far as I understand, the string-based (de)serialization methods will not be used (directly) in Python 3 anymore, but only through their ByteString pendants. That is, the string-based methods are obsolete anyway.
If so, we can make the old string-based functions non-public. For Python 2, ByteString equals std::string and thus there is no API/ABI? difference.

gleichdick:
Oh, I was thinking about anyone else use the `py_bindings_tools` C++ code to glue their own Python/C++ modules together. Removing anything in e.g. serialize_msg.h might break their code.

rhaschke:
As far as I understand, in Python 3 the old, string-based functions serializeMsg and deserializeMsg will not work anymore - due to unicode/byte issues. This applies to any MoveIt-external glue code as well, doesn't it?
Your current proposal replaces _all_ usages of serializeMsg/deserializeMsg in MoveIt with the new functions serializeMsgByteString/deserializeMsg(ByteString), right? A corresponding code adaptation would be necessary for any other code as well that is using this C++ 🡘 python data transfer approach.
So, to reduce this effort to renaming std::string 🡒 ByteString, my suggestion was to keep the original function name.

rhaschke:
I think this PR should target the master branch.

gleichdick:
Okay, thanks, I agree. Rebased onto master, see #1870. Can we close this PR?

rhaschke:
Thanks. Closing here in favour of #1870.

PR #1871, opened 2020-02-02
[jog_arm] Fix initial end effector transform jump

AndyZe:
- Fix a bug where the end effector transform jumped initially, when the first jog command is received.

- Reduce duplicate code between joint jogging and Cartesian jogging.

- Remove the joint velocity low-pass filters. I justify this because joint velocities are derived from joint positions (which do get filtered).

- This required some re-ordering of the algorithm. I think the flow makes a lot more sense now.

### DIFF
--- a/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
@@ -328,7 +328,11 @@ bool JogCalcs::cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, JogArmShared&
   delta_theta_ = pseudo_inverse_ * delta_x;
 
   // If close to a collision or a singularity, decelerate
-  applyVelocityScaling(shared_variables, mutex, delta_theta_, decelerateForSingularity(delta_x, svd_));
+  if (!applyVelocityScaling(shared_variables, mutex, delta_theta_, decelerateForSingularity(delta_x, svd_)))
+  {
+    has_warning_ = true;
+    suddenHalt(outgoing_command_);
+  }
 
   return convertDeltasToOutgoingCmd();
 }
@@ -368,12 +372,11 @@ bool JogCalcs::convertDeltasToOutgoingCmd()
   if (!enforceSRDFJointBounds(outgoing_command_))
   {
     suddenHalt(outgoing_command_);
-    publishWarning(true);
+    has_warning_ = true;
   }
-  else
-  {
-    publishWarning(false);
-  }
+
+  publishWarning(has_warning_);
+  has_warning_ = false;
 
   // done with calculations
   if (parameters_.use_gazebo)
@@ -442,7 +445,7 @@ trajectory_msgs::JointTrajectory JogCalcs::composeJointTrajMessage(sensor_msgs::
 
 // Apply velocity scaling for proximity of collisions and singularities.
 // Scale for collisions is read from a shared variable.
-void JogCalcs::applyVelocityScaling(const JogArmShared& shared_variables, std::mutex& mutex,
+bool JogCalcs::applyVelocityScaling(const JogArmShared& shared_variables, std::mutex& mutex,
                                     Eigen::ArrayXd& delta_theta, double singularity_scale)
 {
   mutex.lock();
@@ -450,6 +453,9 @@ void JogCalcs::applyVelocityScaling(const JogArmShared& shared_variables, std::m
   mutex.unlock();
 
   delta_theta = collision_scale * singularity_scale * delta_theta;
+
+  // Heuristic: flag that we are stuck if velocity scaling is < X%
+  return collision_scale * singularity_scale >= 0.1;
 }
 
 // Possibly calculate a velocity scaling factor, due to proximity of singularity and direction of motion
--- a/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
@@ -45,7 +45,7 @@ namespace moveit_jog_arm
 {
 // Constructor for the class that handles jogging calculations
 JogCalcs::JogCalcs(const JogArmParameters& parameters, const robot_model_loader::RobotModelLoaderPtr& model_loader_ptr)
-  : parameters_(parameters)
+  : parameters_(parameters), default_sleep_rate_(1000)
 {
   // Publish collision status
   warning_pub_ = nh_.advertise<std_msgs::Bool>(parameters_.warning_topic, 1);
@@ -54,7 +54,7 @@ JogCalcs::JogCalcs(const JogArmParameters& parameters, const robot_model_loader:
   while (ros::ok() && !model_loader_ptr)
   {
     ROS_WARN_THROTTLE_NAMED(5, LOGNAME, "Waiting for a non-null robot_model_loader pointer");
-    ros::Duration(WHILE_LOOP_WAIT).sleep();
+    default_sleep_rate_.sleep();
   }
   const robot_model::RobotModelPtr& kinematic_model = model_loader_ptr->getModel();
   kinematic_state_ = std::make_shared<robot_state::RobotState>(kinematic_model);
@@ -65,8 +65,10 @@ JogCalcs::JogCalcs(const JogArmParameters& parameters, const robot_model_loader:
 
 void JogCalcs::startMainLoop(JogArmShared& shared_variables, std::mutex& mutex)
 {
-  // Reset loop termination flag
-  stop_requested_ = false;
+  // Reset flags
+  stop_jog_loop_requested_ = false;
+  halt_outgoing_jog_cmds_ = false;
+  is_initialized_ = false;
 
   // Wait for initial messages
   ROS_INFO_NAMED(LOGNAME, "jog_calcs_thread: Waiting for first joint msg.");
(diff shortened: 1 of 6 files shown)


moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -111,9 +108,28 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables, std::mutex& mutex)
...
+    incoming_joints_ = shared_variables.joints;
+    mutex.unlock();
+
+    kinematic_state_->setVariableValues(joint_state_);
+
+    // Always update the end-effector transform in case the getCommandFrameTransform() method is being used
+    // Get the transform from MoveIt planning frame to jogging command frame
+    // We solve (planning_frame -> base -> robot_link_command_frame)
+    // by computing (base->planning_frame)^-1 * (base->robot_link_command_frame)
+    tf_moveit_to_cmd_frame_ = kinematic_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
```
henningkayser: Does this really fix the issue or is it just much more unlikely to occur? To me it seems like there is still the possibility of a race condition when getCommandFrameTransform() is called before we have an updated joint state. Also, wouldn't it be better to compute `tf_moveit_to_cmd_frame` inside getCommandFrameTransform() in order to reduce redundant computations? I don't see the transform being used here anymore. Maybe we should additionally let `getCommandFrameTransform()` return bool or add a blocking function `waitForValidJointState()`.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -124,25 +140,19 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables, std::mutex& mutex)
...
+    // Always update the joints and end-effector transform for 2 reasons:
+    // 1) in case the getCommandFrameTransform() method is being used
+    // 2) so the low-pass filters are up to date and don't cause a jump
     mutex.lock();
     incoming_joints_ = shared_variables.joints;
     mutex.unlock();
-
+    while (!updateJoints() && ros::ok() && !stop_requested_)
+    {
+      ros::Duration(WHILE_LOOP_WAIT).sleep();
```
henningkayser: Maybe unrelated to this PR, but is there a reason we are not using `ros::Rate`?

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -153,91 +163,96 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables, std::mutex& mutex)
...
-      mutex.lock();
-      // If everything normal, share the new traj to be published
-      if (valid_nonzero_command)
+      // Prioritize cartesian jogging above joint jogging
+      if (!zero_cartesian_cmd_flag)
       {
-        shared_variables.outgoing_command = outgoing_command_;
-        shared_variables.ok_to_publish = true;
+        mutex.lock();
+        cartesian_deltas = shared_variables.command_deltas;
```
henningkayser: All these mutexes make the code unnecessarily hard to read. We should add thread safe accessor functions to shared_variables

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -153,91 +163,96 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables, std::mutex& mutex)
...
-          ++zero_velocity_count;
+        suddenHalt(outgoing_command_);
+        zero_cartesian_cmd_flag = true;
+        zero_joint_cmd_flag = true;
+      }
+
+      bool valid_nonzero_command = !zero_cartesian_cmd_flag || !zero_joint_cmd_flag;
+
+      // Send the newest target joints
+      if (!outgoing_command_.joint_names.empty())
```
henningkayser: Are there ever no joint names specified in the message at this point?

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -153,91 +163,96 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables, std::mutex& mutex)
     shared_variables.tf_moveit_to_cmd_frame = tf_moveit_to_cmd_frame_;
     mutex.unlock();
 
-    // Initialize the position filters to initial robot joints
-    while (!updateJoints() && ros::ok() && !stop_requested_)
+    // Only do jogging calculations if the robot should move, for efficiency
+    if (stop_requested_)
```
henningkayser: Am I reading this wrong or does `stop_requested_` now only mean that the positions are reset, but the loop is kept alive? That's not what `stop_requested_` is supposed to do, we need an API to actually stop the loop as well. Stopping should be done by resetting the velocities in the command.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: Line 94 ensures joint_state_ is updated before this, so no race condition.

AndyZe: Yeah, I guess it does make sense to calculated `tf_moveit_to_cmd_frame` inside the getter method.

AndyZe: However, I did notice that `tf_planning_to_cmd_frame` was redundant with `tf_moveit_to_cmd_frame_` and could be deleted

AndyZe: I added an `isInitialized()` function and `std::atomic<bool> is_initialized_` like you added previously for `stop_requested_`

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: Not that I know of.  ;)  Will update it

AndyZe: I found out there is a reason. The default sleep rate is defined in jog_arm_data.h:

`static const double WHILE_LOOP_WAIT = 0.001;`

If I change that to `static ros::Rate DEFAULT_SLEEP_RATE(1000)'` the error is:

Cannot use ros::Time::now() before the first NodeHandle has been created or ros::start() has been called

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: OK, but in a different PR, right?

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: not that i know of, will try deleting it

AndyZe: deleting it _seems_ fine...

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
jliukkonen: It should go a long way if you refactored JogArmShared into a class with mutexed setters and getters. That should ensure no race conditions can occur. Also note that some of the variables can just be converted into atomics.

Another thing I would recommend, is to use `std::lock_guard<foo>`. The thing about locking and unlocking is, that eventually you forget to unlock and there will be problems. Lock guard always unlocks the mutex at the end of scope so you can't really forget anything. You can also do this: 
```
std::lock_guard<std::mutex> lock(mutex)
{
  // Do things here
} // automatic unlocking.
```

In other words,, you set a temporary scope just to guide the lock_guard.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -153,91 +163,96 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables, std::mutex& mutex)
...
+          shared_variables.ok_to_publish = true;
+        }
+        mutex.unlock();
+
+        // Store last zero-velocity message flag to prevent superfluous warnings.
+        // Cartesian and joint commands must both be zero.
+        if (zero_cartesian_cmd_flag && zero_joint_cmd_flag)
+        {
+          // Avoid overflow
+          if (zero_velocity_count < INT_MAX)
```
jliukkonen: You can use `std::numeric_limits<int>::max()`. That will also warn about incompatible types. With macros I'm not sure whether e.g., signed / unsigned comparisons are caught or not.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: The robot will not move if `stop_requested_` is true but the node will continue to make the end_effector_transform available (for closed loop controllers). It does look like the usage of `stop_requested_` needs updating in some places, though.

I think the way to exit from the loop is to destroy the object. Otherwise, we want it to continue to spin and provide the end effector transform (unless getEndEffectorTransform() should not be provided by this node.)

Over a year ago, the design decision was that you shouldn't need to publish zeros to stop motion. Because
a) ROS1 messages can be dropped
b) if you're spamming zeros to the controller, you shut down other ways of moving the robot. People may have multiple motion control modes, not just jogging. E.g. trajectory control and joint jogging possibly could operate with the same ros_control controller

AndyZe: I guess this confusion was due to the naming of the variable. How about renaming:

`stop_requested_ -> exit_requested_`

To signify that it's not just meant to halt, but to exit.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: I'll work on this in a followup PR. This one's already pretty big

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
henningkayser: I thought it would be enough to just access `shared_variables.joints` for getting the latest state. Anyway, this doesn't really belong into this PR and I'm fine with the current solution

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
henningkayser: Yes, ros::Rate can't be static, you'll have to initialize it before the loop (or inside constructor). A big advantage is that ros::Rate will run the loop in the rate you specified while `ros::Duration::sleep()` adds a constant sleep additional to the computation time.

AndyZe: ok, done

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
jliukkonen: I'm going to approve this so we can move on to "fix mutexes" state :)

PR #1872, opened 2020-02-03
add warnings to jog_arm low_pass filter

tylerjw:
### Description

These warnings and errors handle warning the user of bad input parameter values for the low pass filter in jog arm.  Here is the function of the feedback magnitude with respect to the `low_pass_filter_coeff` value: https://www.wolframalpha.com/input/?i=y+%3D+%281%2F%281%2Bx%29%29+*+%281-x%29

Here is a small python script that shows why these values are problematic:

```python
def run_filter(mu, length):
  x0 = x1 = y1 = 0

  y = x = [0 for i in range(length)]
  x[0] = 1

  for i in range(length):
    x1 = x0
    x0 = x[i]

    y[i] = (1.0 / (1.0 + mu)) * (x1 + x0 + (1.0 - mu)*y1)
    y1 = y[i]

  print(mu, sum(y))

run_filter(2.0, 1000)
run_filter(1.0, 1000)
run_filter(0.5, 1000)
run_filter(0, 1000)
run_filter(-0.5, 1000)
run_filter(-0.999, 1000)
run_filter(-2, 1000)

```

output:
```
(2.0, 0.5000000000000002)
(1.0, 1.0)
(0.5, 1.9999999999999991)
(0, 1999.0)
(-0.5, inf)
(-0.999, inf)
(-2, nan)
```

### Checklist

### DIFF
--- a/moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
@@ -39,6 +39,7 @@
 
 #include <moveit_jog_arm/low_pass_filter.h>
 #include <string>
+#include <ros/ros.h>
 
 static const std::string LOGNAME = "low_pass_filter";
 
@@ -47,6 +48,18 @@ namespace moveit_jog_arm
 LowPassFilter::LowPassFilter(double low_pass_filter_coeff)
 {
   filter_coeff_ = low_pass_filter_coeff;
+  if (low_pass_filter_coeff == 1.0)
+  {
+    ROS_WARN_NAMED(LOGNAME, "LowPassFilter constructed as a 2 tap boxcar FIR");
+  }
+  else if (low_pass_filter_coeff == -1.0)
+  {
+    ROS_FATAL_NAMED(LOGNAME, "LowPassFilter coeff value of -1.0 will result in divide by zero error");
+  }
+  else if (low_pass_filter_coeff < 1.0)
+  {
+    ROS_ERROR_NAMED(LOGNAME, "LowPassFilter constructed as an unstable IIR");
+  }
 }
 
 void LowPassFilter::reset(double data)
--- a/moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
@@ -38,29 +38,45 @@
  */
 
 #include <moveit_jog_arm/low_pass_filter.h>
+#include <cmath>
 #include <string>
 #include <ros/ros.h>
 
-static const std::string LOGNAME = "low_pass_filter";
+const std::string LOGNAME = "low_pass_filter";
+constexpr double EPSILON = 1e-9;
 
 namespace moveit_jog_arm
 {
 LowPassFilter::LowPassFilter(double low_pass_filter_coeff)
-  : filter_coeff_(low_pass_filter_coeff)
+  : previous_measurements_{ 0., 0. }
+  , previous_filtered_measurement_(0.)
+  , filter_coeff_(low_pass_filter_coeff)
   , scale_term_(1. / (1. + low_pass_filter_coeff))
   , feedback_term_(-low_pass_filter_coeff + 1.)
 {
-  if (filter_coeff_ == 1.0)
+  // guarentee this doesn't change because the logic depends on this length implicity
+  static_assert(LowPassFilter::FILTER_LENGTH == 2);
+
+  if (std::abs(feedback_term_) < EPSILON)
   {
-    ROS_WARN_NAMED(LOGNAME, "LowPassFilter constructed as a 2 tap boxcar FIR");
+    ROS_WARN_STREAM_NAMED(
+        LOGNAME, "Filter coefficient value of "
+                     << filter_coeff_
+                     << " resulted in feedback term of 0. "
+                        " This results in a window averaging Finite Impulse Response (FIR) filter with a gain of "
+                     << scale_term_ * LowPassFilter::FILTER_LENGTH);
   }
-  else if (filter_coeff_ == -1.0)
+  if (std::isinf(scale_term_))
   {
-    ROS_FATAL_NAMED(LOGNAME, "LowPassFilter coeff value of -1.0 will result in nan outputs");
+    ROS_FATAL_STREAM_NAMED(LOGNAME, "Filter coefficient value of "
+                                        << filter_coeff_
+                                        << ", outputs from filter will be inf because denominator of scale is 0");
   }
-  else if (filter_coeff_ < 1.0)
+  if (filter_coeff_ < 1.)
   {
-    ROS_ERROR_NAMED(LOGNAME, "LowPassFilter constructed as an unstable IIR");
+    ROS_ERROR_STREAM_NAMED(
+        LOGNAME,
+        "Filter coefficient < 1. makes single order Butterworth an unstable Infinite Impulse Response (IIR) filter, ");
   }
 }
 
--- a/moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
@@ -55,7 +55,7 @@ LowPassFilter::LowPassFilter(double low_pass_filter_coeff)
   , feedback_term_(1. - low_pass_filter_coeff)
 {
   // guarantee this doesn't change because the logic depends on this length implicity
-  static_assert(LowPassFilter::FILTER_LENGTH == 2);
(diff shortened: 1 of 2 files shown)


moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -47,6 +48,18 @@ namespace moveit_jog_arm
 LowPassFilter::LowPassFilter(double low_pass_filter_coeff)
 {
   filter_coeff_ = low_pass_filter_coeff;
+  if (low_pass_filter_coeff == 1.0)
+  {
+    ROS_WARN_NAMED(LOGNAME, "LowPassFilter constructed as a 2 tap boxcar FIR");
```
jliukkonen: Would it be better if the warning was tiny bit more explicit like this:
`"LowPassFilter constructed as a 2 tap boxcar FIR (Finite Impulse Response)."`?

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -47,6 +48,18 @@ namespace moveit_jog_arm
...
+  }
+  else if (low_pass_filter_coeff == -1.0)
+  {
+    ROS_FATAL_NAMED(LOGNAME, "LowPassFilter coeff value of -1.0 will result in divide by zero error");
+  }
+  else if (low_pass_filter_coeff < 1.0)
+  {
+    ROS_ERROR_NAMED(LOGNAME, "LowPassFilter constructed as an unstable IIR");
```
jliukkonen: Would it be better if the warning was tiny bit more explicit like this:
`"LowPassFilter constructed as an unstable IIR (Infinite Impulse Response)."`?

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -76,8 +78,8 @@ double LowPassFilter::filter(double new_measurement)
   previous_measurements_[1] = previous_measurements_[0];
   previous_measurements_[0] = new_measurement;
 
-  double new_filtered_msrmt = (1. / (1. + filter_coeff_)) * (previous_measurements_[1] + previous_measurements_[0] -
-                                                             (-filter_coeff_ + 1.) * previous_filtered_measurement_);
+  double new_filtered_msrmt = scale_term_ * (previous_measurements_[1] + previous_measurements_[0] -
```
jliukkonen: I'm sure we can afford couple more letters here :wink:. I propose `new_filtered_measurement`.

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -46,17 +46,19 @@ static const std::string LOGNAME = "low_pass_filter";
 namespace moveit_jog_arm
 {
 LowPassFilter::LowPassFilter(double low_pass_filter_coeff)
+  : filter_coeff_(low_pass_filter_coeff)
+  , scale_term_(1. / (1. + low_pass_filter_coeff))
+  , feedback_term_(-low_pass_filter_coeff + 1.)
```
jliukkonen: Consider changing the order here to make that minus harder to miss. 
Like this: `feedback_term_(1. - low_pass_filter_coeff)`.

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -38,15 +38,46 @@
...
+  {
+    ROS_FATAL_STREAM_NAMED(LOGNAME, "Filter coefficient value of "
+                                        << filter_coeff_
+                                        << ", outputs from filter will be inf because denominator of scale is 0");
+  }
+  if (filter_coeff_ < 1.)
+  {
+    ROS_ERROR_STREAM_NAMED(
+        LOGNAME,
+        "Filter coefficient < 1. makes single order Butterworth an unstable Infinite Impulse Response (IIR) filter, ");
```
AndyZe: Delete the ", " at the end of this line.

I think this error message might be too detailed for most. How about:

"Filter coefficient < 1. makes the lowpass filter unstable"

tylerjw: done

tylerjw:
The last change was to make it C++11 compatible (the way I used static_assert was from C++17).

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -38,15 +38,44 @@
...
 LowPassFilter::LowPassFilter(double low_pass_filter_coeff)
+  : previous_measurements_{ 0., 0. }
+  , previous_filtered_measurement_(0.)
+  , filter_coeff_(low_pass_filter_coeff)
+  , scale_term_(1. / (1. + low_pass_filter_coeff))
+  , feedback_term_(1. - low_pass_filter_coeff)
 {
-  filter_coeff_ = low_pass_filter_coeff;
+  // guarantee this doesn't change because the logic depends on this length implicity
+  static_assert(LowPassFilter::FILTER_LENGTH == 2, "moveit_jog_arm::LowPassFilter::FILTER_LENGTH should be 2");
```
henningkayser: Isn't that a little paranoid? Also, what if someone initializes `previous_measurements_` with another constant... ;)

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -63,12 +92,12 @@ double LowPassFilter::filter(double new_measurement)
   previous_measurements_[1] = previous_measurements_[0];
   previous_measurements_[0] = new_measurement;
 
-  double new_filtered_msrmt = (1. / (1. + filter_coeff_)) * (previous_measurements_[1] + previous_measurements_[0] -
-                                                             (-filter_coeff_ + 1.) * previous_filtered_measurement_);
+  double new_filtered_measurement = scale_term_ * (previous_measurements_[1] + previous_measurements_[0] -
```
henningkayser: Why add those extra variables? `filter_coeff_` is redundant now, the warnings would also work for `low_pass_filter_coeff`. With the original implementation the terms would be optimized anyway.

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -38,15 +38,44 @@
  */
 
 #include <moveit_jog_arm/low_pass_filter.h>
+#include <cmath>
 #include <string>
+#include <ros/ros.h>
 
-static const std::string LOGNAME = "low_pass_filter";
+const std::string LOGNAME = "low_pass_filter";
```
henningkayser:
```suggestion
constexpr char LOGNAME[] = "low_pass_filter";
```

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -38,15 +38,44 @@
...
+  , previous_filtered_measurement_(0.)
+  , filter_coeff_(low_pass_filter_coeff)
+  , scale_term_(1. / (1. + low_pass_filter_coeff))
+  , feedback_term_(1. - low_pass_filter_coeff)
 {
-  filter_coeff_ = low_pass_filter_coeff;
+  // guarantee this doesn't change because the logic depends on this length implicity
+  static_assert(LowPassFilter::FILTER_LENGTH == 2, "moveit_jog_arm::LowPassFilter::FILTER_LENGTH should be 2");
+
+  if (std::abs(feedback_term_) < EPSILON)
```
henningkayser: you could use ROS_ASSERT_MSG for this

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
tylerjw: I don't think it is too paranoid.  The indexes of the array are explicitly used instead of loops below this and this check is just a compiler check.  It won't add any code to the binary.

I don't know why changing the values of `previous_measurements_` with the `reset` method would affect this?

henningkayser: Sure, initialization using `reset` is not an issue. This check here only warns other developers from changing the value of `FILTER_LENGTH` in the header file. But you could just declare `previous_measurement_` with another constant and this assertion will never be triggered, so it's not really a guarantee that logic doesn't break. I'm not concerned about performance, I just think that a simple comment in the header file is more effective than letting developers run into warnings at compile time.

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
henningkayser: @tylerjw no anwer?

PR #1873, opened 2020-02-04
Add error message on failure to load occupancy map monitor

Dale-Koenig:
### Description

Output an error message if the Occupancy map monitor fails to find the `sensors` parameter namespace with a list of sensors.  The current behavior is for the occupancy map monitor to silently not load.

### DIFF
--- a/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
+++ b/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
@@ -191,8 +191,8 @@ void OccupancyMapMonitor::addUpdater(const OccupancyMapUpdaterPtr& updater)
     if (map_updaters_.size() > 1)
     {
       mesh_handles_.resize(map_updaters_.size());
-      if (map_updaters_.size() ==
-          2)  // when we had one updater only, we passed direcly the transform cache callback to that updater
+      // when we had one updater only, we passed direcly the transform cache callback to that updater
+      if (map_updaters_.size() == 2)
       {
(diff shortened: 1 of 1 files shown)


Dale-Koenig:
> lgtm, except it would be nice if that error was named so it was easier to find what file the error came from. Other errors in this source file aren't named so feel free to ignore that.

I'd be happy to namespace all the messages but it was unclear to me whether there was a reason for not namespacing them.  It seems a lot of the different move group capabilities are only namespaced up to something fairly generic like "manipulation", "ros_moveit_planning", or in this case "ros_moveit_perception"

Dale-Koenig:
@rhaschke Let me know if you would want all the logging namespaced

rhaschke:
Yes, please introduce a logger namespace, following our new policy.
It's common practice to augment missing logger namespaces as soon as somebody touches a source file, where it is still missing. Thanks in advance.

rhaschke:
Great that you took the effort to provide logging scopes for _all_ these files. As this goes far beyond the original scope of this PR, I suggest to file a separate PR for your last commit and we merge the first part as is. Please use `const static` variables for those logging names.

Dale-Koenig:
> Great that you took the effort to provide logging scopes for _all_ these files. As this goes far beyond the original scope of this PR, I suggest to file a separate PR for your last commit and we merge the first part as is. Please use `const static` variables for those logging names.

Figured I already had the search replace setup in my IDE :)

I reverted and force pushed.  I can rebase + make a logging PR then after this is merged.

PR #1883, opened 2020-02-07
Initialize zero dynamics in CurrentStateMonitor

henningkayser:
Currently, when setting `copy_dynamics_` to `true`, the current state would not have any velocity or effort values (i.e. `RobotSate::hasVelocities()` is *false*) if the joint state message only contains zero values for dynamics. IMO the values should be written to the robot state initially so that a caller can check if dynamics have been updated or not even though the values are 0. With this PR, the robot state dynamics will always be updated if `hasVelocities()`/`hasEffort()` are false.


henningkayser:
> As far as I understand, without this PR `hasVelocities` and `hasEffort` never will get true.

Exactly, but only if velocity/effort are 0

> If so, I approve this PR.

Thanks!

> I found another issue in `RobotState::copyFrom()`.

Issue #1888, opened 2020-02-10
default velocity / acceleration factors


v4hn:
I thought I already created an issue for this at some point, but apparently not. So here it goes...

A usual problem with robot configurations is that maintainers decide to specify severely reduced velocity/acceleration limits.
Of course, this ensures that the robot will never run into anything at high speeds when given a very simple motion command.
This is clearly appreciated for new users and makes sense as default behavior!

But it also means that there is no easy way to increase the speed later on without modifying the per-joint configuration (which should only be done with knowledge of the specs, of course).
In the past, I was asked quite a number of times, why MoveIt's trajectories are so slow, when "industry" demos with the same robot arms can move so much faster..
So this is actually bad advertisement for some people who might try it.

As we already have the concept of velocity/acceleration scaling factors, I would propose to add two simple parameters `default_velocity_scaling_factor` and `default_acceleration_scaling_factor` (to our joint_limits.yaml?) and support them throughout the pipeline.
This should allow the average user to speed up their trajectories simply by setting higher factors.

Any thoughts on this @rhaschke @gavanderhoorn ?

gavanderhoorn:
> As we already have the concept of velocity/acceleration scaling factors, I would propose to add two simple parameters `default_velocity_scaling_factor` and `default_acceleration_scaling_factor` (to our joint_limits.yaml?) and support them throughout the pipeline.

I'm not sure I completely understand what the idea is with these new parameters.

Could you clarify?

jschleicher:
Do you think of using demo.launch to reduce robot speed when run from RViz? We solved that by setting default values in the rviz config of [prbt_moveit_config](https://github.com/PilzDE/pilz_robots/tree/melodic-devel/prbt_moveit_config) to 20%. This value reduces the default for first-time users. move_group-requests (programmatically) can still be executed full speed and the user can increase the speed in the plugin as well. But the default(s) shouldn't be 1.0 of course.

![PRBT in RViz](http://wiki.ros.org/pilz_robots/Tutorials/MoveRobotWithPilzCommand_planner?action=AttachFile&do=get&target=execute_edited.png)

Maybe it's also possible to reduce the default velocity/acceleration scaling factors inside MotionPlanRequest message so that one has to override them to run full speed?

I fully support your opinion, that the joint_limits.yaml / URDF limits shouldn't be messed with to reduce to safe speed at all. These values should always match the hardware limits.

rhaschke:
[...]

PR #1889, opened 2020-02-10
WIP: Reduce default velocity and acceleration scaling to 0.2

rhaschke:
First part of implementing #1888. According to https://github.com/ros-planning/moveit/issues/1888#issuecomment-584166598 there are two major locations to modify the default scaling factors:
- the MoveGroupInterface and
- the rviz GUI

This PR just hard-codes these factors to a value of 0.2. According to @v4hn original suggestion, the hard-coded values should be replaced by ROS parameters. It remains to decide their name and a location to set them in some config files. What about augmenting [`joint_limits.yaml`](https://github.com/ros-planning/panda_moveit_config/blob/melodic-devel/config/joint_limits.yaml) with parameters `max_velocity_scaling_factor` and `max_acceleration_scaling_factor` at the root level?

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
@@ -124,8 +124,8 @@ public:
     goal_orientation_tolerance_ = 1e-3;  // ~0.1 deg
     allowed_planning_time_ = 5.0;
     num_planning_attempts_ = 1;
-    max_velocity_scaling_factor_ = 1.0;
-    max_acceleration_scaling_factor_ = 1.0;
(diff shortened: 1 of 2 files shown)


jonbinney:
After this change will all robots run at 20% of their previous speed?

felixvd:
Until you set the scaling factors in your code, yes. I think this is prudent and a good change for now.

rhaschke:
> I think the maximum scaling factors should always be 1.0 in my opinion.

I fully agree. Note that this PR is not ready for merging yet. It requires to replace the hard-coded values with ROS parameters. Any help is welcome. I don't have the time to work on this right now.

rhaschke:
Superseeded by #1890.

PR #1890, opened 2020-02-12
Add default velocity/acceleration scaling factors

felixvd:
### Description

Since I was in the middle of touching adjacent code and @rhaschke mentioned (#1889) that he has no time to fully implement the feature, I quickly wrote this PR and it seems to work fine. As suggested in https://github.com/ros-planning/moveit/issues/1888 (@v4hn), this PR adds default velocity/acceleration scaling factors as ROS parameters to `joint_limits.yaml` so that the initial setting is `0.1` for each, which should ensure safer robot motions for beginners. The parameter is retrieved from the server when starting up a MoveGroup and when adding a MotionPlanning widget in Rviz.

It leaves the setting in the `.ui` file of the MotionPlanning widget somewhat redundant, but I guess it's not a big problem.

I would suggest adding a note to `MIGRATION.md`, throughout the tutorials, and maybe even somewhere else so that new users are aware that the robot is not moving at full speed.

### Checklist


moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -73,6 +73,15 @@ MotionPlanningFrame::MotionPlanningFrame(MotionPlanningDisplay* pdisplay, rviz::
   connect(planning_display_, SIGNAL(queryStartStateChanged()), joints_tab_, SLOT(queryStartStateChanged()));
   connect(planning_display_, SIGNAL(queryGoalStateChanged()), joints_tab_, SLOT(queryGoalStateChanged()));
 
+  // Read the initial max velocity and acceleration scaling factors from ROS parameters
+  double default_velocity_scaling_factor_, default_acceleration_scaling_factor_;
```
rhaschke: local variables shouldn't use a trailing underscore.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -641,7 +641,7 @@
                <double>0.010000000000000</double>
               </property>
               <property name="value">
-               <double>1.000000000000000</double>
+               <double>0.100000000000000</double>
```
rhaschke: I suggest to revert these changes and leave the ui untouched.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: True. Copy-paste collateral damage. Fixed.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
felixvd: Fine with me, but I also think it can't hurt to apply it in both places.

felixvd:
> But, at least in the MoveGroupInterface you should add a corresponding clipping (and a warning).
> My suggestion is:
> 
> * `value < 0` → use default
> * `value > 1` → clip to 1.0

I didn't know that it wasn't already handled that way. I added the check as suggested.

rhaschke:
> I didn't know that it wasn't already handled that way. I added the check as suggested.

Oh, I didn't check explicitly. Indeed, all the time parameterization methods to exactly the same checking (sorry for the noise!):
https://github.com/ros-planning/moveit/blob/6b56cd68d76ae29d1ceb255d06be35543dd3f830/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp#L890-L905

... **falling back to a default of 1.0**.

I think this code should be refactored to have a common planning request adapt for time-parameterization, which in turn uses a specific time-parameterization plugin via a common interface.
The common planning request adapter would then be an ideal candidate to inject the default values, if they are still zero.
To keep things simple (and avoid code duplication), we could even hard-code the factors in rviz to 0.0.

felixvd:
Hmm, I would prefer to keep this PR small since the default parameters will be useful regardless, and it seems like a separate issue. Would you prefer to revert the clipping inside the MoveGroupInterface or keep it? I can see arguments for either.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -1052,14 +1052,26 @@ bool MoveItConfigData::output3DSensorPluginYAML(const std::string& file_path)
 bool MoveItConfigData::outputJointLimitsYAML(const std::string& file_path)
 {
   YAML::Emitter emitter;
+  emitter << YAML::Comment("joint_limits.yaml allows the dynamics properties specified in the URDF "
+                           "to be overwritten or augmented as needed");
+  emitter << YAML::Newline;
+
   emitter << YAML::BeginMap;
 
+  emitter << YAML::Comment("For beginners, down-scale velocity and acceleration limits.") << YAML::Newline;
```
felixvd:
```suggestion
  emitter << YAML::Comment("For beginners, we downscale velocity and acceleration limits.") << YAML::Newline;
```

felixvd:
Looks great, thanks! Agree with everything and only found that typo.

rhaschke:
@v4hn, is this what you intended in #1888?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -1052,8 +1052,26 @@ bool MoveItConfigData::output3DSensorPluginYAML(const std::string& file_path)
...
 {
   YAML::Emitter emitter;
+  emitter << YAML::Comment("joint_limits.yaml allows the dynamics properties specified in the URDF "
+                           "to be overwritten or augmented as needed");
+  emitter << YAML::Newline;
+
   emitter << YAML::BeginMap;
 
+  emitter << YAML::Comment("For beginners, we downscale velocity and acceleration limits.") << YAML::Newline;
+  emitter << YAML::Comment("Increase those values to 1.0 to exploit the full potential of the robot.");
```
v4hn: How about

> For beginners, we downscale velocity and acceleration by default.
> You can always specify higher scaling factors (<= 1.0) in your motion requests.
> Increase those values to 1.0 to always move at maximum speed.

?

The second sentence is rather misleading right now.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -335,14 +337,32 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
...
+  {
+    if (target > 1.0)
+    {
+      ROS_WARN_NAMED("move_group_interface", "Limiting max_%s (%.2f) to 1.0.", name, target);
+      variable = 1.0;
+    }
+    else if (target <= 0.0)
+    {
+      node_handle_.param<double>(std::string("robot_description_planning/default_") + name, variable, fallback);
+      ROS_WARN_NAMED("move_group_interface", "max_%s <= 0.0! Setting to default: %.2f.", name, variable);
```
v4hn: I would expect this to be reasonable behavior for this API, so I would prefer this as a `ROS_DEBUG` instead.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -335,14 +337,32 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
...
   }
 
-  void setMaxAccelerationScalingFactor(double max_acceleration_scaling_factor)
+  void setMaxAccelerationScalingFactor(double factor)
   {
-    max_acceleration_scaling_factor_ = max_acceleration_scaling_factor;
+    setMaxScalingFactor(max_acceleration_scaling_factor_, factor, "acceleration_scaling_factor", 0.1);
+  }
+
+  void setMaxScalingFactor(double& variable, const double target, const char* name, double fallback)
```
v4hn: This method looks quite ugly to me, especially `name` which has to be a suffix to an incomplete ROS parameter name later on. (you could even drop the `_scaling_factor` the way you're doing it now).

For me, this is clearly not worth avoiding 3 lines of code duplication.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -124,8 +124,10 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
     goal_orientation_tolerance_ = 1e-3;  // ~0.1 deg
     allowed_planning_time_ = 5.0;
     num_planning_attempts_ = 1;
-    max_velocity_scaling_factor_ = 1.0;
-    max_acceleration_scaling_factor_ = 1.0;
+    node_handle_.param<double>("robot_description_planning/joint_limits/default_velocity_scaling_factor",
+                               max_velocity_scaling_factor_, 0.1);
+    node_handle_.param<double>("robot_description_planning/joint_limits/default_acceleration_scaling_factor",
+                               max_acceleration_scaling_factor_, 0.1);
```
v4hn: There is quite a number of places with hard-coded `0.1` now, but I'm not sure we have a nice place to define a constant for this.

`0.1` seems overly slow to me though, I would have picked `0.5` myself.
Maybe we want to haggle over the exact number in the maintainer meeting next week?

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
rhaschke: I don't agree. If the user specifies a negative scaling factor, she should get a warning about that.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
rhaschke: I don't get what is ugly about this method. This is 16 lines of code, which would be duplicated.
Code duplication is always error-prone to be not updated in sync in the future.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
felixvd: I think we mostly need to decide which situation the default value should be for. In my opinion, it should be a safe ballpen mode for new users, which leaves them enough time to hit the emergency switch (which they will invariably not have in their hand) in case they accidentally send a motion request that does not do what they expected. E.g. the IK flipped and the robot is doing a large motion, or the robot hits something that wasn't represented in the environment (like the robot base), etc.

Choosing a "very safe" default value and expecting the user to adjust the speed to when they are ready is probably better than the opposite.

I tested a few parameter values with our UR5 to check which speeds would still be surprising a user in this type of situation. `0.5` is barely different from full speed for small motions, `0.2` seemed to be still pretty fast, and `0.1` seemed safe. `0.1` for a UR5 is about `10 cm/s` in most configurations, which I think this is a good compromise.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
felixvd: I agree that clipping the value to [0, 1] is reasonable behavior, but I think users should be alerted if they attempt to set it outside of that range.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
felixvd: Fixed

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
felixvd: Unrelated: shouldn't this be `ROS_WARN_NAMED_STREAM`?

rhaschke: Using stream syntax would be more verbose. But, I don't mind.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
felixvd: I think I'm with @rhaschke on this one, but I made the variable names longer to spite him 😎 (and because I think it's clearer).

rhaschke: I'm fine with this :smile:

felixvd:
I updated the PR with small changes. We can talk about the actual value in the meeting if you like.

A last comment is that I think `max_X_scaling_factor` vs `X_scaling_factor` should probably be unified. The former is used in the existing code, the latter is what I wrote reflexively for the default parameters. I think I prefer the latter, but I doubt it's worth changing the nomenclature everywhere else.

rhaschke:
> A last comment is that I think `max_X_scaling_factor` vs `X_scaling_factor` should probably be unified. The former is used in the existing code, the latter is what I wrote reflexively for the default parameters.

Actually you used `default_X_scaling_factor` for the name in the yaml config file.
For me that is fine.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
v4hn: > I don't agree. If the user specifies a negative scaling factor, she should get a warning about that.

If you insist, I can agree with the negative factor.
In this case, I would prefer to have a separate branch for `0.0` though.
Otherwise there is no warning-free API to reset the factors to their defaults (aside from the Constructor)!

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
v4hn: Ah, I didn't see this was your code @rhaschke . Of course you would argue for more code obscurity over easier-to-understand code :laughing: .
Anyway, it's just a nitpick.

felixvd:
Yes, I used `default_X_scaling_factor` instead of `default_max_X_scaling_factor`. I don't think it's a big deal either, but I figured it's worth noting.

felixvd:
If this is good with you after your last change @v4hn, the only tests that are failing are the code coverage, so I think this is ready to merge.

rhaschke:
We wanted to discuss the actual numbers for the scaling factors... So not ready to merge yet.

v4hn:
@felixvd: The maintainer meeting just agreed to have the default of 0.1.

At the same time, we need prominent information (or better a tutorial page) on how to set the default scaling factors also for existing moveit configurations.
Could you do that too?

felixvd:
I'll do it next week, after the IROS deadline :)

I will add a ToDo list to the first post of this thread.

rhaschke:
@v4hn Is this ready for merging?

felixvd:
I'll update it soon. I was still busy with the video submission deadline that is in 15 minutes.

v4hn:
> @v4hn Is this ready for merging?

We do need some prominent advertisement,
either a discourse announcement with some explanations or a tutorial.

Also, there should be an entry in the `MIGRATION.md` as part of this patch set.

So I want to wait for @felixvd to do this before merging.
Now that the IROS video-submission deadline is over too, he should have time to do that. :grin:

felixvd:
I added a mention in `MIGRATION.md` and the documentation of all interface functions that change the scaling factors, and filed the corresponding PR for the tutorials.

I think the only other thing to do would be to post a pre-emptive ROS answers question "Why is MoveIt so slow?"

v4hn:
> I added a mention in MIGRATION.md

You didn't push a change to the file as part of this request.

> I think the only other thing to do would be to post a pre-emptive ROS answers question "Why is MoveIt so slow?"

Could you do that please? Aside from these things, this is good to go.

felixvd:
I forgot to push the branch from the office 🙄I just fixed that and rebased.

The ROS answers question is [here](https://answers.ros.org/question/346203/why-is-my-robot-moving-so-slow-with-moveit/?answer=346204#post-id-346204).

PR #1892, opened 2020-02-12
Add initial poses to fake_controllers.yaml

felixvd:
### Description

It looks ugly when robots start up in nonsensical poses when running `demo.launch`, and I don't think it should be the default behavior. Setting the robots' initial pose is perfectly possible, but it requires [a manual addition to `kinematics.yaml`](http://docs.ros.org/api/moveit_tutorials/html/doc/fake_controller_manager/fake_controller_manager_tutorial.html#yaml-file-examples) that users have to look up. This PR changes the behavior of the setup assistant so that if group states (robot poses) are defined in the setup assistant/SRDF, the first one is automatically exported to `kinematics.yaml` as the initial pose of that group.

I believe this will increase discoverability of the feature and be a better default setting.

When no pose is set, the default name "home" is exported. I confirmed that setting the parameter this way does not cause problems and tested it with multiple groups.

### Checklist

### DIFF
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -535,18 +535,19 @@ bool MoveItConfigData::outputFakeControllersYAML(const std::string& file_path)
   // Add an initial pose for each group
   emitter << YAML::Key << "initial" << YAML::Comment("Define initial robot poses.");
 
-  bool pose_found = false;
-  for (srdf::Model::Group& group : srdf_->groups_)
+  bool poses_found = false;
+  std::string default_group_name;
+  for (const srdf::Model::Group& group : srdf_->groups_)
   {
-    const robot_model::JointModelGroup* joint_model_group = getRobotModel()->getJointModelGroup(group.name_);
-
-    for (srdf::Model::GroupState& group_state : srdf_->group_states_)
+    if (default_group_name.empty())
+      default_group_name = group.name_;
+    for (const srdf::Model::GroupState& group_state : srdf_->group_states_)
     {
       if (group.name_ == group_state.group_)
       {
-        if (!pose_found)
+        if (!poses_found)
         {
-          pose_found = true;
+          poses_found = true;
           emitter << YAML::Value << YAML::BeginSeq;
         }
         emitter << YAML::BeginMap;
@@ -559,11 +560,17 @@ bool MoveItConfigData::outputFakeControllersYAML(const std::string& file_path)
       }
     }
   }
-  if (pose_found)
+  if (poses_found)
     emitter << YAML::EndSeq;
   else
+  {
     // Add commented lines to show how the feature can be used
-    emitter << YAML::Comment(" - group: panda_arm \n   pose: ready");
+    if (default_group_name.empty())
+      default_group_name = "group";
+    emitter << YAML::Newline;
+    emitter << YAML::Comment(" - group: " + default_group_name) << YAML::Newline;
+    emitter << YAML::Comment("   pose: home") << YAML::Newline;
+  }
 
   emitter << YAML::EndMap;
 
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -529,6 +529,34 @@ bool MoveItConfigData::outputFakeControllersYAML(const std::string& file_path)
     emitter << YAML::EndSeq;
     emitter << YAML::EndMap;
   }
+
+  emitter << YAML::EndSeq;
+
+  // Add an initial pose for each group
+  emitter << YAML::Key << "initial" << YAML::Comment("Default robot poses can be defined here");
+  emitter << YAML::Value << YAML::BeginSeq;
+
+  bool found = false;
+  for (srdf::Model::Group& group : srdf_->groups_)
+  {
+    const robot_model::JointModelGroup* joint_model_group = getRobotModel()->getJointModelGroup(group.name_);
+
+    for (srdf::Model::GroupState& group_state : srdf_->group_states_)
+    {
+      if (group.name_ == group_state.group_)
+      {
+        emitter << YAML::BeginMap;
+        emitter << YAML::Key << "group";
+        emitter << YAML::Value << group.name_;
+        emitter << YAML::Key << "pose";
+        emitter << YAML::Value << group_state.name_;
+        emitter << YAML::EndMap;
+        found = true;
+        break;
+      }
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -529,6 +529,29 @@ bool MoveItConfigData::outputFakeControllersYAML(const std::string& file_path)
...
+    emitter << YAML::BeginMap;
+    const std::vector<const robot_model::JointModel*>& joint_models = joint_model_group->getActiveJointModels();
+    emitter << YAML::Key << "group";
+    emitter << YAML::Value << group.name_;
+
+    emitter << YAML::Key << "pose";
+    if (!srdf_->group_states_.empty())
+      emitter << YAML::Value << srdf_->group_states_[0].name_;
+    else
+      emitter << YAML::Value << "home";  // Default entry
```
rhaschke: If no group state is available, nothing should be written here. `home` is probably not defined...

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -529,6 +529,29 @@ bool MoveItConfigData::outputFakeControllersYAML(const std::string& file_path)
...
+
+  // Add an initial pose for each group
+  emitter << YAML::Key << "initial";
+  emitter << YAML::Value << YAML::BeginSeq;
+
+  for (srdf::Model::Group& group : srdf_->groups_)
+  {
+    const robot_model::JointModelGroup* joint_model_group = getRobotModel()->getJointModelGroup(group.name_);
+    emitter << YAML::BeginMap;
+    const std::vector<const robot_model::JointModel*>& joint_models = joint_model_group->getActiveJointModels();
```
rhaschke: That's not needed, is it?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
felixvd: It isn't, but it also doesn't cause any errors. I left it in as an example, for users to find if they read through the files, so that the option is visible even if they have not defined a group state yet. Feature discovery and that.

I'm not opposed if the consensus is to remove it, but I figured it doesn't hurt.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
felixvd: Whoops. I thought I deleted it. Fixed.

felixvd:
> Of course, it would be great to have a drop-down or checkbox to actively select a specific initial pose from the GUI 😉
Completely agreed :) A third column called "Set as default" in the robot poses widget would be great. I've never touched the widgets and would take too long to do it, though. Maybe someone who has been more involved with the setup assistant would be more efficient, like @MohmadAyman?

I could add something like "The first pose for each robot will be set as its default." to the widget text in the meantime, if you think it helps.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
rhaschke: According to the following code, an unknown pose should trigger a warning:
https://github.com/ros-planning/moveit/blob/6b56cd68d76ae29d1ceb255d06be35543dd3f830/moveit_plugins/moveit_fake_controller_manager/src/moveit_fake_controller_manager.cpp#L164-L169

If it doesn't, there is a bug somewhere else.
For feature discovery, I suggest commenting the corresponding pose line in yaml. Not sure though, how to do this programmatically.

felixvd: True, it might throw a warning, and it's not nice to be flooded with those by default :)

I will change it and add a comment instead. Apparently [YAML::Comment works](https://github.com/jbeder/yaml-cpp/wiki/How-To-Emit-YAML).

felixvd:
I added the changes and fixed a bug that I overlooked because I had defined both my test groups' poses with the same name. This should work.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -529,6 +529,34 @@ bool MoveItConfigData::outputFakeControllersYAML(const std::string& file_path)
...
     emitter << YAML::EndMap;
   }
+
+  emitter << YAML::EndSeq;
+
+  // Add an initial pose for each group
+  emitter << YAML::Key << "initial" << YAML::Comment("Default robot poses can be defined here");
+  emitter << YAML::Value << YAML::BeginSeq;
+
+  bool found = false;
```
rhaschke: `found` isn't yet used. However, you could use it for a commented-out template as above.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -529,6 +529,34 @@ bool MoveItConfigData::outputFakeControllersYAML(const std::string& file_path)
     emitter << YAML::EndSeq;
     emitter << YAML::EndMap;
   }
+
+  emitter << YAML::EndSeq;
+
+  // Add an initial pose for each group
+  emitter << YAML::Key << "initial" << YAML::Comment("Default robot poses can be defined here");
```
rhaschke: I was actually thinking of something like this, if there was no pose defined at all
(otherwise you loose feature recovery):
```yaml
initial:
#  - group: panda_arm
#    pose: home
```
Also, I suggest the following wording:
```suggestion
  emitter << YAML::Key << "initial" << YAML::Comment("Define initial robot poses");
```

MohmadAyman:
> > Of course, it would be great to have a drop-down or checkbox to actively select a specific initial pose from the GUI wink
> 
> Completely agreed :) A third column called "Set as default" in the robot poses widget would be great. I've never touched the widgets and would take too long to do it, though. Maybe someone who has been more involved with the setup assistant would be more efficient, like @MohmadAyman?
> 
> I could add something like "The first pose for each robot will be set as its default." to the widget text in the meantime, if you think it helps.

I like the sound of "Set as default" :wink: , rn I've exams, I should be able help in the first week of March :sweat_smile:

rhaschke:
Good luck for your exams, @MohmadAyman. Would be great if you could help when you're available again.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
felixvd: Done

moveit_setup_assistant/src/tools/moveit_config_data.cpp
felixvd: Sounds good. I committed it like that.

felixvd:
I added the suggestions and a line in the setup assistant widget.

Issue #210, opened 2016-09-09
Specify the robots' movement (LIN / PTP)


mat-l:
Hello,

how can I plan and execute linear (LIN) or point to point (PTP) movements using MoveIt!   
I would like to define the distance between two waypoints using LIN and PTP movements in the command line, C++/Python programms and I would like to integrate it later maybe into the Cartesian Path Planner (https://github.com/ros-industrial-consortium/fermi)

Any ideas?
Best regards

v4hn:
Could you elaborate on this please?

mat-l:
Hi,
the point is, that i would like to tell ROS how the path should be planned for the robot. Common kinds of pathes are synchronous or asynchronous point to point movements or even linear ones.

If you have different waypoints named for example w01, w02, w03, and so one, then you can tell the robot how to move between the waypoints.

For example:
move_lin(w01, w02) //linear movement between the two waypoints   

move_ptp(w02, w03) //point to point movement between the two waypoints

My question is, how to tell ROS how to plan the path between different kinds of waypoints?

Besr regards

CMobley7:
mat-l,

For linear and point to point motion use the compute_cartesian_path() and plan() functions respectively.

V/R,
CMobley7

PR #1893, opened 2020-02-12
Pilz planner merge

ct2034:
### Description

Merging pilz trajectory generation into moveit as suggested here: https://github.com/ros-planning/moveit/issues/1800

## merge before this:
- [x] https://github.com/ros-planning/moveit_msgs/pull/65
- [x] https://github.com/ros-planning/moveit_resources/pull/44 -> then
- [x] https://github.com/ros-planning/moveit_resources/pull/43

## merge after this:
- [ ] https://github.com/ros-planning/moveit_tutorials/pull/461
- [ ] https://github.com/PilzDE/pilz_industrial_motion/pull/244 
- [ ] https://github.com/PilzDE/pilz_robots/pull/316
- [ ] https://github.com/PilzDE/pilz_application_templates/pull/15

### Checklist
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/) -> [see PR](https://github.com/ros-planning/moveit_tutorials/pull/461)


moveit_planners/trapezoidal/CMakeLists.txt
```diff
@@ -0,0 +1,547 @@
...
+
+  # Tests with other robots only for kinetic (for now)
+
+  #add_rostest(test/integrationtest_sequence_action_capability_abb_irb2400.test
+  #  DEPENDENCIES integrationtest_sequence_action_capability
+  #)
+
+  #add_rostest(test/integrationtest_sequence_action_capability_frankaemika_panda.test
+  #  DEPENDENCIES integrationtest_sequence_action_capability
+  #)
```
davetcoleman: Why are Franka tests disabled? we support those in moveit_resources for tests:
https://github.com/ros-planning/moveit_resources/tree/master/panda_moveit_config

moveit_planners/trapezoidal/README.md
```diff
@@ -0,0 +1,236 @@
+# Overview
```
davetcoleman: Can we convert some of this README into a moveit_tutorial?

moveit_planners/trapezoidal/doc/MotionBlendAlgorithmDescription.tex
```diff
@@ -0,0 +1,116 @@
...
+	\end{subfigure}
+	
+\end{figure}
+
+
+\begin{thebibliography}{9}
+\bibitem {lloyd1993trajectory}Lloyd, John and Hayward, Vincent \textit{Trajectory generation for sensor-driven and time-varying tasks}, The International journal of robotics research, 1993
+\bibitem {dantam2014orientation}Dantam, Neil and Stilman, Mike \textit{Spherical Parabolic Blends for Robot Workspace Trajectories}, International Conference on Intelligent Robots and Systems (IROS), 2014
+\bibitem{trapezoidaltrajectorygeneration}\url{https://github.com/TrapezoidalDE/trapezoidal_industrial_motion/tree/melodic-devel/trapezoidal_trajectory_generation}
+\end{thebibliography}
```
davetcoleman: Awesome documentation!

moveit_planners/trapezoidal/include/trapezoidal_trajectory_generation/cartesian_limit.h
```diff
@@ -0,0 +1,139 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef CARTESIAN_LIMIT_H
+#define CARTESIAN_LIMIT_H
```
davetcoleman: we've switched to #pragma once, so that's more preferable:
https://moveit.ros.org/documentation/contributing/code/

moveit_planners/trapezoidal/include/trapezoidal_trajectory_generation/cartesian_limit.h
```diff
@@ -0,0 +1,139 @@
...
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
davetcoleman: Seems like this license has not been updated

moveit_planners/trapezoidal/package.xml
```diff
@@ -0,0 +1,56 @@
...
+  <name>trapezoidal_trajectory_generation</name>
+  <version>0.4.10</version>
+  <description>The trapezoidal_trajectory_generation package containing the MoveIt! plugin trapezoidal_command_planner.</description>
+
+  <maintainer email="a.gutenkunst@pilz.de">Alexander Gutenkunst</maintainer>
+  <maintainer email="c.henkel@pilz.de">Christian Henkel</maintainer>
+  <maintainer email="h.slusarek@pilz.de">Hagen Slusarek</maintainer>
+  <maintainer email="i.martini@pilz.de">Immanuel Martini</maintainer>
+
+  <license>LGPLv3</license>
```
davetcoleman: update here and other package.xmls

moveit_planners/trapezoidal/README.md
ct2034: Yes, this is planned. @jschleicher I think the current documentation matches the style of tutorial sections for other planners well. Although we can probably skip the part that introduced moveit. ;-)

moveit_planners/trapezoidal/src/trajectory_functions.cpp
```diff
@@ -0,0 +1,566 @@
...
+}
+
+bool trapezoidal::computePoseIK(const moveit::core::RobotModelConstPtr& robot_model, const std::string& group_name,
+                                const std::string& link_name, const geometry_msgs::Pose& pose,
+                                const std::string& frame_id, const std::map<std::string, double>& seed,
+                                std::map<std::string, double>& solution, bool check_self_collision,
+                                const double timeout)
+{
+  Eigen::Isometry3d pose_eigen;
+  tf::poseMsgToEigen(pose, pose_eigen);
```
jschleicher: We should update tf to tf2.
See PilzDE/pilz_industrial_motion#163

moveit_planners/trapezoidal/README.md
jschleicher: See ros-planning/moveit_tutorials#461.

davetcoleman:
Should the "WIP" be removed at this point?
Why is Travis still failing?

jschleicher:
ros-planning/moveit_msgs#65 is required first. So travis currently is failing because of missing message(s).
PilzDE/pilz_robots#316 needs to be (at least partially) merged for a successful moveit build as well, see [travis test build](https://travis-ci.org/pilzde/moveit/builds/655278908).

moveit_planners/trapezoidal/CMakeLists.txt
jschleicher: The issue https://github.com/PilzDE/pilz_industrial_motion/issues/176 doesn't tell the problem or link a failed build - we'd have to try enabling them again...

ct2034: I enabled them again. Closes https://github.com/PilzDE/pilz_industrial_motion/issues/176

felixvd:
I'm unclear on the name change?

jschleicher:
Ready for review now @mamoll @v4hn.

Still ros-planning/moveit_msgs#65 needs to be merged first to make travis pass.

moveit_planners/pilz_industrial_motion_planner/package.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>pilz_industrial_motion_planner</name>
+  <version>0.4.10</version>
+  <description>MoveIt! plugin to generate industrial trajectories PTP, LIN, CIRC and sequences thereof.</description>
```
davetcoleman, line 5: unimportant nit: we removed the ! in MoveIt

moveit_planners/pilz_industrial_motion_planner/package.xml
```diff
@@ -0,0 +1,56 @@
...
+
+  <depend>eigen_conversions</depend>
+  <depend>kdl_conversions</depend>
+  <depend>moveit_ros_planning_interface</depend>
+  <depend>moveit_msgs</depend>
+  <depend>moveit_core</depend>
+  <depend>moveit_ros_planning</depend> <!-- RobotModelLoader -->
+  <depend>moveit_ros_move_group</depend> <!-- move_group capability -->
+  <depend>orocos_kdl</depend>
+  <depend>pilz_extensions</depend>
```
davetcoleman, line 29: Is this package going to be released as a debian regularly? having it as a dependency will require that

moveit_planners/trajopt/package.xml
```diff
@@ -6,9 +6,7 @@
   <author email="omid.github@gmail.com">Omid Heidari</author>
 
   <maintainer email="omid.github@gmail.com">Omid Heidari</maintainer>
-  <maintainer email="moveit_releasers@googlegroups.com">MoveIt! Release Team</maintainer>
-
-  <license>BSD</license>
```
davetcoleman, line 11: This shouldn't be changed, please revert

moveit_planners/pilz_industrial_motion_planner/doc/MotionBlendAlgorithmDescription.fls
```diff
@@ -0,0 +1,147 @@
+PWD /home/ch/ros/pilz_ws/src/moveit/moveit_planners/pilz_industrial_motion_planner/doc
```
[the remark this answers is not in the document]
rhaschke: I agree with Dave that the source repo should only comprise source files. Please remove all tex output from the commit history and just keep .tex and image sources.
Particularly, remove:
*.fdb_latexmk
*.fls
*.out
*.synctex.gz
MotionBlendAlgorithmDescription.pdf (created from .tex)
.png and .pdf that are generated from .uxf (at least keep only one of the generated files)
ptp.png (generated from ptp.svg)
*-eps-converted-to.pdf (I suggest to keep pdfs instead of .eps files)

Also, do we really need the email confirmations from @clalancette and @gavanderhoorn for changing license terms in this repo? Isn't it sufficient that you can provide them if necessary?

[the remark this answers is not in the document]
clalancette: @rhaschke @gavanderhoorn I'm happy to review new licensing terms, but I think I'm missing some context.  Can you please point me to what the license was and what it will be in the future?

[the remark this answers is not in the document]
ct2034: @clalancette the change was from LVGLv3 to BSD 3-clause. But we talked about it already, as documented here: https://github.com/PilzDE/moveit/blob/pilz_planner_merge/moveit_planners/pilz_industrial_motion_planner/doc/E-Mail_clalancette.pdf 
(which maybe shows to @rhaschke why it may be handy to have the files around ;-) )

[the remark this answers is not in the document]
ct2034: @rhaschke I do agree with you not to check in the latex auxiliary files. But I would like to check in the final pdf, to be able to make it available directly for documentation purposes. I can not expect from someone looking for information on the planner to first build the tex files.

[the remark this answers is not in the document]
rhaschke: The preferred format for documentation in MoveIt is via sphinx in the tutorials. Is there a possibility to convert/compile latex into .rst files?

moveit_planners/pilz_industrial_motion_planner/package.xml
ct2034: yes, it is released with all our other packages

moveit_planners/trajopt/package.xml
ct2034: sorry about that

moveit_planners/pilz_industrial_motion_planner/doc/MotionBlendAlgorithmDescription.fls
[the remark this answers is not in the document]
ct2034: I have tried to convert the file with [pandoc](https://pandoc.org/). But it does not handle math and graphics very well.
Where would you put this documentation anyways? Under https://moveit.ros.org/documentation/planners/ I assume. Because all deeper links from there seem to pretty inconsistent anyways.

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
```diff
@@ -0,0 +1,530 @@
...
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
+
+## Mark cpp header files for installation
+#install(DIRECTORY include/${PROJECT_NAME}/
+#  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+#  FILES_MATCHING PATTERN "*.h"
+#  PATTERN ".svn" EXCLUDE
+#)
```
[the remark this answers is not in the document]
ct2034: I cleaned up the whole file in 1712678

moveit_planners/pilz_industrial_motion_planner/doc/MotionBlendAlgorithmDescription.fls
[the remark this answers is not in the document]
rhaschke: I was thinking of the MoveIt tutorials. But let's keep this separated:
- We should definitely have a tutorial how to use this planner - probably you already have this.
- In https://moveit.ros.org/documentation/planners we should have a short summary of functionality and there you could link your pdf.

moveit_planners/pilz_industrial_motion_planner/package.xml
```diff
@@ -0,0 +1,56 @@
...
+  <maintainer email="a.gutenkunst@pilz.de">Alexander Gutenkunst</maintainer>
+  <maintainer email="c.henkel@pilz.de">Christian Henkel</maintainer>
+  <maintainer email="i.martini@pilz.de">Immanuel Martini</maintainer>
+  <maintainer email="j.schleicher@pilz.de">Joachim Schleicher</maintainer>
+  <maintainer email="h.slusarek@pilz.de">Hagen Slusarek</maintainer>
+
+  <license>BSD</license>
+
+  <url type="website">https://wiki.ros.org/pilz_trajectory_generation</url>
+  <url type="bugtracker">https://github.com/PilzDE/pilz_industrial_motion/issues</url>
```
rhaschke, line 16: Should these websites be changed to MoveIt? Do you continue to host the source mainly in your organization or here at ros-planning? In the former case, why to merge into MoveIt at all. In this case, it would suffice to provide binary-releases as well as some tutorials/documentation on the MoveIt side.

moveit_planners/pilz_industrial_motion_planner/doc/MotionBlendAlgorithmDescription.fls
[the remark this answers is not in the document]
ct2034: Yes, this was also what I was thinking. 

@jschleicher created the tutorial PR: https://github.com/ros-planning/moveit_tutorials/pull/461

And I will then focus on the https://moveit.ros.org/documentation/planners part

moveit_planners/pilz_industrial_motion_planner_testutils/package.xml
```diff
@@ -0,0 +1,29 @@
...
+  <maintainer email="a.gutenkunst@pilz.de">Alexander Gutenkunst</maintainer>
+  <maintainer email="c.henkel@pilz.de">Christian Henkel</maintainer>
+  <maintainer email="i.martini@pilz.de">Immanuel Martini</maintainer>
+  <maintainer email="j.schleicher@pilz.de">Joachim Schleicher</maintainer>
+  <maintainer email="h.slusarek@pilz.de">Hagen Slusarek</maintainer>
+
+  <license>Apache 2.0</license>
+
+  <url type="website">https://wiki.ros.org/pilz_industrial_motion</url>
+  <url type="bugtracker">https://github.com/PilzDE/pilz_industrial_motion/issues</url>
```
rhaschke, line 16: Same here.

davetcoleman:
@rhaschke the discussion on whether to add this to the moveit repo was already decided on in past maintainer meetings and here:
https://github.com/ros-planning/moveit/issues/1800

The benefits are outlined there. Something I didn't mention there but is also relevant: I don't see why the scope of MoveIt, as well as the scope of core ROS packages, must be limited to what was created at Willow Garage. I believe adding new major features and tightly integrating them into MoveIt, as we're doing here, is beneficial for consistent maintenance, usability, and discoverability.

Overall I think this PR is huge as we're importing a whole project, and our best approach to changes is to do follow up PRs for fixups. 

> Merging this here, I see the risk that it will become harder for Pilz to merge changes, because they need to pass our review process. On the other hand, Pilz will need to continue maintaining this code base, as I think no current MoveIt maintainer will sign to be responsible for this new code base.

For this reason we have plans [here](https://github.com/ros-planning/moveit/issues/1800) for the following:

- Add @jschleicher as a MoveIt maintainer so he can continue to maintain this feature. I think he has proven his skills with the development of pilz_industrial_motion
- Add @ct2034 as a MoveIt Core Contributor so he can also get more involved in maintenance

ct2034:
Thanks, @rhaschke for raising the point of making it harder for us to get our changes into the code. We have also had this discussion internally bu agreed that under the condition of us being maintainer/core contributor it will be manageable. While we expect some support in terms of maintenance we are of cause willing to do our fair share in terms of maintenance, just the same way we maintain all our other code.

I also want to apologize for the sometimes sloppy job with the inclusion of wrong files but it is also a big PR for us.

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
```diff
@@ -0,0 +1,494 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(pilz_industrial_motion_planner)
+
+## Add support for C++11, supported in ROS Kinetic and newer
+add_definitions(-std=c++11)
+add_definitions(-Wall)
+add_definitions(-Wextra)
+add_definitions(-Wno-unused-parameter)
+add_definitions(-Werror)
```
traversaro, line 9: To maintain Windows compatibility, it probably make sense to only  add this GCC/Clang-specific flags under an if, see for example https://github.com/ros-planning/moveit/blob/379acf91aa3630f320db77ba85b1ca294310b0bc/moveit_core/CMakeLists.txt#L8 . 

cc @seanyen

ct2034: could we do this in a separate PR afterwards?

rhaschke:
I'm fine to merge this PR in general, just wanted to raise some concerns, because I stumbled over the issue and repo links pointing to the old github repository. As far as I understand, Pilz will move maintenance to this repo only. Having @jschleicher and @ct2034 in the maintainer/contributor team would be great. @ct2034, could you please comment on the (seamingly) superfluous files mentioned in https://github.com/ros-planning/moveit/pull/1893#pullrequestreview-399252619?

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
traversaro: > 
> 
> could we do this in a separate PR afterwards?

I guess this is up to the maintainers, I just noticed this by looking up the (great) content added.

ct2034:
> * There are several image files: `moveit_planners/pilz_industrial_motion_planner/test/acceptance_tests/img/acceptance_test_lin_img*.png`
>   What is their role?

We have defined acceptance, which are supposed to be executed by hand. These images are included in the test description files, e.g. https://github.com/PilzDE/moveit/blob/pilz_planner_merge/moveit_planners/pilz_industrial_motion_planner/test/acceptance_tests/acceptance_test_lin.md


> * In `moveit_planners/pilz_industrial_motion_planner/test/test_robots/frankaemika_panda` you seem to have a copy of `panda_moveit_config` / `moveit_resources/panda_moveit_config`. Please use the latter one instead of introducing yet another copy here.

We have a couple of own robot-specific settings for testing. But I will remove everything taht is also in moveit_resources. 


> * This PR includes `moveit_planners/trajopt/package.xml`, which is unrelated.

Should be gone now.

rhaschke:
Dear @davetcoleman,
In my opinion, the points listed in https://github.com/ros-planning/moveit/issues/1800 are well taken,  but could be realized in a separate repo as well. I'm hesitant to increase the core MoveIt code base by such a large repository, even further increasing turnaround times of CI. As the package is self-contained and will not be named according to MoveIt's naming scheme (`moveit_*`), but continue to be named `pilz_industrial_motion_planner` (which is perfectly fine to honor Pilz's effort), I think it should stay in a separate repository.
A compromise: what about moving the whole repository from Pilz's organization to `ros-planning`?
This way it becomes more visible in the context of ROS / MoveIt and at the same time we could easily preserve the whole history of the source, which is a big plus for such a code base.

felixvd:
I have a feeling that I can't fully judge the implications of maintaining a monolithic repository (also discussed [here](https://github.com/ros-planning/moveit/issues/1928)). I tend to agree with @davetcoleman about putting everything in one place and calling it "clean", but it is very possible I may be missing something, or that it might cause problems.

One argument in favor of including this in the main repository is that PTP/LIN motions are considered by most users to be a basic, core feature, which any motion planning software should just "come with". They have been requested in [issues as early as 2016](https://github.com/ros-planning/moveit/issues/210) (and possibly earlier) and like I mentioned [here](https://github.com/ros-planning/moveit/issues/797), they are an industry standard. It would make sense to include it in the core, non-optional packages.

I know that things can work out of the box with separate repositories, so this is not a conclusive argument, but I agree that it would improve discoverability and make PRs and maintenance simpler.

*TL;DR:* One vote with caveats for including it here.

rhaschke:
Sorry @felixvd, but I can't follow your argumentation: Of course, I agree with you, that we should increase visibility of this package, provide docs, and ship it frequently.
However, this doesn't require its inclusion here. The pilz planner is a self-contained entity - proven by the fact that it was maintained in a separate repo so far. So, why not keep it like this as this will:
- simplify maintenance with clear responsibilities for this particular package
- reduce bloat of the core repo, which already takes ages ([> 1h](http://build.ros.org/job/Mdev__moveit__ubuntu_bionic_amd64/60/console)) to build and test from scratch on a decent machine (and `pilz_industrial_motion` would [add another 30min](http://build.ros.org/job/Mdev__pilz_industrial_motion__ubuntu_bionic_amd64/55/consoleFull#console-section-18) to this!)

mamoll:
Assuming the technical problem of reducing build time can be worked out, then I'd be in favor of getting the Pilz planner merged in the main repo. Since this code provides functionality that is expected in the industrial robotics community, I suspect the Pilz planner will help MoveIt achieve wider adoption.

rhaschke:
@mamoll, what is your plan to reduce the build time?

v4hn:
I am still somewhat undecided here.

Yes, PTP and LIN support should belong to core MoveIt these days.

Not looking into the code base, I am surprised to hear about *half an hour* of build time for this.
What takes so long there? What's the ratio between build and tests?
I believe the associated test suite is rather extensive?

> @mamoll, what is your plan to reduce the build time?

If we don't find a way to do so, the best way forward indeed seems to be to keep it as a separate repository in `ros-planning`.
We could still depend on it from the `moveit` meta package of course.
But I would actually prefer to split out, for example, CHOMP and allow this planner in instead.

What's the current time required for running all tests in the current repository?
Would it make sense to split of some extensive tests? Off the top of my head,
I seem to recall some time-intense performance tests in the distance field code?

mamoll:
> @mamoll, what is your plan to reduce the build time?

I believe @tylerjw or @JafarAbdi may have some ideas.

davetcoleman:
It is disappointing how much time the Pilz team has put into this PR and how we've followed all the steps of planning this ahead of time - creating an [issue](https://github.com/ros-planning/moveit/issues/1800), discussing at multiple maintainer meetings, having a PR open since February, and now @rhaschke is taking issue with it. They've done everything we've asked of them and more. 

> But I would actually prefer to split out, for example, CHOMP and allow this planner in instead.

This is a good idea. We can also drop some of the repos in the [rosinstall](https://github.com/ros-planning/moveit/blob/master/moveit.rosinstall)

> Not looking into the code base, I am surprised to hear about *half an hour* of build time for this.

Same question...

ct2034:
I want to elaborate a bit on the execution time: You have noticed in the beginning, that we have a lot of tests and also 100% test coverage. Execution time is literally the price you pay for that. ;-) But we have also been working on speeding them up.
There may be some possibilities to speed them up: Especially by getting rid of as many rostests as possible and replacing them by faster running unit tests. But this requires a lot of refactoring within the code as well and I am not sure how much of this we can do in the foreseeable future.

jschleicher:
The build time is [32 seconds](https://travis-ci.org/github/PilzDE/pilz_industrial_motion/jobs/681006365). It's the test base, that takes much of the time: Together with more python tests the [former repo](https://github.com/PilzDE/pilz_industrial_motion/) had

> Function 'run_target_test' returned with code '0' after 19 min 18 sec

That could be optimized - movements run in real time via fake_trajectory_execution and thus take several seconds each. For many of the tests, joint_limits could be drastically increased (use a dummy robot instead of prbt) or we could use shorter movements to just check, that there is robot motion at all. But I'd definitely leave that for follow-up-pull-requests.

rhaschke:
First of all, thanks to @jschleicher for providing better runtime statistics. The 30mins CI time I estimated, was from the ROS buildfarm, which also involves setting up the docker containers. Nevertheless, even 20min is a lot of time to add, particularly if its testing time, because that cannot be reduced by caching.
If caches are fully exploited, Travis can build the current master in [less than 15mins](https://travis-ci.org/github/ros-planning/moveit/builds/678552494). Including the pilz planner in the main repo will more than double this time.

I have no idea, how the overall CI time could be significantly reduced as suggested by @mamoll. Skipping tests is definitely not an option for me, as testing is a crucial part of CI. Furthermore, I think it's kind of wasted time to refactor the code to get from 20mins down to 15mins.

@davetcoleman, yes we agreed in maintainer meetings to migrate the code and I'm still supporting the idea of a) extending MoveIt by this standard feature and b) increasing the visibility of Pilz's planner. However, when we decided on these steps half a year ago, nobody mentioned the 20min build time!

I just argue that this codebase is sufficiently decoupled from the core that we could keep it in a separate repo, but now hosted in the ros-planning organization. This will neither reduce its visibility nor diminish the effort the Pilz team has put into for preparing this migration. (Actually, one could argue that Pilz visibility is even strengthened, because the `pilz-industrial-planner` will become a separate repo in ros-planning. As was argued before, visibility will be achieved by adding corresponding _tutorials_, which - of course - still will be part of `moveit_tutorials` and hosted on the MoveIt websites.

martiniil:
https://api.travis-ci.org/v3/job/681549033/log.txt

I don't understand why the timeout of the clang-tidy check is 20 minutes. The remaining time should have been larger since `MOVEIT_CI_TRAVIS_TIMEOUT=85` and the job ran for 48 minutes.

rhaschke:
The 20min timeout corresponds to the clang-tidy part only. `moveit_ci` keeps track of the overall time consumed so far and only allocates the remaining time for the next subtask.
However, I don't understand right now why the overall time consumed is less than 85min.

jschleicher:
@rhaschke Beeing a repo under the ros-planning umbrella would mean, Pilz is still sole maintainer(s), since you (moveit maintainers and contributors) probably won't see issues and PRs there? What advantage do you see over leaving our planner at the PilzDE organisation and adding a dependency to install our planner from the moveit.rosinstall and moveit debian packages?

I don't want to repeat the discussion of [Migration to one github repo](https://discourse.ros.org/t/migration-to-one-github-repo-for-moveit/266), but I see several parallels: API changes in the last months led to breaking our planner (e.g. PilzDE/pilz_industrial_motion/pull/109), even if it's just a deprecation (and we use `-Werror`). PRs and releases had to happen synchronously etc. Integrating the pilz planner makes development much easier in my opinion regarding that point.

> If caches are fully exploited, Travis can build the current master in [less than 15mins](https://travis-ci.org/github/ros-planning/moveit/builds/678552494). Including the pilz planner in the main repo will more than double this time.

No. It takes 16 minutes now. The tests obviously are only in the `coverage` target:
build-times
Not sure, why clang-tidy time is reduced in this example...?

tylerjw:
While decreasing the run-time of CI is a useful thing to care about, I don't think it should stand in the way of merging code that is well tested.  @jschleicher point about api changes breaking seems like a really good reason to adding this here.

Maybe we could find some way to have travis report the job as succeeding before finishing the code coverage run (as everything that is tested by it is already tested in the quicker release built test run) as a separate improvement.  If we mark the coverage build as allowing it to fail and fast finished as described [here](https://docs.travis-ci.com/user/build-matrix/#fast-finishing) the 20min increase in that won't affect the CI time but will still run the the code coverage job.

tylerjw:
Looking at the travis result, I see you too are being attacked by the kinetic `eigen: too many iterations in Jacobi transform.` bug.  I am looking into this but if anyone else has time/experience with this I'd appreciate help solving this.

rhaschke:
> Beeing a repo under the ros-planning umbrella would mean, Pilz is still sole maintainer(s), since you (moveit maintainers and contributors) probably won't see issues and PRs there?

No, that's a wrong assumption. MoveIt maintainers are responsible for a whole bunch of repos in the ros-planning organization, included (aside of MoveIt) e.g. panda_moveit_config, geometric_shapes, warehouse_ros, ...

Syncing API changes is definitely a good reason to merge into a monolithic repo, I agree. On the other hand we don't have so many API changes for the core planner plugin API, do we?

Comparing Travis results is tricky! Each build has its own cache and thus you should only compare results with a similar level of cache exploitation (which is hard to control in turn).
Best way is to clean cache a build from scratch.

jschleicher:
Regarding the failing kinetic build (needs #2044 first) the remaining issue seems to be API differences in the kinematics plugin. Building `prbt_ikfast_manipulator_plugin` together with master on kinetic [passes Travis](https://travis-ci.org/github/PilzDE/moveit/jobs/682794703).

How should should we integrate the prbt kinematics solver for our test robot?

1. Inside moveit_resources
2. within moveit.rosinstall
3. deprecate kinetic and disable all prbt tests there (to avoid work into an platform that will be deprecated soon)
(4. switch to kdl for tests - seems like larger refactoring)

rhaschke:
> How should should we integrate the prbt kinematics solver for our test robot?

My suggestion is to place the kinematics plugin in a .rosinstall file. As this adds more build time again (probably not that much), I'm even more in favor of a separate repo.

davetcoleman:
I believe the next steps were outlined in our most recent Maintainer Meeting. From the top of my head:

- update moveit_resources to have multiple packages inside, including the robots needed for testing this PR's code
- speed up the integration tests simulation so that CI doesn't time out

is that everything?

jschleicher:
Yep, I think that's all. I just opened https://github.com/ros-planning/moveit_resources/issues/34 for the split in `moveit_resources`. Any help there is greatly appreciated - we won't get through that during the next two weeks :-(

v4hn:
@jschleicher feel free to propose a new description and moveit config for your own pilz robot now to move this forward.

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/command_list_manager.h
```diff
@@ -0,0 +1,222 @@
...
+   * element (until all requests are processed or until the next group change).
+   *
+   * @param planning_scene The planning scene to be used for trajectory generation.
+   * @param req_list List of motion requests containing: PTP, LIN, CIRC
+   * and/or gripper commands.
+   * Please note: A request is only valid if:
+   *    - All blending radii are non negative.
+   *    - The blending radius of the last request is 0.
+   *    - Only the first request of each group has a start state.
+   *    - Non of the blending radii overlapp each other.
```
felixvd, line 90:
```suggestion
   *    - None of the blending radii overlap each other.
```

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/cartesian_limit.h
```diff
@@ -0,0 +1,154 @@
...
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+#pragma once
+
+namespace pilz_industrial_motion_planner
+{
+/**
+ * @brief Set os cartesian limits, has values for velocity, acceleration and deceleration of both the
```
felixvd, line 40:
```suggestion
 * @brief Set of cartesian limits, has values for velocity, acceleration and deceleration of both the
```
?

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/cartesian_limit.h
```diff
@@ -0,0 +1,154 @@
...
+class CartesianLimit
+{
+public:
+  CartesianLimit();
+
+  // Translational Velocity Limit
+
+  /**
+   * @brief Check if translational velocity limit is set.
+   * @return True if limit was set false otherwise
```
felixvd, line 52:
```suggestion
   * @return True if limit was set, false otherwise
```

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/cartesian_limit.h
```diff
@@ -0,0 +1,154 @@
...
+
+  /**
+   * @brief Set the maximal translational velocity
+   * @param Maximum translational velocity [m/s]
+   */
+  void setMaxTranslationalVelocity(double max_trans_vel);
+
+  /**
+   * @brief Return the maximal translational velocity [m/s], 0 if nothing was set
+   * @return maximal translational velocity, 0 if nothing was set
```
felixvd, line 64:
```suggestion
   * @return Maximum translational velocity, 0 if nothing was set
```

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/planning_context_base.h
```diff
@@ -0,0 +1,177 @@
...
+   * descriptions "plan", "simplify", "interpolate"
+   * @param res The detailed response
+   * @return true on success, false otherwise
+   */
+  bool solve(planning_interface::MotionPlanDetailedResponse& res) override;
+
+  /**
+   * @brief Will terminate solve()
+   * @return
+   * @note Currently will not stop a runnning solve but
```
felixvd, line 93: But?

felixvd (commented):
The new commits got me looking again, excited to see it moving. Just scrolled through and found some small mistakes I'm not above nitpicking.

On another note, this seems to depend on `pilz_extensions` only for `pilz_extensions::JointLimit`. Should we implement that type locally to remove the dependency?

ct2034:
> On another note, this seems to depend on `pilz_extensions` only for `pilz_extensions::JointLimit`. Should we implement that type locally to remove the dependency?

Thanks Felix. You have a good point about `pilz_extensions`. I put the topic on our backlog to be solved after the merge is done.

ct2034:
> > On another note, this seems to depend on `pilz_extensions` only for `pilz_extensions::JointLimit`. Should we implement that type locally to remove the dependency?
> 
> Thanks Felix. You have a good point about `pilz_extensions`. I put the topic on our backlog to be solved after the merge is done.

@felixvd https://github.com/ros-planning/moveit/pull/1893/commits/e849f7626f6e8eefe0dd695a6ddd243aebfec944

rfeistenauer:
With  https://github.com/ros-planning/moveit_ci/pull/119 closed the CI should pass after re triggering it.

henningkayser (commented):
I tested all planners and it all worked fine for me without issues. IMO this PR is good to go after my two comments and the remaining requests in https://github.com/ros-planning/moveit_resources/pull/43 have been resolved. For that, please consider adding a template for the planning pipeline launch file and configs [as requested here](https://github.com/ros-planning/moveit_resources/pull/43), or add corresponding TODO+issue and plan for a follow-up. The same applies for the new config `cartesian_limits.yaml`.

TIMING_MEASUREMENTS_LAST_POINT
```diff
@@ -0,0 +1,53 @@
+
```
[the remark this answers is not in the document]
jschleicher: yep, thanks.

moveit_planners/pilz_industrial_motion_planner/README.md
```diff
@@ -0,0 +1,4 @@
+Please consult tutorials and the official [documentation](http://moveit.ros.org/documentation/).
```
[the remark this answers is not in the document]
jschleicher: done.

jschleicher:
Thank you very much @henningkayser ! The latest build failure seems to be an unrelated docker issue.

I'd suggest a squash-merge.

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/pilz_industrial_motion_planner_planning_pipeline.launch.xml
```diff
@@ -0,0 +1,14 @@
...
+  <arg name="planning_plugin" value="pilz_industrial_motion_planner::CommandPlanner" />
+
+  <arg name="planning_adapters" value="" />
+
+  <arg name="start_state_max_bounds_error" value="0.1" />
+
+  <param name="planning_plugin" value="$(arg planning_plugin)" />
+  <param name="request_adapters" value="$(arg planning_adapters)" />
+  <param name="start_state_max_bounds_error" value="$(arg start_state_max_bounds_error)" />
+
```
henningkayser, line 13: Shouldn't this have the new capabilities, like [here](https://github.com/ros-planning/moveit_resources/blob/2d207b4de949c9e4ff80acfd07004133bd054a39/panda_moveit_config/launch/pilz_industrial_motion_planner_planning_pipeline.launch.xml#L15)?

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/pilz_industrial_motion_planner_planning_pipeline.launch.xml
```diff
@@ -0,0 +1,14 @@
+<launch>
+
+  <!-- Pilz Command Planner Plugin for MoveIt! -->
```
henningkayser, line 3:
```suggestion
  <!-- Pilz Command Planner Plugin for MoveIt -->
```

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/pilz_industrial_motion_planner_planning_pipeline.launch.xml
jschleicher: Yep, absolutely. To allow passing in user-defined capabilities as well, I'd suggest to pass the `capabilities` argument down to the included pipeline launch file(s) and append the planner-specific ones there: 99c059f

I'll re-generate the configuration inside moveit_resources based on this template.

davetcoleman:
The RPBT PR is merged! I think this is ready to go.

@jschleicher would still be nice to improve usability for future users via @henningkayser's suggestion:

> For that, please consider adding a template for the planning pipeline launch file and configs as requested here, or add corresponding TODO+issue and plan for a follow-up. The same applies for the new config cartesian_limits.yaml.

henningkayser:
> The RPBT PR is merged! I think this is ready to go.
> 
> @jschleicher would still be nice to improve usability for future users via @henningkayser's suggestion:
> 
> > For that, please consider adding a template for the planning pipeline launch file and configs as requested here, or add corresponding TODO+issue and plan for a follow-up. The same applies for the new config cartesian_limits.yaml.

@davetcoleman Thanks for merging moveit_resources. Actually, the templates have already been added here.
@jschleicher please update the rosinstall so that we can merge this.

felixvd:
Absolutely cannot wait to see this with multiple planning pipelines. Super excited for these features <3

henningkayser (approved):

ct2034:
Thank you very much. @henningkayser

gonzalocasas:
🍾

PR #1894, opened 2020-02-12
added options to start and end state dropdowns

JStech:
The current default behavior is unintuitive, and may be wrong. Changing
the start state in the graphical interface does not impact the plan
generated--it just starts from the robot's current state. The sematics
of `<current>` are unclear, and apparently different for the start state
and the end state.

I added the "Query Start State" option to the Start State dropdown menu,
and similarly "Query Goal State" for the Goal State menu. These options
intuitively correspond to the states defined in the graphical interface.
I also made them the default options.

### Description

### Checklist

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -283,12 +283,14 @@ void MotionPlanningFrame::fillStateSelectionOptions()
   const robot_model::JointModelGroup* jmg = robot_model->getJointModelGroup(group);
   if (jmg)
   {
+    ui_->start_state_combo_box->addItem(QString("Query Start State"));
     ui_->start_state_combo_box->addItem(QString("<random valid>"));
     ui_->start_state_combo_box->addItem(QString("<random>"));
     ui_->start_state_combo_box->addItem(QString("<current>"));
     ui_->start_state_combo_box->addItem(QString("<same as goal>"));
     ui_->start_state_combo_box->addItem(QString("<previous>"));
 
+    ui_->goal_state_combo_box->addItem(QString("Query Goal State"));
     ui_->goal_state_combo_box->addItem(QString("<random valid>"));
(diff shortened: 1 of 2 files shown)


rhaschke:
Sorry, this PR doesn't make sense to me. The comboboxes are used to initialize the corresponding state in the GUI interface. There is no need to add "Query Start State" and "Query Goal State" to the start resp. goal combo box.

> The sematics of "" are unclear

What are you referring to? There shouldn't be an empty string entry. If so, it probably comes from a named pose with an empty name.

JStech:
Sorry, that empty string is supposed to be `<current>`.

I was working through the "MoveIt Quickstart in RViz" tutorial, but changing the Query Start State in the interactive interface did not affect the planner--it always started from the robot's initial state. So, when you get to the point shown in [this image](https://ros-planning.github.io/moveit_tutorials/_images/rviz_plugin_planned_path.png), the green arm is ignored--instead, this happens:
Selection_002

Maybe I'm misunderstanding the semantics of those comboboxes. The only way I could get it to plan from an initial state of my choosing was by selecting `<random>` and then moving the arm.

rhaschke:
Thanks for explaining the underlying issue in more detail. The culprit is the following code:
https://github.com/ros-planning/moveit/blob/bd85e7d78e3992b8671d3a1814bcd7eaee2ac076/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp#L175-L177

which explicitly resets the modified start state to the current state. I introduced this code in #1772 for a good reason. Obviously we need another fix to it.

rhaschke:
Hm. I can't reproduce the issue of #1772 after reverting it. Obviously this solves the issue. I will commit this fix...

rhaschke:
This code, which updates both start and goal state on scene updates, should actually solve #1772 https://github.com/ros-planning/moveit/blob/bd85e7d78e3992b8671d3a1814bcd7eaee2ac076/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp#L1216-L1221

Issue #1835, opened 2020-01-16
Issues with PlanningScene updates


rhaschke:
As pointed out in https://github.com/ros-planning/moveit_task_constructor/issues/122#issuecomment-574659080, PlanningScene updates [performed by MTC's `ExecuteTaskSolutionCapability`](https://github.com/ros-planning/moveit_task_constructor/blob/ef6cb1a746353bf2b11eb986c9f690e1c785e623/capabilities/src/execute_task_solution_capability.cpp#L179), are broken in the master branch (latest binary release still works!).

If I augment the mentioned call with a 0.5s sleep, the spurious collision is gone and trajectory execution finishes. Probably there is a race condition.

felixvd:
I'll double check tomorrow if #1439 is involved, since that involved changes in the `processAttachedCollisionObject` and `processCollisionObject` methods. For some reason I can't run the MTC demo at all at the moment - Rviz does not show a robot or the example task.

rhaschke:
#1439 is not the culprit. I'm currently bisecting the issue. Currently, #1808 is still good.

rhaschke:
#1439 is not the culprit. Bisecting the issue shows that #1584 is the culprit.
@j-petit, do you have any idea, why the unified collision environment might exhibit a race condition?
I remember that we had a similar issue in the past.

j-petit:
Hm, not off the top of my head... Is this is executed as multi-threaded collision checking?  I'll try to look more closely at the code again today.

>  I remember that we had a similar issue in the past.

Before the introduction of the unified environment or after?

rhaschke:
> I remember that we had a similar issue in the past.

After introducing the unified environment.

The issue occurs for a single-thread environment using FCL.

j-petit:
What might be the reason is that [the callback](https://github.com/ros-planning/moveit/blob/06a9e4beac1e94d862c8286a22561529cb03d638/moveit_core/collision_detection/src/world.cpp#L319) which deletes an object when it changes from environment to being attached is executed only after the object is added to the `robot_state` as being attached and a collision check performed.

rhaschke:
The PlanningScene's `World` and `RobotState` is correct before the collision check is ran as the following debug output illustrates (_attached_ and _free_ object lists correctly list the right objects):
```
[DEBUG] ros.moveit_task_constructor_capabilities.ExecuteTaskSolution: apply effect of 8/17
attached: 0x7f14b4001770 object, 
[...]

PR #1899, opened 2020-02-13
"object" collides with "object": unconditionally allocate unpadded collision environments

v4hn:
I just read through https://github.com/ros-planning/moveit/issues/1835 and https://github.com/ros-planning/moveit_task_constructor/issues/122 which describe a problem with collision checking of attached objects ending up in self-collision.

I am sorry. I debugged this problem in a hurry already back in October for a demo,
but did not get around to investigate further or create a thread for it. :frowning_face: 

This patch resolved the problem for me, although it is likely not the best solution.
Without digging deeper into this again, here's what I recall:

- Collision detectors maintain separate collision structures for padded and unpadded collision checking
- Apparently the unpadded collision checking is thought to be less popular and is only allocated when it is requested through API that allows to write to it
- As long as it is not allocated, a standard pattern in the PlanningScene structures is applied and **the parents** unpadded collision structure is used if required
- @j-petit 's global unification of `CollisionRobot` and `CollisionWorld` into `CollisionEnv` seems to have exposed the issue to trigger with MTC's execution capability, although I'm not convinced the logic was perfectly sane before
- Currently, things go wrong the moment an object becomes attached in a diff planning scene
  - it is removed from the collision world
  - it is added to the collision robot structure
  - during *unpadded* collision checking, the parent's (unpadded) collision world is returned which still contains the object
  - Thus the object suddenly collides with itself

The patch addresses this logic bug simply by always allocating the unpadded collision environment.
I am not convinced it is worth the effort to try to fix the lazy allocation scheme here, but I did not benchmark anything and this only becomes relevant when the benchmarks include attach/detach operations...

On top of that, I am not convinced this double-accounting for padded and unpadded checking makes sense anymore(?), but I did not look into the underlying structures deep enough to understand this.

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -204,10 +204,8 @@ PlanningScene::PlanningScene(const PlanningSceneConstPtr& parent) : parent_(pare
     detector->cenv_ = detector->alloc_->allocateEnv(parent_detector->cenv_, world_);
     detector->cenv_const_ = detector->cenv_;
 
-    // leave these empty and use parent collision_env_ unless/until a non-const one
-    // is requested (e.g. to modify link padding or scale)
-    detector->cenv_unpadded_.reset();
-    detector->cenv_unpadded_const_.reset();
+    detector->cenv_unpadded_ = detector->alloc_->allocateEnv(parent_detector->cenv_unpadded_, world_);
+    detector->cenv_unpadded_const_ = detector->cenv_unpadded_;
   }
   setActiveCollisionDetector(parent_->getActiveCollisionDetectorName());
 }
@@ -286,12 +284,8 @@ void PlanningScene::addCollisionDetector(const collision_detection::CollisionDet
     detector->copyPadding(*active_collision_);
   }
 
-  // If we don't have a parent, allocate unpadded versions, otherwise use the parent's cenv_unpadded_.
-  if (!detector->parent_)
-  {
-    detector->cenv_unpadded_ = detector->alloc_->allocateEnv(world_, getRobotModel());
-    detector->cenv_unpadded_const_ = detector->cenv_unpadded_;
-  }
+  detector->cenv_unpadded_ = detector->alloc_->allocateEnv(world_, getRobotModel());
+  detector->cenv_unpadded_const_ = detector->cenv_unpadded_;
 }
 
 void PlanningScene::setActiveCollisionDetector(const collision_detection::CollisionDetectorAllocatorPtr& allocator,
@@ -393,11 +387,11 @@ void PlanningScene::clearDiffs()
 
     if (it.second->parent_)
     {
-      it.second->cenv_unpadded_.reset();
-      it.second->cenv_unpadded_const_.reset();
-
       it.second->cenv_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_, world_);
       it.second->cenv_const_ = it.second->cenv_;
+
+      it.second->cenv_unpadded_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_unpadded_, world_);
+      it.second->cenv_unpadded_const_ = it.second->cenv_unpadded_;
     }
     else
     {
@@ -1138,11 +1132,6 @@ void PlanningScene::decoupleParent()
 
   for (std::pair<const std::string, CollisionDetectorPtr>& it : collision_)
   {
-    if (!it.second->cenv_unpadded_)
-    {
-      it.second->cenv_unpadded_ = it.second->alloc_->allocateEnv(it.second->parent_->getCollisionEnvUnpadded(), world_);
-      it.second->cenv_unpadded_const_ = it.second->cenv_unpadded_;
-    }
     it.second->parent_.reset();
   }
   world_diff_.reset();
--- a/moveit_core/planning_scene/test/test_planning_scene.cpp
+++ b/moveit_core/planning_scene/test/test_planning_scene.cpp
@@ -125,6 +125,13 @@ TEST(PlanningScene, LoadRestoreDiff)
   EXPECT_EQ(next->getWorld()->size(), 2u);
   EXPECT_EQ(ps->getWorld()->size(), 1u);
 
+  /* the worlds used for collision detection contain one and two objects, respectively */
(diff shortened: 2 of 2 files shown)


rhaschke:
@v4hn, I don't see how your explanation and your fix fits my observation of a race condition: When adding a small sleep after a PlanningScene update in MTC's execution capability, the issue is gone.
But, adding a sleep doesn't trigger an extra environment allocation, does it?

simonschmeisser:
I think the right fix would be to find the place where this copy-on-write is broken (ie making a copy is not triggered) instead of removing it? I would expect your changes to have some impact in cases like multi-threaded planners etc.?

v4hn:
> I think the right fix would be to find the place where this copy-on-write is broken (ie making a copy is not triggered) instead of removing it?

I agree but did not have the time to do so, especially as this whole part was rearranged lately...

> @v4hn, I don't see how your explanation and your fix fits my observation of a race condition: When adding a small sleep after a PlanningScene update in MTC's execution capability, the issue is gone.

Did you try the patch?

I am not sure anymore whether an additional sleep fixed my problems, I tried to find the actual problem.
I could imagine the [PlanningSceneMonitor pushes its diff](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L381) if you wait long enough, but sleeping 0.1 seems rather short for that.

> I would expect your changes to have some impact in cases like multi-threaded planners etc.?

You are welcome to setup a benchmark to show that. As I said, a-priori I'm not convinced the whole structural overhead there is justified.

j-petit:
> - Collision detectors maintain separate collision structures for padded and unpadded collision checking
> - Apparently the unpadded collision checking is thought to be less popular and is only allocated when it is requested through API that allows to write to it
> - As long as it is not allocated, a standard pattern in the PlanningScene structures is applied and the parent's unpadded collision structure is used if required

That is also my understanding of the current status-quo. Maybe [this discussion](https://github.com/ros-planning/moveit/pull/1584#discussion_r308044375) helps?

> On top of that, I am not convinced this double-accounting for padded and unpadded checking makes sense anymore(?), but I did not look into the underlying structures deep enough to understand this

When working on this part of MoveIt, I also had the impression that the efficiency gains are not worth the complexity added (anymore?). The code is very difficult to understand.

v4hn:
I'm undecided whether I want to see this merged (temporarily?) and mainly wanted to share the insights.
If you feel like debugging further @rhaschke , please go ahead.

Incidentally, an alternative workaround for the issue is to remove `Unpadded` [from the plan_execution](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/plan_execution/src/plan_execution.cpp#L286), thus enforcing padding for the collision-monitoring too.
I feel this would work around the current problem with less of a hypothetical hard-to-benchmark-but-possibly-existing impact, but of course it still leaves the unpadded worlds in an inconsistent state under some conditions. Maybe you would prefer this as a temporary way to fix upstream @rhaschke ?

v4hn:
As a side note: could someone explain to me how this patch manages to *reduce* coverage in the checker?
It reduces LOC and even branches.

felixvd:
It might be deleting LOC that were covered by tests, so the percentage of uncovered code grows?

rhaschke:
> Incidentally, an alternative workaround for the issue is to remove Unpadded from the plan_execution, thus enforcing padding for the collision-monitoring too.

Yes, this change seems to be less source-intrusive. I would like to avoid workarounds that strongly change the collision code without fully understanding the underlying problem. However, I don't have the resources right now to further investigate. For MTC, introducing a 0.1s sleep "solved" the issue. As there are no other reported failures due to this bug, I prefer this as the least-intrusive workaround. Nevertheless, your analysis will be very helpful.

v4hn:
> > Incidentally, an alternative workaround for the issue is to remove Unpadded from the plan_execution, thus enforcing padding for the collision-monitoring too.
> 
> Yes, this change seems to be less source-intrusive. I would like to avoid workarounds that strongly change the collision code without fully understanding the underlying problem.

To add a quick note here: This does not seem to be a solution as `Padded` collision checking might yield errors when picking up things from a table

v4hn:
> I think the right fix would be to find the place where this copy-on-write is broken (ie making a copy is not triggered) instead of removing it?

I spent a few hours on this the other night and it turns out the whole logic of unpadded collision checking on `diff()` scenes was broken for ages (if it ever worked).
I added [a test](https://github.com/ros-planning/moveit/pull/1899/commits/7d047dca1871402221f4e92c4543343889fa8485) to the planning scene for this.

Note that this bug affects the standard use of the `PlanningSceneMonitor` which maintains diffs.

Whereas a source comment mentions copy-on-write, this is actually not implemented.
Unpadded collision checking always uses the parent's collision world.
The only two ways to get correct unpadded collision checking for a `diff()` scene are
(1) by decoupling it from the parent, which disables the whole concept of the diff.
(2) by pushing the diff to the parent.

To make the scene structure consistent again - objects in the scene must be respected for all types of collision checking - I propose to merge the original patch from this pull-request.
Consistency *has* to come before performance and as CI demonstrates, the patch fixes the test.

I consider the alternative solution, to implement a lazy allocation when unpadded checking is requested, inferior as it would introduce more fragile logic in an overcomplicated datastructure.

@rhaschke

j-petit:
Great work, thanks. Does this also fix #1835?

simonschmeisser:
@v4hn @rhaschke has this been included in this or the previous melodic release? if so, which one? I just got a call about some race conditions and it could save some of my holidays if this were included in the latest release? (Assuming it's actually this what we're seeing ... )

v4hn:
Sorry Simon, but this was never backported to kinetic/melodic. The exact patch requires the new `CollisionEnv` structure.
It should be straight-forward to get the same functionality for melodic though and, as always, a patch is welcome. :)

Sorry about the interruption from your vacation.

simonschmeisser:
Ok, I thought so as well.

Unfortunately there are no copies of `PlanningScenes` involved in where we run into issues so simply backporting this most likely won't fix our issues. Will need more investigations after holidays ...

v4hn:
Sounds like you need to figure out why your robot has an attached object then, when the object is still part of the world as well. That's the second kind of problem where I've seen this error before.

Happy debugging.

PR #1904, opened 2020-02-15
Fix usages of panda_moveit_config

### LINKED ISSUE #1903: why 'rosdep install' install some moveit binary package when i installed moveit from source?
### Description
I followed the direction of https://moveit.ros.org/install/source/ and ran commands step by step. When running `rosdep install -y --from-paths src --ignore-src --rosdistro ${ROS_DISTRO}`, I noticed it started to install `ros-melodic-panda-moveit-config` and also some additional packages like `ros-melodic-moveit-core`, `ros-melodic-moveit-ros-planning`, etc. I was confused why these packages were installed while I had added `--ignore-src` ?

Overview of your issue here.

### Your environment
* ROS Distro: [Melodic]
* OS Version: Ubuntu 18.04
* Source, master-branch 

### Steps to reproduce
Following the direction of https://moveit.ros.org/install/source/ step by step.
When getting `rosdep install -y --from-paths src --ignore-src --rosdistro ${ROS_DISTRO}`.

### Expected behaviour
Some additional packages like `ros-melodic-moveit-core`, `ros-melodic-moveit-ros-planning`, etc have been installed.

### Actual behaviour
Additional packages like `ros-melodic-moveit-core`, `ros-melodic-moveit-ros-planning`, etc should not been installed.

Use [gist.github.com](gist.github.com) to copy-paste the console output or segfault backtrace using gdb.

rhaschke:
As pointed out in https://github.com/ros-planning/moveit/issues/1903#issuecomment-586569789, following our [install-from-source procedure](https://moveit.ros.org/install/source/), will install `panda_move_config` and, as their dependencies, several other MoveIt packages from debians.
This is usually not desired.
Interestingly, none of the MoveIt packages actually package-depend on panda_moveit_config.
However, `moveit_tutorials` does. As this, but not `panda_moveit_config` is listed in `moveit.rosinstall`, all these binary packages get installed.
Searching for panda_moveit_config in the source base, I found several uses of the package, which will get replaced by the corresponding `moveit_resources` package now.

Please merge as a merge-commit to keep all three, independent commits.
Fixes #1903.
Relies on https://github.com/ros-planning/moveit_resources/pull/28

### DIFF
--- a/moveit.rosinstall
+++ b/moveit.rosinstall
@@ -24,3 +24,7 @@
     local-name: moveit_tutorials
     uri: https://github.com/ros-planning/moveit_tutorials.git
     version: master
+- git:
+    local-name: panda_moveit_config
+    uri: https://github.com/ros-planning/panda_moveit_config.git
+    version: melodic-devel
--- a/moveit_planners/trajopt/test/trajectory_test.test
+++ b/moveit_planners/trajopt/test/trajectory_test.test
@@ -3,7 +3,7 @@
 
   <test pkg="moveit_planners_trajopt" type="trajectory_test" test-name="trajectory_test" time-limit="300" args="">
 
-    <rosparam command="load" file="$(find panda_moveit_config)/config/trajopt_planning.yaml"/>
+    <rosparam command="load" file="$(find moveit_resources)/panda_moveit_config/config/trajopt_planning.yaml"/>
 
   </test>
 
--- a/moveit_ros/benchmarks/examples/demo_panda.launch
+++ b/moveit_ros/benchmarks/examples/demo_panda.launch
@@ -3,17 +3,17 @@
   <arg name="bench_opts" default="$(find moveit_ros_benchmarks)/examples/demo1.yaml"/>
 
   <!-- Load robot settings -->
-  <include file="$(find panda_moveit_config)/launch/planning_context.launch">
+  <include file="$(find moveit_resources)/panda_moveit_config/launch/planning_context.launch">
     <arg name="load_robot_description" value="true"/>
   </include>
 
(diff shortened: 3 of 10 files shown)


moveit.rosinstall
```diff
@@ -24,3 +24,11 @@
     local-name: moveit_tutorials
     uri: https://github.com/ros-planning/moveit_tutorials.git
     version: master
+- git:
+    local-name: panda_moveit_config
+    uri: https://github.com/ros-planning/panda_moveit_config.git
+    version: melodic-devel
+- git:
+    local-name: moveit_resources
+    uri: https://github.com/rhaschke/moveit_resources.git
```
mamoll: Shouldn't that be https://github.com/ros-planning/moveit_resources?

Maybe I am missing a bit of MoveIt history here, but why duplicate the panda_moveit_config in moveit_resources if it needs to be kept in sync with the one in https://github.com/ros-planning/panda_moveit_config?

rhaschke: The corresponding commit comment indicates this change as temporary. As long as https://github.com/ros-planning/moveit_resources/pull/28 is not yet merged and released, Travis would fail otherwise.
We maintain a separate moveit_resources package to resolve circular package dependencies due to MoveIt -> panda_moveit_config -> MoveIt. This PR exactly tries to resolve this circular dependency.

mlautman: Now that  ros-planning/moveit_resources#28 is merged, can we revert this?

rhaschke: No, the release is not yet synced.

rhaschke: Melodic release is synced, but Kinetic not yet.

Issue #130, opened 2016-08-23
Moveit Treats Visual Mesh as Collision Mesh When No Collision Mesh is Specified


mfbailey91:
Although, some may argue that this is a check for when URDFs are constructed improperly and links are not given collision meshes I would argue that promoting improper creation of URDF is a negative and it limits the modularity as moveit as a tool.

This prevents the use of virtual links which one may want to use as visual feedback in RVIZ.  Examples of when you may want only a visual mesh is: if you want to use a virtual link visual mesh to outline a robots reachable work space, or specify some other visual indicator in the work space of the manipulator, or if you want some visual indicator that moves with the arm you may add a virtual fixed joint to a manipulator link. 

Because the visual mesh is taken as the collision mesh if one wants to build in some visual feedback that intersects with the workspace they have to create some virtual collision mesh that then doesn't intersect with the workspace, which is a poor work around.

Links with visual meshes and no collision meshes should be treated as such and the visual mesh should not act as a collision mesh when no collision mesh is given.

de-vri-es:
I ran into the same problem. I wanted a visual link without collision model, but had to resort to disabling collisions with the link and every other link in the allowed collision matrix of the SRDF.

Changing this unconditionally could break existing systems though, and especially in the case of collision detection you don't want to suddenly certain checks in existing systems. So this would have to be an option that defaults to the old behaviour I think.

That leaves the question what would be a logical place for such an option. In the URDF on a per-robot or per-link basis, or make it configurable in code and load it from the parameter server where applicable?

gavanderhoorn:
@mfbailey91 wrote:
> Links with visual meshes and no collision meshes should be treated as such and the visual mesh should not act as a collision mesh when no collision mesh is given.

I may be a bit conservative, but I wouldn't be comfortable with changing MoveIt's default behaviour here. Collision checking is part of MoveIt's 'safety features', and I've seen enough configs that actually (implicitly) depend on MoveIt defaulting to using the visual geometry for the collision mesh. That is certainly not nice (and the author of the URDF shouldn't have done that), but with such a large installed base I don't think removing this behaviour is an option.

So it would need to be configurable, as @de-vri-es suggests.

> That leaves the question what would be a logical place for such an option. In the SRDF or URDF on a per-robot or per-link basis, or make it configurable in code and load it from the parameter server where applicable?

[...]

PR #1909, opened 2020-02-21
Ensure consistency when overriding collision geometry with visual

### LINKED ISSUE #1907: rviz collision is inaccurate when robot visual is used to override collision
When there is no collision specified for a link, but there is a visual specified, the visual geometry is taken as the collision geometry (See for example #130).  However, in rviz, the collision for the link is not shown as there is no collision geometry for the object.  This can lead to hard to debug problems as collisions may not appear to be collisions in rviz.

The RobotStateVisualization for rviz should also use the visual geometry as collision geometry when a link is empty.  I also think a warning should be issued when moveit overrides the collision geometry.

Dale-Koenig:
### Description

One approach to fix  #1907 
When visual geometry is specified for a link but not collision geometry, RobotModel overrides the collision geometry using the visual geometry when creating shapes.  This PR changes RobotModel to instead take ownership of it's own copy of the `urdf::ModelInterface`, and to modify it before generating the model.  This ensures that anything that touches the RobotModel in the future, such as the rviz display functionality, uses the same collision geometry.

I'm a bit hesitant to change the constructor for an important class like this so I'm open to other approaches.

Also, would this need special handling for when the visual array is empty?  I tested with a link that had a single visual element and it still had a non-empty Visual array, but it seems like the previous code had special handling for the case where the visual array is empty.

### DIFF
--- a/moveit_core/robot_model/src/robot_model.cpp
+++ b/moveit_core/robot_model/src/robot_model.cpp
@@ -55,9 +55,32 @@ const std::string LOGNAME = "robot_model";
 RobotModel::RobotModel(const urdf::ModelInterfaceSharedPtr& urdf_model, const srdf::ModelConstSharedPtr& srdf_model)
 {
   root_joint_ = nullptr;
-  urdf_ = urdf_model;
+  urdf_.reset(new urdf::ModelInterface(*urdf_model));
   srdf_ = srdf_model;
-  buildModel(*urdf_model, *srdf_model);
+
+  // Override empty collision geometry in stored with visual geometry when only the latter is specified
+  std::vector<urdf::LinkSharedPtr> links;
+  urdf_->getLinks(links);
+  for (const urdf::LinkSharedPtr& link : links)
+  {
+    if (!link->visual_array.empty() && link->collision_array.empty())
+    {
+      ROS_WARN_STREAM_NAMED(LOGNAME, "Link " << link->name << " has visual geometry but no collision geometry "
+                                                              "specified.  Using visual geometry to override "
+                                                              "collision geometry");
+      for (const urdf::VisualSharedPtr& visual : link->visual_array)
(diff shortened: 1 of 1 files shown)


rhaschke:
This PR solves the issue only for MoveIt-related rviz plugins. The rviz-builtin RobotModel will not react to those changes. As explained in https://github.com/ros-planning/moveit/issues/130#issuecomment-589544315, I prefer a clear-cut solution, returning to the standard ROS behavior within a deprecation phase.

Dale-Koenig:
> This PR solves the issue only for MoveIt-related rviz plugins. The rviz-builtin RobotModel will not react to those changes. As explained in [#130 (comment)](https://github.com/ros-planning/moveit/issues/130#issuecomment-589544315), I prefer a clear-cut solution, returning to the standard ROS behavior within a deprecation phase.

Fair enough.  I think it would still be best to have rviz as accurate as possible over the current situation, but as long as there is a scary enough error message it's probably not worth putting in partial measures.

rhaschke:
Could you provide a corresponding PR to implement the suggested behavior in `RobotModel::constructLinkModel`?

Dale-Koenig:
@shaschke It's not completely clear to me that there was a conclusion made in terms of how to handle this, or what that conclusion is.  By a parameter to disable the collision overriding, do you just mean an extra argument to the `RobotModel` constructor?

rhaschke:
In #130, we didn't yet came up with a final conclusion. My suggestion was to use a new ROS parameter `moveit_populate_empty_collision_geometry`, that should be evaluated within `RobotModel`.

Dale-Koenig:
@rhaschke I posted my opinion in the issue thread.  I'm happy to make a PR, but I think I will wait until a conclusion is made (or until it seems this has been forgotten about)

PR #1911, opened 2020-02-21
Jog cpp api fixes

AndyZe:
This cleans up a few things in jog_arm:

- Renames JogCppApi class to JogCppInterface to match the filename

- Fix a crash when the dimensions of the Jacobian changed

- Doxygenize some comments

- Make the inheritance of JogInterfaceBase public. I wanted this so I could call the functions from external C++ code, like this:  `jog_interface_ptr_->changeDriftDimensions(...)`

### DIFF
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -102,8 +102,7 @@ protected:
   // Possibly calculate a velocity scaling factor, due to proximity of
   // singularity and direction of motion
   double decelerateForSingularity(const Eigen::VectorXd& commanded_velocity,
-                                  const Eigen::JacobiSVD<Eigen::MatrixXd>& svd,
-                                  const Eigen::MatrixXd& jacobian);
+                                  const Eigen::JacobiSVD<Eigen::MatrixXd>& svd, const Eigen::MatrixXd& jacobian);
 
   /**
    * Slow motion down if close to singularity or collision.
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -101,8 +101,9 @@ protected:
 
   // Possibly calculate a velocity scaling factor, due to proximity of
   // singularity and direction of motion
-  double decelerateForSingularity(const Eigen::VectorXd& commanded_velocity,
-                                  const Eigen::JacobiSVD<Eigen::MatrixXd>& svd, const Eigen::MatrixXd& jacobian);
+  double velocityScalingFactorForSingularity(const Eigen::VectorXd& commanded_velocity,
+                                             const Eigen::JacobiSVD<Eigen::MatrixXd>& svd,
+                                             const Eigen::MatrixXd& jacobian, const Eigen::MatrixXd& pseudo_inverse);
 
   /**
    * Slow motion down if close to singularity or collision.
@@ -154,9 +155,6 @@ protected:
 
   JogArmParameters parameters_;
 
-  // Preallocate for jacobian calculations
-  Eigen::MatrixXd pseudo_inverse_, matrix_s_;
-  Eigen::JacobiSVD<Eigen::MatrixXd> svd_;
   // Use ArrayXd type to enable more coefficient-wise operations
   Eigen::ArrayXd delta_theta_;
 
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -102,7 +102,8 @@ protected:
   // Possibly calculate a velocity scaling factor, due to proximity of
   // singularity and direction of motion
   double decelerateForSingularity(const Eigen::VectorXd& commanded_velocity,
-                                  const Eigen::JacobiSVD<Eigen::MatrixXd>& svd);
+                                  const Eigen::JacobiSVD<Eigen::MatrixXd>& svd,
+                                  const Eigen::MatrixXd& jacobian);
 
   /**
    * Slow motion down if close to singularity or collision.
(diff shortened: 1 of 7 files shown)


moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -751,17 +750,19 @@ bool JogCalcs::addJointIncrements(sensor_msgs::JointState& output, const Eigen::
 
 void JogCalcs::removeDimension(Eigen::MatrixXd& jacobian, Eigen::VectorXd& delta_x, unsigned int row_to_remove)
 {
-  unsigned int num_rows = jacobian.rows() - 1;
-  unsigned int num_cols = jacobian.cols();
+  /*
```
henningkayser: Why did you remove this?

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -500,8 +500,7 @@ double JogCalcs::decelerateForSingularity(const Eigen::VectorXd& commanded_veloc
   new_theta += pseudo_inverse_ * delta_x;
   kinematic_state_->setJointGroupPositions(joint_model_group_, new_theta);
 
-  jacobian_ = kinematic_state_->getJacobian(joint_model_group_);
-  Eigen::JacobiSVD<Eigen::MatrixXd> new_svd = Eigen::JacobiSVD<Eigen::MatrixXd>(jacobian_);
+  Eigen::JacobiSVD<Eigen::MatrixXd> new_svd = Eigen::JacobiSVD<Eigen::MatrixXd>(jacobian);
```
henningkayser:
```suggestion
  Eigen::JacobiSVD<Eigen::MatrixXd> new_svd(jacobian);
```

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: Oops, it was commented for debugging. Will restore.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -154,8 +154,8 @@ class JogCalcs
 
   JogArmParameters parameters_;
 
-  // For jacobian calculations
-  Eigen::MatrixXd jacobian_, pseudo_inverse_, matrix_s_;
+  // Preallocate for jacobian calculations
```
jliukkonen: I would reword comments since the lines below don't really pre-allocate memory but rather cache results that are calculated once(?) Correct me though if I have misunderstood the way Eigen allocates memory for dynamic matrices.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -474,23 +474,24 @@ bool JogCalcs::applyVelocityScaling(const JogArmShared& shared_variables, std::m
 
 // Possibly calculate a velocity scaling factor, due to proximity of singularity and direction of motion
 double JogCalcs::decelerateForSingularity(const Eigen::VectorXd& commanded_velocity,
```
jliukkonen: Could this function be renamed? It calculates a velocity scaling factor but it sounds like it commands the manipulator to decelerate when close to singularity. Maybe `velocityScalingFactorForSingularity` or something like that.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
mlautman: I don't like pre-allocating memory by creating member variables unless there constructor is non-trivial (eg a RobotState variable). It seems like poor design because you never know how they may be used elsewhere and there is no reason why their values should be preserved. Seems like a use case for a static variable instead..

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_cpp_interface.h
```diff
@@ -44,15 +44,15 @@
 namespace moveit_jog_arm
 {
 /**
-* Class JogCppApi - This class should be instantiated in a new thread
+* Class JogCppInterface - This class should be instantiated in a new thread
 * See cpp_interface_example.cpp
 */
-class JogCppApi : protected JogInterfaceBase
+class JogCppInterface : public JogInterfaceBase
```
mlautman: Since this is in moveit_experimental in the master branch, this is okay but in general, it is really bad practice to change public interfaces without deprecating the old one.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -321,28 +321,28 @@ bool JogCalcs::cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, JogArmShared&
   Eigen::VectorXd delta_x = scaleCartesianCommand(cmd);
 
   // Convert from cartesian commands to joint commands
-  jacobian_ = kinematic_state_->getJacobian(joint_model_group_);
+  Eigen::MatrixXd jacobian = kinematic_state_->getJacobian(joint_model_group_);
```
mlautman: If you are worried about performance, you could make this static as well

moveit_experimental/moveit_jog_arm/src/jog_cpp_interface.cpp
```diff
@@ -39,13 +39,13 @@
 
 #include "moveit_jog_arm/jog_cpp_interface.h"
 
-// TODO(davetcoleman): rename JogCppApi to JogCppInterface to match file name
+// TODO(davetcoleman): rename JogCppInterface to JogCppInterface to match file name
```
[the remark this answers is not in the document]
AndyZe: done

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
AndyZe: These variables get updated frequently in a while loop (like, 100 times/second or so) so performance kind of matters. I guess I'm not clear how static variables would be better, but I want to know

I'll make them all local variables until I hear back otherwise

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: good idea, done

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp

AndyZe: Strangely, making this static causes a crash. Probably due to some obscure Eigen memory management issue?

AndyZe: I think the problem is:

> Static variables when used inside function are initialized only once, and then they hold their value even through function calls.

https://www.studytonight.com/cpp/static-keyword.php

mlautman: Sounds like some janky Eigen stuff for sure

PR #1912, opened 2020-02-25
Adding panda_moveit_config as test_depend

ct2034:
### Description
I had problems because this test https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/test/moveit_cpp_test.test uses `panda_moveit_config` which was no test dependency.


ct2034:
Ok, I see. It is solved by #1904 then: https://github.com/ros-planning/moveit/pull/1904/files#diff-a6c38af7b7eb0f6aa218eca2bda70b42

PR #1915, opened 2020-02-25
[jog_arm] Publish more detailed warnings

AndyZe:
Rather than just a boolean, this publishes a more detailed status message.

I thought about publishing a vector of status codes, but some downsides of that are:

- it would require a custom message type

- the vector would have to be resized, which isn't great from a performance perspective. The code to handle that would be more complicated.

### DIFF
--- a/moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
+++ b/moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
@@ -47,7 +47,7 @@ joint_limit_margin: 0.1 # added as a buffer to joint limits [radians]. If moving
 cartesian_command_in_topic: jog_server/delta_jog_cmds  # Topic for incoming Cartesian twist commands
 joint_command_in_topic: jog_server/joint_delta_jog_cmds # Topic for incoming joint angle commands
 joint_topic:  joint_states
-warning_topic: jog_server/halted # Publish boolean warnings to this topic
+status_topic: jog_server/status # Publish status to this topic
 command_out_topic: joint_group_position_controller/command # Publish outgoing commands here
 
 ## Collision checking for the entire robot body
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -181,7 +181,7 @@ protected:
 
   ros::Publisher status_pub_;
 
-  StatusCode status_ = kNoWarning;
+  StatusCode status_ = NO_WARNING;
 
   JogArmParameters parameters_;
 
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -81,37 +81,48 @@ protected:
   // Flag that robot state is up to date, end effector transform is known
   std::atomic<bool> is_initialized_;
 
+  /** \brief Do jogging calculations for Cartesian twist commands. */
   bool cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, JogArmShared& shared_variables, std::mutex& mutex);
 
+  /** \brief Do jogging calculations for direct commands to a joint. */
   bool jointJogCalcs(const control_msgs::JointJog& cmd, JogArmShared& shared_variables);
 
   /** \brief Update the stashed status so it can be retrieved asynchronously */
   void updateCachedStatus();
 
-  // Parse the incoming joint msg for the joints of our MoveGroup
+  /** \brief Parse the incoming joint msg for the joints of our MoveGroup */
   bool updateJoints(std::mutex& mutex, const JogArmShared& shared_variables);
 
+  /** \brief If incoming velocity commands are from a unitless joystick, scale them to physical units.
+   * Also, multiply by timestep to calculate a position change.
+   */
   Eigen::VectorXd scaleCartesianCommand(const geometry_msgs::TwistStamped& command) const;
 
+  /** \brief If incoming velocity commands are from a unitless joystick, scale them to physical units.
+   * Also, multiply by timestep to calculate a position change.
+   */
   Eigen::VectorXd scaleJointCommand(const control_msgs::JointJog& command) const;
 
   bool addJointIncrements(sensor_msgs::JointState& output, const Eigen::VectorXd& increments) const;
 
-  // Suddenly halt for a joint limit or other critical issue.
-  // Is handled differently for position vs. velocity control.
+  /** \brief Suddenly halt for a joint limit or other critical issue.
+   * Is handled differently for position vs. velocity control.
+   */
   void suddenHalt(trajectory_msgs::JointTrajectory& joint_traj);
   void suddenHalt(Eigen::ArrayXd& delta_theta);
 
+  /** \brief Publish the status of the jogger to a ROS topic */
   void publishStatus(StatusCode status) const;
 
-  // Scale the delta theta to match joint velocity/acceleration limits
+  /** \brief  Scale the delta theta to match joint velocity/acceleration limits */
   void enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta);
 
-  // Avoid overshooting joint limits
+  /** \brief Avoid overshooting joint limits */
   bool enforceSRDFPositionLimits(trajectory_msgs::JointTrajectory& new_joint_traj);
 
-  // Possibly calculate a velocity scaling factor, due to proximity of
-  // singularity and direction of motion
+  /** \brief Possibly calculate a velocity scaling factor, due to proximity of
+   * singularity and direction of motion
+   */
   double velocityScalingFactorForSingularity(const Eigen::VectorXd& commanded_velocity,
                                              const Eigen::JacobiSVD<Eigen::MatrixXd>& svd,
                                              const Eigen::MatrixXd& jacobian, const Eigen::MatrixXd& pseudo_inverse);
@@ -127,10 +138,13 @@ protected:
   bool applyVelocityScaling(const JogArmShared& shared_variables, std::mutex& mutex, Eigen::ArrayXd& delta_theta,
                             double singularity_scale);
 
+  /** \brief Compose the outgoing JointTrajectory message */
   trajectory_msgs::JointTrajectory composeJointTrajMessage(sensor_msgs::JointState& joint_state) const;
 
+  /** \brief Smooth position commands with a lowpass filter */
   void lowPassFilterPositions(sensor_msgs::JointState& joint_state);
 
+  /** \brief Convert an incremental position command to joint velocity message */
   void calculateJointVelocities(sensor_msgs::JointState& joint_state, const Eigen::ArrayXd& delta_theta);
 
   /** \brief Convert joint deltas to an outgoing JointTrajectory command.
@@ -138,6 +152,9 @@ protected:
     */
   bool convertDeltasToOutgoingCmd();
 
+    /** \brief Gazebo simulations have very strict message timestamp requirements.
+     * Satisfy Gazebo by stuffing multiple messages into one.
+     */
   void insertRedundantPointsIntoTrajectory(trajectory_msgs::JointTrajectory& trajectory, int count) const;
 
   /**
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -58,7 +58,7 @@ class JogCalcs
 public:
   JogCalcs(const JogArmParameters& parameters, const robot_model_loader::RobotModelLoaderPtr& model_loader_ptr);
 
-  void startMainLoop(JogArmShared& shared_variables, std::mutex& mutex);
+  void startMainLoop(JogArmShared& shared_variables);
 
   void stopMainLoop();
 
@@ -82,7 +82,7 @@ protected:
   std::atomic<bool> is_initialized_;
 
   /** \brief Do jogging calculations for Cartesian twist commands. */
-  bool cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, JogArmShared& shared_variables, std::mutex& mutex);
+  bool cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, JogArmShared& shared_variables);
 
   /** \brief Do jogging calculations for direct commands to a joint. */
   bool jointJogCalcs(const control_msgs::JointJog& cmd, JogArmShared& shared_variables);
@@ -91,7 +91,7 @@ protected:
   void updateCachedStatus(JogArmShared& shared_variables);
 
   /** \brief Parse the incoming joint msg for the joints of our MoveGroup */
-  bool updateJoints(std::mutex& mutex, const JogArmShared& shared_variables);
+  bool updateJoints(JogArmShared& shared_variables);
 
   /** \brief If incoming velocity commands are from a unitless joystick, scale them to physical units.
    * Also, multiply by timestep to calculate a position change.
@@ -130,13 +130,11 @@ protected:
   /**
    * Slow motion down if close to singularity or collision.
    * @param shared_variables data shared between threads, tells how close we are to collision
-   * @param mutex locks shared data
    * @param delta_theta motion command, used in calculating new_joint_tray
    * @param singularity_scale tells how close we are to a singularity
    * @return false if very close to collision or singularity
    */
-  bool applyVelocityScaling(const JogArmShared& shared_variables, std::mutex& mutex, Eigen::ArrayXd& delta_theta,
-                            double singularity_scale);
+  bool applyVelocityScaling(JogArmShared& shared_variables, Eigen::ArrayXd& delta_theta, double singularity_scale);
 
   /** \brief Compose the outgoing JointTrajectory message */
   trajectory_msgs::JointTrajectory composeJointTrajMessage(sensor_msgs::JointState& joint_state) const;
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -152,9 +152,9 @@ protected:
     */
   bool convertDeltasToOutgoingCmd();
 
-    /** \brief Gazebo simulations have very strict message timestamp requirements.
-     * Satisfy Gazebo by stuffing multiple messages into one.
-     */
+  /** \brief Gazebo simulations have very strict message timestamp requirements.
+   * Satisfy Gazebo by stuffing multiple messages into one.
+   */
   void insertRedundantPointsIntoTrajectory(trajectory_msgs::JointTrajectory& trajectory, int count) const;
 
   /**
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -38,12 +38,18 @@
 
 #pragma once
 
+// System
 #include <atomic>
-#include "jog_arm_data.h"
-#include "low_pass_filter.h"
+
+// ROS
 #include <moveit/planning_scene_monitor/planning_scene_monitor.h>
 #include <moveit/robot_model_loader/robot_model_loader.h>
-#include <std_msgs/Bool.h>
+#include <std_msgs/Int8.h>
+
+// moveit_jog_arm
+#include "jog_arm_data.h"
+#include "low_pass_filter.h"
+#include "status_codes.h"
 
 namespace moveit_jog_arm
 {
@@ -79,6 +85,9 @@ protected:
 
   bool jointJogCalcs(const control_msgs::JointJog& cmd, JogArmShared& shared_variables);
 
+  /** \brief Update the stashed status so it can be retrieved asynchronously */
+  void updateCachedStatus();
+
   // Parse the incoming joint msg for the joints of our MoveGroup
   bool updateJoints(std::mutex& mutex, const JogArmShared& shared_variables);
 
@@ -93,7 +102,7 @@ protected:
   void suddenHalt(trajectory_msgs::JointTrajectory& joint_traj);
   void suddenHalt(Eigen::ArrayXd& delta_theta);
 
-  void publishWarning(bool active) const;
+  void publishStatus(StatusCode status) const;
 
   // Scale the delta theta to match joint velocity/acceleration limits
   void enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta);
@@ -153,10 +162,11 @@ protected:
 
   std::vector<LowPassFilter> position_filters_;
 
-  ros::Publisher warning_pub_;
+  ros::Publisher status_pub_;
 
-  // Flag that a warning should be published
-  bool has_warning_ = false;
+  StatusCode status_ = kNoWarning;
+  // Cache the previous status, for retrieval at any time
+  StatusCode prev_status_ = kNoWarning;
 
   JogArmParameters parameters_;
 
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -112,7 +112,7 @@ protected:
   void suddenHalt(Eigen::ArrayXd& delta_theta);
 
   /** \brief Publish the status of the jogger to a ROS topic */
-  void publishStatus(StatusCode status) const;
+  void publishStatus() const;
 
   /** \brief  Scale the delta theta to match joint velocity/acceleration limits */
   void enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta);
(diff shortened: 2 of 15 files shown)


moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
```diff
@@ -47,7 +47,7 @@ joint_limit_margin: 0.1 # added as a buffer to joint limits [radians]. If moving
 cartesian_command_in_topic: jog_server/delta_jog_cmds  # Topic for incoming Cartesian twist commands
 joint_command_in_topic: jog_server/joint_delta_jog_cmds # Topic for incoming joint angle commands
 joint_topic:  joint_states
-warning_topic: jog_server/halted # Publish boolean warnings to this topic
+status_topic: jog_server/status # Publish string status messages to this topic
```
jliukkonen: The idea is to publish status messages in general but all the messages end up being warnings. I think the naming here is good but for example `WarningCodeEnum` could be just `StatusCode` to make it more expandable in the future. Maybe you want to add info messages or debug or error messages at some point?

moveit_experimental/moveit_jog_arm/test/test_halt_msg/test_jog_arm_halt_msg.py
```diff
@@ -18,8 +18,8 @@
 
 CARTESIAN_JOG_COMMAND_TOPIC = 'jog_server/delta_jog_cmds'
 
-# jog_arm should publish 'true' here if it halts
-HALT_TOPIC = 'jog_server/halted'
+# jog_arm should publish a string warning here if it halts
+HALT_TOPIC = 'jog_server/status'
```
jliukkonen: Should this be just `WARNING_TOPIC` or `STATUS_TOPIC` to be more general?

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -386,12 +391,9 @@ bool JogCalcs::convertDeltasToOutgoingCmd()
   if (!enforceSRDFJointBounds(outgoing_command_))
   {
     suddenHalt(outgoing_command_);
-    has_warning_ = true;
+    status_ = kJointBound;
```
jliukkonen: Wouldn't it be way better if we only publish when there's an issue (relatively rare occurrence) and just replace these status variable updates with publish commands? This would then read `publishStatus(WarningCodeEnum::kJointBound)`. My problem with setting the `status_` variable here and there is that eventually it is not updated correctly and you either miss a warning(s) or get wrong (outdated) warning.

moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
AndyZe: ok, renaming WarningCodeEnum->StatusCode sounds fine to me. Done

moveit_experimental/moveit_jog_arm/test/test_halt_msg/test_jog_arm_halt_msg.py
AndyZe: yeah, `STATUS_TOPIC` to match the config files makes sense

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: I actually prefer to publish once per iteration. If you only publish an error, the subscriber won't be sure how long ago the error occurred and if it's resolved.

I can delete `status_` and just call `publishStatus()` immediately. I think that makes sense.

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
```diff
@@ -94,7 +94,7 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
   error += !rosparam_shortcuts::get("", n, parameter_ns + "/planning_frame", ros_parameters_.planning_frame);
   error += !rosparam_shortcuts::get("", n, parameter_ns + "/use_gazebo", ros_parameters_.use_gazebo);
   error += !rosparam_shortcuts::get("", n, parameter_ns + "/check_collisions", ros_parameters_.check_collisions);
-  error += !rosparam_shortcuts::get("", n, parameter_ns + "/warning_topic", ros_parameters_.warning_topic);
+  error += !rosparam_shortcuts::get("", n, parameter_ns + "/status_topic", ros_parameters_.status_topic);
```
mlautman: Changes to rosparam names using rosparam_shortcuts can break applications that use this. You should use the node handle interface to check for "/status_topic" and if it exists use the value. If it doesn't exist then print a warning about the parameter name change and check for "/warning_topic" and use that if it is found. If it is not found, you can print an error and increment `error`.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -48,7 +48,7 @@ JogCalcs::JogCalcs(const JogArmParameters& parameters, const robot_model_loader:
   : parameters_(parameters), default_sleep_rate_(1000)
 {
   // Publish collision status
-  warning_pub_ = nh_.advertise<std_msgs::Bool>(parameters_.warning_topic, 1);
+  warning_pub_ = nh_.advertise<std_msgs::String>(parameters_.status_topic, 1);
```
tylerjw: I thought the intent was to replace this with a new message type that included both a int32 (enum) and a string.  That way software that interacts could recover from errors returned this way and also print the warning string for humans.

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
AndyZe: done

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: @tylerjw  Summarizing what we talked about yesterday: ideally there would be a ROS message type with an `int` field and a `string` field so I could publish both without making a separate PR to moveit_msgs. But there isn't, so... the new commits publish an int8 status code and provide a C++ interface for checking the status. A human-readable warning message is logged, too.

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
```diff
@@ -105,6 +104,23 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
...
+  {
+    error += !rosparam_shortcuts::get("", n, parameter_ns + "/warning_topic", ros_parameters_.status_topic);
+    ROS_WARN_NAMED(LOGNAME,
+                   "Topic 'warning_topic' was renamed to 'status_topic'. Please update the jogging yaml file.");
+  }
+  else
+  {
+    ROS_WARN_NAMED(LOGNAME,
+                   "Topic 'warning_topic' was renamed to 'status_topic'. Please update the jogging yaml file.");
+  }
```
tylerjw: I don't think you need to test if warning_topic is there, just print the log and call get, if neither are there you want an error.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/status_codes.h
```diff
@@ -0,0 +1,61 @@
...
+  kJointBound = 4
+};
+
+const std::unordered_map<uint, std::string>
+    kStatusCodeMap({ { kNoWarning, "No warnings" },
+                     { kSingularityDecelerate, "Close to a singularity, decelerating" },
+                     { kSingularityHalt, "Very close to a singularity, halting" },
+                     { kCollision, "Close to a collision, halting." },
+                     { kJointBound, "Close to a joint bound (position or velocity), halting" } });
+}  // end namespace trackjoint
```
tylerjw: I like this.  This is a nice interface.

AndyZe:
> You should make a cpp unit test for the cpp interface. It could be mostly a copy of the cpp interface example with a couple asserts.

That's going to be a big task, I think. It would be more of an integration test because it involves loading the robot.

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
AndyZe: That makes sense. Done

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
```diff
@@ -105,6 +104,15 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
...
 
+  // This parameter name was changed recently.
+  // Try retrieving from the correct name. If it fails, then try the deprecated name.
+  if (!rosparam_shortcuts::get("", n, parameter_ns + "/status_topic", ros_parameters_.status_topic))
+  {
+    ROS_WARN_NAMED(LOGNAME, "'status_topic' parameter is missing. Recently renamed from 'warning_topic'. Please update "
+                            "the jogging yaml file.");
+    error += !rosparam_shortcuts::get("", n, parameter_ns + "/warning_topic", ros_parameters_.status_topic);
+  }
+
```
tylerjw: :+1:

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/status_codes.h
```diff
@@ -0,0 +1,61 @@
...
+{
+  kNoWarning = 0,
+  kSingularityDecelerate = 1,
+  kSingularityHalt = 2,
+  kCollision = 3,
+  kJointBound = 4
+};
+
+const std::unordered_map<uint, std::string>
+    kStatusCodeMap({ { kNoWarning, "No warnings" },
```
mlautman: MoveIt Style would dictate that this should be named STATUS_CODE_MAP. 'Status code' is a little generic but since this is in the `moveit_jog_arm` namespace, I think it works. Nevertheless, I might consider naming it something like, `JOG_ARM_STATUS_CODE_MAP`

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -109,8 +110,14 @@ int main(int argc, char** argv)
 
   // Retrieve the current joint state from the jogger
   sensor_msgs::JointState current_joint_state = jog_interface.getJointState();
+  ROS_INFO_STREAM("Current joint state:");
```
mlautman: Always use `_NAMED` so that the user can selectively filter out msgs with the rosconsole

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -109,8 +110,14 @@ int main(int argc, char** argv)
...
   ROS_INFO_STREAM(current_joint_state);
 
+  // Retrieve the current status of the jogger
+  moveit_jog_arm::StatusCode status = jog_interface.getJoggerStatus();
+  ROS_INFO_STREAM("Jogger status:");
+  ROS_INFO_STREAM(status);
```
mlautman:
```suggestion
  ROS_INFO_STREAM_NAMED(LOGGER_NAME, "Jogger status:" << status);
```

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -526,13 +544,16 @@ double JogCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& comm
...
+      status_ = kSingularityDecelerate;
+      ROS_WARN_STREAM_THROTTLE_NAMED(2, LOGNAME, kStatusCodeMap.at(status_));
     }
 
     // Very close to singularity, so halt.
     else if (ini_condition > parameters_.hard_stop_singularity_threshold)
     {
       velocity_scale = 0;
-      ROS_WARN_THROTTLE_NAMED(2, LOGNAME, "Close to a singularity. Halting.");
+      status_ = kSingularityHalt;
```
mlautman: Is there a reason you are using the mutex for reading but not for writing the status?

moveit_experimental/moveit_jog_arm/src/jog_cpp_interface.cpp
```diff
@@ -221,4 +221,13 @@ bool JogCppInterface::getCommandFrameTransform(Eigen::Isometry3d& transform)
   // All zeros means the transform wasn't initialized, so return false
   return !transform.matrix().isZero(0);
 }
+
+StatusCode JogCppInterface::getJoggerStatus()
```
mlautman: Why not provide a private `setJoggerStatus` so you don't have to pass the mutex around?

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
```diff
@@ -105,6 +104,15 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
   error += !rosparam_shortcuts::get("", n, parameter_ns + "/publish_joint_accelerations",
                                     ros_parameters_.publish_joint_accelerations);
 
+  // This parameter name was changed recently.
+  // Try retrieving from the correct name. If it fails, then try the deprecated name.
```
mlautman: Add a TODO to replace this with rosparam_shortcuts::get in the noetic release.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -77,30 +83,44 @@ class JogCalcs
...
-  // Suddenly halt for a joint limit or other critical issue.
-  // Is handled differently for position vs. velocity control.
+  /** \brief Suddenly halt for a joint limit or other critical issue.
+   * Is handled differently for position vs. velocity control.
+   */
   void suddenHalt(trajectory_msgs::JointTrajectory& joint_traj);
 
-  void publishWarning(bool active) const;
+  /** \brief Publish the status of the jogger to a ROS topic */
+  void publishStatus(StatusCode status) const;
```
mlautman: Looks like this is always used with the class variable `status_`. Why not do this?
```suggestion
  void publishStatus() const;
```

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -346,14 +346,17 @@ bool JogCalcs::cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, JogArmShared&
   if (!applyVelocityScaling(shared_variables, mutex, delta_theta_,
                             velocityScalingFactorForSingularity(delta_x, svd, jacobian, pseudo_inverse)))
   {
-    has_warning_ = true;
     suddenHalt(outgoing_command_);
   }
 
+  publishStatus(status_);
```
mlautman: Why pass a local variable?

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -77,30 +83,44 @@ class JogCalcs
...
+  /** \brief Do jogging calculations for Cartesian twist commands. */
   bool cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, JogArmShared& shared_variables, std::mutex& mutex);
 
-  bool jointJogCalcs(const control_msgs::JointJog& cmd, JogArmShared& shared_variables);
+  /** \brief Do jogging calculations for direct commands to a joint. */
+  bool jointJogCalcs(const control_msgs::JointJog& cmd, JogArmShared& shared_variables, std::mutex& mutex);
 
-  // Parse the incoming joint msg for the joints of our MoveGroup
+  /** \brief Update the stashed status so it can be retrieved asynchronously */
+  void updateCachedStatus(JogArmShared& shared_variables, std::mutex& mutex);
```
mlautman: Why not put the mutex in the JogArmShared or maybe even use setters / getters for accessing the shared_variables that manage locking / unlocking the mutex internally?

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/status_codes.h
```diff
@@ -0,0 +1,61 @@
...
+*******************************************************************************/
+
+#pragma once
+
+#include <string>
+#include <unordered_map>
+
+namespace moveit_jog_arm
+{
+enum StatusCode : int8_t
```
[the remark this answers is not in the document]
AndyZe: I try to avoid moveit_msgs because it will take another week or two at least to get it merged. Then I would need to get this PR reviewed again, so that could easily add a month to review time.

To me, it seems like a std_msgs/Int8 is just fine as a message type.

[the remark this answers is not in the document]
AndyZe: will fix ALL_CAPS

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/status_codes.h
AndyZe: ok, JOG_ARM_STATUS_CODE_MAP it is

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
AndyZe: done

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
AndyZe: done

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: Looking through jog_calcs.cpp, status_ gets written to the shared variables here (L385):

```
  mutex.lock();
  shared_variables.status = status_;
  mutex.unlock();
```

Mutexes aren't used anywhere else. I think that makes sense.

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
AndyZe: done

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: I like to keep it explicit. OK, I'll change it to `publishStatus()`

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
AndyZe: done

moveit_experimental/moveit_jog_arm/src/jog_cpp_interface.cpp
AndyZe: I'm not sure what you mean, but I could make the status be a std::atomic<__int8> so I don't even need the mutex for this variable. How about that?

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
AndyZe: fixed by making the status code atomic

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -382,9 +385,19 @@ bool JogCalcs::jointJogCalcs(const control_msgs::JointJog& cmd, JogArmShared& /*
...
   prev_joint_velocity_ = delta_theta_ / parameters_.publish_period;
 
+  publishStatus();
+  // Cache the status so it can be retrieved asynchronously
+  updateCachedStatus(shared_variables, mutex);
+
   return convertDeltasToOutgoingCmd();
 }
 
+void JogCalcs::updateCachedStatus(JogArmShared& shared_variables, std::mutex& mutex)
```
henningkayser: Why are you still passing the mutex around? I don't see it being used anywhere. Please remove it from `jointJogCalcs()` and `updateCachedStatus()`. In general, I would really prefer having a member struct/class that provides thread-save access to all shared member variables using getters and setters. Of course, atomic members are fine as well if applicable. Having mutexes everywhere and passing them around makes this code harder to read and maintain than it has to be.

AndyZe: Sorry, I simply forgot to delete the mutex arguments. Good catch.

I also want to reduce mutex usage. Maybe in a follow-up PR?

henningkayser: Follow-up PR is fine! Thanks for fixing

henningkayser:
Seems like a rebase is missing, then lgtm

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/status_codes.h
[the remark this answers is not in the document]
mlautman: It's a shame that moveit_msgs wasn't merged into the moveit repo when the rest of the packages were merged. It seems like there is plenty of support for merging the two repos now. The only thing blocking it is that we need someone to update `https://github.com/davetcoleman/moveit_merge` to merge moveit and moveit_msgs while preserving the two repos' respective histories.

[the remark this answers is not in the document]
mlautman: > To me, it seems like a std_msgs/Int8 is just fine as a message type.

I don't know it sounds like a shortcut to me. At very least, I would add a TODO that this should be converted to a moveit_msg in the future. That way if you have a Python consumer, it doesn't have to duplicate this file just to make sense of the error codes. Also if you had a JogArmStatus msg, you could put both a std_msgs/Int8 and a std_msgs/String in there so that you could put the strings in the `JOG_ARM_STATUS_CODE_MAP` in the msg itself getting sent around.

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -109,7 +110,13 @@ int main(int argc, char** argv)
...
   // Retrieve the current joint state from the jogger
   sensor_msgs::JointState current_joint_state = jog_interface.getJointState();
-  ROS_INFO_STREAM(current_joint_state);
+  ROS_INFO_STREAM_NAMED(LOGNAME, "Current joint state:");
+  ROS_INFO_STREAM_NAMED(LOGNAME, current_joint_state);
+
+  // Retrieve the current status of the jogger
+  moveit_jog_arm::StatusCode status = jog_interface.getJoggerStatus();
+  ROS_INFO_STREAM_NAMED(LOGNAME, "Jogger status:");
+  ROS_INFO_STREAM_NAMED(LOGNAME, status);
```
mlautman: nit: 
```suggestion
  ROS_INFO_STREAM_NAMED(LOGNAME, "Jogger status:\n" << status);
```

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
mlautman: This doesn't seem to use the mutex..

moveit_experimental/moveit_jog_arm/src/jog_cpp_interface.cpp
mlautman: atomic sounds good. Is there any reason why you would only make status atomic or would it make more sense for the entire shared_variables_ to be an `atomic<JogArmShared>`?

The other option that I originally suggested was to add a `std::mutex lock;` to `JogArmShared` so that you didn't have to pass the mutex around through function interfaces right next to the JogArmShared object all over the place.

A third option, if you want to get fancy would be to take care of locking and unlocking the mutex within the JogArmShared object by using setters and getters for it's internal data. You would probably make it into a class and make all of it's members private and then use the setters and getters across the code base. (eg `shared_variables_.get_ok_to_publish()` instead of `shared_variables.ok_to_publish` and `shared_variables_.set_ok_to_publish(bool ok_to_publish`)

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/status_codes.h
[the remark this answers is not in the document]
AndyZe: A downside is that a new PR to moveit_msgs would be required if we ever wanted to change a status code

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: Thanks, good catch

AndyZe:
> Is there any reason why you would only make status atomic or would it make more sense for the entire shared_variables_ to be an atomic<JogArmShared>?

Apparently you can only do that if the type is "trivially copyable." I'll try it (and ask Tyler what he thinks).

> For example, we cannot use std::atomic<std::map<int, int>>, because std::map<> type is not trivially copyable for any arguments of its template.

https://www.codeproject.com/Articles/1183379/We-Make-Any-Object-Thread-safe

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
AndyZe: done

AndyZe:
@mlautman Let me know what you think about that most recent commit. Passing mutexes around is done, so thanks for that!

I still don't want to use moveit_msgs

Issue #1896, opened 2020-02-13
Disable codecov triggering red 'X' for pull-request CI status


rhaschke:
Dear @tylerjw, @davetcoleman,
while having the codecov report shown in github is great, can we relax the requirement to not decrease coverage with PRs? This creates lots of spurious "not-mergable" flags, I one always needs to inspect the details, why a particular PR is not mergable.
As we currently have extremely low coverage, I suggest to invert the messaging and provide extra positive feedback if the coverage is significantly increased, because a new test case was added.
Is that possible?

rhaschke:
Ping @tylerjw. We should discuss our strategy in the next maintainer meeting.
Could you, @tylerjw, figure out the technical possibilities until then?

felixvd:
Cross-referencing the [discourse post](https://discourse.ros.org/t/moveit-maintainer-meeting-feb-27th/12812/5?u=felixvd) for the meeting.

jschleicher:
See PR #1917

v4hn:
The remaining issue, as discussed at the last maintainer meeting is the red 'x' in every pull-request that fails codecov.

This is not wanted right now and makes the maintainers' job harder looking through all open requests.

jschleicher:
@v4hn I think, this was fixed in #1917, see current PR #2027 for a working example: coverage decreases and it is still green. Suggesting to close here.

PR #1917, opened 2020-02-27
codecov yaml

tylerjw:
### Description

This yaml file configures how codecov comments on pull requests.  It should now "succeed" even when pull requests reduce coverage by up to 0.5%

@rhaschke here is the documentation for this: https://docs.codecov.io/docs/commit-status

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,0 +1,13 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        # allow coverage to drop by this amount and still post success
+        threshold: 0.5%
+        base: auto
+        branches:
+          - master
+        if_not_found: success
+        if_ci_failed: error
+        only_pulls: false


codecov.yaml
```diff
@@ -0,0 +1,13 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        # allow coverage to drop by this amount and still post success
+        threshold: 2%
```
rhaschke: I think it suffices to limit the threshold to 0.5%, given the size of the code base.
Would it be possible to reduce the severity of a codecov failure (from red cross to yellow dot in github)?

tylerjw: I wasn't able to find an option for the yellow dot in the configuration.  From my reading of the documentation I am not sure if this is 2% of the content of the existing pr, 2% of the overall code coverage, or both.  If is of the overall coverage than I agree with you that I could be much smaller.

rhaschke: I guess it's the percentage reported by the codecov reports itself. What I have seen so far, are values like -0.39%. Maybe check some recent PRs to get a better impression of the actual range.

tylerjw: changed it to 0.5

rhaschke:
Fixes #1896.

PR #1927, opened 2020-03-04
allow to filter for joints when creating a RobotTrajectory message

rhaschke:
To execute MTC solutions from the rviz plugin, one needs a possibility to constrain RobotTrajectory messages to the actually moving joints. Usually, this is done via the associated JointModelGroup:
https://github.com/ros-planning/moveit/blob/ece0b7b44da2cfcf178e557557bc9d2693d17d5f/moveit_core/robot_trajectory/src/robot_trajectory.cpp#L226-L227
However, when deserializing received RobotTrajectory messages, I cannot infer the corresponding JointModelGroup anymore. Hence, this information is lost. But, remembering the involved joint names, it's possible to filter the RobotTrajectory for relevant joints during serialization into a message. This is what this PR proposes.

### DIFF
--- a/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
+++ b/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
@@ -213,7 +213,8 @@ public:
 
   double getAverageSegmentDuration() const;
 
-  void getRobotTrajectoryMsg(moveit_msgs::RobotTrajectory& trajectory) const;
+  void getRobotTrajectoryMsg(moveit_msgs::RobotTrajectory& trajectory,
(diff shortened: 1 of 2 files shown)


rhaschke:
@v4hn, you accidentally merged this single commit as a merge-commit.

v4hn:
True, sorry about that.
Why does Github have to remember the state of the merge button between different requests. *sigh*

rhaschke:
I fixed this by force-pushing to the master. Please update locally.

v4hn:
> I fixed this by force-pushing to the master.

-.- You just don't care about policies at all, do you?

I still consider a force-push to `*-devel` much much worse than a single merge commit in the history.

PR #1942, opened 2020-03-08
Provide feedback on success/failure of requestPlanningSceneState()

rhaschke:
Addition to #1934, providing corresponding feedback.

### DIFF
--- a/moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
+++ b/moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
@@ -533,6 +533,7 @@ void PlanningSceneDisplay::loadRobotModel()
   model_is_loading_ = false;
 }
 
+// This should always run in the main GUI thread!
 void PlanningSceneDisplay::onRobotModelLoaded()
 {
   changedPlanningSceneTopic();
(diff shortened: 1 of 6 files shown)


moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
```diff
@@ -370,8 +370,13 @@ void PlanningSceneDisplay::changedPlanningSceneTopic()
     std::string service_name = planning_scene_monitor::PlanningSceneMonitor::DEFAULT_PLANNING_SCENE_SERVICE;
     if (!getMoveGroupNS().empty())
       service_name = ros::names::append(getMoveGroupNS(), service_name);
-    auto bg_func = [=]() { planning_scene_monitor_->requestPlanningSceneState(service_name); };
-    addBackgroundJob(bg_func, "planning_scene_monitor_->requestPlanningSceneState");
+    auto bg_func = [=]() {
+      if (planning_scene_monitor_->requestPlanningSceneState(service_name))
+        addMainLoopJob(boost::bind(&PlanningSceneDisplay::onNewPlanningSceneState, this));
+      else
+        setStatus(rviz::StatusProperty::Warn, "PlanningScene", "Requesting initial scene failed");
```
v4hn: Does it make sense to retry this later?
I could imagine a roslaunch take more than 5 seconds to bring up a `move_group`, while RViz is up much earlier.
This might happen, e.g., if moveit has to wait for hardware controllers.

rhaschke: How often / long do you want to retry? I wouldn't make this too complicated. In case, you can still reset the display.

PR #1946, opened 2020-03-09
jog_arm tests: Fix test timing. Preallocate the received variable

### LINKED ISSUE #1943: flaky unittest in jog_arm
Hi @AndyZe, there is again (still?) a flaky unit test in jog_arm:
https://travis-ci.org/ros-planning/moveit/jobs/659981308?utm_medium=notification&utm_source=github_status
Could you take a look, please?

AndyZe:
Hopefully closes #1943


AndyZe:
> Please replace sleeps with `rospy.wait_for_service(service, timeout=None)` where appropriate.

Would you suggest waiting for that service even if I don't plan on using it? In other words, would it always be a good way to check if the jogger is done initializing?  (I think so)

rhaschke:
Depends on how often you expect the service to be used. If it's more often *not* used than used, only check the availability of the service just before calling. Otherwise, check it on initialization.
In any case you can provide a timeout to not wait forever. On failure you should print an error or warning message, depending on the severity of the problem.

AndyZe:
OK, this looks ready for me.

Since `wait_for_jogger_initialization()` is used several places, it would be nice to import it. Right now, I can't figure out how to do that.

moveit_experimental/moveit_jog_arm/test/test_drift_dimensions_service/test_drift_dimensions_service.py
```diff
@@ -14,23 +14,36 @@
...
 @pytest.fixture
 def node():
     return rospy.init_node('pytest', anonymous=True)
 
 
+def wait_for_jogger_initialization(service_name):
+    try:
+      rospy.wait_for_service(service_name, timeout=15)
+    except rospy.ServiceException as exc:
+      rospy.logerr("The jogger never finished initialization, expected service is not available: " + str(exc))
```
rhaschke: You shouldn't use "never", when you actually only wait for 15s :wink:

AndyZe:
> ```python
> from ..util import wait_for_jogger_initialization
> ```

It was a little more complicated than that ;)  but I figured out a way. I think the main issue was the Python code running as scripts, not modules.

I have to warn you, I won't have tons of time to work on this in the next few days.

AndyZe:
Thanks Robert, your changes look fine

PR #1950, opened 2020-03-11
PSI: don't wait forever for services

rhaschke:
Possible fix for regression described in https://github.com/ros-planning/moveit/pull/1611#issuecomment-597617904.

### DIFF
--- a/moveit_ros/planning_interface/planning_scene_interface/include/moveit/planning_scene_interface/planning_scene_interface.h
+++ b/moveit_ros/planning_interface/planning_scene_interface/include/moveit/planning_scene_interface/planning_scene_interface.h
@@ -52,8 +52,10 @@ MOVEIT_CLASS_FORWARD(PlanningSceneInterface);
 class PlanningSceneInterface
 {
 public:
-  explicit PlanningSceneInterface(const std::string& ns = "");
+  explicit PlanningSceneInterface(const std::string& ns = "", const ros::WallDuration& timeout = ros::WallDuration());
   ~PlanningSceneInterface();
+
+  bool waitForServices(const ros::WallDuration& timeout = ros::WallDuration());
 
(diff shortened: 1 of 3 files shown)


v4hn:
Continuing after https://github.com/ros-planning/moveit/pull/1611#issuecomment-597649907.

> From my point of view, the only option is to provide an optional timeout parameter to the PSI constructor

I do not consider this an option at all.
The issues with timeouts are always:
- what you do after the timeout is up
  - if you would WARN and retry, this is what the code currently does
- How to pick a reasonable timeout
  - is blocking RViz for 1 second better than having the standard use-case of 5-10 seconds to get a standard move_group running?
  - If you want to be able to specify an immediate timeout, this justifies another way of construction instead of a continuous parameter.

> The other two options, I don't really like

You misunderstood my proposal. Of course I would combine these options.
For a very similar effect we could introduce a construction that does not wait at all, but fails to produce a valid instance if the services do not exist. A boolean constructor parameter and a throw or a factory method would do the trick.

> Usually (for non-persistent connections), the service is contacted for each service call, which means that it can handle a crash and restart of the service. So why wait at all in the constructor? Shouldn't we have a timeout for each call instead?

Waiting in the constructor instead of waiting in each method implies an invariant for the services to be available. It is true that this invariant does not hold when the service provider dies later on.
I do not see the need to wait for that invariant in every method call though.
The API design of the class suffers from the inability to discriminate between "nothing" and "failure" already. Adding "reached timeout" to "failure" just makes it worse.

The point of waiting in the constructor in the first place is to ensure the created object satisfies the invariant *initially*. Otherwise the whole data structure might be initialized, but is unusable without ever reporting it.
It will just report empty results the whole time.

v4hn:
Superseded by #1951

PR #1951, opened 2020-03-12
Resolve PSI lock-up in RViz display

v4hn:
Supersedes #1950 .

I implemented an alternative way to construct a PSI, enabling users to detect issues with the ROS services instead of locking up the thread waiting for them.
It does not harm to have this option around.

Looking at the RViz display code again, I decided against making use of this feature here though.
The display already maintains a PSM, so I rewrote the logic to use this instead and  altogether got rid of the PSI.

### DIFF
--- a/moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp
+++ b/moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp
@@ -50,7 +50,7 @@ static const std::string LOGNAME = "planning_scene_interface";
 class PlanningSceneInterface::PlanningSceneInterfaceImpl
 {
 public:
-  explicit PlanningSceneInterfaceImpl(const std::string& ns = "")
+  explicit PlanningSceneInterfaceImpl(const std::string& ns = "", bool wait = true)
   {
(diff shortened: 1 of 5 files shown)


moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp
```diff
@@ -59,8 +59,18 @@ class PlanningSceneInterface::PlanningSceneInterfaceImpl
...
+    if (wait)
+    {
+      waitForService(planning_scene_service_);
+      waitForService(apply_planning_scene_service_);
+    }
+    else
+    {
+      if (!planning_scene_service_.exists() || !apply_planning_scene_service_.exists())
+      {
+        throw std::runtime_error("ROS Services not available");
```
rhaschke:
```suggestion
        throw std::runtime_error("ROS services not available");
```

rhaschke:
Do you want to merge this as a merge commit, keeping the two rather independent commits?

v4hn:
> Do you want to merge this as a merge commit, keeping the two rather independent commits?

That was the idea, yes.

Still waiting for CI though after clang-tidy was still complaining.

PR #1952, opened 2020-03-13
Do not override empty URDF link collision geometry

### LINKED ISSUE #130: Moveit Treats Visual Mesh as Collision Mesh When No Collision Mesh is Specified

### LINKED ISSUE #1907: rviz collision is inaccurate when robot visual is used to override collision

Dale-Koenig:
~~Partially addressing #130.  Add a **warning** so that more drastic action can be taken when this issue is revisited in the future, and to ensure that the user is aware that the collision geometry has been altered.~~

Fixes #130 and fixes #1907.  Changes to no longer override collision geometry.


v4hn:
Thank you for the patch. [The maintainer meeting actually agreed to a more drastic patch in the master branch](https://discourse.ros.org/t/moveit-maintainer-meeting-feb-27th/12812/7), but no one acted on it yet.

Could you change your patch to keep the warning, but *not add the collision geometry*?

The reasoning was as follows:
- If we want to change the standard, we should do this in the `master` branch directly. This is what the branch is for.
- If we add the warning and anyone really *relies* on the old behavior, they will just have to copy their geometry to the collision geometry too
- If people want the new behavior, they will have to ignore the warning for some transition period.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -999,21 +1000,32 @@ LinkModel* RobotModel::constructLinkModel(const urdf::Link* urdf_link)
...
+    }
+  }
+
+  if (has_missing_collision)
+  {
+    // clang-format off
+    ROS_WARN_STREAM_NAMED(LOGNAME, "Link " << urdf_link->name << ": "
+                                   "has visual geometry but no collision geometry. "
+                                   "Collision geometry will be left empty. "
+                                   "Fix your URDF file by explicitly specifying collision geometry.");
```
v4hn: I would actually propose to set this warning to a different logger name, so that affected users can set this logger to ERROR if they want to ignore the specific warning.
How about `LOGNAME+".empty_collision_geometry"`?

```suggestion
    ROS_WARN_STREAM_NAMED(LOGNAME+".empty_collision_geometry", "Link " << urdf_link->name << ": "
                                   "has visual geometry but no collision geometry. "
                                   "Collision geometry will be left empty. "
                                   "If you require collision checking for the link, explicitly set the collision geometry in the URDF."
                                   "Otherwise, consider disabling this logger name via rosconsole.");
```

PR #1953, opened 2020-03-14
KDL IK: constrain wiggled joints to limits

rhaschke:
In #1321 I added random wiggling of joints when the optimization is stuck in a local minimum.
However, I didn't clip those changes to the feasible joint range.

### DIFF
--- a/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
+++ b/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
@@ -428,8 +428,8 @@ int KDLKinematicsPlugin::CartToJnt(KDL::ChainIkSolverVelMimicSVD& ik_solver, con
   KDL::Frame f;
   KDL::Twist delta_twist;
   KDL::JntArray delta_q(q_out.rows()), q_backup(q_out.rows());
-  Eigen::ArrayXd extra_joint_weights;
-  extra_joint_weights.setOnes(joint_weights.rows());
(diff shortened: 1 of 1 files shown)


moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp
```diff
@@ -478,12 +478,14 @@ int KDLKinematicsPlugin::CartToJnt(KDL::ChainIkSolverVelMimicSVD& ik_solver, con
...
     {
-      if (step_size < 0.005)  // cannot reach target
+      if (step_size < epsilon_)  // cannot reach target
         break;
       // wiggle joints
       last_delta_twist_norm = DBL_MAX;
       delta_q.data.setRandom();
       delta_q.data *= std::min(0.1, delta_twist_norm);
+      clipToJointLimits(q_out, delta_q, extra_joint_weights);
+      extra_joint_weights.setOnes();
```
v4hn: Is it justified to throw away the values of `extra_joint_weights`, computed in the call to `clipToJointLimits` above?
Intuitively, I would have expected a separate vector to pass here.

rhaschke: When we wiggle the joints, we essentially start over:
- `last_delta_twist_norm` is reset
- as well as `extra_joint_weights`

v4hn: makes sense, I did not notice the norm reset.

PR #1955, opened 2020-03-16
mgi: add execution methods for moveit_msgs::RobotTrajectory

v4hn:
I just reviewed https://github.com/ros-planning/moveit_tutorials/pull/468 and
thought this would be a nicer way to execute Cartesian trajectories...


v4hn:
Actually, if you want to unify them, I would propose to simply deprecate the old execute method.
The method actually only ever looks at the `trajectory_` field...
I'll implement the changes and provide a deprecated wrapper for the old method.

v4hn:
After thinking about it a bit more, I decided to leave the old interfaces around without deprecation.
They are convenient wrappers when working with the `plan` method and in use all over the place.
It feels stupid to deprecate a one-line convenient wrapper in a class meant to make access convenient.

felixvd:
It makes a lot of sense for the move group to execute a trajectory instead of a plan. I like this.

PR #1964, opened 2020-03-22
Improve performance by calling Eigen::Transform::linear() instead of Eigen::Transform::rotation().

peci1:
### Description

As shown in https://github.com/ros-planning/geometric_shapes/issues/125#issuecomment-602134891, Eigen doesn't yet contain the optimizations that would save computing SVD in case `rotation()` is called on an isometry. However, calling `linear()` gives the same result for isometries and doesn't need to compute SVD. In geometric_shapes, this change cut computation times tenfold for the affected functions.

This PR is more like an opening to a discussion. I don't expect it to be merged as is. Maybe it should be split per package, or there may be other concerns.


peci1:
Fixed. I haven't thought a simple sed with a shorter string could generate formatting errors, but well, one line got shorter so much that a newline could be deleted from the middle of it :)

From my point of view, there are no reasons for not merging this. It just seemed as quite a big change to me (meaning spanning a lot of packages). 

And it will (silently) reveal all the places where somebody uses an invalid rotation, like I did e.g. here: https://github.com/ros-planning/geometric_shapes/pull/126/commits/88734502bdac40c46af4d49cb95ac69576904762 . The `computeRotationScaling()` did actually cope with most kinds of invalid rotation by putting all the wrong numbers into the scaling matrix which is then thrown away. The updated behavior will pass the invalid rotation further completely silently, and the only visible change would be that some code will start acting weird. Passing an invalid transform is of course a problem of the user of the API, but well, some people might feel sad about it.

v4hn:
> It just seemed as quite a big change to me (meaning spanning a lot of packages).

That's no problem.

> I don't know if moveit test coverage is good enough to be sure this didn't actually break anything (which it should not, but...).

No, our test coverage does not suffice.
Feel free to contribute (even) more tests.

> And it will (silently) reveal all the places where somebody uses an invalid rotation. [...]
> The updated behavior will pass the invalid rotation further completely silently, and the only visible change would be that some code will start acting weird

This has to be pointed out to the user and we have to add an item to `MIGRATION.md` and we should consider merging in `master` only. [Eigen does not validate Isometries at all](https://stackoverflow.com/questions/46176533/how-to-ensure-eigen-isometry-stays-isometric), because it is rather expensive to do.
This change in behavior, garbage-in-garbage-out over silent normalization, affects `computeCartesianPath`, as well as user code maintaining their own `CollisionWorld`/adding objects outside the regular `Quaternion` representation, which is/should be normalized on our end via the planning scene.

Notice that users will have to live with the change either way in the long run, as it will automatically take effect with Eigen's next release.

Your test-case shows how easy it is to accidentally break an Isometry, e.g., via `Eigen::AngleAxis3d(phi, Vector3d(1.0,1.0,1.0))` and retracting a normalization guarantee is no lightweight change. It frustrates me that we apparently did not do that as part of the `Isometry3d` transition for melodic.

The same applies to [`geometric_shapes`](https://github.com/ros-planning/geometric_shapes/pull/126), of course.

Personally, I would merge these patches in `melodic-devel`, still, but that is mostly because we keep pushing more and more things in the `master` branch and `noetic` and future release progress is very much uncertain.
Looking for opinions from other maintainers, @gavanderhoorn @rhaschke @jrgnicho @davetcoleman .

gavanderhoorn:
I would be OK with breaking things for others, but as always, breaking things means you'll get unhappy customers.

Customers in an OSS setting are potential contributors.

If this is done in `master` and mentioned in the migration guide, this could be acceptable though.

Breaking melodic is not a good idea. It's going to make a lot of people unhappy, and we would be violating our (implicit) contract with them. `master` is a free-for-all. Until we cut a release from it.

rhaschke:
I agree with @gavanderhoorn. I just noticed that this PR is targeting melodic, not master.
@peci1, could you please rebase onto master, please, and then target this PR against master?

simonschmeisser:
Most of these changes seem not to be user facing? Maybe we can exclude all places where user-generated data is used and then backport the speed gains to melodic. Especially the AABB calculation should be quite beneficial!

peci1:
I retargeted to master.

@simonschmeisser I thought exactly the same. But as I don't see into moveit that much, I probably couldn't do that work myself.

As CI is down now, I ran the tests locally and all are green.

peci1:
A wrap-up from WG meeting: I'll add asserts to the code which will check the isometries. These asserts will be compiled out in release builds. This change should be mentioned in Migration.md and should have a thread on discourse. Having all this done, it should be considered for backporting to melodic-devel.

We also discussed issuing a `ROS_WARN_ONCE()` or something like that with a message to the user that if his code starts behaving weird, he should run in debug mode and see if any of the asserts fails for him and where do the wrong data originate. But I'm not sure there was a broad agreement on this warning message. And it is unclear what's the best place to issue this warning.

Also, tutorials should be checked and notes should be added to places where beginners (or even pros) might create a non-isometry when not being careful.

rhaschke:

gavanderhoorn:
If what we're seeing with Travis is indeed caused by the app vs other integration, than the branch protection rules must be updated to require the new Travis PR builds. I believe `continuous-integration/travis-ci` is the old one, which isn't run any more, so the status will never be updated.

(I've updated 100+ repositories over at `ros-industrial` and `ros-industrial-consortium`)

rhaschke:
@peci1, could you please rebase onto latest master, such that I can merge this?

peci1:
Rebased.

I just wonder whether it is now more efficient to write

    transform.linear().inverse()

or

    transform.inverse().linear()

If I'm not mistaken, after calling `linear()`, the fact that the transform is an isometry and `linear()` is just pure rotation gets lost... So I'd bet the subsequent `inverse` will do the costly full inverse.

On the other hand, `transform.inverse()` is optimized for isometries to actually use the transposition on its linear part. But it creates an unnecessary 4x4 matrix instead of the needed 3x3...

Of course, the most efficient is probably

    transform.linear().transpose()

but that clouds a bit the intention of inverting the matrix. Of course, people with experience will probably see inversion and transposition as equivalent, but still, the semantics gets a little clouded...

Anyways, the only use I found was in https://github.com/ros-planning/moveit/blob/ad88e18ae604e0533456c89eb0c6694bb0b63081/moveit_core/kinematic_constraints/src/utils.cpp#L570 . So maybe it's not that important at all... It just caught my interest.

v4hn:
On Mon, Apr 27, 2020 at 04:25:00PM -0700, Martin Pecka wrote:
>     transform.linear().transpose()
> 
> but that clouds a bit the intention of inverting the matrix.

I clearly prefer this version, and I don't think it's confusing...
If you like, add an appropriate comment.

peci1:
Okay, I changed utils.cpp to use `linear().transpose()`

v4hn:
Personally, I would prefer to have the additional Isometry check in debug mode for the general MoveIt code base too. I guess we could reuse the assert definition in `geometric_shapes`, although we should probably put it in a new header there for that.

Any other opinions?

v4hn:
@peci1 could you please add the Isometry checks to get this merged? It's been dangling for almost two months now...

peci1:
Okay, I'll have a look at it. So what's the preferred way? Should I create a new header file containing the definition of the isometry check and copy the code from geometric_shapes?

v4hn:
> Should I create a new header file containing the definition of the isometry check and copy the code from geometric_shapes?

I would go with a new header in `moveit_core/macros`.

Alternatively factorize the macro in geometric_shapes in a new header there and reuse it here.

Both approaches seem fine to me.

peci1:
I added the isometry checks. They revealed two tests with "homemade" quaternions that did not result in proper rotations. That's all.

Now Travis complains about warnings that macro `ASSERT_ISOMETRY()` is being redefined. So what's the proper way to proceed with reusing the macro from geometric_shapes? Should I first issue a PR to geometric_shapes `master` branch that would expose this macro, and once it is merged, I'd reuse it here?

v4hn:
> Now Travis complains about warnings that macro ASSERT_ISOMETRY() is being redefined.

True. You could add `MOVEIT_*`, but reusing is probably better as we already depend on `geometric_shapes`.

> So what's the proper way to proceed with reusing the macro from geometric_shapes? Should I first issue a PR to geometric_shapes master branch that would expose this macro, and once it is merged, I'd reuse it here?

yes, please. You're talking about `noetic-devel` though.

peci1:
@rhaschke okay, but what are user inputs? Are data read from TF user inputs or not? I think this needs to be better specified.

@v4hn Would it make sense then to create some kind of annotation (at least in function documentation) that would tell whether the returned isometry can be trusted? I was even thinking about creating a new type like `UnsafeIsometry3d` which would be used for user inputs, but that seems like an overkill to me. On the other hand, it would make the API crystal clear.

v4hn:
> but what are user inputs?

As a general rule of thumb, everything is.
After all, MoveIt exports almost everything as public API.
That's the crux of the matter.

> Are data read from TF user inputs or not? I think this needs to be better specified.

Great question. I would argue they can be treated as sanitized because other systems, prominently RViz, but possibly every TF listener should warn if something invalid is published there.

> Would it make sense then to create some kind of annotation (at least in function documentation) that would tell whether the returned isometry can be trusted?

No. Eigen::Isometry3d *is* an annotation already, even if Eigen fails to verify the invariant.
If anything we could wrap the class and add check, but that seems overkill too.

peci1:
I've just pushed a new version which moves the check more towards the places where the values are set. I also updated documentation so that it mentions if a transform can be trusted to be an isometry. There would probably be much more places which would need this update, but these are all that have something to do with calling `linear()` instead of `rotation()` (at least in moveit code).

I tried running tests locally in debug mode, but some of them take enormous time to complete (I tried the same with current master and the behavior is more or less the same, so I assume this isn't a problem introduced in this PR):

```
-- run_tests.py: execute commands
  /media/data/ros/devel-debug/lib/moveit_ros_robot_interaction/locked_robot_state_test --gtest_output=xml:/media/data/ros/build-debug/test_results/moveit_ros_robot_interaction/gtest-locked_robot_state_test.xml
[==========] Running 16 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 16 tests from LockedRobotState
[ RUN      ] LockedRobotState.load
[ INFO] [1590159631.586877101] [ros.moveit_core.robot_model]: Loading robot model 'one_robot'...
[       OK ] LockedRobotState.load (12 ms)
[ RUN      ] LockedRobotState.URDF_sanity
[       OK ] LockedRobotState.URDF_sanity (0 ms)
[ RUN      ] LockedRobotState.robotStateChanged
[       OK ] LockedRobotState.robotStateChanged (1 ms)
[ RUN      ] LockedRobotState.set1
[       OK ] LockedRobotState.set1 (8792 ms)
[ RUN      ] LockedRobotState.set2
[       OK ] LockedRobotState.set2 (15115 ms)
[ RUN      ] LockedRobotState.set3
[       OK ] LockedRobotState.set3 (24244 ms)
[ RUN      ] LockedRobotState.mod1
[       OK ] LockedRobotState.mod1 (282483 ms)
[ RUN      ] LockedRobotState.mod2
[       OK ] LockedRobotState.mod2 (132392 ms)
[ RUN      ] LockedRobotState.mod3
[       OK ] LockedRobotState.mod3 (13359 ms)
[ RUN      ] LockedRobotState.set1mod1
[       OK ] LockedRobotState.set1mod1 (92560 ms)
[ RUN      ] LockedRobotState.set2mod1
[       OK ] LockedRobotState.set2mod1 (128344 ms)
[ RUN      ] LockedRobotState.set1mod2
[       OK ] LockedRobotState.set1mod2 (222076 ms)
[ RUN      ] LockedRobotState.set3mod1
[       OK ] LockedRobotState.set3mod1 (183312 ms)
[ RUN      ] LockedRobotState.set1mod3
[       OK ] LockedRobotState.set1mod3 (399067 ms)
[ RUN      ] LockedRobotState.set3mod3
[       OK ] LockedRobotState.set3mod3 (575074 ms)
[ RUN      ] LockedRobotState.set3mod3c3
[       OK ] LockedRobotState.set3mod3c3 (559277 ms)
[----------] 16 tests from LockedRobotState (2636109 ms total)

[----------] Global test environment tear-down
[==========] 16 tests from 1 test case ran. (2636112 ms total)
[  PASSED  ] 16 tests.
```

Yes, 2636 second to do 16 tests. This is the worst one, though. In release mode, this whole test suite takes about 15 seconds.

I also tried debug-compiled moveit with release-compiled geometric_shapes, but the speedup was negligible.

simonschmeisser:
You could temporarily replace the `#ifdef NDEBUG` by something different and run the tests in release mode. The slowdown is most likely because Eigen does "a few more tests" in debug mode.

peci1:
@simonschmeisser good idea. When built with `-O3` and without `NDEBUG`, the tests run reasonably fast. I just wonder nobody has noticed that debug-build tests are sooo slow. I also tested several past versions (back to 1.0.3), and all of them have the tests so slow, so it is nothing newly introduced.

v4hn:
On Sat, May 23, 2020 at 02:24:51AM -0700, Martin Pecka wrote:
> I just wonder nobody has noticed that debug-build tests are sooo slow.

CI runs tests without debug.
Locally, I notice this issue a lot, but it was not important enough to me to work on it.
There are a number of tests with unreasonably long runtime, in many cases by excessive test iterations, as in the case you cite.

I would appreciate efforts to limit test repetitions to a sane amount.

peci1:
Thank you too, @v4hn and @rhaschke for getting this done ;)

rhaschke:
Thanks goes to you, @peci1, for all your work and patience!

PR #1965, opened 2020-03-22
[WIP] SINGLE distance request collision test

v4hn:
CI fails for this test.

Related to #1963 .


v4hn:
@j-petit any idea why this new test fails for the bullet plugin?

PR #1967, opened 2020-03-23
MP display: change internal shortcut Ctrl+R to Ctrl+I

v4hn:
Ctrl+R is used by RViz to rename displays.
Overlaying the sequence semi-breaks the behavior
and the rename button gets selected, but not pressed.

This is an internal shortcut used in case the interactive markers
of the display fail to update correctly. This is seldomly (never?)
used, so remapping does not hurt anyone.

To the best of my knowledge Ctrl+I is not used in standard RViz.


v4hn:
>  Maybe we should even remove it completely? If I need to reset the display, I usually reset it completely.

It did not harm up to now, so no reason to drop it I guess.
It's not like RViz is running low on shortcuts...

PR #1977, opened 2020-03-26
[jog_arm] Add a binary collision check to the main control loop

AndyZe:
Previously there was distance-based collision checking in a separate thread, but some issues with it were noticed. Namely, starting inside another body did not stop motion.

This is kind of a collision check of last resort. Binary collision checking should be fast enough for the main control loop.

I can probably send video of this working if requested, but it's a private project, so I can't share the video publicly.

### DIFF
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/collision_check_thread.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/collision_check_thread.h
@@ -41,7 +41,6 @@
 #include <atomic>
 #include "jog_arm_data.h"
 #include "low_pass_filter.h"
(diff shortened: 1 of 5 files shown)


v4hn:
I do believe you managed to get this to work on your system, but this very much defeats the purpose of the project and breaks separation of concerns in terms of threads.
Binary collision checking is *not* fast. There are no guarantees on the number of meshes/vertices involved in the computation.
This looks like the sort of change you would do to get your project actually running, but that shouldn't go upstream.

> Namely, starting inside another body did not stop motion.

Could you create a thread for this issue?
This sounds like you can easily add individual binary checks on startup, before?

AndyZe:
OK. I understand, and shared your concern

PR #1978, opened 2020-03-26
[jog_arm] Add a binary collision check

AndyZe:
Add a binary collision check to the distance-based collision check. This helps with edge cases like starting inside another body and reducing overall flakiness.

Also, delete the parameter _hard_stop_singularity_threshold_ which hasn't been used in some time.

### DIFF
--- a/moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
+++ b/moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
@@ -52,5 +52,5 @@ command_out_topic: joint_group_position_controller/command # Publish outgoing co
 
 ## Collision checking for the entire robot body
 check_collisions: true # Check collisions?
-collision_check_rate: 5 # [Hz] Collision-checking can easily bog down a CPU if done too often.
+collision_check_rate: 10 # [Hz] Collision-checking can easily bog down a CPU if done too often.
 collision_proximity_threshold: 0.01 # Start decelerating when a collision is this far [m]
--- a/moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
+++ b/moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
@@ -54,4 +54,3 @@ command_out_topic: joint_group_position_controller/command # Publish outgoing co
 check_collisions: true # Check collisions?
 collision_check_rate: 5 # [Hz] Collision-checking can easily bog down a CPU if done too often.
 collision_proximity_threshold: 0.01 # Start decelerating when a collision is this far [m]
-hard_stop_collision_proximity_threshold: 0.0005 # Stop when a collision is this far [m]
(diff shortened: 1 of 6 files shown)


moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -73,6 +72,14 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
   double velocity_scale_coefficient = -log(0.001) / parameters_.collision_proximity_threshold;
   ros::Rate collision_rate(parameters_.collision_check_rate);
 
+  double collision_distance = 0;
+  bool is_in_collision = false;
```
henningkayser: Why add this extra variable?

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/status_codes.h
```diff
@@ -48,14 +48,16 @@ enum StatusCode : int8_t
...
+  JOINT_BOUND = 5
 };
 
 const std::unordered_map<uint, std::string>
     JOG_ARM_STATUS_CODE_MAP({ { NO_WARNING, "No warnings" },
                               { DECELERATE_FOR_SINGULARITY, "Close to a singularity, decelerating" },
                               { HALT_FOR_SINGULARITY, "Very close to a singularity, halting" },
-                              { COLLISION, "Close to a collision, halting." },
+                              { DECELERATE_FOR_COLLISION, "Close to a collision, decelerating." },
+                              { HALT_FOR_COLLISION, "Close to a collision, halting." },
```
henningkayser: Shouldn't this be "Collision detected, emergency stop" or something similar?

moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
```diff
@@ -54,4 +54,3 @@ command_out_topic: joint_group_position_controller/command # Publish outgoing co
 check_collisions: true # Check collisions?
 collision_check_rate: 5 # [Hz] Collision-checking can easily bog down a CPU if done too often.
```
henningkayser: This rate should probably be increased. Maybe it would be wise to add a warning to the thread initialization in case this rate is very low.

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
AndyZe: good point, will fix

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/status_codes.h
AndyZe: OK

moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
AndyZe: OK, will fix

PR #1979, opened 2020-03-26
small fixes to https://github.com/ros-planning/moveit/pull/1757

rhaschke:
As https://github.com/ros-planning/moveit/pull/1757 was opened on an organizational repository,
I was not able to push directly.
I fixed the existing test. The key was to adapt the `ResolveConstraintFrames` adapter to handle the now global frames returned from `getFrameInfo()`.
Will continue to look at the original tutorial now as this needs an update as well.


rhaschke:
Need to open this against PickNik's repo...

PR #1982, opened 2020-03-26
[jog_arm] Fix low-pass filter initialization

henningkayser:
I ran into an issue where JogArm would initialize the low-pass position filters before processing the initial jog commands. The issue was that there was a separate "waiting loop" that initialized the filter  joint states which have never been updated. That caused an initial jump if the robot has moved between thread initialization and the first commands. Since the waiting loop and the main loop have a very similar structure, I thought it would be cleaner to just use a bool flag "wait_for_initial_commands" that leads to the same logic as before, just that the joint states are being updated, fixing the described issue. While debugging this issue I touched and refactored several other parts, including:

* Removing a duplicate joint increment with cartesian commands 4f47786
* Simplified the joint validation logic by using the active joints only instead of all variable joints
* Added a pause/unpause flag for jog threads

Please don't squash this PR

### DIFF
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -60,14 +60,6 @@ public:
 
   void startMainLoop(JogArmShared& shared_variables);
 
-  void stopMainLoop();
-
-  /** \brief Stop publishing jog commands while keeping the main loop spinning and filters up to date */
-  void pauseOutgoingJogCmds();
-
-  /** \brief Continue publishing jog commands */
-  void unpauseOutgoingJogCmds();
-
   /** \brief Check if the robot state is being updated and the end effector transform is known
    *  @return true if initialized properly
    */
@@ -75,12 +67,6 @@ public:
 
 protected:
   ros::NodeHandle nh_;
-
-  // Loop termination flag
-  std::atomic<bool> stop_jog_loop_requested_;
-
-  // Flag that outgoing commands to the robot should not be published
-  std::atomic<bool> pause_outgoing_jog_cmds_;
 
   // Flag that robot state is up to date, end effector transform is known
   std::atomic<bool> is_initialized_;
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -62,7 +62,11 @@ public:
 
   void stopMainLoop();
 
-  void haltOutgoingJogCmds();
+  /** \brief Stop publishing jog commands while keeping the main loop spinning and filters up to date */
+  void pauseOutgoingJogCmds();
+
+  /** \brief Continue publishing jog commands */
+  void unpauseOutgoingJogCmds();
 
   /** \brief Check if the robot state is being updated and the end effector transform is known
    *  @return true if initialized properly
@@ -76,7 +80,7 @@ protected:
   std::atomic<bool> stop_jog_loop_requested_;
 
   // Flag that outgoing commands to the robot should not be published
-  std::atomic<bool> halt_outgoing_jog_cmds_;
+  std::atomic<bool> pause_outgoing_jog_cmds_;
 
   // Flag that robot state is up to date, end effector transform is known
   std::atomic<bool> is_initialized_;
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_cpp_interface.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_cpp_interface.h
@@ -59,11 +59,8 @@ public:
 
   void stopMainLoop();
 
-  /** \brief Pause processing jog commands while keeping the main loop alive */
-  void pause();
-
-  /** \brief Continue processing jog commands */
-  void unpause();
+  /** \brief Pause or unpause processing jog commands while keeping the main loop alive */
+  void setPaused(bool paused);
 
   /** \brief Provide a Cartesian velocity command to the jogger.
    * The units are determined by settings in the yaml file.
@@ -99,8 +96,5 @@ public:
 
 private:
   ros::NodeHandle nh_;
-
-  std::atomic<bool> stop_requested_;
-  std::atomic<bool> paused_;
 };
 }  // namespace moveit_jog_arm
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_cpp_interface.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_cpp_interface.h
@@ -59,6 +59,12 @@ public:
 
   void stopMainLoop();
 
+  /** \brief Pause processing jog commands while keeping the main loop alive */
+  void pause();
+
+  /** \brief Continue processing jog commands */
+  void unpause();
+
   /** \brief Provide a Cartesian velocity command to the jogger.
    * The units are determined by settings in the yaml file.
    */
@@ -95,5 +101,6 @@ private:
   ros::NodeHandle nh_;
 
   std::atomic<bool> stop_requested_;
+  std::atomic<bool> paused_;
 };
 }  // namespace moveit_jog_arm
--- a/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
@@ -326,9 +326,6 @@ bool JogCalcs::cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, JogArmShared&
 
   enforceSRDFAccelVelLimits(delta_theta_);
 
-  if (!addJointIncrements(internal_joint_state_, delta_theta_))
-    return false;
-
   // If close to a collision or a singularity, decelerate
   applyVelocityScaling(shared_variables, delta_theta_,
                        velocityScalingFactorForSingularity(delta_x, svd, jacobian, pseudo_inverse));
--- a/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
@@ -107,49 +107,20 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables)
 
   is_initialized_ = true;
 
-  // Wait for the first jogging cmd.
-  // Store it in a class member for further calcs, then free up the shared variable again.
-  geometry_msgs::TwistStamped cartesian_deltas;
-  control_msgs::JointJog joint_deltas;
-  while (ros::ok() && (cartesian_deltas.header.stamp == ros::Time(0.)) && (joint_deltas.header.stamp == ros::Time(0.)))
-  {
-    if (stop_jog_loop_requested_)
-      return;
-
-    default_sleep_rate_.sleep();
-
-    // Ensure the low-pass filter matches reality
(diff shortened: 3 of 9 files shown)


jliukkonen:
Did not cause regressions when running on UR5. Seems to be working nicely.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -273,6 +254,11 @@ void JogCalcs::haltOutgoingJogCmds()
   halt_outgoing_jog_cmds_ = true;
 }
 
+void JogCalcs::proceedOutgoingJogCmds()
+{
+  halt_outgoing_jog_cmds_ = true;
+}
```
AndyZe: About this function name again -- it seems backward. `halt_outgoing_jog_cmds_ == true` means the jogger will halt. But the name of this function implies it will do the reverse

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -62,8 +62,12 @@ class JogCalcs
 
   void stopMainLoop();
 
+  /** \brief Stop publishing jog commands while keeping the main loop spinning and filters up to date */
   void haltOutgoingJogCmds();
 
+  /** \brief Continue publishing jog commands */
+  void proceedOutgoingJogCmds();
```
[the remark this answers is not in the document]
AndyZe: I think `unpauseOutgoingJogCmds()` would be a good name, like you used in jog_cpp_interface

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -62,8 +62,12 @@ class JogCalcs
 
   void stopMainLoop();
 
+  /** \brief Stop publishing jog commands while keeping the main loop spinning and filters up to date */
   void haltOutgoingJogCmds();
```
AndyZe: To match with my comment below, a better name would be `pauseOutgoingJogCmds()`

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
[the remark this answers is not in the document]
henningkayser: Fixed

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
henningkayser: Good catch! Of course this should be `false`

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
henningkayser: done

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -345,9 +331,6 @@ bool JogCalcs::cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, JogArmShared&
 
   enforceSRDFAccelVelLimits(delta_theta_);
 
-  if (!addJointIncrements(internal_joint_state_, delta_theta_))
```
henningkayser: @AndyZe Do you agree that this was redundant? I removed it because it was called again in `convertDeltasToOutgoingCmd()` in the end of this function.

AndyZe: Yeah, that seems like a good change

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -577,61 +560,70 @@ double JogCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& comm
 void JogCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
 {
   Eigen::ArrayXd velocity = delta_theta / parameters_.publish_period;
+  const Eigen::ArrayXd acceleration = (velocity - prev_joint_velocity_) / parameters_.publish_period;
```
AndyZe: This should be named ACCELERATION per ROS style guide

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -577,61 +560,70 @@ double JogCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& comm
...
-      // --> delta_theta = (accel * delta_t _ + vel_prev) * delta_t
-      Eigen::ArrayXd acceleration = (velocity - prev_joint_velocity_) / parameters_.publish_period;
-      if ((bounds.min_acceleration_ != 0) && (acceleration(joint_delta_index) < bounds.min_acceleration_))
+      if (clip_acceleration)
       {
-        double relative_change =
-            ((bounds.min_acceleration_ * parameters_.publish_period + prev_joint_velocity_(joint_delta_index)) *
+        // accel = (vel - vel_prev) / delta_t = ((delta_theta / delta_t) - vel_prev) / delta_t
+        // --> delta_theta = (accel * delta_t _ + vel_prev) * delta_t
+        const double relative_change =
```
AndyZe: RELATIVE_CHANGE

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -577,61 +560,70 @@ double JogCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& comm
...
 
-      velocity = delta_theta / parameters_.publish_period;
       // Apply velocity bounds
-      // delta_theta = joint_velocity * delta_t
-      if ((bounds.min_velocity_ != 0) && (velocity(joint_delta_index) < bounds.min_velocity_))
+      if (clip_velocity)
       {
-        double relative_change = (bounds.min_velocity_ * parameters_.publish_period) / delta_theta(joint_delta_index);
+        // delta_theta = joint_velocity * delta_t
+        const double relative_change = (velocity_limit * parameters_.publish_period) / delta_theta(joint_delta_index);
```
AndyZe: RELATIVE_CHANGE

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -577,61 +560,70 @@ double JogCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& comm
...
   Eigen::ArrayXd velocity = delta_theta / parameters_.publish_period;
+  const Eigen::ArrayXd acceleration = (velocity - prev_joint_velocity_) / parameters_.publish_period;
 
   std::size_t joint_delta_index = 0;
-  for (auto joint : joint_model_group_->getJointModels())
+  for (auto joint : joint_model_group_->getActiveJointModels())
   {
     // Some joints do not have bounds defined
-    if (kinematic_state_->getJointModel(joint->getName())->hasVariable(joint->getName()))
+    const auto bounds = joint->getVariableBounds(joint->getName());
```
AndyZe: BOUNDS

AndyZe:
I'll make a state machine diagram to make this logic easier to follow sometime soon

henningkayser:
> Approved assuming you'll make these small naming changes

done

> I'll make a state machine diagram to make this logic easier to follow sometime soon

That would be really nice!

henningkayser:
> I'll make a state machine diagram to make this logic easier to follow sometime soon

Reverted because of clang-tidy

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -170,9 +141,19 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables)
...
 
-    // If paused, just keep the low-pass filters up to date with current joints so a jump doesn't occur when
-    // restarting
-    if (halt_outgoing_jog_cmds_)
+    //  Check for an initial jog command
+    if (wait_for_jog_commands)
+    {
+      shared_variables.lock();
+      // Check if there are any commands with valid timestamp
+      wait_for_jog_commands = shared_variables.command_deltas.header.stamp == ros::Time(0.) &&
```
mlautman: Is it possible to miss this initial jog command?

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -268,9 +249,14 @@ void JogCalcs::stopMainLoop()
   stop_jog_loop_requested_ = true;
 }
 
-void JogCalcs::haltOutgoingJogCmds()
+void JogCalcs::pauseOutgoingJogCmds()
+{
+  pause_outgoing_jog_cmds_ = true;
+}
```
mlautman: nit, combine methods:
```suggestion
void JogCalcs::pauseOutgoingJogCmds(bool pause)
{
  pause_outgoing_jog_cmds_ = pause;
}
```

moveit_experimental/moveit_jog_arm/src/jog_cpp_interface.cpp
```diff
@@ -153,6 +157,22 @@ void JogCppInterface::stopMainLoop()
   stop_requested_ = true;
 }
 
+void JogCppInterface::pause()
+{
+  if (jog_calcs_)
+    jog_calcs_->pauseOutgoingJogCmds();
+
+  paused_ = true;
+}
```
mlautman: nit, combine the two:
```suggestion
void JogCppInterface::pause(bool pause)
{
  if (jog_calcs_)
    jog_calcs_->pauseOutgoingJogCmds(pause);

  paused_ = pause;
}
```

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_cpp_interface.h
```diff
@@ -95,5 +101,6 @@ class JogCppInterface : public JogInterfaceBase
   ros::NodeHandle nh_;
 
   std::atomic<bool> stop_requested_;
+  std::atomic<bool> paused_;
```
mlautman: This should probably not exist if the information already exists in `jog_calcs_`. Instead I would add a get method to jog_calcs.h for getting `pause_outgoing_jog_cmds_`. That way you never need to worry about keeping things in sync

henningkayser: I added these (stop_requested and paused) to the shared_variables so that all threads use the same atomic bools

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
henningkayser: Makes sense, I removed these (using shared variables now) and combined the pause/unpause functions in a single function `setPaused()`

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
henningkayser: I'm not sure I understand what you mean by this. Either a command has been received before this check or not. If not, we didn't really miss it because it was received after this check. In that case nothing happens and it will be picked up in the next loop cycle. Theoretically you would only miss commands if you publish them in a much higher rate than the jog control rate, but this seems unrelated to your question.

henningkayser:
@mlautman I addressed your comments and rebased onto master

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
mlautman: Cool, just wanted to make sure you considered this

PR #1983, opened 2020-03-26
[planning_interface] Unit Test for ByteString class

gleichdick:
### Description

I just found an unit test for the ByteString class in my workspace, dunno whether it makes sense to include it upstream...

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/test/test_serialize_msg_python.cpp
+++ b/moveit_ros/planning_interface/test/test_serialize_msg_python.cpp
@@ -1,0 +1,388 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, RWTH Aachen University.
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of RWTH Aachen University nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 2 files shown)


moveit_ros/planning_interface/test/test_serialize_msg_python.cpp
```diff
@@ -0,0 +1,371 @@
...
+    if (!py_data)
+      return false;
+    Py_ssize_t size = PyBytes_GET_SIZE(obj);
+    if (size != data.size())
+      return false;
+    return std::memcmp(py_data, &data[0], size) == 0;
+  }
+  bool compare_bytes(const ByteString& s)
+  {
+    const std::vector<unsigned char> testdata{ 0xff, 0xef, 0x00, 0x10 };
```
rhaschke: What is actually test here? An arbitrary? Bytestring against a fixed uchar vector?

gleichdick:
These Testcases where made to ensure that the message passing between C++ and Python didn't break. That's why I had to embed the Python Interpreter and to exchange data with it.
I added some comments to give a hint about the test cases, but I don't mind not merging this PR tbh.

moveit_ros/planning_interface/test/test_serialize_msg_python.cpp
gleichdick: this method is exposed to Python and called with `b'\\xff\\xef\\x00\\x10'` as argument from within Python. Then the `bytes` Python instance will be converted to C++ `py_bindings_tools::ByteString` by boost, the converted Object is then tested.
So we know that embedded NULL's won't be dropped.

gleichdick:
Rebased onto master

felixvd:
If you restart Travis it should build now.

rhaschke:

rhaschke (approved):
Just noticed that this PR is still pending. What a shame!
Generally, I approve. However, designing the test a C++ based test only, makes it rather complicated. Splitting it into a C++ part (providing `ByteStringTestHelper`) and a python part (actually performing the tests) would, I think, considerably simplify the code.
If you agree, @gleichdick, and are (still) willing to work on this, I really would appreciate that.
If not, I'm fine to merge this as is. I will wait another few days for your response.

gleichdick:
Thank you for your comments, I will try to do my best. I'm likely to be
quarantined because of Covid-19, so I'll have plenty of time 😅

Robert Haschke <notifications@github.com> schrieb am Fr., 9. Okt. 2020,
10:45:

rhaschke:
Great thanks. I hope you are fine regarding Covid-19!

gleichdick:
Thanks. First Covid-Test was negative :relaxed: 
Splitting the test in Python and C++ components was a very good hint, thank you very much. #2362 should do it.

Issue #1954, opened 2020-03-16
Make Pointcloud Octomap Updater more general


Dale-Koenig:
**Is your feature request related to a problem? Please describe.**
Opening this to foster initial discussion on this idea and collect more suggestions.

Pointcloud octomap updater at the moment seems to be appropriate only for a very specific use case.  That is, where

- The camera location changes
- It is necessary to maintain a permanent updated map of the environment
- Speed/high resolution is not particularly important, and it is NOT necessary to ensure a most recent octomap

It is my opinion that this (especially the last) makes it pretty much unusable for serious applications in its current state.

**Describe the solution you'd like**
I think to make it useful, it should be refactored such that:

- Slower operations (ray tracing, filtering the robot) can be disabled with parameters in the sensor yaml file when they are not needed.  This can change the processing time by an order of magnitude (I was frequently seeing 0.5s with ray tracing, 0.05s without)
- When disallowing ray tracing, the octomap can be generated from scratch from each pointcloud.  I understand that this can almost certainly be done without using MoveIt at all, but it would be convenient to have all the functionality in one place.
- Refactored such that an inner component can be used to synchronously process the pointcloud and add it to a planning scene/psm when desired.  Currently, especially with the large processing time from ray tracing, processing can finish later than expected resulting in plans using an old octomap. Bonus if this could be used without needing the locks/mutexes.
- Perhaps improved interfaces for filtering by shapes, intended for filtering objects to pick rather than just parts of the robot.

These are features that I currently have in my hacked version, but would take a decent amount of work still to integrate back into the upstream version.

v4hn:
> The camera location changes

Isn't this the more general use-case? I believe you might refer to a required `waitForTransform` call?
If the camera is properly registered as a static transform in your tf2 system w.r.t. your  planning frame, it should not wait at all.

> It is necessary to maintain a permanent updated map of the environment

Not entirely true, that's why there is a service to clear the octomap.
In some setups, I previously implemented a selective forwarding of point cloud messages, so that the octomap is only updated when the external system relies on it.

> Speed/high resolution is not particularly important, and it is NOT necessary to ensure a most recent octomap

That is indeed a severe bottleneck in the current system. There is probably potential to improve computation time, e.g., utilizing [GPU-Voxels](https://www.gpu-voxels.org/).
I'm not aware of anyone working along these lines at the moment though.

[...]

PR #1984, opened 2020-03-27
WIP: Improve pointcloud octomaps

Dale-Koenig:
### Description

See #1954 for discussion.  Adds a service to force PointcloudOctomapUpdater to update, and allows remaking the tree with each update, skipping ray tracing.  Requires a corresponding change to `moveit_msgs` to add the service.  Additionally refactors to allow easily maintaining ones own OcTree and attaching PointcloudOctomapUpdaters to it.

setup_assistant_change


Dale-Koenig:
@v4hn @rhaschke I'm leaving WIP in the title to make sure this is not merged before the `.rosinstall` change is reverted, but it is ready for review.

v4hn:
I added @henningkayser as a reviewer. He is eager to look into this and get it merged to use it in one of his own projects. :)

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -348,16 +358,85 @@ void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::PointCloud2::
...
+  std::lock_guard<std::mutex> lock(update_mutex_);
+  if (max_update_rate_ > 0)
+  {
+    // ensure we are not updating the octomap representation too often
+    if (ros::Time::now() - last_update_time_ <= ros::Duration(1.0 / max_update_rate_))
+      return;
+    last_update_time_ = ros::Time::now();
+  }
+
+  processCloud(cloud_msg);
```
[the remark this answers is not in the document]
Dale-Koenig: I'm not completely sure which is better.  That said, afaik with most common causes of failure, failing once means you will fail (almost) every time, so I'm not sure it matters that much.  At the moment the easiest causes of failure are:
- Having weird pointcloud timestamps (for example from a rosbag) or malformed pointcloud messages

- Having objects in the planning scene that do not have poses published in tf when using the full perception pipeline seems to cause problems.  I think this may just be a bug, but it can cause processing the cloud to fail every time.

moveit.rosinstall
```diff
@@ -2,8 +2,8 @@
 # Used with wstool, users can download source of all packages of MoveIt.
 - git:
     local-name: moveit_msgs
-    uri: https://github.com/ros-planning/moveit_msgs.git
-    version: master
+    uri: https://github.com/Dale-Koenig/moveit_msgs.git
+    version: ef4ffa00c48604c02421889fa50c721dcced359b
```
mlautman, line 6: I just merged this. You should be good to revert and remove WIP

moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
```diff
@@ -64,8 +66,29 @@ class PointCloudOctomapUpdater : public OccupancyMapUpdater
   void start() override;
   void stop() override;
   ShapeHandle excludeShape(const shapes::ShapeConstPtr& shape) override;
+
+  /** @brief Exclude a shape from the octomap when processing.  Allows specifying an initial transform to put in the
+     transform cache.
+             For use outside of the perception pipeline
+      @param shape The shape to exclude
+      @param pose Pose of the shape in the sensor frame */
+  ShapeHandle excludeShape(const shapes::ShapeConstPtr& shape, const Eigen::Isometry3d& pose);
```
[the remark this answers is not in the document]
Dale-Koenig: I think I just didn't consider this because `Eigen::Isometry3d::Identity()` does not really make sense as a default value - the default value for the `ShapeHandle excludeShape(const shapes::ShapeConstPtr& shape) override;` signature is actually "uninitialized because it will be set in the updateTransformCache callback", and there are very few situations where we would actually want to use the identity pose as the transform.  We also only want to use the version taking an Isometry3d when we are not using the perception pipeline.  So they are basically two separate APIs, even if the contents of the functions happen to be nearly the same.

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -348,16 +358,85 @@ void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::PointCloud2::
...
+    return false;
+  }
+
+  return processCloud(cloud_msg, sensor_origin_eigen, update_method_);
 }
+
+void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::PointCloud2::ConstPtr& cloud_msg)
+{
+  ROS_DEBUG_NAMED(LOGNAME, "Received a new point cloud message");
+  std::lock_guard<std::mutex> lock(update_mutex_);
```
henningkayser, line 420: Wouldn't it make sense to only lock `processCloud()` to allow thread-safe usage of the C++ API?

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -348,16 +358,85 @@ void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::PointCloud2::
...
+bool PointCloudOctomapUpdater::processCloud(const sensor_msgs::PointCloud2::ConstPtr& cloud_msg)
+{
+  if (monitor_->getMapFrame().empty())
+    monitor_->setMapFrame(cloud_msg->header.frame_id);
+
+  /* get transform for cloud into map frame */
+  Eigen::Isometry3d sensor_origin_eigen;
+
+  if (monitor_->getMapFrame() == cloud_msg->header.frame_id)
+    sensor_origin_eigen = Eigen::Isometry3d::Identity();
```
henningkayser, line 387:
```suggestion
    sensor_origin_eigen.setIdentity();
```

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -297,39 +296,50 @@ void PointCloudOctomapUpdater::cloudMsgCallback(const sensor_msgs::PointCloud2::
...
-  /* cells that overlap with the model are not occupied */
-  for (const octomap::OcTreeKey& model_cell : model_cells)
-    occupied_cells.erase(model_cell);
-
-  /* occupied cells are not free */
-  for (const octomap::OcTreeKey& occupied_cell : occupied_cells)
-    free_cells.erase(occupied_cell);
+  if (update_method == UpdateMethod::INCREMENTAL)
+  {
+    /* occupied cells are not free */
```
henningkayser, line 332: Doesn't this fit in the block above where `occupied_cells` is already being iterated?

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -182,7 +182,7 @@ void OccupancyMapMonitor::initialize()
     }
   }
   else
-    ROS_ERROR_NAMED(LOGNAME, "Failed to find 3D sensor plugin parameters for octomap generation");
+    ROS_WARN_NAMED(LOGNAME, "Failed to find 3D sensor plugin parameters for octomap generation");
```
henningkayser, line 185: Why shouldn't this be an error anymore?

moveit_ros/occupancy_map_monitor/include/moveit/occupancy_map_monitor/occupancy_map_updater.h
```diff
@@ -66,6 +66,13 @@ class OccupancyMapUpdater
   /** \brief This is the first function to be called after construction */
   void setMonitor(OccupancyMapMonitor* monitor);
 
+  /** \brief Bypass the default monitor by setting the linked occupancy map directly.  Updates must be processed
+   * manually*/
+  void setTree(const OccMapTreePtr& tree)
+  {
+    tree_ = tree;
```
henningkayser, line 73: Maybe consider wrapping this in the same update mutex, this could have strange effects when called during an update.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
Dale-Koenig: I had noticed it being output in many of the `moveit_ros` unit tests, so I am worried it is about a bit more than expected in situations where people have some skeleton of the sensor file from the setup assistant but aren't actually planning to use the perception pipeline.  If this isn't the case at all, I can leave it as an error.

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
Dale-Koenig: I believe I separated it out because this loop is only modifying local variables so it can be performed without locking the OcTree

Dale-Koenig: Oh I think I see what you mean.  This loop existed separately in the original so I just brought it over.  You mean this loop 
```     
for (const octomap::OcTreeKey& occupied_cell : occupied_cells)
    if (tree_->computeRayKeys(sensor_origin, tree_->keyToCoord(occupied_cell), key_ray_))
        free_cells.insert(key_ray_.begin(), key_ray_.end());
```
right?  It's possible that these could be combined.  I think combining them might have unexpected consequences if there are somehow multiple occupied cells on the same ray, whereas by performing this separately we guarantee none of `occupied_cells` accidentally get marked free.  But I don't know if that's something we really need to worry about or not.

henningkayser: I see, makes sense!

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
henningkayser: It just seemed unrelated, your explanation checks out. Also, the error is not handled in any way, so I'm fine with a warning.

Dale-Koenig:
(Still want to retest this a bit again since making updates)

v4hn:
> (Still want to retest this a bit again since making updates)

Please tell us once you are ready to see this merged.

Dale-Koenig:
> (Still want to retest this a bit again since making updates)
> Please tell us once you are ready to see this merged.

Should be fine now.  I just wanted to rerun using the old interface to check that it wasn't broken.  The other thing I was worried about it now addressed in #2104

moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
```diff
@@ -64,8 +72,29 @@ class PointCloudOctomapUpdater : public OccupancyMapUpdater
   void start() override;
   void stop() override;
   ShapeHandle excludeShape(const shapes::ShapeConstPtr& shape) override;
+
+  /** @brief Exclude a shape from the octomap when processing.  Allows specifying an initial transform to put in the
+     transform cache.
```
rhaschke, line 77:
```suggestion
  /** @brief Exclude a shape from the octomap when processing.
      Allows specifying an initial transform to put in the transform cache.
```

moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
```diff
@@ -64,8 +72,29 @@ class PointCloudOctomapUpdater : public OccupancyMapUpdater
...
+             For use outside of the perception pipeline
+      @param shape The shape to exclude
+      @param pose Pose of the shape in the sensor frame */
+  ShapeHandle excludeShape(const shapes::ShapeConstPtr& shape, const Eigen::Isometry3d& pose);
+
   void forgetShape(ShapeHandle handle) override;
 
+  /** @brief process a pointcloud message and update the octomap.  Does not update the transform cache.
+             Can be used outside of the perception pipeline
```
rhaschke, line 86:
```suggestion
  /** @brief Process a pointcloud message and update the octomap.
      Does not update the transform cache. Can be used outside of the perception pipeline.
```

moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
```diff
@@ -64,8 +72,29 @@ class PointCloudOctomapUpdater : public OccupancyMapUpdater
...
+      @param cloud_msg The pointcloud to process
+      @param sensor_pose Eigen pose of the frame in which the pointcloud is given
+      @param update_method Whether to update the current octomap probabilities or to forget all previous octomap data */
+  bool processCloud(const sensor_msgs::PointCloud2::ConstPtr& cloud_msg, const Eigen::Isometry3d& sensor_pose,
+                    UpdateMethod update_method);
+
+  /** @brief process a pointcloud message and update the octomap.  Updates transform cache.
+             Requires setTransformCacheCallback to have been called
```
rhaschke, line 94: The difference between these two methods doesn't become clear yet. Please augment the description.
```suggestion
  /** @brief Process a pointcloud message and update the octomap.
      Updates transform cache. Requires setTransformCacheCallback to have been called.
```

moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
```diff
@@ -47,11 +47,19 @@ DIAGNOSTIC_POP
...
 #include <moveit/occupancy_map_monitor/occupancy_map_updater.h>
 #include <moveit/point_containment_filter/shape_mask.h>
+#include <moveit_msgs/UpdatePointcloudOctomap.h>
 
 #include <memory>
+#include <mutex>
 
 namespace occupancy_map_monitor
 {
+enum class UpdateMethod
```
rhaschke, line 57: Please add a better description of the two update modes. What's the difference between them?

moveit_ros/perception/pointcloud_octomap_updater/src/pointcloud_octomap_updater.cpp
```diff
@@ -37,7 +37,7 @@
 #include <cmath>
 #include <moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h>
 #include <moveit/occupancy_map_monitor/occupancy_map_monitor.h>
-#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+#include <tf2_eigen/tf2_eigen.h>
```
rhaschke, line 40: I think this change makes the dependency on `tf2_geometry_msgs` obsolete. Please remove in package.xml.

Dale-Koenig: It seems there are still dependencies in other subfolders of the package, such as the depthmap updater.

aseligmann:
I am in the process of building a MoveIt! system around a robot, and I am using the sensor plugin `occupancy_map_monitor/PointCloudOctomapUpdater` for defining real-world collision objects based on pointcloud data.
I currently see no way of updating or clearing the pointcloud or its associated occupancy tree once it has been published on the topic that the sensor plugin is listening on.

As I understand it, this pull-request would enable such functionality. What is the current status for merging this PR?

Alternatively, is there a workaround to updating/clearing the Octomap pointcloud or occupancy tree from a user-defined node?

v4hn:
> What is the current status for merging this PR?

Some maintainer needs to review and test the patch and sadly nobody had the time to do so.
If you want to support this patch, please try it out and comment here.

Dale-Koenig:
> I am in the process of building a MoveIt! system around a robot, and I am using the sensor plugin `occupancy_map_monitor/PointCloudOctomapUpdater` for defining real-world collision objects based on pointcloud data.
> I currently see no way of updating or clearing the pointcloud or its associated occupancy tree once it has been published on the topic that the sensor plugin is listening on.
> 
> As I understand it, this pull-request would enable such functionality. What is the current status for merging this PR?
> 
> Alternatively, is there a workaround to updating/clearing the Octomap pointcloud or occupancy tree from a user-defined node?

I believe there's already a clear octomap movegroup capability you could look into

JStech:
I was just trying to test this, but it's failing to build with a bunch of warnings about `moveit_resources`: 
```
WARNING: package "moveit_planners_ompl" should not depend on metapackage "moveit_resources" but on its packages instead
```
(It says warning, but then it abandons the build.) Maybe a rebase is needed? Or is there a workaround for this?

rhaschke:
> I was just trying to test this, but it's failing to build with a bunch of warnings about `moveit_resources`

We recently modularized `moveit_resources`. So this just needs a rebase onto latest `master`.

JStech:
Thanks for the rebase. I'd like to see this merged--I think it'll be helpful in a project I'm working on. I built and ran it, and it seems to work as intended.

PR #1985, opened 2020-03-27
remove unused angles.h includes

rhaschke:
angles.h was included at various places, but wasn't used.


rhaschke:
Argh! @davetcoleman, can you please have another look, why Travis and github don't communicate with each other?

PR #1986, opened 2020-03-27
[jog_arm] Remove duplicate collision check

henningkayser:
Removed the second collision check since it's redundant right after the one with distance enabled.

### DIFF
--- a/moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
@@ -67,6 +67,7 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
   // Init collision request
   collision_detection::CollisionRequest collision_request;
   collision_request.group_name = parameters_.move_group_name;
+  collision_request.distance = true;  // enable distance-based collision checking
   collision_detection::CollisionResult collision_result;
 
   // Copy the planning scene's version of current state into new memory
@@ -74,8 +75,6 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
 
   double velocity_scale_coefficient = -log(0.001) / parameters_.collision_proximity_threshold;
   ros::Rate collision_rate(parameters_.collision_check_rate);
-
-  double collision_distance = 0;
 
   // Scale robot velocity according to collision proximity and user-defined thresholds.
   // I scaled exponentially (cubic power) so velocity drops off quickly after the threshold.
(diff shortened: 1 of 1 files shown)


v4hn:
What's going on at your project? :confused:
@henningkayser [you merged that second check](https://github.com/ros-planning/moveit/pull/1978/files#diff-1fe6258c35c0a7605e86f0d03324e9d1L90) yourself yesterday.

The main repository is not meant to see changes like that on a daily basis.
That's what pull-requests (and the option for maintainers to modify them) are for!

henningkayser:
@v4hn Well, we are actively using JogArm a lot right now, potentially not only for one project, and IMO we have added some pretty significant improvements for this package. If we found that many fixes and improvements for other components, we would definitely add them as well right now. I simply missed that duplicate check in the diff, be assured that I still tested it on multiple setups.

> The main repository is not meant to see changes like that on a daily basis.
> That's what pull-requests (and the option for maintainers to modify them) are for!

I'm not sure what you mean by this, we are adding frequent changes and improvements and we are reviewing every single PR + testing it on multiple setups (I'm not sure every maintainer is doing this). Are you suggesting we should split up into a devel or feature branch that is being synced from time to time? Maybe we should have a discussion what it means to work on an experimental package in a master branch.

AndyZe:
Well the [official policy](https://moveit.ros.org/documentation/contributing/pullrequests/) right now is 

> For small or trivial fixes, a single review is sufficient. Additionally, for the moveit_experimental package, a single review is sufficient.

Which I think is cool. Originally I was sad that jog_arm was quarantined to the experimental folder, but now I see the benefit.

I'm neutral about this PR. It seems like the binary collision check is more reliable than distance-based.

henningkayser:
> I'm neutral about this PR. It seems like the binary collision check is more reliable than distance-based.

The distance-based collision check includes the binary one, so a second check additional to the distance-based isn't required.

AndyZe:
I see. OK, looks good to me then.

v4hn:
I was not complaining about the one-review policy, especially as the author is code owner and a core contributor. That's pretty much two reviews already. :+1:

I was complaining that *the core* change from yesterday is undone in a pull request by the reviewer one day after merge. Testing on multiple configurations is awesome (do you test on hardware?), but so is a logical analysis of the patch. ;)

> The distance-based collision check includes the binary one, so a second check additional to the distance-based isn't required.

Could your issues come from *unsigned* distance requests? [Signed distances (in case of collisions are disabled by default](https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h#L247)

henningkayser:
> I was complaining that _the core_ change from yesterday is undone in a pull request by the reviewer one day after merge. Testing on multiple configurations is awesome (do you test on hardware?), but so is a logical analysis of the patch. ;)

But that wasn't the core change at all. The core change was that the bool collision result was actually checked additional to the not sufficient distance value. I'm just removing the second check because it's already included in the first one.

> > The distance-based collision check includes the binary one, so a second check additional to the distance-based isn't required.
> 
> Could your issues come from _unsigned_ distance requests? [Signed distances (in case of collisions are disabled by default](https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h#L247)

That's possible. @AndyZe Did you look at the distance values?

AndyZe:
Not with great care

v4hn:
> But that wasn't the core change at all. The core change was that the bool collision result was actually checked additional to the not sufficient distance value. I'm just removing the second check because it's already included in the first one.

True, although this is 3 lines below this second check you remove here :grin: .

Looking further into the package, adding the signed distance option is probably not a good idea without more modifications. It adds computational overhead in case of collisions, so unless you want to keep moving in "mild" collisions - something you will run into sooner or later - you don't need the penetration depth.

henningkayser:
> At line 100, move `collision_request.distance = true;` outside the loop

Done, also changed to access `collision_result.distance` directly instead of the variable.

PR #1987, opened 2020-03-27
Fix CATKIN_DEPENDS: catkin_lint 1.6.7

henningkayser:
CI was failing because of missing CATKIN_DEPENDS errors from catkin_lint 1.6.7

See https://travis-ci.org/github/ros-planning/moveit/jobs/667658214#L694


v4hn:
Interesting to see a version update (1.6.6-1.6.7) break CI between a pull-request and the merged commit.
Thanks for the cleanup. I keep wondering how such issues have not been noticed/fixed years ago...

PR #1995, opened 2020-03-28
add tests for move_group interface

tylerjw:
### Description

I realized there was no test of the c++ interface of move_group.  This PR fixes that to defend the interface used by our newest users against unintended breakages.

### Checklist
- [x] Moves the CodeCov report in the right direction :arrow_up: :arrow_up: :arrow_up:

### DIFF
--- a/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
+++ b/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
@@ -58,8 +58,8 @@
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 #include <eigen_conversions/eigen_msg.h>
 
-// 1mm acuracy tested for position and orientation
-constexpr double EPSILON = 1e-3;
+// 1cm acuracy tested for position and orientation
+constexpr double EPSILON = 1e-2;
 
 static const std::string PLANNING_GROUP = "panda_arm";
 constexpr double PLANNING_TIME_S = 30.0;
--- a/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
+++ b/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
@@ -240,14 +240,6 @@ TEST_F(MoveGroupTestFixture, JointSpaceGoalTest)
 {
   SCOPED_TRACE("JointSpaceGoalTest");
 
-  // set a custom start state
-  geometry_msgs::Pose start_pose;
-  start_pose.orientation.w = 1.0;
-  start_pose.position.x = 0.55;
-  start_pose.position.y = -0.05;
-  start_pose.position.z = 0.8;
-  PlanAndMoveToPose(start_pose);
-
   // set start state for planning
   move_group_->setStartStateToCurrentState();
 
@@ -258,7 +250,7 @@ TEST_F(MoveGroupTestFixture, JointSpaceGoalTest)
 
   // Now, let's modify the joint positions.  (radians)
   ASSERT_EQ(plan_joint_positions.size(), std::size_t(7));
-  plan_joint_positions = { 1.0, -0.5, 0.5, -1.0, 2.0, 1.0, -1.0 };
+  plan_joint_positions = { 1.2, -1.0, -0.1, -2.4, 0.0, 1.5, 0.6 };
   move_group_->setJointValueTarget(plan_joint_positions);
 
   // plan and move
--- a/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
+++ b/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
@@ -216,6 +216,10 @@ TEST_F(MoveGroupTestFixture, PathConstraintTest)
   start_pose.position.z = 0.8;
   planAndMoveToPose(start_pose);
 
+  // NOTE: If the link names change this will fail.  This was put here because
+  // the tutorial for C++ move_group usese these links.  If the link names are
+  // change this and the tutorial will have to be updated.
+
   // create an orientation constraint
   moveit_msgs::OrientationConstraint ocm;
   ocm.link_name = "panda_link7";
--- a/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
+++ b/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
@@ -73,21 +73,12 @@ public:
     nh_ = ros::NodeHandle("/move_group_interface_cpp_test");
     move_group_ = std::make_shared<moveit::planning_interface::MoveGroupInterface>(PLANNING_GROUP);
 
-    // get the starting pose
-    start_pose_stamped_ = move_group_->getCurrentPose();
-
     // set velocity and acceleration scaling factors (full speed)
     move_group_->setMaxVelocityScalingFactor(MAX_VELOCITY_SCALE);
     move_group_->setMaxAccelerationScalingFactor(MAX_ACCELERATION_SCALE);
 
     // allow more time for planning
     move_group_->setPlanningTime(PLANNING_TIME_S);
-  }
-
-  void retrunToStartPose()
-  {
-    SCOPED_TRACE("retrunToStartPose");
-    planAndMoveToPose(start_pose_stamped_.pose);
   }
 
   void planAndMoveToPose(const geometry_msgs::Pose& pose)
@@ -162,7 +153,6 @@ protected:
   ros::NodeHandle nh_;
   moveit::planning_interface::MoveGroupInterfacePtr move_group_;
   moveit::planning_interface::PlanningSceneInterface planning_scene_interface_;
-  geometry_msgs::PoseStamped start_pose_stamped_;
 };
 
 TEST_F(MoveGroupTestFixture, StartingConditionsTest)
@@ -235,9 +225,6 @@ TEST_F(MoveGroupTestFixture, MoveToPoseTest)
 
   // get the pose after the movement
   testPose(eigen_target_pose);
-
-  // return to start pose for next test
-  retrunToStartPose();
 }
 
 TEST_F(MoveGroupTestFixture, JointSpaceGoalTest)
@@ -262,9 +249,6 @@ TEST_F(MoveGroupTestFixture, JointSpaceGoalTest)
 
   // test that we moved to the expected joint positions
   testJointPositions(plan_joint_positions);
-
-  // return to start pose for next test
-  retrunToStartPose();
 }
 
 TEST_F(MoveGroupTestFixture, PathConstraintTest)
@@ -305,9 +289,6 @@ TEST_F(MoveGroupTestFixture, PathConstraintTest)
 
   // get the pose after the movement
   testPose(target_pose);
-
-  // return to start pose for next test
-  retrunToStartPose();
 }
 
 TEST_F(MoveGroupTestFixture, CartPathTest)
@@ -349,9 +330,6 @@ TEST_F(MoveGroupTestFixture, CartPathTest)
 
   // get the pose after the movement
   testPose(target_waypoint);
-
-  // return to start pose for next test
-  retrunToStartPose();
 }
 
 TEST_F(MoveGroupTestFixture, CollisionObjectsTest)
@@ -421,9 +399,6 @@ TEST_F(MoveGroupTestFixture, CollisionObjectsTest)
   EXPECT_EQ(planning_scene_interface_.getObjects().size(), std::size_t(1));
   planning_scene_interface_.removeCollisionObjects(object_ids);
   EXPECT_EQ(planning_scene_interface_.getObjects().size(), std::size_t(0));
-
-  // return to start pose for next test
-  retrunToStartPose();
 }
 
 int main(int argc, char** argv)
--- a/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
+++ b/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
@@ -108,7 +108,7 @@ public:
   void testPose(const Eigen::Isometry3d& expected_pose)
   {
     SCOPED_TRACE("testPose(const Eigen::Isometry3d&)");
-    // get the pose after the movement
+    // get the pose of the end effector link after the movement
     geometry_msgs::PoseStamped actual_pose_stamped = move_group_->getCurrentPose();
     Eigen::Isometry3d actual_pose;
     tf::poseMsgToEigen(actual_pose_stamped.pose, actual_pose);
--- a/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
+++ b/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
@@ -58,7 +58,8 @@
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 #include <eigen_conversions/eigen_msg.h>
 
-constexpr double EPSILON = 1e-2;
+// 1mm acuracy tested for position and orientation
+constexpr double EPSILON = 1e-3;
 
 static const std::string PLANNING_GROUP = "panda_arm";
 constexpr double PLANNING_TIME_S = 30.0;
@@ -84,7 +85,6 @@ public:
   void planAndMoveToPose(const geometry_msgs::Pose& pose)
   {
     SCOPED_TRACE("planAndMoveToPose");
-    move_group_->setStartStateToCurrentState();
     ASSERT_TRUE(move_group_->setJointValueTarget(pose));
     planAndMove();
   }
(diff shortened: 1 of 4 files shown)


tylerjw:
I've tried rebasing this on earlier versions of master to get it past travis (failing unrelated tests) but gave up when I couldn't find a version where it doesn't fail on clang.  I think this must be as a result of some external dependency changing.  I created an issue for it here: https://github.com/ros-planning/moveit/issues/1996

The clang CI failure is unrelated (this new test passes).

moveit_ros/planning_interface/test/move_group_interface_cpp_test.test
```diff
@@ -0,0 +1,29 @@
...
+  <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher"/>
+
+  <!-- Run the main MoveIt executable with fake trajectory execution -->
+  <include file="$(find panda_moveit_config)/launch/move_group.launch">
+    <arg name="allow_trajectory_execution" value="true"/>
+    <arg name="fake_execution" value="true"/>
+    <arg name="info" value="true"/>
+  </include>
+
+  <!-- Subframes test -->
```
rhaschke: The comment is outdated ;-)

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
```diff
@@ -0,0 +1,463 @@
...
+  EXPECT_EQ(move_group_->getName(), "panda_arm");
+  EXPECT_EQ(move_group_->getPlanningFrame(), "world");
+  EXPECT_EQ(move_group_->getVariableCount(), std::size_t(7));
+  EXPECT_EQ(move_group_->getDefaultPlannerId(), "");
+  EXPECT_EQ(move_group_->getPlanningTime(), 10.0);
+  EXPECT_EQ(move_group_->getGoalJointTolerance(), 0.0001);
+  EXPECT_EQ(move_group_->getGoalPositionTolerance(), 0.0001);
+  EXPECT_EQ(move_group_->getGoalOrientationTolerance(), 0.001);
+
+  const std::vector<std::string> NAMED_TARGETS = { "ready", "extended" };
```
rhaschke: These tests rely on specific details of panda_moveit_config.
Changing these details, e.g. adding more named targets, renaming joints, etc. will make this test fail for no reason.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
```diff
@@ -0,0 +1,463 @@
...
+  std::vector<std::string> named_targets = move_group_->getNamedTargets();
+  ASSERT_EQ(named_targets.size(), NAMED_TARGETS.size());
+  for (size_t i = 0; i < named_targets.size(); ++i)
+    EXPECT_EQ(named_targets[i], NAMED_TARGETS[i]) << "index: " << i;
+
+  const std::vector<std::string> JOINT_MODEL_GROUP_NAMES = { "hand", "panda_arm", "panda_arm_hand" };
```
rhaschke: All these tests follow the same scheme. Please factor that out into a separate function.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+  // set with eigen, get ros message representation
+  move_group_->setPoseTarget(eigen_target_pose);
+  geometry_msgs::PoseStamped set_target_pose = move_group_->getPoseTarget();
+  Eigen::Isometry3d eigen_set_target_pose;
+  tf::poseMsgToEigen(set_target_pose.pose, eigen_set_target_pose);
+
+  // expect that they are identical
+  {
+    std::stringstream ss;
```
rhaschke: Again, please factor this code out into a function to be reused.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
tylerjw: done

moveit_ros/planning_interface/test/move_group_interface_cpp_test.test
tylerjw: fixed

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
tylerjw: Changing parts of the panda_moveit_config could cause the tutorials to stop working.  I'm not sure we need this level of testing on this.  Would you like me to just drop this test?

tylerjw:
This is currently failing one of the tests on travis, and is therefore WIP, I'll post here when I get it fully working again.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
rhaschke: I suggest to adapt the test such that only required prerequisites are validated.
Thus, if you need let's say the named target `ready` in your test, just test for its existence, but don't constrain the list of other named targets.

rhaschke:
Weird, that the bullet collision test are failing now. You just added a test, which should not affect other tests.

tylerjw:
This is no longer WIP as it is passing travis.  @rhaschke please review.  Thank you.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
tylerjw: Ok, I'll update this test to still call those interfaces but loosen what it tests for to just what is needed for the tutorial to work.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+  waypoints.push_back(target_waypoint);  // up and left
+
+  moveit_msgs::RobotTrajectory trajectory;
+  const double jump_threshold = 0.0;
+  const double eef_step = 0.01;
+  move_group_->computeCartesianPath(waypoints, eef_step, jump_threshold, trajectory);
+
+  // Execute trajectory
+  moveit::planning_interface::MoveGroupInterface::Plan cartesian_plan;
+  cartesian_plan.trajectory_ = trajectory;
```
v4hn, line 328: There is no need for this workaround anymore in `master`. You can just run `move_group_->execute(trajectory)`. If a tutorial (in its `master` version) still specifies this, we should adjust that too.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+                      move_group_->getJointNames(), "joint_names");
+  testVectorOfStrings({ "panda_link1", "panda_link2", "panda_link3", "panda_link4", "panda_link5", "panda_link6",
+                        "panda_link7", "panda_link8" },
+                      move_group_->getLinkNames(), "link_names");
+  testVectorOfStrings({ "panda_joint1", "panda_joint2", "panda_joint3", "panda_joint4", "panda_joint5", "panda_joint6",
+                        "panda_joint7" },
+                      move_group_->getActiveJoints(), "active_joints");
+  testVectorOfStrings({ "panda_joint1", "panda_joint2", "panda_joint3", "panda_joint4", "panda_joint5", "panda_joint6",
+                        "panda_joint7", "panda_joint8" },
+                      move_group_->getJoints(), "joints");
```
v4hn, line 193: The only aspects of this test the other tests rely on are (1) there are 7 joints and (2) `panda_link0` and `panda_link7` exist. The latter is only needed because of the `OrientationConstraint` test and I would propose to change it to use `getPlanningFrame()` and `getEndEffectorLink()` instead.

Anything else in here that is actually relevant to stay in sync with a tutorial, please refactor it into a different test case `MoveGroupTutorialSupport` and add a comment at the top stating that these tests do not indicate functional breakage, but should never be adjusted without adjusting the respective tutorials as well.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+// MoveIt
+#include <moveit/planning_scene_interface/planning_scene_interface.h>
+#include <moveit/move_group_interface/move_group_interface.h>
+
+// TF2
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+#include <eigen_conversions/eigen_msg.h>
+
+constexpr double EPSILON = 1e-2;
```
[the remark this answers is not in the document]
tylerjw: I was able to get this to consistently pass locally at (1mm) `1e-3` but nothing smaller.  This epsilon is used for the diff of all values within transforms of final positions after execution.  In my subframe test I used 1cm (1e-2) accuracy for testing final poses as the precision achieved by path plans wouldn't consistently be better than that and would result in a flaky test.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+  EXPECT_EQ(move_group_->getEndEffectorLink(), "panda_link8");
+  EXPECT_EQ(move_group_->getEndEffector(), "");
+  EXPECT_EQ(move_group_->getPoseReferenceFrame(), "world");
+  EXPECT_EQ(move_group_->getName(), "panda_arm");
+  EXPECT_EQ(move_group_->getPlanningFrame(), "world");
+  EXPECT_EQ(move_group_->getVariableCount(), std::size_t(7));
+  EXPECT_EQ(move_group_->getDefaultPlannerId(), "");
+  EXPECT_EQ(move_group_->getGoalJointTolerance(), 0.0001);
+  EXPECT_EQ(move_group_->getGoalPositionTolerance(), 0.0001);
+  EXPECT_EQ(move_group_->getGoalOrientationTolerance(), 0.001);
```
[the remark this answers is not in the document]
tylerjw: I removed these tests.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+    move_group_->setMaxAccelerationScalingFactor(MAX_ACCELERATION_SCALE);
+
+    // allow more time for planning
+    move_group_->setPlanningTime(PLANNING_TIME_S);
+  }
+
+  void planAndMoveToPose(const geometry_msgs::Pose& pose)
+  {
+    SCOPED_TRACE("planAndMoveToPose");
+    move_group_->setStartStateToCurrentState();
```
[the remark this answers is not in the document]
tylerjw: Good catch, removed this.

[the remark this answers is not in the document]
tylerjw: I found it was needed in several places to kill a warning.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
tylerjw: Thank you, here is an issue for tracking the change to the tutorial: https://github.com/ros-planning/moveit_tutorials/issues/479

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
tylerjw: I removed the test that checks the default state of the panda arm.  If changes to the panda config break this test they will also break the tutorial unless both are updated as the tests here test the functionality needed for the C++ move_group tutorial.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
tylerjw: PR: https://github.com/ros-planning/moveit_tutorials/pull/480

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
tylerjw: As to using getPlanningFrame() that returns `world` and `getEndEffectorLink()` returns `panda_link8`.  I'm not sure how I would use those to get the strings `panda_link0` and `panda_link7`.  I think it is reasonable that the `OrientationConstraint` will now fail if those link names are changed as it will cause the tutorial to stop working as this is copied from the tutorial

tylerjw: I added a comment to explicitly call this out.

tylerjw:
Rebased on master, this is ready for review again (assuming it passes CI) @v4hn and @rhaschke.  Thank you for the reviews so far.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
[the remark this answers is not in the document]
tylerjw: Putting it at 1mm caused it to fail here: https://travis-ci.org/github/ros-planning/moveit/jobs/672252650?utm_medium=notification&utm_source=github_status

I put it back at 1cm.  @rhaschke do you have thoughts on this?

tylerjw:
This [succeeded on travis)(https://travis-ci.org/github/ros-planning/moveit/builds/672280250?utm_medium=notification&utm_source=github_status).

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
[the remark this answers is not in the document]
v4hn: What's the warning? I'm not aware of a related one off the top of my head.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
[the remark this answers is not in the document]
v4hn: Could you try to add `move_group_->setGoalTolerance(0);` to the fixture?
Execution runs through the exact waypoints generated by the planner and the size of the valid goal space for the planner is determined by the tolerance padding (which is non-zero by default). Aside from that I would have to look into the code further to find noise sources.

Ensuring we *can* control the noise in the pipeline up to a sufficient epsilon is rather vital for MoveIt and `1e-5` is a typical industry requirement.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
v4hn: You are awesome, thanks :sunglasses:

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
v4hn: panda_frames

Looking at the frames, `world` and `panda_link0` are colocated, `panda_link7` and `panda_link8` are aligned in the z axis.
I don't see why the constraint should fail if you change the frames to the more generic ones, such that the test does not reference implementation details of the model.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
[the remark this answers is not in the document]
tylerjw: I must have been confused.  I can't reproduce the warning so I removed it in the 2 other places I was calling this method.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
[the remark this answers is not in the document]
tylerjw: 0 causes some of the plans to fail, but if I set it to 1e-6 I can set the epsilon to 1e-5.  Thank you for coming back to this, I didn't understand what you meant by the goal tolerance affecting it.

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
tylerjw: ah, done

moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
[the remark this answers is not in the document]
v4hn: Sounds much better!

Still, planning should succeed just as well with tolerances of 0.
The main difference is whether the IK solver gets more attempts on *almost* the same goal position... The only problem I might expect is a target right in a singularity which can pose a problem to some IK solvers. Wiggling around the real target could help there.

I'm fine with having separate tests for behavior w.r.t. different tolerance values in the future though, as long as `1e-5` suffices here now.

tylerjw:
@v4hn Thank you for the review.  [CI is passing with my latest changes.](https://travis-ci.org/github/ros-planning/moveit/builds/672569145)  I am ready for another pass on this.

tylerjw:
@v4hn thank you for the help getting this in.  We are slowly making a dent in the code coverage.

rhaschke:
Sorry, just noticed this now: Our test cases shouldn't use `panda_moveit_config` (see #1904).
@tylerjw, could you please replace this package with `moveit_resources`? The patch https://github.com/ros-planning/panda_moveit_config/pull/63 should be applied there as well.

v4hn:
One more addendum: MoveIt's [fake controllers support multiple simulation modes](https://ros-planning.github.io/moveit_tutorials/doc/controller_configuration/controller_configuration_tutorial.html#fake-controller-manager) and it makes sense to have tests use different modes to avoid unnecessary runtimes.

For this test, I propose to set the types to 'last point', as you do not test the actual execution of the trajectory.

This **should** be achievable via
```
<param name="/move_group/controller_list[0]/type" value="last point"/>
<param name="/move_group/controller_list[1]/type" value="last point"/>
```
or
```
<rosparam param='move_group/controller_list'>[ { type: 'last point'}, { type: 'last point' }]</rosparam>
```
But the syntax to address individual list elements (`[0]`) does not work and the latter command overrides the rest of the yaml tree, removing the main controller configuration...

I just digged through the rosparam APIs and cannot find a way to set an individual ros parameter inside a list at all and this seems to go all the way down to `xmlrpc`, although one could probably write a cumbersome construct to parse this in `rosparam` and add a patch there *grml*.

Alternatively, we could allow a global default for the fake controllers to be set in `/move_group/controller_list/type`, but someone would have to write a patch for that too (assuming it does not work yet :eyes:

PR #1998, opened 2020-03-30
the last 9 months of patches in master

v4hn:
```
git cherry -v melodic-devel master 1.0.1 | grep '^+'
```
yields 267 commits (run the command if you want to see them) that have not been byte-exact backported from `master` to `melodic-devel`.
Of course, we do not want to backport *everything* to avoid breaking API users might depend on.
The list also contains a number of false-positives with changes that were trivially modified during backport.

As requested in #1990, I sieved through the list over the weekend and applied everything I could that would probably not disturb users. The final list includes 67 commits.
I excluded two notable change types:
- clang-tidy cleanup
  - These changes touch way too many files to backport. It would be easier to rerun the commands for the branch if we really need that
  - I had to adjust a number of commits after them by hand (mostly to use index-based notation over ranged loops, nothing critical)
- subframe support
  - as the "final" cleanup on this support was only merged yesterday, this was clearly not a candidate for backporting
  - commits for this are all over the place in the commit history and in source files and I would probably have had to adjust quite a number of patches.
  - In theory this is backportable, as [moveit_msgs contains the required message changes](https://github.com/ros-planning/moveit_msgs/blob/melodic-devel/msg/CollisionObject.msg#L30), but anyone is welcome to do that in a separate pull-request.

### DIFF
--- a/moveit/CMakeLists.txt
+++ b/moveit/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 2.8.3)
 project(moveit)
 find_package(catkin REQUIRED)
 catkin_metapackage()
--- a/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
@@ -47,6 +47,7 @@
 #include <QCloseEvent>
 #include <QMessageBox>
 #include <QString>
+#include <QDir>
 #include <pluginlib/class_loader.hpp>  // for loading all avail kinematic planners
 // Rviz
 #include <rviz/render_panel.h>
@@ -113,13 +114,7 @@ SetupAssistantWidget::SetupAssistantWidget(QWidget* parent, const boost::program
   }
   else
   {
-    // Open the directory where the MSA was started from.
-    // cf. http://stackoverflow.com/a/7413516/577001
-    QString pwdir("");
-    char* pwd;
-    pwd = getenv("PWD");
-    pwdir.append(pwd);
-    start_screen_widget_->stack_path_->setPath(pwdir);
+    start_screen_widget_->stack_path_->setPath(QDir::currentPath());
   }
 
   // Add Navigation Buttons (but do not load widgets yet except start screen)
--- a/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
@@ -47,7 +47,6 @@
 #include <QCloseEvent>
 #include <QMessageBox>
 #include <QString>
-#include <QDir>
 #include <pluginlib/class_loader.hpp>  // for loading all avail kinematic planners
 // Rviz
 #include <rviz/render_panel.h>
@@ -114,7 +113,7 @@ SetupAssistantWidget::SetupAssistantWidget(QWidget* parent, const boost::program
   }
   else
   {
-    start_screen_widget_->stack_path_->setPath(QDir::currentPath());
+    start_screen_widget_->stack_path_->setPath(QString(getenv("PWD")));
   }
 
   // Add Navigation Buttons (but do not load widgets yet except start screen)
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -586,7 +586,7 @@ class MoveGroupCommander(object):
         traj_out.deserialize(ser_traj_out)
         return traj_out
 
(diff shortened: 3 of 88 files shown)


v4hn:
Tests succeeded locally and the RViz plugin behaves well, also handling collision objects, attach/detach.

Further testing by melodic users is encouraged.

We should not keep this pull-request around for long as it almost certainly breaks with the next commit to melodic-devel / gets outdated with more commits to master.

simonschmeisser:
Wow, that's a lot of changes. Would it make sense to release 1.0.3 now and then merge this PR?

moveit/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.3)
+cmake_minimum_required(VERSION 2.8.3)
```
simonschmeisser: kinetic: >= 3.0.2
melodic: >= 3.7.2

v4hn: Wow, good eyes you got there!
You are right, but these are meta packages and the exact content of the file sadly matters:

https://github.com/ros-planning/moveit/pull/1620

moveit_planners/chomp/chomp_motion_planner/include/chomp_motion_planner/chomp_planner.h
```diff
@@ -50,6 +50,7 @@ class ChompPlanner
   ChompPlanner();
   virtual ~ChompPlanner(){};
 
+  // TODO: switch API from moveit_msgs to MotionPlanRequest / MotionPlanDetailedResponse
```
simonschmeisser: the TODO in this comment seems to have happened already

v4hn: True, e017669ef0c3e18cc02c49b2d60080a4cf581552 is an empty patch that was previously applied.

I will fix this later to give people a stable branch for the moment.

v4hn:
> Wow, that's a lot of changes. Would it make sense to release 1.0.3 now and then merge this PR?

It certainly is a lot of (previously reviewed but partially adapted) patches.

Given that there are also quite a number of unreleased patches in the branch already,
I guess it makes sense to release before this arrives.

At the same time, a lot of these are bug fixes and we would release with a lot of known bugs.

@rhaschke @130s @davetcoleman ?

rhaschke:
I suggest to go with a single release after this PR is merged (and tested, let's say for two more weeks).

v4hn:
I will fix issues, if any more come up with fixup commits for now and will clean the history up before merge.

moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
```diff
@@ -113,13 +113,7 @@ SetupAssistantWidget::SetupAssistantWidget(QWidget* parent, const boost::program
...
   else
   {
-    // Open the directory where the MSA was started from.
-    // cf. http://stackoverflow.com/a/7413516/577001
-    QString pwdir("");
-    char* pwd;
-    pwd = getenv("PWD");
-    pwdir.append(pwd);
-    start_screen_widget_->stack_path_->setPath(pwdir);
+    start_screen_widget_->stack_path_->setPath(QString(getenv("PWD")));
```
simonschmeisser: this looks different here: https://github.com/ros-planning/moveit/pull/1998/commits/eb02dcdba7ac9312202b3bdd56b94006e7f7bd17    ie it uses QDir::currentPath() ?

rhaschke: It turned out that `QDir::currentPath()` doesn't return the original working directory (see #1932).

v4hn:
Thank you for the apparently very thorough review @simonschmeisser !

I reworked the branch to include the fixup commits. From my perspective, this request is ready for merge. Only shadow-fixed fails now.

Further user testing would be appreciated, but as we will have a cool-down period before release, we might also

simonschmeisser:
I see some strange behavior at the moment, not sure if this a regression or maybe me doing something strange ...

anyway, display of RobotState display currently fails and just displays nothing. No matter if I want to display something in our own application or in rviz. RobotModel display (from rviz) works, PlanningScene display works as well (both update correctly)

I'll restart the thing now and see if it persists

rhaschke:
I can confirm the issue with the RobotState display.

v4hn:
https://github.com/ros-planning/moveit/pull/2014 should fix the invisible robot in the robot display.

v4hn:
@simonschmeisser Did you figure out where your "strange behavior" stems from?

simonschmeisser:
"strange behavior" = invisible robot, all good if that is fixed

PR #1999, opened 2020-03-30
update/adjust keys to satisfy travis config validator

v4hn:
travis-config

PR #2000, opened 2020-03-30
allow failure for shadow-fixed

v4hn:
As we see once more at the moment, we *don't* **ever** want our CI to block because someone broke shadow-fixed. Assuming I didn't break anything, this should mitigate half [of your worries](https://github.com/ros-planning/moveit/pull/1999#issuecomment-606048430) @rhaschke ?
The other half being the incomplete migration to github apps that leads to github's guards waiting for completed CI jobs.

add a separate gcc shadow-fixed job as we still want the required clang check.

also adjust clang job to use global variables.
Job-variables are *all* included in the jobname on the website.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ jobs:
     - env: ROS_DISTRO=kinetic
     - compiler: clang
       env: TEST=clang-tidy-fix CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-overloaded-virtual"
+    - env: ROS_REPO=ros-shadow-fixed
   allow_failures:
     - env: ROS_REPO=ros-shadow-fixed
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,11 @@ jobs:
   include:
     - env: TEST="clang-format catkin_lint"
     - env: TEST=code-coverage
-    - env: ROS_DISTRO=melodic BEFORE_DOCKER_SCRIPT="source moveit_kinematics/test/test_ikfast_plugins.sh"
+    - env: ROS_DISTRO=melodic
     - env: ROS_DISTRO=kinetic
     - compiler: clang
-      env: TEST=clang-tidy-fix CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-overloaded-virtual"
+      # test_ikfast_plugins takes ~10 minutes. Take the time off clang-tidy to keep the main jobs shorter
+      env: TEST=clang-tidy-fix BEFORE_DOCKER_SCRIPT="source moveit_kinematics/test/test_ikfast_plugins.sh" CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-overloaded-virtual"
     - env: ROS_REPO=ros-shadow-fixed
   allow_failures:
     - env: ROS_REPO=ros-shadow-fixed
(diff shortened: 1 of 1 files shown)


v4hn:
Ok, I did break things. :-)
Support is welcome, if anyone has a moment, I have not worked a lot with the travis configuration so far.

v4hn:
CI is still running, but the configuration looks good now. Sorry for all the noise (so far).

@rhaschke

.travis.yml
```diff
@@ -21,20 +21,17 @@ env:
...
+    - env: ROS_DISTRO=melodic BEFORE_DOCKER_SCRIPT="source moveit_kinematics/test/test_ikfast_plugins.sh"
+    - env: ROS_DISTRO=kinetic
     - compiler: clang
-      env: ROS_REPO=ros-shadow-fixed
-           TEST=clang-tidy-fix
-           BEFORE_DOCKER_SCRIPT="source moveit_kinematics/test/test_ikfast_plugins.sh"
-           CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-overloaded-virtual"
+      env: TEST=clang-tidy-fix CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-overloaded-virtual"
+  allow_failures:
+    - env: ROS_REPO=ros-shadow-fixed
```
rhaschke: You need to list this job in `include` as well. `allow_failures` is intended to mark an existing job - weird Travis logics...

v4hn:
>  Nevertheless, I'm hesitating to add an extra build job. One can easily see the reason for the failure (whether it is clang-tidy or ros-shadow-fixed being broken) from inspection of the build log in Travis.

That is beside the point. The main point of this patch is to mark the shadow-fixed test as `allow_failure` and we do not want the clang job to be allowed to fail. There is no way around having a separate job if we want that change. And, as I said in the first post, we do not want broken releases of other peoples' software to break our regular CI workflow.

>  Also, I deliberately put the ikfast tests into the clang-tidy test, because this tests adds ~10min and having the basic kinetic and melodic tests finish early, speeds up the development cycle (at least mine).

I see. A comment would have been a good idea. I changed it back and added the comment. :)

v4hn:
@rhaschke As you can see, this would also allow everyone else to merge approved pull-requests for master again. :smiling_imp: 

I do see the overhead, but I am annoyed by the need for admin permissions to merge pull-requests
and the alternative would be to disable the shadow-fixed job altogether.

gavanderhoorn:
Off-topic (somewhat), but: we could consider migrating to Github Actions.

There's a lot more resources available, and for the repositories we've done it, CI runs completed sooner, it was easier to parallelise builds, we haven't run into any build time limitations yet and the available automation templates/pipelines is quite extensive.

v4hn:
> Off-topic (somewhat), but: we could consider migrating to Github Actions.

Go ahead. Apparently neither @rhaschke nor I do have the permissions to do so.

gavanderhoorn:
Me neither.

But this is not something which can be decided on a whim.

I believe it would be something to discuss at the next maintainer meeting. Or would at least require some consensus among maintainers.

rhaschke:
If a Travis job is allowed to fail, we risk to not notice this failure at all. I can only speak for myself, but if I see the green checkmark, I don't look into the details of Travis anymore to check that some job failed.
I'm fine to temporarily disable this failing job, but in the long run I prefer to keep the previous config.

v4hn:
Robert, you are currently the only one who can and also does merge requests, and you do that by exploiting your admin permissions.
Keeping ros-shadow-fixed as a required CI for merging requests is **not** an option.

I fully disabled the shadow-fixed job here now, please merge (possibly adjust as you see fit).

Long term, the only alternative to `allow_failure` I can see would be to restrict this job to the main branches (not run them for pull-requests).

PR #2003, opened 2020-03-31
Travis: re-enable ros-shadow-fixed

rhaschke:
A new eigenpy version is released to ros-shadow-fixed...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ jobs:
       env: TEST=clang-tidy-fix
            BEFORE_DOCKER_SCRIPT="source moveit_kinematics/test/test_ikfast_plugins.sh"
            CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-overloaded-virtual"
-#           ROS_REPO=ros-shadow-fixed
+           ROS_REPO=ros-shadow-fixed
 
 before_script:
   - git clone -q --depth=1 https://github.com/ros-planning/moveit_ci.git .moveit_ci


rhaschke:
eigenpy is still failing

v4hn:
Even once the eigenpy issue is resolved, I oppose this as a required check for CI.

Having this job for pull-requests can and does break our whole merge process which can only be overridden by github administrators.

Two alternatives I see:
- allow the job to fail - in theory it should still send mails on failure?
- add the job only for the main branches. This way we would see issues in shadow-fixed but they would not block CI.

gavanderhoorn:
Third option: enable "cron builds". Those would run independently of CI, and could be used to periodically check whether any upstream changes in MoveIt's dependencies have introduced any breaking changes.

rhaschke:
> Having this job for pull-requests can and does break our whole merge process

And we want it to break early and noticable, so that we can take counter measures...

> ... which can only be overridden by github administrators.

Not really true. Every maintainer can (temporarily) disable the ros-testing by editing `.travis.yml`.

> Third option: enable "cron builds".

Interesting idea. Checking the environment variable [`$TRAVIS_EVENT_TYPE == "cron"`](https://docs.travis-ci.com/user/cron-jobs/) we can limit the corresponding job to cron builds.
A remaining question is how to (email) notify the maintainers in case of a failure of this particular job.

v4hn:
> > ... which can only be overridden by github administrators.
> 
> Not really true. Every maintainer can (temporarily) disable the ros-testing by editing .travis.yml.

I don't see how you can propose auxiliary commits to temporarily change travis and at the same time force-push because of merge commits. Either you aim to keep the commit history clean or you don't.

I just filed https://github.com/ros-planning/moveit/pull/2012

rhaschke:
So, you don't like temporarily modifying Travis?

v4hn:
No.

PR #2004, opened 2020-03-31
Fix some clang tidy issues

rhaschke:
Several clang-tidy issues slipped through our CI checker, which only considers changed files.

### DIFF
--- a/moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
+++ b/moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
@@ -97,9 +97,9 @@ MoveItCpp::MoveItCpp(const Options& options, const ros::NodeHandle& /*unused*/,
   ROS_DEBUG_NAMED(LOGNAME, "MoveItCpp running");
 }
 
-MoveItCpp::MoveItCpp(MoveItCpp&& other)
+MoveItCpp::MoveItCpp(MoveItCpp&& other) noexcept
 {
-  other.clearContents();
+  *this = std::move(other);
 }
 
 MoveItCpp::~MoveItCpp()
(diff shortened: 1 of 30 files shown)


moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -108,14 +108,16 @@ MoveItCpp::~MoveItCpp()
...
   {
     this->node_handle_ = other.node_handle_;
-    this->tf_buffer_ = other.tf_buffer_;
-    this->robot_model_ = other.robot_model_;
-    this->planning_scene_monitor_ = other.planning_scene_monitor_;
+    this->tf_buffer_ = std::move(other.tf_buffer_);
+    this->tf_listener_ = std::move(other.tf_listener_);
+    this->robot_model_ = std::move(other.robot_model_);
+    this->planning_scene_monitor_ = std::move(other.planning_scene_monitor_);
+    // TODO: what about the other elements?
```
rhaschke, line 120: @henningkayser, can you comment on this?
Why are the other members of MoveItCpp not handled here?

henningkayser: You're right, the other components should be moved as well. Thanks for looking into this. They were probably added later and moving MoveItCpp was never used in practice, so it wasn't noticed

henningkayser: I can create a fixup PR

tylerjw: Is there any reason the move or copy constructor needs to explicitly defined?  It looks like the default move constructor behavior is what is intended.  Why not use the default move constructor by putting this in the header?

```c++
MoveItCpp(const MoveItCpp& other)            = default;
MoveItCpp(MoveItCpp&& other)                 = default;
MoveItCpp& operator=(const MoveItCpp& other) = default;
MoveItCpp& operator=(MoveItCpp&& other)      = default;
```

rhaschke:
Rebased to resolve conflicts.

Issue #1957, opened 2020-03-17
Add units to Rviz planning plugin, especially "Goal Tolerance"


AndyZe:
It would be nice if this field indicated what the units were. Radians, I guess? Does it apply to every joint?

![path_constraints_units](https://user-images.githubusercontent.com/11284393/76899267-334ea880-6865-11ea-9063-d7f873711ae9.png)

v4hn:
`grep`ing through the source code, this field is not actually used anywhere at all.
Especially as it is listed as part of the `Path Constraints` box, I wouldn't know what it should mean anyway.

Please file a pull-request to remove the option from the gui or move it to the options block and actually fill in the `MoveGroup` action goal (where it would indeed be given in rad per joint).

PR #2007, opened 2020-04-01
Remove the "Goal Constraints" RViz GUI field

AndyZe:
As described in issue #1957, the Goal Constraints field in the RViz plugin didn't make sense. This just removes it.

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
@@ -1330,8 +1330,6 @@ void MotionPlanningDisplay::load(const rviz::Config& config)
       frame_->ui_->velocity_scaling_factor->setValue(d);
     if (config.mapGetFloat("Acceleration_Scaling_Factor", &d))
       frame_->ui_->acceleration_scaling_factor->setValue(d);
-    if (config.mapGetFloat("MoveIt_Goal_Tolerance", &d))
-      frame_->ui_->goal_tolerance->setValue(d);
 
     bool b;
     if (config.mapGetBool("MoveIt_Allow_Replanning", &b))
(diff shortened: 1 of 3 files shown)


rhaschke:
Instead of removing this field, why not setting the goal tolerance via 
`move_group_->setGoalTolerance(double)`?

rhaschke:
I don't really agree. When providing a missing implementation is the same effort as removing the half-implemented feature, why not providing the implementation?

v4hn:
> I don't really agree.
> When providing a missing implementation is the same effort as removing the half-implemented feature,
> why not providing the implementation?

Because the feature would be confusing and make the system behavior appear inconsistent.
You *do* expect the robot to arrive at the visually shown InteractiveMarker state, no matter the parameterization.
The tolerance parameters are mostly of interest for Pose goals and "forgiving" motion requests from code.

The "half-implementation" you talk about is 4 lines of c&p'able code + a qt ui field in the wrong place.

I don't see a single reason to have this parameter available in the GUI. We also don't have an option to specify Pose goals there.

rhaschke:
> I don't see a single reason to have this parameter available in the GUI. We also don't have an option to specify Pose goals there.

The proposed function goal is not limited to (Cartesian) pose goals:
https://github.com/ros-planning/moveit/blob/2a98624b26f7ad0be42c272e0787e538436bfedb/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L1968-L1973

You should have this parameter available in the GUI in order to flawlessly execute a plan on a real robot, whose controllers cannot exactly reach the goal.

v4hn:
> You should have this parameter available in the GUI in order to flawlessly execute a plan on a real robot, whose controllers cannot exactly reach the goal.

This is the *planning* tolerance, which has *nothing* to do with the controller tolerances.

I would like to merge this as-is.

AndyZe:
Closing the stale PR

felixvd:
I'd rather merge than close this to be honest. Evidently no one has the cycles to implement it, but I'd rather remove it than keep a dead interface visible. It would only add noise.

v4hn:
I actually approved and want to see this merged.

The only reason why this was not merged is because @rhaschke disagreed and almost no other maintainer feels able to merge pull-requests in MoveIt these days :grumpy_smurf: .

@AndyZe If you still have it around please reopen the pull-request and push some of your coworkers to merge it @henningkayser @davetcoleman @mlautman @tylerjw

PR #2008, opened 2020-04-02
Split collision proximity threshold

JStech:
### Description

Replaces `collision_proximity_threshold` config parameter with `self_collision_proximity_threshoold` and `scene_collision_proximity_threshold`, so that a lower self-collision threshold can be used. This helps handle near-collisions we suspect are from links that are always near to each other, although unlikely to collide.

### Checklist


moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -62,7 +62,11 @@ class JogCalcs
 
   void stopMainLoop();
 
-  void haltOutgoingJogCmds();
+  /** \brief Stop publishing jog commands while keeping the main loop spinning and filters up to date */
+  void pauseOutgoingJogCmds();
```
henningkayser, line 66: These shouldn't be in here, did you rebase onto the latest version of #1982?

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -99,12 +101,15 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
     // Do a thread-safe distance-based collision detection
     collision_request.distance = true;
     getLockedPlanningSceneRO()->checkCollision(collision_request, collision_result, current_state);
```
henningkayser, line 103: This also still includes the duplicate collision check. Please rebase

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -99,12 +101,15 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
...
-    collision_distance = collision_result.distance;
+    self_collision_distance = collision_result.distance;
+    scene_collision_distance = getLockedPlanningSceneRO()->distanceToCollision(current_state);
```
henningkayser, line 105: Maybe we should keep the planning scene locked for both calls.

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -113,15 +118,20 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
...
       // velocity_scale = e ^ k * (collision_distance - threshold)
       // k = - ln(0.001) / collision_proximity_threshold
       // velocity_scale should equal one when collision_distance is at collision_proximity_threshold.
       // velocity_scale should equal 0.001 when collision_distance is at zero.
-      velocity_scale =
-          exp(velocity_scale_coefficient * (collision_distance - parameters_.collision_proximity_threshold));
+      velocity_scale = exp(scene_velocity_scale_coefficient *
+                           (scene_collision_distance - parameters_.scene_collision_proximity_threshold));
+    }
+    else if (self_collision_distance < parameters_.self_collision_proximity_threshold)
```
henningkayser, line 131: I think I wouldn't split both distance checks as if/else. Instead I would control for the lower velocity scale, even if this probably doesn't have that many practical implications.

moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
```diff
@@ -53,4 +53,5 @@ command_out_topic: joint_group_position_controller/command # Publish outgoing co
 ## Collision checking for the entire robot body
 check_collisions: true # Check collisions?
 collision_check_rate: 10 # [Hz] Collision-checking can easily bog down a CPU if done too often.
-collision_proximity_threshold: 0.01 # Start decelerating when a collision is this far [m]
+self_collision_proximity_threshold: 0.01 # Start decelerating when a self-collision is this far [m]
+scene_collision_proximity_threshold: 0.02 # Start decelerating when a scene collision is this far [m]
```
henningkayser, line 57: Maybe we should increase the default scene proximity threshold a bit?

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
AndyZe, line 131: This second else-if could get skipped if the first else-if is true... even if the robot is very close to self-collision. Maybe it would be better to do:

```
else if ((scene_collision_distance < parameters_.scene_collision_proximity_threshold) || (self_collision_distance < parameters_.self_collision_proximity_threshold))
{
  double self_collision_scale = ...
  double scene_collision_scale = ...
  velocity_scale = std::min(self_collision_scale, scene_collision_scale);
}
```

AndyZe:
Did you test it? Maybe on the UR5 simulation? I can do that, if you want

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
JStech: Sorry, I hadn't. Now up to date.

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
JStech: I did think through this and thought I'd avoided any bad outcomes, but on second thought it's not very maintainable this way. Instead I just made three discrete if blocks, using `velocity_scale = std::min(velocity_scale, ...);` in each, so we always use the slowest result.

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -99,12 +101,15 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
...
     getLockedPlanningSceneRO()->checkCollision(collision_request, collision_result, current_state);
-    collision_distance = collision_result.distance;
+    self_collision_distance = collision_result.distance;
+    scene_collision_distance = getLockedPlanningSceneRO()->distanceToCollision(current_state);
 
     // Do a binary collision detection (helps with strange edge cases like being in collision initially)
     collision_request.distance = false;
     getLockedPlanningSceneRO()->checkCollision(collision_request, collision_result, current_state);
 
+    velocity_scale = 1;
```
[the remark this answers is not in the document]
JStech: Yeah, I felt this was probably a minor bug.

moveit_experimental/moveit_jog_arm/config/ur_simulated_config.yaml
JStech: I bumped it to 5cm.

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
JStech: Yeah--see my comment above. I did something similar to what you suggested.

JStech:
I did test it.

AndyZe:
Looks good to me!

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
```diff
@@ -88,8 +88,10 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
                                     ros_parameters_.lower_singularity_threshold);
   error += !rosparam_shortcuts::get("", n, parameter_ns + "/hard_stop_singularity_threshold",
                                     ros_parameters_.hard_stop_singularity_threshold);
-  error += !rosparam_shortcuts::get("", n, parameter_ns + "/collision_proximity_threshold",
```
henningkayser, line 91: Please add a deprecation warning

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -70,9 +70,13 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
   // Copy the planning scene's version of current state into new memory
   moveit::core::RobotState current_state(getLockedPlanningSceneRO()->getCurrentState());
 
-  double velocity_scale_coefficient = -log(0.001) / parameters_.collision_proximity_threshold;
+  double self_velocity_scale_coefficient = -log(0.001) / parameters_.self_collision_proximity_threshold;
```
henningkayser, line 73: const

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -98,6 +102,10 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
       // Do a thread-safe distance-based collision detection
       getLockedPlanningSceneRO()->checkCollision(collision_request, collision_result, current_state);
```
henningkayser, line 103: Looking at the implementation of [checkCollision()](https://github.com/ros-planning/moveit/blob/master/moveit_core/planning_scene/src/planning_scene.cpp#L482) we are still doing one redundant distance check with the scene. We could remove one check by running `getCollisionEnv()->checkRobotCollision()` and `getCollisionEnvUnpadded()->checkSelfCollision()` separately and then checking the distance and collision results for both.

henningkayser:
Looks good in general, Travis CI seems to require another rebase. Please consider the alternative implementation I suggested.

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
JStech: I tried this, but it's always returning 1.797e308 (i.e. DOUBLE_MAX) for the scene collisions, and 0. or -1. for the self-collisions. I'd appreciate a quick sanity check, if you could give it a look.

henningkayser: Well that's odd. I don't see anything wrong with the code, though. Do you have collision objects in the planning scene (not part of the urdf)? I would expect it to return DOUBLE_MAX in case there are no objects to collide with. For self collisions we should look up how the values are computed. Feel free to start digging in FCL and ping me if you find something, I'll try to debug it later as well.

JStech: I ran it in some more scenarios (two other setups in our current project), and the scene collision distances look believable: about 0.1-0.2m. I'm looking into the self-collision.

JStech: Adding the ACM to the `checkSelfCollision` call fixed it.

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
JStech: done

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
JStech: done

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -92,12 +97,20 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
...
       current_state.updateCollisionBodyTransforms();
 
       // Do a thread-safe distance-based collision detection
-      getLockedPlanningSceneRO()->checkCollision(collision_request, collision_result, current_state);
+      collision_result.clear();
+      getLockedPlanningSceneRO()->getCollisionEnv()->checkRobotCollision(collision_request, collision_result,
+                                                                         current_state);
+      scene_collision_distance = collision_result.distance;
+
+      collision_result.clear();
```
henningkayser, line 108: We need to store the first bool collision result before clearing as well

moveit_core/collision_detection_fcl/src/collision_common.cpp
```diff
@@ -465,19 +465,16 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
                         cd1->getID().c_str(), cd2->getID().c_str());
     }
   }
-  else
+  else if (cd2->type == BodyTypes::ROBOT_LINK && cd1->type == BodyTypes::ROBOT_ATTACHED)
```
henningkayser, line 468: Thanks for the cleanup

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
JStech: Good catch. Done

PR #2010, opened 2020-04-05
[moveit_ros] Change eigenpy to interface (modern CMake)

wxmerkt:
This should future-proof the inclusion of EigenPy by switching to the exported CMake target. 

(Opening PR to run against CI)


wxmerkt:
Closing: We haven't updated eigenpy on Kinetic for a few versions (and the CMake targets aren't exported yet in the version available on Kinetic).

wxmerkt:
New releases are available on both Kinetic and Melodic. Reopening to see if this builds now.

wxmerkt:
Builds on `ros-shadow-fixed`. Need to wait for the next kinetic sync and then I'll retrigger Travis.

Issue #1902, opened 2020-02-14
Eigen RobotState memory issue when linked with OpenVINO


tylerjw:
### Description

I created a minimal example that reproduces this bug here: https://github.com/tylerjw/moveit_memory_bug

OpenVINO is a library provided from intel for ML inference among other things.  If it is linked in (never called) then this short bit of code will generate a memory alignment Eigen error in the constructor for MoveitCpp.  Here is the code in the test:

```
#include <memory>
#include <moveit/moveit_cpp/moveit_cpp.h>
#include <gtest/gtest.h>
#include <ros/ros.h>

namespace {
// Comment out this function that is never called to see the bug go away
void test_function(const planning_scene_monitor::PlanningSceneMonitorPtr
                       &planning_scene_monitor) {
  const moveit::core::RobotStatePtr &kinematic_state =
      planning_scene_monitor->getStateMonitor()->getCurrentState();
  const Eigen::Isometry3d &command_to_camera_frame_transform =
      kinematic_state->getGlobalLinkTransform("base_link").inverse() *
      kinematic_state->getGlobalLinkTransform("camera");
}
}

TEST(ExampleServerTest, ExampleUnitTests) {
  ros::NodeHandle nh("~");
  moveit::planning_interface::MoveItCpp::Options moveit_cpp_options(nh);

  // Initialize MoveIt planning stack using MoveItCpp
  std::shared_ptr<moveit::planning_interface::MoveItCpp> moveit_cpp =
      std::make_shared<moveit::planning_interface::MoveItCpp>(
          moveit_cpp_options, nh);
}

int main(int argc, char **argv) {
  ros::init(argc, argv, "unit_test");
  ::testing::InitGoogleTest(&argc, argv);
  return RUN_ALL_TESTS();
}
```

### Your environment
* ROS Distro: Melodic
* OS Version: e.g. Ubuntu 18.04
* Source build, master branch

### Backtrace/Console output
```
[...]

PR #2011, opened 2020-04-05
WIP: Alignment of vectorizable Eigen member variables

gleichdick:
### Description

According to the [Eigen docs](http://eigen.tuxfamily.org/dox-devel/group__TopicFixedSizeVectorizable.html), fixed-size Eigen members which sizes are a multiple of 16 have to be memory-aligned to be vectorized. Related issues: #1385, #1902

For instances on the stack, the alignment  is expressed in the Eigen classes (probably with `alignas` and friends), the compiler then adds padding to classes which have Eigen members. So this works out of the box, no changes needed. <del>`alignas(EIGEN_MAX_ALIGN_BYTES)` (There is a bunch of macros in Eigen, so I don't know if that's the correct one). </del> In addition, dynamic allocation on the heap requires a custom allocator, provided by Eigen (`Eigen::aligned_allocator<T>`).
IMPORTANT: If these classes are stored in a container, e.g. `std::vector`, the aligned allocator has to be used too.

Furthermore, `std::make_shared` (which creates a shared pointer to a new instance) also does not respect the alignment. This PR provides <del>template specializations</del> the function `aligned_make_shared()` which behaves like `std::make_shared` but with alignment.

<del>To be done: Verify that all containers containing aligned classes use the aligned allocator.</del>

To be discussed: CPU Extensions like AVX512 require further alignment than 16 bytes, the value of the Eigen memory macros depend on the platform. So, alignment issues could occur if Eigen Objects are passed between libraries which were compiled on different platforms. The alignment has to be the same in all compilation units, for example by adding and exporting some Eigen macro definitions in CMakeLists.txt.
So far, I discovered that both `EIGEN_MAX_STATIC_ALIGN_BYTES` and `EIGEN_MAX_ALIGN_BYTES` need to be defined to the same value (16, 32, 64) to hide the platform.

### Checklist

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
@@ -64,7 +64,7 @@ enum Type
   /** \brief A body in the environment */
   WORLD_OBJECT
 };
-}
+}  // namespace BodyTypes
 
 /** \brief The types of bodies that are considered for collision */
 typedef BodyTypes::Type BodyType;
@@ -73,6 +73,7 @@ typedef BodyTypes::Type BodyType;
 struct Contact
 {
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+  using AlignedVector = std::vector<Contact, Eigen::aligned_allocator<Contact>>;
 
   /** \brief contact position */
   Eigen::Vector3d pos;
@@ -147,7 +148,7 @@ struct CollisionResult
   CollisionResult() : collision(false), distance(std::numeric_limits<double>::max()), contact_count(0)
   {
   }
-  typedef std::map<std::pair<std::string, std::string>, std::vector<Contact> > ContactMap;
+  typedef std::map<std::pair<std::string, std::string>, Contact::AlignedVector> ContactMap;
 
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 
@@ -366,7 +367,7 @@ struct DistanceResultsData
 };
 
 /** \brief Mapping between the names of the collision objects and the DistanceResultData. */
-typedef std::map<const std::pair<std::string, std::string>, std::vector<DistanceResultsData> > DistanceMap;
+typedef std::map<const std::pair<std::string, std::string>, std::vector<DistanceResultsData>> DistanceMap;
 
 /** \brief Result of a distance request. */
 struct DistanceResult
@@ -392,4 +393,4 @@ struct DistanceResult
     distances.clear();
   }
 };
-}
+}  // namespace collision_detection
--- a/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_distance_field_types.h
+++ b/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_distance_field_types.h
@@ -69,6 +69,7 @@ enum CollisionType
 
 struct CollisionSphere
 {
+  using AlignedVector = std::vector<CollisionSphere, Eigen::aligned_allocator<CollisionSphere>>;
   CollisionSphere(const Eigen::Vector3d& rel, double radius)
   {
     relative_vec_ = rel;
@@ -187,7 +188,7 @@ public:
    * @param stop_at_first_collision when true the computation is terminated when
    * the first collision is found
    */
-  bool getCollisionSphereGradients(const std::vector<CollisionSphere>& sphere_list,
+  bool getCollisionSphereGradients(const CollisionSphere::AlignedVector& sphere_list,
                                    const EigenSTL::vector_Vector3d& sphere_centers, GradientInfo& gradient,
                                    const CollisionType& type, double tolerance, bool subtract_radii,
                                    double maximum_value, bool stop_at_first_collision);
@@ -199,23 +200,24 @@ protected:
 // determines set of collision spheres given a posed body; this is BAD!
 // Allocation erorrs will happen; change this function so it does not return
 // that vector by value
-std::vector<CollisionSphere> determineCollisionSpheres(const bodies::Body* body, Eigen::Isometry3d& relativeTransform);
+CollisionSphere::AlignedVector determineCollisionSpheres(const bodies::Body* body,
+                                                         Eigen::Isometry3d& relativeTransform);
 
 // determines a set of gradients of the given collision spheres in the distance
 // field
 bool getCollisionSphereGradients(const distance_field::DistanceField* distance_field,
-                                 const std::vector<CollisionSphere>& sphere_list,
+                                 const CollisionSphere::AlignedVector& sphere_list,
                                  const EigenSTL::vector_Vector3d& sphere_centers, GradientInfo& gradient,
                                  const CollisionType& type, double tolerance, bool subtract_radii, double maximum_value,
                                  bool stop_at_first_collision);
 
 bool getCollisionSphereCollision(const distance_field::DistanceField* distance_field,
-                                 const std::vector<CollisionSphere>& sphere_list,
+                                 const CollisionSphere::AlignedVector& sphere_list,
                                  const EigenSTL::vector_Vector3d& sphere_centers, double maximum_value,
                                  double tolerance);
 
 bool getCollisionSphereCollision(const distance_field::DistanceField* distance_field,
-                                 const std::vector<CollisionSphere>& sphere_list,
+                                 const CollisionSphere::AlignedVector& sphere_list,
                                  const EigenSTL::vector_Vector3d& sphere_centers, double maximum_value,
                                  double tolerance, unsigned int num_coll, std::vector<unsigned int>& colls);
 
@@ -238,7 +240,7 @@ public:
 
   Eigen::Isometry3d relative_cylinder_pose_;
 
-  void replaceCollisionSpheres(const std::vector<CollisionSphere>& new_collision_spheres,
+  void replaceCollisionSpheres(const CollisionSphere::AlignedVector& new_collision_spheres,
                                const Eigen::Isometry3d& new_relative_cylinder_pose)
   {
     // std::cerr << "Replacing " << collision_spheres_.size() << " with " <<
@@ -247,7 +249,7 @@ public:
     relative_cylinder_pose_ = new_relative_cylinder_pose;
   }
 
-  const std::vector<CollisionSphere>& getCollisionSpheres() const
+  const CollisionSphere::AlignedVector& getCollisionSpheres() const
   {
     return collision_spheres_;
   }
@@ -291,7 +293,7 @@ protected:
 
   bodies::BoundingSphere relative_bounding_sphere_;
   std::vector<double> sphere_radii_;
-  std::vector<CollisionSphere> collision_spheres_;
+  CollisionSphere::AlignedVector collision_spheres_;
   EigenSTL::vector_Vector3d relative_collision_points_;
 };
 
@@ -302,7 +304,7 @@ public:
 
   PosedBodySphereDecomposition(const BodyDecompositionConstPtr& body_decomposition);
 
-  const std::vector<CollisionSphere>& getCollisionSpheres() const
+  const CollisionSphere::AlignedVector& getCollisionSpheres() const
   {
     return body_decomposition_->getCollisionSpheres();
   }
@@ -374,7 +376,7 @@ public:
   {
   }
 
-  const std::vector<CollisionSphere>& getCollisionSpheres() const
+  const CollisionSphere::AlignedVector& getCollisionSpheres() const
   {
     return collision_spheres_;
   }
@@ -432,7 +434,7 @@ public:
 private:
   PosedBodySphereDecompositionConstPtr empty_ptr_;
   std::vector<PosedBodySphereDecompositionPtr> decomp_vector_;
-  std::vector<CollisionSphere> collision_spheres_;
+  CollisionSphere::AlignedVector collision_spheres_;
   EigenSTL::vector_Vector3d posed_collision_spheres_;
   std::vector<double> sphere_radii_;
   std::map<unsigned int, unsigned int> sphere_index_map_;
@@ -526,4 +528,4 @@ void getCollisionMarkers(const std::string& frame_id, const std::string& ns, con
                          const std::vector<PosedBodySphereDecompositionPtr>& posed_decompositions,
                          const std::vector<PosedBodySphereDecompositionVectorPtr>& posed_vector_decompositions,
                          const std::vector<GradientInfo>& gradients, visualization_msgs::MarkerArray& arr);
-}
+}  // namespace collision_detection
--- a/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_distance_field_types.h
+++ b/moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_distance_field_types.h
@@ -88,6 +88,7 @@ struct GradientInfo
   }
 
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+  using AlignedVector = std::vector<GradientInfo, Eigen::aligned_allocator<GradientInfo>>;
 
   double closest_distance;
   bool collision;
@@ -522,10 +523,10 @@ void getCollisionSphereMarkers(const std_msgs::ColorRGBA& color, const std::stri
 void getProximityGradientMarkers(const std::string& frame_id, const std::string& ns, const ros::Duration& dur,
                                  const std::vector<PosedBodySphereDecompositionPtr>& posed_decompositions,
                                  const std::vector<PosedBodySphereDecompositionVectorPtr>& posed_vector_decompositions,
-                                 const std::vector<GradientInfo>& gradients, visualization_msgs::MarkerArray& arr);
+                                 const GradientInfo::AlignedVector& gradients, visualization_msgs::MarkerArray& arr);
 
 void getCollisionMarkers(const std::string& frame_id, const std::string& ns, const ros::Duration& dur,
                          const std::vector<PosedBodySphereDecompositionPtr>& posed_decompositions,
                          const std::vector<PosedBodySphereDecompositionVectorPtr>& posed_vector_decompositions,
-                         const std::vector<GradientInfo>& gradients, visualization_msgs::MarkerArray& arr);
+                         const GradientInfo::AlignedVector& gradients, visualization_msgs::MarkerArray& arr);
 }  // namespace collision_detection
--- a/moveit_core/utils/include/moveit/utils/aligned_make_shared.h
+++ b/moveit_core/utils/include/moveit/utils/aligned_make_shared.h
@@ -1,0 +1,58 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, RWTH Aachen University
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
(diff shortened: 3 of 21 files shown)


henningkayser:
@gleichdick What a huge effort. In general, I think this goes into the right direction. I also like the use of alias types, but they should probably be a little more descriptive, ideally marking that those are/contain Eigen types. However, this refactoring effort becomes a little more tricky when considering the dynamic memory in classes like RobotState (see https://github.com/ros-planning/moveit/blob/master/moveit_core/robot_state/src/robot_state.cpp#L83). I think we already had the discussion if RobotState could be held by an STL vector/queue inside RobotTrajectory (currently it's a queue of pointers).

> To be discussed: CPU Extensions like AVX512 require further alignment than 16 bytes, the value of the Eigen memory macros depend on the platform. So, alignment issues could occur if Eigen Objects are passed between libraries which were compiled on different platforms. The alignment has to be the same in all compilation units, for example by adding and exporting some Eigen macro definitions in CMakeLists.txt.
So far, I discovered that both EIGEN_MAX_STATIC_ALIGN_BYTES and EIGEN_MAX_ALIGN_BYTES need to be defined to the same value (16, 32, 64) to hide the platform.

This is interesting and would probably explain some issues we found in some of our projects.

moveit_core/utils/include/moveit/utils/aligned_make_shared.h
```diff
@@ -0,0 +1,56 @@
...
+namespace moveit
+{
+namespace core
+{
+/**
+ * \brief memory-aligned version of std::make_shared
+ * This function uses Eigen::aligned_allocator to provide a memory-aligned shared_ptr
+ */
+template <class T, class... Args>
+::std::shared_ptr<T> aligned_make_shared(Args... args)
```
v4hn, line 51: I believe you miss `&&` after `Args` here.
[See Instantiated templates for an rvalue parameter](https://cppinsights.io/lnk?code=I2luY2x1ZGUgPG1lbW9yeT4KCnRlbXBsYXRlIDxjbGFzcyBULCBjbGFzcy4uLiBBcmdzPgo6OnN0ZDo6c2hhcmVkX3B0cjxUPiBhbGlnbmVkX21ha2Vfc2hhcmVkKEFyZ3MuLi4gYXJncykKewogIHJldHVybiA6OnN0ZDo6YWxsb2NhdGVfc2hhcmVkPFQ+KHN0ZDo6YWxsb2NhdG9yPFQ+KCksIDo6c3RkOjpmb3J3YXJkPEFyZ3M+KGFyZ3MpLi4uKTsKfQoKdGVtcGxhdGUgPGNsYXNzIFQsIGNsYXNzLi4uIEFyZ3M+Cjo6c3RkOjpzaGFyZWRfcHRyPFQ+IGFsaWduZWRfbWFrZV9zaGFyZWRfdW5pdmVyc2FsX3JlZihBcmdzJiYuLi4gYXJncykKewogIHJldHVybiA6OnN0ZDo6YWxsb2NhdGVfc2hhcmVkPFQ+KHN0ZDo6YWxsb2NhdG9yPFQ+KCksIDo6c3RkOjpmb3J3YXJkPEFyZ3M+KGFyZ3MpLi4uKTsKfQoKCnN0cnVjdCBXIHsKICBXKGludCYgeCl7fQogIAogIFcoaW50JiYgeCl7fQp9OwoKaW50IG1haW4oKXsKICBpbnQgeCA9IDQyOwogIGF1dG8geSA9IGFsaWduZWRfbWFrZV9zaGFyZWQ8Vz4oc3RkOjptb3ZlKHgpKTsKICBhdXRvIHogPSBhbGlnbmVkX21ha2Vfc2hhcmVkX3VuaXZlcnNhbF9yZWY8Vz4oc3RkOjptb3ZlKHgpKTsKICByZXR1cm4gMDsKfQ==&insightsOptions=cpp17&std=cpp17&rev=1.0)

v4hn:
> To be done: Verify that all containers containing aligned classes use the aligned allocator.

This is going to be a never-ending story.
Is there *any* way to test for these alignment issues automatically in our testbench, so that something will fail if someone ever writes `vector<Contact>` again?
Otherwise I do see this bit-rotting within months...

gavanderhoorn:
> This is going to be a never-ending story.

Not sure whether I've posted it before, but Eigen requires some special care in any case: [Eigen Memory Issues](https://github.com/ethz-asl/eigen_catkin/wiki/Eigen-Memory-Issues).

v4hn:
On Thu, Apr 16, 2020 at 01:54:56AM -0700, G.A. vd. Hoorn wrote:
> Not sure whether I've posted it before, but Eigen requires some special care in any case: [Eigen Memory Issues](https://github.com/ethz-asl/eigen_catkin/wiki/Eigen-Memory-Issues).

Of course. You're also aware of the `eigen_stl_containers` package I guess.
Given how common this is, I would expect *some* automated way to look for these issues.

Alternatively, we might actually want to consider aligning **all** classes (probably very much unreasonable) to avoid the "where is the one broken vector?" egg-hunt.

gavanderhoorn:
The problem is (re: never ending story): iirc, different compiler settings/Eigen alignments across libraries will also result in crashes and subtle problems with data corruption.

In newer C++ versions this is getting better, but in an application which links-in libraries which themselves make use of Eigen, all sorts of nasty problems can crop up.

v4hn:
> In newer C++ versions this is getting better, but in an application which links-in libraries which themselves make use of Eigen, all sorts of nasty problems can crop up.

Yes, but there is nothing we can do about that. I'm not sure it helps a lot to export the constants as @gleichdick proposed. If you have non-default alignment requirements, you will have to compile from scratch anyway. If you have independent libraries you want to link with MoveIt, you have to align those values too and recompile.

Looking at [the eigen bugtracker](https://gitlab.com/libeigen/eigen/issues/1409#note_254723904) it seems this whole mess is not needed anymore with C++17 and a current (unreleased?) version of eigen. Is that right @gleichdick ?

gleichdick:
> This is going to be a never-ending story.
> Is there _any_ way to test for these alignment issues automatically in our testbench, so that something will fail if someone ever writes `vector<Contact>` again?
> Otherwise I do see this bit-rotting within months...

What about some Macros in the constructors/move assignment operators, which do something like `assert(std::ptrdiff_t(this) % EIGEN_MAX_ALIGN_BYTES == 0)` when compiling with e.g. `-DMOVEIT_CHECK_ALIGNMENT=1`? 

@henningkayser RobotState should be fine, because the variables are stored only in dynamic memory which is manually aligned. So the class itself does not need to be aligned.
The most common pitfalls are (in my mind):
 - Storing these classes in the STL (vector, map, queue, ...)
 - create them using `std::make_shared`
 - same with other classes which __contain__ these classes as a member variable or inherit from them.

For testing, I made a complete source install of ROS, FCL, OMPL, Octomap and KDL on Ubuntu Bionic with `CXXFLAGS="-DEIGEN_MAX_STATIC_ALIGN_BYTES=64 -DEIGEN_MAX_ALIGN_BYTES=64"` and build type set to `Debug` (to get the Eigen assertations). My rosinstall file is [here](https://gist.github.com/gleichdick/71599fc5d775c23537212b833c78dccf), I removed the ROS repo to make sure that everything is from scratch. Some packages had to be disabled while running `rosdep`.

Playing around with the Panda shows some errors (when trying to plan and execute):
```
[ INFO] [1587035603.968111835]: Combined planning and execution request received for MoveGroup action. Forwarding to planning and execution pipeline.
[ INFO] [1587035603.970610878]: Planning attempt 1 of at most 1
[ INFO] [1587035603.990377022]: Found a contact between 'panda_link8' (type 'Robot link') and 'panda_link7' (type 'Robot link'), which constitutes a collision. Contact information is not stored.
[ INFO] [1587035603.990811855]: Collision checking is considered complete (collision was found and 0 contacts are stored)
[ INFO] [1587035603.991050046]: Start state appears to be in collision with respect to group panda_arm
[ WARN] [1587035609.658675825]: Unable to find a valid state nearby the start state (using jiggle fraction of 0.050000 and 100 sampling attempts). Passing the original planning request to the planner.                                                                                                  
[ INFO] [1587035609.705192719]: Planner configuration 'panda_arm' will use planner 'geometric::RRTConnect'. Additional configuration parameters will be set when the planner is constructed.
[ WARN] [1587035609.723935107]: panda_arm/panda_arm: Skipping invalid start state (invalid state)
[ERROR] [1587035609.724196294]: panda_arm/panda_arm: Motion planning start tree could not be initialized!
[ WARN] [1587035609.724498756]: panda_arm/panda_arm: Skipping invalid start state (invalid state)
[ERROR] [1587035609.724606679]: panda_arm/panda_arm: Motion planning start tree could not be initialized!
[ WARN] [1587035609.752802602]: panda_arm/panda_arm: Skipping invalid start state (invalid state)
```
However, the KUKA LBR iiwa seems to work.

Many unit tests are failing:
```sh
$ catkin_test_results 
build/moveit_core/test_results/moveit_core/gtest-test_bullet_continuous_collision_checking.xml: 8 tests, 0 errors, 0 failures, 2 skipped
build/moveit_core/test_results/moveit_core/gtest-test_fcl_collision_detection.xml: 22 tests, 0 errors, 2 failures, 0 skipped
build/moveit_planners_chomp/test_results/moveit_planners_chomp/rostest-test_chomp_moveit.xml: 1 tests, 1 errors, 0 failures, 0 skipped
build/moveit_planners_chomp/test_results/moveit_planners_chomp/rosunit-chomp_test.xml: 10 tests, 0 errors, 8 failures, 0 skipped
build/moveit_ros_perception/test_results/moveit_ros_perception/gtest-mesh_filter_test.xml: 48 tests, 0 errors, 24 failures, 0 skipped
build/moveit_ros_planning_interface/test_results/moveit_ros_planning_interface/rosunit-move_group_interface_cpp_test.xml: 10 tests, 0 errors, 10 failures, 0 skipped
build/moveit_ros_planning_interface/test_results/moveit_ros_planning_interface/rosunit-subframes_test.xml: 2 tests, 0 errors, 2 failures, 0 skipped
Summary: 542 tests, 1 errors, 46 failures, 2 skipped
```
The results are [here](https://gist.github.com/gleichdick/5850bff5a6148560758aa22fb92aa6fd)

> Looking at the eigen bugtracker it seems this whole mess is not needed anymore with C++17 and a current (unreleased?) version of eigen. Is that right @gleichdick ?

I hope so, as C++17 has a `new()` operator which [respects alignment](https://en.cppreference.com/w/cpp/memory/new/operator_new)

v4hn:
On Thu, Apr 16, 2020 at 04:52:23AM -0700, Bjar Ne wrote:
> > This is going to be a never-ending story.
> > Is there _any_ way to test for these alignment issues automatically in our testbench, so that something will fail if someone ever writes `vector<Contact>` again?
> > Otherwise I do see this bit-rotting within months...
> 
> What about some Macros in the constructors/move assignment operators, which do something like `assert(std::ptrdiff_t(this) % EIGEN_MAX_ALIGN_BYTES == 0)` when compiling with e.g. `-DMOVEIT_CHECK_ALIGNMENT=1`? 

That's possible, but I was hoping for something on Eigen's side.

> For testing, I made a complete source install of ROS, FCL, OMPL, Octomap and KDL on Ubuntu Bionic with `CXXFLAGS="-DEIGEN_MAX_STATIC_ALIGN_BYTES=64 -DEIGEN_MAX_ALIGN_BYTES=64"` and build type set to `Debug` (to get the Eigen assertations). My rosinstall file is [here](https://gist.github.com/gleichdick/71599fc5d775c23537212b833c78dccf), I removed the ROS repo to make sure that everything is from scratch. Some packages had to be disabled while running `rosdep`.

That sounds much more like a reasonable test for CI, although the Debug build might break some performance tests that would need to be disabled.
Do I understand that right in that you just require an overly big alignment for everything and Eigen's debug asserts catch broken allocations?

> > Looking at the eigen bugtracker it seems this whole mess is not needed anymore with C++17 and a current (unreleased?) version of eigen. Is that right @gleichdick ?
> 
> I hope so, as C++17 has a `new()` operator which [respects alignment](https://en.cppreference.com/w/cpp/memory/new/operator_new)

Then, is it possibly an option to simply move the `master` branch to C++17 instead of trying to maintain the flags in our own code?
People who need specific alignment can then build Eigen from their `master` branch and everything would be fine?


-- 
Michael 'v4hn' Görner, M.Sc. Cognitive Science, PhD Student
Universität Hamburg
Faculty of Mathematics, Informatics and Natural Sciences
Department of Informatics
Group Technical Aspects of Multimodal Systems

Vogt-Kölln-Straße 30
D-22527 Hamburg

Room: F-315
Phone: +49 40 42883-2432
Website: https://tams.informatik.uni-hamburg.de/people/goerner/

rhaschke:
Building MoveIt and all its dependencies from source is not feasible on Travis CI - just takes way too long to compile.

> Then, is it possibly an option to simply move the `master` branch to C++17 instead of trying to maintain the flags in our own code?

I like this suggestion very much. As Michael pointed out, it will be extremely hard to catch all usages of Eigen - now and in future. Switching to C++17 we can avoid that mess.

gavanderhoorn:
It's likely Ubuntu Xenial does not have a GCC which supports C++17 sufficiently for this. That would make `master` not compile any longer on Kinetic.

v4hn:
> It's likely Ubuntu Xenial does not have a GCC which supports C++17 sufficiently for this.

Not just likely. C++17 requires GCC 7+.

> That would make `master` not compile any longer on Kinetic.

Yes. Incidentally that would also allow us to move to boost::python and away from eigenpy apparently.
But deliberately breaking support for kinetic in `master` requires the consent of more maintainers.

And it requires someone to actually go through the trouble to test whether this actually fixes
the discussed alignment problems.

rhaschke:
Obviously, we will also need to wait for the next Eigen release 3.4. Not sure, it will ever make it into Ubuntu Focal.

v4hn:
> Obviously, we will also need to wait for the next Eigen release 3.4. Not sure, it will ever make it into Ubuntu Focal.

I don't see why.  I believe we might be in for a long sit-in if we do wait for that.
For everyday-use alignment is no problem in our own builds. If users care about it for their setup,
they might as well compile Eigen from the repository and build MoveIt et. al. on top.

rhaschke:
Agreed. But then they are also free to enforce C++17 for the MoveIt build, without MoveIt itself enforcing it. So nothing to change at all - just a remark in some documentation?

gavanderhoorn:
> > Obviously, we will also need to wait for the next Eigen release 3.4. Not sure, it will ever make it into Ubuntu Focal.
> 
> I don't see why. I believe we might be in for a long sit-in if we do wait for that. For everyday-use alignment is no problem in our own builds. If users care about it for their setup, they might as well compile Eigen from the repository and build MoveIt et. al. on top.

you don't want to do this. As I mentioned earlier, linking software with different versions of Eigen is really not a good idea.

While MoveIt may be OK, having users install a version of Eigen which is not part of their OS but does show up in the default search paths is asking for trouble.

It *might* work if that newer version of Eigen is contained inside the ROS workspace, but even there it could become problematic, as users (re)compiling packages and linking with packages built by the buildfarm which includes the standard version of Eigen -- potentially with different alignment settings as well -- would still be a problem.

It's all technically possible, but from a support perspective it may not be feasible.

v4hn:
On Thu, Apr 16, 2020 at 06:57:01AM -0700, G.A. vd. Hoorn wrote:
> you don't want to do this.

**I** do want it, if I care about vector optimization.

> As I mentioned earlier, linking software with different versions of Eigen is really not a good idea.

Yes.

> It's all technically possible, but from a support perspective it may not be feasible.

Yes, but users looking for custom aligning do not need hand-holding (and will not get it from us).

moveit_core/utils/include/moveit/utils/aligned_make_shared.h
gleichdick: Thanks, good catch

gleichdick:
Okay, so no further actions needed?

moveit_core/collision_distance_field/include/moveit/collision_distance_field/collision_distance_field_types.h
```diff
@@ -69,6 +69,7 @@ enum CollisionType
 
 struct CollisionSphere
 {
+  using AlignedVector = std::vector<CollisionSphere, Eigen::aligned_allocator<CollisionSphere>>;
```
rhaschke, line 72: I suggest renaming `AlignedVector` to just `Vector`. Throughout the API we are not interested in the reason for this data type anymore. Here, however, a comment will be helpful.

moveit_experimental/collision_distance_field_ros/include/collision_distance_field_ros/collision_distance_field_ros_helpers.h
```diff
@@ -44,7 +44,7 @@ namespace collision_detection
 {
 static inline bool loadLinkBodySphereDecompositions(
     ros::NodeHandle& nh, const planning_models::RobotModelConstPtr& robot_model,
-    std::map<std::string, std::vector<collision_detection::CollisionSphere> >& link_body_spheres)
+    std::map<std::string, collision_detection::CollisionSphere::AlignedVector>& link_body_spheres)
```
rhaschke, line 47: Same here.

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
```diff
@@ -73,6 +73,7 @@ typedef BodyTypes::Type BodyType;
 struct Contact
 {
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+  using AlignedVector = std::vector<Contact, Eigen::aligned_allocator<Contact>>;
```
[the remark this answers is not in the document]
gleichdick: I agree, but if I understood @henningkayser correctly, he asked for more descriptive names in his first comment:
> In general, I think this goes into the right direction. I also like the use of alias types, but they should probably be a little more descriptive, ideally marking that those are/contain Eigen types.

v4hn:
> I agree, but if I understood @henningkayser correctly, he asked for more descriptive names in his first comment:

`Vector` would suffice, although it slightly obfuscates why you would *have* to write `Contact::Vector` over `vector<Contact>`.

> Okay, so no further actions needed?

Bjar, could you state own opinion on this please?
After all, you are the one working on this.
Do you need this alignment to work with C++14?
Did you try the Eigen built from source with C++17?
Does it actually solve the problem?

gleichdick:
Well, I did a whole source-install with `-std=c++17` **without this PR** (had to modify a lot of packages which hard-coded the C++ standard in their CMakeLists.txt). gcc will only respect the rightmost `-std` flag on the CLI, AFAIK.
The `Eigen::aligned_allocator<T>` was modified to be an alias to `std::allocator<T>`, so that `std::vector` and friends will use the default allocator even when told otherwise (like in the EigenSTL package). Some small changes at [octomap](https://github.com/OctoMap/octomap/pull/293), [eigenpy](https://github.com/stack-of-tasks/eigenpy/pull/186) and [ompl](https://github.com/ompl/ompl/pull/711) to fix compilation errors are already merged.

These Eigen assertations didn't show up, so I suspect that compiling with C++17 will indeed do it.
Unfortunately, the same unit tests as listed above still failed, so there might be some issues with not using the pre-built libraries from the repo.

Furthermore, on x86_64 dynamic memory is 16-byte aligned, so Eigen will be happy even without C++17 as long as AVX\* is not enabled.

IMHO, this PR can be closed. Thank you for your feedback.

rhaschke:
Thanks to you for your detailed analysis of the problem. Let's hope that the new Eigen lib is released soon... Closing here.

PR #2012, opened 2020-04-06
run shadow-fixed for main branches only

v4hn:
Followup to #2003 .

I disapprove of shadow-fixed as a required CI check. Yes, we want to know about problems with shadow-fixed early, but not at the cost of contributors who see their jobs failing for no reason.

Instead, this pull-request adds a conditional job for the main branches only, so we still notice breakage after merge.

Here's my playground travis [with the additional job](https://travis-ci.com/github/v4hn/moveit/builds/158638495) and [without it for other branches](https://travis-ci.com/github/v4hn/moveit/builds/158638459).

Receiving email notifications for this particular job, [@rhaschke was worried](https://github.com/ros-planning/moveit/pull/2003#issuecomment-609445786), is not a problem in my opinion. All main web page on GitHub related to MoveIt as well as its travis pages will show a red indicator if the main job breaks.

In addition to this job for new commits, [@gavanderhoorn](https://github.com/ros-planning/moveit/pull/2003#issuecomment-607216991) proposed to add a cron entry on travis.
This acts as a rebuild of the same commit and failure will show up the same way as for new commits.
We can just add the cron build on their web page once this is merged.
We can set it to `Do not run if there has been a build in the last 24 hours` and run it for the main development branches.

We could also further restrict the cron build jobs to shadow-fixed only, but I'm not sure that would play well with the associated update of the commit status. If some other job failed, cron might override the overall status if we do not run all jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,8 @@ jobs:
       env: TEST=clang-tidy-fix
            BEFORE_DOCKER_SCRIPT="source moveit_kinematics/test/test_ikfast_plugins.sh"
            CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-overloaded-virtual"
-           ROS_REPO=ros-shadow-fixed
+    - if: branch =~ /^(.*-devel|master)$/
+      env: ROS_REPO=ros-shadow-fixed
 
 before_script:
   - git clone -q --depth=1 https://github.com/ros-planning/moveit_ci.git .moveit_ci


rhaschke:
I very much like the idea of a conditional build (didn't know this exists). However, I still consider the (email) notification issue as unresolved.

> All main web page on GitHub related to MoveIt as well as its travis pages will show a red indicator if the main job breaks.

On MoveIt's github page, you need to scroll down two pages to see the red badge.
On Travis, you need to switch to the branches tab to see the failure.
So, at least for me, this means that I won't notice a failure (immediately).

v4hn:
> On MoveIt's github page, you need to scroll down two pages to see the red badge.

Right above the file list there is a summary of the last commit in the main branch.
It includes a CI symbol (checkmark or cross). It's not very prominent, granted, but at least I notice if it goes red.

> On Travis, you need to switch to the branches tab to see the failure.

If you open [a random pull-request CI](https://travis-ci.org/github/ros-planning/moveit/builds/671531643?utm_source=github_status&utm_medium=notification), the top badge, right next to the name of the repository indicates the status of the main branch. You see that every time you look at a CI report.

It's true that neither of these show you the CI result of `kinetic-devel`/`melodic-devel` though and a configuration for email notifications that does not spam everyone would be nice.
Maybe something like that can be constructed [with conditions too](https://config.travis-ci.com/ref/notifications/email), but testing a config like that is more involved than adding a conditional job.

Thank you for the approval, merging.

rhaschke:
> Right above the file list there is a summary of the last commit in the main branch.

Great, you are right! Never noticed this before (consciously).

> If you open a random pull-request CI, the top badge, right next to the name of the repository indicates the status of the main branch. 

Looks like, I'm stupid. I cannot see a badge indicating the status of the main branch.
Could you post a screenshot with a highlight, please?

v4hn:
default-push-build

rhaschke:
Thanks a lot. Now I know where to look in the future... I really didn't notice this badge beforehand.

v4hn:
Especially for you Robert, it might even make sense to configure a start page in your browser with the badges of your various repositories...

PR #2014, opened 2020-04-07
Revert "display robot state: do not clutter rviz with default robot o…

v4hn:
…n enable"

This reverts commit 73e6c409c097b5a87a1811b8f39263680fda1a9a.

The patch works together with af106bed1f1188ce85b4c7b6558c9d26ea773a65 which is not backported.

We could write a melodic version of this without the message change, but I don't think we need to backport this with changes.


v4hn:
As Simon and you both approved this reverts the regression in the upstream branch, I will merge it now.

I did not extend the patch to have the feature in melodic because I cannot easily test it over here without a lot of recompiling and it *is* quite a change of behavior users might be confused about. Feel free to implement the feature without the message change again for melodic.

PR #2022, opened 2020-04-13
fix param logic bug for self- and scene-collision proximity thresholds

JStech:
### Description

This is a fix to a logic error in my PR that was just merged (#2008). I had `!A && !B` where `!(A && B)` is correct.

### Checklist

### DIFF
--- a/moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
@@ -104,9 +104,15 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
                             "'self_collision_proximity_threshold' and 'scene_collision_proximity_threshold' "
                             "parameters. Please update the jogging yaml file.");
     if (!have_self_collision_proximity_threshold)
+    {
       ros_parameters_.self_collision_proximity_threshold;
+      have_self_collision_proximity_threshold = true;
+    }
     if (!have_scene_collision_proximity_threshold)
+    {
       ros_parameters_.scene_collision_proximity_threshold;
+      have_scene_collision_proximity_threshold = true;
+    }
   }
   error += !(have_self_collision_proximity_threshold && have_scene_collision_proximity_threshold);
   error += !rosparam_shortcuts::get("", n, parameter_ns + "/move_group_name", ros_parameters_.move_group_name);
--- a/moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
@@ -114,7 +114,8 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
       have_scene_collision_proximity_threshold = true;
     }
   }
-  error += !(have_self_collision_proximity_threshold && have_scene_collision_proximity_threshold);
+  error += !have_self_collision_proximity_threshold;
+  error += !have_scene_collision_proximity_threshold;
   error += !rosparam_shortcuts::get("", n, parameter_ns + "/move_group_name", ros_parameters_.move_group_name);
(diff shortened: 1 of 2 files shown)


moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
```diff
@@ -91,10 +91,10 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
...
-  if (!rosparam_shortcuts::get("", n, parameter_ns + "/self_collision_proximity_threshold",
-                               ros_parameters_.self_collision_proximity_threshold) &&
-      !rosparam_shortcuts::get("", n, parameter_ns + "/scene_collision_proximity_threshold",
-                               ros_parameters_.scene_collision_proximity_threshold))
+  if (!(rosparam_shortcuts::get("", n, parameter_ns + "/self_collision_proximity_threshold",
+                                ros_parameters_.self_collision_proximity_threshold) &&
+        rosparam_shortcuts::get("", n, parameter_ns + "/scene_collision_proximity_threshold",
+                                ros_parameters_.scene_collision_proximity_threshold)))
   {
     if (rosparam_shortcuts::get("", n, parameter_ns + "/collision_proximity_threshold",
```
henningkayser, line 99: I think it makes sense to always print the deprecation warning, even if both new parameters are defined. How about you move this parameter lookup before the new parameters and simply override the values if the new parameters are defined?

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
```diff
@@ -88,28 +88,27 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
...
-      ros_parameters_.scene_collision_proximity_threshold = ros_parameters_.self_collision_proximity_threshold;
-    }
-    else
-    {
-      error += 1;
-    }
+    ROS_WARN_NAMED(LOGNAME, "'collision_proximity_threshold' parameter is deprecated, and has been replaced by separate"
+                            "'self_collision_proximity_threshold' and 'scene_collision_proximity_threshold' "
+                            "parameters. Please update the jogging yaml file.");
+    if (!have_self_collision_proximity_threshold)
```
henningkayser, line 106: I guess you should set those flags to true then?

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
JStech: OK. Took me a bit to get all the cases how I like them, but here's where it stands:

- If the old parameter is defined, the deprecation warning is printed, regardless of whether zero, one, or both new parameters are defined;
- If the old parameter is defined and one or both new parameters are missing, the old parameter is used in their place;
- If, after all that, one or both new parameters is not defined, we shutdown via `rosparams_shortcuts::shutdownIfError`;
- Also, no error is printed if the old parameter is missing, but errors are printed if the new parameters are missing.

moveit_experimental/moveit_jog_arm/src/jog_interface_base.cpp
JStech: I don't follow. What flags? `have_self_collision_proximity_threshold` is set on line 94.

henningkayser: Yes, but the deprecated parameter is never used (if flags are false) since you then increase the error counter a couple lines below.

JStech: Ah yes, you're right. Fixed

JStech:
Yeah--I actually ran into this when running on the real robot, and initially fixed it there.

I also just ran the jog arm test locally a couple times with different parameters set in the config file to make sure it behaves as intended. I made one more small change, so that the shutdown error message reports the correct number of missing parameters when both thresholds are missing.

PR #2023, opened 2020-04-14
Allow dynamic adjustment of the low-pass filter

AndyZe:
So that you can adjust the degree of smoothing on the fly.

It also fixes a nasty bug where the low-pass-filtered positions weren't updated if jog commands stopped coming in. That fix is at jog_calcs.cpp L205

### DIFF
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
@@ -40,6 +40,7 @@
 
 // System
 #include <mutex>
+#include <boost/lockfree/queue.hpp>  // boost::lockfree::queue, share data in a lock-free way
 #include <thread>
 
 // Eigen
@@ -101,6 +102,9 @@ struct JogArmShared : public std::mutex
 
   // Stop jog loop threads - threads are not stopped by default
   std::atomic<bool> stop_requested{ false };
+
+  // Use a lock-free queue to update filter coefficients from other threads
+  boost::lockfree::queue<double> filter_coefficient_queue;
 };
 
 // ROS params to be read. See the yaml file in /config for a description of each.
--- a/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
+++ b/moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
@@ -129,7 +129,7 @@ protected:
   trajectory_msgs::JointTrajectory composeJointTrajMessage(sensor_msgs::JointState& joint_state) const;
 
   /** \brief Smooth position commands with a lowpass filter */
-  void lowPassFilterPositions(sensor_msgs::JointState& joint_state);
+  void lowPassFilterPositions(sensor_msgs::JointState& joint_state, JogArmShared& shared_variables);
 
   /** \brief Convert an incremental position command to joint velocity message */
   void calculateJointVelocities(sensor_msgs::JointState& joint_state, const Eigen::ArrayXd& delta_theta);
@@ -137,7 +137,7 @@ protected:
   /** \brief Convert joint deltas to an outgoing JointTrajectory command.
     * This happens for joint commands and Cartesian commands.
     */
-  bool convertDeltasToOutgoingCmd();
+  bool convertDeltasToOutgoingCmd(JogArmShared& shared_variables);
 
   /** \brief Gazebo simulations have very strict message timestamp requirements.
    * Satisfy Gazebo by stuffing multiple messages into one.
@@ -152,6 +152,13 @@ protected:
    * @param row_to_remove Dimension that will be allowed to drift, e.g. row_to_remove = 2 allows z-translation drift.
    */
   void removeDimension(Eigen::MatrixXd& matrix, Eigen::VectorXd& delta_x, unsigned int row_to_remove);
+
+  /**
+   * Update the filter coefficient to change the degree of smoothing.
+   *
+   * @param new_filter_coeff Must be greater than 1. A larger number will cause more smoothing but a longer delay.
+   */
+  void updateFilterCoeff(double new_filter_coeff);
 
   const moveit::core::JointModelGroup* joint_model_group_;
 
--- a/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
@@ -69,6 +69,10 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables)
   // Reset flags
   is_initialized_ = false;
 
+  // We expect a maximum of 1 value in the filter coefficient queue
+  // It is emptied as soon as a new value arrives
+  shared_variables.filter_coefficient_queue.reserve(2);
+
   // Wait for initial messages
   ROS_INFO_NAMED(LOGNAME, "jog_calcs_thread: Waiting for first joint msg.");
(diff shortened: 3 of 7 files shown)


AndyZe:
@tylerjw

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -822,4 +838,12 @@ void JogCalcs::removeDimension(Eigen::MatrixXd& jacobian, Eigen::VectorXd& delta
   jacobian.conservativeResize(num_rows, num_cols);
   delta_x.conservativeResize(num_rows);
 }
+
+void JogCalcs::updateFilterCoeff(double new_filter_coeff)
+{
+  for (size_t index = 0; index < position_filters_.size(); ++index)
```
JStech, line 844: A bit of a nitpick, but I think this could be a range-based for loop

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
```diff
@@ -101,6 +102,9 @@ struct JogArmShared : public std::mutex
 
   // Stop jog loop threads - threads are not stopped by default
   std::atomic<bool> stop_requested{ false };
+
+  // Use a lock-free queue to update filter coefficients from other threads
+  boost::lockfree::queue<double> filter_coefficient_queue;
```
tylerjw, line 107: Suggest using `spsc_queue` (https://www.boost.org/doc/libs/1_72_0/doc/html/boost/lockfree/spsc_queue.html) if this is only published to by one thread and read from another for a minor performance benefit over queue.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -129,15 +129,15 @@ class JogCalcs
   trajectory_msgs::JointTrajectory composeJointTrajMessage(sensor_msgs::JointState& joint_state) const;
 
   /** \brief Smooth position commands with a lowpass filter */
-  void lowPassFilterPositions(sensor_msgs::JointState& joint_state);
+  void lowPassFilterPositions(sensor_msgs::JointState& joint_state, JogArmShared& shared_variables);
```
tylerjw, line 132: [documentation] update docstring with parameter description.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -129,15 +129,15 @@ class JogCalcs
...
 
   /** \brief Convert an incremental position command to joint velocity message */
   void calculateJointVelocities(sensor_msgs::JointState& joint_state, const Eigen::ArrayXd& delta_theta);
 
   /** \brief Convert joint deltas to an outgoing JointTrajectory command.
     * This happens for joint commands and Cartesian commands.
     */
-  bool convertDeltasToOutgoingCmd();
+  bool convertDeltasToOutgoingCmd(JogArmShared& shared_variables);
```
tylerjw, line 140: [documentation] update docstring with parameter description.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -413,8 +421,16 @@ void JogCalcs::insertRedundantPointsIntoTrajectory(trajectory_msgs::JointTraject
...
-void JogCalcs::lowPassFilterPositions(sensor_msgs::JointState& joint_state)
+void JogCalcs::lowPassFilterPositions(sensor_msgs::JointState& joint_state, JogArmShared& shared_variables)
 {
+  // Check if a new filter coefficient was supplied
+  if (!shared_variables.filter_coefficient_queue.empty())
+  {
+    double new_coefficient;
+    shared_variables.filter_coefficient_queue.pop(new_coefficient);
+    updateFilterCoeff(new_coefficient);
+  }
```
tylerjw, line 432: [nit] this should probably pop all new values to get the last one sent.  To do that check the size of the queue and pop that number of times.  Even though this is unlikely, it is probably the behavior you want?

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -59,19 +59,22 @@ LowPassFilter::LowPassFilter(double low_pass_filter_coeff)
...
-                 LOGNAME);
-  ROS_ASSERT_MSG(low_pass_filter_coeff >= 1., "%s: Filter coefficient < 1. makes the lowpass filter unstable", LOGNAME);
+  if (low_pass_filter_coeff < 1)
+  {
+    ROS_WARN_STREAM_NAMED(LOGNAME, "Filter coefficient < 1. makes the lowpass filter unstable. "
+                                   "Setting to default of 10.");
+    low_pass_filter_coeff = 10;
+    scale_term_ = 1. / (1. + low_pass_filter_coeff);
+    feedback_term_ = 1. - low_pass_filter_coeff;
+  }
```
tylerjw, line 69: [nit] I like the softer error handling here, however I suggest making that an error instead of warn log to make it clearer to the users that the parameter that was set didn't take.

moveit_experimental/moveit_jog_arm/src/low_pass_filter.cpp
```diff
@@ -98,4 +101,17 @@ double LowPassFilter::filter(double new_measurement)
...
+void LowPassFilter::updateFilterCoeff(double new_filter_coeff)
+{
+  if (new_filter_coeff >= 1)
+  {
+    scale_term_ = 1. / (1. + new_filter_coeff);
+    feedback_term_ = 1. - new_filter_coeff;
+  }
+  else
+  {
+    ROS_WARN_STREAM_NAMED(LOGNAME, "Filter coefficient < 1. makes the lowpass filter unstable. Ignoring request.");
```
tylerjw, line 114: This behavior is different from what is in the constructor.  I suggest just calling this from the constructor so there is only one behavior and move the logic from the constructor here.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
jliukkonen: The idea is to have maybe typically one thread adjusting the filter coefficient in a thread-safe manner? Wouldn't it be much easier to just modify the code a bit to make `JogCalcs::updateFilterCoeff` thread-safe and allow threads to use that directly? If you aren't consuming the filter coefficients at the same rate as you are producing them or faster, the queue size will explode unless you set a fixed size for it.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -413,8 +421,16 @@ void JogCalcs::insertRedundantPointsIntoTrajectory(trajectory_msgs::JointTraject
   }
 }
 
-void JogCalcs::lowPassFilterPositions(sensor_msgs::JointState& joint_state)
+void JogCalcs::lowPassFilterPositions(sensor_msgs::JointState& joint_state, JogArmShared& shared_variables)
 {
+  // Check if a new filter coefficient was supplied
+  if (!shared_variables.filter_coefficient_queue.empty())
```
jliukkonen, line 427: Not sure whether it could be an issue here but calling `empty` on a lockfree queue isn't accurate in general. Boost documentation has the following note: 

> The result is only accurate, if no other thread modifies the queue. Therefore it is rarely practical to use this value in program logic.

AndyZe:
I'm gonna close this PR for now. My external motivation to get it done is gone. But, I appreciate the reviewer comments, it will probably come back at some point.

Issue #2025, opened 2020-04-16
CollisionObject shapes and subframes should be defined in an object's reference frame


felixvd:
When I am placing objects in the scene, I find it extremely cumbersome to transform all of their shapes and subframes to the position I want the object to be in. It's really something that the user shouldn't have to take care of. My current workaround is to publish a temporary TF frame for the object, which is awkward, requires waiting for the frames to propagate, and isn't very stable.

I would like to replace the `header` object in the [CollisionObject message](http://docs.ros.org/melodic/api/moveit_msgs/html/msg/CollisionObject.html) to be a `PoseStamped` object called `object_pose`. All the shapes and subframes would be defined relative to that pose.

The current behavior would be easy to replicate by defining a neutral `PoseStamped` with the same `header`. With the change, it would be easier to shift the whole object, and in the future, attaching the object to a scene graph would be simple.

Does anyone see a problem with this design direction? I may start developing this soon otherwise.

rhaschke:
I can understand your frustration and I like your proposal. Actually, MoveIt's internal representation of the collision object should be changed as well, to use this pose. Currently, the "pose" of the object is extracted from the pose of its first shape, which is really awkward:
https://github.com/ros-planning/moveit/blob/4d3062164bf03ebb74c57d5eba0e3e50438481ba/moveit_core/collision_detection/src/world.cpp#L172-L174

By the way, this code doesn't check anymore that shape_poses_ isn't empty!

v4hn:
Sounds good to me.
But we probably want the message change to get in for noetic, otherwise I'm not sure this would ever be released...
So you better speed up :sunglasses: 

> By the way, this code doesn't check anymore that shape_poses_ isn't empty.

True, where's the patch? :stuck_out_tongue: 
Or do you want all three of us to file one in parallel?

rhaschke:
No time for a patch today. I was hoping for Felix...

felixvd:
Like [this](https://github.com/felixvd/moveit/tree/check-for-empty-shapes)?

I would try to get on this first thing and propose something within this month.

rhaschke:
No, more like this:
`if (it != objects_.end() && !it->second->shapes_.empty())`

v4hn:
... You know if you keep replying to everything Robert, you are not really saving time in contrast to just writing the patch.
 I just filed #2026.

felixvd:
[...]

PR #2026, opened 2020-04-16
collision world: check for empty shapes vector before access

v4hn:
See https://github.com/ros-planning/moveit/issues/2025#issuecomment-614455384

### DIFF
--- a/moveit_core/collision_detection/src/world.cpp
+++ b/moveit_core/collision_detection/src/world.cpp
@@ -168,10 +168,15 @@ const Eigen::Isometry3d& World::getTransform(const std::string& name) const
 
 const Eigen::Isometry3d& World::getTransform(const std::string& name, bool& frame_found) const
 {
+  // assume found
   frame_found = true;
+
   std::map<std::string, ObjectPtr>::const_iterator it = objects_.find(name);
   if (it != objects_.end())
-    return it->second->shape_poses_[0];
+  {
+    if (!it->second->shape_poses_.empty())
+      return it->second->shape_poses_[0];
+  }
   else  // Search within subframes
   {
     for (const std::pair<const std::string, ObjectPtr>& object : objects_)
@@ -186,6 +191,7 @@ const Eigen::Isometry3d& World::getTransform(const std::string& name, bool& fram
     }
   }
 
+  // we need a persisting isometry for the API
   static const Eigen::Isometry3d IDENTITY_TRANSFORM = Eigen::Isometry3d::Identity();
   frame_found = false;
   return IDENTITY_TRANSFORM;
--- a/moveit_core/collision_detection/src/world.cpp
+++ b/moveit_core/collision_detection/src/world.cpp
@@ -143,7 +143,7 @@ bool World::knowsTransform(const std::string& name) const
   std::map<std::string, ObjectPtr>::const_iterator it = objects_.find(name);
   if (it != objects_.end())
     // only accept object name as frame if it is associated to a unique shape
-    return it->second->shape_poses_.size() == 1;
+    return !it->second->shape_poses_.empty();
   else  // Then objects' subframes
   {
     for (const std::pair<const std::string, ObjectPtr>& object : objects_)


moveit_core/collision_detection/src/world.cpp
```diff
@@ -186,6 +191,7 @@ const Eigen::Isometry3d& World::getTransform(const std::string& name, bool& fram
     }
   }
 
+  // we need a persisting isometry for the API
```
felixvd, line 194:
```suggestion
  // Return an identity transform so there is a persisting isometry object for the API
```
nit

v4hn: Don't like the suggestion because we *return* a reference, not the transform.

v4hn:
I believe you are talking about [this access operator](https://github.com/felixvd/moveit/commit/48e28d0a2e05d5ea08179755429e695c0cb31fdf#diff-9582c92ebb2c026094ca2349707db3a4R205)?

This is actually guarded by ensuring there exists a shape for this index.
I believe the structure holds the invariant `shapes_.size() == shape_poses_.size()`.

moveit_core/collision_detection/src/world.cpp
```diff
@@ -174,7 +174,7 @@ const Eigen::Isometry3d& World::getTransform(const std::string& name, bool& fram
   std::map<std::string, ObjectPtr>::const_iterator it = objects_.find(name);
   if (it != objects_.end())
   {
-    if (!it->second->shape_poses_.empty())
+    if (it->second->shape_poses_.size() == 1)
```
rhaschke, line 177: We should check for `size == 1`, because of this:
https://github.com/ros-planning/moveit/blob/4d3062164bf03ebb74c57d5eba0e3e50438481ba/moveit_core/collision_detection/src/world.cpp#L146

v4hn: Actually back when we last discussed this, we kept the case with multiple `shape_poses_` a gray area on purpose. If you ask for the frame you get the first one, if you ask whether the frame exists, it says no.
I don't have a use-case for objects with multiple shapes, so I don't care too much.

It does seem uncalled for to change this semantic in a segfault fix though. ;)

The proper way to fix this is @felixvd proposal to have a global pose for the object...

rhaschke: I cannot believe we agreed on inconsistent behavior on purpose. As both fixes are strongly related, I'm fine to have them in a single PR. IMHO it doesn't make sense to split this tiny PR up into even smaller ones.

v4hn:
> I cannot believe we agreed on inconsistent behavior on purpose.

I would be very surprised if you were involved in that discussion.

As far as I remember the rationale was to avoid breaking code that silently relied on the Transforms of objects with multiple shapes and never checked `knowsTransform` - which is quite a common thing to do if you just added the object a few lines above.

After all, it is an arbitrary and restrictive decision to stop treating the first transform of an object as its frame.

rhaschke:
> After all, it is an arbitrary and restrictive decision to stop treating the first transform of an object as its frame.

I agree. We should change both checks to `!it->second->shape_poses_.empty()` then.
My key point was to make it consistent :wink:

v4hn:
> I agree. We should change both checks to `!it->second->shape_poses_.empty()` then.
> My key point was to make it consistent :wink:

I believe the current behavior was present in the `PlanningScene` at least since 2013 (fc1a7e66e14b356d009fdb22549e5245917eb1af).
There has been a warning since then if you access the transform of an object with multiple shapes, but it still works.
We discussed changing it at some point, but did not back then.

I'm fine with clearing up the semantics to define the pose as the pose *of the first shape* altogether.
One reason why I did not do that before is that the user does not always have full control over which shape is the first,
e.g., shapes are generated in the planning scene by type array, so one shape type will always go first.

rhaschke:
If @felixvd is going to change the semantics anyway, I think we don't need to further discuss this topic here...

v4hn:
> If @felixvd is going to change the semantics anyway, I think we don't need to further discuss this topic here...

I adjusted knowsTransform (which by the way makes it consistent with `RobotState`s code.
Feel free to merge once CI finished.

PR #2030, opened 2020-04-18
commander: python3 fix

v4hn:
There is certainly more, but this allows to `use <group>\ngo <named_state>`


v4hn:
> Thank you for working on python3 compatibility. +1 +1 +1

I don't have much of a choice, running on a python-3 based system :dagger: 

> [tangent] Do you know if we could have our python tests run with python3 to find areas in our python code that need updated and keep them compatible?

Once we have noetic CI set up, it's going to test for python 3 automatically.
That again still depends on a number of packages to be released for noetic (not just MoveIt).

PR #2031, opened 2020-04-19
[test] move_group pick place test

tylerjw:
### Description

Another test for basic functionality from our tutorials to protect the experience of new users.

There is a bunch of hard coded strings and magic numbers for the geometry for this test.  I would appreciate any way to reduce the complexity of the test code.  Most of the code for this was copied from the [tutorial](https://github.com/ros-planning/moveit_tutorials/blob/master/doc/pick_place/src/pick_place_tutorial.cpp) that does the same thing.  @mlautman do you see anyway to reduce the complexity of this test and the tutorial?

The actual test itself is just the return values of the `pick` and `place` method calls in the move_group.  Another potential improvement to this test would be to get the new location of the object that was placed and validate it is in the correct new position.  I'm not sure what the return value of the move_group methods actually mean (I'm assuming it means the trajectory ran without error) and if testing the location of the moved object would be redundant.

The last thing is the tutorial sets the planning time to 45s so I did that here too.  I'm assuming needs to be long because on a debug build (what the codecov build is) planning takes a long time.  Is there a good way to know how long to set this to avoid the test being flaky?

Another thing to point out is that I'm using `moveit_resources` but I have tested this with my PR to update it to the current state of the upstream `panda_moveit_config` which is what the tutorial uses so this test should work after my change to `moveit_resources` is updated.

### Checklist
- [x] Moves the CodeCov report in the right direction :arrow_up: :arrow_up: :arrow_up:

### DIFF
--- a/moveit_ros/planning_interface/test/move_group_pick_place_test.cpp
+++ b/moveit_ros/planning_interface/test/move_group_pick_place_test.cpp
@@ -1,0 +1,290 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2020, Tyler Weaver
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of PickNik Robotics nor the
+*     names of its contributors may be used to endorse or promote
+*     products derived from this software without specific prior
+*     written permission.
+*
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+/* Author: Tyler Weaver */
+
+/* These integration tests are based on the tutorials for using move_group to do a pick and place:
+ * https://ros-planning.github.io/moveit_tutorials/doc/pick_place/pick_place_tutorial.html
+ */
+
+// C++
+#include <vector>
+
+// ROS
+#include <ros/ros.h>
+
+// The Testing Framework and Utils
+#include <gtest/gtest.h>
+
+// MoveIt
+#include <moveit/planning_scene_interface/planning_scene_interface.h>
+#include <moveit/move_group_interface/move_group_interface.h>
+
+// TF2
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+
+static const std::string PLANNING_GROUP = "panda_arm";
+constexpr double PLANNING_TIME_S = 45.0;
+constexpr double MAX_VELOCITY_SCALE = 1.0;
+constexpr double MAX_ACCELERATION_SCALE = 1.0;
+
+class PickPlaceTestFixture : public ::testing::Test
+{
+public:
+  void SetUp() override
+  {
+    nh_ = ros::NodeHandle("/move_group_pick_place_test");
+    move_group_ = std::make_shared<moveit::planning_interface::MoveGroupInterface>(PLANNING_GROUP);
+
+    // set velocity and acceleration scaling factors (full speed)
+    move_group_->setMaxVelocityScalingFactor(MAX_VELOCITY_SCALE);
+    move_group_->setMaxAccelerationScalingFactor(MAX_ACCELERATION_SCALE);
+
+    // allow more time for planning
+    move_group_->setPlanningTime(PLANNING_TIME_S);
+  }
+
+protected:
+  ros::NodeHandle nh_;
+  moveit::planning_interface::MoveGroupInterfacePtr move_group_;
+  moveit::planning_interface::PlanningSceneInterface planning_scene_interface_;
+};
+
+TEST_F(PickPlaceTestFixture, PickPlaceTest)
+{
+  // Create vector to hold 3 collision objects.
+  std::vector<moveit_msgs::CollisionObject> collision_objects;
+  collision_objects.resize(3);
+
+  // Add the first table where the cube will originally be kept.
+  collision_objects[0].id = "table1";
+  collision_objects[0].header.frame_id = "panda_link0";
+
+  /* Create identity rotation quaternion */
+  tf2::Quaternion zero_orientation;
+  zero_orientation.setRPY(0, 0, 0);
+  geometry_msgs::Quaternion zero_orientation_msg = tf2::toMsg(zero_orientation);
+
+  /* Define the primitive and its dimensions. */
+  collision_objects[0].primitives.resize(1);
+  collision_objects[0].primitives[0].type = collision_objects[0].primitives[0].BOX;
+  collision_objects[0].primitives[0].dimensions.resize(3);
+  collision_objects[0].primitives[0].dimensions[0] = 0.2;
+  collision_objects[0].primitives[0].dimensions[1] = 0.4;
+  collision_objects[0].primitives[0].dimensions[2] = 0.4;
+
+  /* Define the pose of the table. */
+  collision_objects[0].primitive_poses.resize(1);
+  collision_objects[0].primitive_poses[0].position.x = 0.5;
+  collision_objects[0].primitive_poses[0].position.y = 0;
+  collision_objects[0].primitive_poses[0].position.z = 0.2;
+  collision_objects[0].primitive_poses[0].orientation = zero_orientation_msg;
+
+  collision_objects[0].operation = collision_objects[0].ADD;
+
+  // Add the second table where we will be placing the cube.
+  collision_objects[1].id = "table2";
+  collision_objects[1].header.frame_id = "panda_link0";
+
+  /* Define the primitive and its dimensions. */
+  collision_objects[1].primitives.resize(1);
+  collision_objects[1].primitives[0].type = collision_objects[1].primitives[0].BOX;
+  collision_objects[1].primitives[0].dimensions.resize(3);
+  collision_objects[1].primitives[0].dimensions[0] = 0.4;
+  collision_objects[1].primitives[0].dimensions[1] = 0.2;
+  collision_objects[1].primitives[0].dimensions[2] = 0.4;
+
+  /* Define the pose of the table. */
+  collision_objects[1].primitive_poses.resize(1);
+  collision_objects[1].primitive_poses[0].position.x = 0;
+  collision_objects[1].primitive_poses[0].position.y = 0.5;
+  collision_objects[1].primitive_poses[0].position.z = 0.2;
+  collision_objects[1].primitive_poses[0].orientation = zero_orientation_msg;
+
+  collision_objects[1].operation = collision_objects[1].ADD;
+
+  // Define the object that we will be manipulating
+  collision_objects[2].header.frame_id = "panda_link0";
+  collision_objects[2].id = "object";
+
+  /* Define the primitive and its dimensions. */
+  collision_objects[2].primitives.resize(1);
+  collision_objects[2].primitives[0].type = collision_objects[1].primitives[0].BOX;
+  collision_objects[2].primitives[0].dimensions.resize(3);
+  collision_objects[2].primitives[0].dimensions[0] = 0.02;
+  collision_objects[2].primitives[0].dimensions[1] = 0.02;
+  collision_objects[2].primitives[0].dimensions[2] = 0.2;
+
+  /* Define the pose of the object. */
+  collision_objects[2].primitive_poses.resize(1);
+  collision_objects[2].primitive_poses[0].position.x = 0.5;
+  collision_objects[2].primitive_poses[0].position.y = 0;
+  collision_objects[2].primitive_poses[0].position.z = 0.5;
+  collision_objects[2].primitive_poses[0].orientation = zero_orientation_msg;
+
+  collision_objects[2].operation = collision_objects[2].ADD;
+
+  planning_scene_interface_.applyCollisionObjects(collision_objects);
+
+  // Create a vector of grasps to be attempted, currently only creating single grasp.
+  // This is essentially useful when using a grasp generator to generate and test multiple grasps.
+  std::vector<moveit_msgs::Grasp> grasps;
+  grasps.resize(1);
+
+  // Setting grasp pose
+  // ++++++++++++++++++++++
+  // This is the pose of panda_link8. |br|
+  // Make sure that when you set the grasp_pose, you are setting it to be the pose of the last link in
+  // your manipulator which in this case would be `"panda_link8"` You will have to compensate for the
+  // transform from `"panda_link8"` to the palm of the end effector.
+  grasps[0].grasp_pose.header.frame_id = "panda_link0";
+  tf2::Quaternion grasp_orientation;
+  grasp_orientation.setRPY(-M_PI / 2, -M_PI / 4, -M_PI / 2);
+  grasps[0].grasp_pose.pose.orientation = tf2::toMsg(grasp_orientation);
+  grasps[0].grasp_pose.pose.position.x = 0.415;
+  grasps[0].grasp_pose.pose.position.y = 0;
+  grasps[0].grasp_pose.pose.position.z = 0.5;
+
+  // Setting pre-grasp approach
+  // ++++++++++++++++++++++++++
+  /* Defined with respect to frame_id */
+  grasps[0].pre_grasp_approach.direction.header.frame_id = "panda_link0";
+  /* Direction is set as positive x axis */
+  grasps[0].pre_grasp_approach.direction.vector.x = 1.0;
+  grasps[0].pre_grasp_approach.min_distance = 0.095;
+  grasps[0].pre_grasp_approach.desired_distance = 0.115;
+
+  // Setting post-grasp retreat
+  // ++++++++++++++++++++++++++
+  /* Defined with respect to frame_id */
+  grasps[0].post_grasp_retreat.direction.header.frame_id = "panda_link0";
+  /* Direction is set as positive z axis */
+  grasps[0].post_grasp_retreat.direction.vector.z = 1.0;
+  grasps[0].post_grasp_retreat.min_distance = 0.1;
+  grasps[0].post_grasp_retreat.desired_distance = 0.25;
+
+  // Setting posture of eef before grasp
+  // +++++++++++++++++++++++++++++++++++
+  /* Add both finger joints of panda robot. */
+  grasps[0].pre_grasp_posture.joint_names.resize(2);
+  grasps[0].pre_grasp_posture.joint_names[0] = "panda_finger_joint1";
+  grasps[0].pre_grasp_posture.joint_names[1] = "panda_finger_joint2";
+
+  /* Set them as open, wide enough for the object to fit. */
+  grasps[0].pre_grasp_posture.points.resize(1);
+  grasps[0].pre_grasp_posture.points[0].positions.resize(2);
+  grasps[0].pre_grasp_posture.points[0].positions[0] = 0.04;
+  grasps[0].pre_grasp_posture.points[0].positions[1] = 0.04;
+  grasps[0].pre_grasp_posture.points[0].time_from_start = ros::Duration(0.5);
+
+  // Setting posture of eef during grasp
+  // +++++++++++++++++++++++++++++++++++
+  /* Add both finger joints of panda robot and set them as closed. */
+  grasps[0].grasp_posture = grasps[0].pre_grasp_posture;
+  grasps[0].grasp_posture.points[0].positions[0] = 0.00;
+  grasps[0].grasp_posture.points[0].positions[1] = 0.00;
+
+  // Set support surface as table1.
+  move_group_->setSupportSurfaceName("table1");
+  // Call pick to pick up the object using the grasps given
+  ASSERT_EQ(move_group_->pick("object", grasps), moveit::planning_interface::MoveItErrorCode::SUCCESS);
+
+  // Ideally, you would create a vector of place locations to be attempted although in this example, we only create
+  // a single place location.
+  std::vector<moveit_msgs::PlaceLocation> place_location;
+  place_location.resize(1);
+
+  // Setting place location pose
+  // +++++++++++++++++++++++++++
+  place_location[0].place_pose.header.frame_id = "panda_link0";
+  tf2::Quaternion place_orientation;
+  place_orientation.setRPY(0, 0, M_PI / 2);
+  place_location[0].place_pose.pose.orientation = tf2::toMsg(place_orientation);
+
+  /* For place location, we set the value to the exact location of the center of the object. */
+  place_location[0].place_pose.pose.position.x = 0;
+  place_location[0].place_pose.pose.position.y = 0.5;
+  place_location[0].place_pose.pose.position.z = 0.5;
+
+  // Setting pre-place approach
+  // ++++++++++++++++++++++++++
+  /* Defined with respect to frame_id */
+  place_location[0].pre_place_approach.direction.header.frame_id = "panda_link0";
+  /* Direction is set as negative z axis */
+  place_location[0].pre_place_approach.direction.vector.z = -1.0;
+  place_location[0].pre_place_approach.min_distance = 0.095;
+  place_location[0].pre_place_approach.desired_distance = 0.115;
+
+  // Setting post-grasp retreat
+  // ++++++++++++++++++++++++++
+  /* Defined with respect to frame_id */
+  place_location[0].post_place_retreat.direction.header.frame_id = "panda_link0";
+  /* Direction is set as negative y axis */
+  place_location[0].post_place_retreat.direction.vector.y = -1.0;
+  place_location[0].post_place_retreat.min_distance = 0.1;
+  place_location[0].post_place_retreat.desired_distance = 0.25;
+
+  // Setting posture of eef after placing object
+  // +++++++++++++++++++++++++++++++++++++++++++
+  /* Similar to the pick case */
+  /* Add both finger joints of panda robot. */
+  place_location[0].post_place_posture.joint_names.resize(2);
+  place_location[0].post_place_posture.joint_names[0] = "panda_finger_joint1";
+  place_location[0].post_place_posture.joint_names[1] = "panda_finger_joint2";
+
+  /* Set them as open, wide enough for the object to fit. */
+  place_location[0].post_place_posture.points.resize(1);
+  place_location[0].post_place_posture.points[0].positions.resize(2);
+  place_location[0].post_place_posture.points[0].positions[0] = 0.04;
+  place_location[0].post_place_posture.points[0].positions[1] = 0.04;
+  place_location[0].post_place_posture.points[0].time_from_start = ros::Duration(0.5);
+
+  // Set support surface as table2.
+  move_group_->setSupportSurfaceName("table2");
+  // Call place to place the object using the place locations given.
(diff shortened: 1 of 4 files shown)


tylerjw:
Sadness, this is failing on kinetic with a timeout and a ton this in the output of running the test:
```eigen: too many iterations in Jacobi transform.```

I tried increasing the length of the test but I don't like that solution even if it does work.  Has anyone seen this?  Any ideas how to troubleshoot this?

**Comment by v4hn** [comment]
On Sun, Apr 19, 2020 at 12:52:43AM -0700, Tyler Weaver wrote:
> Sadness, this is failing on kinetic with a timeout and a ton this in the output of running the test:
> ```eigen: too many iterations in Jacobi transform.```

This message has been mentioned in a number of issues and pull-requests over the years.

You could try to setup kinetic and debug it there by hand.

I'm not sure it's worth the effort though. Maybe we can just disable this new test in kinetic?

**Comment by tylerjw** [comment]
@v4hn I'll do that and get back to you.  It'd be nice to squash the kinetic bug too if possible.

One other note is that @henningkayser pointed out that there are talks about deprecating this pick and place interface in favor of something more powerful (MTC maybe).  I think it is still useful to have a test for it until we do that work though.  

The primary reason is I'm concerned about the experience of new users and I think that this is a likely interface for them to start playing with.  The tutorial is the first result to the google search "moveit pick and place".

**Comment by tylerjw** [comment]
I sunk a couple hours into this and don't feel any closer to a solution.  I can reproduce the error locally by checking out docker and running the test locally.  I'm pasting my set of commands here in case someone else with more knowledge of the issue would find these steps useful in debugging the bug in kinetic this exposed:

```bash
## start docker image
docker pull moveit/moveit:kinetic-ci
docker run -it moveit/moveit:kinetic-ci /bin/bash

## Inside docker container
apt-get update
apt-get dist-upgrade
apt-get install -y python-catkin-tools ssh git

## set enviroment variables
export ROS_DISTRO=kinetic
export TEST_BLACKLIST="moveit_ros_perception"

## Craete ROS workspace
CATKIN_WS=/root/ros_ws
mkdir -p ${CATKIN_WS}/src
cd ${CATKIN_WS}/src

## checkout moveit and dependencies library into workspace
cd ${CATKIN_WS}/src
wstool init .
git clone https://github.com/PickNikRobotics/moveit.git -b p/tylerjw/pick_place_test
wstool merge -t . moveit/moveit.rosinstall
wstool update

## Install dependencies
cd ${CATKIN_WS}/src
rosdep install -y -q -n --from-paths . --ignore-src --rosdistro ${ROS_DISTRO}

## Config and build
cd $CATKIN_WS
export CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls"
catkin config --extend /opt/ros/${ROS_DISTRO} --no-install --cmake-args -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS_RELEASE="-O3" -- --blacklist $TEST_BLACKLIST
catkin build --summarize

## build the tests
cd ${CATKIN_WS}
catkin build --summarize --make-args tests

## run the tests
cd ${CATKIN_WS}
catkin build --catkin-make-args run_tests -- --no-status --summarize

## view test results
catkin_test_results --verbose

## run my test that is failing
rostest moveit_ros_planning_interface move_group_pick_place_test.test --text
```

tylerjw:
More poking the bear I found that the move_group binary is crashing when I launch it.  Here is the valgrind output.  It looks to be something eigen related:

```
==14938== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==14938==  General Protection Fault
==14938==    at 0x7D8DD47: Eigen::internal::copy_using_evaluator_innervec_CompleteUnrolling<Eigen::internal::generic_dense_assignment_kernel<Eigen::internal::evaluator<Eigen::Matrix<double, 4, 4, 0, 4, 4> >, Eigen::internal::evaluator<Eigen::Matrix<double, 4, 4, 0, 4, 4> >, Eigen::internal::assign_op<double>, 0>, 0, 16>::run(Eigen::internal::generic_dense_assignment_kernel<Eigen::internal::evaluator<Eigen::Matrix<double, 4, 4, 0, 4, 4> >, Eigen::internal::evaluator<Eigen::Matrix<double, 4, 4, 0, 4, 4> >, Eigen::internal::assign_op<double>, 0>&) (in /root/ros_ws/devel/.private/moveit_core/lib/libmoveit_robot_state.so.1.0.1)
==14938==    by 0x9D7B48C: bodies::ConvexMesh::cloneAt(Eigen::Transform<double, 3, 1, 0> const&, double, double) const (in /root/ros_ws/devel/.private/geometric_shapes/lib/libgeometric_shapes.so)
==14938==    by 0x1FE926AE: point_containment_filter::ShapeMask::addShape(std::shared_ptr<shapes::Shape const> const&, double, double) (in /opt/ros/kinetic/lib/libmoveit_point_containment_filter.so.0.9.18)
==14938==    by 0x1F9FF1F9: occupancy_map_monitor::PointCloudOctomapUpdater::excludeShape(std::shared_ptr<shapes::Shape const> const&) (in /opt/ros/kinetic/lib/libmoveit_pointcloud_octomap_updater_core.so.0.9.18)
==14938==    by 0x50E5833: planning_scene_monitor::PlanningSceneMonitor::excludeRobotLinksFromOctree() (in /root/ros_ws/devel/.private/moveit_ros_planning/lib/libmoveit_planning_scene_monitor.so.1.0.1)
==14938==    by 0x50EA5AD: planning_scene_monitor::PlanningSceneMonitor::startWorldGeometryMonitor(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) (in /root/ros_ws/devel/.private/moveit_ros_planning/lib/libmoveit_planning_scene_monitor.so.1.0.1)
==14938==    by 0x4134A2: main (in /root/ros_ws/devel/.private/moveit_ros_move_group/lib/moveit_ros_move_group/move_group)
```

While it says that it dumped a core... even though I have core dumps turned on there is no core file produced.  If you run it in gdb it crashes before it ever reaches main.  This is something that is going bad while it is trying to load the executable.  I'm going to try to build with debug symbols to see if I can get more detail.

tylerjw:
I might be just complaining to a black hole... but this is the sort of bug you invite when you use dynamic memory everywhere (shared_ptr all the things).  :cry: 

My best guess so far is we are copying a shared pointer that doesn't have a valid control block.

tylerjw:
This might be a good clue for this one:

```
r<shapes::Shape const> const&) (in /opt/ros/kinetic/lib/libmoveit_pointcloud_octomap_updater_core.so.0.9.18)
==17790==    by 0x99F21C4: occupancy_map_monitor::OccupancyMapMonitor::excludeShape(std::shared_ptr<shapes::Shape const> const&) (occupancy_map_monitor.cpp:237)
```

Why would the kinetic build be using a debian of something that originates from this package and not the source built one.  That seems wrong, what could cause that?

tylerjw:
When I try to add the dependency to `moveit_ros_perception` to `moveit_ros_occupancy_map_monitor` I introduce a circular dependency.  Is there a good way to fix this problem so they can depend on eachother's source builds instead of having `moveit_ros_occupancy_map_monitor` depend on the debaian for `moveit_ros_preception` which is who-knows-how-old in kenetic.

The only way I can think to untangle this mess is to refactor the code so there isn't a circular dependency.  Since I don't have any experience with either of these packages does someone else have a better idea?

tylerjw:
Testing blacklist change to moveit_ci: https://github.com/ros-planning/moveit_ci/pull/111

tylerjw:
The `_preception` blacklist issue is no longer a problem because the failing test was fixed and therefore blacklist isn't needed.  That being said there is an unrelated move_group crash on kinetic.  I will report back if I find more information on this.

tylerjw:
I found our smoking gun.  The length of the array is 6 and the code is accessing the 6th element (one past the end of the array) of it (from current_state_monitor.cpp:450 with a ros warn added by me):

```c++
      std::vector<double> new_values(joint->getStateSpaceDimension());
      ROS_WARN_STREAM("new_values size: " << new_values.size());
      const moveit::core::LinkModel* link = joint->getChildLinkModel();
      if (link->jointOriginTransformIsIdentity())
        joint->computeVariablePositions(tf2::transformToEigen(transf), new_values.data());
```

This seems fine except `computeVariablePositions` takes a raw pointer to a double and has this code (floating_joint_model.cpp223):
```c++
void FloatingJointModel::computeVariablePositions(const Eigen::Isometry3d& transf, double* joint_values) const
{
  joint_values[0] = transf.translation().x();
  joint_values[1] = transf.translation().y();
  joint_values[2] = transf.translation().z();
  Eigen::Quaterniond q(transf.rotation());
  joint_values[3] = q.x();
  joint_values[4] = q.y();
  joint_values[5] = q.z();
  joint_values[6] = q.w();
}
```

The logs I get before the segfault have the length of that array at 6.
```
[ WARN] [1588755613.247947347]: new_values size: 6
```

rhaschke:
I think the culprit is that the FloatingJointModel reports the wrong state-space dimension (6 instead of 7):
https://github.com/ros-planning/moveit/blob/2bebf052a76b1dfe06eab3731bf7b460dd883563/moveit_core/robot_model/src/floating_joint_model.cpp#L193-L196

What is really weird, is that this bug was never experienced before... The above code line hasn't changed since the very beginnings of MoveIt...

We should run our CI tests with sanitizing enabled to spot such issues!
[ASan](https://github.com/google/sanitizers/wiki/AddressSanitizer) is a much better sanitizer than valgrind by the way :wink:

v4hn:
Why would the Floating joint have 7 DOF?
It represents an isometry.

tylerjw:
I don't know why it would be 7, if it should be 6 then we shouldn't write into the 7th position.

To me the interface design of `computeVariablePositions` seems like we set ourselves up to fail.

tylerjw:
How have we not discovered this before?  Maybe in most builds, it just harmlessly corrupts some memory that doesn't matter?

tylerjw:
If I change the magic number from 6 to 7... what other code depends on a 6 length type?

tylerjw:
> [ASan](https://github.com/google/sanitizers/wiki/AddressSanitizer) is a much better sanitizer than valgrind by the way

I'll look into that... ty for the tip.

tylerjw:
After searching through the code I now see that this is the only place it is being used to construct an array that is then passed into functions in the joint model.  I made some minor changes to the floating joint model that should fix that bug... now onto the next one.

tylerjw:
Here is a stack trace of the current issue here.  It looks like it is something in fcl:

```c++
move_group: /var/lib/jenkins/workspace/fcl0.5-pkg_builder-master-generic/repo/src/math/transform.cpp:90: void fcl::Quaternion3f::toRotation(fcl::Matrix3f&) const: Assertion `.99 < data [0]*data [0] + data [1]*data [1] + data [2]*data [2] + data [3]*data [3]' failed.

Thread 16 "move_group" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffb2aaa700 (LWP 25431)]
0x00007ffff52fa428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff52fa428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff52fc02a in __GI_abort () at abort.c:89
#2  0x00007ffff52f2bd7 in __assert_fail_base (fmt=<optimized out>, 
    assertion=assertion@entry=0x7fffea0fd878 ".99 < data [0]*data [0] + data [1]*data [1] + data [2]*data [2] + data [3]*data [3]", 
    file=file@entry=0x7fffea0fd818 "/var/lib/jenkins/workspace/fcl0.5-pkg_builder-master-generic/repo/src/math/transform.cpp", line=line@entry=90, 
    function=function@entry=0x7fffea0fd940 "void fcl::Quaternion3f::toRotation(fcl::Matrix3f&) const")
    at assert.c:92
#3  0x00007ffff52f2c82 in __GI___assert_fail (
    assertion=0x7fffea0fd878 ".99 < data [0]*data [0] + data [1]*data [1] + data [2]*data [2] + data [3]*data [3]", 
    file=0x7fffea0fd818 "/var/lib/jenkins/workspace/fcl0.5-pkg_builder-master-generic/repo/src/math/transform.cpp", line=90, 
    function=0x7fffea0fd940 "void fcl::Quaternion3f::toRotation(fcl::Matrix3f&) const") at assert.c:101
#4  0x00007fffe9e7eb0e in fcl::Quaternion3f::toRotation(fcl::Matrix3fX<fcl::details::Matrix3Data<double> >&) const () from /usr/lib/x86_64-linux-gnu/libfcl.so.0.5
#5  0x00007fffe9e7fdd3 in fcl::Transform3f::getRotationInternal() const ()
   from /usr/lib/x86_64-linux-gnu/libfcl.so.0.5
#6  0x00007fffe9e5f290 in fcl::details::boxBoxIntersect(fcl::Box const&, fcl::Transform3f const&, fcl::Box const&, fcl::Transform3f const&, std::vector<fcl::ContactPoint, std::allocator<fcl::ContactPoint> >*) () from /usr/lib/x86_64-linux-gnu/libfcl.so.0.5
#7  0x00007fffe9d1e5f0 in fcl::ShapeCollisionTraversalNode<fcl::Box, fcl::Box, fcl::GJKSolver_libccd>::leafTesting(int, int) const () from /usr/lib/x86_64-linux-gnu/libfcl.so.0.5
#8  0x00007fffe9bd33de in unsigned long fcl::ShapeShapeCollide<fcl::Box, fcl::Box, fcl::GJKSolver_libccd>(fcl::CollisionGeometry const*, fcl::Transform3f const&, fcl::CollisionGeometry const*, fcl::Transform3f const&, fcl::GJKSolver_libccd const*, fcl::CollisionRequest const&, fcl::CollisionResult&) ()
   from /usr/lib/x86_64-linux-gnu/libfcl.so.0.5
#9  0x00007fffe9e7e2f0 in unsigned long fcl::collide<fcl::GJKSolver_libccd>(fcl::CollisionGeometry const*, fcl::Transform3f const&, fcl::CollisionGeometry const*, fcl::Transform3f const&, fcl::GJKSolver_libccd const*, fcl::CollisionRequest const&, fcl::CollisionResult&) ()
   from /usr/lib/x86_64-linux-gnu/libfcl.so.0.5
#10 0x00007fffe9e7df33 in fcl::collide(fcl::CollisionObject const*, fcl::CollisionObject const*, fcl::CollisionRequest const&, fcl::CollisionResult&) () from /usr/lib/x86_64-linux-gnu/libfcl.so.0.5
#11 0x00007fffed96ef15 in collision_detection::collisionCallback (o1=0x7fffb40048b0, 
    o2=0x7fffa801c630, data=0x7fffb2aa74f0)
    at /root/ros_ws/src/moveit/moveit_core/collision_detection_fcl/src/collision_common.cpp:306
#12 0x00007fffea0d53e3 in fcl::details::dynamic_AABB_tree::collisionRecurse(fcl::NodeBase<fcl::AABB>*, fcl::CollisionObject*, void*, bool (*)(fcl::CollisionObject*, fcl::CollisionObject*, void*)) ()
   from /usr/lib/x86_64-linux-gnu/libfcl.so.0.5
#13 0x00007fffea0d9be2 in fcl::DynamicAABBTreeCollisionManager::collide(fcl::CollisionObject*, void*, bool (*)(fcl::CollisionObject*, fcl::CollisionObject*, void*)) const ()
   from /usr/lib/x86_64-linux-gnu/libfcl.so.0.5
#14 0x00007fffed98d16c in collision_detection::CollisionEnvFCL::checkRobotCollisionHelper (
    this=0x7fffa8005830, req=..., res=..., state=..., acm=0x7fffa4002b00)
    at /root/ros_ws/src/moveit/moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp:297
#15 0x00007fffed98cbdb in collision_detection::CollisionEnvFCL::checkRobotCollision (
    this=0x7fffa8005830, req=..., res=..., state=..., acm=...)
    at /root/ros_ws/src/moveit/moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp:269
#16 0x00007ffff21b2c47 in planning_scene::PlanningScene::checkCollision (this=0x7fffa8012b90, req=..., 
    res=..., robot_state=..., acm=...)
    at /root/ros_ws/src/moveit/moveit_core/planning_scene/src/planning_scene.cpp:507
#17 0x00007fffb3ce050d in planning_scene::PlanningScene::checkCollision (this=0x7fffa8012b90, req=..., 
    res=..., robot_state=..., acm=...)
    at /root/ros_ws/src/moveit/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
```

tylerjw:
The bug behaves the same with both the current version of the upstream `panda_moveit_config` and our released `moveit_resources` one in `kinetic`.  I was hoping this was just a config error.

Maybe the bug is in the libfcl that is in kinetic as the moveit code is from master and works in melodic?  Is there some way to get a newer version of libfcl for kinetic to test that theory?

tylerjw:
I found this: https://github.com/flexible-collision-library/fcl/issues/190 and I'm also getting the same error @davetcoleman  calls out in the output before the crash.  I don't know if it is related.

tylerjw:
I found a clue:

```c++
(gdb) print (*state.attached_body_map_.begin()->second)
$20 = {parent_link_model_ = 0x8a6910, id_ = "object", shapes_ = std::vector of length 1, capacity 1 = {
    std::shared_ptr (count 4, weak 2) 0x7fffcc004060}, 
  attach_trans_ = std::vector of length 1, capacity 1 = {{
      m_matrix = {<Eigen::PlainObjectBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<Eigen::MatrixBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<Eigen::DenseBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<Eigen::internal::special_scalar_op_base<Eigen::Matrix<double, 4, 4, 0, 4, 4>, double, double, Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 3>, false>> = {<Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 3>> = {<Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 1>> = {<Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 0>> = {<Eigen::EigenBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, m_storage = {m_data = {array = {
                -nan(0x8000000000000), -nan(0x8000000000000), -nan(0x8000000000000), 0, 
                -nan(0x8000000000000), -nan(0x8000000000000), -nan(0x8000000000000), 0, 
                -nan(0x8000000000000), -nan(0x8000000000000), -nan(0x8000000000000), 0, 
                0.00044243504144125811, 0.00035635795687904981, 0.084564638127313407, 
                1}}}}, <No data fields>}}}, touch_links_ = std::set with 3 elements = {
    [0] = "panda_hand", [1] = "panda_leftfinger", [2] = "panda_rightfinger"}, detach_posture_ = {
    header = {seq = 0, stamp = {<ros::TimeBase<ros::Time, ros::Duration>> = {sec = 0, 
          nsec = 0}, <No data fields>}, frame_id = ""}, 
    joint_names = std::vector of length 0, capacity 0, points = std::vector of length 0, capacity 0}, 
  global_collision_body_transforms_ = std::vector of length 1, capacity 1 = {{
      m_matrix = {<Eigen::PlainObjectBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<Eigen::MatrixBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<Eigen::DenseBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<Eigen::internal::special_scalar_op_base<Eigen::Matrix<double, 4, 4, 0, 4, 4>, double, double, Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 3>, false>> = {<Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 3>> = {<Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 1>> = {<Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 0>> = {<Eigen::EigenBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, m_storage = {m_data = {array = {
                -nan(0x8000000000000), -nan(0x8000000000000), -nan(0x8000000000000), 0, 
                -nan(0x8000000000000), -nan(0x8000000000000), -nan(0x8000000000000), 0, 
                -nan(0x8000000000000), -nan(0x8000000000000), -nan(0x8000000000000), 0, 
                0.49999990181087917, 3.0970443763039292e-07, 0.51923044559676901, 
                1}}}}, <No data fields>}}}, subframe_poses_ = std::map with 0 elements, 
  global_subframe_poses_ = std::map with 0 elements}
```

tylerjw:
When this pose input:

```c++
{position = {x = 0.5, y = 0, z = 0.5}, orientation = {x = 0, y = 0, z = 0, w = 0}}
```

is sent into `PlanningScene::poseMsgToEigen` we get this output:

```c++
{
  m_matrix = {<Eigen::PlainObjectBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<Eigen::MatrixBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<Eigen::DenseBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<Eigen::internal::special_scalar_op_base<Eigen::Matrix<double, 4, 4, 0, 4, 4>, double, double, Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 3>, false>> = {<Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 3>> = {<Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 1>> = {<Eigen::DenseCoeffsBase<Eigen::Matrix<double, 4, 4, 0, 4, 4>, 0>> = {<Eigen::EigenBase<Eigen::Matrix<double, 4, 4, 0, 4, 4> >> = {<No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, <No data fields>}, m_storage = {m_data = {array = {-nan(0x8000000000000), 
            -nan(0x8000000000000), -nan(0x8000000000000), 0, -nan(0x8000000000000), 
            -nan(0x8000000000000), -nan(0x8000000000000), 0, -nan(0x8000000000000), 
            -nan(0x8000000000000), -nan(0x8000000000000), 0, 0.5, 0, 0.5, 1}}}}, <No data fields>}}
```

I think all those `-nan` values look suspect to me.

tylerjw:
for more context this happens when you send this to `planning_scene::PlanningScene::processCollisionObjectAdd`

```c++
$12 = (const moveit_msgs::CollisionObject &) @0x7fffcc0016a0: {header = {seq = 0, 
    stamp = {<ros::TimeBase<ros::Time, ros::Duration>> = {sec = 0, nsec = 0}, <No data fields>}, 
    frame_id = "panda_link0"}, pose = {position = {x = 0, y = 0, z = 0}, orientation = {x = 0, y = 0, 
      z = 0, w = 0}}, id = "object", type = {key = "", db = ""}, 
  primitives = std::vector of length 1, capacity 1 = {{type = 1 '\001', 
      dimensions = std::vector of length 3, capacity 3 = {0.02, 0.02, 0.20000000000000001}}}, 
  primitive_poses = std::vector of length 1, capacity 1 = {{position = {x = 0.5, y = 0, z = 0.5}, 
      orientation = {x = 0, y = 0, z = 0, w = 0}}}, meshes = std::vector of length 0, capacity 0, 
  mesh_poses = std::vector of length 0, capacity 0, planes = std::vector of length 0, capacity 0, 
  plane_poses = std::vector of length 0, capacity 0, 
  subframe_names = std::vector of length 0, capacity 0, 
  subframe_poses = std::vector of length 0, capacity 0, operation = 0 '\000'}
```

tylerjw:
For more context, here is the code from this test that is triggering that:

```c++
  /* Define the primitive and its dimensions. */
  collision_objects[2].primitives.resize(1);
  collision_objects[2].primitives[0].type = collision_objects[1].primitives[0].BOX;
  collision_objects[2].primitives[0].dimensions.resize(3);
  collision_objects[2].primitives[0].dimensions[0] = 0.02;
  collision_objects[2].primitives[0].dimensions[1] = 0.02;
  collision_objects[2].primitives[0].dimensions[2] = 0.2;

  /* Define the pose of the object. */
  collision_objects[2].primitive_poses.resize(1);
  collision_objects[2].primitive_poses[0].position.x = 0.5;
  collision_objects[2].primitive_poses[0].position.y = 0;
  collision_objects[2].primitive_poses[0].position.z = 0.5;

  collision_objects[2].operation = collision_objects[2].ADD;

  planning_scene_interface_.applyCollisionObjects(collision_objects);
```

tylerjw:
To be clear, where it is barfing inside of fcl is when doing some operation on this object.  That is why I suspect it.

tylerjw:
I believe this is ready as it passes Travis.  there is a problem with the moveit_tutorials that I copied from I discovered the hard way.  I'll make a PR to fix that.  A default constructed quaternion message is invalid and can cause FCL on Kinetic to fail during collision checking.

@rhaschke

tylerjw:
In case anyone finds this in the future because you are seeing a ton of these errors and then a crash in FCL: `too many iterations in Jacobi transform`  Make sure you are not adding any objects to the planning scene with a default constructed rotation in the pose.  `{0, 0, 0, 0}` is not a valid pose and when it tries to normalize that quaternion you get a bunch of `nan` in the resulting transform.  FCL barfs on this when it is trying to do collision checking and you get the crash.

v4hn:
This is a problem as old as ROS and pops up in every student's first attempt to specify an orientation.

I did not look at the code, but MoveIt should validate quaternions and reject invalid objects.
There should be an open issue for this, because it clearly doesn't.

Please add a reference to the problematic interface and consider adding input verification there.

felixvd:
Yeah, it is a super old problem, and evidently everyone makes the mistake once in a while. I filed the PR already, it just didn't appear in the email thread you responded to: #2089

moveit_ros/planning_interface/test/move_group_pick_place_test.cpp
```diff
@@ -0,0 +1,297 @@
...
+  grasps[0].pre_grasp_posture.points.resize(1);
+  grasps[0].pre_grasp_posture.points[0].positions.resize(2);
+  grasps[0].pre_grasp_posture.points[0].positions[0] = 0.04;
+  grasps[0].pre_grasp_posture.points[0].positions[1] = 0.04;
+  grasps[0].pre_grasp_posture.points[0].time_from_start = ros::Duration(0.5);
+
+  // Setting posture of eef during grasp
+  // +++++++++++++++++++++++++++++++++++
+  /* Add both finger joints of panda robot. */
+  grasps[0].grasp_posture.joint_names.resize(2);
```
henningkayser, line 215: Seems like some duplicate code to me

moveit_ros/planning_interface/test/move_group_pick_place_test.cpp
```diff
@@ -0,0 +1,297 @@
...
+// MoveIt
+#include <moveit/planning_scene_interface/planning_scene_interface.h>
+#include <moveit/move_group_interface/move_group_interface.h>
+
+// TF2
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+
+static const std::string PLANNING_GROUP = "panda_arm";
+constexpr double PLANNING_TIME_S = 45.0;
+constexpr double MAX_VELOCITY_SCALE = 0.1;
```
henningkayser, line 60: We're not using sim time, right? Maybe set these to 1.0

moveit_ros/planning_interface/test/move_group_pick_place_test.cpp
tylerjw: addressed this, see latest changes

moveit_ros/planning_interface/test/move_group_pick_place_test.cpp
tylerjw: done, see latest changes

v4hn:
@tylerjw could you please cleanup the commit history, I would like to keep separate commits but you added fixups.

tylerjw:
@v4hn I rebased and separated the two logical changes in this PR into separate commits, the test, and the bug fix.  lmk if that is what you wanted.  I originally assumed this would be squashed, but I could see how separating these two things would make for back-porting them simpler if they were separate commits.

v4hn:
> Situations like this make me really miss working on a project with gerrit. Maybe sometime github will support a gerrit like workflow.

What exactly are you missing?

> separated the two logical changes in this PR into separate commits, the test, and the bug fix. lmk if that is what you wanted. I originally assumed this would be squashed, but I could see how separating these two things would make for back-porting them simpler

True, although I simply want to keep commits with new test cases and actual bug fixes apart, also to reference them in years to come.
Thanks, good to go after CI finished.

tylerjw:
> What exactly are you missing?

Gerrit made it seamless to keep your changes in a clean history.  Each PR can only be one commit that can be cleanly cherry-picked and when you push updates you force push.  However, the force push doesn't destroy the history of the PR as gerrit keeps a history of all your changes so it can be easily code reviewed and there is no need for fixup commits to make the review process easy.

With github you have to first do fixup commits until you are through code review so that the reviewers can see the new changes (force push destroys the history on github), and then once the reviewers are happy you can rebase and clean up the history of the commits.

v4hn:
> and there is no need for fixup commits to make the review process easy.

As a reviewer as well as a contributor I do not like fixup commits at all and do not encourage using them.
Having a "Address feedback" commit at the end of a clean history does not help with anything.
If you force-push a pr branch, your changes should address all raised points.
If anything is not addressed, I expect feedback in the corresponding review thread (which still exists).

@henningkayser: I explicitly requested two separate commits and you just merge via squash without a comment. -.-
Please read reviews before merging.

PR #2033, opened 2020-04-21
PlanningComponent: Load plan_request_params and use it when calling p…

JafarAbdi:
This PR do the following:

1- Load the `plan_request_params` parameters and use them as default when calling `PlanningComponent::plan()` if a parameter is not loaded it uses the previous default value
2- Constructs `MoveItCpp`'s node_handle_ member variable with the input nh
3- It uses delegating constructor when no `MoveItCpp` object is passed to `PlanningComponent`

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
+++ b/moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
@@ -113,11 +113,11 @@ public:
     double max_velocity_scaling_factor;
     double max_acceleration_scaling_factor;
 
-    void load(const ros::NodeHandle& nh, const std::string& default_planning_pipeline)
+    void load(const ros::NodeHandle& nh)
     {
       std::string ns = "plan_request_params/";
       nh.param(ns + "planner_id", planner_id, std::string(""));
-      nh.param(ns + "planning_pipeline", planning_pipeline, default_planning_pipeline);
+      nh.param(ns + "planning_pipeline", planning_pipeline, std::string(""));
       nh.param(ns + "planning_time", planning_attempts, 1);
       nh.param(ns + "planning_attempts", planning_time, 5.0);
       nh.param(ns + "max_velocity_scaling_factor", max_velocity_scaling_factor, 1.0);
--- a/moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
+++ b/moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
@@ -108,10 +108,21 @@ public:
   {
     std::string planner_id;
     std::string planning_pipeline;
-    size_t planning_attempts;
+    int planning_attempts;
     double planning_time;
     double max_velocity_scaling_factor;
     double max_acceleration_scaling_factor;
+
+    void load(const ros::NodeHandle& nh, const std::string& default_planning_pipeline)
+    {
+      std::string ns = "plan_request_params/";
+      nh.param(ns + "planner_id", planner_id, std::string(""));
+      nh.param(ns + "planning_pipeline", planning_pipeline, default_planning_pipeline);
+      nh.param(ns + "planning_time", planning_attempts, 1);
+      nh.param(ns + "planning_attempts", planning_time, 5.0);
+      nh.param(ns + "max_velocity_scaling_factor", max_velocity_scaling_factor, 1.0);
+      nh.param(ns + "max_acceleration_scaling_factor", max_acceleration_scaling_factor, 1.0);
+    }
   };
 
   /** \brief Constructor */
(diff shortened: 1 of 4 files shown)


moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -108,10 +108,21 @@ class PlanningComponent
   {
     std::string planner_id;
     std::string planning_pipeline;
-    size_t planning_attempts;
+    int planning_attempts;
```
JafarAbdi, line 111: I changed it because the type of `num_planning_attempts` in `moveit_msgs::MotionPlanRequest` is `int` and `NodeHandle::getParam` don't have an overload for `size_t`

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -108,10 +108,21 @@ class PlanningComponent
...
     double planning_time;
     double max_velocity_scaling_factor;
     double max_acceleration_scaling_factor;
+
+    void load(const ros::NodeHandle& nh, const std::string& default_planning_pipeline)
+    {
+      std::string ns = "plan_request_params/";
+      nh.param(ns + "planner_id", planner_id, std::string(""));
+      nh.param(ns + "planning_pipeline", planning_pipeline, default_planning_pipeline);
```
henningkayser, line 120: I think I would leave `default_planning_pipeline` empty by default and simply check for the MoveItCpp configuration.

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -57,9 +57,9 @@ MoveItCpp::MoveItCpp(const ros::NodeHandle& nh, const std::shared_ptr<tf2_ros::B
 {
 }
 
-MoveItCpp::MoveItCpp(const Options& options, const ros::NodeHandle& /*unused*/,
+MoveItCpp::MoveItCpp(const Options& options, const ros::NodeHandle& nh,
```
henningkayser, line 60: Is this related in any way? Also, the other constructor doesn't seem to initialize `node_handle_` this way

JafarAbdi: Yes, this node handle will be passed to planning components to load the parameters (the parameters are under MoveItCpp namespace

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
henningkayser: Any opinion on this? I think we should not introduce implicit depencencies in the configurations, so I would leave `planning_pipeline` empty just as `planner_id`

JafarAbdi: I made the default planning pipeline as an empty string because [this line](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/moveit_cpp/src/planning_component.cpp#L168) will print an error if it's not specified

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/planning_component.h
```diff
@@ -108,10 +108,21 @@ class PlanningComponent
...
+    void load(const ros::NodeHandle& nh)
+    {
+      std::string ns = "plan_request_params/";
+      nh.param(ns + "planner_id", planner_id, std::string(""));
+      nh.param(ns + "planning_pipeline", planning_pipeline, std::string(""));
+      nh.param(ns + "planning_time", planning_attempts, 1);
+      nh.param(ns + "planning_attempts", planning_time, 5.0);
+      nh.param(ns + "max_velocity_scaling_factor", max_velocity_scaling_factor, 1.0);
+      nh.param(ns + "max_acceleration_scaling_factor", max_acceleration_scaling_factor, 1.0);
+    }
```
jliukkonen, line 125: What about error checking in this function? The `param` method will return false on failure but there is no checking for that here. The default value is then utilized but wouldn't it be better to inform user that the parameter values they set are not being used? For example, would it be possible to typo the `max_velocity_scaling_factor`, default to 1.0 and have the arm move really fast?

henningkayser: I don't think error checking is required because of the default values. You could say that overriding these values is the optional behavior, so I don't think we need extensive user infos here. Sanity checks and warnings are issued by the planning pipeline in case something is wrong. In general, moveit_cpp is currently not meant to provide a fool-proof user interface, but I would agree that we could add more debugging output about configuration and checks.

PR #2035, opened 2020-04-24
[backport to melodic] Assistant: load additional file with kinematics parameters (#1997)

simonschmeisser:
This is a backport of #1997  to melodic-devel. I used
```
git cherry-pick 2a98624b26f7ad0be42c272e0787e538436bfedb
```
Is that correct like this? @v4hn


rhaschke:
Thanks. This looks good.

PR #2041, opened 2020-04-30
[ci] configure travis to report success before code-coverage finishes

tylerjw:
### Description

This should speed up builds by configuring travis to report success before code-coverage job finishes.  This shouldn't reduce the effectiveness of the testing as we are only using code-coverage for reporting and are not failing CI based on that report.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,9 @@ env:
     - WARNINGS_OK=false
 
 jobs:
+  fast_finish: true
   include:
     - env: TEST="clang-format catkin_lint"
-    - env: TEST=code-coverage
     - env: ROS_DISTRO=melodic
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,9 +22,9 @@ env:
     - WARNINGS_OK=false
 
 jobs:
+  fast_finish: true
   include:
     - env: TEST="clang-format catkin_lint"
-    - env: TEST=code-coverage
```
rhaschke, line 27: You need to keep the test definition here as well. That's a pitfall of non-intuitive behavior I have tapped in before as well...

tylerjw: good catch, put it back here.

v4hn:
Thanks @tylerjw .

PR #2045, opened 2020-05-04
clang-tidy fixes

rhaschke:
Due to a bug (https://github.com/ros-planning/moveit_ci/pull/110), clang-tidy was never applied to the header files. Now it is...


rhaschke:
I will redo this, applying individual fixes one-by-one to facilitate reviewing.

PR #2049, opened 2020-05-04
PlanningSceneDisplay speedup

simonschmeisser:
Alternative title: "MoveIt like it's 2020!"

This utilizes the `robot_state.is_diff` field in `PlanningScene` messages to send only joint state updates and teaches `PlanningSceneDisplay` to simply update the transforms in OGRE's scene graph instead of rebuilding the representation of both attached objects and collision objects. This reduces the time required for handling one update from 150-250ms down to 0.5ms in a simple scene typical for us. (one attached object (240kb), a couple more of those as collision objects, two boxes (few vertices) and  an octomap) (Note that updates where the scene geometry changes are still slow ...)

`AttachedObjects` are now attached to the corresponding link in OGRE as well, so their position no longer needs to be updated manually.

Furthermore this increases the default rate limit for joint state updates to ca 33hz, the publishing rate to 30hz and the incoming update limit in `PlanningSceneDisplay` to 100hz. This allows for smooth animations at 30fps

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -240,7 +240,7 @@ void PlanningSceneMonitor::initialize(const planning_scene::PlanningScenePtr& sc
 
   last_update_time_ = last_robot_motion_time_ = ros::Time::now();
   last_robot_state_update_wall_time_ = ros::WallTime::now();
-  dt_state_update_ = ros::WallDuration(0.1);
+  dt_state_update_ = ros::WallDuration(0.03);
 
   double temp_wait_time = 0.05;
 
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -410,7 +410,6 @@ void PlanningSceneMonitor::scenePublishingThread()
     }
     if (publish_msg)
     {
-      rate.reset();
       planning_scene_publisher_.publish(msg);
       if (is_full)
         ROS_DEBUG_NAMED(LOGNAME, "Published full planning scene: '%s'", msg.name.c_str());
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -370,6 +370,11 @@ void PlanningSceneMonitor::scenePublishingThread()
             if (octomap_monitor_)
               lock = octomap_monitor_->getOcTreePtr()->reading();
             scene_->getPlanningSceneDiffMsg(msg);
+            if (new_scene_update_ == UPDATE_STATE)
+            {
+              msg.robot_state.attached_collision_objects.clear();
+              msg.robot_state.is_diff = true;
+            }
           }
           boost::recursive_mutex::scoped_lock prevent_shape_cache_updates(shape_handles_lock_);  // we don't want the
                                                                                                  // transform cache to
--- a/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
+++ b/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
@@ -191,7 +191,10 @@ protected:
   RobotStateVisualizationPtr planning_scene_robot_;
   PlanningSceneRenderPtr planning_scene_render_;
 
+  // full update required
   bool planning_scene_needs_render_;
+  // or only the robot position (excluding attached object changes)
+  bool robot_state_needs_render_;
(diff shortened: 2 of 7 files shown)


simonschmeisser:
Well I think people expect smooth visualizations in 2020 ...

Try adding a `PlanningSceneDisplay` and a `RobotDisplay` or `TFDisplay`and compare the two to know what I mean. People with live capture and constantly updated octomaps can still lower that rate if necessary, all other will simply enjoy progress ;)

v4hn:
> Try adding a PlanningSceneDisplay and a RobotDisplay or TFDisplayand compare the two to know what I mean.

I've never seen a problem with having a PlanningScene "snapshot" every 0.5 seconds and a RobotDisplay showing the latest robot state at the same time... The snapshots simply trail the RobotDisplay during motions. But I do see your motivation for the patch.

> People with live capture and constantly updated octomaps can still lower that rate if necessary, all other will simply enjoy progress ;)

This is not an acceptable default if it almost surely overloads systems with sensor-based live updates of the scene. Actually, even the planning scene monitor allows to subscribe directly to `/joint_states` instead of getting the updates from the scene topic to avoid overloading the topic for no reason.
I would propose to either utilize this method in the PlanningSceneDisplay as a boolean property or make add an update frequency in the PlanningSceneMonitor that applies to joint changes / object position changes.
I guess the latter are also not optimized by your proposal yet?

simonschmeisser:
`RobotDisplay` does not display attached objects.This PR optimizes joint position changes with attached objects so they have no extra cost (a complete rebuild of the scene on every update) any longer.

Scene rebuilds still occur if `GEOMETRY_CHANGED` is emitted/transmitted. I can plan on adding some extra caching here as well but I have no completely straightforward solution yet. I'll do a separate PR for that therefore.

The publish rate is an upper limit btw, ie if nothing happens update rates will be lower. Also all sensor interfaces have update limits themselves, so you can still tune that to what your system can handle (but then you should maybe just reduce the update rate sensor-side). 

All of that being said, I'm fine with bargaining for what we change the publish rate to ;)

(I prefer having the data go via a common path so that there are no "multiple realities" and what the display shows is what the code acts upon. That's why I'm not a fan of using the `/joint_states` directly in the display. Also it would add more sources of confusion)

simonschmeisser:
How about defaulting to 10 or 15Hz? (And postponing a further increase to once CollisionObject updates have been optimized equivalently)

v4hn:
> > This is not an acceptable default if it almost surely overloads systems with sensor-based live updates of the scene. Actually, even the planning scene monitor allows to subscribe directly to /joint_states instead of getting the updates from the scene topic to avoid overloading the topic for no reason.
I would propose to either utilize this method in the PlanningSceneDisplay as a boolean property or make add an update frequency in the PlanningSceneMonitor that applies to joint changes / object position changes.
I guess the latter are also not optimized by your proposal yet?

> How about defaulting to 10 or 15Hz? (And postponing a further increase to once CollisionObject updates have been optimized equivalently)

Are we on a market haggling? updates with less than 30Hz might look even worse than the current version because it might look like RViz stutters. Also it's still a lot of overhead if you have to sent octomaps.

Are you unhappy with both of my proposed solutions?

simonschmeisser:
I'm not particularily keen on consuming the joint_states directly yes, as I'm a bit afraid of the visualization diverging from the logic. Also I think there are too many options already.

For the second proposal that exists already via rate limiting your input. Octomaps are only send when they have changed. They are unfortunately currently redrawn/repopulated whenever the geometry changes (ie not when there is just a joint position changing but when changing any CollisionObject). Selectively rate limiting some part of the output could also create weird divergences (at least for some frames). 

I'll drop changing the publishing rate and re-propose that once I got some input /experiments on #2079 and on octomap "rendering". Would that be an acceptable intermediate step? 

> it might look like RViz stutters
it doesn't now?

v4hn:
> > it might look like RViz stutters
>
> it doesn't now?

My impression was always that it is *obviously* throttled by program logic, but if you suspect your system is under heavy load, it might look like that too, yes...

> I'll drop changing the publishing rate and re-propose that once I got some input /experiments on #2079 and on octomap "rendering". Would that be an acceptable intermediate step?

Sounds good to me. Your other changes here are a definite improvement.

v4hn:
@simonschmeisser Looks good to me.
You added some quite nice standalone commits here but afterwards added fixups.
Could you squash the last three ones into their respective commits?

simonschmeisser:
@v4hn should be clean now, sorry for the mess.

v4hn:
> In #2049 (comment), @simonschmeisser promised to not increase the publishing rate, but the commit is still there: 1df820b. Am I missing something?

The commit you cite decreases the timer that includes update from `joint_state` in the monitored scene.
This is separate from [this parameter](https://github.com/ros-planning/moveit/blob/1df820bff0ccee71c4904f4b532bc05284335702/moveit_ros/planning/planning_scene_monitor/cfg/PlanningSceneMonitorDynamicReconfigure.cfg#L7) which you are worried about.

simonschmeisser:
Thanks for review and merging, I'll prepare a backport PR later on. Now if I could have your opinion on #2079 that would be great (but not sure when I get around to it though)

Issue #2052, opened 2020-05-05
Planning adapter adding state to trajectory but not velocity or acceleration


MarqRazz:
### Description
I am attempting to run back to back motion on my Motoman robot arm. This includes planning for a move, executing it with the `move_group.execute(plan, wait=True)` command then after it finishes planning a second move and executing it the same way. Every once in a while I get a printout from MoveIt stating:
`ros.moveit_ros_planning: Planning adapters have added states at index positions: [ 0 ]`
 followed by the robot arm rejecting the trajectory because the first point is missing velocity data.
`Validation failed: Missing velocity data for trajectory pt 0`

Here is a screen shot of the trajectory sent to the robot arm where you can clearly see the added state that is missing the velocity and acceleration data...
![Screenshot from 2020-04-20 13-21-34](https://user-images.githubusercontent.com/25058794/81017603-1162b280-8e20-11ea-95c4-bc15bc794549.png)

Here is a printout from my terminal.
```
ros.moveit_planners_ompl.model_based_planning_context: Planner configuration 'manipulator[PRM]' will use planner 'geometric::PRM'. Additional configuration parameters will be set when the planner is constructed.
ros.moveit_planners_ompl.ompl: manipulator/manipulator[PRM]: Starting planning with 2 states already in datastructure
ros.moveit_planners_ompl.ompl: manipulator/manipulator[PRM]: Created 2 states
ros.moveit_planners_ompl.ompl: Solution found in 0.503224 seconds
ros.moveit_planners_ompl.ompl: SimpleSetup: Path simplification took 0.000002 seconds and changed from 2 to 2 states
ros.moveit_ros_planning: Planning adapters have added states at index positions: [ 0 ]
ros.moveit_ros_move_group.ExecuteTrajectoryAction: Execution request received
ros.motoman_driver: Validation failed: Missing velocity data for trajectory pt 0
```

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Source build, master branch

### Steps to reproduce
I have attempted to get this error in simulation but I only see it when working with the hardware.

### Expected behaviour
When appending points to the trajectory the planning adapters should also calculate and fill in the velocity and acceleration.

### Possible fixes
[...]

PR #2053, opened 2020-05-05
fix ordering of request adapters

v4hn:
Some auxiliary PRAs can create additional waypoints,
so time parameterization must run *after* them on return of a successful plan,
which means it has to be loaded *before* them in the PRA chain.

Additionally, constraint names have to be resolved *before* fixing the input state w.r.t. them.

Look up the source code and documentation if you don't believe it.

This came up in https://github.com/ros-planning/moveit/issues/2052 .

### DIFF
--- a/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/ompl_planning_pipeline.launch.xml
+++ b/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/ompl_planning_pipeline.launch.xml
@@ -6,12 +6,12 @@
   <!-- The request adapters (plugins) used when planning with OMPL.
        ORDER MATTERS -->
   <arg name="planning_adapters"
-       value="default_planner_request_adapters/FixWorkspaceBounds
+       value="default_planner_request_adapters/AddTimeParameterization
+              default_planner_request_adapters/ResolveConstraintFrames
+              default_planner_request_adapters/FixWorkspaceBounds
               default_planner_request_adapters/FixStartStateBounds
               default_planner_request_adapters/FixStartStateCollision
-              default_planner_request_adapters/FixStartStatePathConstraints
-              default_planner_request_adapters/ResolveConstraintFrames
-              default_planner_request_adapters/AddTimeParameterization"
+              default_planner_request_adapters/FixStartStatePathConstraints"
               />
 
   <arg name="start_state_max_bounds_error" value="0.1" />


rhaschke:
> Look up the source code and documentation if you don't believe it.

Actually I did twice! I also enabled debugging output from the planning adapters. What I get is:
```
...
[DEBUG] ros.moveit_ros_planning: Running 'Fix Workspace Bounds'
[DEBUG] ros.moveit_ros_planning: Running 'Fix Start State Bounds'
[DEBUG] ros.moveit_ros_planning: Running 'Fix Start State In Collision'
[DEBUG] ros.moveit_ros_planning: Running 'Fix Start State Path Constraints'
[DEBUG] ros.moveit_ros_planning: Running 'Resolve constraint frames to robot links'
... ompl planning output ...
[DEBUG] ros.moveit_ros_planning: Running 'Add Time Parameterization'
...
```
which is exactly the expected order. @gavanderhoorn already pointed to the relevant code snippet:
https://github.com/ros-planning/moveit/blob/5553507f97f36d8c34678975b1888fb2cff02960/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp#L146-L152

This builds a call chain, starting from the last adapter. But, eventually, the last `fn` created is the first adapter, which is called first. So, the order isn't reversed at all.

Looking deeper into the code, it's actually even more complicated: Each adapter essentially gets a functionn pointer (`fn`, comprising the chain of the remaining adapters to call). But, each adapter can decide on its own, whether to adapt the request before or after calling this function. We can distinguish to classes of adapters:
1. those, which first adapt the _request_ and subsequently call `fn` (`FixXXX` adapters)
2. those, which first call `fn` and only then adapt the _result_ (all time parameterization adapters)

While the former adapters are always executed before the planner, the latter ones are _always_ execute afterwards - no matter in which order they appear in the list. However, if there are multiple adapters of the 2nd type listed, their order is indeed reversed - due to the fact that they first call their `fn` and then apply themselves.

I think we need more debugging to really understand this bug. I agree with @v4hn, that `ResolveConstraintFrames` should be executed before `FixStartStatePathConstraints`.

v4hn:
> which is exactly the expected order

No it's not. 

You did not mention whether you got this output *without* or *with* my patch.
Judging from the wrong order of `Resolve Constraint Frames` and `Fix Path Constraints` it's apparently the one without my patch.

Your explanation sounds somewhat confused to me, but granted, the code is awful.
I probably expected too much to ask for RTFS here. :-)

Part of the confusion probably arises because the PRA itself is responsible for printing the debugging info and not all plugins print it upon getting called.

The PRA chain sets up a V-style call structure where the "inward pass" towards the solver is defined by the list of parameters. The structure *should* look like that (leaving out a number of PRAs): 

pra-call-structure

Which is exactly what my patch does.

You missed one crucial point in your explanations. PRAs are free to act before **and** after the planner to adjust the request and the result accordingly.
[`fix_start_state_collision` does that](https://github.com/ros-planning/moveit/blob/5553507f97f36d8c34678975b1888fb2cff02960/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp#L155)

gavanderhoorn:
I believe I've found why I remembered things as doing a reverse application of the adapters: it's mentioned briefly in [wiki/industrial_trajectory_filters/Tutorials: Using a planning adapter inside of MoveIt - Example](http://wiki.ros.org/industrial_trajectory_filters/Tutorials/filters_inside_moveit):

> Added `industrial_trajectory_filters/UniformSampleFilter` to the `planning_adapters` list parameter. By adding this filter to the **beginning** of the list, we ensure that it is the **last** filter to be applied. This happens because the list is used for pre-processing trajectories (before planning), in which case the list is executed in order, and post-processing (after planning), in which case the list is executed in reverse order.

v4hn:
> it's mentioned briefly in wiki/industrial_trajectory_filters/Tutorials: Using a planning adapter inside of MoveIt - Example:

I do not really agree with the "is run before or after" part, because they all run as a single call stack, but aside from that the passage you cite is correct.

> We should definitely mention the fact that time parameterization PRAs should always be listed on top, such that they have "highest priority" / access the final result.

That's not necessarily what we want. The only real restriction is that PRAs listed *before* the parameterization have to be aware of it and PRAs listed *after* can freely modify kinematic waypoints. 

I'll merge this and create a WMD issue to improve the code/documentation for it.

PR #2060, opened 2020-05-07
Added support for PS4 joystick

TrippleBender:
### Description

### Checklist

### DIFF
--- a/moveit_ros/visualization/src/moveit_ros_visualization/moveitjoy_module.py
+++ b/moveit_ros/visualization/src/moveit_ros_visualization/moveitjoy_module.py
@@ -322,6 +322,163 @@ class PS3WiredStatus(JoyStatus):
         self.right_analog_y = msg.axes[3]
         self.orig_msg = msg
 
+class PS4Status(JoyStatus):
+    def __init__(self, msg):
+        JoyStatus.__init__(self)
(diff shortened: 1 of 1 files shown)


TrippleBender:
I added support for the PS4 Controller. Also added a description in the tutorial

simonschmeisser:
Thanks for your first PR! This is how we move it! ;)

Please note that the text that is there when you open the PR is meant as a template, so please click edit and exchange `Please explain the changes you made, including a reference to the related issue if applicable` by a description of what you want to change and why. You later wrote that in a comment but it's to find and parse this way.

Also, next time, when you write `Also added a description in the tutorial` please add a link to the PR for the tutorial

simonschmeisser:
In an ideal world I would prefer it if this code would not guess the type of input device from the number of buttons but that does not seem to be exposed by this very spartan msg api, so LGTM

v4hn:
> In an ideal world I would prefer it if this code would not guess the type of input device from the number of buttons but that does not seem to be exposed by this very spartan msg api, so LGTM

You could try to convince `joy` upstream to add the device id (Best case also unique between multiple devices) as `frame_id`. That's pretty much what I would expect from the field in this message.

v4hn:
Merging with two approvals.

PR #2082, opened 2020-05-12
fix signal for planning_attempts

v4hn:
fixup for dfb00da5917f357f56b54db5b0e3038a3180e10a


simonschmeisser:
Did we drop Qt4 already? Then this could be checked at compile time by using modern style connects

rhaschke:
Is there a nice tool available to upgrade all connects?

simonschmeisser:
https://github.com/KDE/clazy can do that

v4hn:
> Did we drop Qt4 already? Then this could be checked at compile time by using modern style connects

Yes we can upgrade the slot connects, I just wanted to keep the old style consistent and did not want to spend even more time on refactoring for now.

Pull-requests that ran automatic tools and upgrade the format are clearly appreciated :sunglasses:

PR #1606, opened 2019-08-08
Separate moveit_experimental packages

rhaschke:
This is a replacement for #1549, factoring out all packages in `moveit_experimental` as proper ROS packages. This allows to build them separately as intended by #1549.
Currently, only `jog_arm` is enabled, all other packages are catkin-ignored (some of them are not even catkinized yet). In contrast to #1549, this PR leaves all folders in their place. 
That is, `moveit_experimental` just acts as a container folder for these packages.
As we don't plan to release `moveit_experimenal` packages, there is no need to define a corresponding meta package.

- [x] This PR builds on top of #1605 and just adds 3c5f3eb. 
Thus this PR depends on the former being merged first.

PR #2083, opened 2020-05-12
[melodic-devel] jog arm backport

tylerjw:
### Description

This is a backport of the jog_arm feature.  This was created using git-filter-repo.


tylerjw:
Rebased to include changes from https://github.com/ros-planning/moveit/pull/1606 outside of moveit_jog_arm folder that changed out moveit_experimental is built to allow moveit_jog_arm to be built as a package.

tylerjw:
I'm not sure this is possible to back-port without the change https://github.com/ros-planning/moveit/pull/1584  which looks like it would break the released API.  @v4hn do I understand this correctly?

v4hn:
On Tue, May 12, 2020 at 01:08:16PM -0700, Tyler Weaver wrote:
> I'm not sure this is possible to back-port without the change https://github.com/ros-planning/moveit/pull/1584

This change must definitely not be backported.

Looking at `moveit_jog_arm`, `CollisionEnv` seems to be used only in two places though.
It looks like this is easy to revert back to use the previous API (which provided the same API split in `CollisionRobot` and `CollisionWorld`).

Talking to @henningkayser though, he was wondering why you are actually so much interested in backporting `moveit_jog_arm`.
Please talk internally to decide whether it's worth having it in melodic with backport modifications.

henningkayser:
@tylerjw wow, great effort going through all of these! I wasn't aware how work we've already spent on JogArm

> > On Tue, May 12, 2020 at 01:08:16PM -0700, Tyler Weaver wrote: I'm not sure this is possible to back-port without the change #1584

> This change must definitely not be backported. Looking at `moveit_jog_arm`, `CollisionEnv` seems to be used only in two places though. It looks like this is easy to revert back to use the previous API (which provided the same API split in `CollisionRobot` and `CollisionWorld`).

I agree, shouldn't be an issue to fix that.

> Talking to @henningkayser though, he was wondering why you are actually so much interested in backporting `moveit_jog_arm`. Please talk internally to decide whether it's worth having it in melodic with backport modifications.

Yep, talked to Tyler and we actually need it in melodic and it makes sense to have a released JogArm version for Ubuntu 18.04.

henningkayser:
@AndyZe Do you mind giving this a quick sanity check?

tylerjw:
Since this contains many commits, please do not squash this.  Land it with a merge or rebase.

AndyZe:
I cloned and built this and started going through the tutorial. Unfortunately the jog node crashes almost immediately without any useful message. It crashes before any command is even sent to the jogger   :(

`[jog_server-2] process has died [pid 20429, exit code -11, cmd /home/andyz/catkin_ws/devel/lib/moveit_jog_arm/jog_server __name:=jog_server __log:=/home/andyz/.ros/log/63b6e36c-954b-11ea-a1a2-10f005ed615b/jog_server-2.log].
log file: /home/andyz/.ros/log/63b6e36c-954b-11ea-a1a2-10f005ed615b/jog_server-2*.log`

AndyZe:
If I comment the main loop of jog_calcs.cpp, it does not crash. So I think the crash is caused there. I started the comment block at jog_calcs.cpp L124

rhaschke:
@tylerjw, To validate that you fetch all relevant commits, you should perform a git diff of the relevant subfolder with the master branch. There are nice tools, e.g. `meld` or VS [`code`](https://medium.com/faun/using-vscode-as-git-mergetool-and-difftool-2e241123abe7) to have a nice side-by-side comparison.

AndyZe:
I see these commits missing:

1982,

tylerjw:
#1982 is in here but doesn't show up in the list because this is a cherry-pick of commits with code changes.  It was merged and therefore the commits contained in it are here and not the merge commit.  Here is a link to the changes from that PR and you can find them in the list of commits for this PR: https://github.com/ros-planning/moveit/pull/1982/commits

#1863 is here as it was squashed and you can find it by just going into the list of commits above and searching for the number on the page.

v4hn:
@AndyZe anything else or can we merge this?

AndyZe:
We really ought to get #2094 into it first. Then it looks good.

AndyZe:
Tested it with the changes in #2094 and the crash is gone. Good to merge after that IMO

tylerjw:
Assuming Travis passes I think this is good to go.  I've tested it locally by going through the tutorials.

In the future, I'd like to add a test or extend an existing test that would have failed without the last change.

Thankyou @rfeistenauer for this bug fix, you might have the record for getting your first PR into a release the quickest.  :)

tylerjw:
Rebased on melodic-devel after bug fix backport PR was merged.

rhaschke:
This merge broke the build farm, because the corresponding changes to moveit_msgs are not yet released: http://build.ros.org/job/Mdev__moveit__ubuntu_bionic_amd64/66/consoleFull#console-section-13

v4hn:
> This merge broke the build farm, because the corresponding changes to moveit_msgs are not yet released:

Thanks for the link. Is that failure of any relevance?
I'm actually wondering why we test on the buildfarm now that `moveit_ci` provides much broader testing (and adheres to our build instructions...).
Releasing `moveit_msgs` is [on the way](https://github.com/ros-planning/moveit_msgs/pull/77) and probably waiting for your feedback?
Releasing `moveit` is scheduled [in one month](https://github.com/ros-planning/moveit/issues/2066).

PR #2087, opened 2020-05-13
WIP: add service to clear octomap

JStech:
### Description

We have an application where we want to generate an octomap from a point cloud, use it for a time, and then clear it and generate a new one (in a different location). To do this, I put together this service to clear the octomap from the point cloud updater.

### Checklist

### DIFF
--- a/moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
+++ b/moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
@@ -47,6 +47,7 @@ DIAGNOSTIC_POP
 #include <sensor_msgs/PointCloud2.h>
 #include <moveit/occupancy_map_monitor/occupancy_map_updater.h>
 #include <moveit/point_containment_filter/shape_mask.h>
+#include <moveit_msgs/ClearMap.h>
 
 #include <memory>
 
@@ -74,6 +75,7 @@ private:
   bool getShapeTransform(ShapeHandle h, Eigen::Isometry3d& transform) const;
   void cloudMsgCallback(const sensor_msgs::PointCloud2::ConstPtr& cloud_msg);
   void stopHelper();
+  bool clearMap(moveit_msgs::ClearMap::Request& req, moveit_msgs::ClearMap::Response& res);
 
   ros::NodeHandle root_nh_;
   ros::NodeHandle private_nh_;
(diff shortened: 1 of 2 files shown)


moveit_ros/perception/pointcloud_octomap_updater/include/moveit/pointcloud_octomap_updater/pointcloud_octomap_updater.h
```diff
@@ -47,6 +47,7 @@ DIAGNOSTIC_POP
 #include <sensor_msgs/PointCloud2.h>
 #include <moveit/occupancy_map_monitor/occupancy_map_updater.h>
 #include <moveit/point_containment_filter/shape_mask.h>
+#include <moveit_msgs/ClearMap.h>
```
henningkayser, line 50: I would probably call the service `ClearOctomap`

henningkayser:
> I think I remember a service to clear Octomap being available already. Looks like this PR added it:
> 
> [ros-planning/moveit_ros#495](https://github.com/ros-planning/moveit_ros/pull/495)
> 
> Can you double-check that does not exist already? Maybe `rostopic list |grep octo` or something like that

Good call! Of course, this capability is currently not available when running MoveItCpp, but maybe we can unify the implementation somehow.

Dale-Koenig:
> Looks good, just consider renaming the message.
> 
> @Dale-Koenig Is this something you would be interested in, in addition to #1984?
> 
> ... now that I think about it, shouldn't this feature be handled by the `UpdatePointcloudOctomap` service? Clearing is also achieved by updating an empty point cloud in snapshot mode. I think this should be an extension of the same service protocol.

At the moment this still wouldn't allow clearing after incrementally updating several times, since there is no option to switch between the modes within the service.  It is true though that maybe the service could be changed to be able to handle both clearing and updating, instead of needing a separate service.

@JStech are you processing multiple pointclouds in each location before switching?

JStech:
Updates aren't necessary in my current application, although we are using MoveItCpp. I'll look at how move_group handles it and see if I can unify them, as @henningkayser suggests.

Dale-Koenig:
> Updates aren't necessary in my current application, although we are using MoveItCpp. I'll look at how move_group handles it and see if I can unify them, as @henningkayser suggests.

If updates aren't necessary then #1984 should work.  No new service or unification needed, just use the octomap in snapshot mode.

Dale-Koenig:
And if there's some reason it doesn't work for you please let me know :)

v4hn:
@Dale-Koenig @JStech can we close this in favor of the more general #1984 ?
Is it possible to send an empty point cloud message with your approach to clear it?

JStech:
I've not gotten back to this work, but it does sound like this PR is unnecessary.

PR #2089, opened 2020-05-14
Check for empty quaternion message when processing CollisionObjects

felixvd:
### Description

To avoid issues like the one @tylerjw encountered in https://github.com/ros-planning/moveit/pull/2031, I think we should check for empty quaternions when adding objects to the scene. This is a lightweight operation compared to the `normalize` that we already apply to every message, so it seems reasonable to me. Please share your opinion.

Fundamentally this is only an issue because ROS quaternion messages default to an invalid value, but that's neither here nor there.

I'll run clang-format etc. when I am at my home machine.

### Checklist

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1687,7 +1687,15 @@ void PlanningScene::poseMsgToEigen(const geometry_msgs::Pose& msg, Eigen::Isomet
 {
   Eigen::Translation3d translation(msg.position.x, msg.position.y, msg.position.z);
   Eigen::Quaterniond quaternion(msg.orientation.w, msg.orientation.x, msg.orientation.y, msg.orientation.z);
-  quaternion.normalize();
+  if ((quaternion.x() == 0) && (quaternion.y() == 0) && (quaternion.z() == 0) && (quaternion.w() == 0))
+  {
+    ROS_WARN_NAMED(LOGNAME, "Empty quaternion found in pose message. Setting to neutral orientation.");
+    quaternion.setIdentity();
+  }
+  else
+  {
+    quaternion.normalize();
+  }
   out = translation * quaternion;
 }
 


v4hn:
Thank you for taking the initiative! I believe we had partial patches for this around already, but I can't find the corresponding request or issue right now...

I am not happy with your method to resolve this though.
In my opinion, the `ApplyPlanningScene` top-level ROS service and the respective `PlanningScene` methods should not succeed when such an object is received.

rhaschke:
It is kind of common practice in ROS to consider uninitialized quaternions as identity, e.g. in rviz:
https://github.com/ros-visualization/rviz/blob/171eb3cfd1ec69cb6ff72da3ca41a1a4caa74eef/src/rviz/validate_quaternions.h#L50-L56

https://github.com/ros-visualization/rviz/blob/171eb3cfd1ec69cb6ff72da3ca41a1a4caa74eef/src/rviz/validate_quaternions.h#L66-L72

v4hn:
> It is kind of common practice in ROS to consider uninitialized quaternions as identity

I am aware of that but not happy with it at all :disappointed:
You are right though, we should probably keep users ~~happy~~happier and use the warning and implicit identity.

tylerjw:
> > It is kind of common practice in ROS to consider uninitialized quaternions as identity
> 
> I am aware of that but not happy with it at all 
> You are right though, we should probably keep users ~happy~happier and use the warning and implicit identity.

This does add extra instructions (a test and jump) in the case where the user did the right thing and didn't give us something invalid.  My pain (and other beginner users) over this comes from two places.  The first is nearly useless errors in FCL when used on Kinetic and even worse, it silently somehow doesn't complain when using Melodic.

If this was a function that users would have to compile and use we could put a static_assert in there but that kind of defeats using it as a library or dynamic linking helping the user because it'll get compiled out and never checked.

The last option would be to add default construction to ROS messages, which is something we have with ROS2 so hopefully as people migrate it won't matter and this is just one more case of us having to implement something downstream that should have been done upstream at compile time in the messages.

As to what the default construction value should be.  I don't know enough about quaternions and if {0,0,0,0} would ever actually be valid and therefore we shouldn't stop that from being tested.  If we want to be even safer we should probably test for anything that'd cause a math exception on division (nan, inf) too, but I don't think we should try to save users who give us quaternions with values set to nan.

Do you propose we set the default quaternion to something other than identity?  Maybe {1, 0, 0, 0} is more reasonable, or {1, 1, 1, 1} since we normalize it after all.

In the end, thank you for merging this.  I hope it saves someone else in the future from copy-pasting from our own tutorials and being greeted by an assert at runtime.  :1st_place_medal:

rhaschke:
@tylerjw, ROS1 message system will not change anymore. So we need to work with such fixes.
Just for your info: The identity quaternion is `(x=y=z=0, w=1)`.

gavanderhoorn:
There is actually *some* support for adding constructors and other functionality to messages.

See [Customizing generated message headers for C++](http://wiki.ros.org/roscpp/Overview/MessagesSerializationAndAdaptingTypes#Customizing_generated_message_headers_for_C.2B-.2B-).

rhaschke:
Interesting to know, but this approach doesn't generalize across all supported languages:

> **IMPORTANT**: be careful, as doing so might break the expected message API and/or compatibility with other languages.

gavanderhoorn:
Adding some convenience constructors to message header does not break compatibility.

Other languages will simply not get those conveniences.

rhaschke:
You are right. What I meant is that users will have a different experience across languages and that's probably nothing we should support.

PR #2091, opened 2020-05-15
Minor logic cleanup

jesseli2002:
### Description

I was skimming through the internals and noticed this bit of logic which I think can be cleaned up.

Due to the check `priority > 0`, there is no time when `best == state_space_factories_.end()` while `priority <= prev_priority`, so `best == state_space_factories_.end()` is unnecessary. Then if you change the starting value of `prev_priority` then you don't need to separately check if `priority > 0`. 

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
@@ -220,7 +220,7 @@ MultiQueryPlannerAllocator::allocatePersistentPlanner<ompl::geometric::LazyPRMst
 {
   return new og::LazyPRMstar(data);
 };
-}
+}  // namespace ompl_interface
 #endif
 
 ompl_interface::PlanningContextManager::PlanningContextManager(moveit::core::RobotModelConstPtr robot_model,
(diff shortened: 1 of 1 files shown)


moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
```diff
@@ -220,7 +220,7 @@ MultiQueryPlannerAllocator::allocatePersistentPlanner<ompl::geometric::LazyPRMst
 {
   return new og::LazyPRMstar(data);
 };
-}
+}  // namespace ompl_interface
```
jesseli2002, line 223: I think this was done by clang-format

mamoll:
This shows that -1 is a bad initial value for prev_priority:
https://github.com/ros-planning/moveit/blob/master/moveit_planners/ompl/ompl_interface/src/parameterization/work_space/pose_model_state_space_factory.cpp#L83
Changing it to `-std::numeric_limits<int>::infinity()` would work, but I'm still not convinced this change would improve the logic.

jesseli2002:
`prev_priority` isn't used later in the function. My logic of changing it to have an initial value of 0 is to merge the checks `priority > prev_priority && priority > 0` - if `prev_priority` is less than 0, then you still need to check that `priority > 0`.

v4hn:
Merging as trivial improvement. Thanks for taking the time @jesseli2002 :+1: 

@mamoll apparently misunderstood the context.

PR #2100, opened 2020-05-21
[jog_arm] time-based collision avoidance

AndyZe:
The old collision avoidance algorithm used manually-defined thresholds. This adds a new algorithm based on how fast the collision is approaching. It's optional, as you can see from the yaml file.

Video of it in action:

https://drive.google.com/file/d/1vMrrtflJme3Hexfo-685rKVS7uWIy14s/view?usp=sharing

### DIFF
--- a/moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
@@ -83,7 +83,7 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
   double velocity_scale = 1;
 
   collision_check_type_ =
-      (parameters_.collision_check_type == "threshold_distance" ? kThresholdDistance : kStopDistance);
+      (parameters_.collision_check_type == "threshold_distance" ? K_THRESHOLD_DISTANCE : K_STOP_DISTANCE);
 
   // Variables for stop-distance-based collision checking
   double current_collision_distance = 0;
@@ -131,7 +131,7 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
         velocity_scale = 0;
       }
       // If threshold distances were specified
-      else if (collision_check_type_ == kThresholdDistance)
+      else if (collision_check_type_ == K_THRESHOLD_DISTANCE)
       {
         // If we are far from a collision, velocity_scale should be 1.
         // If we are very close to a collision, velocity_scale should be ~zero.
--- a/moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
@@ -40,7 +40,8 @@
 #include <moveit_jog_arm/collision_check_thread.h>
 
 static const std::string LOGNAME = "collision_check_thread";
-static const double MIN_RECOMMENDED_COLLISION_RATE = 10;
+static constexpr double MIN_RECOMMENDED_COLLISION_RATE = 10;
+constexpr double EPSILON = 1e-6;  // For very small numeric comparisons
 
 namespace moveit_jog_arm
 {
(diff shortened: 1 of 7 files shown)


v4hn:
You forgot the video :smile:

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -115,36 +124,72 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
...
       {
-        // velocity_scale = e ^ k * (collision_distance - threshold)
-        // k = - ln(0.001) / collision_proximity_threshold
-        // velocity_scale should equal one when collision_distance is at collision_proximity_threshold.
-        // velocity_scale should equal 0.001 when collision_distance is at zero.
-        velocity_scale =
-            std::min(velocity_scale, exp(scene_velocity_scale_coefficient *
-                                         (scene_collision_distance - parameters_.scene_collision_proximity_threshold)));
+        // If we're definitely in collision, stop immediately
+        if (collision_detected)
```
henningkayser, line 132: This should be triggered in both modes

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -115,36 +124,72 @@ void CollisionCheckThread::startMainLoop(JogArmShared& shared_variables)
...
+        if (current_collision_distance < parameters_.min_allowable_collision_distance &&
+            derivative_of_collision_distance <= 0)
+        {
+          velocity_scale = 0;
+        }
+        // Only bother doing calculations if we are moving toward the nearest collision
+        else if (derivative_of_collision_distance < 0)
+        {
+          // At the rate the collision distance is decreasing, how long until we collide?
+          est_time_to_collision = fabs(current_collision_distance / derivative_of_collision_distance);
```
henningkayser, line 177: This should check for division by 0. Also, what happens if the distance values are not continues but a little bit noisy?

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -743,6 +741,47 @@ bool JogCalcs::updateJoints(JogArmShared& shared_variables)
...
+    {
+      if (joint_model->getName() == joint_name)
+      {
+        // Some joints do not have acceleration limits
+        try
+        {
+          kinematic_bounds = joint_model->getVariableBounds();
+          // Be conservative when calculating overall acceleration limit from min and max limits
+          accel_limit =
+              std::min(fabs(kinematic_bounds[0].min_acceleration_), fabs(kinematic_bounds[0].max_acceleration_));
```
henningkayser, line 766: I think we can safely take `min_acceleration_` here as it's defined and implemented as min/max range (with min value being the negative max value by default).

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -743,6 +741,47 @@ bool JogCalcs::updateJoints(JogArmShared& shared_variables)
...
+        }
+        catch (const std::exception& ex)
+        {
+          ROS_WARN_STREAM_NAMED(LOGNAME, "An acceleration limit is not defined for this joint; minimum stop distance "
```
henningkayser, line 770: We could check this using `VariableBounds::acceleration_bounded_`

henningkayser:
General approach looks good to me. I see that it makes a lot of sense to throttle the motion dynamically. Is there a reason we don't apply both methods at the same time so that the user decides how to set this up, i.e. you could throttle scene collisions with the first method and then only use the second method for self collisions by setting the distance/threshold parameters.

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
AndyZe: done

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
AndyZe: I don't think there's a lot I can do about that. From my testing, the distance values have not been noisy, though.

I could apply a low-pass filter but the delay would not be helpful. I think it's best to leave this collision-checking mode optional for that reason. I can think of some scenarios where the user would not want dynamic collision checking (e.g. any contact task).

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: Basically, assume they are symmetric? That's OK with me

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: done

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
AndyZe: I'll handle the zero-check within the `else if ()` above

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AndyZe: Actually, not having an acceleration-bounded joint would be another good reason to make this kind of collision check optional.

PR #2103, opened 2020-05-22
[jog_arm] simplify communication between threads

tylerjw:
### Description

This PR went through many changes before arriving here.  If you find it now this is what the outcome so far is from an external perspective for interacting with jog_arm:

jog_server

The internal changes are there are no move manual threads or shared data between the workers (calcs and collision).  Instead, each maintains its own state and the only communication internally is from the collision thread to the calcs thread through a ros topic using zero-copy queueing.

For external communication (even in the same process) ros pub/sub is used because it provides very fast thread-safe zero-copy queues.  In a future PR I will convert this to be done via a nodelet so it is as simple as editing your launch file to use jog_arm.  In your application, you can then use zero-copy queues (assuming you also write a nodelet, which you should).

### Description (old)

This is a large commit so I've included descriptions of my justification for many of these.  If we can get some good testing of this and others agree I'd consider back-porting it.  It does break the API of jog_arm and would like to not do that on the release.

#### Inter-thread communication
For communication between threads I've converted from using shared memory in a structure with a mutex to lockfree::spsc_queue from boost.  This has the advantage of being statically allocated, lock-free, and simple to use.  There is one case where I find it not intuitive and that is when you want to read the latest value form the queue.  I wrote a template method `popLatest` that I put in jog_arm_data.h for that use case.

#### On `spsc_queue`
One gotcha that you need to be aware of is the commands shouldn't be sent faster than the loop in `JogCalcs` can consume them.  If you do this you will silently lose commands.  This is because the size of the queue is statically allocated so there are no malloc calls in the critical path so when the buffer is full and you try to add another message it just won't.  Also, you should not run `JogCalcs` faster than the loop in `JogArm` that is publishing commands or those commands will be lost.  To make it clear the loop rate should go `JogArm > JogCalcs > Commands`.

#### API changes (refactoring for simplicity and consistency)
In several places I was able to simplify the interface by removing code that was duplicated in multiple files.  This should make it much easier to maintain.  One large one is I removed the ros_interface and cpp_interface derived classes and the base class is just called `JogArm`.  This made so there is only one truth for the main loop and the behavior of jog arm.  For the ros_interface, I moved the little code that was not in the cpp_interface into the jog_server binary.

#### start, stop, and run
Another new change to the API is I added `start` and `stop` methods to `JogArm` to start and stop the threads.  This change allows you tighter control of when your threads are started and stopped.  For running threads I made all the main functions named `run` as this is a common pattern in threading.

#### On `atomic`
I looked through how `atomic` was used and found that it was actually making the code generated less performant and not any safer.  `atomic` is a construct designed to use special machine instructions to protect you from RMW (read-modify-write) statements.  An example of this would be a counter that you increment on two separate threads.  That is where you would use an atomic.  Specifically, if you find yourself doing these operations to something on multiple threads you should use an atomic: `+=, -=, &=, |=, ^=, --, ++`.  It is perfectly safe to write to a bool or read from it on two different threads.

#### On `mutex`
You will notice there are a few places where I placed new mutexes and others where I removed them.  First of all I removed the mutex from `JogArmShared` as it no longer contains data that is unsafe to write and read from on multiple threads.  The couple places where I added mutexes is where an asyc ros callback needs to write to an internal state variable that is used by the thread.  This is needed for types that are not rival (like `bool` or `double`).  The reason I didn't use a `spsc_queue` was for the reason stated above where the publisher will often run at a faster rate than the subscriber and therefore we will often drop the latest values.  In this case a mutex is better.

#### On `lock_guard`
You'll notice that in the few places where I do lock a mutex I use `lock_guard` which will unlock the mutex when it goes out of scope.  In a few places I created a new scope so it would only lock for one more statement.  I believe this makes it much easier to see where a mutex is being used and to reason about it.  It also saves you if you call `return` after lock as it will always get freed when you go out of scope.  You'll see I use that in a few places to copy a mutex protected value into the return value of a function.

#### On `isNonZero`
I moved non-zero message testing into the jog_calc thread and out of the callbacks.  This made the logic simpler to follow and eliminated the need for a couple variables in the shared data.

#### On multiple ROS subscribers to the same topic (joint state)
I noticed that each thread needed its own copy of the latest values of joint state and that it always comes from a ROS subscription.  There is no limit to the number of ROS subscribers for a single topic and each thread can just have their own which will all get a callback triggered when a new message comes in.  The spsc_queue is not a good option for this because when you read from it you erase a value and each thread needed the latest value.

#### Extending the CPP interface test
I added the statements to launch the robot_state_publisher to the launch file to allow it to run for a bit which tests that it doesn't crash or hang when you start it.  Previously it only tested the constructor.

### Checklist


AndyZe:
Wow, this is a lot of change. It will probably take awhile to test and I want to verify that it truly doesn't hurt performance.

I understand that an spsc_queue is for 2 threads (one to read, one to write). Well, jog_arm has 3 threads (interface, calcs, collision). Do you see any issues there or did you simplify it to just 2 threads?

AndyZe:
**On multiple ROS subscribers to the same topic (joint state)**

This worries me a little bit. I'm not sure if that's going to increase bandwidth by a factor of 3 vs. one subscriber.

(I have seen issues with too many joint_states subscribers before when I was running a mobile robot over wifi.)

tylerjw:
> This worries me a little bit. I'm not sure if that's going to increase bandwidth by a factor of 3 vs. one subscriber.
> 
> (I have seen issues with too many joint_states subscribers before when I was running a mobile robot over spotty wifi.)

When you saw issues how was it setup?  Were the multiple subscribers in one process or was it spread over several processes?  I went and read the roscpp code on this and it internally just calls each callback that has registered to the same topic with each message.

tylerjw:
> I understand that an spsc_queue is for 2 threads (one to read, one to write). Well, jog_arm has 3 threads (interface, calcs, collision). Do you see any issues there or were you very careful to use an spsc_queue where only 2 threads share data?

In each case the spsc queue is only being used between two threads (main and calc threads).  The collision thread doesn't use any queues.

AndyZe:
> When you saw issues how was it setup? Were the multiple subscribers in one process or was it spread over several processes? I went and read the roscpp code on this and it internally just calls each callback that has registered to the same topic with each message.

You are correct, it was multiple processes, so I guess this is fine.

moveit_experimental/moveit_jog_arm/src/jog_arm.cpp
```diff
@@ -238,18 +260,106 @@ bool JogInterfaceBase::readParameters(ros::NodeHandle& n)
...
+    else
+    {
+      ROS_DEBUG_STREAM_THROTTLE_NAMED(10, LOGNAME, "All-zero command. Doing nothing.");
+    }
+
+    main_rate.sleep();
+  }
+}
+
+void JogArm::start()
```
AndyZe, line 318: This function name could easily be confused with `run()`. I think you should change the name of one or the other. I think this makes sense:

run() --> processCommands()

AndyZe:
Assuming the code review goes well, I'd be fine with you adding yourself as an author in package.xml

moveit_experimental/moveit_jog_arm/CMakeLists.txt
```diff
@@ -82,13 +86,13 @@ target_link_libraries(cpp_interface_example
...
-  src/jog_interface_base.cpp
-  src/jog_ros_interface.cpp
+  src/jog_arm.cpp
   src/jog_server.cpp
   src/low_pass_filter.cpp
 )
 add_dependencies(jog_server ${catkin_EXPORTED_TARGETS})
-target_link_libraries(jog_server ${catkin_LIBRARIES} ${Eigen_LIBRARIES})
+target_link_libraries(jog_server
+  ${catkin_LIBRARIES} ${Eigen_LIBRARIES}  ${Boost_LIBRARIES})
```
AndyZe, line 95: hit Enter between each entry, for consistency

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm.h
```diff
@@ -36,31 +36,46 @@
...
-class JogInterfaceBase
+class JogArm
 {
 public:
-  JogInterfaceBase();
+  JogArm(const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor);
+
+  ~JogArm();
+
+  /** \breif start jog arm threads */
```
AndyZe, line 71: breif  --> brief

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm.h
```diff
@@ -36,31 +36,46 @@
...
+  void start();
 
-  /** \brief Update the joints of the robot */
-  void jointsCB(const sensor_msgs::JointStateConstPtr& msg);
+  /** \breif stop jog arm threads */
```
AndyZe, line 74: breif --> brief

And several other places, too

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm.h
```diff
@@ -78,20 +93,46 @@ class JogInterfaceBase
...
+   * The units are determined by settings in the yaml file.
+   */
+  void provideTwistStampedCommand(const geometry_msgs::TwistStamped& velocity_command);
+  void provideTwistStampedCommand(const geometry_msgs::TwistStampedConstPtr& msg);
 
-  /** \brief Stop the main calculation thread */
-  bool stopJogCalcThread();
+  /** \brief Send joint position(s) commands */
+  void provideJointCommand(const control_msgs::JointJog& joint_command);
+  void provideJointCommand(const control_msgs::JointJogConstPtr& msg);
```
AndyZe, line 104: Please fill me in on why you overloaded this

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
```diff
@@ -38,34 +38,53 @@
...
-struct JogArmShared : public std::mutex
+// Types for queues between threads
+constexpr size_t QUEUE_SIZE = 10;
+template <typename T, size_t U>
+using MessageQueue = boost::lockfree::spsc_queue<T, boost::lockfree::capacity<U>>;
+using TwistedStampedQueue = MessageQueue<geometry_msgs::TwistStamped, QUEUE_SIZE>;
+using JointJogQueue = MessageQueue<control_msgs::JointJog, QUEUE_SIZE>;
+using JointTrajectoryQueue = MessageQueue<trajectory_msgs::JointTrajectory, QUEUE_SIZE>;
+
+// Helper function for poping the latest value out of a queue
```
AndyZe, line 65: poping --> popping

Unless you mean this:  https://media.irishpost.co.uk/uploads/2018/06/11150144/pope-francis-swag2.jpg

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
```diff
@@ -38,34 +38,53 @@
...
+    }
+    // read the latest value
+    queue.pop(output);
+  }
+  return true;
+}
+
+// Variables to share between threads
+// Be careful to not read-modify-write any of these because they are not
+// protected by a mutex.
```
AndyZe, line 85: woah, how are they protected now, then?

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -56,21 +57,29 @@ namespace moveit_jog_arm
...
+   * Get the MoveIt planning link transform.
+   * The transform from the MoveIt planning frame to robot_link_command_frame
+   *
+   * @param transform the transform that will be calculated
+   * @return true if a valid transform was available
+   */
+  bool getCommandFrameTransform(Eigen::Isometry3d& transform);
 
+private:
   /** \brief Do jogging calculations for Cartesian twist commands. */
```
AndyZe, line 83: spelling of "brief" here, too

moveit_experimental/moveit_jog_arm/CMakeLists.txt
```diff
@@ -54,14 +57,14 @@ include_directories(
 add_library(${LIBRARY_NAME} SHARED
   src/collision_check_thread.cpp
   src/jog_calcs.cpp
-  src/jog_cpp_interface.cpp
-  src/jog_interface_base.cpp
+  src/jog_arm.cpp
```
AndyZe, line 60: I think the name `jog_arm.cpp` is a little vague. I think `jog_arm_master` or `jog_arm_interface` would be better.

Normally I would assume `jog_arm.cpp` contains main(), but I see that's not true here -- main() is in jog_server.cpp.

moveit_experimental/moveit_jog_arm/test/jog_cpp_interface_test.test
```diff
@@ -5,6 +5,12 @@
     <arg name="load_robot_description" value="true"/>
   </include>
 
+  <!-- Initial joint positions -->
+  <node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher">
+    <rosparam command="load" file="$(find moveit_jog_arm)/test/config/initial_position.yaml" />
+    <param name="publish_frequency" type="double" value="50.0"/>
```
AndyZe, line 11: nice

moveit_experimental/moveit_jog_arm/src/jog_server.cpp
```diff
@@ -37,15 +37,108 @@
...
+  auto drift_dimensions_server =
+      nh.advertiseService(nh.getNamespace() + "/" + ros::this_node::getName() + "/change_drift_dimensions",
+                          &moveit_jog_arm::JogArm::changeDriftDimensions, &jog_arm);
+
+  // ROS Server for changing the control dimensions
+  auto dims_server =
+      nh.advertiseService(nh.getNamespace() + "/" + ros::this_node::getName() + "/change_control_dimensions",
+                          &moveit_jog_arm::JogArm::changeControlDimensions, &jog_arm);
+
+  // Start the jog server run ros spinner
```
AndyZe, line 134: this sentence needs a grammar check. Maybe:

// Start the jog server and run ros spinner

moveit_experimental/moveit_jog_arm/src/jog_server.cpp
```diff
@@ -37,15 +37,108 @@
...
+  jog_arm.start();
+
+  // Wait for ros to shutdown
+  ros::waitForShutdown();
```
AndyZe, line 138: I'm fairly sure this is optional, you could delete it

tylerjw: It isn't because I'm running everything in separate threads (not in the constructor).

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
tylerjw: They don't need a mutex because you never do a RMW (read-modify-write) operation on any of these.  They are written to and read in a thread-safe way by default as they are plain old data types.  Where you would need an atomic or mutex is if you do something like this:

shared_data.value += 1;

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
tylerjw: lol, fixed

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm.h
tylerjw: I did this so I can use the same method name for the callback for a ros subscriber and the C++ interface.  If I convert this to use ros style inter-process communication I'd probably change it to only have one method (the Ptr one) and call it a callback and make people using the C++ interface publish to the topic.

moveit_experimental/moveit_jog_arm/src/jog_arm.cpp
tylerjw: `run` is a common name for the main for a thread so I figured it was a good name.  `start`/`stop` are the public methods.

moveit_experimental/moveit_jog_arm/src/jog_server.cpp
tylerjw: If I exited without this ROS would just exit.  This just waits for `ros::ok()` to be false in a blocking way (so it isn't polling).

moveit_experimental/moveit_jog_arm/test/jog_cpp_interface_test.cpp
```diff
@@ -69,14 +67,20 @@ class TestJogCppInterface : public ::testing::Test
...
 {
-  moveit_jog_arm::JogCppInterface jog_cpp_interface(planning_scene_monitor_);
+  moveit_jog_arm::JogArm jog_arm(nh_, planning_scene_monitor_);
+  jog_arm.start();
+  ros::Duration(1).sleep();  // Give the started thread some time to run
+  jog_arm.stop();
+  ros::Duration(1).sleep();  // Give the started thread some time to run
+  jog_arm.start();
   ros::Duration(1).sleep();  // Give the started thread some time to run
+  jog_arm.stop();
```
tylerjw, line 83: I did this so I can test that starting and stopping works correctly.

tylerjw:
@AndyZe After reading the source for roscpp on how the message passing works I believe this is both simpler and more performant without boost::lockfree::spsc_queue.  I removed that and converted the sections using it to just plain old ROS pub/sub.

@AdamPettinger sorry I keep moving the goal post on this one.  I spent time this weekend learning about the internals of ros messaging and have much more confidence using it like this.

One thing to point out is that I did convert to using ros timers instead of threads as I believe that offers the user more control and probably better performance.  This does have the side affect that all of jog_arm will become effectively single-threaded if someone is using it as a CPP library and using a single-threaded ros spinner.  This should probably be called out in the tutorials.  The CPP example has been modified to use it correctly.

gavanderhoorn:
@tylerjw: from our ROS Answers discussion I'd have expected your memory pool approach to be included here.

There don't seem to be any references to `pool`. Was that ultimately not needed?

tylerjw:
I didn't include it because right now the only part of the code where it would be useful is from the user side.  I plan on including it in a tutorial on how to use jog_arm in a performant way.

If you look really closely you'll notice that I should have used it for `trajectory_msgs::JointTrajectory` across the internal namespace but I didn't.  The reason I didn't is I'm trying to limit the scope of this PR (you are welcome to laugh at me) and think that'll be a good starting point for refactoring the logic inside JogCalc that right now mixes parameters and object state for the methods it calls and I find it confusing.  I would like to map out the call stack and try to simplify it to make it easier to understand what the inputs and outputs are to each method.

gavanderhoorn:
> I didn't include it because right now the only part of the code where it would be useful is from the user side. I plan on including it in a tutorial on how to use jog_arm in a performant way.

Makes sense.

I haven't checked the tutorial but then I guess the "user side" would be a nodelet ideally?

> The reason I didn't is I'm trying to limit the scope of this PR (you are welcome to laugh at me) 

This did make me smile, yes ;)

> I would like to map out the call stack and try to simplify it to make it easier to understand what the inputs and outputs are to each method.

:+1:

tylerjw:
> I haven't checked the tutorial but then I guess the "user side" would be a nodelet ideally?

Not necessarily.  I think the real win with nodelets is if you have nodes you already reuse in some way and you wan to compose them in different sets.  For jog_arm without nodelets a reasonable use-case would be to just start the ros spinner, start jog arm, and then go create a thread where you process vision stuff or whatever your inputs are and publish to the jog_arm topics all on the same process.

I think the way I'd have to do that is create a JogArmNodelet that owns a JogArm and passes in the nodelet NodeHandle and nodelet private namespaced handle.  Also, another PR idea I'm trying to keep out of this one.

tylerjw:
Other ideas I'm resisting the urge to include in this PR:

* Using a memory pool allocator for `trajectory_msgs::JointTrajectory`
* Clarifying inputs/outputs and call stack inside `JogCalcs`
* Nodelet interface
* Figuring out a clearer approach for shared data (maybe through a custom internal message).
* Removing the parameter argument allowing users to leave the `frame_id` of commands blank and substituting it with a parameter (this would remove one more copy and I think make the code more standard).  Provide warnings if the user does this.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm.h
v4hn: `ros::Subscriber` is perfectly capable of handling a callback with a single `const pkg_msgs::Msg&` parameter.
This is usually the better option if you do not intend to store the `shared_ptr` somewhere.

tylerjw: See the latest version of this where I only accept the shared ptr for these as I do store the lastest one I've received.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
v4hn: First off, I'm clearly not an expert in threat safe data structures and I did not spent too much time with this code.
Always willing to learn though. :)

Can you actually measure a significant improvement in runtime behavior if you throw out the locking?

API wise, locked access is much better than a comment saying "be careful not to r-m-w".

If you revert to simple datastructures without explicit thread-safe access, do you take care of cache consistency manually? To the best of my knowledge, `<atomic>` also takes care of memory barriers.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm.h
v4hn: :laughing: I did not expect a reply for the next 6 hours at least. ;)

You are right and it's the correct interface now. I've seen a lot of code requiring the `ConstPtr&` for no particular reason over the years, so I thought I'd mention the `const Msg&` interface.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
tylerjw: Cache correctness is also not a problem here because all of these only get modified in one direction.  I will build a profile to show how much faster the loops run with this PR.

tylerjw: In the end I'd like to get rid of the shared variables as they are confusing but with the message types removed and the only types remaining are one direction modifies, these are safe.

tylerjw: I agree that these make for confusing API but even RMW is fine without a lock or atomic if each one is only ever changed on one thread.

AndyZe: I agree with v4hn here. It would be really easy to introduce a bug that's hard to track down in the future

tylerjw:
Here are some profiling results.  To produce this I modified the moveit_cpp example of each to jog a panda in a way that would cause a collision and plotted the time inside ~150 loop interactions of each loop.

main_thread
calcs_thread
collision_thread

moveit_experimental/moveit_jog_arm/src/jog_arm.cpp
AndyZe: I didn't notice `run` is private. OK, that makes sense.

moveit_experimental/moveit_jog_arm/CMakeLists.txt
AndyZe: @tylerjw you didn't address this yet. What do you think?

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -67,14 +70,14 @@ int main(int argc, char** argv)
...
-  std::thread jogging_thread([&]() { jog_interface.startMainLoop(); });
+  moveit_jog_arm::JogArm jog_arm(nh, planning_scene_monitor);
+  jog_arm.start();
 
   // Make a Cartesian velocity message
-  geometry_msgs::TwistStamped velocity_msg;
-  velocity_msg.header.frame_id = "base_link";
-  velocity_msg.twist.linear.y = 0.01;
-  velocity_msg.twist.linear.z = -0.01;
+  auto velocity_msg = boost::make_shared<geometry_msgs::TwistStamped>();
```
AndyZe, line 77: Can you add a comment here for people like me (and probably 90% of C++ users) to clarify why you're using a shared pointer?

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -93,31 +96,25 @@ int main(int argc, char** argv)
   ros::Duration(2).sleep();
 
   // Make a joint command
-  control_msgs::JointJog base_joint_command;
-  base_joint_command.joint_names.push_back("elbow_joint");
-  base_joint_command.velocities.push_back(0.2);
-  base_joint_command.header.stamp = ros::Time::now();
+  auto base_joint_command = boost::make_shared<control_msgs::JointJog>();
```
AndyZe, line 99: Please rename to `elbow_joint_command`

tylerjw:
As to the name of the files.  I think JogArm is a good name and it doesn't need the word Interface because there is no separate interface from the implementation.  There is no longer a difference between the C++ and ROS interface.  To use this you create a JogArm object with a valid planning scene (and config) and call start().  Once you do that you can start sending commands.

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
tylerjw: Thank you for pointing this out.  I'd updated this in my profile to work correctly but hadn't updated here.  There is now a comment to go along with this too.

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
tylerjw: done

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -91,156 +129,161 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables)
...
 
-          if (!jointJogCalcs(joint_deltas, shared_variables))
-            continue;
-        }
+  // Input frame determined by YAML file if not passed with message
+  if (twist_stamped_.header.frame_id.empty())
+  {
+    twist_stamped_.header.frame_id = parameters_.robot_link_command_frame;
+  }
+  if (twist_stamped_.header.frame_id.empty())
```
AdamPettinger, line 208: This if statement is the same as the one above it. I believe it was intended to be `if (joint_jog_.header.frame_id.empty())`, in which case I am not sure it is needed as jogging in the joint space does not have a frame or require a transformation

tylerjw: Good catch, I'll remove this one.

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -67,57 +70,57 @@ int main(int argc, char** argv)
...
-    base_joint_command.header.stamp = ros::Time::now();
-    jog_interface.provideJointCommand(base_joint_command);
+    // Make a joint command
+    // Messages are sent to jogger as boost::shared_ptr to enable zero-copy message_passing.
+    // Because this message is not coppied we should not modify it after we send it.
+    auto elbow_joint_command = boost::make_shared<control_msgs::JointJog>();
+    elbow_joint_command->header.stamp = ros::Time::now();
+    elbow_joint_command->joint_names.push_back("elbow_joint");
+    elbow_joint_command->velocities.push_back(0.2);
+    elbow_joint_command->header.stamp = ros::Time::now();
```
jliukkonen, line 112: Is this line about time stamping duplicate for a reason?

moveit_experimental/moveit_jog_arm/src/jog_arm.cpp
```diff
@@ -0,0 +1,426 @@
...
+  twist_stamped_pub_ = nh_.advertise<geometry_msgs::TwistStamped>(parameters_.cartesian_command_in_topic, 1);
+  joint_jog_pub_ = nh_.advertise<control_msgs::JointJog>(parameters_.joint_command_in_topic, 1);
+
+  // ROS Server for allowing drift in some dimensions
+  drift_dimensions_server_ =
+      nh_.advertiseService(nh_.getNamespace() + "/" + ros::this_node::getName() + "/change_drift_dimensions",
+                           &JogArm::changeDriftDimensions, this);
+
+  // ROS Server for changing the control dimensions
+  dims_server_ =
```
jliukkonen, line 73: This could be called `control_dimensions_server_`.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
jliukkonen: Can you explain a bit more how this data is now modified and accessed? If it is just one thread accessing the values, then it is no longer shared data and that comment shouldn't apply anymore. On the other hand, if there are multiple threads accessing the variables, then won't there definitely be RMW operations on them?

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
tylerjw: no, good catch

moveit_experimental/moveit_jog_arm/src/jog_arm.cpp
AdamPettinger: I agree and like this change

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -917,4 +921,32 @@ void JogCalcs::jointJogCB(const control_msgs::JointJogConstPtr& msg)
...
 
+bool JogCalcs::changeDriftDimensions(moveit_msgs::ChangeDriftDimensions::Request& req,
+                                     moveit_msgs::ChangeDriftDimensions::Response& res)
+{
+  shared_variables_.drift_dimensions[0] = req.drift_x_translation;
+  shared_variables_.drift_dimensions[1] = req.drift_y_translation;
+  shared_variables_.drift_dimensions[2] = req.drift_z_translation;
+  shared_variables_.drift_dimensions[3] = req.drift_x_rotation;
+  shared_variables_.drift_dimensions[4] = req.drift_y_rotation;
+  shared_variables_.drift_dimensions[5] = req.drift_z_rotation;
```
tylerjw, line 932: I tried making these member variables of JogCalcs and it fails a test where it is expecting a warning.  I haven't figured out why so for now they are staying in the shared_variables.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm_data.h
tylerjw: Go search for how it is used.  These are only ever written by one thread and only read by another.  There is never a case where two different threads write for these.  I think we should let this sit for now as it works.  It is confusing and I plan on refactoring this structure out of the system into something easier to reason about later.

tylerjw: Ok, shared_variables is gone.  I replaced it with ros pub/sub using zero-copy messaging.  I profiled this and it produced a similar profile as before.  Now there should be no question of thread safety as there are no mutexes used for communication between the three modules.  I still use a mutex inside the modules to protect some of the pointer copies.  I am unsure if this is necessary but seems to have to affect.

tylerjw:
My latest commit did something I said I wasn't going to do but thought would make it even simpler.  Gone is the SharedVariables struct and in the place of it is ROS pub/sub.  I also minimally extended the c++ test.  I did this by replicating some of the functionality in one of the ros tests.  This tests the c++ interface which only differs from the ROS interface in a few functions.

One of those functions is `setPaused`.  I did replicate, and now test, the previous behavior of setting this.  However, I'm not sure that is the correct behavior.  When this is set the nodes all stop calculating except for the low pass filter in the JogCalcs thread.  Publishing stops abruptly when you command a pause.  There is no halt that happens here.  This may be the intended behavior if this is combined with some other form of trajectory control, I'm not sure.

henningkayser:
> One of those functions is `setPaused`. I did replicate, and now test, the previous behavior of setting this. However, I'm not sure that is the correct behavior. When this is set the nodes all stop calculating except for the low pass filter in the JogCalcs thread. Publishing stops abruptly when you command a pause. There is no halt that happens here. This may be the intended behavior if this is combined with some other form of trajectory control, I'm not sure.

As I understand `setPaused()` should really only pause computations and keep the joint state updated so that you can easily resume running controls without having to launch JogArm again. It's the responsibility of the user to halt the motion before calling this function.

> I profiled the latest commit and the number didn't change. The zero-copy message passing from ROS is really cheap and a nice way to communicate between different components or threads in a program.

Nice

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -153,35 +161,114 @@ class JogCalcs
...
 
-  ros::Rate default_sleep_rate_;
+  // True -> allow drift in this dimension. In the command frame. [x, y, z, roll, pitch, yaw]
+  std::array<bool, 6> drift_dimensions_ = { { false, false, false, false, false, false } };
+
+  // The dimesions to control. In the command frame. [x, y, z, roll, pitch, yaw]
+  std::array<bool, 6> control_dimensions_ = { { true, true, true, true, true, true } };
+
+  // Amount we sleep when waiting
+  ros::Rate default_sleep_rate_ = 1000;
```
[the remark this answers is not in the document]
tylerjw: I left this alone, it was just copy-pasted from somewhere and is the rate we check on things when first starting.  I'll lower it as I agree it is large.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/collision_check_thread.h
```diff
@@ -54,18 +57,58 @@ class CollisionCheckThread
...
+  collision_detection::CollisionResult collision_result_;
+
+  // ROS
+  ros::Timer timer_;
+  ros::Duration period_;
+  ros::Subscriber joint_state_sub_;
+  ros::Publisher collision_velocity_scale_pub_;
+
+  // Latest joint state, updated by ROS callback
+  mutable std::mutex joint_state_mutex_;
```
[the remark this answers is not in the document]
tylerjw: This way is better I think because it uses only ROS for this sort of thread-safe communication and it is already implemented very nicely for us.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/boost_pool_allocation.h
```diff
@@ -36,25 +36,14 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 *******************************************************************************/
 
-// Server node for arm jogging with MoveIt.
-
 #pragma once
 
-#include "jog_interface_base.h"
+#include <boost/pool/pool_alloc.hpp>
```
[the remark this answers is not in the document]
tylerjw: I'm not sure it even belongs in jog_arm proper and not in some utility project.  Figured it was better to separate it because it isn't really related other than that jog_arm is using it.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/collision_check_thread.h
```diff
@@ -54,18 +57,58 @@ class CollisionCheckThread
    *  \param planning_scene_monitor: PSM should have scene monitor and state monitor
    *                                 already started when passed into this class
    */
-  CollisionCheckThread(const moveit_jog_arm::JogArmParameters& parameters,
+  CollisionCheckThread(ros::NodeHandle& nh, const moveit_jog_arm::JogArmParameters& parameters,
```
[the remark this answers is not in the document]
tylerjw: You can't do nearly as much with a const NodeHandle.  You can subscribe but you can't start a service or advertise a publisher.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/collision_check_thread.h
```diff
@@ -54,18 +57,58 @@ class CollisionCheckThread
...
 
-  // Get thread-safe read-only lock of planning scene
-  planning_scene_monitor::LockedPlanningSceneRO getLockedPlanningSceneRO() const;
+  /** \brief start and stop the Thread */
+  void start();
+  void stop();
 
-  void startMainLoop(moveit_jog_arm::JogArmShared& shared_variables);
+  /** \brief Pause or unpause processing jog commands while keeping the threads alive */
+  void setPaused(bool paused);
```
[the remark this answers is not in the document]
tylerjw: This is one part of the API I left alone.  I don't think `setPaused` is a bad name.  It allows you to locally have a paused variable you set and then pass into this so you have a way of tracking the state locally.

I see start/stop more as methods you use on startup and shutdown.

moveit_experimental/moveit_jog_arm/src/jog_server.cpp
```diff
@@ -37,15 +37,50 @@
...
  */
 
-#include <moveit_jog_arm/jog_ros_interface.h>
+#include <moveit_jog_arm/jog_arm.h>
 
-static const std::string LOGNAME = "jog_server";
+namespace
+{
+constexpr char LOGNAME[] = "jog_server";
+constexpr char ROS_THREADS = 8;
```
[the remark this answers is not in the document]
tylerjw: Before this was 4 and then we started 3 more threads.  So in a way, I'm only adding 1 thread to the pool and most processors can actually use 8 threads if they all have something to do.  The more threads you use (assuming your processor has enough cores to run them concurrently) the quicker you'll respond to them being unblocked and therefore the better utilization of your machine and the whole thing will go idle faster.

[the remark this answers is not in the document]
tylerjw: To be consistent I increased the number of threads in the example to 8.

moveit_experimental/moveit_jog_arm/src/jog_arm.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+void JogArm::jointTrajectoryCB(const trajectory_msgs::JointTrajectoryConstPtr& msg)
+{
+  const std::lock_guard<std::mutex> lock(latest_state_mutex_);
+  latest_joint_trajectory_ = msg;
+}
+
+void JogArm::okToPublishCB(const std_msgs::BoolConstPtr& msg)
+{
+  const std::lock_guard<std::mutex> lock(latest_state_mutex_);
```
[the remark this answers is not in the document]
tylerjw: no, good catch... fixed

tylerjw:
One nice side-effect of this change is when you break functionality in jog_arm and have to debug it (as the tests have found several times in this change) you can use a ros_bag to debug it as you can even subscribe to the internal messages.

This does have performance implications because once you do that those messages are no longer nearly free as they have to be serialized and sent over the network to the bag node but it is a nice way to debug something that only changes when running all the threads at once interacting with eachother (harder to do in a debugger nicely).

gavanderhoorn:
> This does have performance implications because once you do that those messages are no longer nearly free as they have to be serialized and sent over the network to the bag node

Well, not if you use [osrf/nodelet_rosbag](https://github.com/osrf/nodelet_rosbag) ;)

tylerjw:
> Well, not if you use [osrf/nodelet_rosbag](https://github.com/osrf/nodelet_rosbag) ;)

Cool!  One thing I'm going to do is create a nodelet interface for jog_arm after this PR so people can add jog_arm to their nodelet based projects easily (just include it in their launch file and talk to the right topics).  I think that'll be the ideal way for people to integrate it.

moveit_experimental/moveit_jog_arm/src/collision_check_thread.cpp
```diff
@@ -46,108 +49,139 @@ namespace moveit_jog_arm
...
+
+void CollisionCheckThread::stop()
+{
+  timer_.stop();
+}
+
+void CollisionCheckThread::run(const ros::TimerEvent& timer_event)
+{
+  // Log last loop duration and warn if it was longer than period
+  if (timer_event.profile.last_duration.toSec() < period_.toSec())
```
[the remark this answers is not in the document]
tylerjw: One thing I could change is publish these durations to an internal topic that is never subscribed to.  That would be almost free and would allow introspection through a bag if you cared (although that would result in a performance hit for measuring performance).  The downsides to that approach could be mitigated by @gavanderhoorn's suggestion down below about using nodelet_bag (if jog_arm was a nodelet itself).

tylerjw:
jog_server

Some more minor changes:

1. Conditional log statement in each of the timers to warn when the last iteration took longer than the period, removed the throttled one as it just adds noise to the logs.
1. JointTrajectory is published with a boost::shared_ptr from pool for zero-copy.
1. ok_to_publish was converted from pub/sub to just a bool with an accessor.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/boost_pool_allocation.h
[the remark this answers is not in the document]
tylerjw: I made this change (moveit::util namespace), lmk what you think?

tylerjw:
Ok, I'm done making further changes here (aside from responding to feedback).  I went ahead and implemented converting this to a nodelet in the next PR but left that out of this one.  I did that with hopes that this one would be easier to review as it just focuses on the communication and not on the layout of the executables as much.

moveit_experimental/moveit_jog_arm/CMakeLists.txt
tylerjw: There is no main any more (in the follow-on PR) except in the cpp demo.  After this PR I implemented a nodelet interface which is better than a main because while you can still run it by itself (see changes in launch files on how) you can also run it in the same process with your client and get all the zero-copy goodness easily (without having to compile against jog_arm as it is now a shared library).

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -467,28 +494,22 @@ trajectory_msgs::JointTrajectory JogCalcs::composeJointTrajMessage(sensor_msgs::
...
 // Apply velocity scaling for proximity of collisions and singularities.
-// Scale for collisions is read from a shared variable.
-void JogCalcs::applyVelocityScaling(JogArmShared& shared_variables, Eigen::ArrayXd& delta_theta,
-                                    double singularity_scale)
+void JogCalcs::applyVelocityScaling(Eigen::ArrayXd& delta_theta, double singularity_scale)
 {
-  shared_variables.lock();
-  double collision_scale = shared_variables.collision_velocity_scale;
-  shared_variables.unlock();
+  double collision_scale = collision_velocity_scale_;
```
AndyZe, line 503: I don't think this intermediate variable `collision_scale` is necessary, just use `collision_velocity_scale_`

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -795,6 +808,7 @@ Eigen::VectorXd JogCalcs::scaleJointCommand(const control_msgs::JointJog& comman
     }
     catch (const std::out_of_range& e)
     {
+      const std::lock_guard<std::mutex> lock(latest_state_mutex_);
```
AndyZe, line 811: Isn't this already locked at Line 733?

AndyZe:
@tylerjw, this PR is too big to review effectively. For example, since you renamed collision_check.cpp, I can’t tell if you changed anything within that file. Renaming that file should have been a separate PR.

At a high level, it makes sense to combine the “interface” thread and the calculations thread, but keep the collision-checking thread separate.

I think the previous method of sharing data was easier to follow, i.e. one common data structure. Now, the data sharing is a bit scattered between many different methods (ROS pubs/subs, plain old double...).  But, I guess the performance gain is worth it.

AndyZe:
Not criticizing this, but just taking notes: CPU load increased from ~35% to ~41% on my PC. I think this is a fair trade given that the calculations are happening faster.

tylerjw:
> this PR is too big to review effectively. For example, since you renamed collision_check.cpp, I can’t tell if you changed anything within that file. Renaming that file should have been a separate PR.

You are right.  I will take a stab at reducing the number of unrelated changes and provide a diagram to hopefully make the communication clearer.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -90,167 +134,164 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables)
...
+  wait_for_jog_commands_ = true;
+  have_nonzero_twist_stamped_ = false;
+  have_nonzero_joint_jog_ = false;
+  have_nonzero_command_ = false;
+}
+
+void JogCalcs::start()
+{
+  stop_requested_ = false;
+  status_pub_ = nh_.advertise<std_msgs::Int8>(parameters_.status_topic, ROS_QUEUE_SIZE);
```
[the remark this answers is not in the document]
tylerjw: I did this so you could wait on the status thread to be ready to signify everything was initialized.  I'm not sure this is good or standard.  Before it was waiting on the services but because I moved them into the constructor here it messed up the timing of some of the tests.  I think maybe the right approach is to wait for the first publish of status and put this back in the constructor.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
tylerjw: I did this so it couldn't be changed "underneath" this function between tests.  If that happened it would make this function confusing and probably broken.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
tylerjw: I don't think so, but this lock_guard is in a weird place... I need to go through the PR and map out how I used them to make sure I didn't create a deadlock.  I'll do that.

tylerjw:
> At a high level, it makes sense to combine the “interface” thread and the calculations thread, but keep the collision-checking thread separate.

Looking at it now the "main" thread (timer) just formats the output based on some setting and could be completely removed as it operates at the same rate as the "thread" that produces the output (JogCalcs) and is therefore redundant.  All the logic is contained within JogCalcs to prepare and output all the data (since it is just going out over a ros publisher anyway).

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
tylerjw: I looked at this and it doesn't do that.  I'm not sure how the code can trace from 733 to here at all.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -90,167 +134,164 @@ void JogCalcs::startMainLoop(JogArmShared& shared_variables)
     position_filters_.emplace_back(parameters_.low_pass_filter_coeff);
   }
 
-  // Initialize the position filters to initial robot joints
-  while (!updateJoints(shared_variables) && ros::ok())
+  // initialization is finished
+  wait_for_jog_commands_ = true;
+  have_nonzero_twist_stamped_ = false;
+  have_nonzero_joint_jog_ = false;
+  have_nonzero_command_ = false;
```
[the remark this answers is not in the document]
tylerjw: They are, removed these.

tylerjw:
jog_server

After my latest push this is what the pub/sub of jog_arm looks like.  

In the end almost all bools (except pause) are unique to each worker (calcs and collision) and were just moved into each object themselves.

The only internal ros pub/sub going on now is the double value from the collision timer to the calcs timer.

Externally there are these interfaces for control which are the same ones that existed before:
* delta_jog_commands & joint_delta_jog_commands (inputs)
* status
* command (output)

In my nodelets change, one more thing is put through roscpp and taken out of the hands of the cpp function calls (pause is in a service).  I'm trying to stop further refactors in this one as it is in a really nice state right now.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/collision_check.h
```diff
@@ -38,34 +38,77 @@
...
 
 private:
+  void run(const ros::TimerEvent& timer_event);
+  planning_scene_monitor::LockedPlanningSceneRO getLockedPlanningSceneRO() const;
+  void jointStateCB(const sensor_msgs::JointStateConstPtr& msg);
+
+  // ROS node handle
+  ros::NodeHandle nh_;
+
+  // Parameters
```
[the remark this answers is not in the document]
tylerjw: done

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/collision_check.h
```diff
@@ -38,34 +38,77 @@
...
-  void startMainLoop(moveit_jog_arm::JogArmShared& shared_variables);
+  /** \brief Pause or unpause processing jog commands while keeping the timers alive */
+  void setPaused(bool paused);
 
 private:
+  void run(const ros::TimerEvent& timer_event);
+  planning_scene_monitor::LockedPlanningSceneRO getLockedPlanningSceneRO() const;
+  void jointStateCB(const sensor_msgs::JointStateConstPtr& msg);
+
+  // ROS node handle
```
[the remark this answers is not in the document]
tylerjw: done

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_arm.h
```diff
@@ -36,78 +36,90 @@
...
+#include "jog_arm_parameters.h"
+#include "jog_calcs.h"
+#include "low_pass_filter.h"
+
 namespace moveit_jog_arm
 {
 /**
- * Class JogInterfaceBase - Base class for C++ interface and ROS node.
- * Handles ROS subs & pubs and creates the worker threads.
+ * Class JogArm - Base class for C++ interface and ROS node.
```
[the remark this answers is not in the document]
tylerjw: What do you think of the updated description for it?

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -56,32 +63,40 @@ namespace moveit_jog_arm
 class JogCalcs
 {
 public:
-  JogCalcs(const JogArmParameters& parameters, const robot_model_loader::RobotModelLoaderPtr& model_loader_ptr);
+  JogCalcs(ros::NodeHandle& nh, const JogArmParameters& parameters,
+           const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor);
 
-  void startMainLoop(JogArmShared& shared_variables);
+  /** \brief Start and stop the timer */
```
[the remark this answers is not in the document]
tylerjw: I added more to the comment to make it clearer what they do.

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -36,19 +36,48 @@
...
+public:
+  StatusMonitor(ros::NodeHandle& nh, const std::string& topic)
+  {
+    sub_ = nh.subscribe(topic, 1, &StatusMonitor::statusCB, this);
+  }
+
+private:
+  void statusCB(const std_msgs::Int8ConstPtr& msg)
+  {
+    moveit_jog_arm::StatusCode latest_stats = static_cast<moveit_jog_arm::StatusCode>(msg->data);
```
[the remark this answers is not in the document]
tylerjw: good catch, fixed

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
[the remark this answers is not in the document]
tylerjw: I was wrong, they already do wait for a first message and this doesn't need to be here.  I moved it with the other ros message interface stuff in the constructors.  Good catch.

tylerjw:
> I find it hard to follow what is shared and what isn't.

The answer is that almost nothing is shared now between the two timers (collision and calcs).  They both only operate on their own data and there is one channel of communication between them through the `internal/collision_velocity_scale` method.

tylerjw:
> At a high level, it makes sense to combine the “interface” thread and the calculations thread, but keep the collision-checking thread separate.

I did this and I wonder if it lowers your CPU usage when running it now.  It should because it eliminates an extra internal queue and timer.

AndyZe:
OK, I tested it out again. It seems fine! Combining those two threads/timers did seem to reduce CPU usage a bit, it's basically indistinguishable from where it was previously. :+1:

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -66,58 +95,61 @@ int main(int argc, char** argv)
...
 
   // Send a few Cartesian velocity commands
   while (ros::ok() && num_commands < 200)
   {
-    ++num_commands;
-    velocity_msg.header.stamp = ros::Time::now();
-    jog_interface.provideTwistStampedCommand(velocity_msg);
+    // Make a Cartesian velocity message
+    // Messages are sent to jogger as boost::shared_ptr to enable zero-copy message_passing.
+    // Because this message is not coppied we should not modify it after we send it.
```
rhaschke, line 118:
```suggestion
    // Because this message is not copied we should not modify it after we send it.
```

moveit_experimental/moveit_jog_arm/src/cpp_interface_example/cpp_interface_example.cpp
```diff
@@ -66,58 +95,61 @@ int main(int argc, char** argv)
...
   // Send a few joint commands
   num_commands = 0;
   while (ros::ok() && num_commands < 200)
   {
-    ++num_commands;
-    base_joint_command.header.stamp = ros::Time::now();
-    jog_interface.provideJointCommand(base_joint_command);
+    // Make a joint command
+    // Messages are sent to jogger as boost::shared_ptr to enable zero-copy message_passing.
+    // Because this message is not coppied we should not modify it after we send it.
```
rhaschke, line 141:
```suggestion
    // Because this message is not copied we should not modify it after we send it.
```

moveit_experimental/moveit_jog_arm/test/jog_cpp_interface_test.cpp
```diff
@@ -45,57 +45,152 @@
...
+TEST_F(JogArmFixture, SendTwistStampedTest)
+{
+  jog_arm_->start();
+  EXPECT_TRUE(waitForFirstStatus()) << "Timeout waiting for Status message";
+
+  auto parameters = jog_arm_->getParameters();
+
+  // count trajectory messages sent by jog_arm
+  size_t received_count = 0;
+  boost::function<void(const trajectory_msgs::JointTrajectoryConstPtr&)> traj_callback =
```
rhaschke, line 113: use `auto` for the type here?

moveit_experimental/moveit_jog_arm/test/jog_cpp_interface_test.cpp
```diff
@@ -45,57 +45,152 @@
...
+TEST_F(JogArmFixture, SendJointJogTest)
+{
+  jog_arm_->start();
+  EXPECT_TRUE(waitForFirstStatus()) << "Timeout waiting for Status message";
+
+  auto parameters = jog_arm_->getParameters();
+
+  // count trajectory messages sent by jog_arm
+  size_t received_count = 0;
+  boost::function<void(const trajectory_msgs::JointTrajectoryConstPtr&)> traj_callback =
```
rhaschke, line 153: use `auto` for the type here?

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/make_shared_from_pool.h
```diff
@@ -36,25 +36,21 @@
...
-namespace moveit_jog_arm
+namespace moveit
+{
+namespace util
 {
-/**
- * Class JogROSInterface - Instantiated in main(). Handles ROS subs & pubs and creates the worker threads.
- */
-class JogROSInterface : protected JogInterfaceBase
+// Useful template for creating messages from a message pool
```
rhaschke, line 47: Interesting feature. Where is the pool (size) defined?
Is such a feature available from C++14's standard lib as well? The pool allocator is the only reason to pull in boost and the ROS community strives to replace boost with std features.

moveit_experimental/moveit_jog_arm/test/jog_cpp_interface_test.cpp
tylerjw: I can't or it'll make a std::function and not a boost::function and ROS won't accept that.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/make_shared_from_pool.h
tylerjw: The pool is allocated by boost as needed using a singleton per type.  Because I used boost::allocate_shared it only does one allocation (both the message and the control block of the shared pointer).  I used a tool to print when malloc is being called and on my machine, the first call to this allocates space for 32 of that type.  After that it will double that storage if I ever am using all 32 and ask for another, and so on and so forth.  The big win here is that you can have all the goodness of the zero-copy from roscpp without the performance hit from memory fragmentation and system calls in every loop iteration (assuming the subscriber doesn't hold onto a ton of them).

C++14 doesn't offer any equivalent functionality to this as far as I know.

moveit_experimental/moveit_jog_arm/test/jog_cpp_interface_test.cpp
tylerjw: See previous comment.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/make_shared_from_pool.h
tylerjw: Also, from my reading of roscpp it seems that the zero-copy implementation depends on a boost::shared_ptr type, not that it couldn't be implemented using std::shared_ptr but that's what I found.

I generally agree that replacing boost with std features where appropriate is the right call.  For now, this functionality doesn't exist in the standard library.

tylerjw: Here is my conversation with @gavanderhoorn about this: https://answers.ros.org/question/353198/pre-allocation-of-messages-for-nodelets/

Here is the documentation on boost::pool_allocator:  https://www.boost.org/doc/libs/1_73_0/libs/pool/doc/html/boost/pool_allocator.html

I used the further specialization `fast_pool_allocator` which is optimized for single allocations (think list and not vector) as I'm never asking for an array of new messages.

tylerjw: `boost` is doing some magic with the type passed in here: `using allocator_t = boost::fast_pool_allocator<boost::shared_ptr<T>>;`  That type is used as a tag for a singleton pool of memory.  They can't deduce the size from this type and instead it is set the first time I create a new chunk from it on the next line.

I don't like adding complexity just for the fun of it.  I think in this case the tradeoffs you get from using this for allocating messages you send through ROS on the same process are completely worth it as it removes malloc from your loops most of the time (ideally only on your first iteration), solves memory fragmentation from calling malloc for messages inside loops, and the interface I created here looks very similar to the `make_shared` everyone is used to already.

rhaschke: I'm all fine with this and already marked the conversation as resolved.
Just a pity that the C++ standard library doesn't support this yet.

tylerjw: Unrelated, what are you thoughts on switching to c++17 in master only (and maybe Noetic as it would be a real stretch to install Noetic on 16.04)?

rhaschke: Noetic still sticks with C++14:
https://www.ros.org/reps/rep-0003.html#noetic-ninjemys-may-2020-may-2025

AndyZe:
@tylerjw it's up to you of course, but IMO it's not worth fixing things for Kinetic. It's going to be EOL in another year anyway

tylerjw:
> @tylerjw it's up to you of course, but IMO it's not worth fixing things for Kinetic. It's going to be EOL in another year anyway

The issue where it fails on kinetic is a segfault resulting from a shared pointer being deleted while I'm trying to copy it's the internals (in a callback where I only have a const ref to it and not an actual copy).   The correct way to handle this is to increase the ref-count so it doesn't get deleted by copying the shared ptr instead of the data inside.  That is what my fix does.  I think this is only an issue with how roscpp works in kinetic and doesn't apply to melodic.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/make_shared_from_pool.h
tylerjw: That's sad... all modern compilers (even MVCPP) fully supports c++17 now.

tylerjw:
I just rebased on master as a result of Andy's latest changes.  If this passes CI I believe it is ready as I have tested this on a client project.

moveit_experimental/moveit_jog_arm/include/moveit_jog_arm/jog_calcs.h
```diff
@@ -153,35 +163,112 @@ class JogCalcs
...
+  ros::Subscriber joint_state_sub_;
+  ros::Subscriber twist_stamped_sub_;
+  ros::Subscriber joint_jog_sub_;
+  ros::Subscriber collision_velocity_scale_sub_;
   ros::Publisher status_pub_;
-
-  StatusCode status_ = NO_WARNING;
-
-  JogArmParameters parameters_;
+  ros::Publisher joint_trajectory_pub_;
```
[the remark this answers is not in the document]
tylerjw: good catch, this was something that I removed the need for.

tylerjw:
My latest change here is fixing an issue with how kinetic deals with multiple subscribers on one topic.  Basically it is broken if you try to use const boost::shared_ptr to keep the lifetime of the message around and just copy the pointer in kinetic.  Thank you @AdamPettinger for debugging that.

v4hn:
This pull-requests keeps growing and is already almost impossible to review from the outside.

Assuming you tested this in your own projects, isn't it about time to merge it?

tylerjw:
I believe this is working correctly however there is some transient deadlock that causes tests to fail ~30% of the time in the project I'm using this PR in for testing (I don't think it is as a result of this pr, however, I can't know until I isolated the cause of the deadlock).  I am hoping to have found the cause of that soon (it has been all I've been working on for several days) so I can know this isn't the cause of it (and then we can merge this).

v4hn:
I remember you advertised this patch as "makes debugging asynchronous issues much simpler" :smile: 

Sounds good.

moveit_experimental/moveit_jog_arm/test/jog_cpp_interface_test.cpp
```diff
@@ -45,57 +45,152 @@
...
+    msg->header.stamp = ros::Time::now();
+    msg->header.frame_id = "panda_link3";
+    msg->velocities.push_back(0.1);
+
+    // Send the message
+    joint_jog_pub.publish(msg);
+    publish_rate.sleep();
+  }
+
+  EXPECT_GT(received_count, num_commands - 20);
```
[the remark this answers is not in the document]
tylerjw: I didn't change this, I copy-pasted this from a python test.

moveit_experimental/moveit_jog_arm/src/jog_server.cpp
```diff
@@ -37,15 +37,50 @@
...
+
+  // Load the planning scene monitor
+  auto planning_scene_monitor = std::make_shared<planning_scene_monitor::PlanningSceneMonitor>("robot_description");
+  if (!planning_scene_monitor->getPlanningScene())
+  {
+    ROS_ERROR_STREAM_NAMED(LOGNAME, "Error in setting up the PlanningSceneMonitor.");
+    exit(EXIT_FAILURE);
+  }
+
+  // Start the planning scene monitor
```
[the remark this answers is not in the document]
tylerjw: That is a good idea, but I'd like to do that in another PR.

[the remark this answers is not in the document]
henningkayser: makes sense

PR #2104, opened 2020-05-23
prevent collision checking segfault if octomap has NULL root pointer

JStech:
### Description

After clearing the octomap, the root node is set to NULL, which then causes FCL to segfault in collision checking. Maybe it's actually an FCL bug, but this was the most straightforward way I found to work around it.

### Checklist

### DIFF
--- a/moveit_core/collision_detection_fcl/src/collision_common.cpp
+++ b/moveit_core/collision_detection_fcl/src/collision_common.cpp
@@ -517,10 +517,11 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
   }
 
   fcl_result.min_distance = dist_threshold;
+  // fcl::distance segfaults when given an octree with a null root pointer (using FCL 0.6.1)
   if ((o1->getObjectType() == fcl::OT_OCTREE &&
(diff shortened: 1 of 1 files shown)


JStech:
Thanks for the guidance. I think I've made the modifications you described.

moveit_core/collision_detection_fcl/src/collision_common.cpp
```diff
@@ -517,6 +517,13 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
   }
 
   fcl_result.min_distance = dist_threshold;
+  if ((o1->getObjectType() == fcl::OT_OCTREE &&
+       !std::static_pointer_cast<const fcl::OcTreed>(o1->collisionGeometry())->getRoot()) ||
+      (o1->getObjectType() == fcl::OT_OCTREE &&
+       !std::static_pointer_cast<const fcl::OcTreed>(o1->collisionGeometry())->getRoot()))
```
v4hn, line 523: I guess the second check should be about `o2`, right?
Also, please add a comment explaining that `fcl::distance` misbehaves for this in the version you use.

JStech: Ah yes, I copy-pasted that and failed to change o1 to o2. Done.

Issue #2097, opened 2020-05-19
Adding collision checking between two collision objects in a planning scene


pradeepr-roboticist:
**Is your feature request related to a problem? Please describe.**
I would like to propose a new feature that allows collision checking between two objects ("object_A", "object_B") in the planning scene. Right now, collision queries can only happen between robot links and world objects. But, collision between world objects cannot be queried.

I recently posted this as a question at [https://answers.ros.org/question/351846/collision-between-planning-scene-objects-eg-collisionobjecta-and-collisionobjectb/?answer=352800#post-id-352800](url).

**Describe the solution you'd like**
After looking at the comments for the question I posted and following advice of the answers, I feel that there should be a new public virtual function in "collision_env.h" called "checkObjectPairCollision".

I have already implemented this function in collision_env_fcl.cpp. The basic idea is to put "object_A" into a FCL world manager called "manager1". And, "object_B" into another world manager called "manager2". And, finally query the "manager1->collide(manager2, ...)".

Please visit the question given in the link above for more details.

I would like to contribute a more elaborate version of the above code. I wanted a go ahead signal before I proceeded with a pull request.

My worries are how this is going to be supported for other collision detectors (i.e., bullet).

**Describe alternatives you've considered**
Here are some of my ideas so far. All of them have some issue.

1) I tried digging through code related AllowedCollisionMatrix. It has methods such as "setDefaultEntry". As far as I understood, it useful only for ignoring collision between robot and a specific object in the scene.

2) A hacky solution is to put the robot in a known non-colliding state. Then, temporarily attach "CollisionObjectA" to the robot. And, see if it collides with "CollisionObjectB" using the existing functions in the API. This solution seems like the easiest but it is also ugly.

3) Define "CollisionObjectA" as link attached to a 6 DOF joint co-located at the root of the robot. Make a planning group for the 6 DOF joint. Set the 6 DOF state of the "CollisionObjectA" and query collision via the existing API (i.e., isValid ). Though it theoretically makes sense, I guess that there can be complications when solving IK for other planning groups.

4) Instantiate FCL collision manager object on my own and perform the necessary collision checks. This has the drawback of ensuring the collision object A and B in planning scene are in sync with the same in my FCL collision manager.

felixvd:
Can you give an example of the use cases you are envisioning? I don't know how many people would use it, and this runs the risk of duplicating collision checking APIs that need to be maintained afterwards.

pradeepr-roboticist:
[...]

PR #2105, opened 2020-05-24
checking collision between two sets of objects in the world

pradeepr-roboticist:
### Description

This PR is related to #2097 .
Here are the changes made:

A new virtual public function (checkCollisionBetweenObjectGroups) was added to collision_env.h.
This function provides the Boolean result of a collision check between objects of one group and another group. Each group is a vector of strings denoting the name of the objects in the group.

A concrete implementation is provided in "collision_env_fcl.cpp" for the new function.
The implementation instantiates a FCL CollisionManager for each group and populates the collision manager with the corresponding objects.
Finally, a collision query is made to the CollisionManager and the result is returned.

A test case is also provided in "test_fcl_env.cpp".
 
### Checklist

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -15,6 +15,7 @@ API changes in MoveIt releases
 - Requests to `get_planning_scene` service without explicitly setting "components" now return full scene
 - `moveit_ros_planning` no longer depends on `moveit_ros_perception`
 - `CollisionRobot` and `CollisionWorld` are combined into a single `CollisionEnv` class. This applies for all derived collision checkers as `FCL`, `ALL_VALID`, `HYBRID` and `DISTANCE_FIELD`. Consequently, `getCollisionRobot[Unpadded] / getCollisionWorld` functions are replaced through a `getCollisionEnv` in the planning scene and return the new combined environment. This unified collision environment provides the union of all member functions of `CollisionRobot` and `CollisionWorld`. Note that calling `checkRobotCollision` of the `CollisionEnv` does not take a `CollisionRobot` as an argument anymore as it is implicitly contained in the `CollisionEnv`.
+- `CollisionEnv` provides a function `checkCollisionBetweenObjectGroups` that allows collision detection between two sets of objects that are already part of the world ([2097](https://github.com/ros-planning/moveit/issues/2097)).
 - `RobotTrajectory` provides a copy constructor that allows a shallow (default) and deep copy of waypoints
 
 ## ROS Melodic
--- a/moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp
+++ b/moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp
@@ -319,12 +319,20 @@ bool CollisionEnvFCL::checkCollisionBetweenObjectGroups(std::vector<std::string>
   for (auto it = object_group1.begin(); it != object_group1.end(); ++it)
   {
     auto const& object_name = *it;
+    if (fcl_objs_.count(object_name) == 0)
+      ROS_ERROR_STREAM_NAMED("collision_detection.fcl",
+                             "FCL object by the name: " << object_name << " does not exist.");
+
     auto fcl_obj = fcl_objs_.at(object_name);
     fcl_obj.registerTo(manager1.get());
   }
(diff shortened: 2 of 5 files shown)


MIGRATION.md
```diff
@@ -15,6 +15,7 @@ API changes in MoveIt releases
 - Requests to `get_planning_scene` service without explicitly setting "components" now return full scene
 - `moveit_ros_planning` no longer depends on `moveit_ros_perception`
 - `CollisionRobot` and `CollisionWorld` are combined into a single `CollisionEnv` class. This applies for all derived collision checkers as `FCL`, `ALL_VALID`, `HYBRID` and `DISTANCE_FIELD`. Consequently, `getCollisionRobot[Unpadded] / getCollisionWorld` functions are replaced through a `getCollisionEnv` in the planning scene and return the new combined environment. This unified collision environment provides the union of all member functions of `CollisionRobot` and `CollisionWorld`. Note that calling `checkRobotCollision` of the `CollisionEnv` does not take a `CollisionRobot` as an argument anymore as it is implicitly contained in the `CollisionEnv`.
+- `CollisionEnv` provides a function `checkCollisionBetweenObjectGroups` that allows collision detection between two sets of objects that are already part of the world ([2097](https://github.com/ros-planning/moveit/issues/2097)).
```
v4hn, line 18: No need for this entry. `MIGRATION.md` is meant to include only API-incompatible changes.

moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp
```diff
@@ -309,6 +309,45 @@ void CollisionEnvFCL::checkRobotCollisionHelper(const CollisionRequest& req, Col
...
+  auto manager1 = std::make_unique<fcl::DynamicAABBTreeCollisionManagerd>();
+  auto manager2 = std::make_unique<fcl::DynamicAABBTreeCollisionManagerd>();
+
+  // Populate managers with the FCL objects corresponding entries in object_groups
+  for (auto it = object_group1.begin(); it != object_group1.end(); ++it)
+  {
+    auto const& object_name = *it;
+    if (fcl_objs_.count(object_name) == 0)
+      ROS_ERROR_STREAM_NAMED("collision_detection.fcl",
+                             "FCL object by the name: " << object_name << " does not exist.");
```
v4hn, line 324:
```suggestion
                             "FCL object by the name: '" << object_name << "' does not exist.");
```

moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp
```diff
@@ -309,6 +309,45 @@ void CollisionEnvFCL::checkRobotCollisionHelper(const CollisionRequest& req, Col
   }
 }
 
+bool CollisionEnvFCL::checkCollisionBetweenObjectGroups(std::vector<std::string> const& object_group1,
+                                                        std::vector<std::string> const& object_group2) const
```
v4hn, line 313: I would expect the possibility to specify only *one* set of objects and check against everything else.

v4hn:
Also,

- I think it makes sense to provide the same interface for the `Distance` requests too. If we can test for collisions, we might as well allow to compute distances.
- In many places, the low-level APIs try to avoid the overhead of lots of string lookups. Is there a reasonable way to collect pointers to the fcl objects beforehand and reuse them across multiple requests?

moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp
pradeepr-roboticist: That sounds like a good idea. I will look into that as well.

pradeepr-roboticist:
> Also,
> 
> * I think it makes sense to provide the same interface for the `Distance` requests too. If we can test for collisions, we might as well allow to compute distances.

I will look into this feature as well.

> * In many places, the low-level APIs try to avoid the overhead of lots of string lookups. Is there a reasonable way to collect pointers to the fcl objects beforehand and reuse them across multiple requests?

Sure. I understand the performance issue with string lookups. I'll find a way to replace it with FCL object pointers.

felixvd:
@pradeepr-roboticist Any progress on this?

felixvd:
I'll close this for lack of activity. Feel free to reopen after addressing the changes.

PR #2116, opened 2020-05-28
Fix Melodic 1.0.3 release

rhaschke:
This attempts to fix broken API in the IKfast plugins, which slipped in with the backport merge #1998.
Unfortunately, it was missed by Travis, because of other issues (missing python-lxml).

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -957,8 +957,8 @@ as the new values that correspond to the group */
                  const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
                  const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions());
   [[deprecated("The attempts argument is not supported anymore.")]] bool
-  setFromIK(const JointModelGroup* group, const geometry_msgs::Pose& pose, unsigned int attempts, double timeout = 0.0,
-            const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
+  setFromIK(const JointModelGroup* group, const geometry_msgs::Pose& pose, unsigned int /*attempts*/,
+            double timeout = 0.0, const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
             const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions())
   {
     return setFromIK(group, pose, timeout, constraint, options);
@@ -976,7 +976,7 @@ as the new values that correspond to the group */
                  const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions());
   [[deprecated("The attempts argument is not supported anymore.")]] bool
   setFromIK(const JointModelGroup* group, const geometry_msgs::Pose& pose, const std::string& tip,
-            unsigned int attempts, double timeout = 0.0,
+            unsigned int /*attempts*/, double timeout = 0.0,
             const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
             const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions())
   {
@@ -993,8 +993,8 @@ as the new values that correspond to the group */
                  const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
                  const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions());
   [[deprecated("The attempts argument is not supported anymore.")]] bool
-  setFromIK(const JointModelGroup* group, const Eigen::Isometry3d& pose, unsigned int attempts, double timeout = 0.0,
-            const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
+  setFromIK(const JointModelGroup* group, const Eigen::Isometry3d& pose, unsigned int /*attempts*/,
+            double timeout = 0.0, const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
             const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions())
   {
     return setFromIK(group, pose, timeout, constraint, options);
@@ -1010,8 +1010,9 @@ as the new values that correspond to the group */
                  double timeout = 0.0, const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
                  const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions());
   [[deprecated("The attempts argument is not supported anymore.")]] bool
-  setFromIK(const JointModelGroup* group, const Eigen::Isometry3d& pose, const std::string& tip, unsigned int attempts,
-            double timeout = 0.0, const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
+  setFromIK(const JointModelGroup* group, const Eigen::Isometry3d& pose, const std::string& tip,
+            unsigned int /*attempts*/, double timeout = 0.0,
+            const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
             const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions())
   {
     return setFromIK(group, pose, tip, timeout, constraint, options);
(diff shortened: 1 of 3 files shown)


rhaschke:
@roehling, do you have an idea why catkin_lint 1.6.8 is complaining in our Travis job about a missing CATKIN_DEPENDS, while the very same catkin_lint version locally doesn't complain?
As none of the moveit_setup_assistant headers exposes an rviz header (checked via 
`find moveit_setup_assistant -iname "*.h" | xargs grep -e "#include.*rviz"`), I don't see a particular reason, why MSA should catkin-depend on rviz.

roehling:
> @roehling, do you have an idea why catkin_lint 1.6.8 is complaining in our Travis job about a missing CATKIN_DEPENDS, while the very same catkin_lint version locally doesn't complain?

This is required for all packages which export messages, services, or actions. `catkin_lint` used to infer this rather crudely from the package name (e.g., `sensor_msgs`), but it is now properly detected by looking for a dependency on `message_runtime`. This is queried from the `rosdep` database.

However, if `catkin_lint` cannot initialize `rosdep`, any check involving a `rosdep` query will be skipped, and you'll see a warning like `catkin_lint: cannot load rosdep database: ...` or `catkin_lint: cannot initialize rosdistro: ...` in the output.

If you don't get such a message, try running `rosdep update`. If the problem still persists, I'd be interested in a bug report, ideally one with the following directories zipped and included:
* `~/.ros/rosdep`
* `~/.ros/catkin_lint`
* `/etc/ros/rosdep`

It would also be instructive to know if the problem vanishes after a `catkin_lint --clear-cache` (but note that this will affect `~/.ros/catkin_lint`).

rhaschke:
After cleaning up the caches in `~/.ros/rosdep` and `~/.ros/catkin_lint`, catkin_lint reports the same issue locally now as well. So this is consistent at least.
However, I still don't understand why rviz should be a required build dependency in this case.
As far as I understand we need to list upstream packages in `catkin_package()` to declare a transitive dependency. However, as moveit_setup_assistant doesn't expose any rviz header in its own headers, it don't see the need for this here. Your explanation, @roehling wasn't particularly helpful in this regard, because rviz isn't related to messages, services, or actions.

roehling:
Surprisingly, [rviz exports](https://github.com/ros-visualization/rviz/blob/171eb3cfd1ec69cb6ff72da3ca41a1a4caa74eef/package.xml#L61) [a service](https://github.com/ros-visualization/rviz/blob/171eb3cfd1ec69cb6ff72da3ca41a1a4caa74eef/CMakeLists.txt#L187).

rhaschke:
That's true (since a few weeks). But why this triggers a `CATKIN_DEPENDS` dependency?
IMHO this should be only required if a header of the downstream package (here MSA) actually uses/includes any header from rviz (message, service, or any other header). What am I missing?

roehling:
That would be nice, and I'd implement it if there where a way to do this without basically creating a full C++ parser. Instead, I'll probably add a way to create a per-user or per-package configuration similar to `.pylintrc` that can be used to disable or change the severity of certain checks.

rhaschke:
> It just occurred to me that if you're not using any headers, you should not declare a build dependency (a plain <depend> implies a build dependency). If you use <exec_depend> instead, catkin_lint won't complain.

That's not what I meant. We are not exposing any rviz headers in our headers, but of course we use rviz headers in our cpp files. So, we require the build dependency.

Can I suppress this particular warning only? I tried both of the following, but neither works:
```
#catkin_lint: ignore_once missing_catkin_depend
#catkin_lint: ignore_once missing_catkin_depend (rviz)
```

roehling:
You probably need to use `ignore` for this, due to a limitation in `catkin_lint` reporting code.

Anyway, the check actually looks for a build_export dependency to trigger the error, so everything I wrote applies, just mentally substitute build_export for build.

JafarAbdi:
I tried changing [rviz](https://github.com/ros-planning/moveit/blob/master/moveit_setup_assistant/package.xml#L31) from `<depend>rviz</depend>` to `<build_depend>rviz</build_depend>` and now I don't get any error -- [build_depend](https://docs.ros.org/melodic/api/catkin/html/howto/format2/catkin_library_dependencies.html#build-depend) should fit our usage of rviz here, right .?

rhaschke:
> I tried changing rviz from `<depend>rviz</depend>` to `<build_depend>rviz</build_depend>`

If that works, let's go for it.

gavanderhoorn:
> As the release 1.0.3 version of MoveIt is broken, we might want to merge this not to the tip of melodic-devel, but onto 1.0.3 and go for an early release.

So you're suggesting to create a patch release on-top of `1.0.3`? With just the fix for the IKFast plugins?

Is there a big difference compared to the "tip of `melodic-devel`"?

> On the other hand, we are preparing the next Melodic release anyway. Looking at [#2066 (comment)](https://github.com/ros-planning/moveit/issues/2066#issue-614097210), it is planned for July 1st, which is too far in the future, I would say. Also, I believe we pushed this date even further (@tylerjw, can you confirm / update the mention issue?).

If we could "just" release the next release earlier, that would have my preference.

We could still release in July as well.

tylerjw:
I'm ok with a quicker release.  If my latest changes (and maybe the pending jog-arm rename) aren't in when you want to release can we pull jog arm out of this faster release and put it back for a July release?

gavanderhoorn:
> I'm ok with a quicker release. If my latest changes (and maybe the pending jog-arm rename) aren't in when you want to release can we pull jog arm out of this faster release and put it back for a July release?

I would say that would be good, yes.

Let's focus on fixing the immediate problem and then add other updates to the next release.

rhaschke:
> Let's focus on fixing the immediate problem and then add other updates to the next release.

That was my feeling as well. Thanks for the backing. As soon as somebody has accepted this PR, I will merge it onto 1.0.3 manually a prepare a new release.

gavanderhoorn:
> I will merge it onto 1.0.3 manually

Could you clarify what you mean exactly by this?

Do you intend to create a new branch starting with 1adbc5a0 and then apply the commits in this PR?

How would you create the release branch?

v4hn:
> That was my feeling as well. Thanks for the backing. As soon as somebody has accepted this PR, I will merge it onto 1.0.3 manually a prepare a new release

I believe @gavanderhoorn was not talking about this approach and would prefer a quicker release with some commits reverted.
If your proposed approach is actually feasible with the tools, I am all for it though.
I guess it might break some things in rosdistro if we release a version for melodic outside the `melodic-devel` branch?

Reverting multiple commits in `melodic-devel` on top of 1.0.3 is a bad idea, making the history unreadable. Especially because the main things are huge merge commits with multiple individual commits for feature backports.
Rushing the next release with multiple changes in critical structures does not seem like a good idea either.
We merged these *after* the release for a reason.

I am still wondering, why this issue was not discovered by the explicit ikfast-plugin test in CI...

v4hn:
I approved, but you might want to rebase. It says I would need administrator privileges to merge right away (Just because it's not branching from the latest commit?). Feel free to move ahead.

rhaschke:
> Could you clarify what you mean exactly by this (manually merge onto 1.0.3)?

* I squashed my results locally into a single commit, adding the final commit message
* I prepared the next release 1.0.4, locally only, not yet pushing to the github repo
* I merged this 1.0.4 release branch into latest melodic-devel, again locally
* I fast-forwarded the github repo to my final result
* I pushed the 1.0.4 tag to the github repo as well

davetcoleman:
> So, I'm asking for feedback, which way to proceed. @v4hn, @davetcoleman, @gavanderhoorn

The discussed route makes sense, so long as all of @tylerjw's work backporting doesn't get totally thrown away.

Slightly off topic: this begs the question: should we start making branches for every release candidate, beyond just kinetic|melodic|noetic? e.g. branch ``1.1.x`` or ``melodic-1.1.x``?

rhaschke:
> The discussed route makes sense, as long as all of @tylerjw's work backporting doesn't get totally thrown away.

Why should we throw that great work away? If you look at the git history tree, you will see that I just merged the "release branch" 1.0.4 back into melodic-devel. There is no explicit need for a release _branch_. Usually we don't have fixups for a release (and if so, only a very few of them, which should all become part of the corresponding devel branch anyway).

image

davetcoleman:
I was agreeing with @v4hn 's concern:

> Reverting multiple commits in melodic-devel on top of 1.0.3 is a bad idea, making the history unreadable. Especially because the main things are huge merge commits with multiple individual commits for feature backports.

rhaschke:
Thanks for clarification.

PR #2118, opened 2020-05-28
[CMake] fix link to eigenpy

nim65s:
# Description

fix https://github.com/stack-of-tasks/eigenpy/issues/195

The issue there is that `eigenpy_INCLUDE_DIRS` doesn't provide numpy include dirs. Moreover, those `eigenpy_INCLUDE_DIRS` & `eigenpy_LIBRARIES` are backward compatibility artifacts from the pkg-config world and are now considered obsolete in modern CMake.

But thanks to https://github.com/ros-planning/moveit/pull/1737, eigenpy is found using `find_package`, so we can get all required information to find all the needed headers and link to all  the needed libraries with a `target_link_libraries` to the `eigenpy::eigenpy` imported CMake target.

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/CMakeLists.txt
+++ b/moveit_ros/planning_interface/CMakeLists.txt
@@ -86,7 +86,6 @@ include_directories(${THIS_PACKAGE_INCLUDE_DIRS}
 
 include_directories(SYSTEM
                     ${EIGEN3_INCLUDE_DIRS}
-                    ${eigenpy_INCLUDE_DIRS}
                     ${PYTHON_INCLUDE_DIRS})
 
 add_subdirectory(py_bindings_tools)
--- a/moveit_ros/planning_interface/move_group_interface/CMakeLists.txt
+++ b/moveit_ros/planning_interface/move_group_interface/CMakeLists.txt
@@ -6,7 +6,7 @@ target_link_libraries(${MOVEIT_LIB_NAME} moveit_common_planning_interface_object
 add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
 
 add_library(${MOVEIT_LIB_NAME}_python src/wrap_python_move_group.cpp)
-target_link_libraries(${MOVEIT_LIB_NAME}_python ${MOVEIT_LIB_NAME} ${eigenpy_LIBRARIES} ${PYTHON_LIBRARIES} ${catkin_LIBRARIES} ${Boost_LIBRARIES} moveit_py_bindings_tools)
+target_link_libraries(${MOVEIT_LIB_NAME}_python ${MOVEIT_LIB_NAME} eigenpy::eigenpy ${PYTHON_LIBRARIES} ${catkin_LIBRARIES} ${Boost_LIBRARIES} moveit_py_bindings_tools)
 add_dependencies(${MOVEIT_LIB_NAME}_python ${catkin_EXPORTED_TARGETS})
 set_target_properties(${MOVEIT_LIB_NAME}_python PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 set_target_properties(${MOVEIT_LIB_NAME}_python PROPERTIES OUTPUT_NAME _moveit_move_group_interface PREFIX "")


acxz:
thx for this PR, Let me give this a shot real quick

nim65s:
Also, this line is useless:

nim65s:
The `catkin_lint` CI check failure is unrelated to this PR:

```
moveit_setup_assistant: CMakeLists.txt(51): error: package 'rviz' must be in CATKIN_DEPENDS in catkin_package()
```

rhaschke:
Could you please update your main description and indicate the Ubuntu, ROS, and MoveIt version you are experiencing issues with? I'm also include @wxmerkt, because he is the maintainer for the ROS eigenpy package providing the mentioned variables. If these variables are inconsistent to what `eigenpy::eigenpy` yields, this should be fixed there.

acxz:
Hey @rhaschke 
I am one of the maintainers for ROS-melodic and ROS-Noetic on Arch Linux
I have been trying to compile ‘ROS-melodic-moveit-ROS-planning-interface’ (Moveit version 1.0.3) but that package is not able to find the correct numpy headers. I posted an issue about this over at eigenpy and it was found that Moveit-ROS-planning-interface is not linking/including eigenpy correctly. Hence this PR. Right now the Arch Linux package of Moveit-ROS-planning-interface has been patched with this PR Until this a new release is made with these commits.

rhaschke:
@acxz, thanks for this clarification. 
I guess our misunderstanding stems from the title of #1737, which states: "treat eigenpy as a system dependency". While it was meant, that we don't handle it via find_package(catkin ...) anymore (but directly), you seem to have interpreted that we don't need the `ros-*-eigenpy` package anymore?
But, at least in Ubuntu 19.04, eigenpy is only shipped via `ros-melodic-eigenpy` from https://github.com/ipab-slmc/eigenpy_catkin-release (there is no _native_ system package for eigenpy):
```bash
$ apt search eigenpy
Sorting... Done
Full Text Search... Done
ros-melodic-eigenpy/bionic,now 2.3.1-1bionic.20200410.133723 amd64 [installed]
  Bindings between Numpy and Eigen using Boost.Python

ros-melodic-eigenpy-dbgsym/bionic 2.3.1-1bionic.20200410.133723 amd64
  debug symbols for ros-melodic-eigenpy
```
By the way, #1737 was required due to https://github.com/fkie/catkin_lint/issues/73.

From the changes of this PR, I understand that `eigenpy::eigenpy` yields different include and library dependencies than `eigenpy_INCLUDE_DIRS` and `eigenpy_LIBRARIES`. If that's the core issue, I argue that this should be (also) fixed by @wxmerkt in https://github.com/ipab-slmc/eigenpy_catkin-release, because the old cmake convention of using `*_INCLUDE_DIRS` and `*_LIBRARIES` variables (which is the default across ROS1) should still work.

acxz:
Ah thank you for clearing that up. Yes I agree with that reasoning. I did try to build moveit-ROS-planning-interface with ROS-melodic-eigenpy and I still received the same error.

nim65s:
@rhaschke : on Ubuntu LTS and on Debian, eigenpy is available in the [robotpkg apt binary repository](http://robotpkg.openrobots.org/debian.html). So you can consider that this is not only shipped via `ros-melodic-eigenpy`, but also from another third party native system package.

About the issue with `*_INCLUDE_DIRS`, the issue is that you added `eigenpy_INCLUDE_DIRS` without `NUMPY_INCLUDE_DIRS`.

Here is the curent `eigenpy.pc` on my system (with `robotpkg-py27-eigenpy`)
```
prefix=/opt/openrobots
exec_prefix=/opt/openrobots
libdir=/opt/openrobots/lib
bindir=/opt/openrobots/bin
pkglibdir=/opt/openrobots/lib/eigenpy
includedir=/opt/openrobots/include
datarootdir=/opt/openrobots/share
pkgdatarootdir=/opt/openrobots/share
docdir=
doxygendocdir=

Name: eigenpy
Description: Bindings between Numpy and Eigen using Boost.Python
URL: http://github.com/stack-of-tasks/eigenpy
Version: 2.4.0
Requires: eigen3 >= 3.0.5
Conflicts: 
Libs: -Wl,-rpath,${libdir} -L${libdir}  -leigenpy -Wl,-rpath,/usr/lib/x86_64-linux-gnu -L/usr/lib/x86_64-linux-gnu -lboost_python-py27
Libs.private: 
Cflags:  -I${includedir} -I/usr/include -I/usr/include/python2.7 -I/usr/lib/python2.7/dist-packages/numpy/core/include
```

So the path to numpy header is there, in a way sufficient to use `g++ $(pkg-config --cflags eigenpy)` to quickly build something that depends on eigenpy.

If you want to properly package something that depend on eigenpy through its .pc file, I think you need to find numpy and add its include dirs in your CMakeLists.txt. Maybe, we should add `numpy` in the `Requires` section of eigenpy.pc, to ensure that all packages dependent on eigenpy will automatically found `numpy`. But I think you'd still have to add `NUMPY_INCLUDE_DIRS`.

nim65s:
Or maybe do you expect numpy include dirs and python include dirs to be included in `eigenpy_INCLUDE_DIRS` ?

wxmerkt:
Hi all, thanks for looking into this. I didn't get a chance yesterday evening to respond to ongoing exchanges on the EigenPy issue. 

Note, this PR is equivalent to https://github.com/ros-planning/moveit/pull/2010 - however back then the updated EigenPy with CMake targets was not yet released to Kinetic. It now is. Thanks for re-raising this issue; this is an important update for future use and improves the overall situation (and the existing code still works and is backwards compatible). As thus, I think we should merge this PR.

> From the changes of this PR, I understand that eigenpy::eigenpy yields different include and library dependencies than eigenpy_INCLUDE_DIRS and eigenpy_LIBRARIES. If that's the core issue, I argue that this should be (also) fixed by @wxmerkt in https://github.com/ipab-slmc/eigenpy_catkin-release, because the old cmake convention of using *_INCLUDE_DIRS and *_LIBRARIES variables (which is the default across ROS1) should still work.

MoveIt correctly builds and links with the currently released EigenPy on Kinetic and Melodic - unless there recently are buildfarm error emails. The EigenPy packages released via the ROS buildfarm are available for Kinetic, Melodic, and Noetic and are released from the upstream repository as a third-party package release.

The described error appears on Arch Linux - I have not investigated the ROS packaging for there or why it happens (it's not via the ROS buildfarm). @acxz did this build in the past and start to fail recently on Melodic? Or has it not been released to Melodic before?

gavanderhoorn:
As an outside observer, this still appears as an issue with `eigenpy` as distributed and maintained by @wxmerkt's organisation.

Especially the issue with numpy not being added to the exported include dirs (reported by @nim65s).

From everything described until now, it does not seem like it's something which should be fixed here in MoveIt. "All" we do here is `find_package(eigenpy REQUIRED)`.

rhaschke:
I agree with @wxmerkt, that we _could_ merge this PR as is. This will allow to use the eigenpy package from [robotpkg](http://robotpkg.openrobots.org/debian.html) as well - without the need for the patched version from @wxmerkt, which provides`eigenpy_INCLUDE_DIRS`. However, I prefer to stick with the ROS1 convention of using old-style cmake.

However, I expect `eigenpy_INCLUDE_DIRS` to include _all_ required include dirs, also those from `numpy` and other transitive upstream dependencies. That's the convention of (old-style) cmake.
If that's not the case (yet), I consider that an issue with @wxmerkt's eigenpy release, which should be fixed there (too). So, in this sense I agree with @gavanderhoorn.

wxmerkt:
Hi @rhaschke and @gavanderhoorn,
Thanks for the comments. I agree we should maintain backwards compatibility. Please note, I _directly released_ from upstream with the [only patches](https://github.com/ipab-slmc/eigenpy_catkin-release/tree/patches/release/melodic/eigenpy) being to install the cmake-config (which gets generated by https://github.com/jrl-umi3218/jrl-cmakemodules) to the correct location. 

The PkgConfig and CmakeConfig get generated by https://github.com/jrl-umi3218/jrl-cmakemodules - I am not following all changes there (which frequently get synced into EigenPy). The relevant macros are https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/package-config.cmake#L108 and the config template is https://github.com/jrl-umi3218/jrl-cmakemodules/blob/master/Config.cmake.in

I am also surprised to see NumPy and transitive dependencies aren't included (there was a PR recently to remove NumPy include dirs, https://github.com/jrl-umi3218/jrl-cmakemodules/pull/383, yet I don't know its impact and EigenPy including this change has not been released to Melodic).

Can we please clarify where this currently fails? The CI and buildfarm are both fine with the current state? This does appear on Arch Linux - and I have not reproduced or tested any release or packaging there.

I want to make sure this gets addressed proper and is not due to the way I've released it directly from upstream. @rhaschke, @gavanderhoorn Do you suggest to add a patch to the release repository to add the NumPy when there is a good chance this is due to https://github.com/jrl-umi3218/jrl-cmakemodules? Or something else? Please advise - happy to provide time and a fix as you suggest.

@rhaschke, the recommended and primarily supported way going forward upstream (https://github.com/jrl-umi3218/jrl-cmakemodules) is the CMake targets `eigenpy::eigenpy`. I am not sure how much the other modes will be supported or tested. @nim65s might be able to advise on that, yet I recall a PR comment recently that they'd recommend moving to targets.

nim65s:
Thanks for your answers. 

> However, I prefer to stick with the ROS1 convention of using old-style cmake.

@rhaschke : I understand the need for backward compatibility, as I am in charge of some robots on 16.04 / kinetic. But `eigenpy::eigenpy` works well on those platforms, so I don't understand why you don't want to move on.

Do you have requirements for older platforms ? Eigenpy and the jrl-cmakemodule are currently not supporting anything older than 16.04. 

This support can be added, but we would need to know precisely what do we need to support and until when. Otherwise, our CI would still have to check 10.04…

> However, I expect eigenpy_INCLUDE_DIRS to include all required include dirs, also those from numpy and other transitive upstream dependencies. That's the convention of (old-style) cmake.

@rhaschke @wxmerkt : do you have any reference for this ? It seems natural for both of you, but it is not for me. I spent painful years to maintain old-style cmake stuff having to include all dependency everywhere, even transitive ones. And I think that's the reason why catkin and the jrl-cmakemodule were even created in the first place, to ease the management of those transitive dependencies.

wxmerkt:
I found an issue with the `CMAKE_INSTALL_LIBDIR` for EigenPy on **Noetic** and added a patch (https://github.com/ipab-slmc/eigenpy_catkin-release/blob/patches/release/noetic/eigenpy/0005-Workaround-for-architecture-issue-with-GNUInstallDir.patch, release: https://github.com/ros/rosdistro/pull/25190). This should fix the original issue by @acxz reported on Noetic on Arch as well. Apologies for the teething issues while sorting out and testing the early releases on Noetic.

Note, this is not an issue with either Kinetic (which has the same patch) or Melodic (which did not need a patch and from which I ported the patches to Noetic).

rhaschke:
@nim65s, I guess you are maintainer of eigenpy.

> But `eigenpy::eigenpy` works well on those platforms, so I don't understand why you don't want to move on.

I'm fine to merge this here, but I would like to push you (as well) to either:
- remove the old-style cmake support completely from eigenpy
- or to fix it.

Currently, it's obviously in a broken state.

nim65s:
@rhaschke : I totally agree with your points. Actually, I am working towards removing the old-style CMake support, but this should be done carefully to avoid breaking anything. This PR is part of it :)

v4hn:
> eigenpy maintainers recommend the modern, target-based dependency style.

This is the crucial point. As the change is backward compatible (by now), let's merge it.

rhaschke:
Thanks Michael.

v4hn:
I'm still a bit frustrated about not having an answer to this question:

> do you have any reference for this? It seems natural for both of you, but it is not for me.

As a Linux distribution maintainer it also "seems natural" to me to include all transitive dependencies. The "old-style" use was fashionable for a long time and replaced `pkg-config` usage (which can resolve dependencies transitively).
No one would want to keep up with manual dependency chains, so the only natural alternative is to include them in these variables...
However, I do have a hard time finding references (or even examples) for this without too much effort. 
It's even harder to find anything as more and more projects migrate to the modern interface approach.

Let's just hope changing this does not break anything in terms of include reordering in the longer run, but as long as CI passes, it should be fine.

nim65s:
I'm also a bit frustrated here and would really like to understand that better.

pkg-config resolves dependencies transitively. If we put `Requires: numpy` in `eigenpy.pc`, moveit would look for `numpy`, even if we don't ask this explicitely in moveit.

But I never saw any `.pc` file with more than one path in `includedir=…`. And I'm not even sure that it would work. So putting a `Requires: numpy` in `eigenpy.pc` would allow to skip the ` pkg_check_modules(numpy REQUIRED)` part, but would not allow skipping the addition of `${numpy_INCLUDE_DIRS}` in moveit.

Of course, if we emulate the `.pc` by manually defining `${eigenpy_INCLUDE_DIRS}`, we can put anything we like into it. But then, what's the point in making CMake generate and install a .pc file in one package, and then making CMake find and parse this .pc file in a dependent package ? Let's agree on generating a .cmake file instead :)

And if no one would want to keep up with manual dependency chains, who really wants to maintain the CMake code that concatenates those variables ? @v4hn this is one more thing that feel natural for you, but not for me :p

In the end, the only natural way for me is the "modern" one.

rhaschke:
@nim65s, as far as I understand https://github.com/ros-planning/moveit/pull/2118#issuecomment-635882287, the stuff exported from the eigenpy cmake files is defined by _your_ repo.

Neither MoveIt nor catkin cmake read from `eigenpy.pc` (if you don't do).
I didn't look into your cmake code, but in order to define the target `eigenpy::eigenpy`, you actually need to list all direct dependencies. Transitive ones are pulled in by cmake.
One option, I see, is to define `INCLUDE_DIRS` and `LIBRARIES` via `get_target_property` from `eigenpy::eigenpy`.

wxmerkt:
@rhaschke This should be handled by https://github.com/jrl-umi3218/jrl-cmakemodules/blob/e715bf761e97dbcd704adecd03c28d8e195e7811/Config.cmake.in#L15-L78 where we search for the full paths for each of the defined pkg-config requirements. These lines were added to address https://github.com/jrl-umi3218/jrl-cmakemodules/issues/346 in order to establish backwards compatibility (particularly for MoveIt as the primary consumer of the traditional CMake way). There is no more catkin-cmake for EigenPy, it's all handled directly by upstream.

rhaschke:
I guess, the mentioned code is also used to derive required include dirs and libs for the eigenpy target? If So, my proposal should work, right?

wxmerkt:
The setting of these variables in the mentioned lines is to allow for variables in the projectConfig.cmake to be set when they are specified with pkg_config_add_dependency downstream - this is for the transition period as there are multiple ways to add dependencies with that cmake-modules.

A few lines below that block is the modern CMake part that then includes the exported targets.

Yes, your proposal does work and will likely be more robust as it builds on the exported targets. I've just verified it using the following `CMakeLists.txt`:

```cmake
cmake_minimum_required(VERSION 3.5)

find_package(eigenpy REQUIRED)

message(STATUS "eigenpy_INCLUDE_DIRS: ${eigenpy_INCLUDE_DIRS}")
message(STATUS "eigenpy_LIBRARIES: ${eigenpy_LIBRARIES}")

# Extract the properties
get_target_property(eigenpy_target_INCLUDE_DIRS eigenpy::eigenpy INTERFACE_INCLUDE_DIRECTORIES)
get_target_property(eigenpy_target_LIBRARIES eigenpy::eigenpy LINK_LIBRARIES)

message(STATUS "eigenpy_target_INCLUDE_DIRS: ${eigenpy_target_INCLUDE_DIRS}")
message(STATUS "eigenpy_target_LIBRARIES: ${eigenpy_target_LIBRARIES}")


###############################################################################
# Code from https://stackoverflow.com/a/51987470 to print all target properties

# Get all propreties that cmake supports
execute_process(COMMAND cmake --help-property-list OUTPUT_VARIABLE CMAKE_PROPERTY_LIST)

# Convert command output into a CMake list
STRING(REGEX REPLACE ";" "\\\\;" CMAKE_PROPERTY_LIST "${CMAKE_PROPERTY_LIST}")
STRING(REGEX REPLACE "\n" ";" CMAKE_PROPERTY_LIST "${CMAKE_PROPERTY_LIST}")
# Fix https://stackoverflow.com/questions/32197663/how-can-i-remove-the-the-location-property-may-not-be-read-from-target-error-i
list(FILTER CMAKE_PROPERTY_LIST EXCLUDE REGEX "^LOCATION$|^LOCATION_|_LOCATION$")
# For some reason, "TYPE" shows up twice - others might too?
list(REMOVE_DUPLICATES CMAKE_PROPERTY_LIST)

# build whitelist by filtering down from CMAKE_PROPERTY_LIST in case cmake is
# a different version, and one of our hardcoded whitelisted properties
# doesn't exist!
unset(CMAKE_WHITELISTED_PROPERTY_LIST)
foreach(prop ${CMAKE_PROPERTY_LIST})
    if(prop MATCHES "^(INTERFACE|[_a-z]|IMPORTED_LIBNAME_|MAP_IMPORTED_CONFIG_)|^(COMPATIBLE_INTERFACE_(BOOL|NUMBER_MAX|NUMBER_MIN|STRING)|EXPORT_NAME|IMPORTED(_GLOBAL|_CONFIGURATIONS|_LIBNAME)?|NAME|TYPE|NO_SYSTEM_FROM_IMPORTED)$")
        list(APPEND CMAKE_WHITELISTED_PROPERTY_LIST ${prop})
    endif()
endforeach(prop)

function(print_properties)
    message ("CMAKE_PROPERTY_LIST = ${CMAKE_PROPERTY_LIST}")
endfunction(print_properties)

function(print_whitelisted_properties)
    message ("CMAKE_WHITELISTED_PROPERTY_LIST = ${CMAKE_WHITELISTED_PROPERTY_LIST}")
endfunction(print_whitelisted_properties)

function(print_target_properties tgt)
    if(NOT TARGET ${tgt})
      message("There is no target named '${tgt}'")
      return()
    endif()

    get_target_property(target_type ${tgt} TYPE)
    if(target_type STREQUAL "INTERFACE_LIBRARY")
        set(PROP_LIST ${CMAKE_WHITELISTED_PROPERTY_LIST})
    else()
        set(PROP_LIST ${CMAKE_PROPERTY_LIST})
    endif()

    foreach (prop ${PROP_LIST})
        string(REPLACE "<CONFIG>" "${CMAKE_BUILD_TYPE}" prop ${prop})
        # message ("Checking ${prop}")
        get_property(propval TARGET ${tgt} PROPERTY ${prop} SET)
        if (propval)
            get_target_property(propval ${tgt} ${prop})
            message ("${tgt} ${prop} = ${propval}")
        endif()
    endforeach(prop)
endfunction(print_target_properties)

###############################################################################

# Print all set properties
print_target_properties(eigenpy::eigenpy)
```

where the output is as follows (note that the `get_target_property` resolves all dependencies while the exported INCLUDE_DIRS do not contain numpy/boost-python):

```
**-- eigenpy_INCLUDE_DIRS: /opt/ros/melodic/include**
-- eigenpy_LIBRARIES: /opt/ros/melodic/lib/libeigenpy.so;/usr/lib/x86_64-linux-gnu/libboost_python.so
-- eigenpy_target_INCLUDE_DIRS: /usr/include;/usr/include/eigen3;/usr/include/python2.7;/usr/lib/python2.7/dist-packages/numpy/core/include;/opt/ros/melodic/include;/usr/include/eigen3;/opt/ros/melodic/include
-- eigenpy_target_LIBRARIES: eigenpy_target_LIBRARIES-NOTFOUND
eigenpy::eigenpy AUTOMOC_COMPILER_PREDEFINES = ON
eigenpy::eigenpy AUTOMOC_MACRO_NAMES = Q_OBJECT;Q_GADGET;Q_NAMESPACE
eigenpy::eigenpy BINARY_DIR = /home/wxm/tmp/test_eigenpy/build
eigenpy::eigenpy BUILD_WITH_INSTALL_RPATH = OFF
eigenpy::eigenpy IMPORTED = TRUE
eigenpy::eigenpy IMPORTED_CONFIGURATIONS = NONE
eigenpy::eigenpy INSTALL_RPATH = 
eigenpy::eigenpy INSTALL_RPATH_USE_LINK_PATH = OFF
**eigenpy::eigenpy INTERFACE_INCLUDE_DIRECTORIES = /usr/include;/usr/include/eigen3;/usr/include/python2.7;/usr/lib/python2.7/dist-packages/numpy/core/include;/opt/ros/melodic/include;/usr/include/eigen3;/opt/ros/melodic/include**
eigenpy::eigenpy INTERFACE_SYSTEM_INCLUDE_DIRECTORIES = /usr/include;/usr/include/eigen3;/usr/include/python2.7;/usr/lib/python2.7/dist-packages/numpy/core/include;include;/usr/include/eigen3
eigenpy::eigenpy NAME = eigenpy::eigenpy
eigenpy::eigenpy POSITION_INDEPENDENT_CODE = True
eigenpy::eigenpy SKIP_BUILD_RPATH = OFF
eigenpy::eigenpy SOURCE_DIR = /home/wxm/tmp/test_eigenpy
eigenpy::eigenpy TYPE = SHARED_LIBRARY
```

nim65s:
TL;DR: This is brilliant. But I find it really sad that we would have to do this. If anyone motivates a use case where the current situation would be an issue, I will push for this solution.


> @nim65s, as far as I understand #2118 (comment), the stuff exported from the eigenpy cmake files is defined by your repo.

@rhaschke  In the current situation, yes, that's totally right. The jrl-cmakemodule is mostly a sum of layers after layers of additions of compatibility stuff.

It does work pretty well. And wherever it doesn't, we add more compatibility code to make it work how we want it to. This fix a huge ton of issues that we don't even know we would have.

Find and link to boost::python privately or publicly on Debian 7 & Fedora 32 for Python 2.6 & 3.4 ? Check.
Generate IDE configuration for the qtcreator on 14.04 & xcode  on catalina ? Check.
Check and release a a version, sign a git tag and a the tarball with your gpg key ? Check.
Generate documentation and test configuration for C++ & Python ? Check.
Workaround issues from CMake 3.1 while not messing with features from CMake 3.17 ? Check.

I have no idea of all the features it actually has. This thing is incredible.

But… The whole existence of this module (and of catkin), is IMO a design issue of C++ that doesn't come with the right tooling modern languages have (not to mention that we also need Python). CMake came to fix this, but was not mature enough for a long time, so we had to hack stuff around to make it work in most of our use cases.

I totally agree with @v4hn in  "as long as CI passes, it should be fine". If it ain't broke, don't fix it.

Now, each time we add another compatiblity layer for yet another reason… How do we ensure that we're not breaking another one ? We don't. We can't. How could we ?

So I'm not joking when I say that if you need anything to work in any different way, we can manage to implement that. And your proposal seems really good for that, I wasn't expecting that much at all, thanks ! And @wxmerkt is taking this really seriously, providing  well documented references, facts and experiments, as always, that's motivating !

But what do we actually need, and what do we really want to maintain ? The most standard stuff, as long as it works well enough. I think that the less CMake code I have, the better. This is not a code golf, but the combinatorial on CMake branches is untestable and a nightmare for any maintainer, don't you think ? How could we check all the CMake variables we export ?

And to whom this solution would benefit ? Someone compiling an old moveit but with a recent eigenpy, while having CMake >= 3.1 as required by recent eigenpy, but not recent enough to be able to use `eigenpy::eigenpy` in moveit ? This is unlikely. Even in the robotics community.


Your point is that if we provide this feature, we should ensure that it work. Actually, to the best of my knowledge, it does, in all places officially supported by ROS. And certainly a few more.
But I hope that we could stop providing this feature without hurting anyone, instead of spending energy trying to bullet-proof it.

rhaschke:
@nim65s, I can understand your reasoning. If you only recommended the new target-based way, I just should - in the long run - remove the old style, (at least partially) broken `INCLUDE_DIR` support.

PR #2119, opened 2020-05-28
MoveItCpp: Use csm to get robot's current state

JafarAbdi:
### Description

I'm sometime getting `ros.moveit_ros_planning.trajectory_execution_manager: 
Invalid Trajectory: start point deviates from current robot state more than 0.01` and after debugging it I found that it is caused by getting not up-to-date robot state when I call `MoveItCpp::getCurrentState`

In the current implementation if wait_seconds is set to zero (which is the default value) it returns the PlanningScene's robot state without checking its time, this PR

1- Use current state monitor directly to get the current state
2- Set default wait time to 1.0 to reflect the default value for the current state monitor's waitForCurrentState function

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
+++ b/moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
@@ -138,7 +138,7 @@ public:
 
   /** \brief Get the current state queried from the current state monitor
       \param wait_seconds the time in seconds for the state monitor to wait for a robot state. */
-  moveit::core::RobotStatePtr getCurrentState(double wait_seconds = 0.0);
+  moveit::core::RobotStatePtr getCurrentState(double wait_seconds = 1.0);
 
   /** \brief Get all loaded planning pipeline instances mapped to their reference names */
   const std::map<std::string, planning_pipeline::PlanningPipelinePtr>& getPlanningPipelines() const;
--- a/moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
+++ b/moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
@@ -195,16 +195,12 @@ const ros::NodeHandle& MoveItCpp::getNodeHandle() const
 
 bool MoveItCpp::getCurrentState(moveit::core::RobotStatePtr& current_state, double wait_seconds)
 {
-  if (wait_seconds > 0.0 &&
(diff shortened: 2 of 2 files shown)


henningkayser:
The error you see is possibly coming from calling this function while the robot is still moving or stopping. The PlanningScene state shouldn't be too old (did you check that), actually it should be up-to-date only limited by CSM's update rate. Also, in cases where it's actually required to use the current (future) state, it's easy to just set a custom wait time.

moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -196,16 +196,12 @@ const ros::NodeHandle& MoveItCpp::getNodeHandle() const
 
 bool MoveItCpp::getCurrentState(moveit::core::RobotStatePtr& current_state, double wait_seconds)
 {
-  if (wait_seconds > 0.0 &&
-      !planning_scene_monitor_->getStateMonitor()->waitForCurrentState(ros::Time::now(), wait_seconds))
+  auto csm = planning_scene_monitor_->getStateMonitor();
+  if (!csm->waitForCurrentState(ros::Time::now(), wait_seconds) || !(current_state = csm->getCurrentState()))
```
mlautman, line 200: In general, I find it useful to separate the different potential sources of error rather than stacking them like this. I find that providing error messages that are specific to the actual failure makes for faster debugging which I find to be worth the cost of a few extra lines of code.

mlautman: Looking into the code, it seems that this doesn't actually change the default behavior so feel free to disregard my previous comment.

moveit_ros/planning_interface/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -134,11 +134,11 @@ class MoveItCpp
 
   /** \brief Get the current state queried from the current state monitor
       \param wait_seconds the time in seconds for the state monitor to wait for a robot state. */
-  bool getCurrentState(moveit::core::RobotStatePtr& current_state, double wait_seconds);
+  bool getCurrentState(moveit::core::RobotStatePtr& current_state, double wait_seconds = 1.0);
 
   /** \brief Get the current state queried from the current state monitor
       \param wait_seconds the time in seconds for the state monitor to wait for a robot state. */
-  moveit::core::RobotStatePtr getCurrentState(double wait_seconds = 0.0);
+  moveit::core::RobotStatePtr getCurrentState(double wait_seconds = 1.0);
```
mlautman, line 141: Do you have a reason for changing the default behavior?

PR #2120, opened 2020-05-29
Fix catkin_lint

JafarAbdi:
### Description

As you can see, right now the master branch is failing due to catkin_lint -- example: https://travis-ci.org/github/ros-planning/moveit/jobs/692499448

### Checklist

### DIFF
--- a/moveit_ros/visualization/CMakeLists.txt
+++ b/moveit_ros/visualization/CMakeLists.txt
@@ -81,6 +81,7 @@ catkin_package(
     moveit_ros_robot_interaction
     object_recognition_msgs
     roscpp
+    rviz
   DEPENDS
     EIGEN3
 )
--- a/moveit_setup_assistant/CMakeLists.txt
+++ b/moveit_setup_assistant/CMakeLists.txt
@@ -57,7 +57,6 @@ catkin_package(
     moveit_ros_planning
     moveit_ros_visualization
     roscpp
-    rviz
 )
 
 # Header files that need Qt Moc pre-processing for use with Qt signals, etc:
--- a/moveit_setup_assistant/CMakeLists.txt
+++ b/moveit_setup_assistant/CMakeLists.txt
@@ -57,6 +57,7 @@ catkin_package(
     moveit_ros_planning
     moveit_ros_visualization
     roscpp
+    rviz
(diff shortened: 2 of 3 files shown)


v4hn:
@rhaschke argues for not having to add this dependency in cmake: https://github.com/ros-planning/moveit/pull/2116#issuecomment-635560400 .

moveit_ros/visualization/CMakeLists.txt
```diff
@@ -81,6 +81,7 @@ catkin_package(
     moveit_ros_robot_interaction
     object_recognition_msgs
     roscpp
+    rviz
```
rhaschke, line 84: This one is correct, because we indeed include rviz headers in our headers.

moveit_setup_assistant/CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ catkin_package(
     moveit_ros_planning
     moveit_ros_visualization
     roscpp
+    rviz
```
rhaschke, line 60: This one is not necessary. See discussion in https://github.com/ros-planning/moveit/pull/2116#issuecomment-635448407.
```suggestion
    rviz
```
However, given that MSA doesn't have downstream dependencies, it's better to add it here than to suppress the warning and risk ignoring future other stuff as well?

moveit_setup_assistant/package.xml
```diff
@@ -22,13 +22,13 @@
...
   <build_depend>libqt5-opengl-dev</build_depend>
   <build_depend>ompl</build_depend>
+  <build_depend>rviz</build_depend>
 
   <depend>moveit_core</depend>
   <depend>moveit_ros_planning</depend>
   <depend>moveit_ros_visualization</depend>
   <depend>rosconsole</depend>
   <depend>roscpp</depend>
-  <depend>rviz</depend>
```
v4hn, line 31: `<depend>` [adds build_depend,exec_depend, and build_export_depend](https://wiki.ros.org/catkin/package.xml).

build_export_depend is the offender here, the other two should still be required. So please add `<exec_depend>` too.

There was a reason why package.xml version 1 did not allow this kind of "generic depend"...

PR #2123, opened 2020-05-31
Bump version to upcoming Noetic release: 1.1.0

### LINKED ISSUE #2036: Version number of master
I'm somewhat confused about the version numbers in master, shouldn't they be bigger than in melodic-devel? I would have expected something like 1.1.0 or similar? I was trying to write an `#ifdef` but that kind of makes it hard?

Tagging the usual suspects @v4hn @rhaschke

rhaschke:
... as agreed on in maintainer meeting.
Fixes #2036.

### DIFF
--- a/moveit/package.xml
+++ b/moveit/package.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package format="2">
   <name>moveit</name>
-  <version>1.0.1</version>
+  <version>1.1.0</version>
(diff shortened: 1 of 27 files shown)


rhaschke:
This is simply the outcome of the automatic version bump performed by `catkin_prepare_release`.
This tool considers only those packages that are not CATKIN_IGNORED.

v4hn:
Fair enough. We discussed to use the tool for that, so I guess I can live with the result.
It doesn't really matter much anyway...

Issue #2062, opened 2020-05-07
RViz panel: parameters list can move right without scrollbar


v4hn:
I guess the picture says it all:

![clipped_parameters](https://user-images.githubusercontent.com/680358/81281955-d3050780-905a-11ea-8501-b979e1be2841.png)

This is on ROS melodic in Ubuntu 18.04.

I believe there is a similar issue, possibly the same, with RViz parameter edits.
Edit: the latter one might be related to my very custom system and I can't force it in Ubuntu right now...

felixvd:
Fixed in #2310

PR #2126, opened 2020-06-02
Fixed Issue: RViz panel-Planner Parameters move right without scrollbars

alexchua0108:
As mentioned in #2062 , RViz panel list will auto-scroll to the right. 

- Disabled the properties of "autoScroll" of planner_param_treeview

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>692</width>
-    <height>414</height>
+    <height>424</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -96,6 +96,9 @@
             </item>
             <item>
              <widget class="moveit_rviz_plugin::MotionPlanningParamWidget" name="planner_param_treeview">
+              <property name="autoScroll">
+               <bool>false</bool>
+              </property>
               <property name="indentation">
                <number>0</number>
               </property>


v4hn:
Please provide a screenshot while having an edit active.
Why do you increase the *height*?
Can you end up with part of the edit widget outside the view when simply disabling auto scroll?

alexchua0108:
> Please provide a screenshot while having an edit active.
> Why do you increase the _height_?
> Can you end up with part of the edit widget outside the view when simply disabling auto scroll?

Please remove the line of increase the _height_
I was using QT designer to edit the property, this might end up increase the Widget Height accidentally.

RViz panel-Planner Parameters will move right without scrollbars is because of the properties of autoScrolls function.
If this property is set to true (the default), the QAbstractItemView automatically scrolls the contents of the view if the user drags within 16 pixels of the viewport edge. If the current item changes, then the view will scroll automatically to ensure that the current item is fully visible.

felixvd:
Have you tested this? Can you confirm that it works as you expect it to, and post a screenshot of the GUI?

felixvd (changes requested):
I tested it and it works, just the line about the height is noise and should be removed as suggested.

I actually have the same issue with the Rviz Displays panel, but I don't see these settings in the repository. @rhaschke Do you know where to look?

rhaschke:
> I actually have the same issue with the Rviz Displays panel.

@felixvd, I'm not sure which issue you are referring to. The (minimal) widget height is automatically calculated by qtdesigner.

felixvd:
Sorry for the vagueness. I meant the auto-scrolling described in #2062. The content of the Displays panel can scroll horizontally so a part on the left ends up hidden. Here are some screenshots (in the bottom one I scrolled back to the left with the arrow keys):

Screenshot 2020-09-23 01:54:40

Screenshot 2020-09-23 01:55:02

rhaschke:
Ah. I noticed this issue before as well. Could you please file an issue in the rviz repo? I don't know by heart where to look...

felixvd:
Certainly.

Also this is fixed via #2310 . Thanks for figuring out the solution @alexchua0108 !

PR #2127, opened 2020-06-02
Support multiple planning pipelines with MoveGroup via MoveItCpp

### LINKED ISSUE #2078: Use MoveItCpp as MoveGroup context
I would like to let MoveGroup use MoveItCpp internally in order to reduce maintenance overhead and support some nice features like:

* Support multiple planning pipelines with MoveGroup
* Unify launch/configuration setup of MoveGroup and MoveItCpp
* Add support for MoveItCpp in MSA

henningkayser:
This effort adds support for running multiple planning pipeline instances with MoveGroup. This allows you to run OMPL, STOMP, CHOMP, etc at the same time or even multiple instances of the same planner (under different configuration namespaces). ~The planning pipeline is specified by a prefixing the planner id - `<pipeline>|<planner_id>` where <pipeline> is the parameter namespace of the pipeline used for MoveItCpp.~ The planning pipeline is specified by passing a separate id with the [MotionPlanRequest](https://github.com/ros-planning/moveit_msgs/pull/95). These changes are backwards compatible (if pipeline is not defined, the default is used) and easy to setup For testing, see https://github.com/ros-planning/moveit_resources/pull/47 for an example. While the old behavior still works, I added deprecation warnings that planning pipelines should be loaded in separate namespaces under `~planning_pipelines/` and that a default planning pipeline should be specified via '~default_planning_pipeline' .

TODOs:

- [x] MoveItCpp -> moveit_ros_planning
- [x] Add MoveItCpp to MoveGroup context
- [x] Initialize MoveItCpp configuration from move_group.launch
- [x] Configure default planning pipeline setup
- [x] Allow specifying non-default planning pipeline via planner_id
- [x] Extend MoveGroupInterface
- [x] Provide example configuration in MSA (multiple planning pipelines)
- [x] Allow specifying planning pipeline with MotionPlanning panel

The MotionPlanning panel now allows selecting non-default planning pipelines loaded by MoveItCpp:

move_group_multi_planner

Fixes #2078


davetcoleman:
I'm excited about this improvement!

henningkayser:
@v4hn ping for review

gavanderhoorn:
> The planning pipeline is specified by a prefixing the planner id - `<pipeline>|<planner_id>` where is the parameter namespace of the pipeline used for MoveItCpp

What's the rationale for using `|` as separator here?

Historically, paths, parameters, namespaces, TF frames and similar concepts which can be "contained" in namespaces have used a forward slash (ie: `/`).

henningkayser:
> > The planning pipeline is specified by a prefixing the planner id - `<pipeline>|<planner_id>` where is the parameter namespace of the pipeline used for MoveItCpp
> 
> What's the rationale for using `|` as separator here?
> 
> Historically, paths, parameters, namespaces, TF frames and similar concepts which can be "contained" in namespaces have used a forward slash (ie: `/`).

Currently, it's technically not forbidden to use `/` inside the planning pipeline name as it can represent absolute ore relative parameter namespaces and I wanted to keep support for the `moveit_cpp.yaml` without the overhead of adding a unique id generator.

gavanderhoorn:
I would probably suggest to standardise naming here such that `/` is only used for separating namespaces and nothing else, as `/` has a long history and will be the most intuitive to most (all?) users.

> Also, the delimiter is not used anywhere to my knowledge and also indicates a specific separation of the parts.

it certainly separates things, but the pipe symbol has historically has quite a different use. It's actually more of a connecting symbol than a separating one.

henningkayser:
> I would probably suggest to standardise naming here such that `/` is only used for separating namespaces and nothing else, as `/` has a long history and will be the most intuitive to most (all?) users.
> 

That's the point, it's already used for separating namespaces (pipeline config). Adding the planner id (which can consist of group name + algorithm or anything arbitrary defined by the planner) as a simple subnamespace would be a break in the representation, therefore I wouldn't use `/`.

> > Also, the delimiter is not used anywhere to my knowledge and also indicates a specific separation of the parts.
> 
> it certainly separates things, but the pipe symbol has historically has quite a different use. It's actually more of a connecting symbol than a separating one.

Hm, the planner_id somehow represents a connection between algorithm and planner instance.

gavanderhoorn:
> That's the point, it's already used for separating namespaces (pipeline config). Adding the planner id (which can consist of group name + algorithm or anything arbitrary defined by the planner) as a simple subnamespace would be a break in the representation, therefore I wouldn't use `/`.

Do I understand you correctly that this would essentially only be a problem if the string which identifies the planner and algorithm contains a `/` and is directly used to index into parameters?

What about decoupling this?

I'm only a single voice here, and it's not a deal breaker for me, but using a `/` instead of `|` seems much more natural to me.

henningkayser:
> > That's the point, it's already used for separating namespaces (pipeline config). Adding the planner id (which can consist of group name + algorithm or anything arbitrary defined by the planner) as a simple subnamespace would be a break in the representation, therefore I wouldn't use `/`.
> 
> Do I understand you correctly that this would essentially only be a problem if the string which identifies the planner and algorithm contains a `/` and is directly used to index into parameters?
> 
> What about decoupling this?

I agree that a separate `planning_pipeline` field inside `MotionPlanRequest` and  `PlannerInterfaceDescription` would be much cleaner. That could co along with corresponding interface changes for the `MoveGroupInterface`, which could.

> 
> I'm only a single voice here, and it's not a deal breaker for me, but using a `/` instead of `|` seems much more natural to me.

True, it might seem more natural, but in the end it's just a delimiter with a specific convention.
To me this approach seems easier than working around handling actual namespaces in the pipeline specifier.

gavanderhoorn:
> I agree that a separate `planning_pipeline` field inside `MotionPlanRequest` and `PlannerInterfaceDescription` would be much cleaner. That could co along with corresponding interface changes for the `MoveGroupInterface`, which could.

Changing the message is not what I meant.

My suggestion was to not think of `planner/algo` as directly related to a ROS parameter namespace, but as an actual identifier. That identifier then could comply with regular ROS naming and users' intuition.

Internally, it would be decomposed and after some post-processing or mapping, could be used to refer to specific namespaces on the ROS parameter server. But it would not be a 1-to-1 mapping, which decouples the occurrence of `/` from being potentially difficult with parameter namespaces.

v4hn:
I agree with Gijs. Using any other symbol than `/` as delimiter is
counterintuitive.
You actually invoke a planner *within* a planner manager, so the
subfolder-analogy is sound as well.

Concerning namespaces, I'm still not sure why you would require more
namespacing with sub-namespaces.
Even if MoveItCpp supports this at the moment, I don't see why we can't change
that for overall better usage and structure.
`move_group/planners/xyz/configuration` should be very much fine.
If you want to invoke a MoveItCpp with alternative configurations, just upload
them to a different namespace too or provide an *optional* list to restrict the
available planners for a concrete MoveItCpp instance.

There is a lot of merit in having *all* planner configurations supported from
the MotionPlanning panel in rviz to try them out and thus they should all be
available in one sub-namespace, e.g., `planners`).

v4hn:
fun fact: I sent the previous message by mail, but github's mailserver rejected it stating:

```
    The user you are trying to contact is receiving mail at a rate that
    550-5.2.1 prevents additional messages from being delivered. For more
    550-5.2.1 information, please visit 550 5.2.1
    https://support.google.com/mail/?p=ReceivingRatePerm
```

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -194,5 +194,7 @@ namespace moveit
 namespace planning_interface
 {
 using MoveItCpp[[deprecated("use moveit_cpp")]] = moveit_cpp::MoveItCpp;
+using MoveItCppPtr[[deprecated("use moveit_cpp")]] = moveit_cpp::MoveItCppPtr;
+using MoveItCppConstPtr[[deprecated("use moveit_cpp")]] = moveit_cpp::MoveItCppConstPtr;
```
v4hn, line 198: I guess you should add `MOVEIT_CLASS_FORWARD` explicitly, otherwise you miss some other types as well.

henningkayser:
@gavanderhoorn 
> My suggestion was to not think of `planner/algo` as directly related to a ROS parameter namespace, but as an actual identifier. That identifier then could comply with regular ROS naming and users' intuition.
> 

Makes sense, we just don't have a unique 'planner' here. We would have to generate or specify this pipeline id which would also require another parameter which maps to an actual config namespace..

> Internally, it would be decomposed and after some post-processing or mapping, could be used to refer to specific namespaces on the ROS parameter server. But it would not be a 1-to-1 mapping, which decouples the occurrence of `/` from being potentially difficult with parameter namespaces.

The 1-to-1 mapping is exactly what should be provided imo, and ids need to be coupled to the parameter namespace. What I'm reading here is that we have to encode an actual namespace (pipeline config) into an id (pipeline name) so that we can represent an id (planner id) as a namespace (pipeline name + planner id). To me this just looks extra complicated and not straight forward at all. The more I think about this, the more I'm actually voting for adding an explicit pipeline field to the message. Especially, since pipeline is a property for MoveGroup/MoveItCpp and planner_id is a property for the planning plugin. Mixing both up in a single string doesn't smell right.

@v4hn 
> I agree with Gijs. Using any other symbol than `/` as delimiter is
> counterintuitive.
> You actually invoke a planner _within_ a planner manager, so the
> subfolder-analogy is sound as well.

Hm, not convinced, sorry. To me `/` represents a parent/child relationship between elements of the same type. `:`, `::`, `-`, `#` would just be as intuitive imo. I'm fine with switching to this representation though, but only if we can still support the full range of options for MoveGroup and MoveItCpp.
That means we have to add a pipeline name identifier to the MoveItCpp options that is then used instead of the namespace. This one could default to the namespace name if it is a direct child of a specified parent namespace (similar to your suggestion below).
 
> Concerning namespaces, I'm still not sure why you would require more
> namespacing with sub-namespaces.
> You would end up with planner strings like `ompl/configuration1|RRTConnect`. That's hard to parse and ambiguous without further insight.

Well, it's supported and I don't see a convincing argument to remove support for it, as it can be very useful for specifying instances running on or using different resources. How is it harder to parse than anything else and with what is this representation ambiguous? I totally understand that it doesn't look pretty - too much information for a simple algorithm identifier - but for now it is the most explicit and still generic representation that supports all planners available with MoveItCpp.

> Even if MoveItCpp supports this at the moment, I don't see why we can't change
> that for overall better usage and structure.
> `move_group/planners/xyz/configuration` should be very much fine.

I'm fine with a convention like this for MoveGroup and generate a valid MoveItCpp PlanningPipelineOptions with it (that would even make one the pipeline_names parameter redundant). It would be still nicer to just be able to use MoveItCpp parameters or yaml files directly.
There is probably no way around adding a `pipeline_name` parameter to the options struct.

> If you want to invoke a MoveItCpp with alternative configurations, just upload
> them to a different namespace too or provide an _optional_ list to restrict the
> available planners for a concrete MoveItCpp instance.

That doesn't sound like it's better usable at all.

> There is a lot of merit in having _all_ planner configurations supported from
> the MotionPlanning panel in rviz to try them out and thus they should all be
> available in one sub-namespace, e.g., `planners`).

Agree, that's what I'm trying to do here ;)

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
henningkayser: good point, thx

felixvd:
Before this gets stale: I feel like I remember there was some discussion about the architecture choices in this PR, but [these notes](https://discourse.ros.org/t/meeting-notes-moveit-manipulation-wg-meeting-may-28th/14344) don't seem to mention it. Could you summarize if you remember and maybe update on what is left to be done?

v4hn:
> Before this gets stale

This **is** stale, to my dismay. :( Henning, could you continue this work anytime soon to get this ready for noetic?

The results of the discussion in the meeting where to take the best of both worlds:

- There should be two separate fields for the Planner-ID and the PlannerManager. Everything else is somewhat messy for no reason.

- The interface should be restricted to have all planners in a fixed namespace on the parameter server, e.g. `move_group/planners`, to make sure they are in predictable places and to allow RViz/etc. to find all configurations automatically.

henningkayser:
@v4hn @felixvd I just updated this and added the pipeline_id field to the corresponding message types and added support for reading/setting the pipeline id via the MGI.

machinekoder:
@henningkayser Something in the noetic branch is causing this error: `TF_REPEATED_DATA ignoring data with redundant timestamp for frame link_2 at time 1603986944.105939 according to authority unknown_publisher` which indicates that there is problem duplicating tf messages somehwere: https://github.com/ros/geometry2/issues/467

moveit_ros/move_group/src/move_group.cpp
```diff
@@ -190,8 +193,70 @@ int main(int argc, char** argv)
...
+      moveit_cpp_options.planning_pipeline_options.pipeline_names = { default_planning_pipeline };
+      moveit_cpp_options.planning_pipeline_options.parent_namespace = pnh.getNamespace();
+    }
+
+    // Reset invalid pipeline parameter for MGI requests
+    pnh.setParam("default_planning_pipeline", default_planning_pipeline);
+  }
+
+  // Initialize MoveItCpp
+  const auto moveit_cpp = std::make_shared<moveit_cpp::MoveItCpp>(moveit_cpp_options, pnh, tf_buffer);
```
machinekoder, line 258: MoveItCpp has its own tf_listener. Passing `tf_buffer` here causes `TF_REPEATED_DATA` errors.

moveit_ros/move_group/src/move_group.cpp
```diff
@@ -190,8 +193,70 @@ int main(int argc, char** argv)
   std::shared_ptr<tf2_ros::Buffer> tf_buffer = std::make_shared<tf2_ros::Buffer>(ros::Duration(10.0));
   std::shared_ptr<tf2_ros::TransformListener> tfl = std::make_shared<tf2_ros::TransformListener>(*tf_buffer, nh);
```
machinekoder, line 194: First tf listener.

moveit_ros/move_group/src/move_group.cpp
machinekoder: https://github.com/ros-planning/moveit/blob/noetic-devel/moveit_ros/planning_interface/moveit_cpp/src/moveit_cpp.cpp#L65

moveit_ros/move_group/src/move_group.cpp
machinekoder: Remove this one, as we listen in `MoveItCpp`

moveit_ros/move_group/src/move_group.cpp
henningkayser: good catch, thanks!

tylerjw:
This is failing on the pilz planner tests.  @henningkayser any idea what would cause that?

henningkayser:
> This is failing on the pilz planner tests. @henningkayser any idea what would cause that?

I wanted to look into it this week. I think the pilz planners depend on certain changes in the launch structure that don't work well with the now deprecated parameter namespace.

machinekoder:
@tylerjw @henningkayser I'm using it with the pilz planner and it works, but I haven't looked at the test to be honest. So the problem might be in the tests.

henningkayser:
I rebased this PR, luckily this was very straight-forward without too many conflicts to resolve. Additional changes are:
* Fixes PlanningSceneMonitor topics in MoveItCpp (before this, the monitor/publish topics were actually flipped) cd3df112
* Support loading capabilities from planning pipeline namespace (simplifies launch structure for loading Pilz planner capabilities) ac4bb59

henningkayser:
> > Before this gets stale
> 
> This **is** stale, to my dismay. :( Henning, could you continue this work anytime soon to get this ready for noetic?

Let's get this merged quickly before it gets stale again ;)

> The results of the discussion in the meeting where to take the best of both worlds:
> 
> * There should be two separate fields for the Planner-ID and the PlannerManager. Everything else is somewhat messy for no reason.

Done! There is now an extra field `pipeline_id` in all relevant message types (https://github.com/ros-planning/moveit_msgs/pull/95). That's using the pipeline namespace and not directly the PlannerManager since that would not allow for different adapter setups like OMPL and OMPL+CHOMP/STOMP.

> * The interface should be restricted to have all planners in a fixed namespace on the parameter server, e.g. `move_group/planners`, to make sure they are in predictable places and to allow RViz/etc. to find all configurations automatically.

Also done, even if all interfaces are using the `MoveGroupQueryPlannersService` afaik.

moveit_ros/move_group/src/move_group.cpp
```diff
@@ -187,11 +203,71 @@ int main(int argc, char** argv)
   spinner.start();
   ros::NodeHandle nh;
```
rhaschke, line 204: This node handle isn't used anymore, is it?
```suggestion
```

moveit_ros/move_group/src/move_group.cpp
```diff
@@ -187,11 +203,71 @@ int main(int argc, char** argv)
...
+      moveit_cpp_options.planning_pipeline_options.pipeline_names = { default_planning_pipeline };
+      moveit_cpp_options.planning_pipeline_options.parent_namespace = pnh.getNamespace();
+    }
+
+    // Reset invalid pipeline parameter for MGI requests
+    pnh.setParam("default_planning_pipeline", default_planning_pipeline);
+  }
+
+  // Initialize MoveItCpp
+  const std::shared_ptr<tf2_ros::Buffer> tf_buffer = std::make_shared<tf2_ros::Buffer>(ros::Duration(10.0));
```
rhaschke, line 268: For symmetry ;-)
```suggestion
  const auto tf_buffer = std::make_shared<tf2_ros::Buffer>(ros::Duration(10.0));
```

moveit_ros/move_group/src/default_capabilities/query_planners_service_capability.cpp
```diff
@@ -58,23 +58,35 @@ void MoveGroupQueryPlannersService::initialize()
...
+    const planning_interface::PlannerManagerPtr& planner_interface = planning_pipeline->getPlannerManager();
+    if (planner_interface)
+    {
+      std::vector<std::string> algs;
+      planner_interface->getPlanningAlgorithms(algs);
+      moveit_msgs::PlannerInterfaceDescription pi_desc;
+      pi_desc.name = planner_interface->getDescription();
+      // Set pipeline id for non-default planners
+      if (planning_pipeline != context_->planning_pipeline_)
+        pi_desc.pipeline_id = pipeline_id;
```
rhaschke, line 74: Why not setting the pipeline_id for the default planner as well?

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_planning.cpp
```diff
@@ -377,7 +377,33 @@ void MotionPlanningFrame::updateQueryStateHelper(moveit::core::RobotState& state
...
+void MotionPlanningFrame::populatePlannersList(const std::vector<moveit_msgs::PlannerInterfaceDescription>& desc)
+{
+  ui_->planning_pipeline_combo_box->clear();
+
+  planner_descriptions_ = desc;
+  size_t default_planner_index = 0;
+  for (auto& d : planner_descriptions_)
+  {
+    // Use planner as default if no pipeline id is given
+    if (d.pipeline_id.empty())
```
rhaschke, line 389: I guess that's the only location, where you use the info that the default pipeline has no description.
Why not accessing the `default_planning_pipeline` parameter to identify it?

rhaschke (commented):
That's a lot of work! I like it very much, but there are some minor remarks:
- The namespace `"planning_pipelines"` is used several times throughout the code.
Wouldn't it make sense to introduce a variable or #define for it? This will facilitate a name change...
- `MoveGroupInterface::getInterfaceDescription()` returns the interface of the first pipeline, not necessarily the default one (as claimed).

What's your plan for merging this huge PR (32 commits)? From my point of view this PR comprises 3 large changes:
1. Moving moveit_cpp from moveit_ros/planning_interface -> moveit_ros/planning
2. Changing MoveGroupContext, MoveGroupExe to use MoveItCpp
3. Introducing support for multiple planning pipelines

Would be great to squash this into 3 corresponding commits.

moveit_ros/move_group/include/moveit/move_group/move_group_capability.h
```diff
@@ -92,6 +93,9 @@ class MoveGroupCapability
   moveit_msgs::PlanningScene clearSceneRobotState(const moveit_msgs::PlanningScene& scene) const;
   bool performTransform(geometry_msgs::PoseStamped& pose_msg, const std::string& target_frame) const;
 
+  bool resolvePlanningPipeline(const std::string& pipeline_id,
+                               planning_pipeline::PlanningPipelinePtr& planning_pipeline) const;
```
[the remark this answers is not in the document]
henningkayser: Agree, it's fixed as suggested

moveit_ros/move_group/include/moveit/move_group/move_group_capability.h
```diff
@@ -37,6 +37,7 @@
 #pragma once
 
 #include <moveit/macros/class_forward.h>
+#include <moveit/moveit_cpp/moveit_cpp.h>
```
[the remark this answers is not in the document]
henningkayser: Not directly, but it was used in the source file and in another capability. I moved the include to the corresponding source files.

moveit_ros/move_group/src/default_capabilities/query_planners_service_capability.cpp
henningkayser: Originally, to support specifying the default and to keep consistent with old launch files. Both points don't really hold true anymore so I fixed it as suggested -> all pipelines are specified by their ids.

henningkayser:
@rhaschke Thanks for your review, I think I've applied all of your suggestions.

As you see I squashed the commits into 4 separate commits. I thought it's not very straight-forward to separate multi-pipeline support from using MoveItCpp in MoveGroup altogether. Let me know if this works for you or not.

rhaschke (approved):
Thanks, @henningkayser, for cleaning up the commit history. I have pushed a few minor fixups. Please squash them with the corresponding main commit. From my point of view this is ready to merge then. :tada: 
I suggest you handle all related PRs at once:
- https://github.com/ros-planning/moveit_msgs/pull/95
- https://github.com/ros-planning/moveit_resources/pull/47
- https://github.com/ros-planning/moveit_tutorials/pull/549

In https://github.com/ros-planning/moveit_resources/pull/47 [you added a default definition](https://github.com/ros-planning/moveit_resources/pull/47/files#diff-a81ed1272c2cf99099b247440de05562cf2a5b23f9f10983fc87db248eefbae4R127) for the OMPL planner. I suggest, we should add corresponding defaults for all pipelines generated by MSA, i.e. in 
https://github.com/ros-planning/moveit/blob/c616ef9aa86b34c4b237ca7b1cdd66c19dfc6f20/moveit_setup_assistant/src/tools/moveit_config_data.cpp#L205

However, this can be handled by a separate PR. @jschleicher: Does the Pilz planner have a default?

jschleicher:
> Does the Pilz planner have a default?

No, it currently doesn't have a default planner id but requires the user to select between "PTP", "LIN" and "CIRC" explicitly. Do you think, there should be a default?

rhaschke:
> No, it currently doesn't have a default planner id but requires the user to select between "PTP", "LIN" and "CIRC" explicitly.

So, you just fail if nothing (or something else than these) is provided, right?

> Do you think, there should be a default?

No, not necessarily. I was just curious. If you had a default, we could highlight it in the rviz GUI as well.

jschleicher:
> you just fail if nothing (or something else than these) is provided, right?

exactly.

v4hn:
@henningkayser assuming you did not just break everything in unexpected ways: Thank you so much! :fireworks: 
It is great to finally have support for, e.g., Pilz and OMPL running in one setup. We had that debate already long before Pilz wrote their planner, but there was clearly not enough incentive to see this through. Also it turned out to be a lot more work than I predicted.

I know a few people who might be annoyed by the config changes and broken messages, but it was worth it! :1st_place_medal:

felixvd:
I wanted to write this the instant you pushed the update but I am *super
excited* about this and can’t wait to add PTP/LIN/CIRC to the Move Group
Interface. Now it was merged so fast that i didn’t even manage to review.

I will test this asap and won’t mind mild breakage at all. Big cheers

rhaschke:
We should prepare a new release asap to minimize issue reports like https://github.com/ros-planning/moveit_tutorials/issues/613. @tylerjw, what's your plan about this?

PR #2133, opened 2020-06-02
Add a function to check if the robot is stopped.

AndyZe:
This adds a function to `current_state_monitor` so we can easily check if the robot is stopped or not. We're using this to wait on planning a trajectory until the robot is stopped.

Credit to @henningkayser who wrote a lot of this code, I am mostly just rolling it into the `current_state_monitor`.

I would be interested in hints on how to write a unit/integration test for this!

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
+++ b/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
@@ -188,7 +188,7 @@ public:
    *  @return false If the check could not be completed
    */
   bool getIsRobotStopped(const std::string& group, double joint_velocity_tol, double state_wait_time,
-                         double& is_stopped);
+                         bool& is_stopped);
 
   /** @brief Allow the joint_state arrays velocity and effort to be copied into the robot state.
    *  This is useful in some but not all applications
--- a/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
+++ b/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
@@ -179,8 +179,19 @@ public:
     return error_;
   }
 
-  /** @brief Allow the joint_state arrrays velocity and effort to be copied into the robot state
-   *  this is useful in some but not all applications
+  /** @brief Check whether the robot is stopped. Two successive states are checked for position change,
+   *  and velocities are checked.
+   *  @param group The joint group of intereset.
+   *  @param joint_velocity_tol Largest allowable joint velocity to be considered stopped [rad/s].
+   *  @param double wait_time Wait time for retrieving robot's current state [seconds].
+   *  @param is_stopped Set to true if the robot is stopped.
+   *  @return false if the check could not be completed within wait_time
+   */
+  bool getIsRobotStopped(const std::string& group, double joint_velocity_tol, double state_wait_time,
+                         double& is_stopped);
+
+  /** @brief Allow the joint_state arrays velocity and effort to be copied into the robot state.
+   *  This is useful in some but not all applications
    */
   void enableCopyDynamics(bool enabled)
   {
--- a/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
+++ b/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
@@ -185,7 +185,7 @@ public:
    *  @param joint_velocity_tol Largest allowable joint velocity to be considered stopped [rad/s].
    *  @param double wait_time Wait time for retrieving robot's current state [seconds].
    *  @param is_stopped Set to true if the robot is stopped.
-   *  @return false if the check could not be completed within wait_time
+   *  @return false If the check could not be completed
    */
   bool getIsRobotStopped(const std::string& group, double joint_velocity_tol, double state_wait_time,
                          double& is_stopped);
--- a/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
@@ -41,6 +41,8 @@
 
 #include <limits>
 
+static const std::string LOGNAME = "current_state_monitor";
+
 planning_scene_monitor::CurrentStateMonitor::CurrentStateMonitor(const moveit::core::RobotModelConstPtr& robot_model,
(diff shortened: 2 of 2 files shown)


moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
```diff
@@ -179,8 +179,19 @@ class CurrentStateMonitor
     return error_;
   }
 
-  /** @brief Allow the joint_state arrrays velocity and effort to be copied into the robot state
-   *  this is useful in some but not all applications
+  /** @brief Check whether the robot is stopped. Two successive states are checked for position change,
+   *  and velocities are checked.
```
gavanderhoorn, line 183: What if the trajectory contains a phase where the robot doesn't physically move, but is still executing the trajectory?

This is very common in certain applications in which the process the robot is performing takes a bit longer.

AndyZe: that's a good point, I'm not sure how I could check that easily. Maybe RobotState should have a flag for whether it's currently executing a trajectory

simonschmeisser: Or if the robot is somewhat slow in starting. There were some funny bugs in industrial_robot_client due to using heuristics like this. I'm not saying this is necessarily a bad idea just thinking you might also want to consider information from `TrajectoryExecutionMonitor` (which should return success only once the movement has actually finished)

AndyZe: OK, you've convinced me this won't work for trajectory monitoring. Thanks for the pointer to `TrajectoryExecutionMonitor`.

We're actually using it to monitor a switch from streaming motion ("jogging") to trajectories, which is a little bit different. But I guess this functionality belongs in ros_control, probably.

AndyZe: I'll give it a couple more days to see if anybody else has comments, then close the issue

henningkayser:
Just for context: we use this approach not only for planned trajectories but for a couple of reasons:

* Robot appears to move (changes joint states) after FollowJointTrajectoryAction returns successfully
  * Cause 1: Low or delayed joint_states update rate causes CSM to be out-of-date (sometimes related to driver/controller)
  * Cause 2: FollowJointTrajectory action returns successfully before the robot has stopped
* Validation that robot comes to a stop when halting position-controlled motions, i.e. using JogArm
* Robot moves after switching ROS controllers (i.e. trajectory controller -> position controller)

I'm sure I'm forgetting some other cases here, but these occur across different robots and setups for various reasons. It is a workaround that really shouldn't be necessary if all hardware/controllers/drivers and ROS messaging was always perfect. But especially when dealing with experimental hardware this is not a good assumption. I see this function more as a smart sleep delay. Moving this to ROS-control would actually not be a bad idea, but this wouldn't fix CSM/`joint_states` not being up-to-date and we would still need a check on the client node side.
I thought it would make sense to bring a version of this upstream, I can't imagine we're the only ones dealing with the issues above.

simonschmeisser:
We have such workarounds as well ... we have them in/near TrajectoryExecutionMonitor where we can check for distance to the expected goal once the action tells us it's done.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -134,13 +136,79 @@ void planning_scene_monitor::CurrentStateMonitor::clearUpdateCallbacks()
...
 
+bool planning_scene_monitor::CurrentStateMonitor::getIsRobotStopped(const std::string& group, double joint_velocity_tol,
+                                                                    double state_wait_time, bool& is_stopped)
+{
+  // Initially assume the robot is not stopped
+  is_stopped = false;
+
+  // Constant joint group variables
+  const auto& joint_models = robot_model_->getActiveJointModels();
+  const std::size_t num_joints = joint_models.size();
```
henningkayser, line 147: This should be the number of joints of the JointModelGroup, not of the full robot model.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -41,6 +41,8 @@
 
 #include <limits>
 
+static const std::string LOGNAME = "current_state_monitor";
```
henningkayser, line 44: :+1:

henningkayser:
> We have such workarounds as well ... we have them in/near TrajectoryExecutionMonitor where we can check for distance to the expected goal once the action tells us it's done.

You mean `waitForRobotToStop()`? This one is really focused on the trajectory context and doesn't check for velocity, instead it checks for 3 consecutive stopped states. Maybe it would make sense to use this approach instead?

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -134,13 +136,79 @@ void planning_scene_monitor::CurrentStateMonitor::clearUpdateCallbacks()
...
+
+          if (joint_velocities_stopped)
+          {
+            // Check if joint distance towards previous state satisfies velocity threshold given the elapsed time
+            const double elapsed_seconds = (current_time - previous_time).toSec();
+            const double max_valid_joint_distance = joint_velocity_tol * num_joints * elapsed_seconds;
+            joint_positions_stopped = previous_state->distance(*current_state, joint_group) < max_valid_joint_distance;
+          }
+
+          if (joint_velocities_stopped && joint_positions_stopped)
```
henningkayser, line 183: I think we still need to check positions for multiple cycles

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -134,13 +136,79 @@ void planning_scene_monitor::CurrentStateMonitor::clearUpdateCallbacks()
...
+  {
+    if (waitForCompleteState(state_wait_time))
+    {
+      const std::pair<robot_state::RobotStatePtr, ros::Time> state_and_time = getCurrentStateAndTime();
+      const auto& current_state = state_and_time.first;
+      const auto& current_time = state_and_time.second;
+
+      if (previous_state)
+      {
+        if (current_state->hasVelocities())
```
henningkayser, line 163: I think in the original implementation we allowed the state to not have velocities (which is the case for some simulations) and just checked for positions instead.

AndyZe:
Closing the stale PR

Issue #2063, opened 2020-05-07
Add collision objects to scene from RViz


v4hn:
It would be useful to add collision objects to the planning scene through the RViz panel.
The "Scene Objects" tab already allows to move and scale objects (no flexible resize though...).

Adding objects, especially resizable cubes and cylinders from this GUI would make setting up a planning scene a much more interactive experience instead of changing parameters in some file and loading it over and over again.

felixvd:
Strongly agree, this would be a great contribution.

dave992:
I will have a look at this.

jrgnicho:
I'm currently addressing this

rhaschke:
Fixed via #2198.

PR #2137, opened 2020-06-02
Add primitive shapes widgets

jrgnicho:
### Description
Addresses issue #2063 

Adds widgets into the moveit motion planning plugin that allows adding primitive shapes into the planning scene
rviz_screenshot_2020_06_02-16_13_31

### Checklist
- [x ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [x ] Include a screenshot if changing a GUI

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -128,7 +128,9 @@ MotionPlanningFrame::MotionPlanningFrame(MotionPlanningDisplay* pdisplay, rviz::
   connect(ui_->planning_scene_tree, SIGNAL(itemChanged(QTreeWidgetItem*, int)), this,
           SLOT(warehouseItemNameChanged(QTreeWidgetItem*, int)));
   connect(ui_->reset_db_button, SIGNAL(clicked()), this, SLOT(resetDbButtonClicked()));
-  connect(ui_->add_shape_button, &QPushButton::clicked, [this]() { addPrimitiveShape(); });
+
+  connect(ui_->add_shape_button, &QPushButton::clicked, this, &MotionPlanningFrame::addPrimitiveShape);
+  connect(ui_->shapes_combo_box, &QComboBox::currentTextChanged, this, &MotionPlanningFrame::shapesComboBoxChanged);
   connect(ui_->export_scene_text_button, SIGNAL(clicked()), this, SLOT(exportAsTextButtonClicked()));
   connect(ui_->import_scene_text_button, SIGNAL(clicked()), this, SLOT(importFromTextButtonClicked()));
   connect(ui_->load_state_button, SIGNAL(clicked()), this, SLOT(loadStateButtonClicked()));
@@ -430,18 +432,17 @@ void MotionPlanningFrame::sceneUpdate(planning_scene_monitor::PlanningSceneMonit
 void MotionPlanningFrame::addPrimitiveShape()
 {
   static const double MIN_VAL = 1e-6;
-  static const std::map<std::string, shapes::ShapeType> SHAPES_MAP = {
-    { "box", shapes::BOX }, { "sphere", shapes::SPHERE }, { "cone", shapes::CONE }, { "cylinder", shapes::CYLINDER }
-  };
 
   if (!planning_display_->getPlanningSceneMonitor())
   {
     return;
   }
 
-  // get side length
-  double side_length = ui_->shape_size_spin_box->value();
-  if (side_length < MIN_VAL)
+  // get size values
+  double x_length = ui_->shape_size_x_spin_box->isEnabled() ? ui_->shape_size_x_spin_box->value() : MIN_VAL;
+  double y_length = ui_->shape_size_y_spin_box->isEnabled() ? ui_->shape_size_y_spin_box->value() : MIN_VAL;
+  double z_length = ui_->shape_size_z_spin_box->isEnabled() ? ui_->shape_size_z_spin_box->value() : MIN_VAL;
+  if (x_length < MIN_VAL || y_length < MIN_VAL || z_length < MIN_VAL)
   {
     QMessageBox::warning(this, QString("Side length value is too small"),
                          QString("Value needs to be  >= '").append(std::to_string(MIN_VAL).c_str()).append("'"));
@@ -453,7 +454,7 @@ void MotionPlanningFrame::addPrimitiveShape()
   if (SHAPES_MAP.count(selected_shape) == 0)
   {
     QMessageBox::warning(this, QString("Unsupported shape"),
-                         QString("The '").append(selected_shape.c_str()).append("' is not supported."));
+                         QString("The '%1' shape is not supported.").arg(selected_shape.c_str()));
     return;
   }
   shapes::ShapeType shape_type = SHAPES_MAP.at(selected_shape);
@@ -461,17 +462,20 @@ void MotionPlanningFrame::addPrimitiveShape()
   switch (shape_type)
   {
     case shapes::BOX:
-      shape = std::make_shared<shapes::Box>(side_length, side_length, side_length);
+      shape = std::make_shared<shapes::Box>(x_length, y_length, z_length);
       break;
     case shapes::SPHERE:
-      shape = std::make_shared<shapes::Sphere>(0.5 * side_length);
+      shape = std::make_shared<shapes::Sphere>(0.5 * x_length);
       break;
     case shapes::CONE:
-      shape = std::make_shared<shapes::Cone>(0.5 * side_length, side_length);
+      shape = std::make_shared<shapes::Cone>(0.5 * x_length, z_length);
       break;
     case shapes::CYLINDER:
-      shape = std::make_shared<shapes::Cylinder>(0.5 * side_length, side_length);
+      shape = std::make_shared<shapes::Cylinder>(0.5 * x_length, z_length);
       break;
+    default:
+      QMessageBox::warning(this, QString("Unsupported shape"),
+                           QString("The '%1' is not supported.").arg(selected_shape.c_str()));
   }
 
   // naming object
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -128,6 +128,9 @@ MotionPlanningFrame::MotionPlanningFrame(MotionPlanningDisplay* pdisplay, rviz::
   connect(ui_->planning_scene_tree, SIGNAL(itemChanged(QTreeWidgetItem*, int)), this,
           SLOT(warehouseItemNameChanged(QTreeWidgetItem*, int)));
   connect(ui_->reset_db_button, SIGNAL(clicked()), this, SLOT(resetDbButtonClicked()));
+  connect(ui_->add_shape_button,&QPushButton::clicked, [this](){
+    addPrimitiveShape();
+  });
   connect(ui_->export_scene_text_button, SIGNAL(clicked()), this, SLOT(exportAsTextButtonClicked()));
   connect(ui_->import_scene_text_button, SIGNAL(clicked()), this, SLOT(importFromTextButtonClicked()));
   connect(ui_->load_state_button, SIGNAL(clicked()), this, SLOT(loadStateButtonClicked()));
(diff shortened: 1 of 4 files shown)


moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -426,6 +427,70 @@ void MotionPlanningFrame::sceneUpdate(planning_scene_monitor::PlanningSceneMonit
...
+  if (side_length < MIN_VAL)
+  {
+    QMessageBox::warning(this, QString("Side length value is too small"),
+                         QString("Value needs to be  >= '").append(std::to_string(MIN_VAL).c_str()).append("'"));
+    return;
+  }
+
+  // get selected shape
+  std::string selected_shape = ui_->shapes_combo_box->currentText().toStdString();
+  if (SHAPES_MAP.count(selected_shape) == 0)
```
JafarAbdi, line 453: You could also use `auto selected_shape_it =  SHAPES_MAP.find(selected_shape)` and test if it's equal to `SHAPES_MAP.end()` and replace `SHAPES_MAP.at(selected_shape);` with the object it point to

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -128,6 +128,7 @@ MotionPlanningFrame::MotionPlanningFrame(MotionPlanningDisplay* pdisplay, rviz::
   connect(ui_->planning_scene_tree, SIGNAL(itemChanged(QTreeWidgetItem*, int)), this,
           SLOT(warehouseItemNameChanged(QTreeWidgetItem*, int)));
   connect(ui_->reset_db_button, SIGNAL(clicked()), this, SLOT(resetDbButtonClicked()));
+  connect(ui_->add_shape_button, &QPushButton::clicked, [this]() { addPrimitiveShape(); });
```
simonschmeisser, line 131:
```suggestion
  connect(ui_->add_shape_button, &QPushButton::clicked, this, &MotionPlanningFrame::addPrimitiveShape);
```

No need for a lambda here. Also please be careful with this short version as `this` or other pointers bound in the lambda could be destroyed elsewhere and then crash on calling the lambda. There is a preferred variant with four arguments where the third denotes the scope of the connection and if either sender or receiver are destructed:
```
connect(ui_->add_shape_button, &QPushButton::clicked, this, [this]() { addPrimitiveShape(); });
```
But as stated above, this shouldn't be necessary here

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -426,6 +427,70 @@ void MotionPlanningFrame::sceneUpdate(planning_scene_monitor::PlanningSceneMonit
...
+                         QString("Value needs to be  >= '").append(std::to_string(MIN_VAL).c_str()).append("'"));
+    return;
+  }
+
+  // get selected shape
+  std::string selected_shape = ui_->shapes_combo_box->currentText().toStdString();
+  if (SHAPES_MAP.count(selected_shape) == 0)
+  {
+    QMessageBox::warning(this, QString("Unsupported shape"),
+                         QString("The '").append(selected_shape.c_str()).append("' is not supported."));
```
simonschmeisser, line 456:
```suggestion
                         QString("The shape '%1' is not supported.").arg(selected_shape.c_str()));
```

This improves readability and also allows rearranging if we should ever start translating MoveIt into languages other than English.

JafarAbdi:
Thanks @jrgnicho for the PR, the CI is failing because you have unhandled ShapeType values [see](https://travis-ci.org/github/ros-planning/moveit/jobs/694034113#L1273), one way to handle it would be

```cpp
  std::string selected_shape_name = ui_->shapes_combo_box->currentText().toStdString();
  auto selected_shape = SHAPES_MAP.find(selected_shape_name);
  shapes::ShapeType shape_type = (selected_shape != SHAPES_MAP.end()) ? selected_shape->second : shapes::UNKNOWN_SHAPE;
  shapes::ShapeConstPtr shape;
  switch (shape_type)
  {
    case shapes::BOX:
      shape = std::make_shared<shapes::Box>(side_length, side_length, side_length);
      break;
    case shapes::SPHERE:
      shape = std::make_shared<shapes::Sphere>(0.5 * side_length);
      break;
    case shapes::CONE:
      shape = std::make_shared<shapes::Cone>(0.5 * side_length, side_length);
      break;
    case shapes::CYLINDER:
      shape = std::make_shared<shapes::Cylinder>(0.5 * side_length, side_length);
      break;
    default:
      QMessageBox::warning(this, QString("Unsupported shape"),
                           QString("The '").append(selected_shape_name.c_str()).append("' is not supported."));
      return;
  }
```

AndyZe:
I would like this feature a lot!

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
jrgnicho: @simonschmeisser  The `this` pointer points to the panel itself so I don't see how it could ever get destroyed in this context but I'll make the change

jrgnicho:
> Thanks @jrgnicho for the PR, the CI is failing because you have unhandled ShapeType values [see](https://travis-ci.org/github/ros-planning/moveit/jobs/694034113#L1273), one way to handle it would be
> 
> ```c++
>   std::string selected_shape_name = ui_->shapes_combo_box->currentText().toStdString();
>   auto selected_shape = SHAPES_MAP.find(selected_shape_name);
>   shapes::ShapeType shape_type = (selected_shape != SHAPES_MAP.end()) ? selected_shape->second : shapes::UNKNOWN_SHAPE;
>   shapes::ShapeConstPtr shape;
>   switch (shape_type)
>   {
>     case shapes::BOX:
>       shape = std::make_shared<shapes::Box>(side_length, side_length, side_length);
>       break;
>     case shapes::SPHERE:
>       shape = std::make_shared<shapes::Sphere>(0.5 * side_length);
>       break;
>     case shapes::CONE:
>       shape = std::make_shared<shapes::Cone>(0.5 * side_length, side_length);
>       break;
>     case shapes::CYLINDER:
>       shape = std::make_shared<shapes::Cylinder>(0.5 * side_length, side_length);
>       break;
>     default:
>       QMessageBox::warning(this, QString("Unsupported shape"),
>                            QString("The '").append(selected_shape_name.c_str()).append("' is not supported."));
>       return;
>   }
> ```

@JafarAbdi thanks for the tip, the logic prior to the switch statement makes it so that an unhandled shape type can never be assigned but I'll make the change to comply with the CI

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
simonschmeisser: You're right, in this context it won't happen. `ui_->shape_button` is a child of this and therefore gets destroyed before `this` and cancels the `connection` ... just wanted to point out the general danger of that variant of `connect()`

My main remark was about the number of brackets and the unnecessary lambda however :)

jrgnicho:
I've made the suggested changes and added double spin boxes to set the x, y and z dimensions which get toggled based on the shape
add_primitive_shapes_widgets_v3

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -1185,66 +1204,132 @@
...
+            </item>
+           </widget>
+          </item>
+          <item>
+           <widget class="QDoubleSpinBox" name="shape_size_x_spin_box">
+            <property name="singleStep">
+             <double>0.100000000000000</double>
+            </property>
+            <property name="value">
+             <double>1.000000000000000</double>
```
felixvd, line 1293: Would much prefer a default value of `0.1` and a step value of `0.01`. It will be a smoother experience and fit the dimensions of the demo/tutorial robot.

felixvd: I guess it's not really important, I just feel like the default will force users to zoom out when their robot fills most of the screen. `0.2` and `0.05` could be a good compromise.

Screenshot from 2020-06-04 18-34-39

Screenshot from 2020-06-04 18-24-21

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -426,6 +427,70 @@ void MotionPlanningFrame::sceneUpdate(planning_scene_monitor::PlanningSceneMonit
     planning_display_->addMainLoopJob(boost::bind(&MotionPlanningFrame::populateCollisionObjectsList, this));
 }
 
+void MotionPlanningFrame::addPrimitiveShape()
+{
+  static const double MIN_VAL = 1e-6;
+  static const std::map<std::string, shapes::ShapeType> SHAPES_MAP = {
+    { "box", shapes::BOX }, { "sphere", shapes::SPHERE }, { "cone", shapes::CONE }, { "cylinder", shapes::CYLINDER }
```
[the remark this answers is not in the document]
jrgnicho: Would't that be redundant since there are already is import buttons?

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -1102,34 +1102,53 @@
...
+            <property name="text">
+             <string>Clear</string>
             </property>
+           </widget>
+          </item>
+          <item row="2" column="0">
+           <widget class="QPushButton" name="import_file_button">
             <property name="text">
-             <string>Status</string>
+             <string>Import &amp;File</string>
```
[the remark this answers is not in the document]
jrgnicho: I don't recall changing this, I made my edits with QT designer so sometimes things get changed without one knowing about it.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -1185,66 +1204,106 @@
...
-          <item row="3" column="0">
-           <widget class="QPushButton" name="remove_object_button">
-            <property name="text">
-             <string>Remove</string>
-            </property>
+         </layout>
+        </widget>
+       </item>
+       <item row="3" column="1">
+        <widget class="QGroupBox" name="groupBox_4">
```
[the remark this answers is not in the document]
jrgnicho: I didn't want to add a second column as there were some concerns with increasing the panel's height.  Furthermore, having yet another "scale" section might be confusing, would't "size" be more appropriate.
See my comments above on the "mesh" option

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
jrgnicho: @felixvd  I agree, I'll update those values

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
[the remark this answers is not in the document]
felixvd: I drew this when the new section was still on the left. It's already good now, I'll resolve

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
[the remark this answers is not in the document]
felixvd: I had a closer look, and you're right: the "Import File" button actually accepts an STL file. I just assumed it's a scene file or something, since nothing tells you about the expected format. I'll fix that in a separate PR.

jrgnicho:
> Nothing to add, this is great.

@felixvd thanks for reviewing

jrgnicho:
This should be ready to merge unless someone has additional change requests

rhaschke:
Great work! I file a new PR resolving the conflicts by a rebase in #2198.

Issue #2075, opened 2020-05-08
memroy leaks for days


tylerjw:
I called out the first chunk of memory leaks found using this PR: #2057

This is to track the fact that there are more and I didn't continue to copy and paste from the logs from them.  Once the others are fixed or resolved some other way we should re-run the test to see what memory leaks persist.

jliukkonen:
I'll take a look into those leaks today as my World MoveIt Day task. Hopefully manage to squash at least couple of them.

PR #2138, opened 2020-06-02
Fix geometric shapes related memory leaks

jliukkonen:
### Couple fixes to memory leaks caused by raw pointer usage with `geometric_shapes` objects.

These fixes are related to the efforts and finding in this issue [https://github.com/ros-planning/moveit/issues/2075](https://github.com/ros-planning/moveit/issues/2075). However, this PR does not fix the leaks reported by @tylerjw, but rather these are new findings made using address sanitizer.

Running address sanitizer revealed couple memory leaks related raw pointer usage with `geometric_shapes` objects. I added two fixes to this PR since they are both very small.

First fix is wrapping return value of `shapes::constructShapeFromText(in)` in shared_ptr so that if the geometry being loaded is corrupted or malformed, the constructed shape will always be freed undepending on the code path.

Second fix is removing call to `clone()` in `BodyDecomposition::init()`. Prior to the fix, each of the cloned objects was heap allocated and raw pointer was passed to `addBody` without calling free on the objects at any point, which caused every single object to leak. Looking more closely at the `addBody` function I realized that there doesn't seem to be any reason to call clone there so I removed the operation altogether. Now, the raw pointers passed to the function point to the `shared_ptr<Shape>` objects which will get freed correctly eventually thanks to reference counting.

### DIFF
--- a/moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
+++ b/moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
@@ -280,7 +280,7 @@ void collision_detection::BodyDecomposition::init(const std::vector<shapes::Shap
   bodies_.clear();
   for (unsigned int i = 0; i < shapes.size(); i++)
   {
-    bodies_.addBody(shapes[i]->clone(), poses[i], padding);
+    bodies_.addBody(shapes[i].get(), poses[i], padding);
   }
 
   // collecting collision spheres
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1018,8 +1018,8 @@ bool PlanningScene::loadGeometryFromStream(std::istream& in, const Eigen::Isomet
       in >> shape_count;
       for (std::size_t i = 0; i < shape_count && in.good() && !in.eof(); ++i)
       {
-        shapes::Shape* s = shapes::constructShapeFromText(in);
-        if (!s)
+        const auto shape = shapes::ShapeConstPtr(shapes::constructShapeFromText(in));
+        if (!shape)
         {
           ROS_ERROR_NAMED(LOGNAME, "Failed to load shape from scene file");
           return false;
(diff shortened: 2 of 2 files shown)


moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
```diff
@@ -280,7 +280,7 @@ void collision_detection::BodyDecomposition::init(const std::vector<shapes::Shap
   bodies_.clear();
   for (unsigned int i = 0; i < shapes.size(); i++)
   {
-    bodies_.addBody(shapes[i]->clone(), poses[i], padding);
+    bodies_.addBody(shapes[i].get(), poses[i], padding);
```
rhaschke, line 283: As the passed `shapes` are a _const_ vector of _const_ ShapePtrs, I'm tempted to assume that `clone()` was correct. It doesn't make sense to extract the raw pointer from the shared ptr and continue using the raw pointer, while the shared pointer might get destroyed...
I think the correct way of handling this is to
a) either store shared ptrs in `bodies_` as well, or
b) free the raw shape pointers before doing `bodies_.clear()`.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1018,8 +1018,8 @@ bool PlanningScene::loadGeometryFromStream(std::istream& in, const Eigen::Isomet
       in >> shape_count;
       for (std::size_t i = 0; i < shape_count && in.good() && !in.eof(); ++i)
       {
-        shapes::Shape* s = shapes::constructShapeFromText(in);
-        if (!s)
+        const auto shape = shapes::ShapeConstPtr(shapes::constructShapeFromText(in));
```
rhaschke, line 1021: These changes are reasonable.

moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
jliukkonen: Can you describe a situation where the shared_pts might get destroyed? What happens there, is that values from those Shape objects are used to construct (and add) new bodies but the shape objects themselves are not stored as part of bodies. That is, they are not required after the function call to init().

JafarAbdi: Shouldn't the [destroctor/clear](https://github.com/ros-planning/geometric_shapes/blob/melodic-devel/src/bodies.cpp#L1311) functions take care of freeing the memory .?

rhaschke: Sorry for the noise. I was thinking that `addBody(Shape*)` just stores a pointer to the passed shape (which then might end up in a dangling pointer). However, actually `addBody()` creates a copy internally:
https://github.com/ros-planning/geometric_shapes/blob/df0478870b8592ce789ee1919f3124058c4327d7/src/body_operations.cpp#L43-L66

PR #2142, opened 2020-06-06
Clean up Rviz plugin, fix mesh scaling

felixvd:
### Description

While reviewing #2137, I realized that the Rviz panel has an option to import meshes that I have never discovered, because the buttons and dialog windows are not very instructive.

This PR should clarify the interface as shown in the screenshot below, and adds a dialog to fix STL files that are encoded with mm units. Sadly, scaling the mesh did not work as I expected, as described [here](https://github.com/ros-planning/geometric_shapes/issues/148). I will remove the WIP when that's fixed.

gui

### Checklist

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -172,8 +172,8 @@ private Q_SLOTS:
   void onClearOctomapClicked();
 
   // Scene Objects tab
-  void import3DObjectFromFileButtonClicked();
-  void import3DObjectFromUrlButtonClicked();
+  void importObjectFromFileButtonClicked();
+  void importObjectFromUrlButtonClicked();
   void clearSceneButtonClicked();
   void sceneScaleChanged(int value);
   void sceneScaleStartChange();
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -97,6 +97,8 @@ static const std::string TAB_SCENES = "Stored Scenes";
 static const std::string TAB_STATES = "Stored States";
 static const std::string TAB_STATUS = "Status";
 
+static const double LARGE_MESH_THRESHOLD = 10.0;
+
 class MotionPlanningFrame : public QWidget
 {
   friend class MotionPlanningDisplay;
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -186,8 +186,8 @@ private Q_SLOTS:
   void collisionObjectChanged(QListWidgetItem* item);
   void imProcessFeedback(visualization_msgs::InteractiveMarkerFeedback& feedback);
   void copySelectedCollisionObject();
-  void exportAsTextButtonClicked();
-  void importFromTextButtonClicked();
+  void exportGeometryAsTextButtonClicked();
+  void importGeometryFromTextButtonClicked();
 
   // Stored scenes tab
   void saveSceneButtonClicked();
@@ -251,8 +251,8 @@ private:
   void renameCollisionObject(QListWidgetItem* item);
   void attachDetachCollisionObject(QListWidgetItem* item);
   void populateCollisionObjectsList();
-  void computeImportFromText(const std::string& path);
-  void computeExportAsText(const std::string& path);
+  void computeImportGeometryFromText(const std::string& path);
+  void computeExportGeometryAsText(const std::string& path);
   visualization_msgs::InteractiveMarker
   createObjectMarkerMsg(const collision_detection::CollisionEnv::ObjectConstPtr& obj);
 
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
@@ -172,8 +172,8 @@ private Q_SLOTS:
   void onClearOctomapClicked();
 
   // Scene Objects tab
-  void importFileButtonClicked();
-  void importUrlButtonClicked();
+  void import3DObjectFromFileButtonClicked();
+  void import3DObjectFromUrlButtonClicked();
   void clearSceneButtonClicked();
   void sceneScaleChanged(int value);
   void sceneScaleStartChange();
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -128,8 +128,8 @@ MotionPlanningFrame::MotionPlanningFrame(MotionPlanningDisplay* pdisplay, rviz::
   connect(ui_->planning_scene_tree, SIGNAL(itemChanged(QTreeWidgetItem*, int)), this,
           SLOT(warehouseItemNameChanged(QTreeWidgetItem*, int)));
   connect(ui_->reset_db_button, SIGNAL(clicked()), this, SLOT(resetDbButtonClicked()));
-  connect(ui_->export_scene_text_button, SIGNAL(clicked()), this, SLOT(exportAsTextButtonClicked()));
-  connect(ui_->import_scene_text_button, SIGNAL(clicked()), this, SLOT(importFromTextButtonClicked()));
+  connect(ui_->export_scene_geometry_text_button, SIGNAL(clicked()), this, SLOT(exportGeometryAsTextButtonClicked()));
+  connect(ui_->import_scene_geometry_text_button, SIGNAL(clicked()), this, SLOT(importGeometryFromTextButtonClicked()));
   connect(ui_->load_state_button, SIGNAL(clicked()), this, SLOT(loadStateButtonClicked()));
   connect(ui_->save_start_state_button, SIGNAL(clicked()), this, SLOT(saveStartStateButtonClicked()));
   connect(ui_->save_goal_state_button, SIGNAL(clicked()), this, SLOT(saveGoalStateButtonClicked()));
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -460,7 +460,8 @@ void MotionPlanningFrame::importResource(const std::string& path)
       if (object_is_very_large)
       {
         QMessageBox msg_box;
-        msg_box.setText("The object is very large (greater than 10 m). The file may be in millimeters instead of meters.");
+        msg_box.setText(
+            "The object is very large (greater than 10 m). The file may be in millimeters instead of meters.");
         msg_box.setInformativeText("Attempt to fix the size by shrinking the object?");
         msg_box.setStandardButtons(QMessageBox::Yes | QMessageBox::No | QMessageBox::Cancel);
         msg_box.setDefaultButton(QMessageBox::Yes);
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -445,6 +445,26 @@ void MotionPlanningFrame::importResource(const std::string& path)
         return;
       }
 
+      // Center the mesh so scaling applies correctly
+      double sx = 0.0, sy = 0.0, sz = 0.0;
+      for (unsigned int i = 0; i < mesh->vertex_count; ++i)
+      {
+        unsigned int i3 = i * 3;
+        sx += mesh->vertices[i3];
+        sy += mesh->vertices[i3 + 1];
+        sz += mesh->vertices[i3 + 2];
+      }
+      sx /= (double)mesh->vertex_count;
+      sy /= (double)mesh->vertex_count;
+      sz /= (double)mesh->vertex_count;
+      for (unsigned int i = 0; i < mesh->vertex_count; ++i)
+      {
+        unsigned int i3 = i * 3;
+        mesh->vertices[i3] -= sx;
+        mesh->vertices[i3 + 1] -= sy;
+        mesh->vertices[i3 + 2] -= sz;
+      }
+
       // If the object is very large, ask the user if the scale should be reduced.
       bool object_is_very_large = false;
       for (unsigned int i = 0; i < mesh->vertex_count; i++)
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -463,39 +463,17 @@ void MotionPlanningFrame::importResource(const std::string& path)
         msg_box.setText(
             "The object is very large (greater than 10 m). The file may be in millimeters instead of meters.");
         msg_box.setInformativeText("Attempt to fix the size by shrinking the object?");
-        msg_box.setStandardButtons(QMessageBox::Yes | QMessageBox::No | QMessageBox::Cancel);
+        msg_box.setStandardButtons(QMessageBox::Yes | QMessageBox::No);
         msg_box.setDefaultButton(QMessageBox::Yes);
-        int ret = msg_box.exec();
-        switch (ret)
+        if (msg_box.exec() == QMessageBox::Yes)
         {
-          case QMessageBox::Yes:
-            // Center the mesh so scaling applies correctly
-            {
-              double sx = 0.0, sy = 0.0, sz = 0.0;
-              for (unsigned int i = 0; i < mesh->vertex_count; ++i)
-              {
-                unsigned int i3 = i * 3;
-                sx += mesh->vertices[i3];
-                sy += mesh->vertices[i3 + 1];
-                sz += mesh->vertices[i3 + 2];
-              }
-              sx /= (double)mesh->vertex_count;
-              sy /= (double)mesh->vertex_count;
-              sz /= (double)mesh->vertex_count;
-              for (unsigned int i = 0; i < mesh->vertex_count; ++i)
-              {
-                unsigned int i3 = i * 3;
-                mesh->vertices[i3] -= sx;
-                mesh->vertices[i3 + 1] -= sy;
-                mesh->vertices[i3 + 2] -= sz;
-              }
-              mesh->scaleAndPadd(0.001, 0);
-            }
-            break;
-          case QMessageBox::No:
-            break;
-          default:
-            return;  // Pressed cancel, do nothing
+          for (unsigned int i = 0; i < mesh->vertex_count; ++i)
+          {
+            unsigned int i3 = i * 3;
+            mesh->vertices[i3] *= 0.001;
+            mesh->vertices[i3 + 1] *= 0.001;
+            mesh->vertices[i3 + 2] *= 0.001;
+          }
         }
       }
 
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -449,8 +449,9 @@ void MotionPlanningFrame::importResource(const std::string& path)
       bool object_is_very_large = false;
       for (unsigned int i = 0; i < mesh->vertex_count; i++)
       {
-        if ((abs(mesh->vertices[i * 3 + 0]) > 10) || (abs(mesh->vertices[i * 3 + 1]) > 10) ||
-            (abs(mesh->vertices[i * 3 + 2]) > 10))
+        if ((abs(mesh->vertices[i * 3 + 0]) > LARGE_MESH_THRESHOLD) ||
+            (abs(mesh->vertices[i * 3 + 1]) > LARGE_MESH_THRESHOLD) ||
+            (abs(mesh->vertices[i * 3 + 2]) > LARGE_MESH_THRESHOLD))
         {
           object_is_very_large = true;
           break;
@@ -459,7 +460,7 @@ void MotionPlanningFrame::importResource(const std::string& path)
       if (object_is_very_large)
       {
         QMessageBox msg_box;
-        msg_box.setText("The object is very large. The file may be in millimeters instead of meters.");
+        msg_box.setText("The object is very large (greater than 10 m). The file may be in millimeters instead of meters.");
         msg_box.setInformativeText("Attempt to fix the size by shrinking the object?");
         msg_box.setStandardButtons(QMessageBox::Yes | QMessageBox::No | QMessageBox::Cancel);
         msg_box.setDefaultButton(QMessageBox::Yes);
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -445,26 +445,6 @@ void MotionPlanningFrame::importResource(const std::string& path)
         return;
       }
 
-      // Center the mesh so scaling applies correctly
-      double sx = 0.0, sy = 0.0, sz = 0.0;
-      for (unsigned int i = 0; i < mesh->vertex_count; ++i)
-      {
-        unsigned int i3 = i * 3;
-        sx += mesh->vertices[i3];
-        sy += mesh->vertices[i3 + 1];
-        sz += mesh->vertices[i3 + 2];
-      }
-      sx /= (double)mesh->vertex_count;
-      sy /= (double)mesh->vertex_count;
-      sz /= (double)mesh->vertex_count;
-      for (unsigned int i = 0; i < mesh->vertex_count; ++i)
-      {
-        unsigned int i3 = i * 3;
-        mesh->vertices[i3] -= sx;
-        mesh->vertices[i3 + 1] -= sy;
-        mesh->vertices[i3 + 2] -= sz;
-      }
-
       // If the object is very large, ask the user if the scale should be reduced.
       bool object_is_very_large = false;
       for (unsigned int i = 0; i < mesh->vertex_count; i++)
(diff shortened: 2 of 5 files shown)


v4hn:
> as shown in the screenshot below

There are no screenshots and CI failed. :)

felixvd:
I mixed up the screenshots and only uploaded the one in the other thread. I fixed the OP.

I think the new commit fixes the build warning, but I don't see the warning locally either way. Do I have to enable something to show it? `catkin config` doesn't have any `--no-warnings` settings or similar that I can see. I did run `clang-format` and `clang-tidy` before submitting the PR but apparently that didn't catch it.

rhaschke:
We are build MoveIt on Travis with flags 
`CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls"`. To see the same issues locally, you should use the same flags.

felixvd:
Thanks. Is that worth adding to the doc somewhere?

mamoll:
💯 This is a great idea! I ran into this issue myself when I was trying to create a realistic planning scene by populating it with some meshes I had downloaded. Besides scaling, you could also add some logic for positioning: if the bounding box of the imported mesh doesn't overlap with the bounding box of the current planning scene, ask the user whether the imported mesh should be centered in the current scene.

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
```diff
@@ -172,8 +172,8 @@ private Q_SLOTS:
   void onClearOctomapClicked();
 
   // Scene Objects tab
-  void importFileButtonClicked();
-  void importUrlButtonClicked();
+  void import3DObjectFromFileButtonClicked();
```
rhaschke, line 175: IMO, you should revert renamings of all (internal) symbols. At most, I agree with import**From**[File|Url]...
If you want to clarify (for the user) that a single object (instead of a full scene) is imported, provide a tooltip to the corresponding buttons.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -448,6 +445,42 @@ void MotionPlanningFrame::importResource(const std::string& path)
         return;
       }
 
+      // If the object is very large, ask the user if the scale should be reduced.
+      bool object_is_very_large = false;
+      for (unsigned int i = 0; i < mesh->vertex_count; i++)
+      {
+        if ((abs(mesh->vertices[i * 3 + 0]) > 10) || (abs(mesh->vertices[i * 3 + 1]) > 10) ||
```
rhaschke, line 452: Just, checking for components larger than 10 is not sufficient. What if a correctly scaled object is just shifted by 10 units? As scaling can be done easily in the rviz gui as well, I actually don't see the need for this extra code here.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -515,7 +548,7 @@ void MotionPlanningFrame::importResource(const std::string& path)
     }
     else
     {
-      QMessageBox::warning(this, QString("Import error"), QString("Unable to import scene"));
+      QMessageBox::warning(this, QString("Import error"), QString("Unable to import 3D object"));
```
rhaschke, line 551: All objects in our context are 3D.
```suggestion
      QMessageBox::warning(this, QString("Import error"), QString("Unable to import object"));
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -73,17 +73,17 @@ QString subframe_poses_to_qstring(const moveit::core::FixedTransformsMap& subfra
 
 namespace moveit_rviz_plugin
 {
-void MotionPlanningFrame::importFileButtonClicked()
+void MotionPlanningFrame::import3DObjectFromFileButtonClicked()
 {
-  QString path = QFileDialog::getOpenFileName(this, tr("Import Object"));
+  QString path = QFileDialog::getOpenFileName(this, tr("Import 3D Object (e.g. .stl, "));
```
rhaschke, line 78: I suggest [using file filters](https://doc.qt.io/qt-5/qfiledialog.html#getOpenFileName):
```suggestion
  QString path = QFileDialog::getOpenFileName(this, tr("Import Object Mesh"), QString(), "CAD files (*.stl *.obj)");
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -1216,9 +1216,9 @@
            </widget>
           </item>
           <item row="1" column="0">
-           <widget class="QPushButton" name="import_file_button">
+           <widget class="QPushButton" name="import_3d_obj_file_button">
             <property name="text">
-             <string>Import &amp;File</string>
+             <string>Import &amp;Object</string>
```
rhaschke, line 1221: An "object" is imported in both cases. Hence, "File" to indicate that we import from file was correct, wasn't it? Please revert and better use tooltips if you want to be more specific.
```suggestion
             <string>Import &amp;File</string>
```

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
felixvd: I'd go for "Object" over "3DObject", but I would like to separate a saved Scene from a saved Object more clearly. Is there a reason not to be descriptive?

I think the fact that these functions are closely related and somewhat easy to confuse is that `computeLoadSceneButton` is defined in `motion_planning_frame_objects.cpp` instead of `motion_planning_frame_scenes.cpp` . I assume the separation was introduced later during development.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: I doubt that many correctly scaled objects are defined 10 meters away from their origin. It sounds like a rare case - surely these users can live with clicking "No" in a dialogue window. 

The more common case is a user loading an STL that is defined in millimeters rather than meters, and then having to scale it by exactly `0.001`. Forcing the user to do that with a slider isn't only bad UX, but it doesn't even work: it leaves a gigantic interactive marker in the scene. Feel free to try with [this file](https://github.com/ros-planning/geometric_shapes/files/4741236/03-PANEL2.zip).

Anyway, in my opinion it's 100% worth checking for this, and I am fine with the check not being perfect, since it only triggers a user dialogue.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
felixvd: Good idea, but we would have to maintain the lists of available file formats though. Do we want that?

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
felixvd: I find the hint that this imports a mesh rather than some other format (a scene? a YAML? some other list of objects?) is hard to discover. I think it should be more visible than a tooltip. How about `Import Mesh`?

To be clear, the problem is not that the button is incorrect, it's that it doesn't convey what a common user would be looking for or understand at first sight. Reading "Import File" as a new user makes me think "What kind of file? Ugh, I'll look into it later", "Import Mesh" makes me think "Oh cool, now I know I can do that".

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
rhaschke: As a fallback, you can still provide an "all files (*.*)" filter.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
rhaschke: The key point is that both buttons essentially import an object mesh, once from a local file, once from an url.
Hence the button names should distinguish file and url.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: Done

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
felixvd: True, it is a bit of a break in consistency, but I value the discoverability higher. The current state leaves the arguably most important piece of information ("this imports an object mesh") unspoken. 

I would prefer to express the main function on the more common button and let the user discover the secondary option via tooltips/dialogue details. E.g. one button "Import Mesh" with tooltip "Import mesh from file" and one button "From URL" with tooltip "Import mesh from URL".

I am making the silent assumption here that importing via URL is uncommon, but I think it's a reasonable one.

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
rhaschke: I agree that we should distinguish scene and object loading/saving if needed. But can we actually import a whole scene?
If so, the corresponding symbols should be renamed to *_scene_* as well?
However, if there is only object loading, why should we rename the symbols?

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
rhaschke: The whole GUI block is dealing with objects only as indicated by the frame header "Curren Scene Objects". As said above, changing the button text from file to object obscures the fact, that we will load from a file in contrast to from url.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
davetcoleman: can you make 10 a named constant such as LARGE_MESH_THRESHOLD?

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -448,6 +445,42 @@ void MotionPlanningFrame::importResource(const std::string& path)
...
+            (abs(mesh->vertices[i * 3 + 2]) > 10))
+        {
+          object_is_very_large = true;
+          break;
+        }
+      }
+      if (object_is_very_large)
+      {
+        QMessageBox msg_box;
+        msg_box.setText("The object is very large. The file may be in millimeters instead of meters.");
```
davetcoleman, line 462:
```suggestion
        msg_box.setText("The object is very large (greater than 10m). The file may be in millimeters instead of meters.");
```

felixvd: Done

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: Done

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
felixvd: In that case is there a difference to not touching this at all?

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
felixvd: From my experience that's not enough to indicate what the button does. It could also mean "Import Scene Objects", for example. The word "Mesh" is more important than "File" in my opinion, and would make the button a lot more expressive.

Another argument: This `Import File` button in the `Current Scene Objects` frame does not actually import a "Scene Object", it imports a *mesh*, and the button that actually imports a "Scene Object" (or multiple ones) is `Import From Text` in the `Scene Geometry` tab.

I just committed an alternative, will post the pictures below.

felixvd: Suggested change:
vnew

Original below:
vold

I might want to take out the "Scene" from "Scene Objects" and add a tooltip to "Publish Scene".

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
felixvd: I think it's sufficiently messy to clear up. There are:  
- "import_file_button" which connects to `importFileButtonClicked` and imports a mesh
- "import_url_button" which connects to `importURLButtonClicked` and imports a mesh
- "import_scene_text_button" which connects to `importFromTextButtonClicked" and imports a .scene file (containing scene geometry (shapes), but no other info)
- "load_scene_button" which connects to `loadSceneButtonClicked` which loads a scene from `moveit_warehouse` (I don't know if this actually works)
- "load_state_button" which connects to `loadStateButtonClicked` and loads robot states (not necessarily clear to first-time codebase readers that this is only the robot state, not the state of the scene)
- "load_query_button" which loads planning queries, and a few corresponding "save" buttons.

The symbols are all in the same big `motion_planning_frame` file, so I think they should be clearer.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: I'll resolve this for now

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -1216,9 +1225,12 @@
            </widget>
           </item>
           <item row="1" column="0">
-           <widget class="QPushButton" name="import_file_button">
+           <widget class="QPushButton" name="import_object_file_button">
             <property name="text">
-             <string>Import &amp;File</string>
+             <string>Import &amp;Mesh</string>
```
rhaschke, line 1230: I still think that file would be better here.

felixvd:
I added the file filter and fixed the scaling problem, which occurred because mesh scaling keeps the mesh's centroid constant, but the mesh centroid was not at the center of the interactive marker. It's fixable as discussed [here](https://github.com/ros-planning/geometric_shapes/issues/148#issuecomment-650801408), but considering that scaling is unlikely to be a very common operation, and there are more things to consider when scaling multi-shape objects, I would probably just leave it as is. For now, meshes imported via the Rviz plugin are shifted so that the center point is at the origin.

felixvd:
Is Travis failing because CHOMP timed out? I tried building with the warnings enabled and ran clang-format and clang-tidy.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
felixvd: Done

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
felixvd: As mentioned below, I renamed some more of these.

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
```diff
@@ -186,8 +186,8 @@ private Q_SLOTS:
   void collisionObjectChanged(QListWidgetItem* item);
   void imProcessFeedback(visualization_msgs::InteractiveMarkerFeedback& feedback);
   void copySelectedCollisionObject();
-  void exportAsTextButtonClicked();
-  void importFromTextButtonClicked();
+  void exportGeometryAsTextButtonClicked();
```
rhaschke, line 189: [export|import]**Scene**[As|From]Text...?

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -128,8 +128,8 @@ MotionPlanningFrame::MotionPlanningFrame(MotionPlanningDisplay* pdisplay, rviz::
   connect(ui_->planning_scene_tree, SIGNAL(itemChanged(QTreeWidgetItem*, int)), this,
           SLOT(warehouseItemNameChanged(QTreeWidgetItem*, int)));
   connect(ui_->reset_db_button, SIGNAL(clicked()), this, SLOT(resetDbButtonClicked()));
-  connect(ui_->export_scene_text_button, SIGNAL(clicked()), this, SLOT(exportAsTextButtonClicked()));
-  connect(ui_->import_scene_text_button, SIGNAL(clicked()), this, SLOT(importFromTextButtonClicked()));
+  connect(ui_->export_scene_geometry_text_button, SIGNAL(clicked()), this, SLOT(exportGeometryAsTextButtonClicked()));
```
rhaschke, line 131: export_scene_as_text_button

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -128,8 +128,8 @@ MotionPlanningFrame::MotionPlanningFrame(MotionPlanningDisplay* pdisplay, rviz::
...
+  connect(ui_->import_scene_geometry_text_button, SIGNAL(clicked()), this, SLOT(importGeometryFromTextButtonClicked()));
```
rhaschke, line 132: import_scene_from_text_button

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -445,6 +445,26 @@ void MotionPlanningFrame::importResource(const std::string& path)
         return;
       }
 
+      // Center the mesh so scaling applies correctly
```
rhaschke, line 448: Centering is not required if no scaling will be applied, is it?
You can merge this code with the scaling code instead.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -1169,19 +1169,19 @@
...
              <string>Export scene geometry to a .scene file (only geometry information is preserved)</string>
             </property>
            </widget>
           </item>
           <item>
-           <widget class="QPushButton" name="import_scene_text_button">
+           <widget class="QPushButton" name="import_scene_geometry_text_button">
             <property name="text">
-             <string>&amp;Import Scene Objects</string>
+             <string>&amp;Import Objects</string>
```
rhaschke, line 1184: Here you even reduced the button text? I think, _scene_ should be mentioned.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -1422,6 +1422,9 @@
               <property name="text">
                <string>Filter</string>
```
rhaschke, line 1423: Is _Filter_ the right text description for the button?

felixvd:
The build just passed, looks like it failed on a CHOMP test before.

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
felixvd: The function it calls in PlanningScene is also called [saveGeometryToStream](https://github.com/ros-planning/moveit/blob/master/moveit_core/planning_scene/src/planning_scene.cpp#L952) and only saves the shapes in the world, ignoring subframes, object poses and even `AttachedBody` objects. I assume it is meant to load obstacle configurations. I added a tooltip to make this clear.

The existing `load_scene_button` seems to preserve the metadata, although I haven't tested it.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
rhaschke: If you rely on the frame title "Scene Geometry", you can shorten the button text to "Import".

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
felixvd: I can't see how it would fit. I changed it.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: See the other comment re. the name change

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: See the other comment re. the name change

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
felixvd: True, why not. I'll do that.

felixvd:
I fixed the comments, the misnamed button "Filter" and changed "State" to "Robot State" in the frame titles since there was space and it is not self-evident that the tab is about *robot* states.

Maybe we should stop before we iterate through the whole plugin. My thoughts on changes that could be a good WMD or newcomer issue:
- The `Stored Scenes` tab should either work from local memory, like `Stored States`, or display a line "Set up a database to use this tab"
- I think the `Planning` tab should be the default one, not `Context`.
- Some functions in `motion_planning_frame_objects` should be moved to `motion_planning_frame_scenes`

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_states.cpp
```diff
@@ -245,8 +245,21 @@ void MotionPlanningFrame::removeStateButtonClicked()
 
 void MotionPlanningFrame::clearStatesButtonClicked()
 {
-  robot_states_.clear();
-  populateRobotStatesList();
+  QMessageBox msg_box;
```
rhaschke, line 248: Why introducing a security message box, if the data is only removed from memory?

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -448,6 +445,64 @@ void MotionPlanningFrame::importResource(const std::string& path)
...
+      }
+      if (object_is_very_large)
+      {
+        QMessageBox msg_box;
+        msg_box.setText(
+            "The object is very large (greater than 10 m). The file may be in millimeters instead of meters.");
+        msg_box.setInformativeText("Attempt to fix the size by shrinking the object?");
+        msg_box.setStandardButtons(QMessageBox::Yes | QMessageBox::No | QMessageBox::Cancel);
+        msg_box.setDefaultButton(QMessageBox::Yes);
+        int ret = msg_box.exec();
```
rhaschke, line 468:
```suggestion
        if (msg_box.exec() == QMessageBox::Yes)
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -448,6 +445,64 @@ void MotionPlanningFrame::importResource(const std::string& path)
...
+              sx /= (double)mesh->vertex_count;
+              sy /= (double)mesh->vertex_count;
+              sz /= (double)mesh->vertex_count;
+              for (unsigned int i = 0; i < mesh->vertex_count; ++i)
+              {
+                unsigned int i3 = i * 3;
+                mesh->vertices[i3] -= sx;
+                mesh->vertices[i3 + 1] -= sy;
+                mesh->vertices[i3 + 2] -= sz;
+              }
```
rhaschke, line 491: This seems overly complicated: finding the centroid, shifting, scaleAndPadd finding the centroid again, ...
What about just scaling w.r.t zero, i.e. just keeping this code:
```suggestion
              for (unsigned int i = 0; i < mesh->vertex_count; ++i)
              {
                unsigned int i3 = i * 3;
                mesh->vertices[i3] *= 0.001;
                mesh->vertices[i3 + 1] *= 0.001;
                mesh->vertices[i3 + 2] *= 0.001;
              }
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_states.cpp
felixvd: Confirmation dialogues in front of destructive operations are just prudent, in my opinion. You never know how many states the user has stored, or if they misunderstood the button. Consider non-native speakers, too.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: Yes. That's actually what I expected `scaleAndPadd` to do to start with. Might as well keep it simple.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: Done

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_states.cpp
rhaschke: For me, confirmation dialogues are just annoying :wink:

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -463,39 +463,18 @@ void MotionPlanningFrame::importResource(const std::string& path)
...
-                mesh->vertices[i3 + 2] -= sz;
-              }
-              mesh->scaleAndPadd(0.001, 0);
-            }
-            break;
-          case QMessageBox::No:
-            break;
-          default:
-            return;  // Pressed cancel, do nothing
+          //
```
rhaschke, line 470: Please cleanup this comment.

felixvd: Fixed and rebased

felixvd:
And I'm in favor of "Import Mesh" and happy to see it merged 😎

v4hn:
Thanks to both of you for all these iterations

PR #2144, opened 2020-06-07
[WIP] github actions ci test

tylerjw:
### Description

This is just a trial of github actions ci and industrial_ci.  Please mostly ignore.

Something probably has to be enabled in this org if we are to try them, but they are on by default on my personal account... https://github.com/tylerjw/moveit/pull/1


tylerjw:
That was easier than I thought it would be.  There is some issue around tf_conversions but otherwise it looks like github actions with industrial_ci is working.  If we were to use this we'd still need to figure out code coverage and clang-tidy and evaluate if github actions provide any benefit over travis.

This was mostly just to see if I could do this.  Here is the link to the build: https://github.com/tylerjw/moveit/pull/1/checks?check_run_id=746378872

PR #2148, opened 2020-06-10
[WIP] convert .travis.yml to industrial_ci

tylerjw:
### Description

This is an attempt to use industrial_ci for everything we are now using moveit_ci.  I think most, if not all of the features exist in industrial_ci that we need along with some nice features that don't exist in moveit_ci.

### Checklist


gavanderhoorn:
@rhaschke has worked quite hard on `moveit_ci` (in addition to a bunch of other ppl) and it would be unfortunate to see that all "replaced" by `industrial_ci`.

There is https://github.com/ros-industrial/industrial_ci/pull/452 which discusses/proposes to merge the two projects, which perhaps should be looked at before migrating `ros-planning/moveit` to `industrial_ci`.

An older issue would be https://github.com/ros-industrial/industrial_ci/issues/98.

Personally I especially like the way the logs are captured/presented by `moveit_ci` and a few other features.

@tylerjw: would it make sense for you to take a look and see what the status is of that issue and the PR?

rhaschke:
Thanks @gavanderhoorn for pointing out https://github.com/ros-industrial/industrial_ci/pull/452.
I would very much appreciate it if we could merge moveit_ci and industrial_ci instead of just migrating.

tylerjw:
I did see that PR.  I'm not sure how a merge would work as the code is so different.  This was more of me just trying to see if I could get it to work easily.

One nice feature is that however they wrote their clang-format check it catches that there are two .cpp files in moveit that are marked executable.  I'm not sure how that happened but I'll make a small PR to fix that.

My favorite feature of industrial ci ( and why I've been using it allot ) is the caching of the docker images for local runs which makes so running a full CI run locally on fairly complex projects is often ~8min.

Also, an interesting new feature is the support for github actions and the seamless support for ros2 and gitlab.

rhaschke:
Of course both projects cannot me "merged" automatically. It's more manually migrating certain features from moveit_ci to industrial_ci. Caching of docker images is a feature of docker itself. Should work for moveit_ci as well, doesn't it?

tylerjw:
The feature I am talking about is the `rerun_ci` script that commits a docker image after the run (successful or not) with a hash of the command you ran (including arguments) and then if you run that same command it reuses that docker image.  It makes so that when you are locally testing you don't have to manage docker yourself to get that benefit.

PR #2151, opened 2020-06-17
[jog_arm] Changes before porting to ROS2

AdamPettinger:
### Description

These are changes I wanted to make (in addition to [https://github.com/ros-planning/moveit/pull/2103](https://github.com/ros-planning/moveit/pull/2103)) before starting on porting the jogger to ROS2 because I want these changes there as well. 

The changes consist of a lot of minor changes, see list below for details. Note that discussion and reviews exist here already: [https://github.com/PickNikRobotics/moveit/pull/8](https://github.com/PickNikRobotics/moveit/pull/8), because I wanted to start the review process before #2103 landed.

- General cleanup: defining `const` in functions where applicable, fixing comments, combining if statements and making them flow better
- I found a stale but nonzero command could make it through the timer callback function without updating the joint position filters, so fixed that and made **sure** the filters are updated every cycle
- Removed the `suddenHalt(Eigen::ArrayXd& delta_theta)` overload. It included comments saying position and velocity controlled robots were handled differently, but this function did not handle them differently (and only set `delta_theta` to all 0 which could be **really bad** for a position controlled robot). It was only called one place with the Eigen Array, and I replaced the call with a `setZero()` there and deleted the function so future users don't use it and accidentally send the robot to 0's. 
- Small fixes in the drift dimensions use and in singularity scaling. Previously the singularity scaling wasn't doing what the algorithm in the comments said it was. Tested after fixing and it looks really good
- Small changes around the `have_nonzero_twist_stamped_` and `have_nonzero_joint_jog_` which were used intermittently in the `run()` timer callback but also set in the command subscriber callbacks. Moved them to mutex protected implementation similar to the rest of subscriber callback variables
- Avoid recalculating the planning -> command transform if the incoming command is empty or in the command frame

### Checklist

### DIFF
--- a/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
+++ b/moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
@@ -629,10 +629,6 @@ void JogCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
   Eigen::ArrayXd velocity = delta_theta / parameters_.publish_period;
   const Eigen::ArrayXd acceleration = (velocity - prev_joint_velocity_) / parameters_.publish_period;
 
-  // We find the joint vel/accel limit that most effects the requested delta_theta
-  // And change requested all of delta_theta by that scaling
-  double most_extreme_relative_change = 1;
-
   std::size_t joint_delta_index = 0;
   for (auto joint : joint_model_group_->getActiveJointModels())
   {
@@ -653,8 +649,8 @@ void JogCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
         acceleration_limit = bounds.max_acceleration_;
       }
 
-      // Find relative change
-      if (clip_acceleration && delta_theta(joint_delta_index) != 0)
+      // Apply acceleration bounds
+      if (clip_acceleration)
       {
         // accel = (vel - vel_prev) / delta_t = ((delta_theta / delta_t) - vel_prev) / delta_t
         // --> delta_theta = (accel * delta_t _ + vel_prev) * delta_t
@@ -662,11 +658,9 @@ void JogCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
             ((acceleration_limit * parameters_.publish_period + prev_joint_velocity_(joint_delta_index)) *
              parameters_.publish_period) /
             delta_theta(joint_delta_index);
-
-        if (relative_change < most_extreme_relative_change)
-        {
-          most_extreme_relative_change = relative_change;
-        }
+        // Avoid nan
+        if (fabs(relative_change) < 1)
+          delta_theta(joint_delta_index) = relative_change * delta_theta(joint_delta_index);
       }
     }
 
@@ -688,21 +682,20 @@ void JogCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
       }
 
       // Apply velocity bounds
-      if (clip_velocity && delta_theta(joint_delta_index) != 0)
+      if (clip_velocity)
       {
         // delta_theta = joint_velocity * delta_t
         const double relative_change = (velocity_limit * parameters_.publish_period) / delta_theta(joint_delta_index);
-        if (relative_change < most_extreme_relative_change)
+        // Avoid nan
+        if (fabs(relative_change) < 1)
         {
-          most_extreme_relative_change = relative_change;
+          delta_theta(joint_delta_index) = relative_change * delta_theta(joint_delta_index);
+          velocity(joint_delta_index) = relative_change * velocity(joint_delta_index);
         }
       }
-      ++joint_delta_index;
     }
+    ++joint_delta_index;
   }
-
-  // Apply the limits
-  delta_theta *= most_extreme_relative_change;
 }
 
 bool JogCalcs::enforceSRDFPositionLimits()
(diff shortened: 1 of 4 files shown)


moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -232,21 +231,31 @@ void JogCalcs::run(const ros::TimerEvent& timer_event)
     if (have_nonzero_twist_stamped_)
     {
       if (!cartesianJogCalcs(twist_stamped_, *joint_trajectory))
+      {
+        resetLowPassFilters(original_joint_state_);
```
tylerjw, line 236: One of these calls to resetLowPassFilters is causing some really weird behavior.

AdamPettinger: I believe this is fixed with the newest commits

AdamPettinger:
Recent update is rebased on master to grab Tyler's #2155, and includes changes:

1. I switched to tracking the "staleness" of joint jog commands and Cartesian twist commands separately. We were getting into trouble when the command was non-stale because of a joint jog command, but the last (and stale) twist command was nonzero, resulting in doing Cartesian jogging calcs on the stale command and ignoring the joint jog (because we prioritize that)
2. One of the problems @tylerjw and @AndyZe were seeing was a bad output trajectory. Tracked it to when no jogging calcs were being performed, and decided to send the previous positions and 0 velocity in this case
3. Slight fix to adding redundant trajectory points if publishing to Gazebo
4. Fixed a stupid Eigen Vector initialization on my part that was causing wild and undesirable joint changes while joint jogging

AndyZe:
I haven't tested it yet but I made a PR against your branch to correct one variable type.

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -225,29 +254,44 @@ void JogCalcs::run(const ros::TimerEvent& timer_event)
...
+    {
+      resetLowPassFilters(original_joint_state_);
+      return;
+    }
+  }
+  else if (have_nonzero_joint_jog_ && !joint_command_is_stale_)
   {
-    // Prioritize cartesian jogging above joint jogging
-    if (have_nonzero_twist_stamped_)
+    if (!jointJogCalcs(joint_jog_, *joint_trajectory))
```
AndyZe, line 269: The name of `joint_jog_` should be more clear. Like, `joint_jog_cmd_`

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -350,14 +393,23 @@ bool JogCalcs::cartesianJogCalcs(geometry_msgs::TwistStamped& cmd, trajectory_ms
...
+      translation_vector = tf_moveit_to_robot_cmd_frame_.linear() * translation_vector;
+      angular_vector = tf_moveit_to_robot_cmd_frame_.linear() * angular_vector;
+    }
+    else
+    {
+      // We solve (planning_frame -> base -> cmd.header.frame_id)
+      // by computing (base->planning_frame)^-1 * (base->cmd.header.frame_id)
+      const auto tf_moveit_to_incoming_cmd_frame =
+          kinematic_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
+          kinematic_state_->getGlobalLinkTransform(cmd.header.frame_id);
```
AndyZe, line 408: These transforms were previously calculated around L231. It looks like `tf_moveit_to_robot_cmd_frame_` is not used around line 231, though. Can you move L231 here? Then delete this local variable `tf_moveit_to_incoming_cmd_frame`

AdamPettinger: These are 2 different transformations:

- `tf_moveit_to_robot_cmd_frame_` which is the moveit planning frame -> robot command frame (specified in yaml file) transformation. I believe it was calculated early (here in L231) so that it would be updated and available if the rest of the jogging process failed / ended early, and would always be available for the C++ API (via `getCommandFrameTransform()`) regardless of jogging status
- `tf_moveit_to_incoming_cmd_frame` is the moveit planning frame -> incoming twist command frame transformation and is calculated and used in the Cartesian jogging

We only use one of these to actually jog every cycle. If the twist command frame is empty, or it's in the same frame given as the "robot command frame" in the parameters, we use the already calculated `tf_moveit_to_robot_cmd_frame_`. If it is a different frame, we jog in that frame and calculate the (local) `tf_moveit_to_incoming_cmd_frame`

I believe this was the desired intent before this PR, and the changes here are to avoid recalculating `tf_moveit_to_robot_cmd_frame_` if the jogging command is in that frame. I can discuss more, but I think this is OK

AndyZe: ok, that sounds good. A comment at Line 230 would be good:

// Calculate this transform to ensure it's available via C++ API

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AdamPettinger: Agree, changed it to `joint_jog_cmd_`, and also changed the similar `twist_stamped_` to `twist_stamped_cmd_`, see latest commit

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
AdamPettinger: Sounds good, see new commit

PR #2154, opened 2020-06-18
[jog_arm] remove duplicate line

tylerjw:
### Description

Another small bug created by my refactor.  I don't think this did anything harmful but it also isn't right.


tylerjw:
ty @gavanderhoorn for the review on this

PR #2155, opened 2020-06-19
[jog_arm] fix access past end of array bug

tylerjw:
### Description

Manually tracing the code I'm not sure how this is happening but it is and this causes odd behavior (sending trajectory data from seemingly random data).


AndyZe:
IMO there's no reason not to add a check like this  :+1:

moveit_experimental/moveit_jog_arm/src/jog_calcs.cpp
```diff
@@ -298,9 +298,9 @@ void JogCalcs::run(const ros::TimerEvent& timer_event)
     else if (parameters_.command_out_type == "std_msgs/Float64MultiArray")
     {
       auto joints = moveit::util::make_shared_from_pool<std_msgs::Float64MultiArray>();
-      if (parameters_.publish_joint_positions)
+      if (parameters_.publish_joint_positions && joint_trajectory->points.size() > 0)
```
henningkayser, line 301: Use `!empty()`?

tylerjw: good catch, done

Issue #694, opened 2017-11-21
[kinetic] Invalid Trajectory: start point deviates from current robot state more than 0.01


Vonstafenberg:
Hi,

I am using Moveit with ROS kinetic for planning the motions for a UR5 arm. I am using the Track_IK kinematic solver. The problem is , I am speicifying simple goal poses for the gripper in cartesian space, but I am getting the error message in the terminal of moveit_planning execution.launch.

> **[ERROR] [1511284205.660285187]: 
Invalid Trajectory: start point deviates from current robot state more than 0.01
joint 'arm_elbow_joint': expected: 0, current: 1.84494
[ INFO] [1511284205.660356791]: Execution completed: ABORTED
[ INFO] [1511284205.700050212]: Execution request received for ExecuteTrajectory action.
[ WARN] [1511284205.708508088]: Controller  failed with error code INVALID_GOAL
[ WARN] [1511284205.708629940]: Controller handle  reports status FAILED
[ INFO] [1511284205.708804955]: Execution completed: FAILED
[ INFO] [1511284205.739949934]: Execution request received for ExecuteTrajectory action.**

I looked into the issue mentioned [here](https://github.com/ros-industrial-consortium/descartes/issues/196). But I ma using trac_Ik instead of Descartes. So I am not sure how to implement a proper solution.

I also referenced [this](https://groups.google.com/forum/#!topic/moveit-users/x5K6lVuiGqM), but then changing the

>  **/move_group/trajectory_execution/allowed_start_tolerance** to 0.0 is giving me another error.

> **[ WARN] [1511285456.138267775]: Controller  failed with error code INVALID_GOAL
[ WARN] [1511285456.138359326]: Controller handle  reports status FAILED
[ INFO] [1511285456.138473683]: Execution completed: FAILED
[ INFO] [1511285456.168054862]: Execution request received for ExecuteTrajectory action.**

in the terminal where I launch the UR_driver to interface with the real robot I have the error

> **[ERROR] [1511285455.959112929]: Goal start doesn't match current pose
[ INFO] [1511285455.995846536]: on_goal
[ERROR] [1511285455.996445868]: Goal start doesn't match current pose
[ INFO] [1511285456.036065982]: on_goal**

and in the terminal where I launch the node I get 

> **[ INFO] [1511285455.930223750]: ABORTED: Solution found but controller failed during execution
[ INFO] [1511285455.968737779]: ABORTED: Solution found but controller failed during execution
[...]

PR #2157, opened 2020-06-21
TEM: fix check for start state position

v4hn:
distance() respects unlimited revolute joints

Addresses an issue pointed out in #694 .


v4hn:
On Wed, Jun 24, 2020 at 09:35:35AM -0700, Mark Moll wrote:
> Looks good to me. I am kind of surprised this has gone unnoticed for so long.

Well, it *is* a damn unlikely edge case to have a
- continuous joint
- start planning from its range edge
- have noise move the joint to the other side in the meanwhile

Thanks for merging.

We should look through backports again soon and merge this also back to `kinetic-devel`, as requested by the reporting user.

PR #2162, opened 2020-06-24
[melodic-devel] moveit_jog_arm (moveit_servo) backport

tylerjw:
### Description

Here is a backport of the latest changes on master into melodic-devel.

Bugfix backports:

* Separate moveit_experimental packages (#1606) 
*  [jog_arm] simplify communication between threads (#2103)
*  [jog_arm] time-based collision avoidance (#2100) 
* Added throttle to jogarm accel limit warning (#2141)
* pragma once in jog_arm.h (#2152) 
* [jog_arm] fix access past end of array bug (#2155)
* [jog_arm] Changes before porting to ROS2 (#2151) 
* [jog_arm] Move and rename to moveit_ros/moveit_servo (#2165)
* Minor moveit_servo header cleanup (#2173)

PR #2165, opened 2020-06-25
[jog_arm] Move and rename to moveit_ros/moveit_servo

AdamPettinger:
### Description

This moves the `moveit_experimental/jog_arm` package and renames it `moveit_ros/moveit_servo` @AndyZe. There is already an open PR for ROS2 ([https://github.com/ros-planning/moveit2/pull/223](https://github.com/ros-planning/moveit2/pull/223)) with some discussion.

@tylerjw will help with migration notes through the name change as well as his big recent PR #2103. 

### Checklist


AndyZe:
I tested it and did a quick verification that "jog" and "Jog" don't appear anywhere (except message types, which we can't change here). Looks good! Approved if you remove that last commit, like Henning pointed out.

AdamPettinger:
>  b48460b shouldn't be included here, though

Just to clarify @henningkayser, you want me to leave the COLCON_IGNORE file here in the ROS1 package?

henningkayser:
> > [b48460b](https://github.com/ros-planning/moveit/commit/b48460b1418d8a7c73257d0afd445b1a014720a4) shouldn't be included here, though
> 
> Just to clarify @henningkayser, you want me to leave the COLCON_IGNORE file here in the ROS1 package?

No, the commits shouldn't include COLCON_IGNORE at all. This repo doesn't include this file so there should be no need to remove it.

AndyZe:
I think once these commits all get squashed it won't matter, then...

AdamPettinger:
> > > [b48460b](https://github.com/ros-planning/moveit/commit/b48460b1418d8a7c73257d0afd445b1a014720a4) shouldn't be included here, though
> > 
> > 
> > Just to clarify @henningkayser, you want me to leave the COLCON_IGNORE file here in the ROS1 package?
> 
> No, the commits shouldn't include COLCON_IGNORE at all. This repo doesn't include this file so there should be no need to remove it.

One of my earlier commits added it (first done in moveit2), and I removed it at the end. I think Andy is right and squashing these commits together would result in it never being here. If we are going to squash before merging anyway, I think leaving it is ok. Otherwise I can fix the commit that added it, and remove the final commit

AdamPettinger:
I removed the COLCON_IGNORE from these commits

henningkayser:
> > > > [b48460b](https://github.com/ros-planning/moveit/commit/b48460b1418d8a7c73257d0afd445b1a014720a4) shouldn't be included here, though
> > > 
> > > 
> > > Just to clarify @henningkayser, you want me to leave the COLCON_IGNORE file here in the ROS1 package?
> > 
> > 
> > No, the commits shouldn't include COLCON_IGNORE at all. This repo doesn't include this file so there should be no need to remove it.
> 
> One of my earlier commits added it (first done in moveit2), and I removed it at the end. I think Andy is right and squashing these commits together would result in it never being here. If we are going to squash before merging anyway, I think leaving it is ok. Otherwise I can fix the commit that added it, and remove the final commit

The commit history is a good documentation here, I would also keep it probably.

rhaschke:
IMHO we should try to avoid merging PRs like this as a merge-commit and instead squash-merge them. For example, this one has many closely-related commits and some cleanup commits:
￼
* Move moveit_experimental/moveit_jog_arm -> moveit_ros/moveit_jog_arm
* Rename filepath moveit_ros/moveit_jog_arm -> moveit_ros/moveit_servo

...
* Rename internal namespaces and package/project name to moveit_servo
* Internal rename interface class JogArm to Servo
* Internal rename JOG_ARM_STATUS_CODE_MAP to SERVO_STATUS_CODE_MAP
* Internal rename jog_server to servo_server
* Rename a bunch of internals
* Internal rename JogCalcs -> ServoCalcs
* Rename servo_calcs internals

...
* Clang formatting

@tylerjw: As I have seen several such commits from your side recently (e.g. also #2103), please carefully balance a clean vs. detailed history in future. Sometimes, of course, it is important and useful to keep individual commits of a PR and perform a proper merge. In these cases, (request to) cleanup the PR commit history before merging. Examples are unrelated commits (e.g. backports) in a single PR. However, as a rule of thumb, our guidelines state that a single PR should only tackle a single "problem" and thus all related commits should be squash-merged by default (see [GitHub Merge Policies](https://moveit.ros.org//documentation/contributing/pullrequests/)).

felixvd:
Thanks for mentioning it, I was close to bringing it up, too. #2162 in particular is very long (54 commits, 15 for renaming, two identical commits one after the other). I would actually support force-pushing a clean version to melodic-devel in this case because it's so large.

PR #2166, opened 2020-06-25
optional cpp version setting

tylerjw:
### Description

This would set the default C++ version to 14 but would enable someone to use 17 if they wanted via the cmake argument `-DCMAKE_CXX_STANDARD=17`.  I'm doing this because I'm debugging an eigen memory issue that I think might be solved if I could build moveit with 17.  This would not harm anyone who wants to continue to use 14.


simonschmeisser:
I'm curious, what eigen-issue are we talking about and how does the C++ standard chosen influence that?

gavanderhoorn:
@simonschmeisser: most likely memory alignment issues with Eigen and the support in C++17 for automatically correctly aligning classes which need it / express the requirement.

PR #2180, opened 2020-07-01
Fixes for upcoming Noetic release

rhaschke:
This PR attempts to fix some new warnings and other issues when building `master` on Noetic. Travis utilizes the new Noetic docker images (#2178).

- [x] I silented new gcc warnings (outside the MoveIt code base) by declaring all external includes as `SYSTEM` includes.
- [x] I fixed all remaining warnings inside our code base
- [x] Travis succeeds
- [x] `mesh_filter` tests were failing again. Looks like the newer version of Mesa OpenGL lost precision (again).
       Fixed by increasing the threshold.


moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -163,7 +163,7 @@ void RobotState::copyFrom(const RobotState& other)
             (robot_model_->getVariableCount() * (1 + ((has_velocity_ || has_acceleration_ || has_effort_) ? 1 : 0) +
                                                  ((has_acceleration_ || has_effort_) ? 1 : 0)) +
              nr_doubles_for_dirty_joint_transforms);
-    memcpy(variable_joint_transforms_, other.variable_joint_transforms_, bytes);
+    memcpy((char*)variable_joint_transforms_, (char*)other.variable_joint_transforms_, bytes);
```
henningkayser, line 166: Is this really a fix? memcpy() can break alignment, no? I also ran into this and would have thought that we should move to stl types for these or use a [Map](https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html).

moveit_ros/perception/CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ endif()
 find_package(Boost REQUIRED thread)
 
 if(WITH_OPENGL)
+  # Prefer newer vender-specific OpenGL library
```
henningkayser, line 19:
```suggestion
  # Prefer newer vendor-specific OpenGL library
```

moveit_core/robot_state/src/robot_state.cpp
rhaschke: This was tuned (though several years ago) to optimize CPU cache hit rates. To this end, it does make sense to keep all the robot state in a single memory block. I don't see how Eigen::Map would help here. It's fine to utilize this memory block directly for Eigen::Isometry. The (new) gcc warning I fixed here is related to the fact that we memcpy the content of Eigen::Isometry, thus by-passing the custom-defined copy-constructor of Eigen::Isometry.
As this copy-constructor just performs a simple check, we are safe to perform the memcpy. Casting to (char*), we pretend (for gcc) that this isn't a Eigen::Isometry array :wink:

moveit_ros/perception/mesh_filter/test/mesh_filter_test.cpp
```diff
@@ -231,7 +231,7 @@ void MeshFilterTest<Type>::test()
     float sensor_depth = sensor_data_[idx] * FilterTraits<Type>::ToMetricScale;
     if (fabs(sensor_depth - distance_ - shadow_) > epsilon_ && fabs(sensor_depth - distance_) > epsilon_)
     {
-      ASSERT_NEAR(filtered_depth[idx], gt_depth[idx], 1e-6);
+      ASSERT_NEAR(filtered_depth[idx], gt_depth[idx], 1e-4);
```
v4hn, line 234: The error you gave [in this post](https://github.com/ros-planning/moveit/issues/2066#issuecomment-653302801)
had 0 values in one of the entries. How can this be solved by reducing the accuracy by factor 1e2?

moveit_core/collision_detection/src/collision_matrix.cpp
```diff
@@ -73,14 +73,6 @@ AllowedCollisionMatrix::AllowedCollisionMatrix(const moveit_msgs::AllowedCollisi
...
 }
 
-AllowedCollisionMatrix::AllowedCollisionMatrix(const AllowedCollisionMatrix& acm)
-{
-  entries_ = acm.entries_;
-  allowed_contacts_ = acm.allowed_contacts_;
-  default_entries_ = acm.default_entries_;
-  default_allowed_contacts_ = acm.default_allowed_contacts_;
-}
-
```
v4hn, line 83: nice!

moveit_core/robot_state/src/robot_state.cpp
v4hn:
```suggestion
    memcpy((void*)variable_joint_transforms_, (void*)other.variable_joint_transforms_, bytes);
```
Officially `memcpy` receives `void*` arguments, so there should be no need for the `char` hack.

moveit_core/CMakeLists.txt
```diff
@@ -175,16 +175,16 @@ catkin_package(
 include_directories(SYSTEM ${EIGEN3_INCLUDE_DIRS}
                            ${LIBFCL_INCLUDE_DIRS}
                            ${BULLET_INCLUDE_DIRS}
-                           )
+                           ${Boost_INCLUDE_DIRS}
+                           ${catkin_INCLUDE_DIRS}
+                           ${urdfdom_INCLUDE_DIRS}
+                           ${urdfdom_headers_INCLUDE_DIRS}
+                           ${OCTOMAP_INCLUDE_DIRS}
+                    )
```
v4hn, line 183: Including `catkin_INCLUDE_DIRS` as SYSTEM feels wrong. This also includes packages in our domain.
Why did you need to do that?

If you insist on making all of them SYSTEM includes, please sort them the way they originally where (previous SYSTEM includes should be appended at the end of the list instead of the beginning).

moveit_core/collision_detection_bullet/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ install(FILES ../collision_detector_bullet_description.xml DESTINATION ${CATKIN_
 
 if(CATKIN_ENABLE_TESTING)
   find_package(moveit_resources REQUIRED)
-  include_directories(${moveit_resources_INCLUDE_DIRS})
+  include_directories(SYSTEM ${moveit_resources_INCLUDE_DIRS})
```
v4hn, line 32: This *is* our repository. We really shouldn't treat these as system includes.

Same in a number of other places.

moveit_ros/perception/mesh_filter/test/mesh_filter_test.cpp
rhaschke: These are unrelated issues. The post you are referring to is related to Melodic. This one is related to Noetic, which utilizes another version of Mesa OpenGL.

moveit_core/CMakeLists.txt
rhaschke: The new gcc compiler introduces a whole bunch of new warnings. Many of them are related to external includes, which we cannot fix. Hence, in order to silent them, we need to declare those includes as external.
My changes keep the includes of each package as local includes.

Where did you notice a reordering of include folders? I thought I took care of placing `catkin_INCLUDE_DIRS` first.

moveit_core/collision_detection_bullet/CMakeLists.txt
rhaschke: I agree for those.

moveit_core/CMakeLists.txt
rhaschke: Looks like, I considered the order only later in my editing process. Will fix the remaining as well.

rhaschke: Fixed.

rhaschke:
Fixes #2212.

moveit_ros/benchmarks/src/BenchmarkExecutor.cpp
```diff
@@ -40,7 +40,10 @@
 #include <tf2_eigen/tf2_eigen.h>
 
 #include <boost/regex.hpp>
+// TODO: Remove if boost > 1.72 (https://github.com/boostorg/timer/issues/12)
```
v4hn, line 43:
```suggestion
// TODO: Remove if boost >= 1.72 (https://github.com/boostorg/timer/issues/12)
```
https://github.com/boostorg/timer/commit/c221a60102ed618a82116f84407de3f1713aa10f is part of 1.72

PR #2181, opened 2020-07-01
Update API to take in revolute and prismatic params for Cartesian Paths

pbeeson:
### Description

Updates the computeCartesianPath() interface of MoveGroup to allow the original jump_threshold but also revolute_jump_threhsold and prismatic_jump_threshold if desired.  This is turn uses the JumpThreshold class of @davetcoleman 
under the hood in the capability in order to first ensure any absolute thresholds are constrained then the "Factor" based relative threshold is still applied if its set. 

This is coded to ensure backwards compatibility with MoveIt APIs, so it will not break existing user code, but allows new code to leverage Dave's work on the absolute joint limiting.

### Checklist
- [ x ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -741,9 +741,8 @@ public:
   /** \brief Compute a Cartesian path that follows specified waypoints with a step size of at most \e eef_step meters
       between end effector configurations of consecutive points in the result \e trajectory. The reference frame for the
       waypoints is that specified by setPoseReferenceFrame(). \e jump_threshold defines a factor off the average
-     configuration
-      space distance between waypoints that is acceptable between any two waypoints (this is to prevent 'jumps' in IK
-      solutions).
+      configuration space distance between waypoints that is acceptable between any two waypoints (this is to prevent
+      'jumps' in IK solutions).
       Collisions are avoided if \e avoid_collisions is set to true. If collisions cannot be avoided, the function fails.
       Return a value that is between 0.0 and 1.0 indicating the fraction of the path achieved as described by the
       waypoints.
@@ -758,9 +757,8 @@ public:
   /** \brief Compute a Cartesian path that follows specified waypoints with a step size of at most \e eef_step meters
       between end effector configurations of consecutive points in the result \e trajectory. The reference frame for the
       waypoints is that specified by setPoseReferenceFrame(). \e jump_threshold defines a factor off the average
-     configuration
-      space distance between waypoints that is acceptable between any two waypoints (this is to prevent 'jumps' in IK
-      solutions).
+      configuration space distance between waypoints that is acceptable between any two waypoints (this is to prevent
+      'jumps' in IK solutions).
       Kinematic constraints for the path given by \e path_constraints will be met for every point along the trajectory,
       if they are not met, a partial solution will be returned.
       Constraints are checked (collision and kinematic) if \e avoid_collisions is set to true. If constraints cannot be
@@ -826,7 +824,7 @@ public:
       acceptable between any two waypoints.
       Collisions are avoided if \e avoid_collisions is set to true. If collisions cannot be avoided, the function fails.
       Return a value that is between 0.0 and 1.0 indicating the fraction of the path achieved as described by the
-     waypoints.
+      waypoints.
       Return -1.0 in case of error. */
   double computeCartesianPath(const std::vector<geometry_msgs::Pose>& waypoints, double eef_step, double jump_threshold,
                               double revolute_jump_threshold, double prismatic_jump_threshold,
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -740,30 +740,115 @@ public:
 
   /** \brief Compute a Cartesian path that follows specified waypoints with a step size of at most \e eef_step meters
       between end effector configurations of consecutive points in the result \e trajectory. The reference frame for the
-      waypoints is that specified by setPoseReferenceFrame(). No more than \e jump_threshold
-      is allowed as change in distance in the configuration space of the robot (this is to prevent 'jumps' in IK
-     solutions).
+      waypoints is that specified by setPoseReferenceFrame(). \e jump_threshold defines a factor off the average
+     configuration
+      space distance between waypoints that is acceptable between any two waypoints (this is to prevent 'jumps' in IK
+      solutions).
+      Collisions are avoided if \e avoid_collisions is set to true. If collisions cannot be avoided, the function fails.
+      Return a value that is between 0.0 and 1.0 indicating the fraction of the path achieved as described by the
+      waypoints.
+      Return -1.0 in case of error. */
+  double computeCartesianPath(const std::vector<geometry_msgs::Pose>& waypoints, double eef_step, double jump_threshold,
+                              moveit_msgs::RobotTrajectory& trajectory, bool avoid_collisions = true,
+                              moveit_msgs::MoveItErrorCodes* error_code = nullptr)
+  {
+    return computeCartesianPath(waypoints, eef_step, jump_threshold, 0, 0, trajectory, avoid_collisions, error_code);
+  }
+
+  /** \brief Compute a Cartesian path that follows specified waypoints with a step size of at most \e eef_step meters
+      between end effector configurations of consecutive points in the result \e trajectory. The reference frame for the
+      waypoints is that specified by setPoseReferenceFrame(). \e jump_threshold defines a factor off the average
+     configuration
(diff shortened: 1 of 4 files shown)


pbeeson:
Requires https://github.com/ros-planning/moveit_msgs/pull/83 to compile.

pbeeson:
Travis is going to fail since this does require a new field in a moveit_msg service to be complete.

pbeeson:
After submitting this, I realized that this still doesn’t “solve” the issue where you want to prevent “rollouts”.  The parameters defined are useful for constraining rollout but are still dependent on the cartesian sampling density (that is you generally don’t want to move more than PI between any two waypoints regardless of the density) but for really dense trajectories, the “rollout” could happen over multiple waypoints so in practice this needs to be something much lower than pi and is dependent on your density.

The correct solution is to walk the final trajectory found from IK on the cartesian lines between goal points and run FK at some user desired spacing (e.g. 0.5 degrees change on any joint) and then determine if the cartesian EE pose deviates from the cartesian line pose (or maybe just the cartesian pose of the two waypoints in the Trajectory you are interpolating between) by more than some user defined distance.

This is how I have coded this outside of Moveit in the past and it works great, is fast (all FK), and is easy for a user to understand (what joint space interval to sample at (radians) and how far the EE can be from the straight line between cartesian points to be considered a failure (in meters)).

I can see the API changes from the current master to do the solution mentioned above requiring 2 doubles and this would be similar to the API changes for this PR.

Thoughts?

v4hn:
As you've noticed by now, these function interfaces are quite messy. There was a long pull-request with changes by @mlautman with a lot of discussion with @rhaschke , but the underlying ideas did not change.

[The PILZ planner](https://github.com/ros-planning/moveit/pull/1893) improves the situation, also because it's an actual pipeline planner, but does not resolve it afaik (not looked into the details myself).

The biggest problem I see with your post-processing approach, which is partially shared by the current absolute joint threshold though, is that it can't differentiate easily between high deviations due to too high eef step size and actual jumps between disjoint manifolds in the IK solutions.

Another way to go about it might be to detect too high absolute jumps during generation and attempt IK for interpolated Cartesian targets between the waypoints in that case, trying to reduce the maximum joint space delta per step.

Or adaptively change the Cartesian step size based on this scheme and stop if current step size << original step size.
In that case you might as well throw away the notion of user-provided step size altogether, simplifying the interface.

Or you just ignore the interpolated IK approach altogether and work with the pseudo-Inverse Jacobian...

Or you rely on a local IK solver without global restarts that will simply not find non-continuous paths.

> I can see the API changes from the current master to do the solution mentioned above requiring 2 doubles and this would be similar to the API changes for this PR.

You probably still want to separate the step size for prismatic and revolute joints, so that's two numbers. You will also need two number for maximum position/orientation displacement of the eef pose.

Personally, I'm always somewhat lost with how to improve these methods without throwing them away altogether.
I believe no maintainer is entirely happy with the current functions and improvements are very welcome.

pbeeson:
> The biggest problem I see with your post-processing approach, which is
partially shared by the current absolute joint threshold though, is that it
can't differentiate easily between high deviations due to too high eef step
size and actual jumps between disjoint manifolds in the IK solutions.

I disagree with this assessment (at least how I interpret what you are
saying).  The approach I advocate for in the comments above actually
interpolates FK across joint space waypoints and determines when these in
fact do have jumps in the space by monitoring the EE location wrt the
original cartesian space path.  If there are large jumps in joint space
that When interpolating at a fine step size do not cause the EE to move off
the expected cartesian space “line” (not just x,y,z but also orientation)
then this is a perfectly acceptable motion, as the EE is doing exactly what
you want.  It doesn’t depend at all on the cartesian space step size but on
the minimum joint delta to use for interpolation and the acceptable
distance from the original cartesian space line.  This is independent from
how far spaced the original cartesian space interpolation was.  So the
parameters here are “how far off the line can the EE move assuming joint
space motion between waypoints and how densely should I be checking to
enforce that.

pbeeson:
In any case, I think this PR is fine to stand o. It’s own and any additional post processing for validity can maybe be a separate PR later, outside of whether this once is accepted or not.

v4hn:
I think you misunderstood my remark. I do understand your proposed approach and would use it myself over the approach we currently have in the repository.
I'm just not sure whether it's a sufficiently better solution for the whole framework if we change the computeCartesianPath interface yet again, because there is a lot of other possible variations that might still yield better results.
It might also be reasonable to provide alternative methods and have a whole zoo of Cartesian generator approaches that differ in details.
After all MoveIt does the same with generic planners. :-)

That being said, I meant the following: If you request a trajectory through the current interface the only thing you require is that it passes through all Cartesian steps, moving reasonably short pathes in joint space between them. Naturally in between these waypoints there will be a non-zero displacement from the ideal Cartesian path. The maximum eef displacement you want to check in your approach has to be big enough to account for that. So if your check fails you basically have to tell the user that either they hit the end of a continuous manifold (they can't reach it, and would have to "rollout" some joints, as you call it, to continue the path, but can't because it would violate the path) or they can reduce their max_step (or increase their allowed pose deviation) and try again. I'm not sure how much of a problem that is in practice though.

And then there is the underlying question why the user would even need to specify *two* step sizes for *one* path. A maximum joint-space distance should suffice with an adaptive eef step size, and an eef step size should suffice with a reliable way to detect invalid jumps. Having both simplifies the computation, but also adds more parameters that can be set to incompatible values... And we already have quite a few parameters in the `computeCartesianPath` methods already.

Either way, I would welcome a pull-request with your proposed extension. Even if we should end up merging it as an alternative approach rather than replacing the current method.

pbeeson:
Thanks for all the useful feedback.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -740,30 +740,113 @@ class MoveGroupInterface
...
+      waypoints.
+      Return -1.0 in case of error. */
+  double computeCartesianPath(const std::vector<geometry_msgs::Pose>& waypoints, double eef_step, double jump_threshold,
+                              moveit_msgs::RobotTrajectory& trajectory,
+                              const moveit_msgs::Constraints& path_constraints, bool avoid_collisions = true,
+                              moveit_msgs::MoveItErrorCodes* error_code = nullptr)
+  {
+    return computeCartesianPath(waypoints, eef_step, jump_threshold, 0, 0, trajectory, path_constraints,
+                                avoid_collisions, error_code);
+  }
```
rhaschke, line 776: I don't like this explosion into various variants of the function at all. What about, just providing a single variant accepting the `JumpThreshold` class? The old variants should be deprecated and removed later.

pbeeson:
I think keeping the old variants and marking them as deprecated and making
a new one with the JumpThreshold class is a fine idea.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -740,30 +740,113 @@ class MoveGroupInterface
...
   /** \brief Compute a Cartesian path that follows specified waypoints with a step size of at most \e eef_step meters
       between end effector configurations of consecutive points in the result \e trajectory. The reference frame for the
-      waypoints is that specified by setPoseReferenceFrame(). No more than \e jump_threshold
-      is allowed as change in distance in the configuration space of the robot (this is to prevent 'jumps' in IK
-     solutions).
+      waypoints is that specified by setPoseReferenceFrame(). No more than \e revolute_jump_threshold radians
+      is allowed on each joint as change in distance in the configuration space of the robot.  Similarly, no more than
+      \e prismatic_jump_threshold m between waypoints is allowed. This is to prevent 'jumps' in IK solutions.
+      \e jump_threshold defines a factor off the average configuration space distance between waypoints that is
+      acceptable between any two waypoints.
```
felixvd, line 840:
```suggestion
      waypoints is that specified by setPoseReferenceFrame(). \e revolute_jump_threshold [rad]
      and \e prismatic_jump_threshold [m] limit the absolute jump distance between waypoints for each joint. \e jump_threshold defines the maximum permissible averaged distance between waypoints in configuration space may not exceed. This is to prevent 'jumps' in IK solutions.
```
Is `jump_threshold` the maximum average distance in c-space, or a factor between two joint configurations? I found that difficult to parse. 

The block was overall unwieldy, so here's a suggestion. The old version works too, except for the typo "off".

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
pbeeson: > I don't like this explosion into various variants of the function at all. What about, just providing a single variant accepting the `JumpThreshold` class? The old variants should be deprecated and removed later.

I think the consensus was that this would be a better solution.  IF people don't mind, I think this PR should be closed and if/when I or somebody else gets around to fixing this in a cleaner way, a new PR can be opened.

pbeeson:
Closing PR as cleaner solution was posited in the comments, and I am no longer using/supporting this solution myself.

PR #2185, opened 2020-07-03
Parameterize input trajectory density of Time Optimal trajectory generation

pbeeson:
### Description

MoveIt wrapper to Time Optimal Trajectory Generation downsamples trajectory, but its hard coded to keep waypoints with at least 0.001 radians of motion on a joint.  For very long and/or dense input trajectories, the downsampled trajectory is still way too large for the iterative optimization algorithm to finish in reasonable time.  I was seeing 3000 point trajectory taking ~15 seconds.  This fix allows the user to set this downsampling density in minimum radians of motion between two waypoints to something else (presumably higher like 0.5 degrees [0.00875 radians]) in order to drastically improve the speed of this algorithm.

### Checklist
- [ x ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)

### DIFF
--- a/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
+++ b/moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
@@ -162,7 +162,8 @@ private:
 class TimeOptimalTrajectoryGeneration
 {
 public:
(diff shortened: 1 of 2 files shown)


v4hn:
Thanks @pbeeson . I was annoyed by that hardcoded value as well when I looked through this, but did not have a reason to make it configurable.

PR #2187, opened 2020-07-03
Fix a parameter switch in moveit_cpp loading

AndyZe:
A simple bug fix. Caught by Michael W, not me (not sure he has a Github account)


v4hn:
This definitely should be part of `melodic-devel` before the release.

PR #2189, opened 2020-07-03
melodic backports

v4hn:
In response to [@tylerjw 's request](https://github.com/ros-planning/moveit/issues/2066#issuecomment-652613378).

At least [one patch](7aa1699f84d38f2881156b0a7466d3d528228e5d) quite close to the current `reviewed-for-backport` tag (7aa1699f84d38f2881156b0a7466d3d528228e5d) was not backported yet, so I went through the open list and collected everything of interest.

Once this request is merged `reviewed-for-backport` can be forwarded at least to a15551f8df441865693dae5e81cf1f3e1679426d (to be updated if further commits are added).

This request can wait until *after* the release.


v4hn:
> Shouldn't we merge #2187 into #2033 to cleanup the history?

No, detecting cherry-picks depends on the individual commits. Of course there should be no need to do that for these commits once this request is merged, but I still strictly prefer to keep individual commits in the backports.

> These look good. I'm good with backporting them before or after this release. Someone with a stronger opinion should choose.

We agreed to have a soft freeze period for releases and none of these changes are urgent.
We can just extend this request with more changes as they are merged in master, so I don't see the need to merge it before the release. :shrug:

rhaschke:
@v4hn, in https://github.com/ros-planning/moveit/pull/2187#issuecomment-653611230 you explicitly stated that at least this fix should be merged before the release. I don't see a reason to not merge these fixes.

v4hn:
> in #2187 you explicitly stated that at least this fix should be merged before the release.

Yes, but it turned out that the change that *introduced* this bug was part of the same badge, so we could have left it out without problems.
Anyway, now its merged.

rhaschke:
Forwarded reviewed-for-backport to 73fa9ea640f507334052bab183163772f1d7b18d.

PR #2190, opened 2020-07-03
melodic backports [pre-release]

v4hn:
Here is the list of things I collected with #2189 that I consider important enough (or at least reasonable) to merge before the release.

No reason to keep memory leaks, a segfault and potential linking problems open issues in the release.

@tylerjw @rhaschke


tylerjw:
This is failing on kinetic for something python related (it seems).

rhaschke:
Looks like the test is flaky. I retriggered the Travis job and it succeeded.

PR #2184, opened 2020-07-02
Changelogs for 1.0.5 Melodic release

tylerjw:
### Description

Here are changelogs I've prepared for the Melodic 1.0.5 release.

PR #2191, opened 2020-07-05
Changelogs for 1.0.5 Melodic release

rhaschke:
This is a rebase of #2184 onto latest Melodic HEAD.
@tylerjw, your work on the changelogs was great. I just merged a few more, related commits and I added the changelogs for the recently merged backports.
I also try to avoid abbreviations in the changelogs as these will be hard for non-maintainers to read. Hence, I replaced them with their long names.
From my point of view, you can proceed with the release as soon as we have fixed https://github.com/ros-planning/moveit/issues/2066#issuecomment-653724081, i.e. replaced the implicit dependency on panda_moveit_config with moveit_resources. I don't have time to look into this today.


tylerjw:
Superseded by https://github.com/ros-planning/moveit/pull/2200

PR #2197, opened 2020-07-06
use moveit_resources in moveit_cpp_test

tylerjw:
### Description

This change is in master and was made here: 2de8471b8ffa8ae2be907c3d820b4144b87c2f4b

I wasn't able to get that commit to nicely apply no melodic-devel and only need the change to the one file so I created this.  This should be all that is needed to get melodic-devel to pass the pre-release tests.


rhaschke:
ROS build farm is back to stable: http://build.ros.org/job/Mdev__moveit__ubuntu_bionic_amd64/82/display/redirect :tada:

Issue #2061, opened 2020-05-07
Remember open tab in RViz panel


v4hn:
Since #1991 a lot of state in the Motion Planning display in rviz is saved.
The currently open tab of the display panel (Context, Planning, Scene Objects...) is not saved though.

Did you leave that out on purpose @rhaschke ?

I would prefer to end up in the Planning tab for most of my configurations, so I'm interested in saving this too.

rhaschke:
I kind of left that out on purpose, yes. I think we should just default the planning tab (instead of the first one).

v4hn:
> I kind of left that out on purpose, yes. I think we should just default the planning tab (instead of the first one).

I thought about that option too.

But that way, most users would probably never look at any of the other tabs and selecting different planning algorithms & parameterizations is in theory one of the main selling points of the framework.

Looking for opinions by other core contributors/maintainers here.

gavanderhoorn:
> But that way, most users would probably never look at any of the other tabs and selecting different planning algorithms & parameterizations is in theory one of the main selling points of the framework.

Personally I'm not sure there is a "best" option here: if users always see the first tab on startup, they may never look at the others. If we save MoveIt configuration pkgs with any tab but the first active, and this is persisted and restored, users may have trouble finding the tab described "everywhere" (ie: the first one).

felixvd:
I think the tabs are sufficiently discoverable without forcing an extra click on every startup, so I'm in favor of defaulting to the Planning tab.

There's a larger discussion to be had about the selling points of the framework, but I suspect that comparing different motion planning algorithms isn't at the forefront anymore necessarily. Prototyping, production and overall usability seem to be becoming more and more important.

rhaschke:
#2198 defines the planning tab as default.

PR #2198, opened 2020-07-06
Add primitive shapes widgets

rhaschke:
Rebase of #2137 required due to conflicting changes in #2142. Great work, @jrgnicho! I was missing this feature several times in the past. In a cleanup commit, I tried to unify the UI to insert primitive objects / meshes.

image


v4hn:
Scrolling through your "Unify" commit, this looks good and improves user experience further.
I'm a bit amused to see how you tend to fix your nagging requests for changes by throwing *more* changes after them, so that no one can argue against them anymore :clown_face: 

I would approve, but I did not try this out, so I will leave it to someone else. Great to finally get primitives support in the GUI!

rhaschke:
> I'm a bit amused to see how you tend to fix your nagging requests for changes

I should have pushed Felix to join forces with Jorge and come up with a unified solution in the first place, yes.
But, I didn't notice #2137 before.

rhaschke:
I just noticed, that the changes a user applies to the PlanningScene through this GUI are _local_ to the rviz plugin only and not (yet) propagated to the move_group's central planning scene. I suggest doing so. I'm working on this right now. So long, I marked this PR as WIP.

rhaschke:
Actually, I have chosen a middle ground solution now: Changes to the local planning scene are handled locally only, but are explicitly noticed (enabling the _Publish_ button). Only before planning, the user is asked whether to publish all local changes if not yet done. This explicitly brings the fact to the user's attention that rviz maintains its own planning scene.

As I added a lot of independent commits now, this PR should be merged as a proper **merge commit**.

v4hn:
Yes, the need to press "publish scene" after changing the scene was not obvious and improving this is a good idea.
Even worse, I'm not sure what happens when a new global scene is received between changing something and publishing it.
I suspect additional issues there.

Thanks for working on it.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -426,131 +437,132 @@ void MotionPlanningFrame::sceneUpdate(planning_scene_monitor::PlanningSceneMonit
...
+  while (planning_display_->getPlanningSceneRO()->getWorld()->hasObject(shape_name))
+  {
+    idx++;
+    shape_name = selected_shape + "_" + std::to_string(idx);
+  }
 
-      shapes::ShapeConstPtr shape(mesh);
-      Eigen::Isometry3d pose;
-      pose.setIdentity();
+  // Actually add object
```
felixvd, line 507: This does not publish the change but only adds the object to the plugin's planning scene copy, right? I would clarify this for first-time readers.
```suggestion
  // Add object to plugin's PlanningScene, but do not publish the change yet (only note that an edit was made)
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -73,24 +73,68 @@ QString subframe_poses_to_qstring(const moveit::core::FixedTransformsMap& subfra
...
 namespace moveit_rviz_plugin
 {
-void MotionPlanningFrame::importObjectFromFileButtonClicked()
+void MotionPlanningFrame::shapesComboBoxChanged(const QString& text)
 {
-  QString path = QFileDialog::getOpenFileName(this, tr("Import Object Mesh"), QString(),
-                                              "CAD files (*.stl *.obj *.dae);;All files (*.*)");
-  if (!path.isEmpty())
-    importResource("file://" + path.toStdString());
+  switch (ui_->shapes_combo_box->currentData().toInt())
```
felixvd, line 78: I guess it's clear from the usage, but I would not have guessed how this function works from just the name.

felixvd:
> Even worse, I'm not sure what happens when a new global scene is received between changing something and publishing it. 

That seems like an argument to keep the plugin's copy and the move_group node's PlanningScene in sync. I wouldn't mind that either.

rhaschke:
> That seems like an argument to keep the plugin's copy and the move_group node's PlanningScene in sync. I wouldn't mind that either.

This would be rather slow: Updates would need to be sent to the move_group's scene and the received back in rviz' scene.

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_frame.h
```diff
@@ -312,8 +303,8 @@ private Q_SLOTS:
   ros::Subscriber update_custom_goal_state_subscriber_;
   // General
   void changePlanningGroupHelper();
-  void addPrimitiveShape();
-  void importResource(const std::string& path);
+  void addSceneObject();
+  shapes::ShapePtr loadMeshResource(const std::string& url);
```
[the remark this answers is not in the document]
rhaschke: No, unify to `url` :wink:

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -67,6 +69,13 @@ MotionPlanningFrame::MotionPlanningFrame(MotionPlanningDisplay* pdisplay, rviz::
 {
   // set up the GUI
   ui_->setupUi(this);
+  ui_->shapes_combo_box->addItem("box", shapes::BOX);
```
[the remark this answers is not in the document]
rhaschke: Done.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
rhaschke: Added a comment.

rhaschke:
Thanks, @felixvd, for your comments.

> On two 18.04 machines (containers) I tested, the + - X icons did not appear. If the "Add", "Del" etc. text is a fallback, the button is not big enough to display it.

Maybe, we need to embed the icons into the plugin? These are default theme icons and should™ be available.

> The Size spinbox fields seem a little wide. Is there a way to make them smaller? It looks like there is a little dead space overall, but it might just be Qt.

Qt computes the minimal size of those spinboxes by the maximal width of numbers that can be potentially entered into them.
Thus, the width is determined by min and max values, which are 99?

> Apart from that, if we're overhauling the GUI we might as well include #2061 :)

I will have a short look.

> And remove the '.00' from the "Planning Attempts" option (it's an integer).

That was already done in #2076?!?

> And figure out why the panel is so wide.

I investigated yesterday: The width essentially boils down to the width of the "object position" spinboxes. As stated above, their width is determined by the min/max values, which were -1000...1000 and now are -999...999. So, I already saved 3x1 digits.

rhaschke:
I have set the default tab to "Planning" now. Fixes #2061.

simonschmeisser:
+1 from me for directly synchronizing, I don't think users need to know that there are separate copies and might actually trigger planning by separate means? The update rate back to rviz could be increased ...

PR #2199, opened 2020-07-07
Adapt repository for new moveit_resources layout

v4hn:
Related to https://github.com/ros-planning/moveit_resources/pull/36 .

There is no reason to include them from cmake anymore.
The resources are not in a single path anymore and there is no header exposed anymore.

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
@@ -233,7 +233,7 @@ TYPED_TEST_P(CollisionDetectorPandaTest, DistanceSelf)
   collision_detection::CollisionResult res;
   this->cenv_->checkSelfCollision(req, res, *this->robot_state_, *this->acm_);
   ASSERT_FALSE(res.collision);
-  EXPECT_NEAR(res.distance, 0.13, 0.01);
+  EXPECT_NEAR(res.distance, 0.022, 0.001);
 }
 
 TYPED_TEST_P(CollisionDetectorPandaTest, DistanceWorld)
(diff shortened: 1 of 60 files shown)


rhaschke:
I'm looking into the remaining (test) build issues right now.

rhaschke:
I hope I fixed most of the numerous issues. Will finish for now.

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
```diff
@@ -233,7 +233,7 @@ TYPED_TEST_P(CollisionDetectorPandaTest, DistanceSelf)
   collision_detection::CollisionResult res;
   this->cenv_->checkSelfCollision(req, res, *this->robot_state_, *this->acm_);
   ASSERT_FALSE(res.collision);
-  EXPECT_NEAR(res.distance, 0.13, 0.01);
+  EXPECT_NEAR(res.distance, 0.022, 0.001);
```
rhaschke, line 236: I had to adapt this value, because the old ACM additionally disabled collisions between link5 and hand as well as link7:
| old | new |
|------|-------|
| image | image |

moveit.rosinstall
```diff
@@ -23,7 +27,7 @@
 - git:
     local-name: moveit_tutorials
     uri: https://github.com/ros-planning/moveit_tutorials.git
-    version: master
+    version: pr-master-rework-resources
 - git:
     local-name: panda_moveit_config
```
jschleicher, line 32: panda_moveit_config is now used from moveit_resources and could be dropped here?!

v4hn: It wasn't used in MoveIt before either, but it's required for the tutorials and thus it's listed in this file.

v4hn:
rebased to resolve conflicts.

SansoneG:
@v4hn Is there anything still holding this PR back from getting merged?

Issue #1807, opened 2019-12-10
RuntimeError: Unable to connect to move_group action server 'move_group' within allotted time (5s)


RDaneelOlivav:
### Description
Hi ,
I am connecting to a robot that the connection is slower than usual, and I suspect that
the timeout could be not letting moveit work.
I've seen other issues posted here but none give a solution for this.
This is the most promising one, but doesnt tell how to access this fucntionality through the MoveGroupCommander or any python modules: [https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h#L76](url), [https://github.com/ros-planning/moveit/issues/636](ISSUE Closed)
And for what I've seen, it's a recurrent issue.

So how can I make the MoveGroupCommander wait more time?
Where do I have to touch, even if its in the source code of moveit,
to allow me to increase or even make infinite the waiting time?

### Your environment
* ROS Distro: [Kinetic|]
* OS Version: e.g. Ubuntu 16.04
* Binary build fro kinetic ROS

### Steps to reproduce
The best way is to launch this [http://www.rosject.io/l/f1bf66e/](ROSject) and follow the instructions at the bottom.
Just launch the simulation and execute the python command sat the end of the notebook to test that it works. Then pause the simulation and launch the exact same code. You will see that if its paused more than 5 seconds, it gives the error once you unpause the simulation. This is the closest there is to reproduce the same thing that happens in the real robot.

### Expected behaviour
It should move the robot arm

### Actual behaviour
Error is located [https://github.com/ros-planning/moveit/blob/590a9c36d80ab55882d77f4c349c6344680a499b/moveit_commander/src/moveit_commander/move_group.py#L53](File Error)

It gives the error : RuntimeError: Unable to connect to move_group action server 'move_group' within allotted time (5s)

rhaschke:
According to the PR you referenced, there was introduced a parameter `trajectory_execution/controller_connection_timeout`, which you could add in your moveit_config's `trajectory_execution.launch` file.
https://github.com/ros-planning/moveit/blob/fe39430d971fdfadcec0c7316fa67d41bf48b281/moveit_plugins/moveit_simple_controller_manager/include/moveit_simple_controller_manager/action_based_controller_handle.h#L79

guiolpei:
I have the same issue.

Ubuntu 16.04 & ROS Kinetic.

[...]

PR #2201, opened 2020-07-08
Exposed missing parameters from the MoveGroup Pyhton interface

gerardcanal:
### Description
The Python interface to MoveCommander was not exposing the `wait_for_servers` parameter as it does in C++. The wrapper already had it, so the only missing part was to add it to the Python interface. This should handle #1807 as now the timeout can be extended from the Python interface.

Edit: I also found that the getPlannerId python method was created but not interfaced, so I added it to too. 

I tested it in ROS melodic, so if this is merged it would be good if it could be ported to melodic too.

### Checklist

### DIFF
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -448,7 +448,7 @@ class MoveGroupCommander(object):
 
     def get_planner_id(self):
         """ Get the current planner_id """
-        self._g.get_planner_id()
+        return self._g.get_planner_id()
 
     def set_num_planning_attempts(self, num_planning_attempts):
         """ Set the number of times the motion plan is to be computed from scratch before the shortest solution is returned. The default value is 1. """
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -48,9 +48,9 @@ class MoveGroupCommander(object):
     Execution of simple commands for a particular group
     """
 
-    def __init__(self, name, robot_description="robot_description", ns=""):
+    def __init__(self, name, robot_description="robot_description", ns="", wait_for_servers=5.0):
         """ Specify the group name for which to construct this commander instance. Throws an exception if there is an initialization error. """
-        self._g = _moveit_move_group_interface.MoveGroupInterface(name, robot_description, ns)
+        self._g = _moveit_move_group_interface.MoveGroupInterface(name, robot_description, ns, wait_for_servers)
 
     def get_name(self):
         """ Get the name of the group this instance was initialized for """
--- a/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
@@ -381,6 +381,11 @@ public:
     return getName().c_str();
   }
 
+  const char* getPlannerIdCStr() const
+  {
+    return getPlannerId().c_str();
+  }
+
   bp::dict getNamedTargetValuesPython(const std::string& name)
   {
     bp::dict output;
@@ -695,6 +700,7 @@ static void wrap_move_group_interface()
   move_group_interface_class.def("set_max_acceleration_scaling_factor",
                                  &MoveGroupInterfaceWrapper::setMaxAccelerationScalingFactor);
   move_group_interface_class.def("set_planner_id", &MoveGroupInterfaceWrapper::setPlannerId);
+  move_group_interface_class.def("get_planner_id", &MoveGroupInterfaceWrapper::getPlannerIdCStr);
   move_group_interface_class.def("set_num_planning_attempts", &MoveGroupInterfaceWrapper::setNumPlanningAttempts);
   move_group_interface_class.def("plan", &MoveGroupInterfaceWrapper::planPython);
   move_group_interface_class.def("compute_cartesian_path", &MoveGroupInterfaceWrapper::computeCartesianPathPython);
--- a/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
@@ -576,7 +576,7 @@ static void wrap_move_group_interface()
   eigenpy::enableEigenPy();
 
   bp::class_<MoveGroupInterfaceWrapper, boost::noncopyable> move_group_interface_class(
-      "MoveGroupInterface", bp::init<std::string, std::string, bp::optional<std::string>>());
+      "MoveGroupInterface", bp::init<std::string, std::string, bp::optional<std::string, double>>());
 
   move_group_interface_class.def("async_move", &MoveGroupInterfaceWrapper::asyncMovePython);
   move_group_interface_class.def("move", &MoveGroupInterfaceWrapper::movePython);


RDaneelOlivav:
LOVE it ;)

gerardcanal:
As this PR has not yet been reviewed, I am adding a fix to add the missing getPlannerId function (it is related to the same issue on missing elements in the move_group python interface). I updated the title accordingly. 

I believe this can be merged if approved (and ported to melodic too).

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
```diff
@@ -381,6 +381,11 @@ class MoveGroupInterfaceWrapper : protected py_bindings_tools::ROScppInitializer
     return getName().c_str();
   }
 
+  const char* getPlannerIdCStr() const
+  {
+    return getPlannerId().c_str();
+  }
+
```
v4hn, line 388: Why do you need the additional wrapper?

[It seems that returning std::string should work just fine](https://stackoverflow.com/questions/44827957/using-stdstring-as-return-type-in-boost-python-exported-function)

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -448,7 +448,7 @@ def set_planner_id(self, planner_id):
 
     def get_planner_id(self):
         """ Get the current planner_id """
-        self._g.get_planner_id()
+        return self._g.get_planner_id()
```
v4hn, line 451: ouch.

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
gerardcanal: I thought the same but I got some compilation errors when I tried.
When looking into it I saw that this is how the other methods returning strings (i.e. getName or getEndEffectorLink) were handling it, so I took the same approach.

v4hn: > I thought the same but I got some compilation errors when I tried.

You do not by any chance have the error message at hand?

It would be nice to see this cumbersome wrapping removed for the other methods as well.

gerardcanal: Hi @v4hn, I do have it. The error made me think that the python bindings didn't have a way to convert the return type from std::string:
```
                 from /home/gerard/code/catkin_ws/src/moveit/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp:38:
/usr/include/boost/python/detail/invoke.hpp: In instantiation of ‘PyObject* boost::python::detail::invoke(boost::python::detail::invoke_tag_<false, true>, const RC&, F&, TC&) [with RC = boost::python::detail::specify_a_return_value_policy_to_wrap_functions_returning<const std::__cxx11::basic_string<char>&>; F = const std::__cxx11::basic_string<char>& (moveit::planning_interface::MoveGroupInterface::*)() const; TC = boost::python::arg_from_python<moveit::planning_interface::MoveGroupInterfaceWrapper&>; PyObject = _object]’:
/usr/include/boost/python/detail/caller.hpp:216:46:   required from ‘PyObject* boost::python::detail::caller_arity<1>::impl<F, Policies, Sig>::operator()(PyObject*, PyObject*) [with F = const std::__cxx11::basic_string<char>& (moveit::planning_interface::MoveGroupInterface::*)() const; Policies = boost::python::default_call_policies; Sig = boost::mpl::vector2<const std::__cxx11::basic_string<char>&, moveit::planning_interface::MoveGroupInterfaceWrapper&>; PyObject = _object]’
/usr/include/boost/python/object/py_function.hpp:38:33:   required from ‘PyObject* boost::python::objects::caller_py_function_impl<Caller>::operator()(PyObject*, PyObject*) [with Caller = boost::python::detail::caller<const std::__cxx11::basic_string<char>& (moveit::planning_interface::MoveGroupInterface::*)() const, boost::python::default_call_policies, boost::mpl::vector2<const std::__cxx11::basic_string<char>&, moveit::planning_interface::MoveGroupInterfaceWrapper&> >; PyObject = _object]’
/home/gerard/code/catkin_ws/src/moveit_gcc/moveit/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp:721:1:   required from here
/usr/include/boost/python/detail/invoke.hpp:86:14: error: no match for call to ‘(const boost::python::detail::specify_a_return_value_policy_to_wrap_functions_returning<const std::__cxx11::basic_string<char>&>) (const std::__cxx11::basic_string<char>&)’
     return rc( (tc().*f)(BOOST_PP_ENUM_BINARY_PARAMS_Z(1, N, ac, () BOOST_PP_INTERCEPT)) );
            ~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

                 from /home/gerard/code/catkin_ws/src/moveit/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp:38:
/usr/include/boost/python/detail/caller.hpp: In instantiation of ‘static const PyTypeObject* boost::python::detail::converter_target_type<ResultConverter>::get_pytype() [with ResultConverter = boost::python::detail::specify_a_return_value_policy_to_wrap_functions_returning<const std::__cxx11::basic_string<char>&>; PyTypeObject = _typeobject]’:
/usr/include/boost/python/detail/caller.hpp:238:19:   required from ‘static boost::python::detail::py_func_sig_info boost::python::detail::caller_arity<1>::impl<F, Policies, Sig>::signature() [with F = const std::__cxx11::basic_string<char>& (moveit::planning_interface::MoveGroupInterface::*)() const; Policies = boost::python::default_call_policies; Sig = boost::mpl::vector2<const std::__cxx11::basic_string<char>&, moveit::planning_interface::MoveGroupInterfaceWrapper&>]’
/usr/include/boost/python/object/py_function.hpp:48:35:   required from ‘boost::python::detail::py_func_sig_info boost::python::objects::caller_py_function_impl<Caller>::signature() const [with Caller = boost::python::detail::caller<const std::__cxx11::basic_string<char>& (moveit::planning_interface::MoveGroupInterface::*)() const, boost::python::default_call_policies, boost::mpl::vector2<const std::__cxx11::basic_string<char>&, moveit::planning_interface::MoveGroupInterfaceWrapper&> >]’
/home/gerard/code/catkin_ws/src/moveit_gcc/moveit/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp:721:1:   required from here
/usr/include/boost/python/detail/caller.hpp:100:98: error: ‘struct boost::python::detail::specify_a_return_value_policy_to_wrap_functions_returning<const std::__cxx11::basic_string<char>&>’ has no member named ‘get_pytype’
         return create_result_converter((PyObject*)0, (ResultConverter *)0, (ResultConverter *)0).get_pytype();
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
```

After looking into int in more detail, I think the issue may be with the return type of all these methods returning strings being `const std::string&`. Boost python doesn't seem to like the reference in the return type. I just tried it changing the `MoveGroupInterface::getPlannerId()` method to return a `const std::string` instead and it compiles just fine. 

I guess there's some reason (efficiency?) for the MoveGroupInterface to return the strings as const refs, so we can also get rid of the `c_str()` making the wrapper methods return the `std::string` instead of `const char*` (although `MoveGroupInterface::getDefaultPlannerId` returns a plain string by value).

If you want I can check all the string returning methods, fix them in either way and add it to this PR before it gets merged (or into a new one if that's better).

v4hn: Argh, the const ref is a problem indeed. There might be some bp magic that could use the return types by their actual value (ignoring the ref), but that's quite a bit deeper than what I hoped for.

Although it is somewhat ridiculous to optimize the MGI on the level of const refs, when it's actually a high-delay adapter that uses ROS communication, that's what we have right now.
In this context, the current approach to extract the `c_str()` from the const ref is probably the most efficient one, so feel free to leave it as-is.

Thank you for following up on this!

gerardcanal: Sound good! Glad I was able to help a bit :-)

v4hn:
Thanks for the contribution @gerardcanal :+1:

PR #2203, opened 2020-07-09
A library for servoing toward a moving pose

AndyZe:
This uses moveit_servo to move toward a (possibly moving) target pose, e.g. visual servoing.

The pose can be updated by ROS topic. The tracking is done with a PID controller for each Cartesian dimension.

Here's a video of tracking several moving waypoints with 1cm position tolerance:

faster_pose_seeking

@AdamPettinger @tylerjw @nbbrooks


moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -7,14 +7,18 @@ endif()
...
+set(SERVO_LIB_NAME moveit_servo_cpp_api)
 
 if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
-find_package(Boost REQUIRED)
+# For library compilation issue. See
+# https://stackoverflow.com/questions/25902519/building-error-cmake-can-not-be-used-when-making-a-shared-object-recompile-wi
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
```
v4hn, line 18: You don't need this if we build libraries as shared objects.
If you add it, you also remove a part of why people even want static libraries to begin with...

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -25,13 +29,13 @@ find_package(catkin REQUIRED COMPONENTS
...
 )
-find_package(Eigen3 REQUIRED)
 
 catkin_package(
   INCLUDE_DIRS
     include
   LIBRARIES
-    ${LIBRARY_NAME}
+    pose_tracking
+    ${SERVO_LIB_NAME}
```
v4hn, line 38: Did you consider making this an independent package in moveit? Isn't it worth it?

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -52,34 +56,48 @@ include_directories(
...
-#########################################
-## A library providing a C++ interface ##
-#########################################
+#########################################################
+## Library to process realtime twist or joint commands ##
+#########################################################
 
-add_library(${LIBRARY_NAME} SHARED
+# This library provides an interface for sending realtime twist or joint commands to a robot
+add_library(${SERVO_LIB_NAME} STATIC
```
v4hn, line 64: :-1: We should stay with `SHARED` libraries for all public interfaces (so literally all...) in MoveIt.

In ROS `SHARED` is the default as well (which users *can* switch if they really want to), so just drop the specifier altogether. I'm still puzzled why that default apparently changed in ROS2/ament.

moveit_ros/moveit_servo/include/moveit_servo/make_shared_from_pool.h
```diff
@@ -38,6 +38,7 @@
 
 #pragma once
 
+#include <boost/make_shared.hpp>
```
v4hn, line 41: Individual commits sometimes actually make sense... anyway, no hard feelings here.

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking/pid_controller.h
```diff
@@ -0,0 +1,83 @@
+#pragma once
+
+namespace moveit_servo
+{
+class PIDController
```
v4hn, line 5: What's wrong with https://github.com/ros-controls/control_toolbox/blob/melodic-devel/include/control_toolbox/pid.h#L117
or even your own https://bitbucket.org/AndyZe/pid/src/8268b7250e40c33a08d7faa7706885eb317eadbd/include/pid/pid.h#lines-52 ?

Having yet another PID implementation in the ecosystem seems senseless, though you might have your reasons for doing this.

AndyZe: Agree, I'll try to get the control_toolbox version working.

"My" PID package only works with ROS topics IIRC, but I want a fast C++ interface

AndyZe:
I'll get back to address these comments in about a week  :+1:

rhaschke:
I don't yet get the purpose of this. Could you please update the initial comment?
What is actually moved? At first glance, just for tracking a Cartesian pose, the code changes seem way to complicated.
But, maybe I misconceived the purpose of this.

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking/pid_controller.h
AndyZe: done

AndyZe:
> I don't yet get the purpose of this. Could you please update the initial comment?
> What is actually moved? At first glance, just for tracking a Cartesian pose, the code changes seem way to complicated.
> But, maybe I misconceived the purpose of this.

It's for applications like visual servoing. For example, picking vegetables from a moving platform. Video link below. I updated the description, too.

I can't think of too many ways to simplify it. v4hn's suggestion to use an external PID library dropped a couple hundred lines of code already. All I can think of would be using one PID controller for all dimensions and vectorizing the calculation, but then you would lose some level of control. For example, maybe you would want less aggressive tuning in the Z-direction so you don't crash into a conveyor belt.

https://www.youtube.com/watch?v=OEiZu0gaP6w

^quite jerky, I want to do better than that

moveit_ros/moveit_servo/CMakeLists.txt
```diff
@@ -52,34 +57,47 @@ include_directories(
...
-  ${Boost_LIBRARIES}
-  ${LIBRARY_NAME}
+  ${SERVO_LIB_NAME}
+)
+
+########################################
+## Library for servoing toward a pose ##
+########################################
+
+add_library(pose_tracking STATIC
```
v4hn, line 94: Please remove the STATIC definitions from the build instructions.

moveit_ros/moveit_servo/include/moveit_servo/pose_tracking/pose_tracking.h
```diff
@@ -0,0 +1,168 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+/*
+   Author: Andy Zelenak
+   Desc: Servoing. Track a pose setpoint in real time.
+*/
+
+#pragma once
+
+#include <moveit_servo/pose_tracking/pid_controller.h>
```
v4hn, line 41: Please [fix CI](https://travis-ci.com/github/ros-planning/moveit/jobs/363006744#L1456). You did not test-compile this patch yet.

v4hn (dismissed):
It's a reasonable bit of code to include in moveit_servo.

It does worry me that this is basically dead code though.
You do not expose it through an executable yet that we might run to verify it and there are no tests for it either.
Do you have plans to add an Interactive Marker application to moveit that integrates pose_tracking and moveit_servo_cpp_api to test with a moveit_resources robot?

Please address my comments on the `STATIC` library definitions.

AndyZe:
> Please address my comments on the `STATIC` library definitions.

Sorry, should have marked it WIP. I definitely will do that.

AdamPettinger:
> I can't think of too many ways to simplify it. v4hn's suggestion to use an external PID library dropped a couple hundred lines of code already. All I can think of would be using one PID controller for all dimensions and vectorizing the calculation, but then you would lose some level of control. For example, maybe you would want less aggressive tuning in the Z-direction so you don't crash into a conveyor belt.

Certainly with reasonably tuned PID's, both cases (PID for each direction, and 1 PID) wouldn't crash into the conveyor belt as long as the target pose did not collide with the object, right? I guess the controller could overshoot the target with poor PID parameters, and the one-in-each-direction method allows being careful in specific directions. 

Also on the topic of collisions, if this is being run via moveit_servo, you will get the benefits of it: collision checking and joint limit enforcement. So if the conveyor belt is in the collision scene, and the target pose gets placed inside of the collision object, servo should stop that collision. 

I don't have very strong opinions either way, but if the performance is close, I would choose the one that involves tuning less PID's

AndyZe:
> Certainly with reasonably tuned PID's, both cases (PID for each direction, and 1 PID) wouldn't crash into the conveyor belt as long as the target pose did not collide with the object, right?

@AdamPettinger and I talked about this today. For our current pick/place project, it does make sense to tune the pose tracker differently for different dimensions, so I want to keep 3 PIDs for X/Y/Z. It makes more parameters but the user can copy/paste the three settings if he wants to.

AndyZe:
> and there are no tests for it either.

@v4hn It was a C++ challenge for me but I added some integration tests

AndyZe:
OK @v4hn, the tests are finally passing except an unrelated Bullet collision detection test. My main problem was launching a std::thread() from within a ROS gtest, apparently that causes all kinds of chaos.

The AMD64 CI run had this warning, not sure what can be done about it:

`/root/ros_ws/src/moveit/moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h:149:21: error: field ‘stop_requested_’ has incomplete type ‘std::atomic<bool>’`

https://travis-ci.com/github/ros-planning/moveit/jobs/364939964

AndyZe:
the failing CI tests are unrelated

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -0,0 +1,304 @@
...
+  std::string target_pose_topic = "/" + parameter_ns_ + "/target_pose";
+  target_pose_sub_ =
+      nh_.subscribe<geometry_msgs::PoseStamped>(target_pose_topic, 1, &PoseTracking::targetPoseCallback, this);
+
+  // Publish outgoing twist commands to the Servo object
+  twist_stamped_pub_ =
+      nh_.advertise<geometry_msgs::TwistStamped>(servo_->getParameters().cartesian_command_in_topic, 1);
+}
+
+int8_t PoseTracking::moveToPose(const Eigen::Vector3d& positional_tolerance, const double angular_tolerance)
```
AdamPettinger, line 81: I am a bit confused by the `PoseTracking` interface. It seems the only way to tell `PoseTracking` to start going to a pose is by directly calling this function, and the only way to set the pose is by publishing it via ROS. I also think the first bit of this function means you have to publish the pose within a specific time of calling the function.

Would it make sense to allow the C++ call here to include a pose to move to (so the call is `tracker_->moveToPose(pos_tol, ang_tol, pose)`)?

Other thoughts: the `moveToPose` could be made into a timer callback that ensure the tracker is always moving to the current pose target (set via publishing). Then there are `start()` and `stop()` interfaces for controlling that timer, as in ServoCalcs/CollisionCheck

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+    parameter_ns_ = yaml_namespace;
+  }
+
+  // Wait for ROS parameters to load
+  ros::Time begin = ros::Time::now();
+  while (!ros::param::has(parameter_ns_ + "/planning_frame") && ((ros::Time::now() - begin).toSec() < ROS_STARTUP_WAIT))
+  {
+    ROS_WARN_STREAM_NAMED(LOGNAME, "Waiting for parameter: " << parameter_ns_ + "/planning_frame");
+    ros::Duration(0.1).sleep();
```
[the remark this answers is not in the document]
AndyZe: that is a good question. I'll wait for somebody else to weigh in or not

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -0,0 +1,304 @@
...
+constexpr double ROS_STARTUP_WAIT = 10;       // sec
+constexpr double DEFAULT_POSE_TIMEOUT = 0.1;  // sec
+}
+
+namespace moveit_servo
+{
+PoseTracking::PoseTracking(const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor,
+                           const std::string& parameter_ns)
+  : planning_scene_monitor_(planning_scene_monitor)
+  , loop_rate_(DEFAULT_LOOP_RATE)
```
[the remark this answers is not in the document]
AndyZe: good point, that stuff isn't necessary

moveit_ros/moveit_servo/src/pose_tracking.cpp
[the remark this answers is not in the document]
v4hn: Duration.sleep constants are not that essential and with better design (possibly of the ROS interfaces...) there would be no need to have it at all. I don't think it's worth a name here.

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -0,0 +1,304 @@
...
+  }
+  target_pose_ = msg;
+  target_pose_.header.stamp = ros::Time::now();
+}
+
+geometry_msgs::TwistStampedConstPtr PoseTracking::calculateTwistCommand()
+{
+  // use the shared pool to create a message more efficiently
+  auto msg = moveit::util::make_shared_from_pool<geometry_msgs::TwistStamped>();
+  msg->header.frame_id = target_pose_.header.frame_id;
```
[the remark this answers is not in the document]
AndyZe: I believe so, since targetPoseCallback() transforms target_pose_ to the planning frame.

moveit_ros/moveit_servo/src/pose_tracking.cpp
AndyZe: The reason I want to publish the target pose on a ROS topic is to update it in real time. You can't do that with a call like this:

`tracker_->moveToPose(pos_tol, ang_tol, pose)`

I'll think about stuffing this in a timer callback

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -0,0 +1,304 @@
...
+  joint_model_group_ = robot_model_->getJointModelGroup(move_group_name_);
+
+  // Initialize PID controllers
+  initializePID(x_pid_config_, cartesian_position_pids_);
+  initializePID(y_pid_config_, cartesian_position_pids_);
+  initializePID(z_pid_config_, cartesian_position_pids_);
+  initializePID(angular_pid_config_, cartesian_orientation_pids_);
+
+  // Use the C++ interface that Servo provides
+  servo_ = std::make_unique<moveit_servo::Servo>(nh_, planning_scene_monitor_, parameter_ns_);
```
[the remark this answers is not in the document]
AndyZe: Passing a shared ptr to a Servo object makes a lot of sense.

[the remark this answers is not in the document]
AndyZe: It looks like there are some thread safety issues here. Have not been able to figure it out in a few hours.

moveit_ros/moveit_servo/src/pose_tracking.cpp
AndyZe: @AdamPettinger (and @tylerjw) can you be more specific about the benefits of moving this into a Timer callback?

AndyZe:
Marking as WIP until the demo Adam made works well

nbbrooks (commented):
In servo_calcs.h

`bool stop_requested_ = false;`

should be an atomic bool. Possibly some other vars there as well.

AndyZe:
> `bool stop_requested_ = false;`
> 
> should be an atomic bool. Possibly some other vars there as well.

Good point. Fixed it, as well as some other member variables that could be modified in public member functions

moveit_ros/moveit_servo/src/pose_tracking.cpp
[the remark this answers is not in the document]
AndyZe: I finally made the Servo instance public because I wanted access to public member functions like `setPaused()`. Thanks for that suggestion!

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+    while (msg_count < 100)
+    {
+      target_pose_pub_.publish(target_pose);
+      ros::Duration(0.01).sleep();
+      ++msg_count;
+    }
+  });
+
+  ros::Duration(ROS_PUB_SUB_DELAY).sleep();
+  tracker_->moveToPose(translation_tolerance_, ROTATION_TOLERANC, 1 /* target pose timeout */);
```
henningkayser, line 141:
```suggestion
  tracker_->moveToPose(translation_tolerance_, ROTATION_TOLERANCE, 1 /* target pose timeout */);
```

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -0,0 +1,346 @@
...
+         (fabs(z_error) < positional_tolerance(2) && fabs(angular_error_) < angular_tolerance);
+}
+
+void PoseTracking::targetPoseCallback(const geometry_msgs::PoseStampedConstPtr& msg)
+{
+  // Transform to MoveIt planning frame
+  target_pose_ = *msg;
+  if (target_pose_.header.frame_id != planning_frame_)
+  {
+    auto target_to_planning_frame = transform_buffer_.lookupTransform(planning_frame_, target_pose_.header.frame_id,
```
henningkayser, line 244: This function can throw exceptions, better wrap it in a try/catch block.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -69,12 +69,39 @@ bool isNonZero(const control_msgs::JointJog& msg)
   };
   return !all_zeros;
 }
+
+// Helper function for converting Eigen::Isometry3d to geometry_msgs/TransformStamped
+geometry_msgs::TransformStamped convertIsometryToTransform(const Eigen::Isometry3d& eigen_tf,
```
henningkayser, line 74: Better use https://github.com/ros2/geometry2/blob/ros2/tf2_eigen/include/tf2_eigen/tf2_eigen.h#L92 and set the frames separately.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -982,6 +1026,42 @@ bool ServoCalcs::getCommandFrameTransform(Eigen::Isometry3d& transform)
   return !transform.matrix().isZero(0);
 }
 
+bool ServoCalcs::getCommandFrameTransform(geometry_msgs::TransformStamped& transform)
+{
+  const std::lock_guard<std::mutex> lock(latest_state_mutex_);
+  // All zeros means the transform wasn't initialized, so return false
+  if (tf_moveit_to_robot_cmd_frame_.matrix().isZero(0))
```
henningkayser, line 1033: Shouldn't we initialize the matrix as zero matrix explicitly for this to work reliably with all compiler settings?

moveit_ros/moveit_servo/test/pose_tracking_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+  target_pose.pose.orientation.w = 1;
+
+  // Republish the target pose in a new thread, as if the target is moving
+  std::thread target_pub_thread([&] {
+    size_t msg_count = 0;
+    while (msg_count < 100)
+    {
+      target_pose_pub_.publish(target_pose);
+      ros::Duration(0.01).sleep();
+      ++msg_count;
```
henningkayser, line 136: Nit, this could go into the loop

moveit_ros/moveit_servo/test/servo_cpp_interface_test.cpp
```diff
@@ -59,6 +60,7 @@ class ServoFixture : public ::testing::Test
 public:
   void SetUp() override
   {
+    ros::Duration(ROS_SETUP_DELAY).sleep();
```
henningkayser, line 63: It would be more robust to wait for certain topics/nodes to be available. Fixed timeouts can easily lead to flaky CI tests.

henningkayser (commented):
It's really awesome to get this feature available in MoveIt. I didn't see any major issues with the implementation, just some minor comments.

@v4hn could you give this a re-review to unblock this PR?

v4hn:
I don't have time to review this at the moment, but I dismissed my review (you could have done so to, all feedback I asked for got addressed).

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: good idea, done.

moveit_ros/moveit_servo/test/servo_cpp_interface_test.cpp
AndyZe: I like the idea but (in practice) CI still fails. I'll add a check for a few key topics/parameters on top of the delay, though.

If you're curious, here's the CI run that failed without a startup delay.  https://github.com/ros-planning/moveit/runs/1210638771

AndyZe: oh, it was a different issue. Done

henningkayser (approved):

henningkayser:
These are a lot of commits. Do you want to break this into a few readable chunks? Otherwise, we should do a squash merge.

AndyZe:
> These are a lot of commits. Do you want to break this into a few readable chunks? Otherwise, we should do a squash merge.

I don't see well-defined ways to chunk it. Suppose I will squash-merge it. Thanks for the review

PR #2204, opened 2020-07-09
Check collisions during joint motions, too

AndyZe:
- Check collisions during joint motions (just like they are with Cartesian motions)

- Print colliding objects, for debugging

### DIFF
--- a/moveit_ros/moveit_servo/src/collision_check.cpp
+++ b/moveit_ros/moveit_servo/src/collision_check.cpp
@@ -212,15 +212,15 @@ void CollisionCheck::printCollisionPairs(collision_detection::CollisionResult::C
   if (!contact_map.empty())
   {
     // Throttled error message about the first contact in the list
-    ROS_ERROR_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "Objects in collision (among others, possibly): "
-                                                                          << contact_map.begin()->first.first << ", "
-                                                                          << contact_map.begin()->first.second);
+    ROS_WARN_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "Objects in collision (among others, possibly): "
+                                                                         << contact_map.begin()->first.first << ", "
+                                                                         << contact_map.begin()->first.second);
     // Log all other contacts if in debug mode
     ROS_DEBUG_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "Objects in collision:");
     for (auto contact : contact_map)
     {
-      ROS_DEBUG_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, contact.first.first << ", "
-                                                                                            << contact.first.second);
+      ROS_DEBUG_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "\t" << contact.first.first << ", "
+                                                                             << contact.first.second);
     }
   }
 }
--- a/moveit_ros/moveit_servo/src/collision_check.cpp
+++ b/moveit_ros/moveit_servo/src/collision_check.cpp
@@ -209,11 +209,11 @@ void CollisionCheck::run(const ros::TimerEvent& timer_event)
 
 void CollisionCheck::printCollisionPairs(collision_detection::CollisionResult::ContactMap& contact_map)
 {
-  if (!collision_result_.contacts.empty())
+  if (!contact_map.empty())
   {
     ROS_ERROR_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME,
-      "Ojects in collision (among others, possibly): " << collision_result_.contacts.begin()->first.first << ", "
-      << collision_result_.contacts.begin()->first.second);
+      "Ojects in collision (among others, possibly): " << contact_map.begin()->first.first << ", "
+      << contact_map.begin()->first.second);
   }
 }
 
--- a/moveit_ros/moveit_servo/src/collision_check.cpp
+++ b/moveit_ros/moveit_servo/src/collision_check.cpp
@@ -211,7 +211,7 @@ void CollisionCheck::printCollisionPairs(collision_detection::CollisionResult::C
 {
   if (!contact_map.empty())
   {
-    ROS_ERROR_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "Ojects in collision (among others, possibly): "
+    ROS_ERROR_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "Objects in collision (among others, possibly): "
                                                                           << contact_map.begin()->first.first << ", "
                                                                           << contact_map.begin()->first.second);
   }
--- a/moveit_ros/moveit_servo/src/collision_check.cpp
+++ b/moveit_ros/moveit_servo/src/collision_check.cpp
@@ -211,9 +211,9 @@ void CollisionCheck::printCollisionPairs(collision_detection::CollisionResult::C
 {
   if (!contact_map.empty())
   {
-    ROS_ERROR_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME,
-      "Ojects in collision (among others, possibly): " << contact_map.begin()->first.first << ", "
-      << contact_map.begin()->first.second);
+    ROS_ERROR_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "Ojects in collision (among others, possibly): "
+                                                                          << contact_map.begin()->first.first << ", "
+                                                                          << contact_map.begin()->first.second);
   }
 }
 
--- a/moveit_ros/moveit_servo/src/collision_check.cpp
+++ b/moveit_ros/moveit_servo/src/collision_check.cpp
@@ -64,6 +64,7 @@ CollisionCheck::CollisionCheck(ros::NodeHandle& nh, const moveit_servo::ServoPar
   // Init collision request
   collision_request_.group_name = parameters_.move_group_name;
   collision_request_.distance = true;  // enable distance-based collision checking
+  collision_request_.contacts = true;  // Record the names of collision pairs
 
   if (parameters_.collision_check_rate < MIN_RECOMMENDED_COLLISION_RATE)
     ROS_WARN_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME,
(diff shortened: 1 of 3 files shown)


moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -203,6 +207,16 @@ void CollisionCheck::run(const ros::TimerEvent& timer_event)
...
+void CollisionCheck::printCollisionPairs(collision_detection::CollisionResult::ContactMap& contact_map)
+{
+  if (!collision_result_.contacts.empty())
+  {
+    ROS_ERROR_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME,
+      "Ojects in collision (among others, possibly): " << collision_result_.contacts.begin()->first.first << ", "
+      << collision_result_.contacts.begin()->first.second);
+  }
+}
+
```
tylerjw, line 219: This is nice, if we get here it'd be nice to know what is in collision in the logs.

henningkayser: This is really helpful. Maybe we should print all collisions but make it a DEBUG statement?

AndyZe: Just noticed that the argument here isn't being used. Will fix that

AndyZe: @henningkayser I tried to think about how I would prefer to go about debugging collisions. You're right, I think printing every collision pair might be too much. But switching to DEBUG logging is a little bit of a pain, and it may be an unexpected collision and the user was not in DEBUG mode beforehand. Also, I'm not sure how many ROS users even know how to switch console logging modes.

moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -203,6 +207,16 @@ void CollisionCheck::run(const ros::TimerEvent& timer_event)
   }
 }
 
+void CollisionCheck::printCollisionPairs(collision_detection::CollisionResult::ContactMap& contact_map)
+{
+  if (!contact_map.empty())
+  {
+    ROS_ERROR_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "Ojects in collision (among others, possibly): "
```
jliukkonen, line 214: Typo in word objects

moveit_ros/moveit_servo/src/collision_check.cpp
henningkayser: I would consider using the debug log a ROS basic so for me it's more a matter of usefulness. For debugging collisions DEBUG seems like the perfect logger level. We should still have a (throtttled) warning if we are decelerating and there we could add the number of collisions and possibly the first collision pair.

AndyZe: done

moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -64,6 +64,7 @@ CollisionCheck::CollisionCheck(ros::NodeHandle& nh, const moveit_servo::ServoPar
   // Init collision request
   collision_request_.group_name = parameters_.move_group_name;
   collision_request_.distance = true;  // enable distance-based collision checking
+  collision_request_.contacts = true;  // Record the names of collision pairs
```
mlautman, line 67: I'm not sure if this slows down collision checking or not... It might be worth profiling this method to see

moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -203,6 +207,24 @@ void CollisionCheck::run(const ros::TimerEvent& timer_event)
   }
 }
 
+void CollisionCheck::printCollisionPairs(collision_detection::CollisionResult::ContactMap& contact_map)
+{
+  if (!contact_map.empty())
+  {
+    // Throttled error message about the first contact in the list
+    ROS_ERROR_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "Objects in collision (among others, possibly): "
```
mlautman, line 215: Is this really an error or should it be a warning?

mlautman: I wouldn't want to see people filtering out all debug info if this gets spammy. I think a rosparam for verbose collision checking would be a nice addition.

moveit_ros/moveit_servo/src/collision_check.cpp
mlautman: You may also want to add a rosparam for optionally storing and logging the collision contacts.

mlautman: I feel like @mamoll might now if storing all of the collision pairs slows down collision checking (versus potentially returning after the collision checker finds the first collision)

moveit_ros/moveit_servo/src/collision_check.cpp
mlautman: If you modify the `collision_request_.contacts = verbose_collision_checking_` (where `verbose_collision_checking_` is set from a rosparam) than maybe `if (!contact_map.empty())` will always be false and you won't have to add additional logic in this method. Something to investigate

moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -203,6 +207,24 @@ void CollisionCheck::run(const ros::TimerEvent& timer_event)
...
+                                                                          << contact_map.begin()->first.first << ", "
+                                                                          << contact_map.begin()->first.second);
+    // Log all other contacts if in debug mode
+    ROS_DEBUG_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "Objects in collision:");
+    for (auto contact : contact_map)
+    {
+      ROS_DEBUG_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, contact.first.first << ", "
```
mlautman, line 222: super nit: 
```suggestion
      ROS_DEBUG_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME, "\t" << contact.first.first << ", "
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -598,6 +596,12 @@ void ServoCalcs::applyVelocityScaling(Eigen::ArrayXd& delta_theta, double singul
   }
 
   delta_theta = collision_scale * singularity_scale * delta_theta;
+
+  if (status_ == StatusCode::HALT_FOR_COLLISION)
+  {
+    ROS_ERROR_STREAM_THROTTLE_NAMED(5, LOGNAME, "Halting for collision!");
```
mlautman, line 602: nit: 5 feels really really long. Like long enough that the user might not see this message before assuming that their code must be buggy. Maybe 2 or 3 is a more sensible period.

mlautman: Also, this isn't really an error because halting for potential collisions is an expected behavior. 
A good shorthand is:
 - FATAL for when the code should stop executing because something terrible has happened
 - ERROR for things that are outside normal operating conditions but the error is recoverable enough that code can keep executing.
 - WARNING for when something bad is happening that should be avoided but it is within normal operating conditions. 
 - INFO for output that might be important to a casually observer looking at the console output to understand what is going on with the code
 - DEBUG for output that might be interesting to a developer who is debugging this section of code

moveit_ros/moveit_servo/src/collision_check.cpp
mamoll: I think @mlautman's intuition is right. Definitely something to check. I see that you already ask for distance information. This adds significant computational cost. The difference with just a boolean collision check can be orders of magnitude. The code path in FCL is entirely different depending on whether you want a collision check or distance info. Not sure how much worse it gets if you add contact info on top of that.

AndyZe: I dropped a high-resolution timer in the CC loop and checked runtime, with and without contacts recorded. The difference is really small.

with contacts recorded:  0.017-0.026s per loop
without contacts recorded: 0.016-0.028s

I think there would be a bigger difference if the robot were in contact, but it would have stopped moving by then anyway. So we don't care about that.

moveit_ros/moveit_servo/src/collision_check.cpp
AndyZe: done!

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: OK, I'll make it 3 seconds. And roger that, I'll turn it into a WARNING. Thanks

moveit_ros/moveit_servo/src/collision_check.cpp
AndyZe: will change it to a warning.

Not sure about making in configurable. It's already throttled so it doesn't get printed very often, and we already have a bunch of parameters to wade through (~40 of them)

moveit_ros/moveit_servo/src/collision_check.cpp
mlautman: I'm not sure I understand why we wouldn't care about run time if the robot is in contact. Maybe I am not quite understanding how this feature will be used. 

My assumption is that this would be used to check if a joint motion would result in a collision before sending a joint command and wait until the user sends a command that isn't in collision. If that is the case, wouldn't we still want the cycle time to be really fast so that it can keep up with new commands at a high rate? Is the assumption that even if it takes longer in a cycle when the joint command would result in a collision, that eventually a new command would come in with a non-colliding command and the cycle time would be really fast again. This seems non-deterministic to me but I am also not overly familiar with the majority of the code base here.

AndyZe: What you're describing would be nice but it doesn't work like that. The jogger just stops if it's in collision. There's an optional buffer so it will start decelerating before it collides.

So yeah, once it collides it's stuck

AndyZe: Also, collision checking is pretty slow (about 50 Hz). Usually robot commands will go out at a higher rate than that

mlautman: Okay, this sounds reasonable. I still don't fully understand how the collision checking and the jogging are run at different rates but I trust that the design decision you are making here are sound. If you are confident that this won't affect performance and that there isn't much need for a parameter that would disable verbose collision checking than I'm happy to merge.

AndyZe:
@mlautman @tylerjw @henningkayser  Is anybody going to click the merge button?

rhaschke:
@tylerjw: Why did you merge this? This introduced a Travis regression in the Noetic build - as indicated by the failing build.

tylerjw:
@rhaschke It was passing.  It should have been rebased on master as the noetic build was added to travis after this PR was starting.  I'll look into the regression and see if I can fix that.

tylerjw:
If you look through the output of the failing Noetic build it seems like it has something to do either with a robot description or python: https://travis-ci.com/github/ros-planning/moveit/jobs/369428306

I don't see anything in there about servo.

rhaschke:
> It was passing. It should have been rebased on master as the noetic build was added to travis after this PR was starting. I'll look into the regression and see if I can fix that.

@tylerjw, sorry for blaming you. I concluded from the last failing Travis build in this PR.
I fixed Travis by triggering a new build of the Noetic docker image(s).

tylerjw:
Thank you, I'm glad you figured it out.  I couldn't figure out what was broken.  :+1:

PR #2211, opened 2020-07-17
Add named logging with LOGNAME to OMPL interface

JeroenDM:
### Description

As is done in other parts of MoveIt #1079 #1179, this pull request adds named logging to all source files of the `moveit_planners_ompl` package. (For some files with a single log statement, it might be overkill, but it is consistent...)

**Important** I'm not confident whether it is ok to put this outside a namespace. But in most source files of the OMPL interface, all code is put outside of a namespace and all names are prepended with `ompl_interface::`. Maybe this is worth another cleanup? Or is there a reason for this?

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp
@@ -41,7 +41,7 @@
 
 #include <utility>
 
-static const std::string LOGNAME{ "constrained_goal_sampler" };
+constexpr char LOGNAME[] = "constrained_goal_sampler";
 
 ompl_interface::ConstrainedGoalSampler::ConstrainedGoalSampler(const ModelBasedPlanningContext* pc,
                                                                kinematic_constraints::KinematicConstraintSetPtr ks,
--- a/moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp
@@ -41,7 +41,10 @@
 
 #include <utility>
 
+namespace ompl_interface
+{
 constexpr char LOGNAME[] = "constrained_goal_sampler";
+}  // namespace ompl_interface
 
 ompl_interface::ConstrainedGoalSampler::ConstrainedGoalSampler(const ModelBasedPlanningContext* pc,
                                                                kinematic_constraints::KinematicConstraintSetPtr ks,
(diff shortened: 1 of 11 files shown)


JeroenDM:
I think I used the wrong clang-format configurations, I will fix it.

JeroenDM:
Tavis reports failing tests in `moveit/moveit_ros/perception/mesh_filter/test/mesh_filter_test`. I don't know what to make of it.

rhaschke:
The failing mesh_filter tests are unrelated (see #2212 and a pending fix in #2180).

JeroenDM:
Just curious, is there another way to quickly check which tests failed on Travis, besides reading through the XML output?

v4hn:
The short summaries are listed at the bottom of the log: https://travis-ci.com/github/ros-planning/moveit/jobs/362068846#L1837

JeroenDM:
Aaah, I looked over it, sorry. Thank you :)

moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp
```diff
@@ -41,6 +41,8 @@
 
 #include <utility>
 
+static const std::string LOGNAME{ "constrained_goal_sampler" };
```
mamoll, line 44: Please put this inside the ompl_interface namespace (the same applies to all cases below). In addition, write it as
```
constexpr char LOGNAME[] = "constrained_goal_sampler";
```
Constexprs offer the compilers more options for optimization. A std::string cannot be created via constexpr.

JeroenDM: Ok!

Can I put all class method definitions inside the namespace (as is done in many other parts of MoveIt, but not [here](https://github.com/ros-planning/moveit/blob/877e1d51560f701b60b79e04a852ec104ad77de0/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp#L88) for example). Or is the preferred way to do `ompl_interface::LOGNAME` in all the log statements?

v4hn:
> Or is the preferred way to do `ompl_interface::LOGNAME` in all the log statements?

LOGNAME should not be used with a namespace definition. If you do, I guess it would mean you're lying. :-)

JeroenDM:
I must admit I don't really understand what I did.

From what I understand (after some googling) the constexpr adheres to normal namespace behavior in this context. The LOGNAME's in the rest of the file work because they are used inside class methods that are defined inside the same namespace, and therefore LOGNAME is available.

Ideally, I would like to extend the namespace scope to the end of the source files, as done in [ompl_planner_manager.cpp](https://github.com/ros-planning/moveit/blob/84cce2ae2a342abd180b4ff97063c3c6dc8b7aa8/moveit_planners/ompl/ompl_interface/src/ompl_planner_manager.cpp#L313). It would make the code less verbose. But that's probably worth a separate issue / pull request.

PR #2216, opened 2020-07-20
WIP: OMPL Constrained Planning (Part 1)

JeroenDM:
### Description

This pull request is part of a larger effort to support [OMPL's Constrained planning capabilities](http://ompl.kavrakilab.org/constrainedPlanning.html) in MoveIt.

In the current OMPL interface, the [StateValidityChecker](https://github.com/ros-planning/moveit/blob/edc9cf6e55ffcd2f3dbe18f2eb8602d913fdde48/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/state_validity_checker.h) decides which state is valid and can be used in a solution path. It [checks](https://github.com/ros-planning/moveit/blob/edc9cf6e55ffcd2f3dbe18f2eb8602d913fdde48/moveit_planners/ompl/ompl_interface/src/detail/state_validity_checker.cpp#L126) whether the state is:

- within the joint limits
- satisfies the path constraints
- is collision free
- is feasible. (Based on a user provides feasibility metric?)

This pull request adds the option to disable checking the path constraints. This is necessary to integrate OMPL’s constrained planners, where the path constraints are handled by the state space and do not need to be checked by the StateValidityChecker.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/state_validity_checker.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/state_validity_checker.h
@@ -67,6 +67,14 @@ public:
   virtual double cost(const ompl::base::State* state) const;
   double clearance(const ompl::base::State* state) const override;
 
+  /** \brief Enable / disable checking path constraints in isValid method.
+   * 
+   * (default) Set true for rejection sampling of path constraints.
+   * Set false if the path constraints are handled by OMPL's state space.
+   * 
+   * */
+  void setCheckPathConstraints(bool flag);
+
   void setVerbose(bool flag);
 
 protected:
@@ -85,6 +93,7 @@ protected:
   collision_detection::CollisionRequest collision_request_with_distance_verbose_;
 
   collision_detection::CollisionRequest collision_request_with_cost_;
+  bool check_path_constraints_;
   bool verbose_;
 };
 }  // namespace ompl_interface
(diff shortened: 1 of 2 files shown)


v4hn:
Please [fix CI](https://travis-ci.com/github/ros-planning/moveit/jobs/362827753#L1398)

JeroenDM:
It's embarrassing that my clang formatting always goes wrong... I fixed some problems with my local setup, I hope it doesn't happen again now.

JeroenDM:
@v4hn Just for clarity, my intention is to only merge this after we agree on the overall approach. This means I'll first do *WIP: OMPL Constrained Planning (Part 3)* to have an idea of the whole picture. I'll add this to the description.

v4hn:
> my intention is to only merge this after we agree on the overall approach. This means I'll first do *WIP: OMPL Constrained Planning (Part 3)* to have an idea of the whole picture. I'll add this to the description.

Ok.

ommmid:
To elaborate the purpose of this PR: what is wrong to have another layer of validity check for satisfying constraints (even if the constraints are handled by state space)? is it time consuming or any other reason? why is that necessary?

mamoll:
Validity checking can be time consuming, esp. if it involves collision checks.

JeroenDM:
@ommmid but this pull request is indeed not strictly necessary to make contained planning work. It will be interesting to see the performance difference of turning this on and off.

PR #2223, opened 2020-07-22
Fix: checkbox selection of displaying for each link is not saved by deferred loading

### LINKED ISSUE #2222: RViZ/Trajectory: checkbox selection of displaying for each link is not saved
### Description

In `Trajectory` visualization on RViZ, displaying of specific link can be enabled/disabled by  marking/unmarking checkbox for each link in "Links" tree which is really useful.
However, the selection is not saved to config file and all checkboxes are enabled on starting RViZ.

Untitled

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Source or Binary build?: Both
* If binary, which release version?: 1.0.5
* If source, which branch?: melodic-devel

### Steps to reproduce

Please refer to the image above

### Expected behaviour

The selection of checkboxes are saved.

### Actual behaviour

The selection is not saved to config file and all checkboxes are enabled on starting RViZ

furushchev:
Closes #2222 

The cause of the issue #2222 is that in the original codes the UI parts are initialized on `update` function instead of `onEnable` callback which occurs resetting saved selection.
This PR fixes deferred loading by using `std::async` and prevents calling `update` method until robot model is loaded to avoid displaying lots of warning messages.


furushchev:
@v4hn Hi Michael!
Thank you for the quick feedback!
It's great pleasure being involved with the great project! :sunglasses: 
It looks it is better to follow the manner of using `addBackgroundJob` if it also works on the other classes. I'll check if it works.

moveit_ros/visualization/trajectory_rviz_plugin/src/trajectory_display.cpp
```diff
@@ -93,7 +95,8 @@ void TrajectoryDisplay::reset()
 void TrajectoryDisplay::onEnable()
 {
   Display::onEnable();
-  load_robot_model_ = true;  // allow loading of robot model in update()
+  // deferred loading
+  std::async(std::launch::async, &TrajectoryDisplay::loadRobotModel, this);
```
rhaschke, line 99: `loadRobotModel()` must not be called in a thread, but in the main GUI thread.

furushchev:
@rhaschke Thank you for the feedback!
True. The deferred loading was introduced [here](https://github.com/ros-planning/moveit/commit/4e72ac5f96863e7c37a47706645c6634d7672f7b) for avoiding warning messages on `loadRobotModel`, which can now be avoided by using `atomic<bool> robot_model_loaded_` and preventing calling `update` until `loadRobotModel` is completed.
I removed the `std::async` part and now confirmed that it works fine.
(And allso applied the same change to `RobotStateDisplay` class as @v4hn mentioned.)

moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
```diff
@@ -351,7 +351,7 @@ void RobotStateDisplay::unsetLinkColor(rviz::Robot* robot, const std::string& li
 // ******************************************************************************************
 void RobotStateDisplay::loadRobotModel()
 {
-  load_robot_model_ = false;
+  robot_model_loaded_ = false;
```
v4hn, line 354: I'm not sure in which thread `Display::reset()` is processed, but it can call `robot_->clear()` and will call `loadRobotModel()` only *afterwards*.
Shouldn't you set `robot_model_loaded_ = false` *before* clearing the model, instead of in here?

furushchev: @v4hn Thanks for the comment. Added setting `robot_model_loaded_` to `false` also in the top of lines in `reset`.

furushchev:
```
build/moveit_ros_perception/test_results/moveit_ros_perception/gtest-mesh_filter_test.xml: 48 tests, 0 errors, 24 failures, 0 skipped
```

Some tests fail because of non-related errors?

v4hn:
It's not just the mesh_filter failing.
Looks like this fails for kinetic with some issues related to `std::atomic`: https://travis-ci.com/github/ros-planning/moveit/jobs/365579117#L933

Any idea @tylerjw ? Could you please resolve this @furushchev ? The failing mesh_filter tests are an independent issue we need to fix elsewhere. Sorry for the late feedback.

moveit_ros/visualization/robot_state_rviz_plugin/include/moveit/robot_state_rviz_plugin/robot_state_display.h
```diff
@@ -132,7 +132,7 @@ private Q_SLOTS:
   robot_state::RobotStatePtr robot_state_;
   std::map<std::string, std_msgs::ColorRGBA> highlights_;
   bool update_state_;
-  bool load_robot_model_;  // for delayed robot initialization
```
tylerjw, line 135: You need to add:

```c++
#include <atomic>
```

to the top of this file for this to compile on the version of the compiler used for kinetic.

moveit_ros/visualization/trajectory_rviz_plugin/include/moveit/trajectory_rviz_plugin/trajectory_display.h
```diff
@@ -89,7 +89,7 @@ private Q_SLOTS:
   rdf_loader::RDFLoaderPtr rdf_loader_;
   robot_model::RobotModelConstPtr robot_model_;
   robot_state::RobotStatePtr robot_state_;
-  bool load_robot_model_;  // for delayed robot initialization
+  std::atomic<bool> robot_model_loaded_;
```
tylerjw, line 92: same for this file, add the include for atomic

v4hn:
@tylerjw You should try to fixup a user-provided pull-request with a commit one of these days.
Now that you do have maintainer status you can push to these branches too. :sunglasses:

moveit_ros/visualization/robot_state_rviz_plugin/include/moveit/robot_state_rviz_plugin/robot_state_display.h
tylerjw: done

moveit_ros/visualization/trajectory_rviz_plugin/include/moveit/trajectory_rviz_plugin/trajectory_display.h
tylerjw: done

tylerjw:
@v4hn I tried figuring out how to use the github cli for doing that (gh) but it doesn't seem to have nice support for this feature.  I figured it out though, just had to manually add the user's remote and push to it.

v4hn:
> the github cli for doing that (gh)

Is it worth using that over `hub`? Would be the same workflow there though.

Awesome! :-)

tylerjw:
> Is it worth using that over `hub`?

It does have some nice features for checking out PRs that `hub` isn't as nice for.

rhaschke:
I think to resolve the chicken-egg-issue we need to provide a custom load() method, which first loads the robot_description parameter before calling the parent's `load()`.

v4hn:
> I think this reverts essentially the attempt of #239.

You are right. @furushchev found a shortcoming of the referenced patch that fixes a synchronization issue, but we sadly can't just revert it.

To sum up:
- Loading the robot model requires the robot_description parameter to be loaded
- The RobotStateVisualization (loaded with the model) provides properties for the individual links that need to exist when the properties are loaded from the config, apparently.

After digging through 100 layers of code, it looks like RViz systematically enables Displays before loading *any* Properties.
The display itself is a Boolean Property whose value represents the enabled state (in addition to the superfluous property "Enabled").
This value is loaded [here](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/display.cpp#L231) and because the display [subscribes to the properties `changed()` signal](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/display.cpp#L70) it calls `onEnabled`.

Incidentally [a few lines later](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/display.cpp#L240) the explicit property `"Enabled"` is checked and would trigger `onEnabled` again (but does not due to an equality check). This second property seems fully redundant to me after spending some time with it in gdb.

I think this is a definite bug in RViz which mixes `onEnable` calls with `load` structure the wrong way around.
From the outside it looks like at some point a patch was merged that disabled calling onEnable twice, but actually disabled the *wrong* onEnable call...
The fix @rhaschke proposed is a simple workaround on MoveIt's side and I added a corresponding patch.

Please verify that this still solves your problem @furushchev .
Also, could you provide a version of this pull-request for the master-branch as well please?
We will not merge it into melodic if it does not arrive in master at the same time.

furushchev:
@v4hn @rhaschke @tylerjw Thank you very much for the feedback and comments!
It is deeper issue than I thought.
I pulled the latest commits and confirmed that the original issue I opened is now resolved with this patch.

@v4hn I'll open the new pull request that is targeted to the main (master) branch.

v4hn:
@rhaschke are you happy(ier) with the current patch?
Please have a look at my rather extended commit message in ca77a99 .

rhaschke:
Thanks, @v4hn, for taking care of this. I didn't had time myself... 
Yes, I am happy with the implemented approach - that's what I had in mind.
I think we can drop the robot_model_loaded_ parameter altogether, because the robot is loaded in onEnable() now...

> I think this is a definite bug in RViz which mixes `onEnable` calls with load structure the wrong way around.

Some other displays might expect that the display was enabled (and initialized) before loading parameters.
In principle, it should work to modify parameters on the fly.
But yes, I stumbled over this order issue already several times in the past.

> From the outside it looks like at some point a patch was merged that disabled calling onEnable twice, but actually disabled the wrong onEnable call...

Could you provide a reference for this patch? Was it reverted or not applied at all?

v4hn:
> Some other displays might expect that the display was enabled (and initialized) before loading parameters.

I understand the conflict. Yes, something might break if it's changed.
Assuming the display is *initialized* is a reasonable assumption, but I don't think it should need to be *enabled*.
After all this is the main difference between initialization and enabling!

> Could you provide a reference for this patch? Was it reverted or not applied at all?

Sorry, I did not investigate enough. Dave relied on "two calls to onEnable" in his original patch and you rejected the approach "because it might change". Skipping through the code in gdb, I could not reproduce a second call anymore, so I assumed it was deliberately changed.
My first candidate for the patch would have been [this sanity-check](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/display.cpp#L258), but it's already really old.

I would expect the two original calls were triggered [here](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/display.cpp#L231) (The display already connects [to the changed() signal](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/display.cpp#L70) in the constructor) and [here](https://github.com/ros-visualization/rviz/blob/melodic-devel/src/rviz/display.cpp#L240) right below. At the moment the whole "Enabled" property and this second check is redundant, but it would actually do the correct thing if the signal would not be connected for the first `BoolProperty::load(config)`.

v4hn:
> I think we can drop the robot_model_loaded_ parameter altogether, because the robot is loaded in onEnable() now...

I don't think so, because [reset() clears `*robot_`](https://github.com/ros-planning/moveit/pull/2223/files#diff-8b87682e810c000fb0509f875f971979R118)  (not the pointer `robot_`), and `update` would break [without the additional check](https://github.com/ros-planning/moveit/pull/2223/files#diff-8b87682e810c000fb0509f875f971979R415).
Unless, of course, RViz ensures these calls cannot happen in parallel. Feel free to propose a different solution, but otherwise this looks reasonable to me right now and I would merge it.

rhaschke:
> > I think we can drop the robot_model_loaded_ parameter altogether, because the robot is loaded in onEnable() now...
> 
> I don't think so, because [reset() clears `*robot_`](https://github.com/ros-planning/moveit/pull/2223/files#diff-8b87682e810c000fb0509f875f971979R118), and `update` would break [without the additional check](https://github.com/ros-planning/moveit/pull/2223/files#diff-8b87682e810c000fb0509f875f971979R415).

`reset` and `update` cannot happen in parallel: They (should) be both triggered by user actions and thus execute in the main GUI thread. I thoroughly tested #2244 and #2245, resetting and disabling/enabling the displays in virtually any situation and it works now as expected. 
I'm closing this PR in favour of #2244 and #2245. @furushchev, thanks for pointing out the issue and proposing an initial solution.

PR #2219, opened 2020-07-20
WIP: OMPL Constrained Planning (FINAL, merge last)

JeroenDM:
### Description

In the context of a [Google Summer of Code project](https://github.com/ros-planning/moveit/issues/2092), I would like to add the constrained planning capabilities of OMPL to the existing interface. Because it involves a significant amount of change, I have a plan to divide it into several smaller pull requests.

### Why
It is one of the projects on the [future projects page](https://moveit.ros.org/documentation/contributing/future_projects/).

It can solve the long-standing [joint flip bug](https://github.com/ros-planning/moveit/issues/562) when planning with path constraints. (Although it does not solve the problem with the `PoseModelStateSpace`, it introduces a new state space that does not have this problem and is potentially faster than the `enforce_joint_model_state_space` hack. But I have not benchmarked it yet.)

(TODO insert some gifs of the planning case examples I have?)

I also started working on a tutorial that shows how to use the new capabilities in parallel with coding, [here](https://github.com/JeroenDM/moveit_tutorials/commit/edcbb9ec771cad34d23a65d3b7dadadf74c6ec33).

#### Phase 1: Cleanup (optional)

- [Add named logging with LOGNAME to OMPL interface](#2211)
- [Remove dead code from ompl interface (related to subspaces)](#2214 )
- And possible [remove factory abstraction to create state spaces](https://github.com/JeroenDM/moveit/pull/2)?

#### Phase 2: Small independent additions of functionality (required)

- [OMPL Constrained Planning (Part 1)](#2216): add the option to disable path constraint checking in StateValidityChecker.
- [OMPL Constrained Planning (Part 2)](#2218): add a new state space that inherits from ModelBasedStateSpace to handles the StateType of a `ompl::base::ConstrainedStateSace`.
- [OMPL Constrained Planning (Part 2)]() (no pull request yet): add different specializations of `ompl::base::Constraint` to handle position constraints, orientation constraints, ...

#### Phase 3: Make it functional

Do the last required changes in this pull request. For a preview of my work in progress, see [this pull request](https://github.com/JeroenDM/moveit/pull/1) on my personal fork.

### Checklist
- [ ] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)
[...]

Issue #2092, opened 2020-05-17
GSoC: Motion planning with general end-effector constraints in MoveIt


JeroenDM:
**mentors**: @mamoll @ommmid @felixvd 

This summer I would like to add collision-free motion planning with general end-effector constraints to MoveIt. Application examples include robot arc welding, grinding, and painting. The original discussion on discourse can be found [here](https://discourse.ros.org/t/gsoc-2020-cartesian-planning-some-trajopt/13045). I probed for interest in other issues [here](https://github.com/ros-planning/moveit/issues/1969) and [here](https://github.com/ros-planning/moveit/issues/467). The goal of this issue is to log and discuss my progress in the coming months.

## Initial project description
Many industrial robot applications specify a Cartesian end-effector path, that could be under-constrained. The MoveIt’s RobotState function, “computeCartesianPath”, fulfills only a small part of this demand. OMPL has advanced capabilities to handle constraints that are not integrated with MoveIt yet. Examples of applications that can benefit from this integration are robot welding, grinding, painting, and many more.

Concretely this would mean integrating OMPL’s Cartesian planning capabilities in the existing planning plugin. If time allows it, I would like to add support for Cartesian path constraints in TrajOpt. All deliverables should include documentation, tutorials, and tests.

## Planning
**Phase 1: Example cases and interface discussion (June)**
- Find/create example cases that use external planners to solve the planning problem and showcase what will be possible with MoveIt at the end of the project. An example could be the puzzle piece demo from [trajot_ros](https://github.com/ros-industrial-consortium/trajopt_ros) or a welding example such as the last one in [hybrid_planning_examples](https://github.com/Jmeyer1292/hybrid_planning_experiments).
- The current interface to specify path constraints in Moveit is not entirely clear, I suspect it will take some time to find a common interface to specify these planning problems.
- As a learning experience, I could refactor the current “computeCartesianPath” as a MoveIt planning plugin, as discussed [here](https://github.com/ros-planning/moveit/issues/797).

**Phase 2: Implementation (July)**
- Start implementing the features that we agreed upon in the first month. This will most likely mean working in the code of the current OMPL interface. I don’t know the exact approach yet, as explained below in the comments from the kick-off meeting.
- @v4hn also pointed out that [the broken pose_model_state_space](https://github.com/ros-planning/moveit/tree/master/moveit_planners/ompl/ompl_interface/src/parameterization/work_space) should be replaced. (I did not look into the details of this yet.)

[...]

PR #2224, opened 2020-07-22
WIP: OMPL Constrained Planning (Part 3) 

JeroenDM:
### Description

**This is only intended to be merged after discussing the overall approach in the main discussion thread: #2219.**

This pull request is part of a larger effort to support [OMPL's Constrained planning capabilities](http://ompl.kavrakilab.org/constrainedPlanning.html) in MoveIt.

To integrate OMPL's constrained state space, we need a model of the constraints that inherits from [ompl::base::Constraint](http://ompl.kavrakilab.org/classompl_1_1base_1_1Constraint.html). In general, this can be any equality constraint. However, many interesting constraints are **not** equality constraints (==), but inequality constraints (< >). This pull request provides a base class `BaseConstraint` and a class `Bounds` that together model bounds on a scalar variable as an equality constraint. That is, a generic vector-valued function `F(joint_positions) = 0`. For a more detailed description, see [this pdf document](https://github.com/ros-planning/moveit/files/4960396/constraints_model.pdf).

Using this base class, two specific types of constraints are implemented:

- **Position constraints**: specify constraints on the position of the link by passing in a primitive shape of type `BOX` into the motion planning request. A specific dimension can be left unconstrained by setting it to `-1`.
- **Orientation constraints**: this turned out to be a rabbit hole of different possibilities. For the first version, I've chosen to adopt the orientation error used in [TrajOpt](https://github.com/ros-industrial-consortium/trajopt_ros), using exponential coordinates, because this seemed to perform reasonable, in contrast to roll-pitch-yaw angles.

### Examples
**Position constraints**: Using a very slender box to get line following.
kuka_boxes2

**Orientation constraints** (slightly cheating / future work, as this uses a [specific implementation](https://github.com/JeroenDM/elion/commit/3b981fc8d3ac007aeef40bdfae6e3ced9813b83a) of orientation constraints in combination with `AtlasStateSpace` and `RRTstar`.)
kuka_aa_con_atlas_prmstar

**Position and orientation constraints** (I'll be adding the code for this soon)
welding_example_collision

**Blooper** Orientation constrained sometimes behaves strangely, especially with collision objects close to the path and using roll-pitch-yaw angles for orientation error. I left out the robot for artistic purposes.)
art1

### Future possibilities

Combine position and orientation constraints. This is in progress, I just have to push the code to this branch and add some tests.

To create an equality constraint on a link's position you can set a dimension of the box to a very small value. However, it will be far more efficient to bypass the `Bounds` for equality constraints and implement them directly in a new class that inherits from `ompl::base::Constraint`. I [have done this](https://github.com/JeroenDM/moveit/blob/39cbff77b5ccd62c6630dfc1e73718d1256d3187/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraint.h#L222) for a specific example, but have not found a clean way to integrate it yet. The logic to select the right constraint model could become overly complex.

In general, it would be cool if we could pass any user-defined constraint to OMPL through some C++ interface but gets us too far off track for now.

The current constrained model only supports OMPL's `ProjectedStateSpace` and not `AtlasStateSpace` or `TangentBundleStateSpace`. This is an issue that is probably not trivial to solve for now. (If you're interested, you can search for the word "atlas" in my GSoC thread #2092 for more info.)

Different ways to express orientation error could be more appropriate for certain types of applications. MoveIt already has the functionality to calculate the Jacobian for unit Quaternions, so this could be straightforward to add.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
@@ -1,0 +1,264 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, Jeroen De Maeyer
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of Willow Garage nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Jeroen De Maeyer */
+
+#pragma once
+
+#include <ompl/base/Constraint.h>
+
+#include <moveit/robot_model/robot_model.h>
+#include <moveit/robot_state/robot_state.h>
+#include <moveit_msgs/Constraints.h>
+
+namespace ompl_interface
+{
+namespace ob = ompl::base;
+
+/** \brief Represents upper and lower bound on a scalar value (double).
+ *
+ * Equality constraints can be represented by setting
+ * the upper bound and lower bound almost equal.
+ * I (jeroendm) assume that it is better to not have them exactly equal
+ * for numerical reasons. Not sure.
+ * **/
+struct Bounds
+{
+  double lower, upper;
+
+  /** \brief Distance to region inside bounds
+   *
+   * Distance of a given value outside the bounds,
+   * zero inside the bounds.
+   *
+   * Creates a penalty function that looks like this:
+   *
+   *  \         /
+   *   \       /
+   *    \_____/
+   * (how does ascii art work??)
+   * */
+  double distance(double value) const;
+};
+
+/** \brief Abstract base class for differen types of constraints, implementations of ompl::base::Constraint
+ *
+ * To create a constrained state space in OMPL, we need a model of the constraints.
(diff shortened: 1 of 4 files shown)


JeroenDM:
Note to self: I could avoid having to call the init function after the construction of a constraint class. See p.50 in "Effective C++ Third Edition, Scott Meyers". Although it would not be that nice, as we need static methods in all derived classes for `bounds_`, `target_position_`, `target_orientation_`, and `link_name_`. These methods would then be called in the derived class constructor to call the base class constructor that takes all these parameters as input.

JeroenDM:
I added [EIGEN_MAKE_ALIGNED_OPERATOR_NEW](https://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html). It not being there could have been responsible for some of the stranger segmentation faults I got, but I can't point to a specific case where this caused an issue.

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -0,0 +1,349 @@
...
+  // put the exponential coordinates in a skew symmetric matrix
+  Eigen::Matrix3d r_skew;
+  r_skew << 0, -r[2], r[1], r[2], 0, -r[0], -r[1], r[0], 0;
+
+  // calculate the absolute value of the rotation angle as an intermediate value for the complex expression below
+  double t{ std::abs(angle) };
+
+  // calculate to 3x3 conversion matrix to convert an angular velocity into exponential coordinates
+  return Eigen::Matrix3d::Identity() - 0.5 * r_skew +
+         r_skew * r_skew / (t * t) * (1 - 0.5 * t * std::sin(t) / (1 - std::cos(t)));
```
ommmid, line 347: Can you mention a reference for this equation?

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
```diff
@@ -0,0 +1,293 @@
...
+                                                 const moveit_msgs::Constraints& constraints);
+
+/** \brief Conversion matrix to go from angular velocity in the world frame to
+ * angle axis equivalent.
+ *
+ * Based on:
+ * https://ethz.ch/content/dam/ethz/special-interest/mavt/robotics-n-intelligent-systems/rsl-dam/documents/RobotDynamics2016/RD2016script.pdf
+ *
+ * */
+Eigen::Matrix3d angularVelocityToAngleAxis(double angle, const Eigen::Vector3d& axis);
```
JeroenDM, line 291: I should change the name, as it is not converting to Angle-Axis representation, but to exponential coordinates. Although the difference is really minimal.

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -0,0 +1,351 @@
...
+#include <moveit/robot_model/robot_model.h>
+#include <moveit/robot_state/robot_state.h>
+#include <moveit_msgs/Constraints.h>
+#include <eigen_conversions/eigen_msg.h>
+
+namespace ompl_interface
+{
+constexpr char LOGNAME[] = "ompl_constraint";
+
+double Bounds::distance(double value) const
```
JeroenDM, line 48: I will be removing all the code related to `Bounds` in the first version and focus on simple equality constraints.

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
```diff
@@ -0,0 +1,293 @@
...
+  virtual Eigen::MatrixXd calcErrorJacobian(const Eigen::Ref<const Eigen::VectorXd>& x) const override;
+};
+
+/** \brief Position and orientation constraint
+ *
+ * Combine position constraints modelled as a box with orientation constraints
+ * using exponential coordinates, but ignore the last rotation around the z-axis.
+ * This is a common constraint for grasping, welding, ...
+ * */
+class PoseConstraint : public BaseConstraint
```
JeroenDM, line 236: I will change this base class to handle simple equality constraints for the first version. Then I can add this current base class as a new derived class that handles bounds, later on, something like `BoundsConstraint` or something.

moveit_planners/ompl/ompl_interface/src/detail/ompl_constraints.cpp
```diff
@@ -0,0 +1,351 @@
...
+  {
+    ROS_ERROR_NAMED(LOGNAME, "No path constraints found in planning request.");
+    return nullptr;
+  }
+}
+
+/******************************************
+ * Angular velocity conversion
+ * ****************************************/
+Eigen::Matrix3d angularVelocityToAngleAxis(double angle, const Eigen::Vector3d& axis)
```
JeroenDM, line 332: @mamoll I suspect this is another bottleneck, runtime wise. (Similar to the one discussed in the meeting yesterday in `calcError`.) But I did not measure it. Based on running some examples and a rough estimate, it seemed the have a similar runtime to using the default numerical Jacobian for the constraints.

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/detail/ompl_constraints.h
ommmid: Yes, I would change the name. The output is not exactly the same as what the name suggest.

PR #2226, opened 2020-07-28
OMPL interface: add documentation

JeroenDM:
### Description

The `ompl_interface::ModelBasedPlanningContext` has two methods with the same name `getPlanningContext`. I think this is confusing because they do different things. Especially given that the `planning_interface::PlannerManager` also has [two of such methods](https://github.com/ros-planning/moveit/blob/dc909df5d2cf93868a1773db869e6767e339cf7a/moveit_core/planning_interface/include/moveit/planning_interface/planning_interface.h#L182), where overloading is used for a completely different purpose (both versions are public).

I also added docstrings to the header, to hopefully clarify a bit was is going on in these methods.

An alternative name could be `constructPlanningContext`.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/planning_context_manager.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/planning_context_manager.h
@@ -170,6 +170,15 @@ public:
     return robot_model_;
   }
 
+  /** \brief Returns a planning context to OMPLInterface, which in turn passes it to OMPLPlannerManager.
+   *
+   * This function checks the input and reads planner specific configurations.
+   * Then it creates the planning context with PlanningContextManager::createPlanningContext.
+   * Finally, it puts the context into a state appropriate for planning.
+   * This last step involves setting the start, goal, and state validity checker using the method
+   * ModelBasedPlanningContext::configure.
+   *
+   * */
   ModelBasedPlanningContextPtr getPlanningContext(const planning_scene::PlanningSceneConstPtr& planning_scene,
                                                   const planning_interface::MotionPlanRequest& req,
                                                   moveit_msgs::MoveItErrorCodes& error_code, const ros::NodeHandle& nh,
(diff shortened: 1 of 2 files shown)


JeroenDM:
I agree that the name change does not due much. It a small thing, left from spending too much time thinking about how these methods could be simplified or clarified. I'll just have to accept that some systems are complex (al least to me) and I just have to deal with it :)

v4hn:
> I'll just have to accept that some systems are complex (al least to me) and I just have to deal with it :)

If things are complex, simplify them. People spent way more time on adding complexity to MoveIt than on reducing it. :+1:
PS: The reason why reviews in some parts of MoveIt take forever is because the maintainers also think these parts are too complex for them. ;)

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/planning_context_manager.h
```diff
@@ -209,9 +218,9 @@ class PlanningContextManager
   void registerPlannerAllocatorHelper(const std::string& planner_id);
 
   /** \brief This is the function that constructs new planning contexts if no previous ones exist that are suitable */
-  ModelBasedPlanningContextPtr getPlanningContext(const planning_interface::PlannerConfigurationSettings& config,
-                                                  const StateSpaceFactoryTypeSelector& factory_selector,
-                                                  const moveit_msgs::MotionPlanRequest& req) const;
+  ModelBasedPlanningContextPtr createPlanningContext(const planning_interface::PlannerConfigurationSettings& config,
```
mamoll, line 221: Why the name change? This method will actually return a planning context from a cache if one is available. In that case no new planning context is created.

JeroenDM: The name change probably only made sense to me personally, I'll revert it.

I think it can be confusing that both methods do part of the construction of the planning context. For examples, after the context is created, or a cached context is retrieved, `context.clear()` is called. And both methods call a bunch of `context->setXXX(..)` methods.

And then as a cherry on top of the cake, `context->configure()` is called to do even more configuration :)

I think it can be confusing to have to the configuration of the context so spread out in the code base, without clear names indications what is done where. But I don't have a solution...

v4hn:
> I think it can be confusing that both methods do part of the construction of the planning context. [...]

I do think the method rename makes sense because of that.
There is a difference between constructing a context and configuring it.

Issue #2170, opened 2020-06-26
Include What You Use


v4hn:
For fun and profit I just ran [include-what-you-use](https://include-what-you-use.org/) on our code base and put the results [here](https://tams.informatik.uni-hamburg.de/people/goerner/files/iwyu-moveit.log). I wanted to make it a gist, but it seem 23k lines is too much for that :)

Please have a look and tell me what you think of incorporating these changes (partially?) into MoveIt.

The idea is
1. to reduce compile time by dropping unnecessary (or unnecessarily general) includes and
2. improve correctness by not relying on transitive includes

@JafarAbdi @henningkayser @tylerjw @rhaschke and whoever else feels responsible :-)

henningkayser:
Nice! I really like all the added forward declarations and that it checks for transitive MoveIt includes. All those extra Eigen/boost/std includes seem a little excessive, though. What's the performance implication of that? I think it would be really helpful to get a less-restrictive check added to moveit_ci.

v4hn:
> I really like all the added forward declarations and that it checks for transitive MoveIt includes.

Same here. That's what made me look into it.

> All those extra Eigen/boost/std includes seem a little excessive, though.

They are, in theory, not excessive but just explicit. At the moment all of them are probably included transitively, but directly used in the respective sources.

> What's the performance implication of that?

Not sure. I did not look into automatically applying the changes.
I just wanted to get the discussion started here and maybe Jafar (or anyone else really) could have a look once some maintainers gave their opinion about this.
After all it is also a general question if we, e.g., are willing to go away from `Eigen/Dense` includes in favor of deep includes for all the things we actually use.

> I think it would be really helpful to get a less-restrictive check added to moveit_ci.

This would likely involve a lot of NOLINT pragmas to avoid excessive complains about petty stuff. I don't see that for now.

rhaschke:
In general, I like the idea to check (and auto-fix) our includes in this fashion. But I agree with Henning, that deep includes for boost or Eigen stuff is counter-productive. These meta includes were introduced for convenience and I remember that you added a similar meta include for MTC recently (stages.h?).
I'm curious how the tool finds missing includes...

v4hn:
> In general, I like the idea to check (and auto-fix) our includes in this fashion.

As I said, automating this might be a lot of work with unstable results. So I don't think that's a good project for now.

> But I agree with Henning, that deep includes for boost or Eigen stuff is counter-productive.

[...]

PR #2229, opened 2020-07-30
Optimize includes

mvieth:
### Description

Remove most obvious unused includes, as recommended by include-what-you-use (see https://github.com/ros-planning/moveit/issues/2170)

### Checklist

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/world.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/world.h
@@ -41,7 +41,6 @@
 #include <string>
 #include <vector>
 #include <map>
-#include <memory>
 #include <boost/function.hpp>
 #include <Eigen/Geometry>
 #include <eigen_stl_containers/eigen_stl_vector_container.h>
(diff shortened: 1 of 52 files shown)


mvieth:
> Thanks! Great to see cleanup-ideas being picked up!

You are very welcome!

> Did you look through the logs I posted or generate them from scratch?

I used your logs. Regarding that: how did you generate them? Did you set the option in CMakeLists.txt? Did you use the `--check_also` switch? Because include-what-you-use [does not always check all header files](https://github.com/include-what-you-use/include-what-you-use#plugging-into-make).

> Should we expect more such patches in the future or is this everything "obvious" from the include-what-you-use approach?

Well, yes, it is the most obvious stuff. But there is more that could be done:
- So far, I did not replace includes with forward-declares. You would have to decide whether you [find them desirable or not](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md#why-forward-declare)
- As I mentioned above, some header files might not have been checked. If any unused includes could be removed from them, that could have a major impact
- Some Boost includes could be replaced with more specific ones (as long as they are meant to be included by the user as per Boost docs). For Eigen, no deep includes should be used as you noted in a comment in the issue. The only possibility would be to replace `<Eigen/Eigen>` and `<Eigen/Dense>` with more specific includes, which I partly already did. For ROS, `<ros/ros.h>` could be replaced with more specific includes like `<ros/node_handle.h>`. I think that would be ok, what do you think?
- I also so far did not add any includes to remove transitive includes, because there would be **many** new includes. I am not sure how big of a problem these transitive includes are, anyway
- And some more misc stuff

PR #2231, opened 2020-08-01
Added missing variants of place from list of PlaceLocations and Poses

gerardcanal:
### Description

The Python interface for move_group was missing the place options where a list of possible places is given either as PlaceLocations or Poses.

I tested it in Melodic (would be good to have it ported there too).

### Checklist


moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -571,7 +571,7 @@ def pick(self, object_name, grasp=[], plan_only=False):
     def place(self, object_name, location=None, plan_only=False):
         """Place the named object at a particular location in the environment or somewhere safe in the world if location is not provided"""
         result = False
-        if location is None:
+        if not location:
```
tylerjw, line 574: This change makes this true with an empty list.  Is that what you wanted here?

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -582,8 +582,16 @@ def place(self, object_name, location=None, plan_only=False):
             result = self._g.place(object_name, conversions.pose_to_list(location), plan_only)
         elif type(location) is PlaceLocation:
             result = self._g.place(object_name, conversions.msg_to_string(location), plan_only)
+        elif type(location) is list:
```
tylerjw, line 585: Would this also work for a tuple type?

moveit_commander/src/moveit_commander/move_group.py
gerardcanal: Yes, that was intended as I later need to check the type of objects inside the list. To do so I take the first element to check the type of objects inside the list (with the assumption that it will be a homogeneous list of objects of the same type). 
Therefore, it's safe to check for location[0] as if the list was empty it would have entered through this first if.

moveit_commander/src/moveit_commander/move_group.py
gerardcanal: It won't as it is right now. To accept tuple types we should add the check there as `elif type(location) is list or type(location) is tuple`. I didn't think about the option of adding a tuple of locations instead of a list, but we can check that too if needed.

gerardcanal:
I wanted to add a comment regarding the pick/place interfaces in Python (maybe I should open an issue for this?). After looking a bit at them, I think it may be worth to reimplement them natively in Python, given that the C++ versions that are called mainly perform a call to the actionlib servers for move_group/{pick/place}.
It seems that calling pick() from python blocks the whole node (therefore the rest of services/topics are not handled even when rospy runs them in different threads). I assume it's because of the python/c++ bridge. This may be the case with other calls from Pyhton, but so far I only used the pick and place.
In my case, this was preventing any possibility of stopping the execution of a trajectory.

felixvd:
They are slated to be deprecated in favor of the MoveIt Task Constructor capability, so that would probably be effort better spent elsewhere.

gerardcanal:
> They are slated to be deprecated in favor of the MoveIt Task Constructor capability, so that would probably be effort better spent elsewhere.

That makes sense then, thank you for the update!

v4hn:
> They are slated to be deprecated in favor of the MoveIt Task Constructor capability, so that would probably be effort better spent elsewhere.

I'm not sure I agree. It is very much an option to have the MTC capability expose the exact same action server API.
It's just that someone has to actually do that and contribute it back to us. An interface for a more general MTC capability would have to look very different of course.

gerardcanal:
In such case and if you think it can be helpful I can open a new PR to call the pick and place through the actionlib as it is done in the C++ interface (I already kind of implemented it in my node, so it should be easy to integrate back to moveit).

v4hn:
Actually, your reported hangup is quite a straight-forward issue in our code base:

The python wrapper methods use a [GILReleaser pattern](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp#L436) to let python run other code while computing/waiting in C++-only code. But this pattern was not implemented for the pick/place methods. Could you provide a pull-request that adds this where needed?

gerardcanal:
Sure, this one looks quite easy! I am not very familiar with mixing python with c++ so I was not aware of this.
I will now test it and add it to this PR if that sounds good.

gerardcanal:
I added the GILReleaser. Some tests seem to have failed in noetic though I can't see why.

v4hn:
Aside from the possibly unrelated CI failure this looks good to me.

@gleichdick Could you comment on 3a5025072cd865619ced7ea44e6942cd130c8b8d please?
Did you forget to add the releaser to these methods or was there some reason to leave them out?

felixvd:
> I'm not sure I agree. It is very much an option to have the MTC capability expose the exact same action server API.
> It's just that someone has to actually do that and contribute it back to us. An interface for a more general MTC capability would have to look very different of course.

My bad, I read too quickly and overlooked that this only concerns an action call, not additional development to the pick/plan pipeline. Making the action call Python-native would not be difficult, but using the `ActionClient` already created by the `move_group` is probably more efficient.

> The python wrapper methods use a [GILReleaser pattern](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp#L436) to let python run other code while computing/waiting in C++-only code. But this pattern was not implemented for the pick/place methods. Could you provide a pull-request that adds this where needed?

Interesting. I never even noticed the pattern, since it only appears in methods I haven't touched. If @gleichdick also left it out because they didn't know about it, could we add a comment to the GILRelease lines? I think would be helpful for everyone unfamiliar with binding Python and C++ code.

gleichdick:
@v4hn the commit looks good to me.
In #1947 I was being asked not to use the GILReleaser in every wrapped function because of the cost of saving the thread state. So I used them in just a few functions.

gerardcanal:
Thanks @gleichdick, I guess it makes sense to not have it in all of them. 

Is there's something else that shall to be done for this PR?

gerardcanal:
Kind ping for @v4hn to check if he can review this PR :-)

PR #2177, opened 2020-06-30
[Discussion] clang-format-10

rhaschke:
Adding to #2169, I tweaked the clang-format config even more.

PR #2235, opened 2020-08-05
[maint] clang-format-10

tylerjw:
This is a rebase of #2177 which itself was based on #2169 

clang-format-10 is available via apt on ubuntu 18.04 and 20.04.  it is also possible to install on 16.04 using the instructions from the clang project itself.  This is blocking work on moveit2.  Please consider this quickly so we can get it done before it needs another rebase.

Co-authored-by: Robert Haschke <rhaschke@techfak.uni-bielefeld.de>
Co-authored-by: Henning Kayser <henningkayser@picknik.ai>

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,66 +1,75 @@
 ---
 BasedOnStyle:  Google
+ColumnLimit:    120
+MaxEmptyLinesToKeep: 1
+SortIncludes: false
+
+Standard:        Auto
+IndentWidth:     2
+TabWidth:        2
+UseTab:          Never
 AccessModifierOffset: -2
 ConstructorInitializerIndentWidth: 2
+NamespaceIndentation: None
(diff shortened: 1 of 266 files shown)


tylerjw:
This is failing CI because moveit_ci doesn't use clang-format-10.  I'll make a PR to fix that.

henningkayser:
> This is failing CI because moveit_ci doesn't use clang-format-10. I'll make a PR to fix that.

Just wanted to mention that. I think we should also keep legacy support for clang-format-3.9 to not break too many setups at once.

rhaschke:
I don't like that you squashed the 4 commits of #2177 / #2169 into a single commit. I think, adapting and applying the config should be two separate commits.

rhaschke:
Continuing with #2177.

PR #2239, opened 2020-08-06
fix memcpy bug in copyJointToOMPLState in ompl interface

JeroenDM:
### Description

A single line bug fix (that was punching far above its weight class in terms of causing debugging fun.)

Example of the issue in [an online C++ shell](http://cpp.sh/6snnu).
Previous [related discussion](https://github.com/JeroenDM/moveit/pull/3) with @tylerjw, where he created the example that I use in the C++ shell link above.

### What (I think) happens in this piece of code

The joint data for a specific joint is copied from MoveIt to OMPL, it looks something like this:

Robot model memory (array of doubles), the bold values are copied.
| x1 | x2 | ... | ... **| y1 | y2 | y3|** ... | ... | ... | z1 | .. |

Copies into OMPL state memory (also an array of doubles)
| x1 | x2 **| y1 | y2 | y3 |** z1 |

The | ... | represent joint values for fixed joints (or mimic joints?) that are not relevant for OMPL.

### Test
I extended an existing test in `test_state_space.cpp` to make it fail without this fix. I mirrored the approach of the other tests. But maybe it is not the best way to test this. If anyone agrees, I can add another test that just copies fixed, known joint values.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/parameterization/model_based_state_space.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/parameterization/model_based_state_space.cpp
@@ -340,7 +340,7 @@ void ompl_interface::ModelBasedStateSpace::copyJointToOMPLState(ompl::base::Stat
 {
   // Copy one joint (multiple variables possibly)
   memcpy(getValueAddressAtIndex(state, ompl_state_joint_index),
-         robot_state.getVariablePositions() + joint_model->getFirstVariableIndex() * sizeof(double),
+         robot_state.getVariablePositions() + joint_model->getFirstVariableIndex(),
          joint_model->getVariableCount() * sizeof(double));
 
   // clear any cached info (such as validity known or not)
--- a/moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
+++ b/moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
@@ -167,6 +167,34 @@ TEST_F(LoadPlanningModelsPr2, StateSpaceCopy)
     EXPECT_TRUE(robot_state.distance(robot_state2) < 1e-12);
   }
 
+  // repeat the above test with a different method to copy the state to OMPL
+  for (int i = 0; i < 10; ++i)
+  {
+    // create two equal states
+    moveit::core::RobotState robot_state2(robot_state);
+    EXPECT_TRUE(robot_state.distance(robot_state2) < 1e-12);
+
+    // copy the first state to OMPL as backup (this is where the 'different' method comes into play)
+    const moveit::core::JointModelGroup* joint_model_group =
+        robot_state.getRobotModel()->getJointModelGroup(ss.getJointModelGroupName());
+    std::vector<std::string> joint_model_names = joint_model_group->getActiveJointModelNames();
+    for (std::size_t joint_index = 0; joint_index < joint_model_group->getVariableCount(); ++joint_index)
+    {
+      const moveit::core::JointModel* joint_model = joint_model_group->getJointModel(joint_model_names[joint_index]);
+      EXPECT_TRUE(joint_model != nullptr);
+      ss.copyJointToOMPLState(state, robot_state, joint_model, joint_index);
+    }
+
+    // and change the joint values of the moveit state, so it is different that robot_state2
+    robot_state.setToRandomPositions(robot_state.getRobotModel()->getJointModelGroup(ss.getJointModelGroupName()));
+    EXPECT_TRUE(robot_state.distance(robot_state2) > 1e-12);
+
+    // copy the backup values in the OMPL state back to the first state,
+    // and check if it is still equal to the second
+    ss.copyToRobotState(robot_state, state);
+    EXPECT_TRUE(robot_state.distance(robot_state2) < 1e-12);
+  }
+
   ss.freeState(state);
 }
 
--- a/moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
+++ b/moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
@@ -172,7 +172,7 @@ TEST_F(LoadPlanningModelsPr2, StateSpaceCopy)
   {
     // create two equal states
     moveit::core::RobotState robot_state2(robot_state);
-    EXPECT_TRUE(robot_state.distance(robot_state2) < 1e-12);
+    EXPECT_LT(robot_state.distance(robot_state2), 1e-12);
 
     // copy the first state to OMPL as backup (this is where the 'different' method comes into play)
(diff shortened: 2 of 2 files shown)


moveit_planners/ompl/ompl_interface/src/parameterization/model_based_state_space.cpp
```diff
@@ -340,7 +340,7 @@ void ompl_interface::ModelBasedStateSpace::copyJointToOMPLState(ompl::base::Stat
 {
   // Copy one joint (multiple variables possibly)
   memcpy(getValueAddressAtIndex(state, ompl_state_joint_index),
-         robot_state.getVariablePositions() + joint_model->getFirstVariableIndex() * sizeof(double),
+         robot_state.getVariablePositions() + joint_model->getFirstVariableIndex(),
```
tylerjw, line 343: Good catch, when you use `operator+` on a pointer it advances based on the type of the pointer.  Since getVariablePositions returns a double you want to advance the pointer by amounts of `sizeof(double)` not `sizeof(double) * sizeof(double)`.

moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
```diff
@@ -167,6 +167,34 @@ TEST_F(LoadPlanningModelsPr2, StateSpaceCopy)
     EXPECT_TRUE(robot_state.distance(robot_state2) < 1e-12);
   }
 
+  // repeat the above test with a different method to copy the state to OMPL
+  for (int i = 0; i < 10; ++i)
+  {
+    // create two equal states
+    moveit::core::RobotState robot_state2(robot_state);
+    EXPECT_TRUE(robot_state.distance(robot_state2) < 1e-12);
```
tylerjw, line 175:
```suggestion
    EXPECT_LT(robot_state.distance(robot_state2), 1e-12);
```

I see that the old code in this file also does the comparison this way.  The advantage of using `EXPECT_LT` is that it will print the value when it fails instead of just saying that it was false.  Also, since 1e-12 seems to be being used as an epsilon you could put a constant at the top of the file and replace all the uses with the constant.

This is more of just a nit as you are following the style already in this test.

moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
```diff
@@ -167,6 +167,34 @@ TEST_F(LoadPlanningModelsPr2, StateSpaceCopy)
...
+
+    // copy the first state to OMPL as backup (this is where the 'different' method comes into play)
+    const moveit::core::JointModelGroup* joint_model_group =
+        robot_state.getRobotModel()->getJointModelGroup(ss.getJointModelGroupName());
+    std::vector<std::string> joint_model_names = joint_model_group->getActiveJointModelNames();
+    for (std::size_t joint_index = 0; joint_index < joint_model_group->getVariableCount(); ++joint_index)
+    {
+      const moveit::core::JointModel* joint_model = joint_model_group->getJointModel(joint_model_names[joint_index]);
+      EXPECT_TRUE(joint_model != nullptr);
```
tylerjw, line 184:
```suggestion
      EXPECT_NE(joint_model, nullptr);
```

moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
```diff
@@ -167,6 +167,34 @@ TEST_F(LoadPlanningModelsPr2, StateSpaceCopy)
...
+      ss.copyJointToOMPLState(state, robot_state, joint_model, joint_index);
+    }
+
+    // and change the joint values of the moveit state, so it is different that robot_state2
+    robot_state.setToRandomPositions(robot_state.getRobotModel()->getJointModelGroup(ss.getJointModelGroupName()));
+    EXPECT_TRUE(robot_state.distance(robot_state2) > 1e-12);
```
tylerjw, line 190:
```suggestion
    EXPECT_GT(robot_state.distance(robot_state2), 1e-12);
```

moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
```diff
@@ -167,6 +167,34 @@ TEST_F(LoadPlanningModelsPr2, StateSpaceCopy)
...
+
+    // copy the backup values in the OMPL state back to the first state,
+    // and check if it is still equal to the second
+    ss.copyToRobotState(robot_state, state);
+    EXPECT_TRUE(robot_state.distance(robot_state2) < 1e-12);
```
tylerjw, line 195:
```suggestion
    EXPECT_LT(robot_state.distance(robot_state2), 1e-12);
```

moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
```diff
@@ -167,6 +167,34 @@ TEST_F(LoadPlanningModelsPr2, StateSpaceCopy)
...
+      ss.copyJointToOMPLState(state, robot_state, joint_model, joint_index);
+    }
+
+    // and change the joint values of the moveit state, so it is different that robot_state2
+    robot_state.setToRandomPositions(robot_state.getRobotModel()->getJointModelGroup(ss.getJointModelGroupName()));
+    EXPECT_TRUE(robot_state.distance(robot_state2) > 1e-12);
+
+    // copy the backup values in the OMPL state back to the first state,
+    // and check if it is still equal to the second
+    ss.copyToRobotState(robot_state, state);
```
tylerjw, line 194: [nit] again, just a style thing, `ss` isn't a nice descriptive name.  You are just using what was already here, so you are welcome to just ignore this comment.

tylerjw:
This is failing CI for unrelated issues with python.

JeroenDM:
@tylerjw Agree to all :) (in the spirit of leaving things better than you found them). Thanks for the review!

Yeah, I went through the CI, but it's not ideal to go through at the moment... :)

moveit_planners/ompl/ompl_interface/test/test_state_space.cpp
JeroenDM: I changed it to `std::numeric_limits<double>::epsilon();`, but used a variable name that is a bit more reasonable in length.

The tests seem to work, but I decreased the test tolerance from `1e-12` to `1e-16` ish. I'm a bit worried this will cause random test failures because the distance calculation probably contains summations. Maybe increase it a bit?

moveit_planners/ompl/ompl_interface/src/parameterization/model_based_state_space.cpp
JeroenDM: Ah, now I really understand it, finally. Thanks!

PR #2246, opened 2020-08-10
A ROS service to reset the Servo status

AndyZe:
We were experiencing an issue where the arm starts in contact with a table. That initial collision needs to be reset before Servo works. This adds a ROS service to perform that reset.

### DIFF
--- a/moveit_ros/moveit_servo/CMakeLists.txt
+++ b/moveit_ros/moveit_servo/CMakeLists.txt
@@ -23,6 +23,7 @@ find_package(catkin REQUIRED COMPONENTS
   rosparam_shortcuts
   sensor_msgs
   std_msgs
+  std_srvs
   trajectory_msgs
 )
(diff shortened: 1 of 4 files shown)


AdamPettinger:
I like this. Also can handle something else I have noticed:

Before this PR, nothing resets the status back to normal (NO_WARNING). Meaning if you ever get into a status, you are stuck there until a different (non-normal) status occurs. 

Example: If you approach a collision and get the DECELERATE_FOR_COLLISION warning, then back away from the collision, you will still be in the DECELERATE_FOR_COLLISION status (and it's published continuously) even though you may now be fine and in a state I think should have status NO_WARNING. 

This PR should at least allow the user to reset that manually. I have thought about resetting the status every cycle after it is published, but this would still cause the problem Andy is having (but would be nicer to the user in situations except for needing to move while in collision)

PR #2249, opened 2020-08-11
[Servo] Update last_sent_command_ at ServoCalcs start

AdamPettinger:
### Description

I moved the initialization of `last_sent_command_` from the ServoCalcs constructor to the `start()` function. The `last_sent_command_` is stored and published if Joint or Cartesian servoing is not done. 

This PR fixes a logic error resulting in bad behavior:

1. Start servo
2. Servo around
3. Stop servo
4. Move robot with something besides servo
5. Start servo
6. `last_sent_command_` is stored as where servo stopped in 3 and if published may result in the manipulator jumping back to that position.

This PR fixes by saving `last_sent_command_` as the current position every time servo is started.

There could still be some unexpected behavior if you start the servo while the arm is moving from another input. This will suddenly stop the arm as the `last_sent_command_` is set to the current position and 0 velocity. I am unsure what **should** happen in this case, but am open to suggestions.


AndyZe:
Thanks for a good description of what's going on. It makes sense; I don't have any suggestion for the edge case you pointed out.

AndyZe:
I just noticed some folders are not being installed. Would you mind adding that here? In CMakeLists, it would be:

```
install(DIRECTORY launch DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
install(DIRECTORY config DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
```

AdamPettinger:
> I just noticed some folders are not being installed. Would you mind adding that here? 

Done!

AndyZe:
@tylerjw you will probably want to merge this before the Noetic release. It adds these key lines:

```
install(DIRECTORY launch DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
install(DIRECTORY config DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
```

gavanderhoorn:
I read the title of this PR as "Update [..] at ServoCakes start" and was rather confused for a while.

:cake:

PR #2260, opened 2020-08-16
Refactor velocity limit enforcement and add a unit test

AndyZe:
The way I was doing acceleration & velocity limit enforcement before had some flaws. It was only applied to one joint at a time, so it could affect the direction of motion of the robot. It was causing some serious direction swings.

This simplifies the whole limit checking, only enforcing velocity limits.

Also renamed the functions from  enforceSRDFAccelVelLimits -> enforceVelLimits. I think this function name was misleading because limits actually get stored in joint_limits.yaml or the URDF, not the SRDF.

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -726,13 +726,11 @@ void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
   // Apply the velocity scaling to all joints
   if (velocity_limit_scaling_factor < 1)
   {
-    joint_delta_index = 0;
-    for (auto joint : joint_model_group_->getActiveJointModels())
+    for (joint_delta_index = 0; joint_delta_index < joint_model_group_->getActiveJointModels().size();
+         ++joint_delta_index)
     {
       delta_theta(joint_delta_index) = velocity_limit_scaling_factor * delta_theta(joint_delta_index);
       velocity(joint_delta_index) = velocity_limit_scaling_factor * velocity(joint_delta_index);
-
-      ++joint_delta_index;
     }
   }
 }
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -685,7 +685,7 @@ void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
 
   std::size_t joint_delta_index = 0;
   // Track the smallest velocity scaling factor required, across all joints
-  double min_relative_velocity_all_joints = 1;
+  double velocity_limit_scaling_factor = 1;
 
   for (auto joint : joint_model_group_->getActiveJointModels())
   {
@@ -712,25 +712,25 @@ void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
       // Apply velocity bounds
       if (clip_velocity)
       {
-        const double relative_change =
+        const double scaling_factor =
             fabs(velocity_limit * parameters_.publish_period) / fabs(delta_theta(joint_delta_index));
 
         // Store the scaling factor if it's the smallest yet
-        if (relative_change < min_relative_velocity_all_joints)
-          min_relative_velocity_all_joints = relative_change;
+        if (scaling_factor < velocity_limit_scaling_factor)
+          velocity_limit_scaling_factor = scaling_factor;
       }
     }
     ++joint_delta_index;
   }
 
   // Apply the velocity scaling to all joints
-  if (min_relative_velocity_all_joints < 1)
+  if (velocity_limit_scaling_factor < 1)
   {
     joint_delta_index = 0;
     for (auto joint : joint_model_group_->getActiveJointModels())
     {
-      delta_theta(joint_delta_index) = min_relative_velocity_all_joints * delta_theta(joint_delta_index);
-      velocity(joint_delta_index) = min_relative_velocity_all_joints * velocity(joint_delta_index);
+      delta_theta(joint_delta_index) = velocity_limit_scaling_factor * delta_theta(joint_delta_index);
+      velocity(joint_delta_index) = velocity_limit_scaling_factor * velocity(joint_delta_index);
 
       ++joint_delta_index;
     }
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -456,7 +456,7 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::TwistStamped& cmd,
 
   delta_theta_ = pseudo_inverse * delta_x;
 
-  enforceSRDFAccelVelLimits(delta_theta_);
+  enforceVelLimits(delta_theta_);
 
   // If close to a collision or a singularity, decelerate
   applyVelocityScaling(delta_theta_, velocityScalingFactorForSingularity(delta_x, svd, pseudo_inverse));
@@ -482,7 +482,7 @@ bool ServoCalcs::jointServoCalcs(const control_msgs::JointJog& cmd, trajectory_m
   // Apply user-defined scaling
   delta_theta_ = scaleJointCommand(cmd);
 
-  enforceSRDFAccelVelLimits(delta_theta_);
+  enforceVelLimits(delta_theta_);
 
   // If close to a collision, decelerate
   applyVelocityScaling(delta_theta_, 1.0 /* scaling for singularities -- ignore for joint motions */);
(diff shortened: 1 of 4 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -676,45 +676,19 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
...
 
-void ServoCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
+void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
 {
   Eigen::ArrayXd velocity = delta_theta / parameters_.publish_period;
   const Eigen::ArrayXd acceleration = (velocity - prev_joint_velocity_) / parameters_.publish_period;
 
   std::size_t joint_delta_index = 0;
+  // Track the smallest velocity scaling factor required, across all joints
+  double min_relative_velocity_all_joints = 1;
```
nbbrooks, line 686: Your comment is very descriptive - I'd rename the var accordingly to something like `velocity_limit_scaling_factor` or `velocity_limit_enforcement_scaling_factor`

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -736,21 +710,32 @@ void ServoCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
...
-        // delta_theta = joint_velocity * delta_t
-        const double relative_change = (velocity_limit * parameters_.publish_period) / delta_theta(joint_delta_index);
-        // Avoid nan
-        if (fabs(relative_change) < 1)
-        {
-          delta_theta(joint_delta_index) = relative_change * delta_theta(joint_delta_index);
-          velocity(joint_delta_index) = relative_change * velocity(joint_delta_index);
-        }
+        const double relative_change =
+            fabs(velocity_limit * parameters_.publish_period) / fabs(delta_theta(joint_delta_index));
```
nbbrooks, line 714: nit I would rename `relative_change` to something like `scaling_factor`

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -736,21 +710,32 @@ void ServoCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
...
+        if (relative_change < min_relative_velocity_all_joints)
+          min_relative_velocity_all_joints = relative_change;
       }
     }
     ++joint_delta_index;
   }
+
+  // Apply the velocity scaling to all joints
+  if (min_relative_velocity_all_joints < 1)
+  {
```
nbbrooks, line 726: Should there be a ROS_WARN or ROS_INFO here or would that be too spammy?

moveit_ros/moveit_servo/test/servo_cpp_interface_test.cpp
```diff
@@ -180,6 +182,61 @@ TEST_F(ServoFixture, SendJointServoTest)
...
+  ros::Rate publish_rate(1. / publish_period);
+
+  // Send a few Cartesian commands with very high velocity
+  for (size_t i = 0; i < num_commands && ros::ok(); ++i)
+  {
+    auto msg = moveit::util::make_shared_from_pool<geometry_msgs::TwistStamped>();
+    msg->header.stamp = ros::Time::now();
+    msg->header.frame_id = "panda_link0";
+    msg->twist.linear.x = 10.0;
+    msg->twist.angular.y = 10.0;
```
nbbrooks, line 219: Instead of `10.0`, what about `2. * LARGEST_ALLOWABLE_PANDA_VEL`

moveit_ros/moveit_servo/test/servo_cpp_interface_test.cpp
```diff
@@ -180,6 +182,61 @@ TEST_F(ServoFixture, SendJointServoTest)
...
+  auto parameters = servo_->getParameters();
+
+  // count trajectory messages sent by servo
+  trajectory_msgs::JointTrajectory joint_command_from_servo;
+  trajectory_msgs::JointTrajectory prev_joint_command_from_servo;
+  boost::function<void(const trajectory_msgs::JointTrajectoryConstPtr&)> traj_callback =
+      [&joint_command_from_servo, &prev_joint_command_from_servo](const trajectory_msgs::JointTrajectoryConstPtr& msg) {
+        // Store a series of two commands so we can calculate velocities from positions
+        prev_joint_command_from_servo = joint_command_from_servo;
+        joint_command_from_servo = *msg;
```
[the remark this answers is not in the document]
AndyZe: I think you're saying I should test every new command coming out of Servo. OK, done. Also added the check on received_count

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: too spammy IMO

gavanderhoorn: There are throttled versions of `ROS_WARN(..)` and `ROS_INFO(..)`, which could reduce the *spammy-ness*.

See `*_THROTTLE` et al.

moveit_ros/moveit_servo/test/servo_cpp_interface_test.cpp
```diff
@@ -50,6 +50,7 @@
 #include <moveit_servo/make_shared_from_pool.h>
 
 static const std::string LOGNAME = "servo_cpp_interface_test";
+static constexpr double LARGEST_ALLOWABLE_PANDA_VEL = 2.8710;  // to test joint velocity limit enforcement
```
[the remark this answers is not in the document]
AndyZe: Servo gets it from:

joint_model_group_->getActiveJointModels();
const std::vector<moveit_msgs::JointLimits> limits = joint->getVariableBoundsMsg();

I guess I could make a getter method to access that more easily. Worth it?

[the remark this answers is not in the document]
nbbrooks: Sounds like scope creep, but that would be a nice contribution for a new MoveIt dev!

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -676,45 +676,19 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
   return velocity_scale;
 }
 
-void ServoCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
+void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
 {
   Eigen::ArrayXd velocity = delta_theta / parameters_.publish_period;
   const Eigen::ArrayXd acceleration = (velocity - prev_joint_velocity_) / parameters_.publish_period;
```
jliukkonen, line 682: I believe the accelaration variable ends up being unused with the refactored velocity limits function.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -736,21 +710,30 @@ void ServoCalcs::enforceSRDFAccelVelLimits(Eigen::ArrayXd& delta_theta)
...
-        {
-          delta_theta(joint_delta_index) = relative_change * delta_theta(joint_delta_index);
-          velocity(joint_delta_index) = relative_change * velocity(joint_delta_index);
-        }
+        const double scaling_factor =
+            fabs(velocity_limit * parameters_.publish_period) / fabs(delta_theta(joint_delta_index));
+
+        // Store the scaling factor if it's the smallest yet
+        if (scaling_factor < velocity_limit_scaling_factor)
+          velocity_limit_scaling_factor = scaling_factor;
```
jliukkonen, line 718: Maybe I misunderstood but it looks to me like the function first looks for the smallest velocity limit scaling factor (which can end up being very small) and then, couple lines down, applies that lowest velocity limit to all the joints. What if there is a use case where someone wants, let's say limit the maximum velocity of the largest joint while keeping the other joints moving much faster. Wouldn't this approach then slow the whole robot down since one of the joints is limited to very low velocity?

AndyZe: Yeah, the whole robot will slow down.

But, we want the robot to keep moving in the right direction. IMO that trumps speed. Here's a video of what I mean.

nonlinear_limit_clipping

tylerjw:
Kinetic is causing your CI failure, if you rebase this on master it will get my change to stop testing on kinetic.

AndyZe:
Marking WIP while I test it a bit more

tylerjw (approved):
I tested this and it works for me.  Thank you for adding the test :1st_place_medal: :1st_place_medal: :1st_place_medal:

henningkayser (approved):

henningkayser:
@AndyZe I'm fine with merging this as this version seems to be safer than the old implementation. In the future we should still check for acceleration values, though.

PR #2266, opened 2020-08-19
[maint] add soname version to moveit_servo

tylerjw:
### Description

I went and verified that this line exists in all the other cmake files in moveit.


v4hn (approved):
I'm surprised jog_arm was merged without this! Thanks for pointing it out!

PR #2271, opened 2020-08-20
Add missing functions to create/attach collision objects in PlanningScene's python API

rhaschke:
We have seen several PRs added a missing functions `add_xxx()` or `attach_xxx()` to the python API of PSI, e.g. #2270.
This is an attempt to reduce the high redundancy in the corresponding code and define all the functions once for all.


moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -306,14 +239,59 @@ def __make_cylinder(name, pose, height, radius):
...
+        co = make_method(name, pose, *args, **kwargs)
+        self._PlanningSceneInterface__submit(co)
+
+    def attach(self, link, name, pose=None, *args, **kwargs):
+        touch_links = kwargs.pop('touch_links', [link])
+        aco = AttachedCollisionObject()
+        if pose is not None:
+            try:
+                aco.object = make_method(name, pose, *args, **kwargs)
+            except TypeError:  # thrown if touch_links was specified as an unnamed last argument
```
felixvd, line 282: I'm not sure how we know this to be true, but I don't object to the fallback.

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -306,14 +239,59 @@ def __make_cylinder(name, pose, height, radius):
...
+        if isinstance(collision_object, AttachedCollisionObject):
             scene.robot_state.attached_collision_objects = [collision_object]
         else:
             scene.world.collision_objects = [collision_object]
         planning_scene_diff_req = ApplyPlanningSceneRequest()
         planning_scene_diff_req.scene = scene
         return planning_scene_diff_req
+
+
+def __create_generic_methods(shape):
```
felixvd, line 268:
```suggestion
def __create_add_attach_methods(shape):
```
Since the methods themselves are not generic?

Also accordingly further down.

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -306,14 +239,59 @@ def __make_cylinder(name, pose, height, radius):
...
+    """Add methods add_xxx() and attach_xxx() to class PlanningSceneInterface for a given shape"""
+    make_method = getattr(PlanningSceneInterface, '_PlanningSceneInterface__make_' + shape)
+
+    def add(self, name, pose, *args, **kwargs):
+        co = make_method(name, pose, *args, **kwargs)
+        self._PlanningSceneInterface__submit(co)
+
+    def attach(self, link, name, pose=None, *args, **kwargs):
+        touch_links = kwargs.pop('touch_links', [link])
```
felixvd, line 277: Doesn't this [raise a KeyError](https://stackoverflow.com/questions/49218302/python-difference-between-kwargs-pop-and-kwargs-get) if `touch_links` is given as an unnamed argument?

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -306,14 +239,59 @@ def __make_cylinder(name, pose, height, radius):
         return co
 
     @staticmethod
-    def __make_planning_scene_diff_req(collision_object, attach=False):
+    def __make_plane(self, name, pose, normal=(0, 0, 1), offset=0):
```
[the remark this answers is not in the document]
rhaschke: Well spotted!

moveit_commander/src/moveit_commander/planning_scene_interface.py
rhaschke: What do you mean: What is true?

moveit_commander/src/moveit_commander/planning_scene_interface.py
rhaschke: No, because I provide a default value.

v4hn: So I guess there is no way to specify touch_links as an **unnamed** argument (as Felix asked)?

rhaschke: Of course, there is (as you probably just figured out): If the argument is provided as last, unnamed argument, the first call will fail with a TypeError and we try again.

v4hn: Yes I did, sorry for the noise.

Relying on the fact that the `__make` function family does not accept a list of string argument seems rather implicit though, and I guess that' what confused @felixvd as well.

Could you please add a comment that explains *why* the exception is thrown and not just *that* it is thrown?

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -306,14 +239,59 @@ def __make_cylinder(name, pose, height, radius):
...
+        co = make_method(name, pose, *args, **kwargs)
+        self._PlanningSceneInterface__submit(co)
+
+    def attach(self, link, name, pose=None, *args, **kwargs):
+        touch_links = kwargs.pop('touch_links', [link])
+        aco = AttachedCollisionObject()
+        if pose is not None:
+            try:
+                aco.object = make_method(name, pose, *args, **kwargs)
+            except TypeError:  # thrown if touch_links was specified as an unnamed last argument
```
rhaschke, line 282: @v4hn Is that what you are thinking of?
```suggestion
            except TypeError:  # thrown if touch_links was specified as an unnamed last argument
                               # (not matching the signature of make_method, which doesn't expect this additional argument)
```

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -306,14 +239,59 @@ def __make_cylinder(name, pose, height, radius):
...
         planning_scene_diff_req = ApplyPlanningSceneRequest()
         planning_scene_diff_req.scene = scene
         return planning_scene_diff_req
+
+
+def __create_add_attach_methods(shape):
+    """Add methods add_xxx() and attach_xxx() to class PlanningSceneInterface for a given shape"""
+    make_method = getattr(PlanningSceneInterface, '_PlanningSceneInterface__make_' + shape)
+
+    def add(self, name, pose, *args, **kwargs):
```
v4hn, line 272: If possible, please add back the method doc strings

        """
        Add a sphere to the planning scene
        """

moveit_commander/src/moveit_commander/planning_scene_interface.py
v4hn: Something like that, yes.

felixvd:
This PR is blocked because we're not sure if `doxygen` will generate documentation for the dynamically generated functions. If it does, this can be merged. If it doesn't, I think we have to drop it because the functions will be practically invisible.

Anyone who can test the `doxygen` output can unblock this.

PR #2275, opened 2020-08-23
Add a utility to print collision pairs

AndyZe:
I had to create my own implementations for this several times while debugging. So I figure it is time to make easier collision debugging available publicly.

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
@@ -45,12 +45,6 @@
 #include <Eigen/Core>
 #include <moveit/robot_model/robot_model.h>
 
-namespace
-{
-static const char LOGNAME[] = "collision_common";
-constexpr double ROS_LOG_THROTTLE_PERIOD = 5;
-}  // namespace
-
 namespace collision_detection
 {
 MOVEIT_CLASS_FORWARD(AllowedCollisionMatrix);
@@ -150,7 +144,12 @@ struct CostSource
 /** \brief Representation of a collision checking result */
 struct CollisionResult
 {
-  CollisionResult() : collision(false), distance(std::numeric_limits<double>::max()), contact_count(0)
+  CollisionResult()
+    : collision(false)
+    , distance(std::numeric_limits<double>::max())
+    , contact_count(0)
+    , collision_log_period(5)
+    , collision_log_name("collision_common")
   {
   }
   typedef std::map<std::pair<std::string, std::string>, std::vector<Contact> > ContactMap;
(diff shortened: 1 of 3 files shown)


moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
```diff
@@ -161,6 +167,31 @@ struct CollisionResult
     cost_sources.clear();
   }
 
+  /** \brief Throttled log of the first collision pair, if any */
+  void logFirstCollisionPair()
+  {
+    if (!contacts.empty())
+    {
+      ROS_WARN_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME,
+                                     "Objects in collision (among others, possibly): "
```
felixvd, line 176:
```suggestion
                                     "Objects in collision (printing first collision pair of " << contacts.size() << "): "
```

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
```diff
@@ -45,6 +45,12 @@
 #include <Eigen/Core>
 #include <moveit/robot_model/robot_model.h>
 
+namespace
```
felixvd, line 48: What is this namespace for?

felixvd (commented):
I don't really understand what this change will improve. It seems like the previous functions were already printing out the same things, and they were using the DEBUG level logger in a sensible way, too. Is there a use case I am missing?

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
v4hn: an anonymous namespace *and* a static definition are somewhat weird indeed.
But the problem is that this exposes a LOGNAME variable *in a common MoveIt header* which it of course mustn't.

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
```diff
@@ -161,6 +167,31 @@ struct CollisionResult
...
+    if (!contacts.empty())
+    {
+      ROS_WARN_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME,
+                                     "Objects in collision (among others, possibly): "
+                                         << contacts.begin()->first.first << ", " << contacts.begin()->first.second);
+    }
+  }
+
+  /** \brief Throttled log of all collision pairs */
+  void logAllCollisionPairs()
```
v4hn, line 182: I quite liked the distinction between WARN for first and DEBUG for all further collisions.
Do you think it's too much to expect people to understand/use the logger level system for that?

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
AndyZe: The "anonymous namespace" is supposed to ensure LOGNAME is local to this file.

`Having something in an anonymous namespace means it's local to this translation unit (.cpp file and all its includes) this means that if another symbol with the same name is defined elsewhere there will not be a violation of the One Definition Rule (ODR).`

But yeah, it's not working right. I think the problem might be using it in a header file.

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
AndyZe: thanks! Will make this change.

You're right, the previous functions printed the same thing but that was only in moveit_servo. The change here would allow somebody (for example) to print the links that are in collision while working through this tutorial:

https://ros-planning.github.io/moveit_tutorials/doc/planning_scene/planning_scene_tutorial.html

They just need to add `collision_request.logFirstCollisionPair();`

felixvd: Oh, I didn't see that the deleted function was in `moveit_servo` and not `collision_common`. A printout like that would be useful, I agree. 

I don't have a strong opinion, but I also liked that DEBUG printed all collision pairs by default. I would keep it combined and call the function `print()`. Maybe @v4hn has a comment.

Also, if `CollisionResult` gets its own member function it should be a class and not a struct ([Google style guidelines](https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes)).

AndyZe: I have to push back on making it a class.

1) there already was a helper function there, clear()
2) the google style guide you linked says this about structs:  `Constructors, destructors, and helper methods may be present; however, these methods must not require or enforce any invariants.`
3) if it was a class, I would have to rename the members, and that would be a huge change

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
AndyZe: i suppose it's not. I'll do what you and Felix both suggested - warn for the first, debug for the rest

moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
felixvd: I guess a `print()` function classifies as a helper function. I'm fine with this, just relaying it.

PR #2252, opened 2020-08-13
Melodic backports

rhaschke:
Current melodic-devel is [broken on Travis](https://travis-ci.com/github/ros-planning/moveit/builds/174775901), because of the missing cherry-pick e894d3892ba220746c3f4dca183ff655ba9e3d32.
Thus, I decided to cherry-pick all recent changes on master (up to 88712bb7f5ab2c4153440c5c55f915798ca83dd3). Here is the full list of PRs that I considered (critical ones are bold-face):

- #2193: Fix segfault in PSM::clearOctomap()
- #2194: use moveit_resources for testing
- #2207: Remove urdf package as build_depend from package.xml
- #2180: Various fixes for upcoming Noetic release (only parts, ignored commits are crossed out):
  - a617923cb7adc499aa7c2863ac0ddf597c431f9a Prefer vendor-specific OpenGL library
  - e894d3892ba220746c3f4dca183ff655ba9e3d32 **Further increase acceptance threshold for mesh-filter test**
  - ~~93cdeabf2c774d1a8066a3736e47ebdc68ec11c7 use SYSTEM includes to suppress external warnings~~
  - ~~a8ee3cb11433716baaf9bc23d8aa2a64d799d5c0 Fix warnings~~
  - ~~7c0d067fef49498812947df5e2308691fc05dfb4 Travis: Enable Noetic build~~
  - ~~9dd66f905c93b8c5ca86d6d4df531bfb66495973 Remove moveit_noetic.rosinstall~~
- #2198: Allow adding planning scene shapes from rviz panel
- #2215: Fix docstrings in robot_state.h
- #2221: Travis: Update dist to Bionic
- #2201: Exposed missing parameters from the MoveGroup Python interface
- #2229: Optimize includes
- #2232: Remove unused yaml param
- #2204: Check collisions during joint motions, too
- #2239: Fix memcpy bug in copyJointToOMPLState in ompl interface
- #2242: Add planning_adapters for chomp in setup assistant
- #2177: **Migration to clang-format-10**
- #2246: A ROS service to reset the Servo status
- #2199: **Adapt repository for splitted moveit_resources layout**

This is the list of ignored PRs (because they didn't apply easily)
- #2211: Add named logging with LOGNAME to OMPL interface
- #2214: Remove dead code from ompl interface (related to subspaces)
- #2226: Add documentation to ompl interface
- #2227: OMPL interface: fix / ignore clang-tidy warnings

[...]

PR #2277, opened 2020-08-26
Melodic backports in ompl interface (cleanup)

JeroenDM:
### Description

I manually backported the commits mentioned in #2252:

- #2211: Add named logging with LOGNAME to OMPL interface
- #2214: Remove dead code from ompl interface (related to subspaces)
- #2226: Add documentation to ompl interface
- #2227: OMPL interface: fix / ignore clang-tidy warnings

There was not always a clear one-to-one mapping, so some minor things are different.

- Add named logging to a script that is removed in the latest version: 0c0d8aa
- Fixed another clang-tidy warning: `performance-implicit-conversion-in-loop` in [constraint_library.cpp](https://github.com/ros-planning/moveit/blob/8b5782cce98e93b9efa33a3269d11c0c84c2f536/moveit_planners/ompl/ompl_interface/src/constraints_library.cpp#L373).
- Correct some spelling errors, while I was going through it anyway: 8b5782c

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -59,6 +59,11 @@
 #include "ompl/base/objectives/StateCostIntegralObjective.h"
 #include "ompl/base/objectives/MaximizeMinClearanceObjective.h"
 
+namespace ompl_interface
+{
+constexpr char LOGNAME[] = "model_based_planning_context";
+}  // namespace ompl_interface
+
 ompl_interface::ModelBasedPlanningContext::ModelBasedPlanningContext(const std::string& name,
                                                                      const ModelBasedPlanningContextSpecification& spec)
   : planning_interface::PlanningContext(name, spec.state_space_->getJointModelGroup()->getName())
@@ -91,7 +96,7 @@ void ompl_interface::ModelBasedPlanningContext::setProjectionEvaluator(const std
 {
   if (!spec_.state_space_)
   {
-    ROS_ERROR_NAMED("model_based_planning_context", "No state space is configured yet");
+    ROS_ERROR_NAMED(LOGNAME, "No state space is configured yet");
     return;
   }
   ob::ProjectionEvaluatorPtr pe = getProjectionEvaluator(peval);
@@ -108,7 +113,7 @@ ompl_interface::ModelBasedPlanningContext::getProjectionEvaluator(const std::str
     if (getRobotModel()->hasLinkModel(link_name))
       return ob::ProjectionEvaluatorPtr(new ProjectionEvaluatorLinkPose(this, link_name));
     else
-      ROS_ERROR_NAMED("model_based_planning_context",
+      ROS_ERROR_NAMED(LOGNAME,
                       "Attempted to set projection evaluator with respect to position of link '%s', "
                       "but that link is not known to the kinematic model.",
                       link_name.c_str());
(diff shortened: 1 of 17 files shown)


JeroenDM:
I fixed some other clang-tidy warnings that showed up on Travis.

JeroenDM:
@v4hn @rhaschke @mamoll 
I'm not sure why there are no reviewers assigned. Hence this manual notification.

v4hn (approved):
Looks good to me and CI passes (aside from coverage). :+1: 

If you spend the time to fix typos in the documentation, you should really do it in `master` though.
8b5782cce98e93b9efa33a3269d11c0c84c2f536 applies there as well :-)

JeroenDM:
I never really understood the coverage check. (Admiditly, I did not spend much time digging into the details.)

`If you spend the time` might be an overstatement :) I just quickly fixed the ones I happened to read.

Should I remove the commit here and do it in master first then?

v4hn:
> I never really understood the coverage check. (Admiditly, I did not spend much time digging into the details.)

The codecov configuration is not very old yet.. It requires that patches are covered by tests roughly by the same amount as the whole project. But because the repositories current coverage is relatively low (51.8%) it's hard to write patches that are covered without adding totally independent new tests that do not make any sense to add in the respective pull-requests.

@tylerjw @rhaschke @davetcoleman
Looking at the coverage check in more detail it also looks like it uses [a damn outdated commit as base for comparison](https://codecov.io/gh/ros-planning/moveit/commit/1c3fa4aba6e6493aab30df8103716faa6329caa8) instead of the current head. I don't know who set the exact configuration up, but this looks like a bug to me and means that we do not see the *actual* coverage change of a pull-request but also of the last ~200 commits in `melodic-devel` as well. That's totally useless.

> Should I remove the commit here and do it in master first then?

Ideally yes. But I don't care for a few typos :-) .
My point is that `master` will live on, but `melodic-devel` will be discarded at some point.
So adding small patches here means they will not stay around for future generations unless you also add them to `master`.

JeroenDM:
> Ideally yes. But I don't care for a few typos :-) .
My point is that master will live on, but melodic-devel will be discarded at some point.
So adding small patches here means they will not stay around for future generations unless you also add them to master.

That makes sense. I also added it to master. (Extremely small pull requests are becoming my signature move...)

v4hn:
> Extremely small pull requests are becoming my signature move...

These are the best because they are trivial to review and still improve something! As you noticed yourself by now reviewing bigger pull-requests in MoveIt takes a lot of time and people are not always willing to invest that when the patch is not directly related to their current projects...

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -59,6 +59,11 @@
 #include "ompl/base/objectives/StateCostIntegralObjective.h"
 #include "ompl/base/objectives/MaximizeMinClearanceObjective.h"
 
+namespace ompl_interface
+{
+constexpr char LOGNAME[] = "model_based_planning_context";
+}  // namespace ompl_interface
+
```
tylerjw, line 66: [nit] Should this be an anonymous namespace? This is inside a cpp file so it won't be included anywhere.

It looks like this whole file could just be put in the `ompl_interface` namespace and then you'd be able to drop that namespace from before `ModelBasedPlanningContext` like on line 62.

Here is an example of what I'm suggesting: https://github.com/ros-planning/moveit/blob/7ad2bc7b86dad08061d98668ba34feba54bb05cc/moveit_ros/moveit_servo/src/low_pass_filter.cpp#L47

tylerjw (approved):
Thank you for this cleanup.  It is really nice to have a more consistent and clean codebase.  I did find one small thing.  But since this is a backport please only do the change in a followup commit if you like that change and make it target both master and melodic-devel.

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
JeroenDM: > It looks like this whole file could just be put in the ompl_interface namespace

This is something I've been wanting to do for a while. It could also make some other files cleaner, but I thought there might be some guidelines against putting the whole source file in a namespace. (No idea why, just that I do not see another reason why these files are written this way.)

The anonymous namespace thing is interesting! It's a feels a bit counterintuitive that you can access variables inside an anonymous namespace :)

JeroenDM:
@tylerjw thanks. I rebased it.

tylerjw:
@JeroenDM Would you like me to squash this into one commit on melodic-devel or merge it?

JeroenDM:
I think the commit descriptions are clear enough to merge it. But I don't have a strong preference for one or the other.

PR #2283, opened 2020-08-27
[maint] adjust code coverage yaml to make it more forgiving

tylerjw:
### Description

These changes are in response to the discussion today in the meeting @v4hn.  These should make the codecov tests always pass.

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -7,9 +7,7 @@ coverage:
       default:
         target: auto
         threshold: 5%
-        informational: true
     patch:
       default:
         target: auto
         threshold: 5%
-        informational: true
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,13 +1,15 @@
 coverage:
+  precision: 2
+  round: up
+  range: "45...70"
   status:
     project:
       default:
         target: auto
-        # allow coverage to drop by this amount and still post success
-        threshold: 0.5%
-        base: auto
-        branches:
-          - master
-        if_not_found: success
-        if_ci_failed: error
-        only_pulls: false
+        threshold: 5%
+        informational: true
+    patch:
+      default:
+        target: auto
+        threshold: 5%
+        informational: true


felixvd (approved):
Looks reasonable. Have you also looked into [this issue](https://github.com/ros-planning/moveit/pull/2277#issuecomment-681852476)?

codecov.yaml
```diff
@@ -1,13 +1,15 @@
 coverage:
+  precision: 2
+  round: up
+  range: "45...70"
```
v4hn, line 4: This should be "45...100", right? :-)

codecov.yaml
```diff
@@ -1,13 +1,15 @@
...
-        # allow coverage to drop by this amount and still post success
-        threshold: 0.5%
-        base: auto
-        branches:
-          - master
-        if_not_found: success
-        if_ci_failed: error
-        only_pulls: false
+        threshold: 5%
+        informational: true
```
v4hn, line 10: defining `threshold` **and** `informational` does not make sense to me.
Either we want it to succeed *always* or only if coverage does not drop by a lot.
Given that 5% coverage drop in the whole MoveIt repository with a single PR should be pretty hard to do, I would propose to comment out `informational` for now.

codecov.yaml
```diff
@@ -1,13 +1,15 @@
...
+    patch:
+      default:
+        target: auto
+        threshold: 5%
+        informational: true
```
v4hn, line 15: Same here, I would propose to comment `informational`.

Notice that even if CI fails, this does not block maintainers from merging the change because Codecov testing is not configured as mandatory for the github repo.

v4hn (commented):
Thanks for looking into this.

The weird base for comparison Felix mentioned might be related to these changes (we did not specify values for `patch` before), or is yet another issue...

codecov.yaml
tylerjw: this is the color gradient applied to files (from red to green).  At our low code coverage this would make 70% or greater green.

codecov.yaml
tylerjw: OK, I'll remove the informational.

codecov.yaml
tylerjw: See previous comment.  I'll remove this.  I'm ok with seeing a red x for a 5% drop.

tylerjw:
> Looks reasonable. Have you also looked into [this issue](https://github.com/ros-planning/moveit/pull/2277#issuecomment-681852476)?

I did look into that and I still don't understand it.  I think the patch section I added in this might help with that.

codecov.yaml
v4hn: fair enough!

v4hn (approved):

PR #2293, opened 2020-09-03
MSA: only write default_planner_config fields if any is selected

v4hn:
Otherwise we get

```
arm:
    default_planner_config: None
```

entries in the ompl_planning.yaml, where "None" of course does not describe an existing planner configuration

Why the hell do I have to write at least two issues and one pull-request every single time I actually use the setup assistant!
We desperately need more thorough tests and users reporting issues with the assistant.

### DIFF
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -242,7 +242,8 @@ bool MoveItConfigData::outputOMPLPlanningYAML(const std::string& file_path)
     emitter << YAML::Key << group.name_;
     emitter << YAML::Value << YAML::BeginMap;
     // Output associated planners
-    emitter << YAML::Key << "default_planner_config" << YAML::Value << group_meta_data_[group.name_].default_planner_;
+    if (!group_meta_data_[group.name_].default_planner_.empty())
+      emitter << YAML::Key << "default_planner_config" << YAML::Value << group_meta_data_[group.name_].default_planner_;
     emitter << YAML::Key << "planner_configs";
     emitter << YAML::Value << YAML::BeginSeq;
     for (const std::string& pconfig : pconfigs)
@@ -272,7 +273,7 @@ bool MoveItConfigData::outputOMPLPlanningYAML(const std::string& file_path)
     return false;
   }
 
-  output_stream << emitter.c_str();
+  output_stream << emitter.c_str() << std::endl;
   output_stream.close();
 
   return true;  // file created successfully
--- a/moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
@@ -1240,7 +1240,7 @@ bool PlanningGroupsWidget::saveGroupScreen()
   config_data_->group_meta_data_[group_name].kinematics_solver_search_resolution_ = kinematics_resolution_double;
   config_data_->group_meta_data_[group_name].kinematics_solver_timeout_ = kinematics_timeout_double;
   config_data_->group_meta_data_[group_name].kinematics_parameters_file_ = kinematics_parameters_file;
-  config_data_->group_meta_data_[group_name].default_planner_ = default_planner;
+  config_data_->group_meta_data_[group_name].default_planner_ = (default_planner == "None" ? "" : default_planner);
   config_data_->changes |= MoveItConfigData::GROUP_KINEMATICS;
 
   // Reload main screen table


jliukkonen (approved):
:+1:

moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
```diff
@@ -1240,7 +1240,7 @@ bool PlanningGroupsWidget::saveGroupScreen()
   config_data_->group_meta_data_[group_name].kinematics_solver_search_resolution_ = kinematics_resolution_double;
   config_data_->group_meta_data_[group_name].kinematics_solver_timeout_ = kinematics_timeout_double;
   config_data_->group_meta_data_[group_name].kinematics_parameters_file_ = kinematics_parameters_file;
-  config_data_->group_meta_data_[group_name].default_planner_ = default_planner;
+  config_data_->group_meta_data_[group_name].default_planner_ = (default_planner == "None" ? "" : default_planner);
```
rhaschke, line 1243: Instead of this, I would have used a comparison against `None` in the above guard. I think this would be more intuitive.
```suggestion
  config_data_->group_meta_data_[group_name].default_planner_ = default_planner;
```

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -242,7 +242,8 @@ bool MoveItConfigData::outputOMPLPlanningYAML(const std::string& file_path)
     emitter << YAML::Key << group.name_;
     emitter << YAML::Value << YAML::BeginMap;
     // Output associated planners
-    emitter << YAML::Key << "default_planner_config" << YAML::Value << group_meta_data_[group.name_].default_planner_;
+    if (!group_meta_data_[group.name_].default_planner_.empty())
```
rhaschke, line 245:
```suggestion
    if (!group_meta_data_[group.name_].default_planner_ != "None")
```

rhaschke (approved):
Nitpick, otherwise looks good.

moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
v4hn: That would break old configurations that did not specify the default planner and thus are empty until the user edits the group.

rhaschke: I'm not sure yet how this breaks old configs. This function is about storing the contents from the GUI:
https://github.com/ros-planning/moveit/blob/8b4bdef46403c67ac13ece608a7af37c2bde5533/moveit_setup_assistant/src/widgets/planning_groups_widget.cpp#L1087
into the `config_data_` structure:
https://github.com/ros-planning/moveit/blob/8b4bdef46403c67ac13ece608a7af37c2bde5533/moveit_setup_assistant/src/widgets/planning_groups_widget.cpp#L1243
isn't it? Maybe, you consider the case, where `default_planner_` is initialized from the existing file:
https://github.com/ros-planning/moveit/blob/8b4bdef46403c67ac13ece608a7af37c2bde5533/moveit_setup_assistant/src/tools/moveit_config_data.cpp#L1328-L1335
and later rewritten to file? In that case, shouldn't the parser handle both, empty strings and `None` as well?

v4hn: > Maybe, you consider the case, where default_planner_ is initialized from the existing file:

That's exactly what I meant.

> In that case, shouldn't the parser handle both, empty strings and None as well?

Why have two null values to take care of, when you can have one for the GUI ("None") and one for the internal datastructure (empty string)? That's why I went for the current patch.

rhaschke: Old configs _can_ exhibit either of both null values, i.e. `None` or an empty string. Thus, without explicitly handling this in the parser code, we will write back `None` (if it was there before). I think that's what you want to avoid?
Your comment pointed me to another open issue: If the existing config file has an empty string, the data structure will be initialized as an empty string and the GUI won't show `None` (but an empty string), right? Thus, I think, we should stick with _None_ internally.

v4hn: > Old configs can exhibit either of both null values, i.e. None or an empty string.

No they can't. Our generator was simply broken. If users generated configurations with the bug, they get error messages about this on each and every startup until they remove the entries. I don't think it makes any sense to provide "fixup" logic in MSA for obvious former bugs. Aside from that it would turn "None" into a magic string with differing semantics.

> If the existing config file has an empty string, the data structure will be initialized as an empty string and the GUI won't show None (but an empty string), right?

No, this currently works as expected.

You are bikeshedding.

rhaschke: > > Old configs can exhibit either of both null values, i.e. None or an empty string. 
> 
> No they can't.

Thanks for this clarification. That was the point I was missing :wink:

> > If the existing config file has an empty string, the data structure will be initialized as an empty string and the GUI won't show None (but an empty string), right?
> 
> No, this currently works as expected.

Good.

PR #2304, opened 2020-09-10
Noetic was released

v4hn:


v4hn:
> We still have build issues on Debian Buster with this release: ros/rosdistro#26484 (comment)!

Yes, thanks for working on this with Jochen. I don't see why this should be a reason not to merge this patch...

rhaschke (approved):

PR #2305, opened 2020-09-10
WIP: Add cartesian limit to FixStartStateCollision

### LINKED ISSUE #2268: FixStartStateCollision can cause sudden jumps
### Description

Ever since I first started using MoveIt, the `FixStartStateCollision` adapter has confused me. Every time I notice it, it makes me think "Thank god that didn't hit anything".

On a UR5e, the "jiggle" it introduces can result in very fast motions of up to 10 cm. With `iiwa_stack` running on a KUKA iiwa, the jumps can go even further because of the lazy tracking. Anecdotally, I observed up to 40 cm offset from the trajectory and many surprised faces.

One can argue that if the environment is set up correctly, this won't be an issue. But there are plenty of reasons why the real world is not perfectly aligned to the planning scene, and the robot's reaction to that sort of mismatch should be to stop, not to attempt a potentially dangerous repair, so I don't think this behavior should be the default. I would suggest to:

**a)** Remove the FixStartStateCollision completely, or  
**b)** Reduce the speed of the trajectory at the start to ensure a low end effector speed, or  
**c)** Evaluate the cartesian distance of the end effector instead of the "jiggle_fraction"

One could also rewrite the adapter, but I doubt anyone has cycles for that. Would be interested in opinions.

### Steps to reproduce
Spawn an object in the scene, move it so it very slightly touches the robot, then move the robot.

### Expected behaviour
The robot should either stop or move away slowly.

### Actual behaviour
The evasive movement can be both large and very fast.

felixvd:
### Description

This is a short PR that restricts the maximum cartesian motion of the robot links in FixStartStateCollision, as discussed in #2268. I made two commits because the first one only checked the robot links but not the end effector link. I don't have a good test case on hand so this is untested (sorry), but at least it is a short and simple PR. I was also unsure about a few things that should be easy to review, like how to calculate the distance with `Eigen::Isometry3d`. Thanks in advance for taking a look.

I noticed that `checkCollision` was called after every update to a joint (and not once all joints have been changed), which is both counter-intuitive and made links closer to the root more likely to move. Collisions are checked much less frequently with this PR.

Fixes #2268. Pinging @simonschmeisser , @v4hn, @rhaschke because you commented on the issue.

### Checklist


v4hn:
@tylerjw I do think the new codecov configuration spams too many comments, what do you think?

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -71,6 +72,14 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
     else
       ROS_INFO_STREAM("Param '" << JIGGLE_PARAM_NAME << "' was set to " << jiggle_fraction_);
 
+    if (!nh.getParam(JIGGLE_PARAM_NAME, jiggle_fraction_))
+    {
+      max_jiggle_distance_squared_ = 0.01*0.01; // = 1 cm
```
v4hn, line 77: Sorry, but please test your commits before asking for detailed feedback. The else-case misses setting max_jiggle_distance_squared_ .

felixvd: Alright alright, I was in a rush, my bad. I fixed that and the rest.

felixvd:
I updated the original post.

felixvd:
The build fails because of clang-tidy, which somehow fails to work correctly in my container. I'll try to figure this out tomorrow.

> @tylerjw I do think the new codecov configuration spams too many comments, what do you think?

And yes, this is sort of extreme:

Screenshot from 2020-09-11 13-05-46

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +126,86 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
+
+          if (dist > max_dist)
+            max_dist = dist;
+          if (dist > max_jiggle_distance_squared_)
           {
-            found = true;
-            ROS_INFO("Found a valid state near the start state at distance %lf after %d attempts",
-                     prefix_state->distance(start_state), c);
+            // Reduce jiggle fraction if it caused excessive movement
+            local_jiggle_fraction_ = local_jiggle_fraction_ * ((sampling_attempts_ - 2) / sampling_attempts_);
```
v4hn, line 191: This breaks if `sampling_attempts_ <= 2`.

I'm not sold on the simulated annealing. Did you weigh up against alternatives, e.g. no annealing, local_jiggle_fraction's for each joint, stochastic sample projection to within the maximum allowed distance?

Alternatively I could imaging changing to sample a joint diff vector (or a number of them) for directions and add this vector in a loop until we exceed the Cartesian limit, checking each intermediate joint.
Of course the best solution would be an informed procedure using Jacobian and penetration depth information, but I think that would end up with different problems in terms of complexity.

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
```diff
@@ -16,7 +16,8 @@
   <arg name="allow_trajectory_execution" default="true"/>
   <arg name="fake_execution" default="false"/>
   <arg name="max_safe_path_cost" default="1"/>
-  <arg name="jiggle_fraction" default="0.05" />
+  <arg name="jiggle_fraction" default="0.02" />
+  <arg name="max_jiggle_distance" default="0.02" />
```
v4hn, line 20: These parameters both desperately need descriptive comments!

v4hn (commented):
Thanks for working on this! I added feedback inline

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
felixvd: True, I blindly assumed sampling_attempts (that's actually the max attempts, not the counter) would not be that low.

No, I didn't do any quantitative investigation, I just added it like that because I figured it couldn't hurt. I don't think it's worth the effort to look into it that deeply (this is an adapter for edge cases). It would be better spent on using penetration depth and making this work in Cartesian space as you described.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +126,86 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
+      std::vector<const moveit::core::LinkModel*> link_models;
+      link_models.resize(jmodels.size());
+      for (std::size_t i = 0; i < jmodels.size(); ++i)
+        link_models[i] = jmodels[i]->getChildLinkModel();
+
+      // Also record end effector link pose if one is set
+      if (!req.goal_constraints[0].position_constraints.empty())
+        link_models.push_back(start_state.getLinkModel(req.goal_constraints[0].position_constraints[0].link_name));
+      else if (!req.goal_constraints[0].orientation_constraints.empty())
+        link_models.push_back(start_state.getLinkModel(req.goal_constraints[0].orientation_constraints[0].link_name));
```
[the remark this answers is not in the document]
felixvd: Nice catch and good call on reversing the link order. It's only a max of two since it only adds the link from the position _or_ the orientation constraint, but it's still redundant. I added a solution that isn't very elegant, but probably still better than wasting calculations.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -71,6 +72,16 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
 
+    if (!nh.getParam(DISTANCE_PARAM_NAME, max_jiggle_distance_squared_))
+    {
+      max_jiggle_distance_squared_ = 0.02;
+      ROS_INFO_STREAM("Param '" << DISTANCE_PARAM_NAME
+                                << "' was not set. Using default value: " << max_jiggle_distance_squared_);
+    }
+    else
+      ROS_INFO_STREAM("Param '" << DISTANCE_PARAM_NAME << "' was set to " << max_jiggle_distance_squared_);
+    max_jiggle_distance_squared_ = pow(max_jiggle_distance_squared_, 2);  // Square the rosparam
```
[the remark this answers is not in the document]
felixvd: Agree, and done

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +126,86 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
+
+      EigenSTL::vector_Isometry3d original_link_poses;
+      original_link_poses.resize(link_models.size());
+      Eigen::Isometry3d new_link_pose;
+
+      for (std::size_t i = 0; i < link_models.size(); ++i)
+        original_link_poses[i] = start_state.getGlobalLinkTransform(link_models[i]);
+
+      double dist, max_dist;
+      double local_jiggle_fraction_ = jiggle_fraction_;
```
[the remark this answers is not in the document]
felixvd: Done

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
felixvd: I added some

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
```diff
@@ -16,7 +16,8 @@
   <arg name="allow_trajectory_execution" default="true"/>
   <arg name="fake_execution" default="false"/>
   <arg name="max_safe_path_cost" default="1"/>
-  <arg name="jiggle_fraction" default="0.05" />
+  <arg name="jiggle_fraction" default="0.02" />  <!--  [rad] Maximum joint space movement allowed in FixStartStateCollision adapter -->
+  <arg name="max_jiggle_distance" default="0.02" />  <!-- [m] Maximum cartesian distance allowed in FixStartStateCollision adapter -->
```
v4hn, line 20:
```suggestion
  <!-- [m] Maximum Cartesian distance any link is allowed to be displaced by the FixStartStateCollision adapter -->
  <arg name="max_jiggle_distance" default="0.02" />
```

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +127,97 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
+
+          if (dist > max_dist)
+            max_dist = dist;
+          if (dist > max_jiggle_distance_squared_)
           {
-            found = true;
-            ROS_INFO("Found a valid state near the start state at distance %lf after %d attempts",
-                     prefix_state->distance(start_state), c);
+            // Reduce jiggle fraction if it caused excessive movement
+            local_jiggle_fraction = local_jiggle_fraction * ((max_sampling_attempts_ - 2) / max_sampling_attempts_);
```
v4hn, line 203: As these are arbitrary choices anyway, why not `max_sampling_attempts_ / (max_sampling_attempts_ + 2)`?
Could you also add a comment explaining that this is an arbitrary way to implement it and other ways might be more efficient?

I'm not really happy with this mechanism as-is, but I also do not want to burden you with implementing a more sophisticated approach because, as you say, this is an edge-case.

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
```diff
@@ -16,7 +16,8 @@
   <arg name="allow_trajectory_execution" default="true"/>
   <arg name="fake_execution" default="false"/>
   <arg name="max_safe_path_cost" default="1"/>
-  <arg name="jiggle_fraction" default="0.05" />
+  <arg name="jiggle_fraction" default="0.02" />  <!--  [rad] Maximum joint space movement allowed in FixStartStateCollision adapter -->
```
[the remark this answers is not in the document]
felixvd: Oops, I was about to go to bed. Fixed.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +127,97 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
+      if (!req.goal_constraints[0].position_constraints.empty())
+        ee_link_model = start_state.getLinkModel(req.goal_constraints[0].position_constraints[0].link_name);
+      else if (!req.goal_constraints[0].orientation_constraints.empty())
+        ee_link_model = start_state.getLinkModel(req.goal_constraints[0].orientation_constraints[0].link_name);
+
+      // Collect the links that will be checked
+      std::vector<const moveit::core::LinkModel*> link_models;
+      link_models.resize(jmodels.size());
+      for (std::size_t i = jmodels.size() - 1; i > 0; --i)
+        link_models[i] = jmodels[i]->getChildLinkModel();
```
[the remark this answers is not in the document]
felixvd: I know it's supoptimal. It's to keep the `ee_link` at the front of the vector, which is a good idea, I think. There is no way to check if `ee_link_model` is part of the robot links without obtaining `link_models` (or at least iterating through `jmodels[i]->getChildLinkModel()`).

I could do one loop to check if `ee_link_model` is in `jmodels[i]->getChildLinkModel()` without assigning them a vector, and then create the vector afterwards, if you prefer:

```
      bool ee_link_separate = true;
      for (std::size_t i = jmodels.size() - 1; i > 0; --i)
      {
        if (ee_link_model == jmodels[i]->getChildLinkModel())
        {
          ee_link_separate = false;
          break;
        }
      }

      // Collect the links that will be checked
      std::vector<const moveit::core::LinkModel*> link_models;

      if (ee_link_separate)
      { 
        link_models.resize(jmodels.size() + 1);
        link_models[0] = ee_link_model;
        for (std::size_t i = jmodels.size() - 1; i > 0; --i)
          link_models[i + 1] = jmodels[i]->getChildLinkModel();
      }
      else
      {
        link_models.resize(jmodels.size());
        link_models[0] = ee_link_model;
        for (std::size_t i = jmodels.size() - 1; i > 0; --i)
          link_models[i] = jmodels[i]->getChildLinkModel();
      }
```

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
felixvd: Done. I don't mind taking it out either though, I doubt that it does much.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +127,97 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
 
+      // Record the cartesian pose of every link to ensure no sudden movements will occur.
+      // The "furthest" links are checked first, as their movement is the largest.
+
+      // Check if an end effector link is set that is not part of the robot
+      const moveit::core::LinkModel* ee_link_model;
+      if (!req.goal_constraints[0].position_constraints.empty())
+        ee_link_model = start_state.getLinkModel(req.goal_constraints[0].position_constraints[0].link_name);
+      else if (!req.goal_constraints[0].orientation_constraints.empty())
+        ee_link_model = start_state.getLinkModel(req.goal_constraints[0].orientation_constraints[0].link_name);
```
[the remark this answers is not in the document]
felixvd: Can a plan with no goal exist? I added a sanity check with `!goal_constraints.empty()`, but I would rather not iterate over them, since then we would have to clean for duplicates. Using the first constraint should cover 98%+ of cases.

Let's not forget that the failure case here is that the plan remains in collision and the robot doesn't move, so we shouldn't spend too much effort (neither calculation nor development, imo)

felixvd:
Rebased to get rid of the codecov errors

felixvd:
Ping

JafarAbdi (approved):

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -71,19 +72,30 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
-    if (!nh.getParam(ATTEMPTS_PARAM_NAME, sampling_attempts_))
+    if (!nh.getParam(DISTANCE_PARAM_NAME, max_jiggle_distance_))
     {
-      sampling_attempts_ = 100;
-      ROS_INFO_STREAM("Param '" << ATTEMPTS_PARAM_NAME << "' was not set. Using default value: " << sampling_attempts_);
+      max_jiggle_distance_ = 0.02;
+      ROS_INFO_STREAM("Param '" << DISTANCE_PARAM_NAME
+                                << "' was not set. Using default value: " << max_jiggle_distance_);
+    }
+    else
```
simonschmeisser, line 81: I'm not sure if that is part of the coding style but I prefer always placing brackets even for a single term after `if`/`else`

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +127,106 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
+      if (!req.goal_constraints.empty())
+      {
+        if (!req.goal_constraints[0].position_constraints.empty())
+          ee_link_model = start_state.getLinkModel(req.goal_constraints[0].position_constraints[0].link_name);
+        else if (!req.goal_constraints[0].orientation_constraints.empty())
+          ee_link_model = start_state.getLinkModel(req.goal_constraints[0].orientation_constraints[0].link_name);
+      }
+      else
+      {
+        ROS_ERROR("No goal constraint set; aborting FixStartStateCollision.");
```
simonschmeisser, line 157: When would this happen? Not changing anything will make it fail at the end, right?

simonschmeisser (approved):
Looks like a very necessary improvement to me that shouldn't stale!

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +127,106 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
+        // Confirm that no links move far after jiggling
+        max_dist = -1;
+        bool euclidean_distance_ok = true;
+        for (std::size_t i = 0; !found && i < link_models.size(); ++i)
+        {
+          new_link_pose = start_state.getGlobalLinkTransform(link_models[i]);
+          dist = (new_link_pose.translation() - original_link_poses[i].translation()).squaredNorm();
+
+          if (dist > max_dist)
+            max_dist = dist;
```
rhaschke, line 207: `max_dist` isn't used yet. But I think, it might be useful: My suggestion is to traverse all links and compute the largest distance. If any link moves too far, i.e. if `max_dist` is too large, try to downscale the jiggling motion. For small joint motions, the relationship between joint and Cartesian deltas is linear. Thus downscaling the joint delta by the factor `0.9 * sqrt(max_dist) / max_jiggle_distance` should result in a valid Cartesian motion. The factor 0.9 is a safety margin to handle nonlinearities. Maybe you can even drop it.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +127,106 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
+        original_link_poses[i] = start_state.getGlobalLinkTransform(link_models[i]);
+
+      double dist, max_dist;
+      double local_jiggle_fraction = jiggle_fraction_;
       bool found = false;
-      for (int c = 0; !found && c < sampling_attempts_; ++c)
+      for (int c = 0; !found && c < max_sampling_attempts_; ++c)
       {
+        // Change each joint slightly ("jiggle")
         for (std::size_t i = 0; !found && i < jmodels.size(); ++i)
```
rhaschke, line 189: `found` doesn't need to be checked in this loop as it isn't changed by the loop.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +127,106 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
           start_state.setJointPositions(jmodels[i], sampled_variable_values);
-          collision_detection::CollisionResult cres;
-          planning_scene->checkCollision(creq, cres, start_state);
-          if (!cres.collision)
+        }
+
+        // Confirm that no links move far after jiggling
+        max_dist = -1;
+        bool euclidean_distance_ok = true;
+        for (std::size_t i = 0; !found && i < link_models.size(); ++i)
```
rhaschke, line 201: `found` doesn't need to be checked here as well.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_collision.cpp
```diff
@@ -115,37 +127,106 @@ class FixStartStateCollision : public planning_request_adapter::PlanningRequestA
...
+          if (dist > max_dist)
+            max_dist = dist;
+          if (dist > max_jiggle_distance_squared_)
           {
-            found = true;
-            ROS_INFO("Found a valid state near the start state at distance %lf after %d attempts",
-                     prefix_state->distance(start_state), c);
+            // Reduce jiggle fraction if it caused excessive movement
+            // This implementation is an arbitrary choice and not tested for performance. Other approaches may be more efficient.
+            local_jiggle_fraction = local_jiggle_fraction * (max_sampling_attempts_ / (max_sampling_attempts_ + 2));
```
rhaschke, line 212: This annealing scheme decays `jiggle_fraction` only very slowly. 
Obviously, `jiggle_fraction` and `max_jiggle_distance` are strongly coupled variables (and the user is probably not aware of that fact). Assuming that `max_jiggle_distance` is the more important variable, you should quickly scale down `jiggle_fraction` if the robot moved too far, e.g using the formula suggested above.

A fundamental issue with the existing jiggling approach is that all joints use the same jiggling fraction, although they contribute differently to the overall motion of distal links. I think one should compute the Jacobian (once at the initial config) and estimate joint contributions from the linear Cartesian velocity parts.  But, that's maybe a different PR.

rhaschke (changes requested):
In my point of view, this adapter had some fundamental design issues and you even introduced some new ones:
- Jiggling is performing a random walk. It's not randomly exploring from the same original state. Thus, you might strongly diverge from the original state and the Cartesian checks will always fail afterwards. A jiggling motion should be reverted/ignored if the Cartesian check failed!
- The two parameters `jiggle_fraction` and `max_jiggle_distance` are strongly coupled, which is not considered yet.
- Jiggling is independent of the Cartesian motion contribution of individual joints.

I'm sorry to reject this as is. But I think we should first discuss in more detail how to proceed. I'm open for a skype/zoom call on this. Just email me.

PR #2306, opened 2020-09-11
[maint] Remove patch status from codecov

tylerjw:
### Description

This disables the "patch" status reporting from codecov.  I believe this is what is causing this:

https://github.com/ros-planning/moveit/pull/2305#issuecomment-690861772

Patch status documentation is here: https://docs.codecov.io/docs/commit-status#patch-status
Here is the documentation for disabling a status: https://docs.codecov.io/docs/commit-status#disabling-a-status

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -7,7 +7,4 @@ coverage:
       default:
         target: auto
         threshold: 5%
-    patch:
-      default:
-        target: auto
-        threshold: 5%
+    patch: off


v4hn:
The documentation for it [seems to be here](https://docs.codecov.io/docs/commit-status#patch-status), right?
From the documentation I would think we can disable them either way (as we only use codecov for checking pr's) but what makes you think it's responsible for the new comments?

tylerjw:
I updated the issue to actually include the links to the documentation.  The comments themselves say they are coming from the patch check:

codecov_patch_comment

lol, codecov is rebelling, this **decreases** code coverage :rofl:

felixvd:
I see. I looked through the doc and couldn't find anything that seemed to apply, and thought it might have been [this](https://docs.codecov.io/docs/github-checks-beta#disabling-github-checks-patch-annotations). This is worth trying anyway.

v4hn (approved):

PR #2307, opened 2020-09-12
[fix] Servo heap-buffer-overflow bug

tylerjw:
### Description

This was fun to track down.  The short story here is that when `suddenHalt` is called sometimes there is a point in the `points` array so the code inside the if statement doesn't get run.  Then the code below uses the `operator[]` and writes past the end of heap memory allocations that are not large enough.

What made this fun is this was corrupting memory so that later the program would crash sometimes with an error about deallocating memory that was already deallocated and the stack trace was inside ompl.  In the end what helped me find this was building with address sanitizer.

```bash
catkin config --cmake-args -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_STANDARD=17 -DCMAKE_CXX_FLAGS="-fno-omit-frame-pointer -fsanitize=address" -DCMAKE_C_FLAGS="-fno-omit-frame-pointer -fsanitize=address" -DCMAKE_LINKER_FLAGS="-fno-omit-frame-pointer -fsanitize=address" -- --no-install
```

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -263,7 +263,6 @@ void ServoCalcs::run(const ros::TimerEvent& timer_event)
   // Create new outgoing joint trajectory command message
   auto joint_trajectory = moveit::util::make_shared_from_pool<trajectory_msgs::JointTrajectory>();
   joint_trajectory->header.frame_id = parameters_.planning_frame;
-  joint_trajectory->header.stamp = ros::Time::now();
   joint_trajectory->joint_names = internal_joint_state_.name;
 
   // Prioritize cartesian servoing above joint servoing
@@ -796,19 +795,20 @@ bool ServoCalcs::enforceSRDFPositionLimits()
 void ServoCalcs::suddenHalt(trajectory_msgs::JointTrajectory& joint_trajectory)
 {
   joint_trajectory.points.clear();
-  joint_trajectory.points.push_back(trajectory_msgs::JointTrajectoryPoint());
-  joint_trajectory.points[0].positions.resize(num_joints_);
-  joint_trajectory.points[0].velocities.resize(num_joints_);
+  joint_trajectory.points.emplace_back();
+  trajectory_msgs::JointTrajectoryPoint& point = joint_trajectory.points.front();
+  point.positions.resize(num_joints_);
+  point.velocities.resize(num_joints_);
 
   for (std::size_t i = 0; i < num_joints_; ++i)
   {
     // For position-controlled robots, can reset the joints to a known, good state
     if (parameters_.publish_joint_positions)
-      joint_trajectory.points[0].positions[i] = original_joint_state_.position[i];
+      point.positions.at(i) = original_joint_state_.position.at(i);
 
     // For velocity-controlled robots, stop
     if (parameters_.publish_joint_velocities)
-      joint_trajectory.points[0].velocities[i] = 0;
+      point.velocities.at(i) = 0;
   }
 }
 
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -262,6 +262,9 @@ void ServoCalcs::run(const ros::TimerEvent& timer_event)
   // Do servoing calculations only if the robot should move, for efficiency
   // Create new outgoing joint trajectory command message
   auto joint_trajectory = moveit::util::make_shared_from_pool<trajectory_msgs::JointTrajectory>();
+  joint_trajectory->header.frame_id = parameters_.planning_frame;
+  joint_trajectory->header.stamp = ros::Time::now();
+  joint_trajectory->joint_names = internal_joint_state_.name;
 
   // Prioritize cartesian servoing above joint servoing
   // Only run commands if not stale and nonzero
@@ -566,10 +569,6 @@ void ServoCalcs::calculateJointVelocities(sensor_msgs::JointState& joint_state, 
 void ServoCalcs::composeJointTrajMessage(const sensor_msgs::JointState& joint_state,
                                          trajectory_msgs::JointTrajectory& joint_trajectory) const
 {
-  joint_trajectory.header.frame_id = parameters_.planning_frame;
-  joint_trajectory.header.stamp = ros::Time::now();
-  joint_trajectory.joint_names = joint_state.name;
-
   trajectory_msgs::JointTrajectoryPoint point;
   point.time_from_start = ros::Duration(parameters_.publish_period);
   if (parameters_.publish_joint_positions)
@@ -796,12 +795,10 @@ bool ServoCalcs::enforceSRDFPositionLimits()
 // Is handled differently for position vs. velocity control.
 void ServoCalcs::suddenHalt(trajectory_msgs::JointTrajectory& joint_trajectory)
 {
-  if (joint_trajectory.points.empty())
-  {
-    joint_trajectory.points.push_back(trajectory_msgs::JointTrajectoryPoint());
-    joint_trajectory.points[0].positions.resize(num_joints_);
-    joint_trajectory.points[0].velocities.resize(num_joints_);
-  }
+  joint_trajectory.points.clear();
+  joint_trajectory.points.push_back(trajectory_msgs::JointTrajectoryPoint());
+  joint_trajectory.points[0].positions.resize(num_joints_);
+  joint_trajectory.points[0].velocities.resize(num_joints_);
 
   for (std::size_t i = 0; i < num_joints_; ++i)
   {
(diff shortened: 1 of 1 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -796,12 +795,10 @@ bool ServoCalcs::enforceSRDFPositionLimits()
...
-  if (joint_trajectory.points.empty())
-  {
-    joint_trajectory.points.push_back(trajectory_msgs::JointTrajectoryPoint());
-    joint_trajectory.points[0].positions.resize(num_joints_);
-    joint_trajectory.points[0].velocities.resize(num_joints_);
-  }
+  joint_trajectory.points.clear();
+  joint_trajectory.points.push_back(trajectory_msgs::JointTrajectoryPoint());
+  joint_trajectory.points[0].positions.resize(num_joints_);
+  joint_trajectory.points[0].velocities.resize(num_joints_);
```
felixvd, line 801: I see. The failure mode would be a `joint_trajectory` with a point that had only the positions allocated, but not the velocities (or vice versa). Makes sense to flush the trajectory every time this is called.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -262,6 +262,9 @@ void ServoCalcs::run(const ros::TimerEvent& timer_event)
   // Do servoing calculations only if the robot should move, for efficiency
   // Create new outgoing joint trajectory command message
   auto joint_trajectory = moveit::util::make_shared_from_pool<trajectory_msgs::JointTrajectory>();
+  joint_trajectory->header.frame_id = parameters_.planning_frame;
```
felixvd, line 265: I see that these were called through `cartesianServoCals` or `jointServoCalcs` before. Why did they move up here? Is this related to the PR title?

felixvd (commented):
LGTM although I'm not sure if one of the changes really belongs in this PR.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -796,12 +795,10 @@ bool ServoCalcs::enforceSRDFPositionLimits()
...
 void ServoCalcs::suddenHalt(trajectory_msgs::JointTrajectory& joint_trajectory)
 {
-  if (joint_trajectory.points.empty())
-  {
-    joint_trajectory.points.push_back(trajectory_msgs::JointTrajectoryPoint());
-    joint_trajectory.points[0].positions.resize(num_joints_);
-    joint_trajectory.points[0].velocities.resize(num_joints_);
-  }
+  joint_trajectory.points.clear();
+  joint_trajectory.points.push_back(trajectory_msgs::JointTrajectoryPoint());
```
v4hn, line 799: nit
```suggestion
  joint_trajectory.points.emplace_back();
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
v4hn: filling in `ros::Time::now()` here slightly changes the stamp. I believe the concrete time stamp is relevant in ros_control, so maybe we should really fill this in only right before publishing the result?

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -796,12 +795,10 @@ bool ServoCalcs::enforceSRDFPositionLimits()
...
+  joint_trajectory.points[0].positions.resize(num_joints_);
+  joint_trajectory.points[0].velocities.resize(num_joints_);
 
   for (std::size_t i = 0; i < num_joints_; ++i)
   {
```
AndyZe, line 804: in this block, do you want to change [i] to .at(i)?  Would have made this a lot easier to track down

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: I agree with v4hn here... the Stamp is filled at L345, right before publishing. No need to do it here.

AndyZe (commented):
LGTM, and I tested it. I'll +1 after the [i] is changed to .at(i)

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: Yes, and I was thinking yesterday we should try to use `at(i)` by default wherever we can to avoid these sorts of issues in the future.  The runtime cost is not that bad if the array size is small.  If the array is large it will prevent vectorization.

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: Removed setting the timestamp.  I moved the header frame_id and joint_names here to initialize them right after creation, I did this mostly just to make it easier to reason about it being fully initialized.

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: The best part of at(i) is that it will throw an exception here if we try to access past the end of the array.

tylerjw:
I addressed the comments and rebased on master (to get the removal of the codecov patch check).  This is ready for another review.

moveit_ros/moveit_servo/src/servo_calcs.cpp
v4hn: That's the whole point of `at()`, yes. I'm undecided on whether to use `at()` or `operator[]`.
I would be surprised to see the former in MoveIt's most real-time capable module, but it's probably hardly measurable with our use-cases.

AndyZe (approved):

jliukkonen (approved):

moveit_ros/moveit_servo/src/servo_calcs.cpp
felixvd: Agree with @v4hn, would be good to have an idea of the cost. Remember someone might want this to run at 500-1000 Hz or more

AndyZe: Replacing it right here is 100% fine because it should only get called rarely. Not part of the main control loop.

I tested this out for the heck of it. On the master branch, the control loop rate varies from about 4.95 to 10.1 kHz. Occasional spikes as slow as 98.4 Hz.  :O

Blindly replacing [i] with .at(i) everywhere doesn't change much. 3.9 kHz - 11.29 kHz, with occasional spikes as slow as 98.6 Hz

felixvd: If you can add instructions to benchmark this (or maybe even automate it), [here](https://github.com/ros-planning/moveit/tree/master/moveit_ros/planning/planning_components_tools/src) would be a great place to put your script.

tylerjw: If I remove the switch to `at()` and the other unrelated refactor of the initialization of the other fields of `joint_trajectory` can we accept this change and revisit those in a future issue?  I would like to get the bug fix accepted and then move on the less trivial discussions around tradeoffs in implementation separately.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -796,22 +798,28 @@ bool ServoCalcs::enforceSRDFPositionLimits()
...
+  joint_trajectory.points.clear();
+  joint_trajectory.points.emplace_back();
+  trajectory_msgs::JointTrajectoryPoint& point = joint_trajectory.points.front();
+  point.positions.resize(num_joints_);
+  point.velocities.resize(num_joints_);
+
+  // Assert the following loop is safe to execute
+  assert(point.positions.size() >= num_joints_);
+  assert(point.velocities.size() >= num_joints_);
+  assert(original_joint_state_.position.size() >= num_joints_);
```
tylerjw, line 811: These will only be compiled in a debug build and will not affect runtime performance in release mode.

tylerjw:
@v4hn I removed the unrelated changes that were causing discussion so we can get this bug fix accepted.  :pleading_face:

tylerjw:
**Squash and merge** is how I'd like this accepted (PR includes fixup commits just so reviewers can see how I responded to comments).

moveit_ros/moveit_servo/src/servo_calcs.cpp
v4hn: I'm sorry but testing the specified effect of `vector::resize` does not make any sense! The last assert is probably reasonable, but please remove the others.

I approve already so you can merge after changing this.

v4hn (approved):
Please see my last nit.

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: > I'm sorry but testing the specified effect of `vector::resize` does not make any sense! The last assert is probably reasonable, but please remove the others.

Good point, fixed.  Thank you for reviewing this.

PR #2273, opened 2020-08-21
OMPL Constrained State Space Planning

JeroenDM:
### Description

In the context of a [Google Summer of Code project](https://github.com/ros-planning/moveit/issues/2092), I would like to add the constrained planning capabilities of OMPL to the existing interface. Previously, I split up the code in several pull requests. This was not entirely clear and impractical for testing. Hence this new approach, a single pull request that can be reviewed on a per commit basis.

**Edit** this pull request is quite large, but can be made smaller by first merging these tests separately:
- Add test to ompl interface for StateValidityChecker #2247 (already approved)
- WIP: add test for PlanningContextManager in ompl interface #2248 (needs some discussion)

### Why
It is one of the projects on the [future projects page](https://moveit.ros.org/documentation/contributing/future_projects/).

It can solve the long-standing [joint flip bug](https://github.com/ros-planning/moveit/issues/562) when planning with path constraints. It is especially useful in cases where the constrained region has zero volume in Cartesian space and the ` enforce_joint_model_state_space` (rejection sampling) does not work. For example planning with the end-effector constrained to a plane:

case_2

Or a line:

case_3

The current default planner that uses the `PoseModelStateSpace` can solve these problems, but the solution will often contain large jumps in joint space because it plans in Cartesian space. (Especially when adding obstacles in the scene.) This new planner plans in joint space. The example below illustrates the difference.

With the default `PoseModelStateSpace`:

case_4_pose_model

With OMPL's `ConstrainedStateSpace`:

case_4

Finally (and maybe most importantly) once this new planning approach is integrated, the possibilities for adding new constraints are endless. It would be really nice if we could pass arbitrary user-defined constraints to OMPL through some C++ interface in the future.

### Checklist

- [x] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [x] [Extend the tutorials.](https://github.com/ros-planning/moveit_tutorials/pull/518)
- [ ] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes
- [x] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html)
- [ ] Include a screenshot if changing a GUI
- [ ] While waiting for someone to review your request, please help review [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers

Issue #807, opened 2018-03-23
Connect rviz using moveit! to database


thisyearspairo:
### Description
Hello, I'm working with the Universal_Robot package in order to control the UR3 robot and I would like to know how to connect Rviz to a database in order to save the states I'm creating, because when I create them the program gives me: "Not connected to a database. The state will be created but not stored"
I'm using Ubuntu 16.04 and Ros Kinetic. 
Thank you very much in advance.

k-okada:
I think this is because of removing mongdb warehouse plugin, someone (including me) is tring to setup sql based plugin, if they have time ... https://github.com/ros-planning/moveit/issues/123

Krebsileinchen:
Are there any news about this topic? I started having the same issue. By now I installed MongoDB Warehouse and got a step further (http://moveit.ros.org/install/source/dependencies/). It starts connecting but then I get the error 
[ERROR] [1525254202.219795349]: Unable to connect to the database at '127.0.0.1:33829'. If you just created the database, it could take a while for initial setup.
When trying 
roslaunch MYROBOT_moveit_config default_warehouse_db.launch 
I get
[mongo_wrapper_ros_...] process has died [pid 980, exit code 1, cmd /home/../catkin_ws/src/warehouse_ros_mongo/src/mongo_wrapper_ros.py __name:=mongo_wrapper_ros_63_6095246583607931054 __log:=/home/../.ros/log/b3e2694e-4dcf-11e8-a39c-0010182ca071/mongo_wrapper_ros_sew_ws_963_6095246583607931054-1.log].
log file: /home/de6ak529/.ros/log/b3e2694e-4dcf-11e8-a39c-0010182ca071/mongo_wrapper_ros_.._6095246583607931054-1*.log
I would appreciate any ideas!

PR #2310, opened 2020-09-13
Clean up Rviz Motion Planning plugin, add tooltips

felixvd:
### Description

I had a go at making the plugin smaller. This version isn't perfect either, but what is? Here are the before and after images:

**Before:**
Screenshot from 2020-09-13 14-19-51 Screenshot from 2020-09-13 14-20-03

**After:**
Screenshot from 2020-09-13 14-21-00 Screenshot from 2020-09-13 14-24-09

Changes:  
- Added tooltips
- Reduced the minimum size of entry fields so they can partially hide numbers if required (users will resize the panel when they need to change the values). This should only affect the layout when the user resizes to _under_ the previous minimum size. The previous sizes should look the same as before.
- Reordered and renamed options in the Planning tab. Some of these had confused me when I started out. Most of these are just suggestions and up for discussion though. I just took the liberty while I was at it.

Notes:  
- I don't know if `allowReplanning` and `allowLooking` have any effect. I have a feeling that this is stub functionality added by Ioan, but that was never fully implemented. Because it seems like no one uses these checkboxes (and they are not mentioned in the tutorials afaik) I moved them down.
- There is a progress bar at the bottom that appears during planning, but not execution. It doesn't seem very useful. If it can't work during execution (I haven't looked into it), it might be better to remove it.
- I would actually like to turn off `Use Collision-Aware IK` by default. The robot looks like it freaks out whenever you move it into a collision, and you always lose your IK configuration. Opinions?
- [Qt SizePolicy documentation](https://doc.qt.io/archives/qt-4.8/qsizepolicy.html#Policy-enum) 

This should be cherry-picked to Kinetic and Melodic if possible.

### Checklist

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>507</width>
-    <height>389</height>
+    <height>380</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -26,7 +26,7 @@
       <bool>false</bool>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
      </property>
      <widget class="QWidget" name="context">
       <attribute name="title">
@@ -526,7 +526,7 @@
            <property name="sizeHint" stdset="0">
             <size>
              <width>20</width>
-             <height>41</height>
+             <height>32</height>
             </size>
            </property>
           </spacer>
@@ -586,7 +586,7 @@
            </property>
            <layout class="QVBoxLayout" name="verticalLayout_10">
             <property name="spacing">
-             <number>2</number>
+             <number>1</number>
             </property>
             <property name="bottomMargin">
              <number>4</number>
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -797,9 +797,9 @@
                </size>
               </property>
               <property name="toolTip">
-               <string>If checked, the IK solver tried to avoid collisions when searching for the state set via the interactive marker.
+               <string>If checked, the IK solver tries to avoid collisions when searching for a start/end state set via the interactive marker.
 
-This is usually achieved by random seeing, which can flip the robot configuration (e.g. elbow up/down). Note that motion planning is always collision-aware, regardless of this checkbox.</string>
+This is usually achieved by random seeding, which can flip the robot configuration (e.g. elbow up/down). Note that motion planning is always collision-aware, regardless of this checkbox.</string>
               </property>
               <property name="text">
                <string>Collision-aware IK</string>
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -26,7 +26,7 @@
       <bool>false</bool>
      </property>
      <property name="currentIndex">
-      <number>1</number>
+      <number>0</number>
      </property>
      <widget class="QWidget" name="context">
       <attribute name="title">
@@ -96,6 +96,9 @@
             </item>
             <item>
              <widget class="moveit_rviz_plugin::MotionPlanningParamWidget" name="planner_param_treeview">
+              <property name="autoScroll">
+               <bool>false</bool>
+              </property>
               <property name="indentation">
                <number>0</number>
               </property>
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -417,7 +417,7 @@
                  <bool>false</bool>
                 </property>
                 <property name="toolTip">
-                 <string>Send a stop() command to the move_group.</string>
+                 <string>Stop execution by sending a stop command to the move_group.</string>
                 </property>
                 <property name="text">
                  <string>&amp;Stop</string>
@@ -549,7 +549,7 @@
             <item>
              <widget class="QComboBox" name="path_constraints_combo_box">
               <property name="toolTip">
-               <string>Set path constraints from the database. </string>
+               <string>Select path constraints from the database. </string>
               </property>
              </widget>
             </item>
@@ -610,7 +610,7 @@
               <item>
                <widget class="QLabel" name="label_9">
                 <property name="toolTip">
-                 <string>The maximum amount of time the planner is allowed to take. This is the total amount, not the amount of time per planning attempt.</string>
+                 <string>Total time allowed for planning.  Planning stops if time or number of attempts is exceeded, or goal is reached.</string>
                 </property>
                 <property name="text">
                  <string>Planning Time (s):</string>
@@ -652,7 +652,7 @@
               <item>
                <widget class="QLabel" name="label_10">
                 <property name="toolTip">
-                 <string>Maximum allowed number of planning attempts.</string>
+                 <string>Allowed number of planning attempts. Planning stops if time or number of attempts is exceeded, or goal is reached.</string>
                 </property>
                 <property name="text">
                  <string>Planning Attempts:</string>
@@ -691,7 +691,7 @@
               <item>
                <widget class="QLabel" name="label_11">
                 <property name="toolTip">
-                 <string>Velocity scaling (applied to each joint)</string>
+                 <string>Factor (between 0 and 1) that scales maximum joint velocities</string>
                 </property>
                 <property name="text">
                  <string>Velocity Scaling:</string>
@@ -727,7 +727,7 @@
               <item>
                <widget class="QLabel" name="label_20">
                 <property name="toolTip">
-                 <string>Acceleration scaling (applied to each joint)</string>
+                 <string>Factor (between 0 and 1) that scales maximum joint accelerations</string>
                 </property>
                 <property name="text">
                  <string>Accel. Scaling:</string>
@@ -770,7 +770,9 @@
                </size>
               </property>
               <property name="toolTip">
-               <string>Check this to generate a linear path in Cartesian (3D) space</string>
+               <string>Check this to generate a linear path in Cartesian (3D) space.
+
+ This does not plan around obstacles. </string>
               </property>
               <property name="text">
                <string>Use Cartesian Path</string>
@@ -795,10 +797,12 @@
                </size>
               </property>
               <property name="toolTip">
-               <string>If checked, a new IK solution is calculated whenever the robot state is in collision. This setting only affects the state set via the interactive marker. Planning is always aware of collisions.</string>
+               <string>If checked, the IK solver tried to avoid collisions when searching for the state set via the interactive marker.
+
+This is usually achieved by random seeing, which can flip the robot configuration (e.g. elbow up/down). Note that motion planning is always collision-aware, regardless of this checkbox.</string>
               </property>
               <property name="text">
-               <string>Reset IK on collision</string>
+               <string>Collision-aware IK</string>
               </property>
               <property name="checked">
                <bool>false</bool>
@@ -820,10 +824,35 @@
                </size>
               </property>
               <property name="toolTip">
-               <string>Allow approximate IK solutions</string>
+               <string>Allow approximate IK solutions, which deviate from the target pose. Useful when an exact solution cannot be found, e.g. for underactuated arms or when trying to reach an unreachable target.</string>
               </property>
               <property name="text">
                <string>Approx IK Solutions</string>
+              </property>
+             </widget>
+            </item>
+            <item>
+             <widget class="QCheckBox" name="allow_external_program">
+              <property name="sizePolicy">
+               <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
+                <horstretch>0</horstretch>
+                <verstretch>0</verstretch>
+               </sizepolicy>
+              </property>
+              <property name="minimumSize">
+               <size>
+                <width>30</width>
+                <height>0</height>
+               </size>
+              </property>
+              <property name="toolTip">
+               <string>Allow controlling the interactive marker through ROS topics (under /rviz/moveit/marker/ ) published other nodes.</string>
+              </property>
+              <property name="text">
+               <string>External Comm.</string>
+              </property>
+              <property name="checked">
+               <bool>false</bool>
               </property>
              </widget>
             </item>
@@ -842,7 +871,10 @@
                </size>
               </property>
               <property name="toolTip">
-               <string>Allows the robot to replan if the plan to be executed collides with the environment.</string>
+               <string>EXPERIMENTAL: Trigger replanning if new collisions are detected during plan execution. Only works for the &quot;Plan &amp; Execute&quot; button. </string>
+              </property>
+              <property name="autoFillBackground">
+               <bool>false</bool>
               </property>
               <property name="text">
                <string>Replanning</string>
@@ -867,35 +899,10 @@
                </size>
               </property>
               <property name="toolTip">
-               <string>Allows the robot to reposition its sensor to look around. A sensor needs to be set up and the planning algorithm has to support this, otherwise it has no effect.</string>
+               <string>EXPERIMENTAL: Allows the robot to reposition its sensor to look around to resolve apparent collisions. A sensor needs to be set up and the new measurements have to affect the collision scene, otherwise it has no effect. See tutorial.</string>
               </property>
               <property name="text">
                <string>Sensor Positioning</string>
-              </property>
-              <property name="checked">
-               <bool>false</bool>
-              </property>
-             </widget>
-            </item>
-            <item>
-             <widget class="QCheckBox" name="allow_external_program">
-              <property name="sizePolicy">
-               <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
-                <horstretch>0</horstretch>
-                <verstretch>0</verstretch>
-               </sizepolicy>
-              </property>
-              <property name="minimumSize">
-               <size>
-                <width>30</width>
-                <height>0</height>
-               </size>
-              </property>
-              <property name="toolTip">
-               <string>Allow communication with an external program that sends an interactive marker position to control the robot.</string>
-              </property>
-              <property name="text">
-               <string>External Comm.</string>
               </property>
               <property name="checked">
                <bool>false</bool>
@@ -2166,7 +2173,6 @@
   <tabstop>acceleration_scaling_factor</tabstop>
   <tabstop>allow_replanning</tabstop>
   <tabstop>allow_looking</tabstop>
-  <tabstop>allow_external_program</tabstop>
   <tabstop>path_constraints_combo_box</tabstop>
   <tabstop>goal_tolerance</tabstop>
   <tabstop>place_button</tabstop>
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -523,7 +523,7 @@
            <property name="sizeHint" stdset="0">
             <size>
              <width>20</width>
-             <height>6</height>
+             <height>41</height>
             </size>
            </property>
           </spacer>
@@ -552,20 +552,6 @@
                <string>Select path constraints from the database. </string>
               </property>
              </widget>
-            </item>
-            <item>
-             <layout class="QHBoxLayout" name="horizontalLayout_11">
-              <item>
-               <widget class="QLabel" name="label_6">
-                <property name="text">
-                 <string>Goal Tolerance:</string>
-                </property>
-               </widget>
-              </item>
-              <item>
-               <widget class="QDoubleSpinBox" name="goal_tolerance"/>
-              </item>
-             </layout>
             </item>
            </layout>
           </widget>
@@ -2174,7 +2160,6 @@ This is usually achieved by random seeding, which can flip the robot configurati
   <tabstop>allow_replanning</tabstop>
   <tabstop>allow_looking</tabstop>
   <tabstop>path_constraints_combo_box</tabstop>
-  <tabstop>goal_tolerance</tabstop>
   <tabstop>place_button</tabstop>
   <tabstop>roi_center_x</tabstop>
   <tabstop>roi_center_y</tabstop>
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
@@ -2160,21 +2160,35 @@ This is usually achieved by random seeding, which can flip the robot configurati
   <tabstop>database_port</tabstop>
   <tabstop>reset_db_button</tabstop>
   <tabstop>database_connect_button</tabstop>
+  <tabstop>wcenter_x</tabstop>
+  <tabstop>wcenter_y</tabstop>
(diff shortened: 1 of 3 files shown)


moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -704,59 +756,149 @@
...
-              <property name="checked">
-               <bool>false</bool>
+              <property name="minimumSize">
+               <size>
+                <width>30</width>
+                <height>0</height>
+               </size>
+              </property>
+              <property name="toolTip">
+               <string>Allow approximate IK solutions</string>
```
rhaschke, line 823:
```suggestion
               <string>Allow approximate IK solutions. This might be helpful if an exact solution cannot be found, e.g. for an underactuated arm.</string>
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -704,59 +756,149 @@
...
+               </sizepolicy>
+              </property>
+              <property name="minimumSize">
+               <size>
+                <width>30</width>
+                <height>0</height>
+               </size>
+              </property>
+              <property name="toolTip">
+               <string>Allows the robot to replan if the plan to be executed collides with the environment.</string>
```
rhaschke, line 845: I guess this option refers to online monitoring of the executed trajectory and triggering replanning if a new collision pops up due to a changed scene. This should be clarified.
I'm not sure when this checkbox is considered at all. Only for `Plan & Execute`?
```suggestion
               <string>Trigger replanning if new collisions pop up during plan execution.</string>
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -704,59 +756,149 @@
...
+               </sizepolicy>
+              </property>
+              <property name="minimumSize">
+               <size>
+                <width>30</width>
+                <height>0</height>
+               </size>
+              </property>
+              <property name="toolTip">
+               <string>Allows the robot to reposition its sensor to look around. A sensor needs to be set up and the planning algorithm has to support this, otherwise it has no effect.</string>
```
rhaschke, line 870: Again, referring to a tutorial might be helpful here.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -704,59 +756,149 @@
...
+               </sizepolicy>
+              </property>
+              <property name="minimumSize">
+               <size>
+                <width>30</width>
+                <height>0</height>
+               </size>
+              </property>
+              <property name="toolTip">
+               <string>Allow communication with an external program that sends an interactive marker position to control the robot.</string>
```
rhaschke, line 895:
```suggestion
               <string>Allow control of start and end poses as well as triggering of planning (and execution?) from external programs via ROS topics.</string>
```

rhaschke (commented):
I like these changes very much, particularly the reordering of checkboxes.
I'm also fine with reducing the minimal size of text fields - as long as the initial/preferred size still shows all content.
See inline comments for some suggestions regarding the tooltips.

v4hn:
A small lobbying comment: Why not incorporate https://github.com/ros-planning/moveit/pull/2007 here @felixvd? :-)
I still think that pr should have been merged.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -538,7 +547,11 @@
              <number>6</number>
             </property>
             <item>
-             <widget class="QComboBox" name="path_constraints_combo_box"/>
+             <widget class="QComboBox" name="path_constraints_combo_box">
+              <property name="toolTip">
+               <string>Set path constraints from the database. </string>
```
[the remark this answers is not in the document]
v4hn: Adding links to tooltips does not make sense. An appropriate link could of course be added to the RViz tutorial(s) if it does not exist there yet.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -704,59 +756,149 @@
...
+                <width>30</width>
+                <height>0</height>
+               </size>
+              </property>
+              <property name="toolTip">
+               <string>If checked, a new IK solution is calculated whenever the robot state is in collision. This setting only affects the state set via the interactive marker. Planning is always aware of collisions.</string>
+              </property>
               <property name="text">
-               <string>Allow Sensor Positioning</string>
+               <string>Reset IK on collision</string>
```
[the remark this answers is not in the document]
v4hn: Please keep the name as-is. "Reset IK on collision" is technically also wrong, because what exactly happens depends on the IK solver.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
v4hn:
```suggestion
               <string>Accept IK solutions, even if they deviate from the target pose. While technically incorrect, these solutions are often still intuitive when moving the interactive marker to unreachable targets.</string>
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
v4hn: >  I'm not sure when this checkbox is considered at all. Only for Plan & Execute?

Yes.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
v4hn:
```suggestion
               <string>Allow to reposition a sensor to resolve apparent collisions. A sensor needs to be set up and the new measurements have to affect the collision scene, otherwise it has no effect.</string>
```

v4hn:
I would propose to move "Replanning" and "Sensor Positioning" all the way to the bottom of the list and explicitly add something like `Experimental Feature:` to the beginning of the resp. tooltips.
I know these have been experimental for years, but nobody works on them and most maintainers don't even know what they do. :-)

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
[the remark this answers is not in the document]
felixvd: I don't actually know how to set it up. We don't have a tutorial, and googling "moveit database" turns up [completely unrelated content](https://docs.ipswitch.com/MOVEit/Transfer2018/Help/Admin/en/index.htm#23428.htm) and the stale issue #807.  I have a strong feeling the database feature is just dead. Maybe something to discuss at the next meeting.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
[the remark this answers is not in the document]
felixvd: Ok. I had changed it because "Collision-aware" made me think "Why would I ignore collisions? Is this dangerous?" when I first saw it, but if it comes up elsewhere then I'm fine with it.

felixvd:
Thanks for the helpful comments and history lessons :) Especially helpful because I also had to guess what some of the features were. I merged the suggestions in a way that should address the main concerns.

I'm happy to tweak the Path Constraints area while I'm at it, especially because it is useless without the extra work of setting up a database, but now that @JeroenDM has put so much work into #2273 , I would prefer to make it visible by making that part of the UI useful rather than removing it.

How about providing default constraints in the dropdown list, such as:  
- No end effector rotation
- X-rotation unconstrained, YZ fixed
- Y-rotation unconstrained, XZ fixed
- Z-rotation unconstrained, XY fixed  
- Keep EE link within sphere containing start and goal
- Keep EE link within 5 cm around linear path to goal

Currently the constraints are defined by passing a string to the move_group. For the default constraints, I would create the constraint messages in the plugin and hand those to the move_group. Additional constraint choices could still be populated from a database, even though I doubt that feature is used by anyone, as mentioned in the comment.

@JeroenDM If you write expressions for the default constraints given a start/goal state, I'll put it into the GUI.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
felixvd: Ah I see, that's what happens. It looks like it only makes the plugin subscribe to the interactive marker topics with [this callback function](https://github.com/ros-planning/moveit/blob/master/moveit_ros/robot_interaction/src/robot_interaction.cpp#L674), but that doesn't trigger a plan or execution. I changed the text, please have a look.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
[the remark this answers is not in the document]
felixvd: I guess [this](http://wiki.ros.org/warehouse_ros/Tutorials) comes closest? [Only this link comes up in our tutorials](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/benchmarking/benchmarking_tutorial.html), but it just "How to install dependencies" and not very helpful. I'm not sure if this feature is worth maintaining at all to be honest.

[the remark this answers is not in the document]
rhaschke: @v4hn, Is [this tutorial](https://ros-planning.github.io/moveit_tutorials/doc/planning_with_approximated_constraint_manifolds/planning_with_approximated_constraint_manifolds_tutorial.html) you recently contributed related?
Skimming through this tutorial, it seems to be related to constraints. But the database created seems to be unrelated to the mongo database used e.g. for poses and scenes. Can you confirm this?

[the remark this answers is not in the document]
rhaschke: > Adding links to tooltips does not make sense.

I agree, tooltips cannot support links. However, referring to a tutorial would be very helpful I think.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
rhaschke: I never used this feature and it is completely unclear to me, which strategy is used to reposition the sensor.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
rhaschke: The checkbox triggers the following function, which also enables topics to trigger planning and execution:
https://github.com/ros-planning/moveit/blob/91e1f095c2e68187e19f975951011c20f0d8b51e/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp#L257-L279

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -691,7 +691,7 @@
               <item>
                <widget class="QLabel" name="label_11">
                 <property name="toolTip">
-                 <string>Velocity scaling (applied to each joint)</string>
+                 <string>Factor (between 0 and 1) that scales maximum joint velocities</string>
```
rhaschke, line 694: To emphasize that maximal velocities can only be reduced:
```suggestion
                 <string>Factor (between 0 and 1) to scale down maximum joint velocities</string>
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -727,7 +727,7 @@
               <item>
                <widget class="QLabel" name="label_20">
                 <property name="toolTip">
-                 <string>Acceleration scaling (applied to each joint)</string>
```
rhaschke: Same here.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -795,10 +797,12 @@
                </size>
               </property>
               <property name="toolTip">
-               <string>If checked, a new IK solution is calculated whenever the robot state is in collision. This setting only affects the state set via the interactive marker. Planning is always aware of collisions.</string>
+               <string>If checked, the IK solver tried to avoid collisions when searching for the state set via the interactive marker.
+
+This is usually achieved by random seeing, which can flip the robot configuration (e.g. elbow up/down). Note that motion planning is always collision-aware, regardless of this checkbox.</string>
```
rhaschke, line 802:
```suggestion
               <string>If checked, the IK solver tries to avoid collisions when searching for a start/end state set via the interactive marker.

This is usually achieved by random seeding, which can flip the robot configuration (e.g. elbow up/down). Note that motion planning is always collision-aware, regardless of this checkbox.</string>
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
```diff
@@ -842,18 +846,18 @@
                </size>
               </property>
               <property name="toolTip">
-               <string>Allows the robot to replan if the plan to be executed collides with the environment.</string>
+               <string>Allow controlling the interactive marker through ROS topics (under /rviz/moveit/marker/ ) published other nodes.</string>
```
rhaschke, line 849: See my previous comment. Triggering planning and execution becomes also possible.
Not sure though, this is a useful feature. If I want to control the robot from an app, I would directly talk to the move_group node instead...

rhaschke (approved):
I fixed some types. Otherwise looks good to me.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
felixvd: I thought I changed that. I must have forgotten to save. Will fix tomorrow.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
v4hn: The logic is quite straight-forward: If the cost of obstacles close to the path is too high, direct the sensor towards a cost source. It was used to get rid of kinect-based artefact in the octomap in the past and is implemented [here](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/plan_execution/src/plan_with_sensing.cpp#L206).

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
v4hn: Users of this feature reported in github issues before. I agree this is not ideal, but I guess it can simplify User-Machine-Interaction pipelines.

v4hn:
> I'm happy to tweak the Path Constraints area while I'm at it (it's annoying me that it is useless without the extra work of setting up a database) but now that @JeroenDM has put so much work into #2273 , I would prefer to make it visible by making that part of the UI useful rather than removing it.

Nobody ever talked about removing path constraints! The whole argument was against the (still unused) "goal tolerances" field.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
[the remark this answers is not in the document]
v4hn: This is warehouse functionality and not related to the constraint approximation database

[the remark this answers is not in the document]
felixvd: I had no idea those two were different.

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
felixvd: I figured it could be useful for some interface, even if it seems a bit strange. I updated the comment.

felixvd:
You're right. I misread because the other PR is called "Goal Constraints". 

I removed the goal constraints and moved the Clear Octomap button down, since it shouldn't be in the Commands section to start with. I also made the distances in the Query group a little bit smaller to match the size to the Commands group:

Screenshot 2020-09-20 00:04:03

I won't do anything about the Path Constraints in this PR, but the offer stands @JeroenDM .

moveit_ros/visualization/motion_planning_rviz_plugin/src/ui/motion_planning_rviz_plugin_frame.ui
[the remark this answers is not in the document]
v4hn: the warehouse refers to the mongodb that can store a number of MoveIt-related datastructures.

The constraint approximation database stores a set of *joint states* that approximates the manifold of a specific constraint.

rhaschke (commented):
I do consider "Clear Octomap" a command, though it's not a planning command. So, I suggest bringing this button back into the commands group.
Also, you suggested changing the default of "Collision-Aware IK" to unticked. That's not yet done.

JeroenDM:
> @JeroenDM If you write expressions for the default constraints given a start/goal state, I'll put it into the GUI.

I like your list of example default constraints! Ideally, these could be implemented independently of which planner plugin is used. (So we can even reuse them for TrajOpt in the future.) But I'm not sure where these default implementations would go.

It would look like a factory method that takes in a start and goal state and spits out a [moveit_msgs/Constraints](http://docs.ros.org/latest/api/moveit_msgs/html/msg/Constraints.html) message.

But maybe we should move this discussion to a separate issue?

felixvd:
> Also, you suggested changing the default of "Collision-Aware IK" to unticked. That's not yet done.

The checkbox is inactive for me when I add a Motion Planning display in a new `panda_moveit_config demo.launch` Rviz window, and when I open the ui file in QT Creator. Is there something else that needs to be changed?

> I do consider "Clear Octomap" a command, though it's not a planning command. So, I suggest bringing this button back into the commands group.

It is a command in the same way that  "Clear Planning Scene" is, but it seems very different from the ones that generate or execute a robot trajectory. Grouping them together doesn't seem intuitive, so I assumed the placement was due to the lack of space. I am not dead set on this layout, but I think separating it makes sense. Maybe someone else has an opinion?

> It would look like a factory method that takes in a start and goal state and spits out a moveit_msgs/Constraints message. But maybe we should move this discussion to a separate issue?

Yes, to both. Feel free to start a PR. I think you still have access to my fork.

v4hn:
> Maybe someone else has an opinion?

In your first version you left some space above `Clear Octomap` as well. For me that suffices to make it clear they refer to different things.
The alternative layout makes the button much bigger than it should be, given that few people use it.
I would prefer it if we had another button right next to it to share the space, but I can't think of useful functionality to add there.

Imho, the more annoying issue with the button is that it is active *even when there is no octomap in the planning scene*.

rhaschke:
> The checkbox was ticked in my last screenshot because I ran `moveit_resources_fanuc_moveit_config demo.launch`, which loads the plugin from its `.rviz` file, where the checkbox is saved that way.

That screenshot was indeed confusing to me.

felixvd:
> In your first version you left some space above `Clear Octomap` as well. For me that suffices to make it clear they refer to different things.

That's the text/label where the planning result appears, so to me there's still no semantic separation. It's true that it's a space though.

How about a "Reset Goal/Start" button next to the Octomap button? Sometimes clicking the dropdown menus to select "Current" is tedious, especially if the GUI bugs out and garbles the options. 

> Imho, the more annoying issue with the button is that it is active _even when there is no octomap in the planning scene_.

Yes! Let me fix that before merging this.

v4hn:
> How about a "Reset Goal/Start" button next to the Octomap button?
> Sometimes clicking the dropdown menus to select "Current" is tedious, especially if the GUI bugs out and garbles the options.

I'm undecided here. Guess I would use the button though instead of clicking "current" explicitly. :-)
Of course, if the GUI bugs out, we should fix that bug instead. ;D

> Yes! Let me fix that before merging this.

:+1:

rhaschke:
> The more annoying issue with the button is that it is active _even when there is no octomap in the planning scene_.

I agree, but I think that should be tackled in a separate PR.

> How about a "Reset Goal/Start" button next to the Octomap button?

The semantics of this button isn't apparent to me: Do you want to reset to current?

> Sometimes clicking the dropdown menus to select "Current" is tedious, especially if the GUI bugs out and garbles the options.

Can you reproduce this somehow? Of course, would be great to fix any bug there... (in a separate PR as well).

felixvd:
> The semantics of this button isn't apparent to me: Do you want to reset to current?

Yes, that's what I thought.

> Can you reproduce this somehow? Of course, would be great to fix any bug there... (in a separate PR as well).

Sadly I've never been able to reproduce it. It appears randomly from what I can tell, and restarting Rviz generally fixes the problem. It seems to be some issue with Qt/the GUI pipeline, not this plugin.

> I'm undecided here. Guess I would use the button though instead of clicking "current" explicitly. :-)

I'm also not convinced that it's necessary, but it might be convenient and save some clicks.

felixvd:
The button enabling/disabling seemed to work when I tested it with `roslaunch moveit_tutorials detect_and_add_cylinder_collision_object_demo.launch` from [this tutorial](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/perception_pipeline/perception_pipeline_tutorial.html), but the octomap wasn't cleared when I pressed the button. Does it work for you?

v4hn:
This might be an independent hiccup. I do remember an issue about the cleared (empty) octomap not showing up in RViz although it was actually cleared.
The tutorial keeps publishing the *same* octomap every 10 seconds, so it might be that you just never see a change but everything else is fine.
I agree with @rhaschke by now that it makes sense to look into this in a separate pull-request. Thanks for working on it! Could you migrate it to a separate PR? :-)

felixvd:
Yes, but even when I cleared it, the Collision-Aware IK was considering the Octomap. Come to think of it, it's probably because the octomap is cleared via the service of the move group capability, but the planning scene inside the plugin isn't updated. Looking a bit closer, [this function](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L532) doesn't remove the Octomap from the world, but only from the `octomap_monitor`(which is optional...?), and doesn't trigger a scene update.

rhaschke:
I'm still in favour of moving the "clear octomap" button back into the commands group.

felixvd:
Eh, I don't care to make that new button, so I moved the Clear Octomap one back and just reduced the spacings. If that looks too cramped, just revert the last commit.

Screenshot 2020-09-23 00:41:46

Does anyone know why the two columns match up in QtCreator but not in my Rviz window? That issue was already there for me on the master branch.

rhaschke (approved):

rhaschke:
> Does anyone know why the bottoms of the two columns match up in QtCreator but not in my Rviz window?

Probably QtCreator and rviz use different fonts. The vertical space between commands and path constraints is a fixed spacer and thus doesn't adapt when resizing the window. Moving the spacers below those columns to a reasonable position, solves the problem:
image image

v4hn:
Looks good to me. Thanks for refining minute details such as checkbox order, tab stops, et cetera as well!

PR #2313, opened 2020-09-14
planning context is already loaded in move_group.launch

ymollard:
It looks that the `planning_context.launch` is included twice when running `demo*.launch` (in `move_group.launch` and `demo*.launch`. Here's a simplification proposal.


rhaschke (approved):
LGTM

tylerjw (approved):
I tested this change and it seems to work for me.  Good catch.

PR #2314, opened 2020-09-14
Fix Servo interruption

AndyZe:
Tweak the interruption logic

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -258,7 +258,7 @@ private:
   bool ok_to_publish_ = false;
   double collision_velocity_scale_ = 1.0;
 
-  std::atomic_bool stop_requested_{false};
+  std::atomic_bool stop_requested_{ false };
 
   // Use ArrayXd type to enable more coefficient-wise operations
   Eigen::ArrayXd delta_theta_;
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -252,12 +252,13 @@ private:
 
   // Status
   StatusCode status_ = StatusCode::NO_WARNING;
-  bool stop_requested_ = false;
   bool paused_ = false;
   bool twist_command_is_stale_ = false;
   bool joint_command_is_stale_ = false;
   bool ok_to_publish_ = false;
   double collision_velocity_scale_ = 1.0;
+
+  std::atomic_bool stop_requested_{false};
 
   // Use ArrayXd type to enable more coefficient-wise operations
   Eigen::ArrayXd delta_theta_;
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -66,9 +66,8 @@ public:
              const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor,
              const std::shared_ptr<JointStateSubscriber>& joint_state_subscriber);
 
-  /** \brief Start and stop the timer where we do work and publish outputs */
+  /** \brief Start the timer where we do work and publish outputs */
   void start();
-  void stop();
 
   /**
    * Get the MoveIt planning link transform.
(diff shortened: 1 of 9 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -207,13 +207,16 @@ void ServoCalcs::run(const ros::TimerEvent& timer_event)
...
-    if (stop_requested_)
-      return;
     default_sleep_rate_.sleep();
   }
 
+  if (stop_requested_)
+  {
+    return;
+  }
+
```
tylerjw, line 219: Is the fix basically that if ros::ok() is false amd stop_requested_ is true this continues instead of returning?

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -335,7 +338,7 @@ void ServoCalcs::run(const ros::TimerEvent& timer_event)
     zero_velocity_count_ = 0;
   }
 
-  if (ok_to_publish_)
+  if (ok_to_publish_ && !stop_requested_)
   {
     // Put the outgoing msg in the right format
```
tylerjw, line 343: Or is it that sometimes we'll publish one more message than we should be after stop is requested because it happened while we were executing this?

tylerjw (commented):
These changes look ok.  I'm trying to clarify what they fix.  Also, you need to run clang_format to get it to pass travis.

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: yep. Also, if updateJoints was successful immediately, there wasn't a check on stop_requested_

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: yep, that was also an issue

AndyZe:
We could also discuss if we need all three of these:

- start()
- stop()
- setPaused()

The difference seems to be, setPaused() allows the thread to keep spinning so the joints stay up to date. There is more overhead involved with stop() but it stops spinning completely

AndyZe:
I'm working on making the tests more repeatable now. It seems like the issue had to do with initializing ROS in one test file for multiple tests. Separate executables for each test is 100% reliable from what I've seen so far. Will work on reducing duplicate code next.

AndyZe:
OK, this is working well locally. I stuck a header file (`common_test_setup.h`) in the the `/test` folder. Not sure if there's any convention for where to put header files related to testing.

tylerjw (commented):
I really like what you are doing with the tests here.  However, I feel that they don't belong in this PR.  Would you mind splitting the changes to the tests into a separate PR from the one to change the logic for stopping?

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -252,13 +251,13 @@ class ServoCalcs
...
   // Status
   StatusCode status_ = StatusCode::NO_WARNING;
-  bool stop_requested_ = false;
-  bool paused_ = false;
   bool twist_command_is_stale_ = false;
   bool joint_command_is_stale_ = false;
   bool ok_to_publish_ = false;
   double collision_velocity_scale_ = 1.0;
 
+  std::atomic_bool paused_{ false };
```
tylerjw, line 259: This is basically the reason I've was hesitant to approve this pr when it was small.  As far as I could tell this value does not need to be atomic.  As far as I can tell this is written to by one thread and read from the other and there is no RMU (read-modify-update) operations done on it.

tylerjw:
It seems that there was some issue caused by the change of logic in the tests that prompted refactoring the tests.  Could you help me understand what the reason was?

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
AndyZe: done!

AndyZe:
> It seems that there was some issue caused by the change of logic in the tests that prompted refactoring the tests. Could you help me understand what the reason was?

Thanks for asking this question! It turns out the deletion of timer_.stop() is what was causing the flaky tests. I put that back in the destructor and tests are back to normal.

tylerjw (approved):
lgtm, tested with customer setup.

henningkayser (approved):

PR #2315, opened 2020-09-14
Add comment to MOVEIT_CLASS_FORWARD

felixvd:
### Description

This adds a comment to every call of `MOVEIT_CLASS_FORWARD`, making it easier to discover for new users, and allowing e.g. `RobotModelPtr` to be found through text search. I think the name of the macro could be more expressive, and I would prefer if all the definitions could be found by text, but this change should be helpful already.

This question came up on the Discord/[ROS Answers](https://answers.ros.org/question/361537/where-is-the-definition-of-the-ptr-naming-standard/), and @tylerjw and I both remember having trouble figuring this out ourselves.

The regex if you want to change wording: `find . -name '*.h' -or -name '*.hpp' -or -name '*.cpp' | xargs sed -i -E 's/(MOVEIT_CLASS_FORWARD)\(([A-Za-z]+)\);/\1(\2);  \/\/ Defines \2Ptr, ConstPtr, WeakPtr... etc/g' $1`

### Checklist

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h
@@ -47,7 +47,7 @@
 
 namespace collision_detection
 {
-MOVEIT_CLASS_FORWARD(AllowedCollisionMatrix);
+MOVEIT_CLASS_FORWARD(AllowedCollisionMatrix);  // Defines AllowedCollisionMatrixPtr, ConstPtr, WeakPtr... etc
 
 /** \brief The types of bodies that are considered for collision */
 namespace BodyTypes
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_detector_allocator.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_detector_allocator.h
@@ -41,7 +41,7 @@
 
 namespace collision_detection
 {
-MOVEIT_CLASS_FORWARD(CollisionDetectorAllocator);
+MOVEIT_CLASS_FORWARD(CollisionDetectorAllocator);  // Defines CollisionDetectorAllocatorPtr, ConstPtr, WeakPtr... etc
 
 /** \brief An allocator for a compatible CollisionWorld/CollisionRobot pair. */
 class CollisionDetectorAllocator
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_env.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_env.h
@@ -45,7 +45,7 @@
 
 namespace collision_detection
 {
-MOVEIT_CLASS_FORWARD(CollisionEnv);
+MOVEIT_CLASS_FORWARD(CollisionEnv);  // Defines CollisionEnvPtr, ConstPtr, WeakPtr... etc
 
 /** \brief Provides the interface to the individual collision checking libraries. */
 class CollisionEnv
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_matrix.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_matrix.h
@@ -71,7 +71,7 @@ enum Type
  * CONDITIONAL) */
 typedef boost::function<bool(collision_detection::Contact&)> DecideContactFn;
 
-MOVEIT_CLASS_FORWARD(AllowedCollisionMatrix);
+MOVEIT_CLASS_FORWARD(AllowedCollisionMatrix);  // Defines AllowedCollisionMatrixPtr, ConstPtr, WeakPtr... etc
 
 /** @class AllowedCollisionMatrix
  *  @brief Definition of a structure for the allowed collision matrix. All elements in the collision world are referred
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_plugin.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_plugin.h
@@ -40,7 +40,7 @@
 
 namespace collision_detection
 {
-MOVEIT_CLASS_FORWARD(CollisionPlugin);
+MOVEIT_CLASS_FORWARD(CollisionPlugin);  // Defines CollisionPluginPtr, ConstPtr, WeakPtr... etc
 
 /**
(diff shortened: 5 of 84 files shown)


tylerjw (approved):
Thank you, this seems useful to me (and hopefully will help new users of the code).

rhaschke (approved):

rhaschke:
You should really use a C++ capable IDE, which would guide you to the macro definition.

gavanderhoorn:
> You should really use a C++ capable IDE, which would guide you to the macro definition.

Ha :) that's exactly what I commented on [the ROS Answers question](https://answers.ros.org/question/361537) as well.

v4hn:
It's still a valid approach to use an editor that does not overheat your laptop, lags a lot or is actually updated to current versions. ;)
But if you have to ask about where things are then, you should consider alternatives.

I'm actually surprised to see this patch merged, as I would have rejected it. :-)

tylerjw:
I think a better approach would be to remove shared_ptr from everywhere except when the management of the lifetime of something explicitly needs to be shared.  Also, if you are using an editor that has decent tab-completion the syntax for `std::shared_ptr<thing>` really isn't that bad to type.  Even vim does this well enough for me.  If your fancy editor can't tab-complete a common word in the language maybe you should consider a better editor?  :P

Generally, the less we can use of macros that modify the code in the precompiler and the more we can use of modern c++ types the better imo.  I know these are baked into MoveIt everywhere and everyone's code that uses them, I for one find them odd and not particularly helpful (I think they encourage using shared_ptr in places where it shouldn't be and it was annoying for me to originally find the macro).

_rand over_

v4hn:
> remove shared_ptr from everywhere except when the management of the lifetime of something explicitly needs to be shared.

Yes, MoveIt is full of unnecessary shared_ptr's. So is ROS in general. Roboticists often don't want to think about memory. But then our APIs also require them in places, so it's not that easy to reduce them even in your own code. At least it's not as bad as PCL yet :laugh:

> Also, if you are using an editor that has decent tab-completion the syntax for `std::shared_ptr<thing>` really isn't that bad to type.

The main reason for adding aliases in many places a few years back was to support both `boost::shared_ptr` and `std::shared_ptr` without relying on `auto`. Believe it or not, this mix was still an issue for me in a workspace last week. But aside from that I also find `RobotStatePtr` much nicer to read than `std::shared_ptr<RobotState>` and I'm sure it's not just me. Remember, it's about *readable*, not about *writable*. :-)

> Even vim does this well enough for me.

`<Ctrl>+n`, `+p` work well enough, but YouCompleteMe can take it's sweet time on my system. I'm happy to try out other plugin recommendations. :-)

> If your fancy editor can't tab-complete a common word in the language maybe you should consider a better editor?  :P

Personally, I run with the ros_qtc_plugin for some projects. But on an i5 laptop it's quite slow, randomly shows high UI latencies and is broken with newer software versions for a while now.

Sorry for going off-topic so far.

tylerjw:
> Remember, it's about *readable*, not about *writable*. :-)

That is a fair point, I mostly just don't like that `shared_ptr` is the **default** pointer type.

> `<Ctrl>+n`, `+p` work well enough, but YouCompleteMe can take it's sweet time on my system. I'm happy to try out other plugin recommendations. :-)

I use [clang-complete](https://github.com/xavierd/clang_complete).  I find it much faster than YouCompleteMe.  ymmv 

> Yes, MoveIt is full of unnecessary shared_ptr's. So is ROS in general. Roboticists often don't want to think about memory. But then our APIs also require them in places, so it's not that easy to reduce them even in your own code. At least it's not as bad as PCL yet :laugh:

But wouldn't `unique_ptr` be just as good at cleaning up when you are done with something.  Only downside is you would then have to reason about ownership **gasp**.  That would make it harder to write, but we already agreed that being **readable** is what we care about.  All these `shared_ptr` things do not make for code that is friendly to read and reason about.

tylerjw:
> The main reason for adding aliases in many places a few years back was to support both `boost::shared_ptr` and `std::shared_ptr` without relying on `auto`.

But doesn't the macro only support one or the other?  You still have to reason about which one you are using in different places (ROS still uses boost::shared_ptr for several things).  All this macro does is hide which one you are using.  It also has the downside that it encourages you to create them to manage memory created with the new operator instead of the technically much better `make_shared` syntax.  The only downside about `make_shared` is you have to know if you are creating the boost or stl one.

I don't find this easier to read in any way.  In many ways `auto` makes for more readable code than `Ptr`.  I think the problem that most people have with `auto` is that they want to avoid reasoning about type deduction and how it works.  It isn't that scary and type-deduction is a powerful tool every C++ author should familiarize themselves with.

One of the few places in ROS where smart pointers make a lot of sense to me is how it is used in the messaging system to allow for sending messages between parts of the same process without copying them.  This is really nice for large message types like images.  And, even in that case, `unique_ptr` would work just fine.  It seems that this is the way ros2 is going thankfully.  Zero-copy messaging seems to only be implemented for `unique_ptr` in ros2 last time I tried to use it.

tylerjw:
> would you have any ideas how to do that? Because the only way I can think of would be to not use the macro at all, and then manually expand it for all forward-declared types.
> 
> I hope you agree that would be a maintenance 'nightmare'?

@gavanderhoorn would that actually be that bad?  It might help us carefully think before putting `Ptr` in some external interface.

v4hn:
@tylerjw I did not expect to hit such a bad nerve. :-)

> But doesn't the macro only support one or the other?

Yes, but it abstracts it from the API of a method/library, so users can write code that doesn't care about which one is used. Converting them into each other was never a reasonable option.

> And, even in that case, unique_ptr would work just fine

Not if you have multiple subscribers.

> In many ways auto makes for more readable code than Ptr.

In my experience `auto` blurs the line between *writable* and *readable* a lot.
Just because I think it reads well now when I write less characters, doesn't mean that next year some user (that would be me again) would stare at it for way too long... In the past we agreed not to allow `auto` in MoveIt outside of contexts that mention the type in the same line.

tylerjw:
> Not if you have multiple subscribers.

A shared_ptr can be made from a unique_ptr easily without copying the data.  Perhaps the ROS middleware could do that conversion in this case.  The downside is the subscriber side would always have to present both a unique_ptr and shared_ptr interface or if they presented only the unique_ptr the middleware would have to copy the data.  That would also be gross.

PR #2320, opened 2020-09-21
Fix Clear Octomap button, disable when no octomap is published

felixvd:
### Description

As discussed in #2310, the Clear Octomap button shouldn't be active when there is no octomap to clear. This PR disables the button by default and enables it when an octomap is in the planning scene. 

You can test this with `roslaunch moveit_tutorials detect_and_add_cylinder_collision_object_demo.launch` which publishes an octomap every 10 seconds, ~although the octomap doesn't actually seem to be cleared when pushing the button~.

Screenshot 2020-09-22 01:44:45

**edit:** I think I fixed part of the problem that caused the octomap not to be cleared. There's still something quirky, though.

### Checklist

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -531,6 +531,8 @@ void PlanningSceneMonitor::newPlanningSceneCallback(const moveit_msgs::PlanningS
 
 void PlanningSceneMonitor::clearOctomap()
 {
+  if (scene_->getWorldNonConst()->removeObject(scene_->OCTOMAP_NS))
+    triggerSceneUpdateEvent(UPDATE_SCENE);
   if (octomap_monitor_)
   {
     octomap_monitor_->getOcTreePtr()->lockWrite();
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -531,8 +531,8 @@ void PlanningSceneMonitor::newPlanningSceneCallback(const moveit_msgs::PlanningS
 
 void PlanningSceneMonitor::clearOctomap()
 {
-  scene_->getWorldNonConst()->removeObject(scene_->OCTOMAP_NS);
-  triggerSceneUpdateEvent(UPDATE_SCENE);
+  if (scene_->getWorldNonConst()->removeObject(scene_->OCTOMAP_NS))
+    triggerSceneUpdateEvent(UPDATE_SCENE);
   if (octomap_monitor_)
   {
     octomap_monitor_->getOcTreePtr()->lockWrite();
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -531,8 +531,8 @@ void PlanningSceneMonitor::newPlanningSceneCallback(const moveit_msgs::PlanningS
 
 void PlanningSceneMonitor::clearOctomap()
 {
-  if (scene_->getWorldNonConst()->removeObject(scene_->OCTOMAP_NS))
-    triggerSceneUpdateEvent(UPDATE_SCENE);
+  scene_->getWorldNonConst()->removeObject(scene_->OCTOMAP_NS);
+  triggerSceneUpdateEvent(UPDATE_SCENE);
   if (octomap_monitor_)
   {
     octomap_monitor_->getOcTreePtr()->lockWrite();
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -431,12 +431,7 @@ void MotionPlanningFrame::changePlanningGroup()
 void MotionPlanningFrame::sceneUpdate(planning_scene_monitor::PlanningSceneMonitor::SceneUpdateType update_type)
 {
   if (update_type & planning_scene_monitor::PlanningSceneMonitor::UPDATE_GEOMETRY)
-  {
     planning_display_->addMainLoopJob(boost::bind(&MotionPlanningFrame::populateCollisionObjectsList, this));
-    // Enable Clear Octomap button only if octomap exists in the world
-    const planning_scene_monitor::LockedPlanningSceneRO& ps = planning_display_->getPlanningSceneRO();
-    ui_->clear_octomap_button->setEnabled(ps->getWorld()->hasObject(ps->OCTOMAP_NS));
-  }
 }
 
 void MotionPlanningFrame::addSceneObject()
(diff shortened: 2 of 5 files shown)


felixvd:
Without that second commit, the planning scene in move_group would be up to date after calling `clearOctomap`, but the one in the MotionPlanning plugin would still have the octomap. In that situation, planning requests that the Collision-Aware IK reported to be in collision with the Octomap would succeed.

I'm not convinced that this is completely error-free (I don't use octomaps myself), but this is surely no worse than before and shouldn't cause trouble.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -433,7 +433,12 @@ void MotionPlanningFrame::changePlanningGroup()
 void MotionPlanningFrame::sceneUpdate(planning_scene_monitor::PlanningSceneMonitor::SceneUpdateType update_type)
 {
   if (update_type & planning_scene_monitor::PlanningSceneMonitor::UPDATE_GEOMETRY)
```
AndyZe, line 435: Should this be `&&`?

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -433,7 +433,12 @@ void MotionPlanningFrame::changePlanningGroup()
...
+  {
     planning_display_->addMainLoopJob(boost::bind(&MotionPlanningFrame::populateCollisionObjectsList, this));
+    // Enable Clear Octomap button only if octomap exists in the world
+    const planning_scene_monitor::LockedPlanningSceneRO& ps = planning_display_->getPlanningSceneRO();
+    ui_->clear_octomap_button->setEnabled(ps->getWorld()->hasObject(ps->OCTOMAP_NS));
+  }
```
AndyZe, line 441: Is this called very frequently? It might not be worth the cost of updating the button, if so

felixvd: Only when the scene geometry changes. If this impacted performance I would think that `populateCollisionObjectsList` has a larger effect than updating the button.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
felixvd: No, I think that's meant to be a bitwise AND. It compares `update_type` with the enum value `...UPDATE_GEOMETRY`. Would be easier to read if it was `==` imo, but this works so I don't see a need to change it.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
AndyZe: (working on testing this now)

Sorry, what does this mean exactly:  `Only when the scene geometry changes`. Is that every time a new point cloud arrives (if octomap is enabled)? Or just when a new object is added to the scene?

felixvd: That's supposed to mean whenever an object is added/removed/changed in the PlanningScene's `world` (= observers are called), but since afaik the octomap is also an object in the world (`OCTOMAP_NS`), I guess it's yes to both.

AndyZe:
Tested it, it works nicely. I vote to keep the changes in planning_scene_monitor.cpp but delete the rest. That's because I value performance more than keeping the button display up-to-date. Don't love the idea of querying the octomap status at, let's say, 15 Hz.

Let's wait for a second opinion on that?

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -531,8 +531,8 @@ void PlanningSceneMonitor::newPlanningSceneCallback(const moveit_msgs::PlanningS
 
 void PlanningSceneMonitor::clearOctomap()
 {
-  if (scene_->getWorldNonConst()->removeObject(scene_->OCTOMAP_NS))
```
rhaschke: For me this guard makes sense. Why did you remove it?

rhaschke (commented):
I understand @AndyZe concerns. However, as @felixvd already said, `populateCollisionObjectsList()` is the costly function.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -431,7 +431,12 @@ void MotionPlanningFrame::changePlanningGroup()
 void MotionPlanningFrame::sceneUpdate(planning_scene_monitor::PlanningSceneMonitor::SceneUpdateType update_type)
 {
   if (update_type & planning_scene_monitor::PlanningSceneMonitor::UPDATE_GEOMETRY)
+  {
     planning_display_->addMainLoopJob(boost::bind(&MotionPlanningFrame::populateCollisionObjectsList, this));
+    // Enable Clear Octomap button only if octomap exists in the world
+    const planning_scene_monitor::LockedPlanningSceneRO& ps = planning_display_->getPlanningSceneRO();
+    ui_->clear_octomap_button->setEnabled(ps->getWorld()->hasObject(ps->OCTOMAP_NS));
```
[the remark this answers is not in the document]
felixvd: Moving it to populateCollisionObjectsList is a good idea. I'll do that when I can.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
felixvd: I had issues where the octomap would still be displayed in Rviz, but not be considered for collisions. I assume that the MotionPlanning plugin's planning scene was not updated somehow.

felixvd: I couldn't reproduce those issues, so I reverted this change.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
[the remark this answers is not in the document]
felixvd: Done

rhaschke (approved):

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -531,6 +531,8 @@ void PlanningSceneMonitor::newPlanningSceneCallback(const moveit_msgs::PlanningS
 
 void PlanningSceneMonitor::clearOctomap()
 {
+  if (scene_->getWorldNonConst()->removeObject(scene_->OCTOMAP_NS))
+    triggerSceneUpdateEvent(UPDATE_SCENE);
```
simonschmeisser, line 535: I'm somewhat late to the party, but why is that `UPDATE_SCENE` instead of `UPDATE_GEOMETRY`? (we have been using that in our workaround code that should have been contributed for some while ...)

simonschmeisser:
I'm also wondering if `PlanningSceneMonitor::clearOctomap()` could lock the `PlanningScene` itself or if it's required for me to do this?

PR #2323, opened 2020-09-24
fix some more python3 issues

### LINKED ISSUE #2322: IKfast plugin not working
### Description

Hi, I installed the new version of Moveit for Noetic, but can't make IKFast plugin work.
Running the package as it is I get this error:

```
  File "/home/usr/catkin_ws/src/moveit/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py", line 50, in <module>
    import yaml
ImportError: No module named yaml
```

I imagined it was due to the compatibility with python3 so I tried changing the shebang line of the scritps `create_ikfast_moveit_plugin.py `and `round_collada_numbers.py` to `#!/usr/bin/env python3` and now I get:

```
  File "/home/usr/catkin_ws/src/moveit/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py", line 399, in <module>
    main()
  File "/home/usr/catkin_ws/src/moveit/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py", line 390, in main
    create_ikfast_package(args)
  File "/home/usr/catkin_ws/src/moveit/moveit_kinematics/ikfast_kinematics_plugin/scripts/create_ikfast_moveit_plugin.py", line 186, in create_ikfast_package
    etree.ElementTree(root).write(f, xml_declaration=True, pretty_print=True, encoding="UTF-8")
  File "src/lxml/etree.pyx", line 2057, in lxml.etree._ElementTree.write
  File "src/lxml/serializer.pxi", line 758, in lxml.etree._tofilelike
  File "src/lxml/etree.pyx", line 318, in lxml.etree._ExceptionContext._raise_if_stored
  File "src/lxml/serializer.pxi", line 682, in lxml.etree._FilelikeWriter.write
TypeError: write() argument must be str, not bytes
```

Which is the correct way to overcome these errors?

### Your environment
* ROS Distro: Noetic
* OS Version: Ubuntu 20.04
* Source build, branch: noetic-devel

rhaschke:
Fixes #2322:
* `ikfast` was missing some python dependencies
* [python scripts should be installed via `catkin_install_python()`](http://wiki.ros.org/UsingPython3/SourceCodeChanges#Changing_shebangs)

### DIFF
--- a/moveit_ros/visualization/src/moveit_ros_visualization/moveitjoy_module.py
+++ b/moveit_ros/visualization/src/moveit_ros_visualization/moveitjoy_module.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # ********************************************************************
 # Software License Agreement (BSD License)
 #
--- a/moveit_commander/CMakeLists.txt
+++ b/moveit_commander/CMakeLists.txt
@@ -7,7 +7,8 @@ catkin_python_setup()
 
 catkin_package()
 
-install(PROGRAMS bin/${PROJECT_NAME}_cmdline.py
-        DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
+catkin_install_python(
+  PROGRAMS bin/${PROJECT_NAME}_cmdline.py
+  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
 
 add_subdirectory(test)
--- a/moveit_kinematics/ikfast_kinematics_plugin/CMakeLists.txt
+++ b/moveit_kinematics/ikfast_kinematics_plugin/CMakeLists.txt
@@ -7,6 +7,12 @@ set(MOVEIT_LIB_NAME moveit_ikfast_kinematics_plugin)
 install(
   PROGRAMS
     scripts/auto_create_ikfast_moveit_plugin.sh
+  DESTINATION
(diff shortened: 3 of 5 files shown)


moveit_ros/visualization/src/moveit_ros_visualization/moveitjoy_module.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
```
tylerjw, line 1: Does this impact the ability to run this with python3 or would you need to define python3 as your system python for that to work?

tylerjw (approved):
I tested this on my 20.04 install with python3 and it seems to work.

moveit_ros/visualization/src/moveit_ros_visualization/moveitjoy_module.py
v4hn: This is a module file without a main, so [it does not need the shebang](https://stackoverflow.com/questions/26650427/should-i-put-the-shebang-line-in-every-python-file).

I don't think this breaks anything w.r.t. python3, it's just general (debatable) cleanup.
@rhaschke It would have been nice to see a comment indicating that in the commit message.

PR #2325, opened 2020-09-26
Handle multiple link libraries for FCL

rhaschke:
This is in preparation to use fcl 0.6.1 in Noetic. See https://github.com/ros/rosdistro/issues/26527 for details.
Our cmake code to find absolute library paths of FCL libs was only handling a single lib (namely `libfcl`). However, future releases of `libfcl` declare several link libraries. To correctly link to those, we need to find absolute paths for all of them.

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -48,10 +48,13 @@ endif()
 
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(LIBFCL_PC REQUIRED fcl)
-# find *absolute* paths to LIBFCL_INCLUDE_DIRS and LIBFCL_LIBRARIES
-find_path(LIBFCL_INCLUDE_DIRS fcl/config.h HINTS ${LIBFCL_PC_INCLUDE_DIR} ${LIBFCL_PC_INCLUDE_DIRS})
-find_library(LIBFCL_LIBRARIES fcl HINTS ${LIBFCL_PC_LIBRARY_DIRS})
-
+set(LIBFCL_INCLUDE_DIRS ${LIBFCL_PC_INCLUDE_DIRS})
+# find *absolute* paths to LIBFCL_LIBRARIES
+set(LIBFCL_LIBRARIES)
+foreach(_lib ${LIBFCL_PC_LIBRARIES})
+  find_library(_lib_${_lib} ${_lib} HINTS ${LIBFCL_PC_LIBRARY_DIRS})
+  list(APPEND LIBFCL_LIBRARIES ${_lib_${_lib}})
+endforeach()
 
 find_package(octomap REQUIRED)
 find_package(urdfdom REQUIRED)


moveit_core/CMakeLists.txt
```diff
@@ -48,10 +48,13 @@ endif()
 
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(LIBFCL_PC REQUIRED fcl)
-# find *absolute* paths to LIBFCL_INCLUDE_DIRS and LIBFCL_LIBRARIES
-find_path(LIBFCL_INCLUDE_DIRS fcl/config.h HINTS ${LIBFCL_PC_INCLUDE_DIR} ${LIBFCL_PC_INCLUDE_DIRS})
-find_library(LIBFCL_LIBRARIES fcl HINTS ${LIBFCL_PC_LIBRARY_DIRS})
-
+set(LIBFCL_INCLUDE_DIRS ${LIBFCL_PC_INCLUDE_DIRS})
```
simonschmeisser, line 51: Newer versions of fcl (ie not the system version in noetic ... what a mess ...) also provide a cmake config. So maybe
```
find_package(fcl REQUIRED)
```
already does the trick?

rhaschke: Yes, I tried this before. It works here, locally. But exporting the required `fcl` libs to downstream packages didn't work:
catkin provides two mechanisms to do this: 
1. For catkin packages via `CATKIN_DEPENDS` and 
2. for 3rd-party packages via `DEPENDS` in `catkin_package()`. 

The latter - required here - requires to declare `LIBFCL_INCLUDE_DIRS` and `LIBFCL_LIBRARIES`. But, I failed to assemble `LIBFCL_LIBRARIES` because fcl's cmake config declares an [imported library target](https://cmake.org/cmake/help/latest/command/add_library.html#imported-libraries). Manually accessing [`IMPORTED_LOCATION`](https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED_LOCATION.html) and [`INTERFACE_LINK_LIBRARIES`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES.html) of that target was a nightmare and I gave up.

simonschmeisser: Hmm, who besides `collision_detection_fcl` needs that dependency? But looking at the documentation of `catkin_package` I see your problem

rhaschke: > Who besides `collision_detection_fcl` needs that dependency?

`moveit_compare_collision_checking_speed_fcl_bullet` was failing to link.

simonschmeisser: Then the clean and lean `find_package` could be moved to these two directly? Normally the fcl dependency should be abstracted away right?

rhaschke: Maybe. I'm not sure `moveit_compare_collision_checking_speed_fcl_bullet` explicitly uses fcl functionality. If so, it should `find_package(fcl)` itself, I agree.

rhaschke: Given the fact that we need to handle FCL 0.5 for Melodic, we are stuck to pkgconfig for now. 
Hence, I would like to get this basic fix merged as is.

v4hn (approved):
Thanks for working on this. Makes sense to me.

PR #2335, opened 2020-10-01
MSA: Fix disappearing robot on change of reference frame

rhaschke:
This fixes https://github.com/ros-planning/moveit_tutorials/issues/533.

- set visibility of robot-state-display to true after reset()
- correctly update reference frame also on deletion of virtual joint

This needs to be backported to Melodic as well!

### DIFF
--- a/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
@@ -187,6 +187,7 @@ void SetupAssistantWidget::virtualJointReferenceFrameChanged()
   {
     rviz_manager_->setFixedFrame(QString::fromStdString(config_data_->getRobotModel()->getModelFrame()));
     robot_state_display_->reset();
+    robot_state_display_->setVisible(true);
   }
 }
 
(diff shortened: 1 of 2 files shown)


moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
```diff
@@ -187,6 +187,7 @@ void SetupAssistantWidget::virtualJointReferenceFrameChanged()
   {
     rviz_manager_->setFixedFrame(QString::fromStdString(config_data_->getRobotModel()->getModelFrame()));
     robot_state_display_->reset();
+    robot_state_display_->setVisible(true);
```
gavanderhoorn, line 190: Was this just never there, or was it removed somehow in a previous change?

rhaschke: It was never there because the RobotStateDisplay was showing the robot by default. But this was changed, I think in #1625.

gavanderhoorn (approved):

tylerjw (approved):
Thank you for this.

PR #2342, opened 2020-10-02
Changelogs for Noetic 1.1.1

tylerjw:
### Description

Changelogs for our next Noetic release.


tylerjw:
One thing of note is that I did not add any lines to the moveit meta package.  Please let me know if you think any of the changes since the last release are large enough they should be added and I'll make that change (or you can if you are a maintainer, as I did this on my fork).

AndyZe (approved):
I spot-checked this, did not see any issues. Arguably, fixing the robot display in MSA is significant enough to include in the moveit metapackage changelog.

tylerjw:
> I spot-checked this, did not see any issues. Arguably, fixing the crashes in MSA are significant enough to include in the moveit metapackage changelog.

That fix was merged today.  I'm not sure we wrote this down anywhere but my impression was that we would let noetic-devel get 1week behind master to during the "release proeccess" window to make things like these changelogs easier as they wouldn't be a moving target.  My understanding is that the noetic-devel branch would be released on this Friday (the 9th) containing the changes up until last Friday (the 2nd).

rhaschke:
I discourage branching off master for a release. This makes it much harder to track in history when (after which release) something was added. So, please take the few minutes to update the changelogs to the latest master when releasing.

tylerjw:
> I discourage branching off master for a release. This makes it much harder to track in history when (after which release) something was added. So, please take the few minutes to update the changelogs to the latest master when releasing.

I think I'm still confused about the reason for the `noetic-devel` branch.  Is it currently still just to support docker images or are we also going to use it for release?  When I do the release, should I do it off of master?

If we are releasing off of master, I'll update this friday morning and hopefully, we are at a good point and can just cut releases of each repo then.

rhaschke:
master and noetic-devel should be in sync until the end of the year or so, when we really fix API.

tylerjw:
@rhaschke I updated these, I'd really appreciate if you skim through them.

henningkayser (approved):

tylerjw:
idk what is wrong with travis, but since I'll probably be updating these again as the noetic sync is today and we need the fcl change before we bloom noetic we can put off merging this just yet.  :disappointed:

PR #2348, opened 2020-10-05
MSA: fix pose definitions

### LINKED ISSUE #2344: [msa] missing finter joint
### Description

When configuring default poses, if I try to add one for the end effector the last joint doesn't have a slider.  See the images.  Something else I noticed is that if I try to edit the joints for the arm the last joint is a fixed joint and doesn't show up in the list (that's probably fine) but might be why we didn't notice this before.

missing_finger_joint
missing_finger_joint_2

rhaschke:
Fix #2344 and other issues:
- The joint values of the initial robot pose (shown in rviz window after loading) were not taking into account for saving poses.
- mimic joints were written to SRDF (with wrong values), overriding auto-updates of mimic joints (causing #2344)
- Cleaning up code to use a RobotState for robot poses (instead of a separate joint_states_map_)

### DIFF
--- a/moveit_setup_assistant/src/widgets/robot_poses_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/robot_poses_widget.cpp
@@ -806,36 +806,24 @@ void RobotPosesWidget::updateRobotModel(const std::string& name, double value)
 // ******************************************************************************************
 void RobotPosesWidget::publishJoints()
 {
-  // Change the scene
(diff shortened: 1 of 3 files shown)


v4hn (approved):
Great simplifications!

v4hn:
merged via merge commit to retain individual commits.

tylerjw:
Thank you for this.  I just tested it and it fixes the mimic joint issue, also it looks like the code is much simpler now.

PR #2349, opened 2020-10-05
MSA: Fix layout

### LINKED ISSUE #2339: [moveit_setup_assistant] larger default window size
Here is a minor UI tweak that would be nice.  The `Default Planning Groups` needs to be adjusted to have more vertical space.  This is easy enough to resize the window by dragging the bottom of the window, but if the default size of the window was slightly taller it would make this look nicer.

Another page that could help with a larger default size is the `Define Robot Poses` when you are editing a pose, at the current default size it has awkward scrollbars in the center column.  To make it nicer for that window a wider default would also help.

default_planning_groups_layout

rhaschke:
Hunting down #2339, I came across several improvements of the (manual :cry:) GUI creation.
However, I didn't yet find the underlying reason why Qt cannot correctly layout the widget - the minimal size is incorrectly computed. Finally, I just increase the minimum height of the main widget...
Fixes #2339 for now.

### DIFF
--- a/moveit_setup_assistant/src/setup_assistant_main.cpp
+++ b/moveit_setup_assistant/src/setup_assistant_main.cpp
@@ -95,8 +95,8 @@ int main(int argc, char** argv)
 
   // Load Qt Widget
   moveit_setup_assistant::SetupAssistantWidget saw(nullptr, vm);
-  saw.setMinimumWidth(980);
-  saw.setMinimumHeight(550);
+  saw.setMinimumWidth(1090);
+  saw.setMinimumHeight(600);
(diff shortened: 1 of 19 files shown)


tylerjw (approved):
Thank you for doing this.  My only feedback is that when I ran this the window for the joint sliders still has a horizontal bar, if the window was slightly wider that'd be gone (the vertical one is probably not worth making the window taller for).

edit_poses

rhaschke:
> If the window was slightly wider that (the horizontal slider in the robot poses widget) would be gone.

I need to increase the minimum width from 980 to 1090 pixels to fit this. Should I?

PR #2350, opened 2020-10-05
MSA: Fix group editing

### LINKED ISSUE #2292: MSA: cancel components dialog for new planning group
In the "Planning Groups" tab of the MoveIt Setup Assistant, users can add new planning groups to their configuration:

dialog

As the last step of this process, the user should click any of the four buttons on the bottom to assign an actual structure to the robot.
(Someone should really add the "Add Kin. Chain" button to "Recommended" as well. It's the most generic/user-friendly way to describe simple arms..)

Clicking on any of the buttons opens a new form to assign the structures with a "Save" and "Cancel" button on the bottom:

dialog

Pressing the "Cancel" button should revert the dialog to the previous dialog, but it actually cancels the whole process to add a new group, throwing away the information entered already.

rhaschke:
Fix #2292. Note, that this builds on top of #2349!

### DIFF
--- a/moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/planning_groups_widget.cpp
@@ -468,7 +468,7 @@ void PlanningGroupsWidget::editSelected()
     loadJointsScreen(plan_group.group_);
 
     // Switch to screen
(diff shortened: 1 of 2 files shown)


v4hn (approved):
works perfectly and makes the code much more readable.

Looks like you spent quite some time with the MSA code base recently...

PR #2352, opened 2020-10-06
MSA: Allow to show both, visual and collision geometry

### LINKED ISSUE #780: Setup assistant does not show collision geometries
The preview robot state in the moveit setup assistant shows the visual meshes attached to links.
This is all good and nice for many of the dialogs (setting up the planning groups, poses, etc),
but it is not a good idea when the user tries to setup the allowed collision matrix for their models.
In this dialog, at least optionally, we should visualize the collision geometry instead.

This is especially important if you have "invisible walls" in your description - fixed links that have a collision geometry, but don't have a visual geometry associated. Currently they cannot be highlighted.

rhaschke:
Fixes #780. Note, that this builds on top of #2349!

### DIFF
--- a/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
@@ -45,6 +45,7 @@
 #include <QFont>
 #include <QLabel>
 #include <QPushButton>
+#include <QCheckBox>
 #include <QCloseEvent>
 #include <QMessageBox>
 #include <QString>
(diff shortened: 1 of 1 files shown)


tylerjw (approved):
This is a real nice usability improvement.  My only feedback is that it would be nice if it showed joints red if in collision and green when not in the edit pose screen (instead of all red for the group).  But I'm approving as that is extra work and this is a major quality life of improvement as is.

rhaschke:
> My only feedback is that it would be nice if it showed joints red if in collision and green when **not?** in the edit pose screen.

Please file a corresponding issue ;-)
Not sure, though, rviz will colorize the collision mesh at all. (I never have seen it other than red.)

PR #2354, opened 2020-10-07
Change servo namespacing logic

jliukkonen:
The namespacing logic seemed odd when running multiple servo servers within their own namespaces. This PR attempts to fix/change the namespacing logic within Servo to make it more intuitive. I listed before/after comparisons below to better illustrate the changes made in this PR.

**Before**

- Service namespaces were a bit surprising for the services Servo advertises. Namespacing the node that made use of Servo into namespace `foo` resulted to service topics like this: `/foo/foo/<nodeName>/change_drift_dimensions` when I would have expected simple `/foo/<nodeName>/change_drift_dimensions`.
- User was forced to load parameters into separate parameter namespace **that matched the node name**. This was a bit confusing since the node namespace is where the parameters are loaded by default.
- The topic names that Servo subscribed to had, by default, in the example files the name of the Servo server node (for example `servo_server/delta_twist_cmds`). And to be more precise, the topic names had to match the node name. I think it it better to keep relative topic names in config files without the requirement to use node name as part of them. This way user doesn't have to change config file topic names if the node name is changed in the launch file.
- Joint states subscriber subscribed to topic relative to the node handle namespace. Normally this is OK but if you wanted to place all the topics into the node namespace by initializing Servo using `ros::NodeHandle nh("~")` the joint states subscriber would start listening to `/<nodeName>/joint_states` topic which, I would argue, is rarely what one wants. And if more complex namespacing was used, there was no easy way (that I know of) to drop the `<nodeName>` part out of the full topic name just by modifying the topic name in the config file.

**After**

- The service namespacing problem is solved by always advertising the services in the namespace of the node.
- I made parameter namespacing optional since it is nice to be able to sometimes stash the Servo specific parameter under a sub-namespace within the node namespace. So user can now leave the parameters into node namespace for example like this `/servo_server/<parameters>` or use sub-namespace `/servo_server/my_own_servo_param_namespace/<parameters>`
- The topic names in the Servo config file do not need to know about the node name that calls Servo. One can still use absolute names such as`/foo/bar/servo/delta_twist_cmds` just like before. Only the implicit requirement of specifying the calling node name as part of the topic names is removed, that is, `delta_twist_cmds` will now suffice for as value for `cartesian_command_in_topic` parameter.
- Joint states subscriber listens always to the node parent namespace + topic name specified in the configuration file. For nodes running in global namespace the topic will default to `/joint_states`, for more complex namespaces the subscribed topic might be, for example, `/foo/bar/baz/joint_states`. Note that user is still able to specify absolute topic names in config file just like before.

In addition to the changes described in before/after examples, the servo server and cpp example files as well as the launch and config files have been modified to work with the changes in this PR.

### DIFF
--- a/moveit_ros/moveit_servo/config/ur_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/ur_simulated_config.yaml
@@ -46,10 +46,10 @@ hard_stop_singularity_threshold: 30 # Stop when the condition number hits this
 joint_limit_margin: 0.1 # added as a buffer to joint limits [radians]. If moving quickly, make this larger.
 
 ## Topic names
-cartesian_command_in_topic: servo_server/delta_twist_cmds  # Topic for incoming Cartesian twist commands
-joint_command_in_topic: servo_server/delta_joint_cmds # Topic for incoming joint angle commands
-joint_topic:  joint_states
-status_topic: servo_server/status # Publish status to this topic
+cartesian_command_in_topic: delta_twist_cmds  # Topic for incoming Cartesian twist commands
+joint_command_in_topic: delta_joint_cmds # Topic for incoming joint angle commands
+joint_topic: joint_states
+status_topic: status # Publish status to this topic
 command_out_topic: joint_group_position_controller/command # Publish outgoing commands here
 
 ## Collision checking for the entire robot body
--- a/moveit_ros/moveit_servo/config/ur_simulated_config.yaml
+++ b/moveit_ros/moveit_servo/config/ur_simulated_config.yaml
@@ -50,7 +50,7 @@ cartesian_command_in_topic: delta_twist_cmds  # Topic for incoming Cartesian twi
 joint_command_in_topic: delta_joint_cmds # Topic for incoming joint angle commands
 joint_topic: joint_states
 status_topic: status # Publish status to this topic
-command_out_topic: joint_group_position_controller/command # Publish outgoing commands here
+command_out_topic: /joint_group_position_controller/command # Publish outgoing commands here
 
 ## Collision checking for the entire robot body
 check_collisions: true # Check collisions?
--- a/moveit_ros/moveit_servo/launch/cpp_interface_example.launch
+++ b/moveit_ros/moveit_servo/launch/cpp_interface_example.launch
@@ -2,8 +2,8 @@
   <!-- Launch an example that sends commands via C++ API. -->
 
   <node name="servo_server" pkg="moveit_servo" type="cpp_interface_example" output="screen" >
-    <param name="parameter_ns" type="string" value="servo_server" />
-    <rosparam command="load" file="$(find moveit_servo)/config/ur_simulated_config.yaml" />
+    <param name="parameter_ns" type="string" value="optional_parameter_namespace" />
+    <rosparam ns="optional_parameter_namespace" command="load" file="$(find moveit_servo)/config/ur_simulated_config.yaml" />
   </node>
 
 </launch>
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -52,14 +52,8 @@ Servo::Servo(ros::NodeHandle& nh, const planning_scene_monitor::PlanningSceneMon
   if (!readParameters())
     exit(EXIT_FAILURE);
 
-  // By default joint topic name is relative to the node parent namespace. E.g. if the joint topic specified in the
-  // configuration files is "joint_states" and the node is launched in namespace "bar", which exists within
-  // namespace "foo", then the subscribed topic would be "/foo/bar/joint_states".
-  //
-  // Fully custom joint name topics can be used by using absolute topic names in config files. For example,
-  // joint topic parameter can be explicitly set to "/baz/my_joint_state_topic".
-  // **Note!** Make sure your planning scene monitor subscribes to correct joint states topic when using absolute
-  // joint state topic name *or* custom relative joint state topic name (for example, "my_joint_states").
+  // By default, joint topic name is relative to the node parent namespace. Fully custom joint name topics can be
+  // set by using absolute topic names in config files. For example, "/foo/my_joint_state_topic".
   ros::NodeHandle nh_parent_ns = ros::NodeHandle("");
   joint_state_subscriber_ = std::make_shared<JointStateSubscriber>(nh_parent_ns, parameters_.joint_topic);
 
--- a/moveit_ros/moveit_servo/src/servo.cpp
+++ b/moveit_ros/moveit_servo/src/servo.cpp
@@ -76,67 +76,64 @@ bool Servo::readParameters()
 {
   std::size_t error = 0;
 
-  // Check for parameter namespace from launch file. All other parameters will be read from this namespace.
-  std::string yaml_namespace;
-  if (ros::param::get("~parameter_ns", yaml_namespace))
-  {
-    if (!parameter_ns_.empty())
-      ROS_WARN_STREAM_NAMED(LOGNAME,
-                            "A parameter namespace was specified in the launch file AND in the constructor argument.");
+  // Optional parameter sub-namespace specified in the launch file. All other parameters will be read from this namespace.
+  std::string parameter_ns;
+  ros::param::get("~parameter_ns", parameter_ns);
 
-    parameter_ns_ = yaml_namespace;
-  }
+  // If parameters have been loaded into sub-namespace within the node namespace, append the parameter namespace
+  // to load the parameters correctly.
+  ros::NodeHandle nh = parameter_ns.empty() ? nh_ : ros::NodeHandle(nh_, parameter_ns);
 
-  error += !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/publish_period", parameters_.publish_period);
-  error += !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/collision_check_rate", parameters_.collision_check_rate);
-  error += !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/num_outgoing_halt_msgs_to_publish",
+  error += !rosparam_shortcuts::get(LOGNAME, nh, "publish_period", parameters_.publish_period);
+  error += !rosparam_shortcuts::get(LOGNAME, nh, "collision_check_rate", parameters_.collision_check_rate);
+  error += !rosparam_shortcuts::get(LOGNAME, nh, "num_outgoing_halt_msgs_to_publish",
                                     parameters_.num_outgoing_halt_msgs_to_publish);
-  error += !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/scale/linear", parameters_.linear_scale);
-  error += !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/scale/rotational", parameters_.rotational_scale);
-  error += !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/scale/joint", parameters_.joint_scale);
+  error += !rosparam_shortcuts::get(LOGNAME, nh, "scale/linear", parameters_.linear_scale);
+  error += !rosparam_shortcuts::get(LOGNAME, nh, "scale/rotational", parameters_.rotational_scale);
+  error += !rosparam_shortcuts::get(LOGNAME, nh, "scale/joint", parameters_.joint_scale);
   error +=
-      !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/low_pass_filter_coeff", parameters_.low_pass_filter_coeff);
-  error += !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/joint_topic", parameters_.joint_topic);
-  error += !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/command_in_type", parameters_.command_in_type);
-  error += !rosparam_shortcuts::get("", nh_, parameter_ns_ + "/cartesian_command_in_topic",
+      !rosparam_shortcuts::get(LOGNAME, nh, "low_pass_filter_coeff", parameters_.low_pass_filter_coeff);
+  error += !rosparam_shortcuts::get(LOGNAME, nh, "joint_topic", parameters_.joint_topic);
+  error += !rosparam_shortcuts::get(LOGNAME, nh, "command_in_type", parameters_.command_in_type);
+  error += !rosparam_shortcuts::get(LOGNAME, nh, "cartesian_command_in_topic",
                                     parameters_.cartesian_command_in_topic);
   error +=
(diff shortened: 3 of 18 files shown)


moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -52,7 +52,16 @@ Servo::Servo(ros::NodeHandle& nh, const planning_scene_monitor::PlanningSceneMon
...
-  joint_state_subscriber_ = std::make_shared<JointStateSubscriber>(nh_, parameters_.joint_topic);
+  // By default joint topic name is relative to the node parent namespace. E.g. if the joint topic specified in the
+  // configuration files is "joint_states" and the node is launched in namespace "bar", which exists within
+  // namespace "foo", then the subscribed topic would be "/foo/bar/joint_states".
+  //
+  // Fully custom joint name topics can be used by using absolute topic names in config files. For example,
+  // joint topic parameter can be explicitly set to "/baz/my_joint_state_topic".
+  // **Note!** Make sure your planning scene monitor subscribes to correct joint states topic when using absolute
+  // joint state topic name *or* custom relative joint state topic name (for example, "my_joint_states").
+  ros::NodeHandle nh_parent_ns = ros::NodeHandle("");
```
henningkayser, line 63: Shouldn't this still be a private NodeHandle following the described logic?

moveit_ros/moveit_servo/config/ur_simulated_config.yaml
```diff
@@ -43,10 +43,10 @@ hard_stop_singularity_threshold: 30 # Stop when the condition number hits this
...
 ## Topic names
-cartesian_command_in_topic: servo_server/delta_twist_cmds  # Topic for incoming Cartesian twist commands
-joint_command_in_topic: servo_server/delta_joint_cmds # Topic for incoming joint angle commands
-joint_topic:  joint_states
-status_topic: servo_server/status # Publish status to this topic
+cartesian_command_in_topic: delta_twist_cmds  # Topic for incoming Cartesian twist commands
+joint_command_in_topic: delta_joint_cmds # Topic for incoming joint angle commands
+joint_topic: joint_states
+status_topic: status # Publish status to this topic
 command_out_topic: joint_group_position_controller/command # Publish outgoing commands here
```
henningkayser, line 50: Aren't the controller topics namespaced as well? A usual setup would have these at root level.

moveit_ros/moveit_servo/launch/cpp_interface_example.launch
```diff
@@ -2,8 +2,8 @@
   <!-- Launch an example that sends commands via C++ API. -->
 
   <node name="servo_server" pkg="moveit_servo" type="cpp_interface_example" output="screen" >
-    <param name="parameter_ns" type="string" value="servo_server" />
-    <rosparam command="load" file="$(find moveit_servo)/config/ur_simulated_config.yaml" />
+    <param name="parameter_ns" type="string" value="optional_parameter_namespace" />
```
henningkayser, line 5: Just curious, what exactly is the purpose of this optional subnamespace?

jliukkonen: Suppose you had a node named `foobar_node`. Normally parameters would be loaded into `/foobar_node/<parameters>`. If the node already has many parameters maybe you want to make things a bit clearer by having `foobar_node` parameters inside namespace `/foobar_node/<parameters>` and all the servo parameters (there are quite a few of them) within subnamespace `/foobar_node/servo_params/<parameters>`. You can also stash multiple servo configurations within single config file and place them into named sections and use that subnamespace to access them.

moveit_ros/moveit_servo/config/ur_simulated_config.yaml
jliukkonen: Good point, I need to check the namespacing for this one.

moveit_ros/moveit_servo/src/servo.cpp
jliukkonen: I may have used wrong terminology, so just to confirm that we are talking about the same thing: If the node would have name `servo_node` and it would exist within namespace `foo`. Then I would argue that the `joint_states` it listens to should be `/foo/joint_states` and not `/foo/servo_node/joint_states`. If the explanation is confusing, can you propose an alternative wording?

moveit_ros/moveit_servo/launch/cpp_interface_example.launch
AndyZe: I'm OK with this sub-namespace approach. Seems like a good idea

AndyZe (commented):
Thanks! Fixing the service namespacing is important. Please rebase.

Unfortunately, it looks like this tutorial does not work with the changes--

https://ros-planning.github.io/moveit_tutorials/doc/realtime_servo/realtime_servo_tutorial.html

I think the problem is, Servo is publishing on this topic:  `/servo_server/joint_group_position_controller/command`

But the controller expects this topic:  `/joint_group_position_controller/command`

So make sure topics in the yaml files have a leading `/`.

I'll need to test on our dual-arm projects after you rebase.


> The service namespacing problem is solved by always advertising the services in the namespace of the node.

What if we're using the C++ interface, and two Servo objects are instantiated in the node?

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -52,7 +52,16 @@ Servo::Servo(ros::NodeHandle& nh, const planning_scene_monitor::PlanningSceneMon
...
 
-  joint_state_subscriber_ = std::make_shared<JointStateSubscriber>(nh_, parameters_.joint_topic);
+  // By default joint topic name is relative to the node parent namespace. E.g. if the joint topic specified in the
+  // configuration files is "joint_states" and the node is launched in namespace "bar", which exists within
+  // namespace "foo", then the subscribed topic would be "/foo/bar/joint_states".
+  //
+  // Fully custom joint name topics can be used by using absolute topic names in config files. For example,
+  // joint topic parameter can be explicitly set to "/baz/my_joint_state_topic".
+  // **Note!** Make sure your planning scene monitor subscribes to correct joint states topic when using absolute
+  // joint state topic name *or* custom relative joint state topic name (for example, "my_joint_states").
```
AndyZe, line 62: IMO this comment is too long. I think you could condense it to:

  // By default, joint topic name is relative to the node parent namespace.  Fully custom joint name topics can be set by using absolute topic names in config files. For example, "/baz/my_joint_state_topic".

AndyZe:
Overall I'm kind of lukewarm about this PR. It would be really nice to fix service namespacing but I think this will lead to a lot of existing users who are missing `/` in their yaml files.

jliukkonen:
The pose tracking PR brought in more changes than I realized. I have to go over those changes and see how they fit together with this PR. The pose tracking interface seems to utilize separate parameter namespace variable to its constructor while I would argue for using node handle. Sorry I didn't bring this up earlier, I simply hadn't considered namespacing back then. But I'll make the required changes here and then we can continue discussion when there is something more concrete to talk about.

As for breaking changes, anyone who depends on `master` branch of any repo implicitly accepts that sometimes their code breaks and they have to accommodate the changes in the repos they rely on. Only breaking changes that should not be merged are the ones that can potentially be dangerous with, for instance, old configuration files.

And, if this PR grows much more, then maybe it is best to concentrate on just the services but I will at least try to get this right in one PR first.

jliukkonen:
Well, this is a bit of a wall of text but I'm hoping this helps in the reviewing process. I listed below examples of how to use Servo with this PR and how the different paramters, topics and services end up looking like. As for the tutorials, I would rather concentrate on getting the namespacing right and once it is right, then it should be trivial to make the tiny fixes to the tutorials and examples that may still need updating.

### How to run two Servo instances in a single node:

```
  ros::NodeHandle nh("~");
  
  // Left hand side
  ros::NodeHandle left_nh = ros::NodeHandle(nh, "left_side");
  moveit_servo::Servo left_servo(left_nh, planning_scene_monitor);
  left_servo.start();

  // Right hand side
  ros::NodeHandle right_nh = ros::NodeHandle(nh, "right_side");
  moveit_servo::Servo right_servo(right_nh, planning_scene_monitor);
  right_servo.start();
```

**Example launch file:**

```
<group ns="foo">
   	<node name="servo_server" pkg="moveit_servo" type="cpp_interface_example" output="screen" >
    	<param name="parameter_ns" type="string" value="sub_namespace" /> 
    	<rosparam command="load" file="$(find moveit_servo)/config/ur_simulated_config.yaml" ns="left_side/sub_namespace"/>
    	<rosparam command="load" file="$(find moveit_servo)/config/pose_tracking_settings.yaml" ns="left_side/sub_namespace" />
    	<rosparam command="load" file="$(find moveit_servo)/config/ur_simulated_config.yaml" ns="right_side/sub_namespace"/>
    	<rosparam command="load" file="$(find moveit_servo)/config/pose_tracking_settings.yaml" ns="right_side/sub_namespace" />
  	</node>
</group>
```

**Parameters:**

One example per each instance.
```
/foo/servo_server/left_side/sub_namespace/<param_name>
/foo/servo_server/right_side/sub_namespace/<param_name>
```

**Topics:**

Note! Especially pay attention to Servo internal topics which are fixed by this PR to truly be internal to that specific Servo instance. Also, The `joint_group_position_controller/command` is still within the node namespace to make sure the receiving end and differentiate between sources. If both instances published to /foo/servo_server/joint_group_position_controller/command` I'd say that will not work correctly.

```
/foo/attached_collision_object
/foo/collision_object
/foo/joint_states
/foo/planning_scene
/foo/planning_scene_world
/foo/servo_server/left_side/delta_joint_cmds
/foo/servo_server/left_side/delta_twist_cmds
/foo/servo_server/left_side/internal/collision_velocity_scale
/foo/servo_server/left_side/internal/worst_case_stop_time
/foo/servo_server/left_side/joint_group_position_controller/command
/foo/servo_server/left_side/status
/foo/servo_server/planning_scene_monitor/parameter_descriptions
/foo/servo_server/planning_scene_monitor/parameter_updates
/foo/servo_server/right_side/delta_joint_cmds
/foo/servo_server/right_side/delta_twist_cmds
/foo/servo_server/right_side/internal/collision_velocity_scale
/foo/servo_server/right_side/internal/worst_case_stop_time
/foo/servo_server/right_side/joint_group_position_controller/command
/foo/servo_server/right_side/status
```

**Services:**

```
/foo/servo_server/get_loggers
/foo/servo_server/left_side/change_control_dimensions
/foo/servo_server/left_side/change_drift_dimensions
/foo/servo_server/left_side/reset_servo_status
/foo/servo_server/planning_scene_monitor/set_parameters
/foo/servo_server/right_side/change_control_dimensions
/foo/servo_server/right_side/change_drift_dimensions
/foo/servo_server/right_side/reset_servo_status
/foo/servo_server/set_logger_level
```

### How to run single Servo instance in a single node:

```
ros::NodeHandle nh("~");

// Left hand side
moveit_servo::Servo servo(nh, planning_scene_monitor);
servo.start();
```

**Parameters:**

```
/foo/servo_server/sub_namespace/<parameter_name>
```

**Topics:**

```
/foo/attached_collision_object
/foo/collision_object
/foo/joint_states
/foo/planning_scene
/foo/planning_scene_world
/foo/servo_server/delta_joint_cmds
/foo/servo_server/delta_twist_cmds
/foo/servo_server/internal/collision_velocity_scale
/foo/servo_server/internal/worst_case_stop_time
/foo/servo_server/joint_group_position_controller/command
/foo/servo_server/planning_scene_monitor/parameter_descriptions
/foo/servo_server/planning_scene_monitor/parameter_updates
/foo/servo_server/status
```

**Services:**
```
/foo/servo_server/change_control_dimensions
/foo/servo_server/change_drift_dimensions
/foo/servo_server/get_loggers
/foo/servo_server/planning_scene_monitor/set_parameters
/foo/servo_server/reset_servo_status
/foo/servo_server/set_logger_level
```

### Multiple nodes

Running multiple nodes with multiple servos (or just one) in each of them is just a special case of the examples above.
Only thing that would change is the `servo_server` node name which would have to be unique for each node.

What about running the PoseTracker then? Well, it is exactly the same as the examples above except there is one more topic:
`/foo/servo_server/target_pose`

**More than one tracker?**
Want to have two trackers (or more) inside a single node? Same logic as above applies:

```
ros::NodeHandle nh("~");

// Left hand side
ros::NodeHandle left_nh = ros::NodeHandle(nh, "left_side");
moveit_servo::PoseTracking right_tracker(left_nh, planning_scene_monitor);

// Make a publisher for sending pose commands
ros::Publisher target_pose_pub =
  left_nh.advertise<geometry_msgs::PoseStamped>("target_pose", 1 /* queue */, true /* latch */);

// Subscribe to servo status (and log it when it changes)
StatusMonitor status_monitor(left_nh, "status");

// Right hand side
ros::NodeHandle right_nh = ros::NodeHandle(nh, "right_side");
moveit_servo::PoseTracking right_tracker(right_nh, planning_scene_monitor);

ros::Publisher target_pose_pub =
  right_nh.advertise<geometry_msgs::PoseStamped>("target_pose", 1 /* queue */, true /* latch */);

StatusMonitor status_monitor(right_nh, "status");
```

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: @jliukkonen Please address this

moveit_ros/moveit_servo/config/ur_simulated_config.yaml
```diff
@@ -46,11 +46,11 @@ hard_stop_singularity_threshold: 30 # Stop when the condition number hits this
...
 
 ## Topic names
-cartesian_command_in_topic: servo_server/delta_twist_cmds  # Topic for incoming Cartesian twist commands
-joint_command_in_topic: servo_server/delta_joint_cmds # Topic for incoming joint angle commands
-joint_topic:  joint_states
-status_topic: servo_server/status # Publish status to this topic
-command_out_topic: joint_group_position_controller/command # Publish outgoing commands here
+cartesian_command_in_topic: delta_twist_cmds  # Topic for incoming Cartesian twist commands
+joint_command_in_topic: delta_joint_cmds # Topic for incoming joint angle commands
+joint_topic: joint_states
```
AndyZe, line 51: Let's put a / here. The tutorial will require it:  `/joint_states`

AndyZe: Oh, this works as-is, actually. Please ignore.

moveit_ros/moveit_servo/src/servo.cpp
jliukkonen: done

AndyZe:
I'm waiting to test this on two more robot setups (one with dual arms). Looks good otherwise!

AndyZe (approved):
Actually I don't have a good dual-arm setup to test with any more, but the changes seem like a net benefit. +1

moveit_ros/moveit_servo/src/servo.cpp
henningkayser: I see, your solution totally makes sense

henningkayser (approved):

PR #2319, opened 2020-09-18
[Kinetic] Let the max number of contacts be the amount of world objects + link models with geometry

LoyVanBeek:
### Description

Fix for https://github.com/ros-planning/moveit/issues/1673, where 
> MoveIt 'check_state_validity' service does not return the contact points of the collision if there is no extra collision object added to the panning scene.

The maximum number of contacts was set to the total number of world objects, but in case there is an empty world (number of world objects == 0) and the `RobotState`-to-check has a self collision, the number of contacts would still be 0. 
This is incorrect, the max number of contacts should be number of world objects) + number of robot links

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/) :arrow_right: not applicable
- [x] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes :arrow_right: not applicable
- [x] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html)
- [x] Include a screenshot if changing a GUI :arrow_right: not applicable
- [x] While waiting for someone to review your request, please help review [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers :arrow_right: I don't feel qualified to do so, all PRs have better reviews than I can give

PR #2355, opened 2020-10-08
[Master] Let the max number of contacts be the amount of world objects + link models with geometry

LoyVanBeek:
Duplicate of PR https://github.com/ros-planning/moveit/pull/2319 but targeting the master branch.


v4hn (approved):
It's somewhat embarrassing that this has been open for so long... Thanks for putting such an amount of work into such a tiny fix! :+1: 

I just (partially force-)pushed some cleanup. If CI passes, I approve.

henningkayser (approved):

PR #2359, opened 2020-10-10
Depend on ros-noetic-fcl (0.6) in Noetic

rhaschke:
After [having fcl 0.6 released](https://github.com/ros/rosdistro/pull/26706) into the ROS repos (additionally to fcl 0.5 in the standard Ubuntu/Debian repos) we can switch to this newer version of fcl now.
Probably, Travis will need to be switched to ros-testing (because it's not synced yet).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
   global:
     - MOVEIT_CI_TRAVIS_TIMEOUT=85  # Travis grants us 90 min, but we add a safety margin of 5 min
     - ROS_DISTRO=noetic
-    - ROS_REPO=ros
+    - ROS_REPO=ros-testing
     - UPSTREAM_WORKSPACE=moveit.rosinstall
     - CXXFLAGS="-Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls"
     - WARNINGS_OK=false
--- a/moveit_core/package.xml
+++ b/moveit_core/package.xml
@@ -26,7 +26,8 @@
   <depend>bullet</depend>
   <depend>eigen_stl_containers</depend>
   <depend>eigen_conversions</depend>
-  <depend>libfcl-dev</depend>
+  <depend condition="$ROS_DISTRO != noetic">libfcl-dev</depend>
+  <depend condition="$ROS_DISTRO == noetic">fcl</depend>
   <depend version_gte="0.5.2">geometric_shapes</depend>
   <depend>geometry_msgs</depend>
   <depend>kdl_parser</depend>


v4hn:
Is this urgent for you? Otherwise I would just wait for the next sync first.

The [build instructions](https://moveit.ros.org/install/source/dependencies/) should probably be adapted as well.

tylerjw:
@v4hn I think this is needed for releasing on Noetic.  I'm going to try to test this today.

rhaschke:
Yes, @tylerjw is right: This is needed to resolve link conflicts on Buster at least. 
Regarding build instructions: We could just mention that 0.6.1 is part of the binary distro and thus automatically installed with MoveIt. If one needs even newer fcl (master), one just needs to clone the repo into the MoveIt workspace as well. On Noetic, with this PR here, dependencies should be resolved automatically, as fcl 0.6 provides a correct package.xml already.

v4hn:
Ok, as the MoveIt release is apparently still broken in noetic until this is patched this should be merged before the sync.

>  I'm going to try to test this today.

Thanks @tylerjw! I will leave the review & merge to you then.

>  Regarding build instructions: We could just mention that 0.6.1 is part of the binary distro and thus automatically installed with MoveIt. If one needs even newer fcl (master), one just needs to clone the repo into the MoveIt workspace as well. On Noetic, with this PR here, dependencies should be resolved automatically, as fcl 0.6 provides a correct package.xml already.

For *noetic* that's right, but people will still want to build master on melodic as well.

tylerjw:
I just tested this in 20.04 and found a problem.  People who build moveit from source are going to have a problem when they do this:

```bash
$ rosdep install --from-paths . --ignore-src -y --rosdistro=noetic
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
moveit_core: Cannot locate rosdep definition for [fcl]
```

What should we do to get around this?  I realize that is because it is in ros-testing, but I'm having trouble figuring out how to test that.  I changed my sources file to point to ros-testing but when I try to use rosdep install it still can't find it (It works after I manually install `ros-noetic-fcl`).  For people using moveit from source in noetic should we provide instructions on how to switch to ros-testing?  Am I missing something in my setup?

tylerjw:
After changing to the testing repo and manually installing ros-noetic-fcl it apears to work on 20.04.  I'm going to test doing the same on buster but I'm assuming it'll just work.

The largest issue I see is we should provide instructions for anyone building master on noetic for how to get the ros-noetic-fcl package until it is in the main repos.

tylerjw (commented):
I added the dependency to the source instructions [in this PR](https://github.com/ros-planning/moveit.ros.org/pull/517/files).  I'm not sure how to make travis use ros-testing.  @rhaschke could you make that change as part of this PR?

I also tested this in Buster and it works the same as in 20.04.

rhaschke:
The issue of `rosdep` not knowing about the `fcl` key will automatically resolve with the next sync. 
Hence (if at all) a comment in the build instructions to install `ros-noetic-fcl` manually from `ros-testing` is required.

tylerjw:
> Hence (if at all) a comment in the build instructions to install `ros-noetic-fcl` manually from `ros-testing` is required.

Isn't that exactly what my PR to the website is?

.travis.yml
```diff
@@ -14,7 +14,7 @@ env:
   global:
     - MOVEIT_CI_TRAVIS_TIMEOUT=85  # Travis grants us 90 min, but we add a safety margin of 5 min
     - ROS_DISTRO=noetic
-    - ROS_REPO=ros
+    - ROS_REPO=ros-testing
```
tylerjw, line 17: Oh, I didn't realize that was what this variable was for.  That is nice, thank you.

tylerjw (approved):

rhaschke:
> Isn't that exactly what my PR to the website is?

Yes, sorry. I only coarse skimmed over the past conversation and missed the PR link.

> I'm not sure how to make travis use ros-testing. @rhaschke could you make that change as part of this PR?

Done.

tylerjw:
Yea, I think in the conversation before there was talk of adding it to the dependencies build instructions.  I looked over those and thought it made more sense in the source build section to meantion it because that's where someone might encounter an issue.

PR #2360, opened 2020-10-10
[backport] Let the max number of contacts be the amount of world objects + link models with geometry

v4hn:
Backport of #2355 to kinetic as requested by @LoyVanBeek .

PR #2361, opened 2020-10-10
[backport] Let the max number of contacts be the amount of world objects + link models with geometry

v4hn:
Backport of #2355 for melodic.


v4hn:
Merging together with #2360 .

PR #2372, opened 2020-10-14
Cleanup current state handling in servo

tylerjw:
### Description

This changes servo to use the CurrentStateMonitor to get the joint variable positions instead of manually handling the messages.  This fixes a bug where if you sent updates to joints outside of the joint group we cared about the servo_calcs timer would block for 100ms while it waited for the joints it cared about (only if the count of that group was different).  This also reduces the subscribers to the joint positions and removes a bunch of unneeded code in servo. 

### Checklist


moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -53,13 +53,9 @@ Servo::Servo(ros::NodeHandle& nh, const planning_scene_monitor::PlanningSceneMon
   if (!readParameters())
     exit(EXIT_FAILURE);
 
-  joint_state_subscriber_ = std::make_shared<JointStateSubscriber>(nh_, parameters_.joint_topic);
+  servo_calcs_ = std::make_unique<ServoCalcs>(nh_, parameters_, planning_scene_monitor_, parameter_ns_);
```
henningkayser, line 56: At this point we should probably check if the state monitor is actually running (check `PSM::getCurrentStateMonitor()`), otherwise call `PSM::startStateMonitor()`. Also, we are using velocity values, so we should call `CSM::enableCopyDynamics()`

henningkayser (commented):
I think this simplification really makes a lot of sense.Thanks for preparing this!

moveit_ros/moveit_servo/src/servo.cpp
tylerjw: Good catch, I also needed to wait for the robot state to be complete (i was reading from uninitialized data in one of the tests (casing a bunch of nan's to show up in the output and that was causing the test to fail).  It should all be fixed now.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -198,12 +185,11 @@ void ServoCalcs::start()
 
   // TODO(adamp or andyz): after rebasing on upstream master, update this section. Maybe put the tf updates in a tiny
```
AdamPettinger, line 186: Found this TODO comment while looking through the changes.

I added it, and am pretty sure it's ok to remove. At the time, I noticed the calculating the transforms below happen in 2 separate places but are identical lines and thought we might want to make it a function. I think its ok to leave as it or do the function, but we should probably remove the TODO comment

AdamPettinger (commented):
I like this change. The joint state subscriber always felt hacky to me and I am not sorry to see it go

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: done, thank you for the review :)

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -326,7 +346,15 @@ const ServoParameters& Servo::getParameters() const
...
-  return joint_state_subscriber_->getLatest();
+  auto msg = moveit::util::make_shared_from_pool<sensor_msgs::JointState>();
+  auto current_state_time = planning_scene_monitor_->getStateMonitor()->getCurrentStateAndTime();
+
+  msg->header.stamp = current_state_time.second;
+  msg->name = current_state_time.first->getJointModelGroup(parameters_.move_group_name)->getActiveJointModelNames();
+  current_state_time.first->copyJointGroupPositions(parameters_.move_group_name, msg->position);
+  current_state_time.first->copyJointGroupPositions(parameters_.move_group_name, msg->velocity);
+
+  return msg;
```
tylerjw, line 357: This is now a convenience function that retains the original functionality but should probably not be used.  The reason for that is it does a bunch of unnecessary work if all you wanted was the latest joint values from the current robot state.  In that case, whatever is calling this should instead just use the CSM through the PSM it already has.

I did this because I didn't want to brake any code that depends on this API.  I think if this change is adopted into MoveIt2 this API function should just be removed.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -665,10 +645,10 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
 
   // Calculate a small change in joints
   Eigen::VectorXd new_theta;
-  kinematic_state_->copyJointGroupPositions(joint_model_group_, new_theta);
+  current_state_->copyJointGroupPositions(joint_model_group_, new_theta);
   new_theta += pseudo_inverse * delta_x;
-  kinematic_state_->setJointGroupPositions(joint_model_group_, new_theta);
-  auto new_jacobian = kinematic_state_->getJacobian(joint_model_group_);
+  current_state_->setJointGroupPositions(joint_model_group_, new_theta);
+  auto new_jacobian = current_state_->getJacobian(joint_model_group_);
```
jliukkonen, line 651: Although this `auto` is not a new change, I think it would be good idea to make it explicit Eigen type just to avoid any "Eigen + auto type" pitfalls.

jliukkonen (approved):
Looks good to me. However, note that one comment about the Eigen + auto. It might be that that use case is safe but having wasted so much time with Eigen problems in the past, I've become paranoid about them. :)

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -38,6 +38,9 @@
 
 #pragma once
 
+// C++
+#include <mutex>
```
AndyZe, line 42: I don't think you needed to add this #include because it compiled previously, and you didn't add any new mutexes

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -326,7 +346,15 @@ const ServoParameters& Servo::getParameters() const
...
 sensor_msgs::JointStateConstPtr Servo::getLatestJointState() const
 {
-  return joint_state_subscriber_->getLatest();
+  auto msg = moveit::util::make_shared_from_pool<sensor_msgs::JointState>();
+  auto current_state_time = planning_scene_monitor_->getStateMonitor()->getCurrentStateAndTime();
+
+  msg->header.stamp = current_state_time.second;
+  msg->name = current_state_time.first->getJointModelGroup(parameters_.move_group_name)->getActiveJointModelNames();
+  current_state_time.first->copyJointGroupPositions(parameters_.move_group_name, msg->position);
+  current_state_time.first->copyJointGroupPositions(parameters_.move_group_name, msg->velocity);
```
AndyZe, line 355: Should be copyJointGroupVelocities, right?

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: Since this function isn't used anywhere, I think you should delete it.

AndyZe:
Nice cleanup! I gave it a test on one of our projects and the tutorial -- seems fine all the way around.

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: Cheers to unit tests catching something

moveit_ros/moveit_servo/src/servo.cpp
tylerjw: good catch

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: done

tylerjw:
@AndyZe I fixed the velocities thing.  I'll go ahead and remove that method but add notes to this issue on how one would get the joint angles from the planning scene monitor.

moveit_ros/moveit_servo/src/servo.cpp
tylerjw: I removed the method and added a `MIGRATION.md` to moveit_servo with a note about how to get the latest joint positions using the PSM.

AndyZe (approved):

henningkayser (approved):

tylerjw:
I just pushed a rebase of this PR as another was merged and there were conflicts.

PR #2375, opened 2020-10-17
Fix servo trajectory point timestamping

jliukkonen:
This PR removes trajectory point timestamping which causes warnings about trajectory points being dropped due to having old timestamps:
```
ros.joint_trajectory_controller: Dropping all 1 trajectory point(s), as they occur before the current time.
Last point is 0.000064s in the past.
```
By removing explicit timestamping, the stamps will default to 0, meaning the trajectory execution defaults to _start now_. For more details, see section 2.3: [http://wiki.ros.org/joint_trajectory_controller](http://wiki.ros.org/joint_trajectory_controller) 

Small time from start is added to the trajectory point generated in `suddenHalt` to prevent warnings about trajectory point being 0.0000 seconds in the past (not sure if this should be considered as a bug or not, but it is clearly a corner case).

Finally, there is a tiny optimization added to the way twist command is passed to publisher from `calculateTwistCommand` function. It is not strictly related but felt too small for its own PR.

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -169,6 +169,9 @@ void ServoCalcs::start()
   // We will update last_sent_command_ every time we start servo
   auto initial_joint_trajectory = moveit::util::make_shared_from_pool<trajectory_msgs::JointTrajectory>();
 
+  // When a joint_trajectory_controller receives a new command, a stamp of 0 indicates "begin immediately"
+  // See http://wiki.ros.org/joint_trajectory_controller#Trajectory_replacement
+  initial_joint_trajectory->header.stamp = ros::Time(0);
   initial_joint_trajectory->header.frame_id = parameters_.planning_frame;
   initial_joint_trajectory->joint_names = internal_joint_state_.name;
   trajectory_msgs::JointTrajectoryPoint point;
@@ -360,6 +363,9 @@ void ServoCalcs::run(const ros::TimerEvent& timer_event)
     // (trajectory_msgs/JointTrajectory or std_msgs/Float64MultiArray).
     if (parameters_.command_out_type == "trajectory_msgs/JointTrajectory")
     {
+      // When a joint_trajectory_controller receives a new command, a stamp of 0 indicates "begin immediately"
+      // See http://wiki.ros.org/joint_trajectory_controller#Trajectory_replacement
+      joint_trajectory->header.stamp = ros::Time(0);
       outgoing_cmd_pub_.publish(joint_trajectory);
     }
     else if (parameters_.command_out_type == "std_msgs/Float64MultiArray")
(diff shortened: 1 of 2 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -170,7 +170,6 @@ void ServoCalcs::start()
   auto initial_joint_trajectory = moveit::util::make_shared_from_pool<trajectory_msgs::JointTrajectory>();
 
   initial_joint_trajectory->header.frame_id = parameters_.planning_frame;
-  initial_joint_trajectory->header.stamp = ros::Time::now();
```
AndyZe, line 173: A comment here about why header.stamp isn't filled in would be great. I assume it's because it defaults to 0

AndyZe: Let's set it to 0 explicitly, too. Why not, right?  Something like:

// When a joint_trajectory_controller receives a new command, a stamp of 0 indicates "begin immediately"
// See http://wiki.ros.org/joint_trajectory_controller#Trajectory_replacement
initial_joint_trajectory->header.stamp = 0;

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -595,7 +593,6 @@ void ServoCalcs::composeJointTrajMessage(const sensor_msgs::JointState& joint_st
                                          trajectory_msgs::JointTrajectory& joint_trajectory) const
 {
   joint_trajectory.header.frame_id = parameters_.planning_frame;
```
AndyZe, line 595: I think the same comment as well as explicitly setting the stamp to 0 would be good here:

// When a joint_trajectory_controller receives a new command, a stamp of 0 indicates "begin immediately"
// See http://wiki.ros.org/joint_trajectory_controller#Trajectory_replacement
initial_joint_trajectory->header.stamp = 0;

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -810,6 +807,8 @@ void ServoCalcs::suddenHalt(trajectory_msgs::JointTrajectory& joint_trajectory)
   joint_trajectory.points.clear();
   joint_trajectory.points.emplace_back();
   trajectory_msgs::JointTrajectoryPoint& point = joint_trajectory.points.front();
+  point.time_from_start.fromNSec(1);
```
AndyZe, line 810: This seems a little dangerous. Wouldn't we want to set it to 0, to ensure the joint_trajectory_controller begins stopping as soon as it receives the message?  (per the same wiki section you mentioned, http://wiki.ros.org/joint_trajectory_controller#Trajectory_replacement)

AndyZe (commented):
Will wait to +1 until we get some related issues sorted out on a PickNik project

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: Oh, never mind. That's a different message field (header.stamp). I think this is fine.

AndyZe: Please add a comment about why you did this.

moveit_ros/moveit_servo/src/servo_calcs.cpp
jliukkonen: Sounds reasonable. I'll change it.

moveit_ros/moveit_servo/src/servo_calcs.cpp
jliukkonen: If `time_from_start` is zero and the trajectory start time is also zero, then it's all zeros and that leads to warnings about points being 0 seconds in the past (and like I said in the PR description, it's a bit weird). This is a way to work around that problem.

AndyZe (approved):

henningkayser (approved):
This is actually a weird issue but your fix is a reasonable workaround.

PR #2378, opened 2020-10-20
Order PlaceLocations by quality during planning

mvieth:
### Description

The quality field was added by https://github.com/ros-planning/moveit_msgs/commit/cdb9e4aa964269f1a159a8f35c7b634365e86038.
The changes are adapted from pick.cpp, where the Grasps are also ordered by quality.

### Checklist

### DIFF
--- a/moveit_ros/manipulation/pick_place/src/place.cpp
+++ b/moveit_ros/manipulation/pick_place/src/place.cpp
@@ -51,6 +51,20 @@ PlacePlan::PlacePlan(const PickPlaceConstPtr& pick_place) : PickPlacePlanBase(pi
 
 namespace
 {
+struct OrderPlaceLocationQuality
+{
+  OrderPlaceLocationQuality(const std::vector<moveit_msgs::PlaceLocation>& places) : places_(places)
+  {
+  }
+
+  bool operator()(const std::size_t a, const std::size_t b) const
+  {
+    return places_[a].quality > places_[b].quality;
+  }
+
+  const std::vector<moveit_msgs::PlaceLocation>& places_;
+};
+
 bool transformToEndEffectorGoal(const geometry_msgs::PoseStamped& goal_pose,
                                 const moveit::core::AttachedBody* attached_body, geometry_msgs::PoseStamped& place_pose)
 {
@@ -304,11 +318,18 @@ bool PlacePlan::plan(const planning_scene::PlanningSceneConstPtr& planning_scene
 
   pipeline_.start();
 
+  // order the place locations by quality
+  std::vector<std::size_t> places_order(goal.place_locations.size());
+  for (std::size_t i = 0; i < goal.place_locations.size(); ++i)
+    places_order[i] = i;
+  OrderPlaceLocationQuality oq(goal.place_locations);
+  std::sort(places_order.begin(), places_order.end(), oq);
+
   // add possible place locations
   for (std::size_t i = 0; i < goal.place_locations.size(); ++i)
   {
     ManipulationPlanPtr p(new ManipulationPlan(const_plan_data));
-    const moveit_msgs::PlaceLocation& pl = goal.place_locations[i];
+    const moveit_msgs::PlaceLocation& pl = goal.place_locations[places_order[i]];
 
     if (goal.place_eef)
       p->goal_pose_ = pl.place_pose;
@@ -325,7 +346,7 @@ bool PlacePlan::plan(const planning_scene::PlanningSceneConstPtr& planning_scene
     p->approach_ = pl.pre_place_approach;
     p->retreat_ = pl.post_place_retreat;
     p->retreat_posture_ = pl.post_place_posture;
-    p->id_ = i;
+    p->id_ = places_order[i];
     if (p->retreat_posture_.joint_names.empty())
       p->retreat_posture_ = attached_body->getDetachPosture();
     pipeline_.push(p);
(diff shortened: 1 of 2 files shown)


moveit_ros/manipulation/pick_place/src/place.cpp
```diff
@@ -304,11 +318,18 @@ bool PlacePlan::plan(const planning_scene::PlanningSceneConstPtr& planning_scene
 
   pipeline_.start();
 
+  // order the place locations by quality
+  std::vector<std::size_t> places_order(goal.place_locations.size());
```
felixvd, line 322:
```suggestion
  std::vector<std::size_t> place_locations_order(goal.place_locations.size());
```

felixvd (commented):
I haven't tested this, but if it works then it is a fine change.

Only suggested to use a longer variable name.

rhaschke (commented):
+1 after changing the variable name as suggested by @felixvd.

felixvd (approved):

rhaschke:
Just a final question before merging: If all qualities are equal (e.g. to zero), the order isn't reversed, right?

mvieth:
> Just a final question before merging: If all qualities are equal (e.g. to zero), the order isn't reversed, right?

The [documentation of std::sort](https://en.cppreference.com/w/cpp/algorithm/sort) says that "The order of equal elements is not guaranteed to be preserved." If you would like that guarantee, `std::stable_sort` could be an option.

moveit_ros/manipulation/pick_place/src/place.cpp
```diff
@@ -51,6 +51,20 @@ PlacePlan::PlacePlan(const PickPlaceConstPtr& pick_place) : PickPlacePlanBase(pi
...
 {
+struct OrderPlaceLocationQuality
+{
+  OrderPlaceLocationQuality(const std::vector<moveit_msgs::PlaceLocation>& places) : places_(places)
+  {
+  }
+
+  bool operator()(const std::size_t a, const std::size_t b) const
+  {
+    return places_[a].quality > places_[b].quality;
```
rhaschke, line 62: To preserve order of equal-quality elements:
```suggestion
    if (places_[a].quality > places_[b].quality) return true;
    if (places_[a].quality == places_[b].quality) return a > b;
    return false;
```

moveit_ros/manipulation/pick_place/src/place.cpp
```diff
@@ -59,7 +59,9 @@ struct OrderPlaceLocationQuality
 
   bool operator()(const std::size_t a, const std::size_t b) const
   {
-    return places_[a].quality > places_[b].quality;
+    if (places_[a].quality > places_[b].quality) return true;
+    if (places_[a].quality == places_[b].quality) return a > b;
```
rhaschke, line 63: Sorry, `sort` expects a less operator. The name `OrderPlaceLocationQuality` should reflect this...
What about the name `PlaceLocationLess`?
```suggestion
    if (places_[a].quality == places_[b].quality) return a < b;
```
You should definitely verify that this is the right combination. I just suggested this from the back of my head.

mvieth: What about `std::stable_sort` with the original compare operator?

rhaschke: `stable_sort` [erzeugt nochmal einen temporären buffer](https://github.com/gcc-mirror/gcc/blob/d9375e490072d1aae73a93949aa158fcd2a27018/libstdc%2B%2B-v3/include/bits/stl_algo.h#L4977)... Daher ist die vorgeschlagene Lösung auf jeden Fall effizienter.

mvieth: German :smile: the temporary buffer should be negligible since the sorting algorithm sorts the vector of `std::size_t`, not the `PlaceLocations` itself. Plus using `stable_sort` should be easier to understand than a rather complicated compare logic.

felixvd: The cost of that buffer is surely negligible considering the rest of this pipeline. In favor of `stable_sort` (with a comment if necessary).

Also a quick reminder that this is slated to be replaced by MTC, so this can be revisited to your heart's content :0

rhaschke: Ok, if there are two votes pro, go for it.

rhaschke (approved):

moveit_ros/manipulation/pick_place/src/place.cpp
mvieth: Should I change `sort` to `stable_sort` in pick.cpp as well?

rhaschke: Makes sense. I wouldn't expect the order to change...

mvieth: Done

tylerjw (approved):
I read throught the code and it seems like it should work as expected.  I left one small comment asking for a change to the test so it will test the sort operator but won't block merging it as is.

moveit_ros/manipulation/pick_place/src/place.cpp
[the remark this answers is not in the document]
mvieth: Hm, this isn't tested for pick, either. I am not sure how this internal sorting could be tested? Currently it is only tested whether the single Grasp/PlaceLocation returns a success. I think there is no way to find out which of several supplied Grasps/PlaceLocations was a success?

AndyZe:
@mvieth I just wanted to reiterate, the pick/place pipeline is probably going to be replaced by MoveIt Task Constructor in the future (especially in ROS2). Might not be worth spending a ton of time on this.

v4hn:
This just got stuck but was approved before. Sorry for the long wait...
Merging.

PR #2392, opened 2020-10-23
Fix QObject::connect: Cannot queue arguments of type 'QVector<int>'

rhaschke:
Fix annoying Qt warning about unregistered `QVector<int>`. The reason was that the planning scene name was changed from a non-GUI thread, causing a `dataChanged()` signal emitted from this thread, which then needs to be queued for processing in the main event loop.
I fixed that by emitting a signal from the thread to change the name. Thus, a `QString` event is queued, which doesn't pose an issue.

### DIFF
--- a/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
+++ b/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
@@ -112,9 +112,6 @@ public:
   planning_scene_monitor::LockedPlanningSceneRW getPlanningSceneRW();
   const planning_scene_monitor::PlanningSceneMonitorPtr& getPlanningSceneMonitor();
 
-Q_SIGNALS:
-  void changeSceneName(const QString& name);
-
 private Q_SLOTS:
 
   // ******************************************************************************************
--- a/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
+++ b/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
@@ -112,6 +112,9 @@ public:
   planning_scene_monitor::LockedPlanningSceneRW getPlanningSceneRW();
   const planning_scene_monitor::PlanningSceneMonitorPtr& getPlanningSceneMonitor();
 
+Q_SIGNALS:
+  void changeSceneName(const QString& name);
+
 private Q_SLOTS:
 
   // ******************************************************************************************
@@ -130,6 +133,7 @@ private Q_SLOTS:
   void changedSceneDisplayTime();
   void changedOctreeRenderMode();
   void changedOctreeColorMode();
+  void setSceneName(const QString& name);
 
 protected Q_SLOTS:
   virtual void changedAttachedBodyColor();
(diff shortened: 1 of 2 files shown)


tylerjw (approved):
I tested this and it works for me to remove the warning.  👍

moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h
```diff
@@ -112,6 +112,9 @@ class PlanningSceneDisplay : public rviz::Display
   planning_scene_monitor::LockedPlanningSceneRW getPlanningSceneRW();
   const planning_scene_monitor::PlanningSceneMonitorPtr& getPlanningSceneMonitor();
 
+Q_SIGNALS:
+  void changeSceneName(const QString& name);
```
simonschmeisser, line 116:
```suggestion
  void sceneNameChanged(const QString& name);
```
would be following the Qt naming scheme

rhaschke: MoveIt uses the naming scheme `changedXXX`:
https://github.com/ros-planning/moveit/blob/38852558d749a029826e827cf9165a5b458b6e87/moveit_ros/visualization/planning_scene_rviz_plugin/include/moveit/planning_scene_rviz_plugin/planning_scene_display.h#L123-L125

However, in contrast to the _passive notification_ slot `changedSceneName()`, the new slot `changeSceneName(QString)` will `actively` change the name.

simonschmeisser: Ok good. But still this is a `signal` where I suggested the change so it should be _passive_ ?

simonschmeisser: Also I cannot find a `connect` in that diff that actually uses the `signal` `changeSceneName`?

rhaschke: > But still this (`changeSceneName`) is a `signal` where I suggested the change so it should be _passive_ ?

Maybe, it should be named `triggerSceneNameChange` ...

> Also I cannot find a `connect` in that diff that actually uses the `signal` `changeSceneName`?

Good catch! I missed adding this in a hurry. Fixed now.

simonschmeisser: You can also call slots directly with `invokeMethod` if your signal is not actually a signal ...

rhaschke:
@simonschmeisser, I addressed your feedback. Thanks for pointing me to `invokeMethod`.

moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
```diff
@@ -583,14 +583,17 @@ void PlanningSceneDisplay::sceneMonitorReceivedUpdate(
 void PlanningSceneDisplay::onSceneMonitorReceivedUpdate(
     planning_scene_monitor::PlanningSceneMonitor::SceneUpdateType /*update_type*/)
 {
-  bool old_state = scene_name_property_->blockSignals(true);
   getPlanningSceneRW()->getCurrentStateNonConst().update();
-  scene_name_property_->setStdString(getPlanningSceneRO()->getName());
-  scene_name_property_->blockSignals(old_state);
-
+  QMetaObject::invokeMethod(this, "setSceneName", Qt::QueuedConnection,
```
simonschmeisser, line 587:
```suggestion
  QMetaObject::invokeMethod(this, &PlanningSceneDisplay::setSceneName, Qt::QueuedConnection,
```
probably doesn't work due to the argument right?

simonschmeisser (approved):

moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
rhaschke: Yeah, tried this first. But invokeMethod doesn't support methods, only lambdas.

PR #2395, opened 2020-10-27
Fix pose tracking race condition

jliukkonen:
This PR fixes issues where `target_pose` timestamp rewriting caused surprising and unwanted behaviour. This PR also fixes possible data races on the variable where it's value can be written in the subscriber callback function while it is simultaneously being read in another function.

- Timestamp rewriting was removed from the subscriber callback function to prevent outdated target poses being used in `moveToPose` function.
- Another instance of timestamp rewriting was removed from the `moveToPose` function itself to prevent situations where sending just one target pose message would occasionally result in situation where target pose would be sent with a reasonable timestamp, callback function would then process and store it, and finally moveToPose would be called which would rewrite the timestamp and then wait until failure for a target pose with a correct timestamp. 

In addition, access to class member variable `target_pose_` was protected with a mutex to prevent data races on it.

### DIFF
--- a/moveit_ros/moveit_servo/src/pose_tracking.cpp
+++ b/moveit_ros/moveit_servo/src/pose_tracking.cpp
@@ -231,20 +231,21 @@ void PoseTracking::targetPoseCallback(const geometry_msgs::PoseStampedConstPtr& 
 {
   std::lock_guard<std::mutex> lock(target_pose_mtx_);
   target_pose_ = *msg;
-  geometry_msgs::TransformStamped target_to_planning_frame;
+
+  // If the target pose is not defined in planning frame, transform the target pose.
   if (target_pose_.header.frame_id != planning_frame_)
   {
     try
     {
-      target_to_planning_frame = transform_buffer_.lookupTransform(planning_frame_, target_pose_.header.frame_id,
-                                                                   ros::Time(0), ros::Duration(0.1));
+      geometry_msgs::TransformStamped target_to_planning_frame = transform_buffer_.lookupTransform(
+          planning_frame_, target_pose_.header.frame_id, ros::Time(0), ros::Duration(0.1));
+      tf2::doTransform(target_pose_, target_pose_, target_to_planning_frame);
     }
-    catch (tf2::TransformException& ex)
+    catch (const tf2::TransformException& ex)
     {
-      ROS_WARN_NAMED(LOGNAME, "%s", ex.what());
+      ROS_WARN_STREAM_NAMED(LOGNAME, ex.what());
       return;
     }
-    tf2::doTransform(target_pose_, target_pose_, target_to_planning_frame);
   }
 }
 
--- a/moveit_ros/moveit_servo/src/pose_tracking.cpp
+++ b/moveit_ros/moveit_servo/src/pose_tracking.cpp
@@ -79,9 +79,6 @@ PoseTracking::PoseTracking(const ros::NodeHandle& nh,
 PoseTrackingStatusCode PoseTracking::moveToPose(const Eigen::Vector3d& positional_tolerance,
                                                 const double angular_tolerance, const double target_pose_timeout)
 {
-  // Roll back the target pose timestamp to ensure we wait for a new target pose message
-  target_pose_.header.stamp = ros::Time::now() - ros::Duration(2 * target_pose_timeout);
-
   // Wait a bit for a target pose message to arrive.
   // The target pose may get updated by new messages as the robot moves (in a callback function).
   ros::Time start_time = ros::Time::now();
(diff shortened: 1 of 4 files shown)


moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -220,59 +216,67 @@ bool PoseTracking::haveRecentEndEffectorPose(const double timespan)
...
+
+  // If the target pose is not defined in planning frame, transform the target pose.
   if (target_pose_.header.frame_id != planning_frame_)
   {
     try
     {
-      target_to_planning_frame = transform_buffer_.lookupTransform(planning_frame_, target_pose_.header.frame_id,
-                                                                   ros::Time(0), ros::Duration(0.1));
+      geometry_msgs::TransformStamped target_to_planning_frame = transform_buffer_.lookupTransform(
+          planning_frame_, target_pose_.header.frame_id, ros::Time(0), ros::Duration(0.1));
```
jliukkonen, line 239: I would actually argue that instead of `ros::Time(0)`, the transform timestamp should match the timestamp of the latest pose message. The reason being that if the target frame and planning frame move quickly w.r.t each other, the difference between latest transform and the transform at the time target pose was sent might be noticeable.

henningkayser: I would argue that's exactly the point why we should take the current time. Otherwise, we would add a lag when tracking a moving target.

henningkayser (approved):

moveit_ros/moveit_servo/src/pose_tracking.cpp
AndyZe: Well, the most recent available pose is going to be tracked regardless.

I'm inclined to say we should use target_pose_.header.stamp rather than `ros::Time(0)`, so we check if the target pose was _sent_ recently. In other words, account for network latency.

But, I'm ok either way.

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -79,12 +79,9 @@ PoseTracking::PoseTracking(const ros::NodeHandle& nh,
 PoseTrackingStatusCode PoseTracking::moveToPose(const Eigen::Vector3d& positional_tolerance,
                                                 const double angular_tolerance, const double target_pose_timeout)
 {
-  // Roll back the target pose timestamp to ensure we wait for a new target pose message
-  target_pose_.header.stamp = ros::Time::now() - ros::Duration(2 * target_pose_timeout);
```
AndyZe, line 83: I tried to explain on Slack that this line accounts for the use case where multiple waypoints are tracked, one after another. If a fresh target pose is late to arrive, we could begin moving toward the previous waypoint until a new target pose finally arrives (network latency). I saw this happen on a different project.

How about adding a public member function called `resetTargetPose()` where `target_pose_.header.stamp` is set to `ros::Time::now() - ros::Duration(2 * target_pose_timeout)` and all other fields are set to empty. That would give an entirely optional way to reset the target pose between waypoints.

If anybody has a better way to indicate that a std_msgs::Time field hasn't been initialized, that would be great. Here's the header file:  http://docs.ros.org/en/electric/api/std_msgs/html/Time_8h_source.html

AndyZe: We could use a separate Boolean flag, too

moveit_ros/moveit_servo/src/pose_tracking.cpp
jliukkonen: @henningkayser you are right. If the target pose frame and the planning frame happen to be different yet static w.r.t. each other, then it makes no difference. If the frames are different and _are_ moving, to eliminate the effect of the network latency and possible other delays that happen before call to `publish`, latest possible transform should be used. Now that I slept on it, I'm not sure what made me think we should use old timestamps.

moveit_ros/moveit_servo/src/pose_tracking.cpp
jliukkonen: My thinking here was that since the function claims to servo the arm to a pose that is fresh enough, it should do only that. The target pose is already updated asynchronously by using callback function, so doesn't the caller side already have everything it needs? Going through waypoints can be achieved by calling `moveToPose` once and then after that again with a new target.

AndyZe: `Going through waypoints can be achieved by calling moveToPose once and then after that again with a new target.`

The problem is, how do you know when the target pose is new? Will make a PR against your branch

jliukkonen: PRs are welcomed! And I think I finally understood what the problem is. You don't know whether the `target_pose_` has been updated prior to `moveToPose` function call. Would it make sense to remove the internal state altogether and just pass the latest pose as a parameter into the function? Or resetting as you proposed and waiting for a new pose inside the while loop.

AndyZe (approved):

PR #2397, opened 2020-10-28
WIP: Introduce interactive marker offset

machinekoder:
### Description

This patch introduces a new interactive marker offset property for the RViz plugin.

In the process I found a bug in the InteractionHandler, which applied the marker offset incorrectly.

This PR is work in progress, I still need to make screenshots and update the docs if applicable.

### Checklist

### DIFF
--- a/moveit_ros/robot_interaction/src/interaction_handler.cpp
+++ b/moveit_ros/robot_interaction/src/interaction_handler.cpp
@@ -368,19 +368,19 @@ bool InteractionHandler::transformFeedbackPose(const visualization_msgs::Interac
 {
   tpose.header = feedback->header;
   tpose.pose = feedback->pose;
+
+  tf2::Transform tf_tpose;
+
   if (feedback->header.frame_id != planning_frame_)
   {
     if (tf_buffer_)
+    {
       try
       {
         geometry_msgs::PoseStamped spose(tpose);
         // Express feedback (marker) pose in planning frame
         tf_buffer_->transform(tpose, spose, planning_frame_);
-        // Apply inverse of offset to bring feedback pose back into the end-effector support link frame
-        tf2::Transform tf_offset, tf_tpose;
-        tf2::fromMsg(offset, tf_offset);
         tf2::fromMsg(spose.pose, tf_tpose);
-        tf2::toMsg(tf_tpose * tf_offset.inverse(), tpose.pose);
       }
       catch (tf2::TransformException& e)
       {
@@ -388,6 +388,7 @@ bool InteractionHandler::transformFeedbackPose(const visualization_msgs::Interac
                   planning_frame_.c_str());
         return false;
       }
+    }
     else
     {
       ROS_ERROR("Cannot transform from frame '%s' to frame '%s' (no TF instance provided)",
@@ -395,6 +396,16 @@ bool InteractionHandler::transformFeedbackPose(const visualization_msgs::Interac
       return false;
     }
   }
+  else
+  {
+    tf2::fromMsg(tpose.pose, tf_tpose);
+  }
+
+  // Apply inverse of offset to bring feedback pose back into the end-effector support link frame
+  tf2::Transform tf_offset;
+  tf2::fromMsg(offset, tf_offset);
+  tf2::toMsg(tf_tpose * tf_offset.inverse(), tpose.pose);
+
   return true;
 }
 
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
@@ -74,6 +74,8 @@ class FloatProperty;
 class RosTopicProperty;
 class EditableEnumProperty;
 class ColorProperty;
+class VectorProperty;
+class QuaternionProperty;
 class MovableText;
 }  // namespace rviz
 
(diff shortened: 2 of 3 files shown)


felixvd:
Could you explain what problem this is meant to solve? I have difficulty imagining the use case.

machinekoder:
@felixvd Yes, tool-offsets run-time customizable by the user. 

And 16ab505 fixes a bug for a feature that doesn't seem to be used so far (but is used by this patch).

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
```diff
@@ -142,6 +144,16 @@ MotionPlanningDisplay::MotionPlanningDisplay()
                               plan_category_, SLOT(changedQueryMarkerScale()), this);
   query_marker_scale_property_->setMin(0.0f);
 
+  query_marker_offset_category_ = new rviz::Property("Interactive Marker Offset", QVariant(), "", plan_category_, nullptr, this);
```
rhaschke, line 147: Please provide more description details here. W.r.t. which coordinate system is the offset described?

rhaschke (commented):
Could you please explain in more detail, which bug you fixed in the first commit?
I didn't test this yet. But being able to specify a tool offset on the fly will be a useful feature. Thanks!

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
```diff
@@ -142,6 +144,16 @@ MotionPlanningDisplay::MotionPlanningDisplay()
...
+  query_marker_offset_position_property_ =
+      new rviz::VectorProperty("Position", Ogre::Vector3(0, 0, 0),
+                               "Specifies the interactive marker offset position.",
+                               query_marker_offset_category_, SLOT(changedQueryMarkerOffset()), this);
+  query_marker_offset_orientation_property_ =
+      new rviz::QuaternionProperty("Orientation", Ogre::Quaternion(1, 0, 0, 0),
+                               "Specifies the interactive marker offset orientation.",
+                               query_marker_offset_category_, SLOT(changedQueryMarkerOffset()), this);
+
```
tylerjw, line 156: This is just me being bothered by how QT works.  But I don't like the fact that these are class variables and we allocate them using new and then don't handle deleting them when this class is destroyed if they still exist.  If the lifetime of them is managed by some qt object we hand them to, maybe they should be local variables instead of class variables.

What you are doing is unrelated to the code smell of this file, carry on.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
```diff
@@ -820,11 +832,28 @@ void MotionPlanningDisplay::publishInteractiveMarkers(bool pose_update)
     }
     else
     {
+      geometry_msgs::Pose pose;
+      const auto position = query_marker_offset_position_property_->getVector();
+      pose.position.x = position.x;
+      pose.position.y = position.y;
+      pose.position.z = position.z;
+      const auto orientation = query_marker_offset_orientation_property_->getQuaternion();
```
tylerjw, line 840: Going back to my previous comment.  These two new members of the class (`query_marker_offset_position_property_` and `query_marker_offset_orientation_property_`) don't seem to be used anywhere else (they aren't being passed to anything that would own their lifetime).  As a result we should deallocate them when the destructor for MotionPlanningDisplay is called.

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
```diff
@@ -295,6 +298,9 @@ private Q_SLOTS:
   rviz::BoolProperty* query_start_state_property_;
   rviz::BoolProperty* query_goal_state_property_;
   rviz::FloatProperty* query_marker_scale_property_;
+  rviz::Property* query_marker_offset_category_;
+  rviz::VectorProperty*  query_marker_offset_position_property_;
+  rviz::QuaternionProperty* query_marker_offset_orientation_property_;
```
tylerjw, line 303: These are allocated on the heap, where is the lifetime of them being managed (where are they deallocated)?

tylerjw (commented):
Left a few comments regarding memory management.

moveit_ros/robot_interaction/src/interaction_handler.cpp
```diff
@@ -368,33 +368,44 @@ bool InteractionHandler::transformFeedbackPose(const visualization_msgs::Interac
 {
   tpose.header = feedback->header;
   tpose.pose = feedback->pose;
+
+  tf2::Transform tf_tpose;
```
felixvd, line 372: Please use more descriptive variable names. Same for `spose`

felixvd (commented):
I am happy to add this to Rviz for on-the-fly adjustments. Tool changing should be introduced differently though.

felixvd:
@machinekoder Do you think you will get around to this?

moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
simonschmeisser: This is handled by Qt automatically via parent-child relations. A parent QObject will delete all of it's children upon destruction. Children will sign off at their parent if they are destroyed before.

Issue #2262, opened 2020-08-18
Moveit finds solution that leads to collision between primitive and mesh collision objects


YannickRiou:
### Description

While using Moveit Task Constructor framework (based on Moveit), it appears that Moveit doesn't care about collision between a primitive collision object and a mesh collision object, at it find a "successful" plan solution. But when coming to execution, Moveit stops it as it find the primitive object is crashing into mesh collision object.

### Your environment
* ROS Distro: Kinetic
* OS Version: Ubuntu 16.04
* Moveit branch : Master
* Robot : PR2

Issue also open for MTC as it may be related (dont know if it's moveit or MTC the problem for now...) :  https://github.com/ros-planning/moveit_task_constructor/issues/180

Might be also related to this issue #1217  that is still open.

### Steps to reproduce

Load [this box as collision object](https://catdrop.drycat.fr/r/PQvBMYDm#LlTgUfNTZHzzrc2+CDQpe376j5hXl4jbj1shmBQRRDE=), and try to place a cube primitive object into it.

### Expected behaviour
Planner shouldn't find a solution (if too complicated) or should avoid the mesh box.

### Actual behaviour

The two gif below shows the **planned solution** to place the grasped primitive object into the mesh box.
No robot parts (gripper fingers) seems to touch the mesh box but the primitive collision object does. 

![crash_MTC_2](https://user-images.githubusercontent.com/59170639/90493576-d1aba680-e142-11ea-8b7d-56ec8798b222.gif)
![crash_MTC_3](https://user-images.githubusercontent.com/59170639/90493588-d53f2d80-e142-11ea-82e8-e4fe95f79a78.gif)

We thought at first that the problem was about the meshes files (square box) that were wrongly created (inverted normals), but we tried with STL and dae files with the same result.

We tried also to tweak the planner settings in order to avoid the trajectory waypoints to be far from each other and skip the collision object, but still having this problem. 

Here is the TRRT configuration we are using (extract from om_planning.yaml file): 

TRRT:  
type: geometric::TRRT  
range: 0.0  
goal_bias: 0.05  
max_states_failed: 10  
temp_change_factor: 1.0  
min_temperature: 10e-10  
init_temperature: 10e-6  
frountier_threshold: 0.0  
frountierNodeRatio: 0.1  
k_constant: 0.0  

### Backtrace or Console output

[...]

PR #2398, opened 2020-10-29
FixStartStateBounds: Copy attached bodies when adapting the start state

rhaschke:
This fixes https://github.com/ros-planning/moveit_task_constructor/issues/211, where an attached object was "lost" during planning. Turned out that the `FixStartStateBounds` adapter, while creating a new _non-diff_ `start_state`, did not copy all the attached objects, which are part of the RobotState as well. Thus with new joints, also attached objects were overridden.
Interestingly this bug was undetected over all the years...

### DIFF
--- a/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_bounds.cpp
+++ b/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_bounds.cpp
@@ -181,7 +181,7 @@ public:
     if (change_req)
     {
       planning_interface::MotionPlanRequest req2 = req;
-      moveit::core::robotStateToRobotStateMsg(start_state, req2.start_state, false);
+      moveit::core::robotStateToRobotStateMsg(start_state, req2.start_state);
       solved = planner(planning_scene, req2, res);
     }
     else


rhaschke:
I guess this will fix https://github.com/ros-planning/moveit/issues/2262 as well, because the object was attached there too.

felixvd:
Wow. Could you summarize how to reproduce this problem? What kind of test would have caught this?

rhaschke:
Seems to be a rare corner case. You need to have:
- a start state that is slightly outside the joint limits
- an object attached to the robot
- the attached object (and only that one) causing a collision that will be slip through unnoticed during planning
- FixStartStateBounds adapter enabled

Interestingly, the rviz display seems to correctly display the motion - including the attached object going through an obstacle. I have seen corresponding videos in the past, e.g. in #2262.
However, planning actually didn't observe a collision, because the attached object wasn't there during planning.

felixvd:
I see. If you have that case prepared somewhere, how about making it into a test? :)

rhaschke:
I don't have a test case lying around for this. As written in the initial comment, the issue arose from https://github.com/ros-planning/moveit_task_constructor/issues/211.

JafarAbdi (approved):

AndyZe (approved):

PR #2401, opened 2020-11-03
[feature] Low latency mode

tylerjw:
### Description

This change allows the user to tie the output of servo to the input.  By using a condition variable in this new `low_latency_mode` servo will run each time it gets input.

Includes a test that validates this works as expected.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -312,5 +312,6 @@ private:
 
   // input condition variable used for low latency mode
   std::condition_variable input_cv_;
+  bool new_input_cmd_ = false;
 };
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -39,7 +39,9 @@
 #pragma once
 
 // C++
+#include <condition_variable>
 #include <mutex>
+#include <thread>
 
 // ROS
 #include <control_msgs/JointJog.h>
@@ -69,10 +71,7 @@ public:
   ServoCalcs(ros::NodeHandle& nh, ServoParameters& parameters,
              const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor);
 
-  ~ServoCalcs()
-  {
-    timer_.stop();
-  }
+  ~ServoCalcs();
 
   /** \brief Start the timer where we do work and publish outputs */
   void start();
@@ -106,8 +105,14 @@ public:
   void changeRobotLinkCommandFrame(const std::string& new_command_frame);
 
 private:
-  /** \brief Timer method */
-  void run(const ros::TimerEvent& timer_event);
+  /** \brief thread main */
+  void mainCalcLoop();
+
+  /** \brief Logic of Servo, publishes one output */
+  void run();
+
+  /** \brief Stop the currently running thread */
+  void stop();
 
   /** \brief Do servoing calculations for Cartesian twist commands. */
   bool cartesianServoCalcs(geometry_msgs::TwistStamped& cmd, trajectory_msgs::JointTrajectory& joint_trajectory);
@@ -257,8 +262,6 @@ private:
   trajectory_msgs::JointTrajectoryConstPtr last_sent_command_;
 
   // ROS
-  ros::Timer timer_;
-  ros::Duration period_;
   ros::Subscriber joint_state_sub_;
   ros::Subscriber twist_stamped_sub_;
   ros::Subscriber joint_cmd_sub_;
@@ -269,6 +272,10 @@ private:
   ros::ServiceServer drift_dimensions_server_;
   ros::ServiceServer control_dimensions_server_;
   ros::ServiceServer reset_servo_status_;
+
+  // Main tracking / result publisher loop
+  std::thread thread_;
+  bool stop_requested_;
 
   // Status
   StatusCode status_ = StatusCode::NO_WARNING;
@@ -292,8 +299,8 @@ private:
   // The dimesions to control. In the command frame. [x, y, z, roll, pitch, yaw]
   std::array<bool, 6> control_dimensions_ = { { true, true, true, true, true, true } };
 
-  // latest_state_mutex_ is used to protect the state below it
-  mutable std::mutex latest_state_mutex_;
+  // input_mutex_ is used to protect the state below it
+  mutable std::mutex input_mutex_;
   Eigen::Isometry3d tf_moveit_to_robot_cmd_frame_;
   Eigen::Isometry3d tf_moveit_to_ee_frame_;
   geometry_msgs::TwistStampedConstPtr latest_twist_stamped_;
@@ -302,5 +309,8 @@ private:
   ros::Time latest_joint_command_stamp_ = ros::Time(0.);
   bool latest_nonzero_twist_stamped_ = false;
   bool latest_nonzero_joint_cmd_ = false;
+
+  // input condition variable used for low latency mode
+  std::condition_variable input_cv_;
 };
 }  // namespace moveit_servo
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -105,11 +105,11 @@ public:
   void changeRobotLinkCommandFrame(const std::string& new_command_frame);
 
 private:
-  /** \brief thread main */
+  /** \brief Run the main calculation loop */
   void mainCalcLoop();
 
-  /** \brief Logic of Servo, publishes one output */
-  void run();
+  /** \brief Do calculations for a single iteration. Publish one outgoing command */
+  void calculateSingleIteration();
 
   /** \brief Stop the currently running thread */
   void stop();
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -86,7 +86,7 @@ geometry_msgs::TransformStamped convertIsometryToTransform(const Eigen::Isometry
 // Constructor for the class that handles servoing calculations
 ServoCalcs::ServoCalcs(ros::NodeHandle& nh, ServoParameters& parameters,
                        const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor)
-  : nh_(nh), parameters_(parameters), planning_scene_monitor_(planning_scene_monitor), period_(parameters.publish_period)
+  : nh_(nh), parameters_(parameters), planning_scene_monitor_(planning_scene_monitor), stop_requested_(true)
 {
   // MoveIt Setup
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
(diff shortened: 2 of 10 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -182,25 +192,59 @@ void ServoCalcs::start()
...
 }
 
-void ServoCalcs::run(const ros::TimerEvent& timer_event)
+void ServoCalcs::main()
 {
-  // Log warning when the last loop duration was longer than the period
-  if (timer_event.profile.last_duration.toSec() > period_.toSec())
+  ros::Rate rate(1.0 / parameters_.publish_period);
+
+  while (ros::ok() && !stop_requested_)
```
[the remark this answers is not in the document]
AndyZe: We should be able to resume the main ServoCalcs() loop easily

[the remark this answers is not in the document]
AndyZe: Also, doesn't `main()` have kind of a special meaning in C++? Not sure this function should be named so

AndyZe (commented):
Overall, I'm not sure I like this. Say the incoming command rate is 250 Hz (4 ms). If I understand, this method might save 3 ms of latency on the first command. But the long-term average latency is still going to be 4 ms. So there's very little gain but more complexity and mutexes involved.

AndyZe:
I'm wondering if this could be simpler:

`std::atomic_bool_ got_first_command_ = false;`

```
void ServoCalcs::twistStampedCB(const geometry_msgs::TwistStampedConstPtr& msg)
{
  ...

  // notify that we have a new input
 got_first_command_ = true;
}
```

Reset got_first_command_ to false every time a stop is requested

moveit_ros/moveit_servo/src/servo_calcs.cpp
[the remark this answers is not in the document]
tylerjw: main doesn't have a special meaning when it is part of a class... it is a pattern i've seen to use main as the function you start a thread with.

If this function returns it can be joined (the thread is finished and could be joined) which will happen if someone calls start again.

tylerjw:
`condition_variable` is a stupid name for what it does.  It makes it easier for me to think of it as a mutex protected signal you can send from one thread to unblock another.  What this does is in `low_latency_mode` the main loop blocks and resumes imidiatly when it receives a new command.  If servo was previously running at 50Hz with another loop commanding it (like the pose tracker) at 50Hz turning this mode on would eliminate 20ms of input delay.  This mode is useful if it is being sent commands by another high rate loop as it just ties the rate of servo to that loop without any of the input latency.

> I'm wondering if this could be simpler:
> 
> `std::atomic_bool_ got_first_command_ = false;`
> 
> ```
> void ServoCalcs::twistStampedCB(const geometry_msgs::TwistStampedConstPtr& msg)
> {
>   ...
> 
>   // notify that we have a new input
>  got_first_command_ = true;
> }
> ```
> 
> Reset got_first_command_ to false every time a stop is requested

I don't understanding what you are suggesting with a `got_first_command_` bool.  What I'm trying to achieve here is to have servo unblock and calculate an output imidiatly after receiving a new input while still preserving the thread separation and nice ros interface it already has.

> Say the incoming command rate is 250 Hz (4 ms) and the controller accepts commands at 100 Hz (10 ms). If I understand, this method might save 3 ms of latency on the first command. But the long-term average command rate to the controller is still going to be 10 ms. So there's very little gain but more complexity.

I am probably misunderstanding something but I don't think it would be useful to run servo at a higher rate than the ros controller accepts commands as you would be calculating commands that would just be dropped at best.  Generally wouldn't you want your robot controller to be running at a higher (or the same) rate than servo or am I misunderstanding something?  Similarly if you are feeding servo commands at a higher rate than it is running roughly half of your commands will be dropped.  That also seems like it could have non-linear affects that would be generally bad.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -182,25 +190,79 @@ void ServoCalcs::start()
...
                            current_state_->getGlobalLinkTransform(parameters_.ee_frame_name);
   tf_moveit_to_robot_cmd_frame_ = current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
                                   current_state_->getGlobalLinkTransform(parameters_.robot_link_command_frame);
+
+  stop_requested_ = false;
+  thread_ = std::thread([this] { main(); });
 }
 
-void ServoCalcs::run(const ros::TimerEvent& timer_event)
+void ServoCalcs::stop()
```
AndyZe, line 203: This overlaps with the existing functionality, `setPaused()`

AndyZe:
It looks like it takes about 50 to a few hundred milliseconds to start a thread. Sorry Tyler, I'm not digging it

https://stackoverflow.com/questions/18274217/how-long-does-thread-creation-and-termination-take-under-windows

AndyZe:
> I don't understand what you are suggesting with a `got_first_command_` bool.

I'll expand my comment

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: It is a private function and not part of the exposed interface.  I created this function because I needed the block of code in the destructor and in start().

AndyZe:
On the bright side, one issue this PR brought to mind is that we should decrease this period just a bit in ServoCalcs so there's always a fresh command waiting to be published:

`  timer_ = nh_.createTimer(period_, &ServoCalcs::run, this);`

Should be something like:

`  timer_ = nh_.createTimer(**0.9 *** period_, &ServoCalcs::run, this);`

tylerjw:
> It looks like it takes about 5 to a few hundred milliseconds to start a thread. Sorry Tyler, I'm not digging it
> 
> https://stackoverflow.com/questions/18274217/how-long-does-thread-creation-and-termination-take-under-windows

Yes, we are already eating that cost with the AsyncSpinner.  I don't see how that matters though.  That time is spent in the start() method not in the loop.

tylerjw:
> On the bright side, one issue this PR brought to mind is that we should decrease this period just a bit in ServoCalcs so there's always a fresh command waiting to be published:
> 
> ` timer_ = nh_.createTimer(period_, &ServoCalcs::run, this);`
> 
> Should be something like:
> 
> ` timer_ = nh_.createTimer(0.9 * period_, &ServoCalcs::run, this);`

The user could just decrease the period in the config by 10% if that is what they wanted.  That will reduce the latency by 10% as that would just increase the frequency of the loop by 10%.  That is completely different from using a condition variable to tie running servo calcs to the input.

AndyZe:
> I don't understand what you are suggesting with a `got_first_command_` bool

You're right, this is hard to do now that the work is started by a timer:   `timer_ = nh_.createTimer(period_, &ServoCalcs::run, this);`

But it used to be done in a while loop, and then it would have been a lot easier:

```
while (// jog loop conditions)
{
  // Wait for first commands
  while(ros::ok() && !got_first_command_)
  {
    ros::Duration(0.001).sleep():
  }

  // Do jog calculations here

  ros::Duration(0.001).sleep():
}
```

AndyZe:
> The user could just decrease the period in the config by 10% if that is what they wanted. That will reduce the latency by 10% as that would just increase the frequency of the loop by 10%. That is completely different from using a condition variable to tie running servo calcs to the input.

Well, that's not what I'm talking about. We'll have to sync offline, I guess

tylerjw:
> But it used to be done in a while loop, and then it would have been a lot easier:
> 
> ```
> while (// jog loop conditions)
> {
>   // Wait for first command
>   while(ros::ok() && !got_first_command_)
>   {
>     ros::Duration(0.001).sleep():
>   }
> 
>   // Do jog calculations here
> 
>   ros::Duration(0.001).sleep():
> }
> ```

The condition variable does this without having to poll a variable.  The condition variable triggers the continuation of the calc thread as soon as the input callback returns (without tying up a thread just polling on a bool like this would do).

The advantage of using a timer is it doesn't create "yet another thread" and allows you to more optimally use your cpu in a external to servo way.  There is no way to trigger the running of a timer in this way.  One could put the calcuations in the callback but that will have a negative affect on the system of making everything in the process less responsive to new ros messages received over the network because one of the closures for servicing the callbacks will now take a relatively long time to return.  To me using a thread and condition variable seems like a nice compromise.

One thing to note is that this PR does not externally change servo for existing use cases where someone doesn't drive it with another high rate (or similar rate) loop.

AndyZe:
Can we take advantage of `timer_.start()` and `timer_.stop()` to do this in a less disruptive way?

tylerjw:
> Can we take advantage of `timer_.start()` and `timer_.stop()` to do this in a less disruptive way?

I don't think there is any guarentee that when you call start() it imidiatly runs.  It seems to me that using a timer for a single run on a separate thread is counter to what the timer is for.  The way I'm achieving that is to block the calc thread and unblock it in the callback.

This is not disrutive to the external user of servo and I think a condition variable is a good solution.  The largest downside to condition variables is they are not intuitive to use (I have to read the documentation and write tests to validate my understanding each time I use them).

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -182,25 +190,79 @@ void ServoCalcs::start()
...
+      if (stop_requested_)
+      {
+        break;
+      }
+    }
+
+    // run servo calcs
+    auto start_time = ros::Time::now();
+    run();
+    auto run_duration = ros::Time::now() - start_time;
```
jliukkonen, line 245: Nit: set both the start time and run duration as const.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -182,25 +190,79 @@ void ServoCalcs::start()
...
+
+  while (ros::ok() && !stop_requested_)
+  {
+    // lock the input state mutex
+    std::unique_lock<std::mutex> input_lock(input_mutex_);
+
+    // low latency mode
+    if (parameters_.low_latency_mode)
+    {
+      input_cv_.wait(input_lock);
```
jliukkonen, line 235: This is something I don't really know an answer to, but I assume the author does and curious to hear the answer. So, wouldn't this be affected by spurious wakeups? And would it be better to add a check for, e.g., a flag that would be set by either the `twistStampedCB` and `jointCmdCB` callback functions? That is, add a predicate to the `.wait()` method?

jliukkonen (commented):
Implementation and the idea both look very good to me. Left couple comments and after clearing those, I'm okay with merging this.

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: Yes, this could be triggered by a spurious wakeup.  I can add a predicate to this.

jliukkonen (approved):

moveit_ros/moveit_servo/src/servo_calcs.cpp
tylerjw: I added the predicate.  It now is not susceptible to the spurious wakeups a condition variable can be.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -106,8 +105,14 @@ class ServoCalcs
   void changeRobotLinkCommandFrame(const std::string& new_command_frame);
 
 private:
-  /** \brief Timer method */
-  void run(const ros::TimerEvent& timer_event);
+  /** \brief thread main */
```
AndyZe, line 108: Let's improve these comments and rename the functions:

  /** \brief Run the main calculation loop */
  void mainCalcLoop();

/** \brief Do calculations for a single iteration. Publishes one outgoing command */
void calculateSingleIteration();

AndyZe (commented):
Here's some fresh feedback after letting this simmer awhile. I still think this functionality is borderline not worthwhile (we're talking about time savings of maybe 2 ms for a typical scenario).

Let's get rid of the yaml parameter and always use "low latency mode" by default, since I can't think of a case where we wouldn't want to use it.

Also, delete the new test since it's not a new mode any more -- it will just be the default mode.

I still need to test it, too

tylerjw:
I'm sorry I didn't include this with the PR but I wrote some profiling code and here is the result from a modified version of the cpp_interface_test to publish the time between when servo receives a command (the callbacks) to when it publishes based on that input:

low_latency_v_normal

If you want the profiliing code message me, it's an annoyingly long diff.

AndyZe (approved):
Made a cleanup commit. Looks good to me now.

PR #2403, opened 2020-11-04
PlanExecution: Reset preempt_requested_ when entering executeAndMonitor()

rhaschke:
So far, `executeAndMonitor()` was only called via `planAndExecuteHelper()`, which was resetting the `preempt_requested_` flag:
https://github.com/ros-planning/moveit/blob/f85db808303cb7787320dc48162eabdc07593fdc/moveit_ros/planning/plan_execution/src/plan_execution.cpp#L164

However, this should also work if the function is called directly.
Fixes https://github.com/ros-planning/moveit_task_constructor/issues/217.

### DIFF
--- a/moveit_ros/planning/plan_execution/include/moveit/plan_execution/plan_execution.h
+++ b/moveit_ros/planning/plan_execution/include/moveit/plan_execution/plan_execution.h
@@ -133,7 +133,7 @@ public:
 
       In case there is no \e planning_scene or \e planning_scene_monitor set in the \e plan they will be set at the
       start of the method. They are then used to monitor the execution. */
-  moveit_msgs::MoveItErrorCodes executeAndMonitor(ExecutableMotionPlan& plan);
+  moveit_msgs::MoveItErrorCodes executeAndMonitor(ExecutableMotionPlan& plan, bool reset_preempted = true);
 
   void stop();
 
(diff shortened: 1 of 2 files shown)


rhaschke:
There is a small drawback of this change: If a `planAndExecute` call (going through the helper function) is preempted just between checking for the flag and calling the `executeAndMonitor()` function, the preemption signal will be essentially lost:
https://github.com/ros-planning/moveit/blob/f85db808303cb7787320dc48162eabdc07593fdc/moveit_ros/planning/plan_execution/src/plan_execution.cpp#L220-L224

felixvd (commented):
I don't think it will be dangerous, but it doesn't seem to be best practice to overwrite the flag. Losing a stop signal like this would be a very confusing bug. What about an input parameter like `bool initialize_preempt_flag` for executeAndMonitor, so the helper function can call `executeAndMonitor` as before?

rhaschke:
Yeah, I thought about that solution as well. Thanks for (independently) supporting this idea, which of course breaks ABI.
I committed an appropriate change.

felixvd (approved):
I would rather break API than introduce a potential time bomb :)

rhaschke:

rhaschke:
Closing in favor of #2554.

PR #2404, opened 2020-11-04
[Servo] Prevent race condition during initialization

AndyZe:

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -182,13 +182,13 @@ void ServoCalcs::start()
   initial_joint_trajectory->points.push_back(point);
   last_sent_command_ = initial_joint_trajectory;
 
-  timer_ = nh_.createTimer(period_, &ServoCalcs::run, this);
-
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
   tf_moveit_to_ee_frame_ = current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
                            current_state_->getGlobalLinkTransform(parameters_.ee_frame_name);
   tf_moveit_to_robot_cmd_frame_ = current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
                                   current_state_->getGlobalLinkTransform(parameters_.robot_link_command_frame);
+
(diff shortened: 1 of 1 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -182,13 +182,13 @@ void ServoCalcs::start()
...
 
-  timer_ = nh_.createTimer(period_, &ServoCalcs::run, this);
-
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
   tf_moveit_to_ee_frame_ = current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
                            current_state_->getGlobalLinkTransform(parameters_.ee_frame_name);
   tf_moveit_to_robot_cmd_frame_ = current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
                                   current_state_->getGlobalLinkTransform(parameters_.robot_link_command_frame);
+
+  timer_ = nh_.createTimer(period_, &ServoCalcs::run, this);
```
AndyZe, line 191: I moved this down to ensure that current_state_ and so on get initialized before `run()` begins

moveit_ros/moveit_servo/src/servo.cpp
```diff
@@ -132,27 +132,6 @@ bool Servo::readParameters()
...
-    {
-      parameters_.self_collision_proximity_threshold = collision_proximity_threshold;
-      have_self_collision_proximity_threshold = true;
-    }
-    if (!have_scene_collision_proximity_threshold)
-    {
-      parameters_.scene_collision_proximity_threshold = collision_proximity_threshold;
-      have_scene_collision_proximity_threshold = true;
-    }
-  }
```
tylerjw, line 155: Is there any harm in leaving this here for a couple of releases (it was in the last one) and removing it after the Nov 20 release?

tylerjw, line 155: Is there any harm in leaving this here for a couple of releases (it was in the last one) and removing it after the Nov 20 release?
@tylerjw

tylerjw (approved):

moveit_ros/moveit_servo/src/servo.cpp
AndyZe: No problem, I'll revert this block of code

Issue #2410, opened 2020-11-12
CollisionRequest min_cost_density arg isn't used


AndyZe:
I grepped for `min_cost_density` and it only appears in a [header file](https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection/include/moveit/collision_detection/collision_common.h). So I think the data member could be deleted.

Here's the description from (http://docs.ros.org/en/melodic/api/moveit_core/html/structcollision__detection_1_1CollisionRequest.html):

`When costs are computed, this is the minimum cost density for a CostSource to be included in the results.`

JafarAbdi:
Close by https://github.com/ros-planning/moveit/pull/2411

PR #2411, opened 2020-11-12
Delete CollisionRequest min_cost_density (unused)

AndyZe:
Fixes issue #2410


tylerjw (commented):
You only deleted the initialization and not the variable.  Maybe someone outside of the repo uses this?

AndyZe:
> Maybe someone outside of the repo uses this?

I think that's unlikely. If you google `min_cost_density`, all results are related to this header file. There are only 8 results

AndyZe:
> You only deleted the initialization and not the variable.

Geez, that's embarrassing. Fixed and tested again

tylerjw (approved):

PR #2037, opened 2020-04-27
Add object pose

felixvd:
~As discussed in https://github.com/ros-planning/moveit/issues/2025, this defines `CollisionObject`s with a `PoseStamped` instead of a `header`, so that an object can be moved together with its subframes and shapes.~

~I would like to get this in before the Noetic release and make it easy to review, so this PR includes only the very minimum that I believe this functionality needs. The tests passed locally, so I'd like to hear your thoughts.~

~Depends on https://github.com/ros-planning/moveit_msgs/pull/69.~

### New summary [here](https://github.com/ros-planning/moveit/pull/2037#issuecomment-709442682).

---

### Checklist
- [x] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)
- [X] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes
- [x] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html)
- [x] While waiting for someone to review your request, please help review [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers

PR #2415, opened 2020-11-17
WIP: Use Bullet for Servo collision detection

AndyZe:
This switches from FCL collision detection to Bullet collision detection. It's nice for a couple reasons:

- It doesn't lock the planning scene. No `getLockedPlanningSceneRO()`
- It runs faster. I dropped a high resolution timer around collision_check.cpp, Lines 115-129. Here are the results:

Bullet:  cycle time varied from 0.9-2.7 ms
FCL:    cycle time varied from 15-69 ms

Wow!

### DIFF
--- a/moveit_ros/moveit_servo/src/collision_check.cpp
+++ b/moveit_ros/moveit_servo/src/collision_check.cpp
@@ -85,9 +85,12 @@ CollisionCheck::CollisionCheck(ros::NodeHandle& nh, const moveit_servo::ServoPar
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
   acm_ = getLockedPlanningSceneRO()->getAllowedCollisionMatrix();
 
+  // Maintain this local copy of the planning scene, to reduce locking of planning_scene_monitor_
+  local_planning_scene_ = std::make_unique<planning_scene::PlanningScene>(
+    getLockedPlanningSceneRO()->getRobotModel()
+    );
   // Use Bullet collision detection, for speed
-  planning_scene_monitor::LockedPlanningSceneRW(planning_scene_monitor_)->
-    setActiveCollisionDetector(collision_detection::CollisionDetectorAllocatorBullet::create());
+  local_planning_scene_->setActiveCollisionDetector(collision_detection::CollisionDetectorAllocatorBullet::create());
 }
 
 planning_scene_monitor::LockedPlanningSceneRO CollisionCheck::getLockedPlanningSceneRO() const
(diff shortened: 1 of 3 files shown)


felixvd:
Thanks @j-petit and GSoC 🎉

j-petit:
Great to see Bullet in action!

moveit_ros/moveit_servo/src/collision_check.cpp
```diff
@@ -80,6 +80,12 @@ CollisionCheck::CollisionCheck(ros::NodeHandle& nh, const moveit_servo::ServoPar
 
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
   acm_ = getLockedPlanningSceneRO()->getAllowedCollisionMatrix();
+
+  // Set up Bullet collision detection
+  collision_det_allocation_ = std::make_shared<collision_detection::CollisionDetectorAllocatorBullet>();
+  const collision_detection::WorldPtr& world =
+      planning_scene_monitor::LockedPlanningSceneRW(planning_scene_monitor_)->getWorldNonConst();
+  collision_env_ = collision_det_allocation_->allocateEnv(world, getLockedPlanningSceneRO()->getRobotModel());
```
henningkayser, line 88: Won't this just take an initial snapshot of the collision world so that changes in the scene aren't handled at all? I definitely agree that we should decouple collision checks from the PSM. One way to achieve this would be to create a decoupled child scene that we keep updated for any changes in the parent. What do you think?

henningkayser (commented):
Awesome! I'm really excited to see this in action. How reliable are the distance measures?

moveit_ros/moveit_servo/src/collision_check.cpp
AndyZe: Transforms are updated but (I believe) new objects aren't detected. I'll test it.

Obviously I'd rather avoid copying the planning scene if this simple approach works.

henningkayser: If this approach works (and doesn't allow race conditions) I'm perfectly fine with this.

AndyZe:
Here are a couple videos from testing. It all looks really promising so far. Let me know if I should test anything else.

The videos are:

- [a simple test of self collisions. Distance to self-collision is streaming in the terminal](https://drive.google.com/file/d/1h6iBp_JmI8DV_LkZeFqJtg7m-bGIZzDA/view?usp=sharing)
- [self collision while moving quickly](https://drive.google.com/file/d/1LybUMVvl5mQaRkI4mtTECKz_sDZ6yAFA/view?usp=sharing)
- [collision with an object added to the planning scene](https://drive.google.com/file/d/1PnvxbDKYHbPvE8dyC0DmcDwm9UIA8pyg/view?usp=sharing)
- [remove a collision object, then re-add it several times](https://drive.google.com/file/d/15N-2gn3eKscSgGj2eUvehjUa8RzVtC1P/view?usp=sharing)

AndyZe:
I changed from Bullet back to FCL because there is an issue with Bullet checkSelfCollision() on a client project. May need to look into this later. This is still much faster than the previous implementation because it doesn't lock the planning scene.

AndyZe:
@j-petit Self collisions are driving me a bit nutty. It works great with a standalone UR5 simulation. But on a more complicated setup with some additional collision bodies added to the URDF, Bullet thinks most of the links are always in collision. Could you take a look at these functions and see if anything jumps out?

- [collision_detection::CollisionEnvBullet::checkSelfCollisionHelper](http://docs.ros.org/en/latest-available/api/moveit_core/html/classcollision__detection_1_1CollisionEnvBullet.html#a02fd324c9389c16fd1d101d569d4bf60)   I echoed the transforms here and they all seem reasonable. There are no cows (attached objects), only links from the URDF.

- [collision_detection_bullet::BulletDiscreteBVHManager::contactTest](https://docs.ros.org/en/noetic/api/moveit_core/html/classcollision__detection__bullet_1_1BulletDiscreteBVHManager.html#a5c6b8bbf8a2303c96cb3af2da2b927bd)

- [BulletDiscreteBVHManager::contactTest](https://docs.ros.org/en/noetic/api/moveit_core/html/bullet__discrete__bvh__manager_8cpp_source.html#l00089)

felixvd:
@AndyZe Could you add some tests in a separate PR? I would love to look into that for #2037. You said that they fail for you with Bullet but not FCL? If @j-petit has any cycles to spare that would be great, too.

j-petit:
Hm, I had a quick look at the functions but could not find anything particular related to the issue. As Felix said, a separate failing test would make it easier for me to look into it.

AndyZe:
Alright, I'll see if I have time to write a failing test. The latest commit (8dc4ee0) works great but it's not quite as clean as I would like

AndyZe:
It still seems that checkSelfCollision() just doesn't work. Made a little demo here, based on the tutorial.  https://github.com/ros-planning/moveit_tutorials/pull/560

AndyZe:
@j-petit you'll be happy to know, the issues on one particular setup have been figured out and Bullet wasn't the cause. :+1:  I'm going to open a new PR because I accidentally deleted this branch   :/

j-petit:
But [this](https://github.com/ros-planning/moveit_tutorials/pull/560) is still relevant, right?

AndyZe:
Yep

PR #2428, opened 2020-11-23
Make a plugin for differential IK solvers

AndyZe:
This converts Servo's differential inverse Jacobian IK solver into a plugin. This should make it interchangeable with other differential IK solvers.

No API/ABI changes. If this didn't work, the existing tests would catch it.

@sjahr


AndyZe:
it looks like there's no momentum to merge this or need for it, really

Issue #2285, opened 2020-08-30
Should we use M_PI or boost::math::constants::pi<double>() ?


JeroenDM:
They are used both, somethings in the same source file. For example `revolute_joint_model.cpp` uses boost pi on [this line](https://github.com/ros-planning/moveit/blob/7ad2bc7b86dad08061d98668ba34feba54bb05cc/moveit_core/robot_model/src/revolute_joint_model.cpp#L155) and just a couple of lines further on, `M_PI` is used.

A quick search on Github gives 15 occurrences of `M_PI` and 13 occurrences of `boost::math::constants::pi<double>()`. So cleaning it up would be easy.

Personally I feel that `M_PI` is much more readable. But there could be other issues I'm not aware of. If we have to use `boost::math::constants::pi<double>()`, I would still propose to make it into a const to make the code easier to reason about.

I'm sorry if I missed any earlier discussion on this, I could not find one.

rhaschke:
I guess this is a matter of taste and depends on the contributor. I would prefer `M_PI` for readability as well.

gavanderhoorn:
From a C++ perspective, using an untyped macro defined in a C header isn't necessarily the best idea.

If readability is important, aliasing the namespace could be an option (`namespace bmc = boost::math::constants`) or even creating a utility header with often used constants (`static constexpr pi = bmc::pi<double>()`) (pseudo-code, haven't checked).

JeroenDM:
@gavanderhoorn I don't know how I missed it, but the [MoveIt Style Guidelines](https://moveit.ros.org/documentation/contributing/code/) already agree with your point:
> Avoid C-style functions such as FLT_EPSILON - instead use std::numeric_limits<double>::epsilon()

So indeed we could go for `static constexpr double PI{ boost::math::constants::pi<double>() };` inside the namespace of the file where it is used. Then we can avoid code such as:

```C++
    {
      if (diff > 0.0)
        diff = 2.0 * boost::math::constants::pi<double>() - diff;
      else
        diff = -2.0 * boost::math::constants::pi<double>() - diff;
      state[0] = from[0] - diff * t;
      // input states are within bounds, so the following check is sufficient
      if (state[0] > boost::math::constants::pi<double>())
        state[0] -= 2.0 * boost::math::constants::pi<double>();
      else if (state[0] < -boost::math::constants::pi<double>())
        state[0] += 2.0 * boost::math::constants::pi<double>();
    }
```

[...]

PR #2434, opened 2020-11-27
Fix orientation constraints failing on satisfied configurations

JafarAbdi:
### Description

After applying the following fix https://github.com/ros-planning/moveit/pull/2414 a pipeline I’m working on isn’t working anymore (with a bunch of errors about constraints not statisfied one of them is `ros.moveit_planners_ompl.constrained_goal_sampler: More than 80% of the sampled goal states fail to satisfy the constraints imposed on the goal sampler. Is the constrained sampler working correctly?`)

The reason is, right now we're calculating the error rotation matrix and call eulerAngles from Eigen to find the angles’ errors which return them in the following ranges `[0:pi]x[-pi:pi]x[-pi:pi]` ([link](https://eigen.tuxfamily.org/dox/group__Geometry__Module.html#ga17994d2e81b723295f5bc3b1f862ed3b))

The issues are
1- As shown in the first test, despite the error is 1.57rad around Z-axis, eulerAngles will return x=3.139981, y=3.139841, z=1.570810 causing the constraint to fail
2- Another example in the second test `0.15, -pi/2, 0.15` eulerAngles will return `3.05343, -1.5708, 3.05343` which again will cause the constraint to fail ([link](https://gcc.godbolt.org/z/4fvd5z))
3- This function isn’t stable around the identity rotation [link#1](https://github.com/stack-of-tasks/pinocchio/pull/1158#issuecomment-615105106) and [link#2](https://gcc.godbolt.org/z/Ycboa8)

~Even if we use the function from `unsupported/Eigen` it will cause similar problems but in different orientations for example if the error rotation matrix correspond to pure rotation about Y with angles bigger than pi/2 the returned roll/yaw angles will cause a satisfied constraint to fails again~

~To fix this I used the Euler angles of the desired and the current orientations to find the angles’ errors~

To fix this I did copy the implementation from [unsupported/Eigen/EulerSystem.h](https://gitlab.com/libeigen/eigen/-/blob/master/unsupported/Eigen/src/EulerAngles/EulerSystem.h#L187) since it’s not released yet which will return the angles in the following ranges `[-pi:pi]x[-pi/2:pi/2]x[-pi:pi]` and handled the singularities differently when we compare the error with the tolerances

~TODO: I need to add more tests before merging this~

### Checklist

### DIFF
--- a/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
+++ b/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
@@ -66,11 +66,6 @@ static double normalizeAbsoluteAngle(const double angle)
   return std::min(2 * M_PI - normalized_angle, normalized_angle);
 }
 
-static Eigen::Vector3d normalizeAbsoluteAngles(const Eigen::Vector3d& angles)
-{
-  return angles.unaryExpr(&normalizeAbsoluteAngle);
-}
-
 /**
  * This's copied from
  * https://gitlab.com/libeigen/eigen/-/blob/master/unsupported/Eigen/src/EulerAngles/EulerSystem.h#L187
@@ -692,7 +687,7 @@ ConstraintEvaluationResult OrientationConstraint::decide(const moveit::core::Rob
     }
   }
   // Account for angle wrapping
-  xyz = normalizeAbsoluteAngles(xyz);
+  xyz = xyz.unaryExpr(&normalizeAbsoluteAngle);
 
   // 0,1,2 corresponds to XYZ, the convention used in sampling constraints
   bool result = xyz(2) < absolute_z_axis_tolerance_ + std::numeric_limits<double>::epsilon() &&
--- a/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
+++ b/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
@@ -91,10 +91,10 @@ CalcEulerAngles(const Eigen::MatrixBase<Derived>& R)
   const Index j = 1;
   const Index k = 2;
   Eigen::Matrix<Scalar, 3, 1> res;
-  const Scalar Rsum = sqrt((R(i, i) * R(i, i) + R(i, j) * R(i, j) + R(j, k) * R(j, k) + R(k, k) * R(k, k)) / 2);
-  res[1] = atan2(R(i, k), Rsum);
+  const Scalar rsum = sqrt((R(i, i) * R(i, i) + R(i, j) * R(i, j) + R(j, k) * R(j, k) + R(k, k) * R(k, k)) / 2);
+  res[1] = atan2(R(i, k), rsum);
   // There is a singularity when cos(beta) == 0
-  if (Rsum > 4 * Eigen::NumTraits<Scalar>::epsilon())
+  if (rsum > 4 * Eigen::NumTraits<Scalar>::epsilon())
   {  // cos(beta) != 0
     res[0] = atan2(-R(j, k), R(k, k));
     res[2] = atan2(-R(i, j), R(i, i));
@@ -663,14 +663,17 @@ ConstraintEvaluationResult OrientationConstraint::decide(const moveit::core::Rob
   std::tuple<Eigen::Vector3d, bool> euler_angles_error;
   if (mobile_frame_)
   {
+    // getFrameTransform() returns a valid isometry by contract
     Eigen::Matrix3d tmp = state.getFrameTransform(desired_rotation_frame_id_).linear() * desired_rotation_matrix_;
+    // getGlobalLinkTransform() returns a valid isometry by contract
     Eigen::Isometry3d diff(tmp.transpose() * state.getGlobalLinkTransform(link_model_).linear());  // valid isometry
-    euler_angles_error = CalcEulerAngles(diff.rotation());
+    euler_angles_error = CalcEulerAngles(diff.linear());
   }
   else
   {
+    // diff is valid isometry by construction
     Eigen::Isometry3d diff(desired_rotation_matrix_inv_ * state.getGlobalLinkTransform(link_model_).linear());
-    euler_angles_error = CalcEulerAngles(diff.rotation());
+    euler_angles_error = CalcEulerAngles(diff.linear());
   }
 
   // Converting from a rotation matrix to an intrinsic XYZ euler angles have 2 singularities:
@@ -688,6 +691,7 @@ ConstraintEvaluationResult OrientationConstraint::decide(const moveit::core::Rob
       xyz(0) = 0;
     }
   }
+  // Account for angle wrapping
   xyz = normalizeAbsoluteAngles(xyz);
 
   // 0,1,2 corresponds to XYZ, the convention used in sampling constraints
--- a/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
+++ b/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
@@ -58,7 +58,7 @@ static double normalizeAngle(double angle)
   return v;
 }
 
-// Normalizes an angles to an absolute angle.
+// Normalizes an angle to the interval [-pi, +pi] and then take the absolute value
 // The returned values will be in the following range [0, +pi]
 static double normalizeAbsoluteAngle(const double angle)
 {
--- a/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
+++ b/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
@@ -58,6 +58,63 @@ static double normalizeAngle(double angle)
   return v;
 }
 
+// Normalizes an angles to an absolute angle.
+// The returned values will be in the following range [0, +pi]
+static double normalizeAbsoluteAngle(const double angle)
+{
+  const double normalized_angle = std::fmod(std::abs(angle), 2 * M_PI);
+  return std::min(2 * M_PI - normalized_angle, normalized_angle);
+}
+
+static Eigen::Vector3d normalizeAbsoluteAngles(const Eigen::Vector3d& angles)
+{
+  return angles.unaryExpr(&normalizeAbsoluteAngle);
+}
+
+/**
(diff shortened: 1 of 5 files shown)


mamoll:
@JeroenDM this is relevant for you, too.

mamoll (approved):
This is some nice detective work, @JafarAbdi!

JeroenDM:
@JafarAbdi Wow, thank you for figuring this out!

Since #2414 was merged my tests for #2402 failed. I suspect you also found the solution to my problem :)
I used some hardcoded test data [here](https://github.com/JeroenDM/moveit/blob/49df9efcbf81cff25ab6875023d57931869ed773/moveit_core/kinematic_constraints/test/test_orientation_constraints.cpp#L52), generated with a [Python script](https://gist.github.com/JeroenDM/f8ba5f1a60821c4cacf7c137eff6f409). If you think this is a good approach I can add the test for Euler angles to this PR.

moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
```diff
@@ -607,23 +650,26 @@ ConstraintEvaluationResult OrientationConstraint::decide(const moveit::core::Rob
...
+    const Eigen::Vector3d current_rpy = CalcEulerAngles(state.getGlobalLinkTransform(link_model_).linear());
+    xyz = current_rpy - desired_rpy;
   }
-
   // Account for angle wrapping
-  xyz(0) = std::min(fabs(xyz(0)), boost::math::constants::two_pi<double>() - fabs(xyz(0)));
-  xyz(1) = std::min(fabs(xyz(1)), boost::math::constants::two_pi<double>() - fabs(xyz(1)));
-  xyz(2) = std::min(fabs(xyz(2)), boost::math::constants::two_pi<double>() - fabs(xyz(2)));
+  xyz = xyz.unaryExpr([](const double angle) {
+    const double normalized_angle = std::fmod(std::abs(angle), boost::math::constants::two_pi<double>());
```
JeroenDM, line 668: A small nit that was already in the existing code, but would you mind using `namespace bmc = boost::math::constants` or `M_PI` depending on your stance on  #2285. I think the extremely long name for `pi` really makes it harder to reason about the code.

(I know this is a code base wide issue, so feel free to ignore it in this PR.)

moveit_core/utils/src/robot_model_test_utils.cpp
```diff
@@ -110,7 +110,7 @@ RobotModelBuilder::RobotModelBuilder(const std::string& name, const std::string&
 }
 
 void RobotModelBuilder::addChain(const std::string& section, const std::string& type,
-                                 const std::vector<geometry_msgs::Pose>& joint_origins)
+                                 const std::vector<geometry_msgs::Pose>& joint_origins, urdf::Vector3 joint_axis)
```
JeroenDM, line 113: Nice! I've missed this but it never occurred to me that I could just add it myself :)

moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
```diff
@@ -58,6 +58,63 @@ static double normalizeAngle(double angle)
   return v;
 }
 
+// Normalizes an angles to an absolute angle.
```
JeroenDM, line 61:
```suggestion
// Normalizes an angle to the interval [-pi, +pi] and then take the absolute value
```

I know this explanation is not how the code does it, but I think it better explains what is happening. Or am I wrong in what the code does?

JeroenDM:
+1 For explicitly handling and explaining the singular case, great work! @JafarAbdi 

I was thinking the following case could maybe surprise some users:
- Orientation tolerance of (0, 0, 1) on roll-pitch-yaw angles.
- Robot at the singular case with theta = 0.5.
In this case, the yaw tolerance is not violated by theta, and therefore theta will be interpreted as pure roll.
This results in roll tolerance being violated.

Is this a correct interpretation?

JafarAbdi:
> I used some hardcoded test data [here](https://github.com/JeroenDM/moveit/blob/49df9efcbf81cff25ab6875023d57931869ed773/moveit_core/kinematic_constraints/test/test_orientation_constraints.cpp#L52), generated with a [Python script](https://gist.github.com/JeroenDM/f8ba5f1a60821c4cacf7c137eff6f409). If you think this is a good approach I can add the test for Euler angles to this PR.

Sorry, I missed this comment, I'm okay with adding them here, or if you prefer I could open my PR against your branch to merge them together

JafarAbdi:
> I was thinking the following case could maybe surprise some users:
> 
> * Orientation tolerance of (0, 0, 1) on roll-pitch-yaw angles.
> * Robot at the singular case with theta = 0.5.
>   In this case, the yaw tolerance is not violated by theta, and therefore theta will be interpreted as pure roll.
>   This results in roll tolerance being violated.
> 
> Is this a correct interpretation?
> 
> **Edit**:
> I don't know how often this singular case is relevant in practical problems, but if my interpretation is correct, an alternative could be to check theta against the maximum of the roll and yaw tolerance.

As you can see in the image below given the tolerances this should pass, I added it as a test case

Screenshot from 2020-12-07 15-36-05

Also, I forgot to mention that in the last commits I reverted my previous approach and used the error rotation matrix to check for tolerances since it's easier to handle singularities

moveit_core/kinematic_constraints/test/test_orientation_constraints.cpp
```diff
@@ -0,0 +1,244 @@
...
+  robot_state.setVariablePositions(getJointValues(0.21, boost::math::constants::half_pi<double>(), 0.0));
+  robot_state.update();
+  EXPECT_TRUE(oc.configure(ocm, tf));
+  EXPECT_FALSE(oc.decide(robot_state).satisfied);
+
+  robot_state.setVariablePositions(getJointValues(0.0, boost::math::constants::half_pi<double>(), 0.21));
+  robot_state.update();
+  EXPECT_TRUE(oc.configure(ocm, tf));
+  EXPECT_FALSE(oc.decide(robot_state).satisfied);
+
```
JeroenDM, line 143:
```suggestion

  // In the non-singular case the absolute_x_axis_tolerance is not violated
  robot_state.setVariablePositions(getJointValues(0.0, boost::math::constants::half_pi<double>() - 0.01, 0.21));
  robot_state.update();
  EXPECT_TRUE(oc.configure(ocm, tf));
  EXPECT_TRUE(oc.decide(robot_state).satisfied);

```

JeroenDM:
> Also, I forgot to mention that in the last commits I reverted my previous approach and used the error rotation matrix to check for tolerances since it's easier to handle singularities

Ah, ok, now I understand it :) I glanced over that change :s
I suggested an extra test case to clarify the behaviour.

> Sorry, I missed this comment, I'm okay with adding them here, or if you prefer I could open my PR against your branch to merge them together

Given you already wrote extensive tests now, I don't' think my other tests would add value here. I will just extend your test file for the new parameterization once this is merged.

JeroenDM (approved):

moveit_core/kinematic_constraints/test/test_orientation_constraints.cpp
```diff
@@ -0,0 +1,244 @@
...
+  ocm.orientation.y = 0.0;
+  ocm.orientation.z = 0.0;
+  ocm.orientation.w = 1.0;
+  ocm.absolute_x_axis_tolerance = 0.2;
+  ocm.absolute_y_axis_tolerance = 2.0;
+  ocm.absolute_z_axis_tolerance = 0.3;
+  ocm.weight = 1.0;
+
+  moveit::core::RobotState robot_state(robot_model_);
+  // Singularity: roll + yaw = theta
```
JeroenDM, line 179:
```suggestion
// Singularity: roll - yaw = theta
```

JeroenDM: Because the tests below use -pi / 2.

JeroenDM:
Also, next time I will properly organize my comments in a review... sorry.

AndyZe:
I just want to make sure you have especially tested several values aroud pitch = pi/2. That's where singularities occur for the Euler angle roll/pitch/yaw convention, I believe. See [here](https://www.astro.rug.nl/software/kapteyn-beta/_downloads/attitude.pdf), page 13.

singularities

moveit_core/kinematic_constraints/src/kinematic_constraint.cpp
```diff
@@ -58,6 +58,58 @@ static double normalizeAngle(double angle)
   return v;
 }
 
+// Normalizes an angle to the interval [-pi, +pi] and then take the absolute value
+// The returned values will be in the following range [0, +pi]
```
AndyZe, line 62: Minor correction:

// Normalizes the absolute value of an angle to [0, +pi]
// Returned values will be in the range [0, +pi]

AndyZe (approved):
Overall I think a better way to do this would be to convert to a rotation vector because it is unique and does not have singularities. But I'm OK getting this merged in the meantime because it is breaking a lot of CI setups.

rhaschke (commented):
Looks reasonable. Will fix #2449 as well.
However, there is a failing (ikfast) test here. Not sure that is related. I retriggered Travis.

AndyZe:
Let's run CI one more time

gavanderhoorn:
Off-topic, but you don't need to close and re-open a PR to rerun Travis: if you don't see a "rerun" or "restart" button on the *Checks* tab of the PR, you can follow the `details` link to the `travis-ci.com` site, login with your Github account and then restart a run there.

AndyZe:
I should have squash-merged  :/

rhaschke:
> I should have squash-merged :/

Yes, indeed. I cleaned up and force-pushed. Please update your devel branches @tylerjw, @AndyZe, @JafarAbdi.

rhaschke:
@AndyZe: You pushed a branch `revert-2434-pr-fix_orientation_constraints` to the main repo. Was that intended to revert the accidental rebase merge? That would have made it even worse (adding even more commits...) If that branch is not needed anymore, please delete it.

AndyZe:
done!

PR #2441, opened 2020-12-01
[WIP] Switch to Bullet collision detection in Servo.

AndyZe:
This is a follow-up to PR #2415  (closed accidentally). It uses Bullet collision detection instead of FCL.

It's been tested pretty heavily on two different setups now, so I feel confident in the changes.

Bullet runs much faster than FCL. Here are some run times for one collision checking cycle:

Bullet:  0.2-0.5ms
Previous implementation with FCL: 15-69ms

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/collision_check.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/collision_check.h
@@ -39,6 +39,8 @@
 #pragma once
 
 #include <moveit/collision_detection/collision_common.h>
+#include <moveit/collision_detection_bullet/collision_detector_allocator_bullet.h>
+#include <moveit/collision_detection_bullet/collision_env_bullet.h>
 #include <moveit/robot_model_loader/robot_model_loader.h>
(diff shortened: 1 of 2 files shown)


AndyZe:
Videos in action:

- [a simple test of self collisions. Distance to self-collision is streaming in the terminal](https://drive.google.com/file/d/1h6iBp_JmI8DV_LkZeFqJtg7m-bGIZzDA/view?usp=sharing)
- [self collision while moving quickly](https://drive.google.com/file/d/1LybUMVvl5mQaRkI4mtTECKz_sDZ6yAFA/view?usp=sharing)
- [collision with an object added to the planning scene](https://drive.google.com/file/d/1PnvxbDKYHbPvE8dyC0DmcDwm9UIA8pyg/view?usp=sharing)
- [remove a collision object, then re-add it several times](https://drive.google.com/file/d/15N-2gn3eKscSgGj2eUvehjUa8RzVtC1P/view?usp=sharing)

jliukkonen:
Tested this briefly. Looks like collisions are being reported between all(?) links. So something is still missing. :-)

PR #2453, opened 2020-12-08
Upgrade cmake_minimum_required to 3.1

rhaschke:
to suppressing CMP0048 warning in Noetic.
@jschleicher


tylerjw (approved):
This causes no issues in Melodic either, as the default cmake for 18.04 is 3.10.2.

PR #2456, opened 2020-12-09
Provide a function to set the position of active joints in a JointModelGroup

### LINKED ISSUE #2442: RobotState::setJointGroupPositions: Check for matching size of JointModelGroup and passed vector
https://github.com/ros-planning/moveit/blob/bdd4bde4938c7bdac3a57a791cf738146c8834e0/moveit_core/robot_state/include/moveit/robot_state/robot_state.h#L625

would be nice to have an assert here and might have hinted to https://github.com/ros-planning/moveit_visual_tools/issues/82  and/or similar bugs in our products code

Will open PR

simonschmeisser:
### Description

this fixes #2442 and provides an less error prone alternative for https://github.com/ros-planning/moveit_visual_tools/pull/83

The general problem is that while usually you only concern yourself with the values of active joints when storing positions in a vector, the function `setJointGroupPositions` actually expects a vector/array containing the variables for *all* joints (including mimic joints). Ignoring this will lead to reading out-of-bounds memory at least and to a mismatch at worst (yes, there where physical crashes ...). Also the code might crash.

@felixvd suggested fixing `setJointGroupPositions` to actually work with only the variables of active joints. I have no objections except this being quite hidden and incompatible if anybody actually uses these functions correctly (I found 12 places in Mikado where we did not ...)

The second commit contains a proof of concept for adding another function with the intended behavior, this would preserve compatibility.

I'm open for either way, let the discussions begin!

### Checklist

---
Has been rebased to master, please squash merge as I failed to properly clang-format the first commit anyway

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -614,7 +614,8 @@ public:
   void setJointGroupPositions(const std::string& joint_group_name, const std::vector<double>& gstate)
   {
     const JointModelGroup* jmg = robot_model_->getJointModelGroup(joint_group_name);
-    if (jmg) {
+    if (jmg)
+    {
       assert(gstate.size() == jmg->getVariableCount());
       setJointGroupPositions(jmg, &gstate[0]);
     }
@@ -637,7 +638,8 @@ public:
   void setJointGroupPositions(const std::string& joint_group_name, const Eigen::VectorXd& values)
   {
     const JointModelGroup* jmg = robot_model_->getJointModelGroup(joint_group_name);
-    if (jmg) {
+    if (jmg)
+    {
       assert(values.size() == jmg->getVariableCount());
       setJointGroupPositions(jmg, values);
     }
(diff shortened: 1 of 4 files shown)


JeroenDM:
Just to mention it explicitly here: I think the conclusion in the MoveIt WG meeting today was to go for the new function `setJointGroupActivePositions`. But correct me if I'm wrong :) @simonschmeisser @felixvd

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -635,13 +639,52 @@ class RobotState
...
+
+  /** \brief Given positions for the variables of active joints that make up a group,
+   * in the order found in the group (excluding values of mimic joints), set those
+   * as the new values that correspond to the group */
+  void setJointGroupActivePositions(const std::string& joint_group_name, const std::vector<double>& gstate)
+  {
+    const JointModelGroup* jmg = robot_model_->getJointModelGroup(joint_group_name);
+    if (jmg)
+    {
+      assert(gstate.size() == jmg->getVariableCount());
```
rhaschke, line 665: Shouldn't this read:
```suggestion
      assert(gstate.size() == jmg->getActiveVariableCount());
```

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -635,13 +639,52 @@ class RobotState
...
+
+  /** \brief Given positions for the variables of active joints that make up a group,
+   * in the order found in the group (excluding values of mimic joints), set those
+   * as the new values that correspond to the group */
+  void setJointGroupActivePositions(const std::string& joint_group_name, const Eigen::VectorXd& values)
+  {
+    const JointModelGroup* jmg = robot_model_->getJointModelGroup(joint_group_name);
+    if (jmg)
+    {
+      assert(values.size() == jmg->getVariableCount());
```
rhaschke, line 683: Same here.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -511,6 +511,30 @@ void RobotState::setJointGroupPositions(const JointModelGroup* group, const Eige
   updateMimicJoints(group);
 }
 
+void RobotState::setJointGroupActivePositions(const JointModelGroup* group, const std::vector<double>& gstate)
+{
+  assert(gstate.size() == group->getActiveVariableCount());
```
rhaschke, line 516: ...because, here, you test against `getActiveVariableCount()`.

rhaschke (commented):
I'm confused about your asserts. Please double check.

moveit_core/robot_state/src/robot_state.cpp
simonschmeisser: then thankfully I got it right at least one time ...

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
simonschmeisser: fixed, thanks for spotting it!

rhaschke (approved):

moveit_core/robot_state/src/robot_state.cpp
simonschmeisser: Actually this was not right either ... this function only exists in documentation but not in code ... So I could either add it (iterate over `getActiveJointModels` and count `variableSize()` **or** use `getActiveJointModels().size()` and add this to the number of places where multi-variable joints are not correctly supported  **or**  add a member variable that caches the number but that would break ABI compatibility when backporting (but then we don't guarantee that anyway)

rhaschke: Hm. I think, I would prefer the last option: adding a member variable to cache that value.

simonschmeisser: done

rhaschke (approved):

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -511,6 +511,30 @@ void RobotState::setJointGroupPositions(const JointModelGroup* group, const Eige
...
+  std::size_t i = 0;
```
AndyZe, line 517: Random tidbit: you don't need `std::`

`size_t` alone is fine. There are already some places in moveit_core which do this. For example:

robot_state/test/robot_state_benchmark.cpp:    `for (size_t i = 0; i < runs; ++i)`

But of course there is no harm leaving it as is. [Stackoverflow link](https://stackoverflow.com/a/5813736/3499467)

AndyZe (approved):
Looks good to me. I'll let it sit for a day or so, then merge if nobody else does

PR #2461, opened 2020-12-16
Fix doxygen documentation for setToIKSolverFrame

### LINKED ISSUE #2460: Confusing documentation on setToIKSolverFrame()
I think the documentation [here](http://docs.ros.org/en/melodic/api/moveit_core/html/classmoveit_1_1core_1_1RobotState.html#a34d8fc133b4a21527ea0875fdb423e25) on setToIKSolverFrame() must be wrong. Here's what it says:

```
bool moveit::core::RobotState::setToIKSolverFrame ( Eigen::Isometry3d& pose, const std::string& ik_frame)

pose - the input to change
ik_frame - the name of frame of reference of base of ik solver
```

If that were correct, how would we know what the original frame of `pose` is? I think it should say:

```
bool moveit::core::RobotState::setToIKSolverFrame | ( Eigen::Isometry3d& pose, const std::string& ik_frame)

pose - the input to change
ik_frame - the **original** frame of reference of pose
```

AndyZe:
I think the documentation [here](http://docs.ros.org/en/melodic/api/moveit_core/html/classmoveit_1_1core_1_1RobotState.html#a34d8fc133b4a21527ea0875fdb423e25) on setToIKSolverFrame() must be wrong. Here's what it says:

```
bool moveit::core::RobotState::setToIKSolverFrame ( Eigen::Isometry3d& pose, const std::string& ik_frame)

pose - the input to change
ik_frame - the name of frame of reference of base of ik solver
```

If that were correct, how would we know what the original frame of `pose` is? I think it should say:

```
bool moveit::core::RobotState::setToIKSolverFrame | ( Eigen::Isometry3d& pose, const std::string& ik_frame)

pose - the input to change
ik_frame - the **previous** frame of reference of the pose
```

Fixes #2460

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -900,14 +900,6 @@ public:
    *  @{
    */
 
-  /**
-   * \brief Transform pose from the robot model's base frame to the reference frame of the IK solver
-   * @param pose - the input to change
-   * @param solver - a kin solver whose base frame is important to us
-   * @return true if no error
-   */
-  bool setToIKSolverFrame(Eigen::Isometry3d& pose, const kinematics::KinematicsBaseConstPtr& solver);
-
   /** \brief If the group this state corresponds to is a chain and a solver is available, then the joint values can be
      set by computing inverse kinematics.
       The pose is assumed to be in the reference frame of the kinematic model. Returns true on success.
@@ -1752,10 +1744,18 @@ private:
   /**
    * \brief Transform pose from the robot model's base frame to the reference frame of the IK solver
    * @param pose - the input to change
+   * @param solver - a kin solver whose base frame is important to us
+   * @return true if no error
+   */
+  inline bool setToIKSolverFrame(Eigen::Isometry3d& pose, const kinematics::KinematicsBaseConstPtr& solver);
+
+  /**
+   * \brief Transform pose from the robot model's base frame to the reference frame of the IK solver
+   * @param pose - the input to change
    * @param ik_frame - the name of frame of reference of base of ik solver
    * @return true if no error
    */
-  bool setToIKSolverFrame(Eigen::Isometry3d& pose, const std::string& ik_frame);
+  inline bool setToIKSolverFrame(Eigen::Isometry3d& pose, const std::string& ik_frame);
 
   void markDirtyJointTransforms(const JointModel* joint)
   {
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -901,20 +901,12 @@ public:
    */
 
   /**
-   * \brief Transform the frame of reference of the pose to that same frame as the IK solver expects
+   * \brief Transform pose from the robot model's base frame to the reference frame of the IK solver
    * @param pose - the input to change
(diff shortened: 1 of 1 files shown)


moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -901,17 +901,17 @@ class RobotState
    */
 
   /**
-   * \brief Convert the frame of reference of the pose to that same frame as the IK solver expects
+   * \brief Transform the frame of reference of the pose to that same frame as the IK solver expects
```
rhaschke, line 904:
```suggestion
   * \brief Transform pose from the robot model's base frame to the reference frame of the IK solver
```

rhaschke (changes requested):
No. Looking into the code, `ik_frame` indeed refers to the solver's frame, while it is assumed that pose is given w.r.t. the robot model's base frame.

I suggest making this function private (`static`) within `robot_state.cpp`. It's only used therein and probably not very useful externally.

AndyZe:
> I suggest making this function private (`static`) within `robot_state.cpp`. It's only used therein and probably not very useful externally.

Can't be static because it uses a member variable (robot_model_). Sounds good to me, otherwise

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -901,21 +901,13 @@ class RobotState
    */
 
   /**
-   * \brief Convert the frame of reference of the pose to that same frame as the IK solver expects
+   * \brief Transform pose from the robot model's base frame to the reference frame of the IK solver
    * @param pose - the input to change
    * @param solver - a kin solver whose base frame is important to us
    * @return true if no error
    */
   bool setToIKSolverFrame(Eigen::Isometry3d& pose, const kinematics::KinematicsBaseConstPtr& solver);
```
rhaschke, line 909: Please make this other function private as well. It could become inline also...

rhaschke (changes requested):
Please make both functions private.

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
AndyZe: You mean, make both functions versions `inline`, right?

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1749,13 +1741,21 @@ class RobotState
...
+  inline bool setToIKSolverFrame(Eigen::Isometry3d& pose, const kinematics::KinematicsBaseConstPtr& solver);
+
   /**
    * \brief Transform pose from the robot model's base frame to the reference frame of the IK solver
    * @param pose - the input to change
    * @param ik_frame - the name of frame of reference of base of ik solver
    * @return true if no error
    */
-  bool setToIKSolverFrame(Eigen::Isometry3d& pose, const std::string& ik_frame);
+  inline bool setToIKSolverFrame(Eigen::Isometry3d& pose, const std::string& ik_frame);
```
rhaschke, line 1758: This one shouldn't be inline. It's a longer function. However, the other one is just forwarding to this one here.
```suggestion
  bool setToIKSolverFrame(Eigen::Isometry3d& pose, const std::string& ik_frame);
```

rhaschke (approved):
One additional change. Otherwise, I approve.

PR #2463, opened 2020-12-18
Servo: add an important sleep()

AndyZe:
Forgot to put a sleep in the loop. The loop was running at 2kHz before this  :O

@Abishalini -- mentioned since you're porting this to ROS2


jliukkonen (approved):
Seems to have done the trick (tested on hardware). :+1:

tylerjw (approved):

PR #2468, opened 2020-12-25
Catch exceptions during RobotModel loading and display them in rviz

### LINKED ISSUE #2238: Crash on negative sizes in URDF
### Description

Somewhat recently (https://github.com/ros-planning/geometric_shapes/commit/07a5254b698ecc89e511dc5b3182ecb6c7d2258e) a check was introduced whether `Shapes` have negative lengths. This crashes whoever displays that URDF (`RobotStateDisplay`)

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Version 1.0.5

### Steps to reproduce
modify some URDF to have a eg cylinder with negative length

### Expected behaviour
Display a warning in the side pane. I guess. For experienced user I guess a crash is a fine way to notify them but it also brings down our URDF editor which is used by novice people who might be frustrated then. I'll therefore add a try catch in our wrapper of `RobotStateDisplay` but was wondering if I should also open a PR

### Actual behaviour
it crashes

simonschmeisser:
### Description

This fixes #2238 : 
> Somewhat recently (https://github.com/ros-planning/geometric_shapes/commit/07a5254b698ecc89e511dc5b3182ecb6c7d2258e) a check was introduced whether `Shapes` have negative lengths. This crashes whoever displays that URDF (`RobotStateDisplay`)

grafik
You can see the error message from the exception and also that negative dimensions are not checked in the code path for `rviz::Robot` and Ogre actually handles them quite fine (in contrast to fcl so it's good that we do have those checks).

### Checklist

### DIFF
--- a/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
+++ b/moveit_ros/visualization/robot_state_rviz_plugin/src/robot_state_display.cpp
@@ -375,20 +375,27 @@ void RobotStateDisplay::loadRobotModel()
 
   if (rdf_loader_->getURDF())
   {
-    const srdf::ModelSharedPtr& srdf =
-        rdf_loader_->getSRDF() ? rdf_loader_->getSRDF() : srdf::ModelSharedPtr(new srdf::Model());
-    robot_model_.reset(new moveit::core::RobotModel(rdf_loader_->getURDF(), srdf));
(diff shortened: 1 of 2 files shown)


felixvd (approved):
Haven't tested, but it looks good to me.

rhaschke (approved):

PR #2469, opened 2020-12-30
WIP: standalone clang tidy script

tylerjw:
### Description

This is a standalone adaption of the script in moveit_ci for running clang-tidy against a specific branch testing packages with changed source files since that branch.  One feature this does not support yet is running clang-tidy against the full repo.  However, that is much simpler (can be done with a one-liner) and there are instructions on how to do that on our website.

It might be nice to write this in python or maybe even extend ament_clang_tidy to support the feature of only testing packages that have changed.  I started down that path and then realized that there was no simple way to use ament in a ros1 environment.

To test this simply run it from the workspace directory after building with compile commands exported.

The goal of this script is an easy way for people to locally test and fix clang-tidy issues with their changes using the same script used by CI to test their changes.

### DIFF
--- a/tools/clang_tidy.sh
+++ b/tools/clang_tidy.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+
+# Software License Agreement - BSD 3-Clause License
+#
+# Author:  Robert Haschke and Tyler Weaver
+#
+# Adapted from script in moveit_ci
+# https://github.com/ros-planning/moveit_ci/blob/master/check_clang_tidy.sh
+#
+# Intended to be run from workspace directory.  Takes one argument, the base
+# git tag or branch to compare against for finding file changes.  Leaving
+# that argument unchanged defaults it to master.
+
+WS_DIR=$PWD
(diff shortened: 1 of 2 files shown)


tools/clang_tidy.sh
```diff
@@ -0,0 +1,83 @@
...
+            fi
+        else
+            echo $(colorize YELLOW "$file not found, skipping")
+        fi
+    done
+
+    pushd $REPO_DIR > /dev/null
+    # if there are workspace changes, print broken pkg to file descriptor 3
+    _have_fixes && 1>&3 ezcho $pkg || true
+    popd > /dev/null
```
tylerjw, line 53: Because I wanted people to be able to run this from a directory with changes and not reset those changes I removed the checkout line in _have_fixes.  As a result of doing that I moved this out of the loop.  The problem is now that $pkg variable being sent to fd3 isn't set and is meaningless.

I need to figure out how to get the packages that have diffs, or maybe just the list of files with diffs to make this work again.

tylerjw:
Closing in favor of python version.

PR #2470, opened 2020-12-30
instructions for running ament linters and repos file

tylerjw:
## Update

I've updated this to be just the instructions I use locally.  lmk if anyone things it should be somewhere else?  Maybe we should just put these instructions on the moveit website.  However, having this repos file here makes it easier for using it in ci systems and docker images.

### Description

Here is a standalone script for running clang-tidy with the time-saving features from moveit_ci.  It is a modified version of the ament_clang_tidy script you can find here: https://github.com/ament/ament_lint/blob/master/ament_clang_tidy/ament_clang_tidy/main.py
 
Because ament_lint is not easily installable for ros1 projects I suggest we maintain this fork of the clang-tidy script.

This is intended to be used from the workspace directory after you have built with compile commands exported (`catkin config --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=ON`).  Here is an example of how you would run this to test source files that have changed on your branch since master:
```bash
python3 src/moveit/tools/ament_clang_tidy.py --filter-repo-path src/moveit --filter-diff master
```

Here is an example of how to use it to test all the packages in the moveit repo:
```bash
python3 src/moveit/tools/ament_clang_tidy.py --filter-repo-path src/moveit 
```

I also added a verbose option that prints out each command as it is called.  This could be useful for debugging if you this script is not calling the commands you expect.

### Upstream

Here is my PR for committing this change upstream: https://github.com/ament/ament_lint/pull/280  It might be easier to review this by looking at what I changed from the ament_clang_tidy script itself.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,0 +1,58 @@
+# MoveIt tools
+
+This documents tools for local lint testing and CI.
+
+## Build and install for local use
+
+Create workspace to build the ament packages.
+
+    mkdir -p /opt/ws_ament_lint/src
+    cd /opt/ws_ament_lint
+    vcs import src --input https://raw.githubusercontent.com/tylerjw/moveit/clang_tidy_ament_script/tools/ament_lint.repos
+
+Build the workspace
+
+    colcon build
+
+Add this line to your `.bashrc`:
+
+    source /opt/ws_ament_lint/install/setup.bash
+
+## Run ament_clang_format
+
+Run this command from the root of the repo to check if your code is formatted correctly.  Make sure you have installed clang-format-10.
+
+    ament_clang_format --config ./.clang-format --clang-format-version 10
+
+To have it reformat your code inplace use the `--reformat` flag.
+
+## Run ament_clang_tidy
+
+`clang-tidy` is a static analysis linter that takes a long time to execute.  As a result we limit testing to packages that have changed in any given PR.  To run a similar test locally you need to first build your project with this option:
+
+    --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+
+Then run `ament_clang_tidy` from the workspace directory like what follows.
+
+    ament_clang_tidy --filter-repo-path src/moveit --filter-diff master
+
+To have it attempt to fix the code in place you can use the option `--reformat`.
+
+### Background
+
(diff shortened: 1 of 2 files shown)


henningkayser:
I'm not fully convinced we should add this in here. How about just cloning the (forked) repo in moveit_ci or installing the official package in docker once it's synced?

tylerjw:
I do agree that we should instead depend on upstream if possible.

tylerjw:
Repeating some of the conversation in the upstream PR: https://github.com/ament/ament_lint/pull/280

@tfoote doesn't think this is the correct approach.  He suggests we do this as a layer above colcon (or catkin presumably) as that layer could have all the information to do this.  However, I did not originally take that aproach because of these two reasons:

* I wanted a standalone script we could use (much like what ament_clang_tidy is) to test locally and make the step of running that test in CI separate, yet simple to invoke.
* I don't know how to add the cmake magic to have a ros1 project use ament_cmake features (adding ament_clang_tidy with arguments to ctest tests) without having to build the ament packages from source as a dependency.  This is not something I think we want for moveit (source dependency on ament for building/testing).

In the short term I think the best solution (that moves the goalpost towards having better ci for moveit) is to use my fork of ament_lint for ci and for me to create a readme on how to build and run those tests locally.  There are several really nice parts of the ament_lint tests I think makes this worth it:

* They generate xunit output (circle ci parses these as test output and will make easier to understand test failures)
* They are standalone tests you can use in any of your projects to do things like clang-tidy, clang-format, etc.
* The process of building them and adding them to your enviroment is fairly straightforward.

To this end I'll update this PR to instead include a README.md and a repos file.  Please let me know if anyone has a better idea for how to do this.  Maybe this README.md and repos file should go in a separate repo as it is more generic than moveit itself and could be used for other projects.  I propose we put it in moveit and migrate it out later if that becomes true, at this point it is basically just some documentation on how to use my fork of ament_lint.

tylerjw:
## Update

I've updated this to be just the instructions I use locally.  lmk if anyone things it should be somewhere else?  Maybe we should just put these instructions on the moveit website.  However, having this repos file here makes it easier for using it in ci systems and docker images.

tylerjw:
Closing this as it is no longer relevant.

Issue #2096, opened 2020-05-19
[jog_arm] cartesian jog one linear axis adds rotation as well


rfeistenauer:
### Description

By Following the [tutorial](https://github.com/ros-planning/moveit/blob/master/moveit_experimental/moveit_jog_arm/README.md) of jog_arm with current moveit master branch (ad88e18ae604e0533) and then publishing twists with 1 and -1 in y axis, the robot gets high angular twists as well.
Depending on the chosen robot (ur like in tutorial or prbt) there are different rotations in size and magnitude.

for example, achieved with jogging only linear y axis!
![Screenshot from 2020-05-19 14-13-45](https://user-images.githubusercontent.com/45563842/82325208-1bb5ac80-99db-11ea-808b-54f8486c0f16.png)

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Source build
* ad88e18ae604e0533

### Steps to reproduce
Run the steps of the [ReadMe](https://github.com/ros-planning/moveit/blob/master/moveit_experimental/moveit_jog_arm/README.md)

Then for example use a script like the following
```
    delta_jog_cmds_publisher = rospy.Publisher("jog_server/delta_jog_cmds", TwistStamped, queue_size=10)
    rospy.sleep(1)
    for i in range(10):
        for p in range(100):
            delta_jog_cmds_publisher.publish(TwistStamped(header=Header(stamp=rospy.Time.now()),
                                                          twist=Twist(linear=Vector3(y=1))))
            rospy.sleep(.01)
        for m in range(100):
            delta_jog_cmds_publisher.publish(TwistStamped(header=Header(stamp=rospy.Time.now()),
                                                          twist=Twist(linear=Vector3(y=-1))))
            rospy.sleep(.01)
```

### Expected behaviour
robot moves on y axis and ends somewhere near the starting position, without rotating the tcp

### Actual behaviour
unexpected rotations and movements increasing over the time

### Comment
Its hard to believe, a bug like this has gone unnoticed, so if you can not reproduce this, please tell me. I tested it on a second PC with newly setup environment, still maybe there is something that i missed here.

AndyZe:
I don't see this behavior  :/

You haven't changed anything in the yaml file, right?

Can you try publishing via command line instead of with your Python script?

> rostopic pub -r 100 /jog_server/delta_jog_cmds geometry_msgs/TwistStamped "header: auto
> twist:
>   linear:
>     x: 0.0
[...]

PR #2471, opened 2021-01-05
Refactor Servo velocity bounds enforcement

jliukkonen:
The `enforceVelLimits` function struck me as bit odd so I figured I make this PR to start a conversation about what I think it is doing and I why I think it looks like it might not be correct.

Prior to my changes the function would calculate joint angle velocity for each joint based on the delta theta (joint angle change) and would determined the smallest scaling factor for the velocities based on the biggest out-of-bounds velocity value. So if one would have joint velocities of, say, `[1, 1, 1, 2, 3, 100]`, and each joint was bounded to `[-1, 1]` range, then the scaling factor would be `0.01` . That's at least how I interpreted it. This scaling factor was then applied to the delta thetas meaning if one joint received a command that would translate into very large velocity, the other joints would slow to a crawl.

I felt it would make more sense and be more correct to just apply the velocity limits as they are used elsewhere in MoveIt, that is, make them upper and lower bounds for the joint velocities by clamping the velocities to `[min_vel, max_vel]` range.

One thing I don't currently understand is how the delta thetas and joint bounds are matched together. Is it guaranteed that the bounds are in correct order with respect to the `delta_theta` vector and just applying them in order is enough?

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -758,58 +758,23 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
 
 void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
 {
+  // Convert to joint angle velocities for checking and applying joint specific velocity limits.
   Eigen::ArrayXd velocity = delta_theta / parameters_.publish_period;
 
   std::size_t joint_delta_index = 0;
-  // Track the smallest velocity scaling factor required, across all joints
-  double velocity_limit_scaling_factor = 1;
-
-  for (auto joint : joint_model_group_->getActiveJointModels())
+  for (const moveit::core::JointModel* joint : joint_model_group_->getActiveJointModels())
   {
-    // Some joints do not have bounds defined
-    const auto bounds = joint->getVariableBounds(joint->getName());
-
+    const auto& bounds = joint->getVariableBounds(joint->getName());
     if (bounds.velocity_bounded_)
     {
-      velocity(joint_delta_index) = delta_theta(joint_delta_index) / parameters_.publish_period;
-
-      bool clip_velocity = false;
-      double velocity_limit = 0.0;
-      if (velocity(joint_delta_index) < bounds.min_velocity_)
-      {
-        clip_velocity = true;
-        velocity_limit = bounds.min_velocity_;
-      }
-      else if (velocity(joint_delta_index) > bounds.max_velocity_)
-      {
-        clip_velocity = true;
-        velocity_limit = bounds.max_velocity_;
-      }
-
-      // Apply velocity bounds
-      if (clip_velocity)
-      {
-        const double scaling_factor =
-            fabs(velocity_limit * parameters_.publish_period) / fabs(delta_theta(joint_delta_index));
-
-        // Store the scaling factor if it's the smallest yet
-        if (scaling_factor < velocity_limit_scaling_factor)
-          velocity_limit_scaling_factor = scaling_factor;
-      }
-    }
-    ++joint_delta_index;
-  }
-
-  // Apply the velocity scaling to all joints
-  if (velocity_limit_scaling_factor < 1)
-  {
-    for (joint_delta_index = 0; joint_delta_index < joint_model_group_->getActiveJointModels().size();
-         ++joint_delta_index)
-    {
-      delta_theta(joint_delta_index) = velocity_limit_scaling_factor * delta_theta(joint_delta_index);
-      velocity(joint_delta_index) = velocity_limit_scaling_factor * velocity(joint_delta_index);
+      velocity(joint_delta_index) =
+          std::min(std::max(velocity(joint_delta_index), bounds.min_velocity_), bounds.max_velocity_);
+      ++joint_delta_index;
     }
   }
+
+  // Convert back to joint angle increments.
+  delta_theta = velocity * parameters_.publish_period;
 }
 
 bool ServoCalcs::enforcePositionLimits()
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -762,19 +762,21 @@ void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
   Eigen::ArrayXd velocity = delta_theta / parameters_.publish_period;
 
   std::size_t joint_delta_index = 0;
+  double velocity_scaling_factor = 1;
   for (const moveit::core::JointModel* joint : joint_model_group_->getActiveJointModels())
   {
     const auto& bounds = joint->getVariableBounds(joint->getName());
     if (bounds.velocity_bounded_)
     {
-      velocity(joint_delta_index) =
-          std::min(std::max(velocity(joint_delta_index), bounds.min_velocity_), bounds.max_velocity_);
+      const auto unbounded_velocity = velocity(joint_delta_index);
+      velocity(joint_delta_index) = std::min(std::max(unbounded_velocity, bounds.min_velocity_), bounds.max_velocity_);
+      velocity_scaling_factor = std::min(velocity_scaling_factor, unbounded_velocity / velocity(joint_delta_index));
       ++joint_delta_index;
     }
   }
 
   // Convert back to joint angle increments.
-  delta_theta = velocity * parameters_.publish_period;
+  delta_theta = velocity_scaling_factor * velocity * parameters_.publish_period;
 }
 
 bool ServoCalcs::enforcePositionLimits()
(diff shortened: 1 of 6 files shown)


AndyZe:
The reason it scales velocities down uniformly is so the robot follows a straight line. Thus, I'm not a huge fan of this change.

If you can show it's really important from a performance standpoint, we could your new method optional.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -758,58 +758,25 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
...
-  double velocity_limit_scaling_factor = 1;
-
-  for (auto joint : joint_model_group_->getActiveJointModels())
+  double velocity_scaling_factor = 1;
+  for (const moveit::core::JointModel* joint : joint_model_group_->getActiveJointModels())
   {
-    // Some joints do not have bounds defined
-    const auto bounds = joint->getVariableBounds(joint->getName());
-
+    const auto& bounds = joint->getVariableBounds(joint->getName());
```
AndyZe, line 768: I appreciate this little optimization of auto**&**

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -758,58 +758,25 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
...
-            fabs(velocity_limit * parameters_.publish_period) / fabs(delta_theta(joint_delta_index));
-
-        // Store the scaling factor if it's the smallest yet
-        if (scaling_factor < velocity_limit_scaling_factor)
-          velocity_limit_scaling_factor = scaling_factor;
-      }
+      const auto unbounded_velocity = velocity(joint_delta_index);
+      velocity(joint_delta_index) = std::min(std::max(unbounded_velocity, bounds.min_velocity_), bounds.max_velocity_);
+      velocity_scaling_factor = std::min(velocity_scaling_factor, unbounded_velocity / velocity(joint_delta_index));
+      ++joint_delta_index;
```
AndyZe, line 774: it seems like there is potential here (`if (!bounds.velocity_bounded_)`) that joint_delta_index doesn't get incremented properly

AndyZe: should it be moved outside of the if()?

jliukkonen-pn: Good catch. Would have definitely been a bug and I'll need to fix it. But this comes back to my question in the description about how are the joint bounds and `delta_theta`s guaranteed to match? Is it simply because the `joint_model_group_` is initialized once and all its class methods access the joint in the same order? I marked this as WIP because I want to convince myself that the indices and deltas do indeed match but I haven't yet had the time.

AndyZe:
@jliukkonen I hope you can finish this PR! I know you're pretty busy but it's a nice simplification.

jschleicher:
> The reason it scales velocities down uniformly is so the robot follows a straight line. Thus, I'm not a huge fan of this change.

Definitely! According to our previous tests it is essential, that you keep the direction, when scaling the velocity vector. Nevertheless I suspect, there is some problem hidden in `enforceVelLimits()` leading to artifacts as described in #2096.

AndyZe (approved):
I added some commits for unit testing. Looks good to me, now. Thanks for a nice simplification.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -758,58 +758,25 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
...
-        const double scaling_factor =
-            fabs(velocity_limit * parameters_.publish_period) / fabs(delta_theta(joint_delta_index));
-
-        // Store the scaling factor if it's the smallest yet
-        if (scaling_factor < velocity_limit_scaling_factor)
-          velocity_limit_scaling_factor = scaling_factor;
-      }
+      const auto unbounded_velocity = velocity(joint_delta_index);
+      velocity(joint_delta_index) = std::min(std::max(unbounded_velocity, bounds.min_velocity_), bounds.max_velocity_);
+      velocity_scaling_factor = std::min(velocity_scaling_factor, unbounded_velocity / velocity(joint_delta_index));
```
jliukkonen, line 773: Commenting on my own mistakes here, but I think I got at least this wrong. The line above (772), first clamps the velocities to a range of `[min_velocity, max_velocity]`. That is correct. However, when calculating the scaling factor, the formula is definitely wrong and should be `velocity(joint_delta_index) / unbounded_velocity` instead.

Now, even with that fix, it seems to be possible to construct a situation where the `velocity_scaling_factor` ends up being nonsensical. For instance: if we have a set of joint velocities `[-1, -3, 2, 3, 4, 5]` and `max_velocity` of 2 and `min_velocity` of 1, then we would first clamp the velocities down to a set of `[1, 1, 2, 2, 2, 2]` and then get the minimum scaling factor of `-1` which would switch the directions the joints are moving. Now, I'm not entirely sure if this kind of scenario is possible with the way velocity limits are constructed. @AndyZe is it possible to have both the max and min velocities to be positive or are they guaranteed to always have opposing signs?

AndyZe: I think there is only a "velocity_limit" in the URDF. It is always a positive value. Hence, min and max are symmetric. Will try to find a reference to confirm.

AndyZe: This [page](http://wiki.ros.org/urdf/XML/joint) and this [page](http://wiki.ros.org/pr2_controller_manager/safety_limits) confirm that the velocity limit from the URDF is always given as a positive value. So the limits are (-velocity_limit, +velocity_limit).

jliukkonen:
Thanks for fixing some of the bugs and adding unit tests @AndyZe! I think I managed to nail all the remaining issues. I added some comments to the code, please check them out when re-reviewing.

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -758,58 +758,26 @@ double ServoCalcs::velocityScalingFactorForSingularity(const Eigen::VectorXd& co
...
-        const double scaling_factor =
-            fabs(velocity_limit * parameters_.publish_period) / fabs(delta_theta(joint_delta_index));
-
-        // Store the scaling factor if it's the smallest yet
-        if (scaling_factor < velocity_limit_scaling_factor)
-          velocity_limit_scaling_factor = scaling_factor;
-      }
+      const double unbounded_velocity = velocity(joint_delta_index);
+      // Clamp each joint velocity to a joint specific [min_velocity, max_velocity] range.
+      const auto bounded_velocity = std::min(std::max(unbounded_velocity, bounds.min_velocity_), bounds.max_velocity_);
```
jliukkonen, line 773: This was wrong in my previous version. The velocities were bounded _before_ applying velocity scaling factor so velocity was scaled down twice. First by applying velocity bounds and then by applying scaling factor to the already bounded velocities. Now the velocity bounds are only used to determine the scaling factor.

moveit_ros/moveit_servo/test/basic_servo_tests.cpp
```diff
@@ -170,6 +174,71 @@ TEST_F(ServoFixture, SendJointServoTest)
...
+  delta_theta[4] = 0.04;
+  delta_theta[5] = 0.05;
+  delta_theta[6] = 0.06;
+
+  // Store the original joint commands for comparison before applying velocity scaling.
+  Eigen::ArrayXd orig_delta_theta = delta_theta;
+  enforceVelLimits(delta_theta);
+
+  // From Panda arm MoveIt joint_limits.yaml. The largest velocity limits for a joint.
+  const double panda_max_joint_vel = 2.610;  // rad/s
```
jliukkonen, line 199: Since `joint_limits.yaml` take precedence over the URDF the value is actually `2.61` and not `2.871`. Took me a while to understand that that was the source of errors. :)

AndyZe: good catch there

moveit_ros/moveit_servo/test/basic_servo_tests.cpp
```diff
@@ -170,6 +174,71 @@ TEST_F(ServoFixture, SendJointServoTest)
...
 }
+
+// This a friend test of a private member function
+TEST_F(ServoFixture, EnforceVelLimitsTest)
+{
+  auto parameters = servo_->getParameters();
+  const double publish_period = parameters.publish_period;
+
+  // Request joint angle changes that are too fast, given the control period in servo settings YAML file.
+  Eigen::ArrayXd delta_theta(7);
```
AndyZe, line 185: i see you made this a 7-vector because the joint group has 7 joints. Good catch.

AndyZe (approved):

AndyZe:
Small numerical errors are throwing off the unit test, for example:

`Expected equality of these values:&#x0A;  orig_delta_theta(i)&#x0A;    Which is: -0.012999999999999999&#x0A;  delta_theta(i)&#x0A;    Which is: -0.012293478260869564"`

Prob need to use a combination of `EXPECT_NEAR (expected, actual, absolute_range)` and `EXPECT_LT()`

AndyZe:
I'll push a commit to do that with, say, `1e-3` tolderance

henningkayser (approved):
This looks much cleaner than the previous implementation.

PR #2474, opened 2021-01-06
Fix OpenMP compilation on OSX with new version of clang

Tobias-Fischer:
### Description

This PR fixes compilation on OSX with Clang - see See https://stackoverflow.com/questions/46414660/macos-cmake-and-openmp

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/CMakeLists.txt
+++ b/moveit_planners/ompl/ompl_interface/CMakeLists.txt
@@ -21,6 +21,24 @@ add_library(${MOVEIT_LIB_NAME}
 )
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 
+if(APPLE AND CMAKE_C_COMPILER_ID MATCHES "Clang")
+  set(OpenMP_C "${CMAKE_C_COMPILER}")
+  set(OpenMP_C_FLAGS "-fopenmp=libomp -Wno-unused-command-line-argument")
+  set(OpenMP_C_LIB_NAMES "libomp" "libgomp" "libiomp5")
+  set(OpenMP_libomp_LIBRARY ${OpenMP_C_LIB_NAMES})
+  set(OpenMP_libgomp_LIBRARY ${OpenMP_C_LIB_NAMES})
+  set(OpenMP_libiomp5_LIBRARY ${OpenMP_C_LIB_NAMES})
+endif()
+
+if(APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  set(OpenMP_CXX "${CMAKE_CXX_COMPILER}")
+  set(OpenMP_CXX_FLAGS "-fopenmp=libomp -Wno-unused-command-line-argument")
+  set(OpenMP_CXX_LIB_NAMES "libomp" "libgomp" "libiomp5")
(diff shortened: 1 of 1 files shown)


tylerjw (approved):
Looks good to me but I don't have a Mac.  If someone with one could test this that would be apreciated.

Tobias-Fischer:
We use this in https://github.com/RoboStack/ros-noetic where compilation without this patch fails on OSX, but with the patch it compiles fine.

rhaschke (changes requested):
Sorry, but this isn't convincing to me. All these settings should be handled by `find_package(OpenMP)`. If there is a bug in OSX's `FindOpenMP.cmake`, it should be fixed there instead of providing a laborious workaround here.
Also, most of the set variables are not used by MoveIt. So why setting them at all?
Could you please report the actual error message(s) as well as the OpenMP variables set in `CMakeCache.txt`?

Tobias-Fischer:
Hello, I agree that the proper place would be fixing cmake, but years will pass before that fixed version becomes available so it can be used and set as minimum version here.

The error and fix is described there: https://stackoverflow.com/questions/46414660/macos-cmake-and-openmp

We observed the error in our efforts for bringing ROS in conda: https://github.com/RoboStack/ros-noetic
There we have a CI for Linux, Win and OSX. Patching MoveIt with the contents of this PR builds successfully, while not applying the PR leads to the error above.

rhaschke:
I looked into the stackoverflow link before and the solution you picked is one possible workaround discussed there. But, as I said, this solution is not very convincing to me and it clutters our codebase. There are also comments on stackoverflow stating: 

> CMake can detect the correct OpenMP options for Apple Clang and standard Clang.

So, please, provide the exact error message, so that we can really help.
In any case, you should (also) file an issue / PR to fix `FindOpenMP.cmake` upstream - even if that's only a solution in the long run.

Tobias-Fischer:
The exact error message is: 
```
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) (found version "1.0")
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) (found version "1.0")
```

I can understand that you're worried re code cluttering. Please feel free to either leave a stale PR here or close the PR - I guess people who run into this problem will find the PR and can apply the patch manually. We manually fix it in the Robostack and hopefully this will be the main way of using ROS/MoveIt on OSX soon, there users won't have to worry.

rhaschke:
The underlying reason of the failure on MacOS seems to be this: http://stechschulte.net/2016/03/20/openmp-osx-cmake.html.
Please, try to use clang-omp as your compiler as suggested there.
In any case, manually setting all the OpenMP variables defeats the purpose of `find_package`! Thus, I'm still not supporting this PR.

mamoll:
The failure to correctly detect OpenMP appears to be a Home Brew failure, not a macOS failure. The same `find_package(OpenMP)` seems to work just fine with clang installed via MacPorts:
```
env CXX=/opt/local/bin/clang++-mp-11 cmake ...
...
-- Found OpenMP_CXX: -fopenmp=libomp (found version "5.0") 
-- Found OpenMP: TRUE (found version "5.0")  
...
```

rhaschke:
So, can we close this PR?

v4hn:
Close due to inactivity.

PR #2481, opened 2021-01-10
Lock the Bullet collision environment, for thread safety

AndyZe:
Lock the collision checking methods of Bullet, for thread safety.

The main problem was updating the scene with a callback while doing a collision check at the same time.

This PR locks all three possible types of collision check:  self-collision, environment collision, and continuous collision detection.

### DIFF
--- a/moveit_core/collision_detection_bullet/README.md
+++ b/moveit_core/collision_detection_bullet/README.md
@@ -5,4 +5,4 @@ To use Bullet as a collision checker you can manually switch to Bullet using the
 For speed comparisons to FCL please see (relative link to README of benchmark script will be added).
 
 ### Current Limitations
-The collision detection using Bullet is not thread safe as the internal collision managers are `mutable` members of the collision environment.
+Collision detection using Bullet is thread safe but parallelization does not improve performance.
--- a/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_env_bullet.h
+++ b/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_env_bullet.h
@@ -39,6 +39,7 @@
 #include <moveit/collision_detection/collision_env.h>
 #include <moveit/collision_detection_bullet/bullet_integration/bullet_discrete_bvh_manager.h>
 #include <moveit/collision_detection_bullet/bullet_integration/bullet_cast_bvh_manager.h>
+#include <mutex>
 
 namespace collision_detection
 {
@@ -113,12 +114,12 @@ protected:
   void addLinkAsCollisionObject(const urdf::LinkSharedPtr& link);
 
   /** \brief Handles self collision checks */
-  mutable collision_detection_bullet::BulletDiscreteBVHManagerPtr manager_{
+  collision_detection_bullet::BulletDiscreteBVHManagerPtr manager_{
     new collision_detection_bullet::BulletDiscreteBVHManager()
   };
 
   /** \brief Handles continuous robot world collision checks */
-  mutable collision_detection_bullet::BulletCastBVHManagerPtr manager_CCD_{
+  collision_detection_bullet::BulletCastBVHManagerPtr manager_CCD_{
     new collision_detection_bullet::BulletCastBVHManager()
   };
 
@@ -140,5 +141,8 @@ private:
   void notifyObjectChange(const ObjectConstPtr& obj, World::Action action);
 
   World::ObserverHandle observer_handle_;
+
+  // Lock the collision environment, for thread safety
+  mutable std::mutex collision_env_mutex_;
 };
 }  // namespace collision_detection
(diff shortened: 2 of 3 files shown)


moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
```diff
@@ -186,6 +190,12 @@ void CollisionEnvBullet::checkRobotCollisionHelper(const CollisionRequest& req,
         cow->getName(), state.getAttachedBody(cow->getName())->getGlobalCollisionBodyTransforms()[0]);
   }
 
+  // updating link positions with the current robot state
```
AndyZe, line 193: I need to do some more testing to see whether this is necessary or not.

AndyZe: It seems like this has no effect. Reverted it.

AndyZe:
I modified the tutorial to test multi-threaded Bullet collision tests [here](https://github.com/AndyZe/moveit_tutorials/tree/andyz/bullet_link_distance). It still crashes   :(

AndyZe:
This is working now in our multi-threaded application. Removing the WIP label.

v4hn:
@j-petit could you please comment on this?
I remember we had discussions about this during the GSoC project.

j-petit:
Yes, I am a bit surprised that removing `mutable` works. I [thought that it is necessary](https://github.com/ros-planning/moveit/issues/1641) because the manager is [manipulated in the collision checking process](https://github.com/ros-planning/moveit/blob/bd1a0eeafff5dc04f3f2812f3500776fed986933/moveit_core/collision_detection_bullet/src/bullet_integration/bullet_discrete_bvh_manager.cpp#L81). But CI and Andy's tests indicate that everything is working, so dropping `mutable` is good. And adding the locks for multi-threaded checking is even better! See also [here](https://github.com/ros-planning/moveit/pull/1657#issuecomment-532725798).

@AndyZe , you might want to change [the readme](https://github.com/ros-planning/moveit/tree/bd1a0eeafff5dc04f3f2812f3500776fed986933/moveit_core/collision_detection_bullet#current-limitations) to indicate that it is now thread safe.

j-petit:
The collision manager is added to the environment via a pointer, therefore `const` does not affect it. LGTM, if the readme is adapted.

moveit_core/collision_detection_bullet/README.md
```diff
@@ -3,6 +3,3 @@ To use Bullet as a collision checker you can manually switch to Bullet using the
 
 ### Speed Benchmarks
 For speed comparisons to FCL please see (relative link to README of benchmark script will be added).
-
-### Current Limitations
-The collision detection using Bullet is not thread safe as the internal collision managers are `mutable` members of the collision environment.
```
v4hn, line 8: Instead of removing this section, could you please change the sentence to say that the bullet collision environment does not profit from parallel collision checking. The mutex gets rid of undefined behavior in parallel use, but it does not magically give you a speedup if you *do* parallelize.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_env_bullet.h
```diff
@@ -140,5 +141,8 @@ class CollisionEnvBullet : public CollisionEnv
   void notifyObjectChange(const ObjectConstPtr& obj, World::Action action);
 
   World::ObserverHandle observer_handle_;
+
+  // Lock the collision environment, for thread safety
+  mutable std::mutex collision_env_mutex_;
```
v4hn, line 146: Does this lock need to protect anything beside `manager_` / `manager_CCD_`?
If these are the only things that need to be locked, please say so in this comment.
Locking the entire class does not make a huge difference I guess, but is imprecise.

v4hn (commented):
I discussed with @henningkayser recently and we were unsure about whether we have to force this to be thread-safe, but [Herb Sutter convinced me](https://channel9.msdn.com/posts/C-and-Beyond-2012-Herb-Sutter-You-dont-know-blank-and-blank). People (including me) just *expect* const methods to be thread-safe.

Assuming the mutex only protects the managers you could probably move some locks a few lines down to block less time, but the change in lock-time would probably be negligible.

Instead of removing the documentation, please modify it as mentioned below.

v4hn:
Any chance you can fixup & merge this @AndyZe ?

AndyZe:
> Any chance you can fixup & merge this @AndyZe ?

I will get back to it soon!  (It has been awhile...)

moveit_core/collision_detection_bullet/README.md
AndyZe: done

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/collision_env_bullet.h
AndyZe: Unfortunately, whatever caused the crash in our application does not occur anymore so this is hard for me to test.

I don't know the Bullet setup well enough to guess if it is necessary to protect anything else, so I guess we should leave it locking the whole class.

v4hn: Of course locking only the `contactTest` will fail, because the methods also call `manager_->add/removeCollisionObject`, among others.
Did you include these calls in the same lock?

My message above was mainly about the comment.
There is only one other local class member (`observer_handle_`) and it's not locked by your request.
I'll push a commit to change the comment accordingly.

PR #2483, opened 2021-01-12
pilz planner: add string includes

v4hn:
On my (non-Ubuntu) system I get error messages about `std::string` being undefined in these two places.
Not sure why this does not happen to everyone, but maybe some upstream header changed its includes recently.


rhaschke (approved):

jschleicher:
Thank you! 👍

PR #2484, opened 2021-01-12
Update move group tests

bostoncleek:
The current move group interface c++ tests will sometimes fail. The failure could be a result of a long sequence of planning steps (~10). I combined a pose goal, path (orientation) constraint, and collision detection into a single test. The success of this test should be more reliable and it also tests multiple move group interface functionalities. The reliability comes from the fact that one planning problem is being solved and that it is not overly complex. In terms of complexity the collision object is not blocking the arm's shortest path and the test does not require the arm to contort into an extreme position.  The orientation constraint keeps the gripper facing up as the arm moves along the x-axis while avoiding a box.

Bellow is the output of the test result. @tylerjw 

test1

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
+++ b/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp
@@ -33,7 +33,7 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  *********************************************************************/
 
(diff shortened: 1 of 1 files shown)


tylerjw (approved):
This looks really good, thank you for doing this.  Hopefully, this will make this a much more reliable test on Travis.

JafarAbdi (approved):

PR #2487, opened 2021-01-14
Adding a missing dependency declaration on joint_limits_interface

ct2034:
Adding a missing dependency declaration on joint_limits_interface

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
+++ b/moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
@@ -9,6 +9,7 @@ add_definitions(-Wno-unused-parameter)
 
 find_package(catkin REQUIRED COMPONENTS
   eigen_conversions
+  joint_limits_interface
   kdl_conversions
   moveit_core
   moveit_msgs
--- a/moveit_planners/pilz_industrial_motion_planner/package.xml
+++ b/moveit_planners/pilz_industrial_motion_planner/package.xml
@@ -19,6 +19,7 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <depend>eigen_conversions</depend>
+  <depend>joint_limits_interface</depend>
   <depend>kdl_conversions</depend>
   <depend>moveit_ros_planning_interface</depend>
   <depend>moveit_msgs</depend>


gavanderhoorn:
The code seems to be using it, but why does a planner have a direct dependency on something from `ros_control` which is typically only used by `hardware_interface` implementations?

martiniil:
> 
> 
> The code seems to be using it, but why does a planner have a direct dependency on something from `ros_control` which is typically only used by `hardware_interface` implementations?

It allows us to represent joint limits and populate them from the parameter server.

jschleicher:
@gavanderhoorn What would you suggest? We need the limits for time parametrization of the trajectory. Is there a `URDF::JointLimits` class or something more appropriate here?

gavanderhoorn:
We have [RobotModel::getVariableBounds(..)](https://github.com/ros-planning/moveit/blob/023b11def6329b165ed7509e5de9aec4c1e29c6c/moveit_core/robot_model/include/moveit/robot_model/robot_model.h#L408) fi, which you already appear to be using ([here](https://github.com/ros-planning/moveit/blob/cf218879dbc23aadf88dd56b8abe7970b7d61030/moveit_planners/pilz_industrial_motion_planner/src/joint_limits_aggregator.cpp#L104) fi).

My comment is not an "official review". I just happened to notice that we're adding a direct dependency on `ros_control` to a (core) planner of MoveIt, while MoveIt itself also loads joint limits. Using parallel functionality from another package seemed strange to me, hence the question.

ct2034:
Thanks for the feedback @gavanderhoorn . Would you still agree to merge it like this? I agree that a solution with RobotModel::getVariableBounds(..) would be neater but I would prefer to fix that separately and put it onto our backlog.

simonschmeisser (approved):
I agree with @gavanderhoorn that this dependency should be killed asap but it should also not be a hidden surprise

AndyZe:
^ i agree with that logic

AndyZe (approved):

AndyZe:
An issue should probably be created to remove the dependency.

rhaschke (approved):
Approving and merging. @jschleicher: Could you please work on a PR to replace this interface and use the existing RobotModel interface instead?

jschleicher:
Thanks for merging. Yes, I've put it on our backlog.

PR #2488, opened 2021-01-14
Update subframes test

bostoncleek:
Sometime the SubframesTests will fail. Planning on this particular problem
takes ~10s and the allotted planning time for move_group was set to 10s.
The planning time has been increased to 30s and the target pose offset was
increased from 1cm to 5cm. This should make the planning problem easier to
solve and therefore more likely to pass the test.

The result from the test is bellow: @tylerjw 
subframes

### Checklist

### DIFF
--- a/moveit_ros/planning_interface/test/subframes_test.cpp
+++ b/moveit_ros/planning_interface/test/subframes_test.cpp
@@ -33,7 +33,7 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  *********************************************************************/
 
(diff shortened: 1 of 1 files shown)


tylerjw (approved):
Awesome, thank you.  I tested this locally and it works.  This should make CI more reliable.

mamoll (approved):
30s seems like a long time to plan for a problem without obstacles, but that seems like a separate issue...

tylerjw:
> 30s seems like a long time to plan for a problem without obstacles, but that seems like a separate issue...

the travis machines are sad

PR #2489, opened 2021-01-18
Use kinematics solver timeout if not specified otherwise

### LINKED ISSUE #2459: pilz_industrial_motion_planner: does not use IK timeout parameters
### Description

We recently switched to `pilz_industrial_motion_planner` in combination with `bio_ik` for our setup. In most cases the system works perfectly fine. However, sometimes we noticed IK solution not found errors, especially on slower systems.

Digging into the Pilz industrial motion planner code, I found that there is a hardcoded timeout for the IK. Instead, MoveIt's IK timeout parameters should be used. `move_group/<name>/kinematics_solver_timeout`

https://github.com/ros-planning/moveit/blob/master/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_functions.h#L70

### Your environment
* ROS Distro: Noetic
* OS Version: Ubuntu 20.20
* Source Build
* noetic-devel

### Steps to reproduce
Constrain your hardware (e.g. VM with CPU cap) and run the pilz industrial motion planner.

### Expected behaviour
IK timeout is respected and can be increased with ROS param.

### Actual behaviour
IK timeout of 0.1s is hardcoded and therefore planning will fail at exactly 0.1s if IK solution is hard to find.

Output by pilz planner:
```
[ INFO] [1607987134.054880577] [/move_group] [ros.pilz_trajectory_generation]: Generating PTP trajectory...
[ERROR] [1607987134.156704179] [/move_group] [ros.pilz_trajectory_generation]: Inverse kinematics for pose 
 0.813189
-0.577424
  0.50653 has no solution.
[ INFO] [1607987134.156867075] [/move_group] [ros.pilz_trajectory_generation]: No IK solution for goal pose
```
From our ROS program:
```
robot_command.rpl.exceptions.MovePlanningError: Planning move failed, time: 0.10122579500000001s, error_code: val: -31
[ERROR] [1608055764.203764] [/interpreter_node] [rosout]: Error executing program:
File "rando_pose_movej.py", line 14: Planning move failed, time: 0.10122579500000001s, error_code: val: -31
```

martiniil:
### Description

This should fix #2459.

Setting the default timeout to zero prevents it from overriding the kinematics solver timeout in the call to `robot_state::setFromIK`. See https://github.com/ros-planning/moveit/blob/664ae01803abf5e0b4649063102357262de9e05c/moveit_core/robot_state/src/robot_state.cpp#L1747

This makes the default `0.5`. See https://github.com/ros-planning/moveit/blob/664ae01803abf5e0b4649063102357262de9e05c/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h#L73

### Checklist

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_functions.h
+++ b/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_functions.h
@@ -61,18 +61,18 @@ namespace pilz_industrial_motion_planner
  * @param solution: solution of IK
  * @param check_self_collision: true to enable self collision checking after IK
  * computation
- * @param max_attempt: maximal attempts of IK
+ * @param timeout: timeout for IK, if not set the default solver timeout is used
  * @return true if succeed
  */
 bool computePoseIK(const robot_model::RobotModelConstPtr& robot_model, const std::string& group_name,
                    const std::string& link_name, const Eigen::Isometry3d& pose, const std::string& frame_id,
                    const std::map<std::string, double>& seed, std::map<std::string, double>& solution,
-                   bool check_self_collision = true, const double timeout = 0.1);
+                   bool check_self_collision = true, const double timeout = 0.0);
 
 bool computePoseIK(const robot_model::RobotModelConstPtr& robot_model, const std::string& group_name,
                    const std::string& link_name, const geometry_msgs::Pose& pose, const std::string& frame_id,
                    const std::map<std::string, double>& seed, std::map<std::string, double>& solution,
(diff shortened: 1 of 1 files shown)


martiniil:
@machinekoder Could you try this out?

rhaschke (approved):
This looks reasonable to me.

AndyZe:
Can you point me toward the line where the default is used, if `timeout == 0`?

(I'm sure it works, just want to verify)

AndyZe:
Ah, here it is at line 1722.  http://docs.ros.org/en/noetic/api/moveit_core/html/robot__state_8cpp_source.html

AndyZe (approved):

PR #2490, opened 2021-01-18
[ompl] Partially Thread safe model based state space type

gleichdick:
### Description

#2273 shows that the current implementation of the OMPL interface requires to modify const-qualified state objects,
by `const_cast`ing away the constness (which is UB). In this PR, some members of this state class are put into an atomic container (`AtomicCache`) which is declared as `mutable`. In this way, `const_cast` is no longer needed.

Note that the container 16 bytes big, clang requires x64 processors to have the `CMPXCHG16B` instruction (which is not the case for very old x64 processors). I've read that steam on Linux requires this instruction, and there were some discussions.
The container can't be smaller, as it has to contain at least the `flags` and the `distance` member (which are 12 bytes). If only the `flags` were atomic, data races could occur (Thread A reads flags and learns that the distance is valid, gets preempted, thread B changes the distance and the flag, gets preemted, now thread A reads the (now invalid) distance).
GCC wraps this instruction into libatomic and has a fallback.

The other members (e.g. `values` pointer) are not protected, as I expect them to be unmodified and thus not affected by potential data races.

Right now, all `load()` / `store()`s are using the default memory orders, which are safe defaults. To gain some performance, they can be adapted (e.g. to `std::memory_order_relaxed`), but I don't have enough experience to see all the side effects of altering them.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/parameterization/model_based_state_space.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/parameterization/model_based_state_space.h
@@ -41,6 +41,7 @@
 #include <moveit/robot_state/robot_state.h>
 #include <moveit/kinematic_constraints/kinematic_constraint.h>
 #include <moveit/constraint_samplers/constraint_sampler.h>
+#include <atomic>
 
 namespace ompl_interface
 {
@@ -74,6 +75,13 @@ OMPL_CLASS_FORWARD(ModelBasedStateSpace);
 class ModelBasedStateSpace : public ompl::base::StateSpace
 {
 public:
+  struct StateTypeAtomicBits
+  {
+    int tag;
+    int flags;
+    double distance;
+  };
+
   class StateType : public ompl::base::State
   {
   public:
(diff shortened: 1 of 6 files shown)


mamoll:
I like this a lot! Did you do any testing to see if this affects performance at all?

rhaschke (approved):
Ouch. Nobody noticed this before?

gleichdick:
@mamoll I didn't do any tests, just played a bit on godbolt: https://godbolt.org/z/5fxPah
Just by looking at the assemby, I'd say that there will be an impact, but I can't tell how big it is.

Side note: Good that we have code coverage analysis. I was wondering why the coverage of pose_model_state_space.cpp had decreased drastically, I checked my code again and found a bad copy-pasta bug (inverting a flag before setting it). It's fixed now.

gleichdick:
Retriggering CI

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/parameterization/model_based_state_space.h
```diff
@@ -86,84 +87,148 @@ class ModelBasedStateSpace : public ompl::base::StateSpace
       IS_GOAL_STATE = 16
     };
 
-    StateType() : ompl::base::State(), values(nullptr), tag(-1), flags(0), distance(0.0)
+    struct alignas(16) AtomicCache
```
JeroenDM, line 90: I think adding some of the explanation from the PR here as a docstring is useful. It could contain:
- Why this struct exists
- Why the `alignas(16)` is there (or a link to an external resource that explains this).

Having a docstring there compared to having to dig up this PR is maybe easier for future readers?

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/parameterization/model_based_state_space.h
```diff
@@ -86,84 +87,148 @@ class ModelBasedStateSpace : public ompl::base::StateSpace
...
+     *
+     * \param func Lambda which takes exactly one parameter, an AtomicCache Reference.
+     *
+     */
+    template <class Func>
+    void modifyCache(Func&& func) const
     {
-      flags |= IS_GOAL_STATE;
+      AtomicCache desired, expected = atomic_cache.load();
+      do
```
JeroenDM, line 223: I do not understand why the loop is necessary. Could you maybe point to an article or doc page I can read to understand this? :)

JeroenDM (changes requested):
I added a single suggestion for a little bit of documentation. (Sorry I was a bit late to the party.)

Thank you for fixing the review comment on my PR yourself, I couldn't have done it :)

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/parameterization/model_based_state_space.h
gleichdick: This is a compare-and-swap loop, for atomic Read-Modify-Write. Basically you make two copies of the atomic variable and modify one of them. `<atomic>` Then, you compare the unaltered copy with the atomic variable and if they match, you swap the atomic it with the modified variable.`</atomic>`. If the atomic variable has been altered, you have to start over. The CAS can also fail spuriously, so a loop is always needed.
In our case, we use it to set the flag and the distance at the same time. If only `flags` were atomic, thread B could change the `distance` between reading `flags` and `distance` in thread A.

gleichdick:
I added some comments.

What about 32 bit platforms (e.g. ROS2)? I've read that there are algorithms for 16 byte CAS for 32bit architectures, but I don't know whether they are better than a mutex or a read-write lock. On x64 (with `CMPXCHG16B`) and ARMv8.1 (`CASP`) we're fine.

PR #2492, opened 2021-01-20
Add method to get Servo status via C++, thread safety for paused_

AndyZe:
Add a simple method, `getServoStatus()`. Also, fix thread safety on the `paused_` flag. We originally thought it would be safe as a regular bool, but I found in testing that was not true.

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -285,7 +285,7 @@ private:
 
   // Status
   StatusCode status_ = StatusCode::NO_WARNING;
-  bool paused_ = false;
+  std::atomic<bool> paused_;
   bool twist_command_is_stale_ = false;
   bool joint_command_is_stale_ = false;
   bool ok_to_publish_ = false;
--- a/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
@@ -104,6 +104,12 @@ public:
    */
   void changeRobotLinkCommandFrame(const std::string& new_command_frame);
 
+  /** \brief Retrieve the status code */
+  StatusCode getStatus() const
+  {
+    return status_;
+  }
+
 private:
   /** \brief Run the main calculation loop */
   void mainCalcLoop();
--- a/moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/pose_tracking.h
@@ -118,6 +118,12 @@ public:
   /** \brief Re-initialize the target pose to an empty message. Can be used to reset motion between waypoints. */
   void resetTargetPose();
 
+  /** \brief Retrieve Servo status code */
+  StatusCode getServoStatus() const
+  {
+    return servo_->getStatus();
+  }
+
   // moveit_servo::Servo instance. Public so we can access member functions like setPaused()
   std::unique_ptr<moveit_servo::Servo> servo_;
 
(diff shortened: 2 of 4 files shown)


tylerjw (commented):
Wouldn't it make more sense to expose pause/unpause as a service?  Also, the status is something that is published and can be subscribed to so I'm not sure we need a C++ interface for that.

The reasoning behind this is as we transition to ros2 we should design parts of moveit to have ros interfaces instead of c++ so they can easily be used as composable nodes.  For moveit1 it would be good to make interactions with servo go through ros interfaces instead of c++ function calls because that will make the transition to ros2 easier.

As to the paused_ flag.  What was the threading issue you were experiencing?  Was it reporting paused when it wasn't or vice versa because there was a different value for it in ram vs in the registers for the thread?  Or was it not pausing when you asked it to because it was modifying a different copy of it (ram or register) when you asked for it to pause?  Can you point me to an example where this was happening?  It seems odd to me because we don't seem to have any RMW (read-modify-write) operations on that variable which are (in my understanding) the type of operations that the `atomic` type protects against.

AndyZe:
> As to the paused_ flag. What was the threading issue you were experiencing? Was it reporting paused when it wasn't or vice versa because there was a different value for it in ram vs in the registers for the thread?

I was calling `setPaused()` from another thread while printing the value of `paused_` here in servoCalcs. And I saw that it often didn't change. (I _think_ it was printing 0 when it should have printed 1). Will point you toward our client project where that was happening.

Will respond to your other comments later. But you know as well as I do, speed is important on that project so it was both convenient and fast to have a C++ method to `getStatus()`.

tylerjw:
> Will respond to your other comments later. But you know as well as I do, speed is important on that project so it was both convenient and fast to have a C++ method to `getStatus()`.

Because of how we are using servo subscribing to the status wherever we need it and updating the state there should be just as fast (and asycronous) because ros messaging will use shard memory within the same process when you pub/sub with shared_ptr versions of messages.

tylerjw:
> I was calling `setPaused()` from another thread while printing the value of `paused_` here in servoCalcs. And I saw that it often didn't change. (I _think_ it was printing 0 when it should have printed 1). Will point you toward our client project where that was happening.

Ok, that's prof that it isn't working the way I thought it did, I'm sorry.  I'll approve that change if you want to split it out from the get status function.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -104,6 +104,12 @@ class ServoCalcs
    */
   void changeRobotLinkCommandFrame(const std::string& new_command_frame);
 
+  /** \brief Retrieve the status code */
+  StatusCode getStatus() const
+  {
+    return status_;
+  }
```
jliukkonen-pn, line 111: I think there is a data race on the `status_` variable is you do it this way. The reason being that the variable is being modified in a thread that executes `mainCalcLoop` whereas the getter can be (and will be) called from another thread or various different threads. However, the `calculateSingleIteration` function already has ROS publisher for the Servo status so I think it would work just fine if the PoseTracking interface had a subscriber for that topic and would update it's own copy of status information using that subscriber.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -104,6 +104,12 @@ class ServoCalcs
...
+
```
jliukkonen, line 112: I think there is a data race on the status_ variable if you do it this way. The reason being that the variable is being modified in a thread that executes mainCalcLoop whereas the getter can be (and will be) called from another thread or various different threads. However, the calculateSingleIteration function already has a ROS publisher for the Servo status so I think it would work just fine if the PoseTracking interface had a subscriber for that topic and would update it's own copy of the status information using that subscriber.

PR #2496, opened 2021-01-21
Servo: Do not break out of loop -- need to update low pass filters

AndyZe:
We implemented this short circuit for efficiency, but in the case of `stop_requested_` we want to continue updating the current robot state and the low pass filters. This prevents a jerk from occurring when Servoing resumes.


tylerjw (approved):
Looks good, I didn't think of this use case when I created low_latency_mode.  Good catch!

PR #2497, opened 2021-01-22
Fix logic in PlanningScene::clearDiffs() for multiple CollisionDetectors

### LINKED ISSUE #2485: Segfault from the PlanningScene after allocating for Bullet collision detection
Attempting to allocate for Bullet collision detection like this:

```
planning_scene_monitor_->getPlanningScene()->setActiveCollisionDetector(collision_detection::CollisionDetectorAllocatorBullet::create());
```

or like this:

`
  {
     planning_scene_monitor::LockedPlanningSceneRW ls(planning_scene_monitor_);
      ls->setActiveCollisionDetector(collision_detection::CollisionDetectorAllocatorBullet::create());
  }
`
Causes a seg fault. Per gdb:

```
[Switching to Thread 0x7fff86ffb700 (LWP 25923)]
0x00007ffff20a1b9b in planning_scene::PlanningScene::clearDiffs() ()
   from /home/andyz/ws_al/devel/lib/libmoveit_planning_scene.so.1.1.1

```

AndyZe:
Fix a bug for the case when `it.second->parent_` is null (i.e. if the current Collision Detector is the parent or the only one).

Provide better Doxygen comments for `clearDiffs()`

This seems like a pretty deep bug. I guess it wasn't found previously because nobody is using >1 collision detection algorithm.

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -389,17 +389,15 @@ void PlanningScene::clearDiffs()
 
     if (it.second->parent_)
     {
-      it.second->cenv_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_, world_);
-      it.second->cenv_const_ = it.second->cenv_;
-
       it.second->cenv_unpadded_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_unpadded_, world_);
       it.second->cenv_unpadded_const_ = it.second->cenv_unpadded_;
-    }
-    else
-    {
-      it.second->copyPadding(*parent_->active_collision_);
 
       it.second->cenv_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_, world_);
+      it.second->cenv_const_ = it.second->cenv_;
+    }
+    else  // This is the parent CollisionDetector
+    {
+      it.second->copyPadding(*parent_->active_collision_);
       it.second->cenv_const_ = it.second->cenv_;
     }
   }
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -774,8 +774,10 @@ public:
   void removeObjectType(const std::string& id);
   void getKnownObjectTypes(ObjectTypeMap& kc) const;
 
-  /** \brief Clear the diffs accumulated for this planning scene, with respect to the parent. This function is a no-op
-   * if there is no parent specified. */
+  /** \brief Clear the diffs accumulated for this planning scene, with respect to:
+   * the parent PlanningScene (if it exists)
+   * the parent CollisionDetector (if it exists)
+   * This function is a no-op if there is no parent planning scene. */
   void clearDiffs();
 
   /** \brief If there is a parent specified for this scene, then the diffs with respect to that parent are applied to a


moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -387,20 +387,23 @@ void PlanningScene::clearDiffs()
...
+    if (parent_->getActiveCollisionDetectorName() != getActiveCollisionDetectorName())
     {
-      it.second->cenv_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_, world_);
-      it.second->cenv_const_ = it.second->cenv_;
+      ROS_ERROR_NAMED(LOGNAME, "Active collision detector of parent does not match this child scene's type");
+      return;
+    }
+
+    it.second->cenv_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_, world_);
+    it.second->cenv_const_ = it.second->cenv_;
```
AndyZe, line 397: These two lines were repeated in the if...else, so I moved them outside of it.

AndyZe:
I'm going to reduce the scope of this PR to the small code simplification only and let #2499 handle the issue with `parent_` pointer.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -389,17 +389,15 @@ void PlanningScene::clearDiffs()
...
+
+      it.second->cenv_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_, world_);
+      it.second->cenv_const_ = it.second->cenv_;
     }
-    else
+    else  // This is the parent CollisionDetector
     {
       it.second->copyPadding(*parent_->active_collision_);
-
-      it.second->cenv_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_, world_);
```
AndyZe, line 402: problem here because `it.second->parent_` is null

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -389,17 +389,15 @@ void PlanningScene::clearDiffs()
...
     if (it.second->parent_)
     {
-      it.second->cenv_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_, world_);
-      it.second->cenv_const_ = it.second->cenv_;
-
       it.second->cenv_unpadded_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_unpadded_, world_);
       it.second->cenv_unpadded_const_ = it.second->cenv_unpadded_;
+
+      it.second->cenv_ = it.second->alloc_->allocateEnv(it.second->parent_->cenv_, world_);
+      it.second->cenv_const_ = it.second->cenv_;
```
AndyZe, line 396: move these 2 lines down to match the order below: unpadded, then padded

moveit_core/planning_scene/src/planning_scene.cpp
felixvd: I'm not sure I understand. Why is this line deleted?

felixvd (commented):
Appreciate the improved comments, but don't know why a line of code was deleted.

moveit_core/planning_scene/src/planning_scene.cpp
AndyZe: If you look at line 390, `if (it.second->parent_)` --> the pointer is null, so this will segfault.

That also means there is no parent CollisionDetector, so copying the parent's collision environment doesn't make sense.

I figure the original author copy/pasted this from L395 but forgot to update it.

jliukkonen (approved):

j-petit (approved):
Looks good to me.

tylerjw (approved):
lgtm

JafarAbdi (approved):

PR #2498, opened 2021-01-24
Clear model-based planning context immediately after planning finishes.

tahsinkose:
### Description

I have encountered this while designing a profile test for our custom constrained sampler. We needed verbose output logged in the destructor of it after each planning attempt. Current behavior in MoveIt is to deallocate constrained samplers in the _next_ planning attempt implicitly via `mbpc->clear()`, when the cached `ModelBasedPlanningContext` (`mbpc`) is retrieved at `PlanningContextManager` . With this change, deallocation is done just after planning finishes. 

### Checklist

### DIFF
--- a/moveit_setup_assistant/CMakeLists.txt
+++ b/moveit_setup_assistant/CMakeLists.txt
@@ -81,9 +81,11 @@ set(HEADERS
 add_library(${PROJECT_NAME}_tools
   src/tools/compute_default_collisions.cpp
   src/tools/moveit_config_data.cpp
+  src/tools/xml_manipulation.cpp
   src/tools/collision_linear_model.cpp
   src/tools/collision_matrix_model.cpp
   src/tools/rotated_header_view.cpp
+  src/tools/xml_syntax_highlighter.cpp
 )
 set_target_properties(${PROJECT_NAME}_tools PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
 target_link_libraries(${PROJECT_NAME}_tools
--- a/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
+++ b/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
@@ -199,8 +199,9 @@ public:
     POSES = 1 << 6,
     END_EFFECTORS = 1 << 7,
     PASSIVE_JOINTS = 1 << 8,
-    AUTHOR_INFO = 1 << 9,
-    SENSORS_CONFIG = 1 << 10,
+    SIMULATION = 1 << 9,
+    AUTHOR_INFO = 1 << 10,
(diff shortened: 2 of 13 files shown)


felixvd:
Can't look at it right away, but @JeroenDM might know this part of the codebase.

JeroenDM:
The change sounds reasonable to me, but don't fully understand it yet. Which line in `ModelBasedPlanningContext::clear()` deallocation the constraint sampler? Is it this [ompl_simple_setup_->clear();](https://github.com/ros-planning/moveit/blob/0e5911316ed51a765fc6b08a35f48c35f275e026/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp#L555) call?

```C++
if (!multi_query_planning_enabled_)
    ompl_simple_setup_->clear();
```
Or am I missing the point?


 It's a bit unfortunate that we have to clear it in 4 different places from a maintenance point of view, but I also don't see an alternative to implement this feature.

tahsinkose:
> Is it this ompl_simple_setup_->clear(); call?

Yes. With this I move the deallocation into the lifetime of actual planning request. Previously it was like:

1. Request A. (No cached context, create new one)
2. Plan A (constrained sampler allocated in a new context)
3. Request B. (Use cached context)
4. Plan B. (deallocate previously allocated sampler, allocate a new one).

So we can coin it as lazy deallocation :slightly_smiling_face:

tahsinkose:
@JeroenDM PTAL. I've splitted OMPL-specific cleanup and MBPC-specific cleanup. Internally OMPL-specific cleanup is used, whereas the exposed API still keeps cleaning everything as previously.

mamoll:
This change would make it impossible to analyze the exploration that was performed after planning is complete. In the development of a new planner for tuning an existing planner for a new robot / environment it can be useful to visualize the tree / roadmap produced by OMPL. The data for this can be obtained with the ompl::geometric::SimpleSetup::getPlannerData method. That data issue cleared immediately if this PR is merged.

tahsinkose:
Thanks @mamoll for the detailed explanation. Now I know the rationale. I still believe deallocating samplers immediately after planning seems reasonable to me. All the other data can persist. 

The statement I'm after for RRTConnect is at https://github.com/ompl/ompl/blob/c7fc1452bb77d5dee417e25a6ce2b2c5feb6fd5c/src/ompl/geometric/planners/rrt/src/RRTConnect.cpp#L106. If it is possible to call this from MoveIt codespace at MBPC for all the sampling-based planners, it would be the solution.

simonschmeisser:
@mamoll do we have a tutorial about how to do this? Or does the visualization still need to be implemented?

JafarAbdi:
> @mamoll do we have a tutorial about how to do this? Or does the visualization still need to be implemented?

This package https://github.com/davetcoleman/ompl_visual_tools helps with the visualization, but it's not maintained anymore

PR #2499, opened 2021-01-24
Initialize the PlanningScene parent to nullptr

AndyZe:
The `parent_` pointer was uninitialized. That can lead to some problems later in planning_scene.cpp when `parent_` is checked, e.g.:

```
void PlanningScene::clearDiffs()
{
  if (!parent_)
    return;
```


v4hn (approved):
If you want to have the explicit initialization, we can add it. But there is no bug. [`shared_ptr()` and `shared_ptr(nullptr_t)` both create a nullptr-like shared_ptr object](https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr)

I approve the change in case you want to merge this patch for readability.

AndyZe:
Thanks. That makes sense. I will leave it up to somebody else to close or merge as they see fit.

PR #2500, opened 2021-01-26
ensure thread safety in clear octomap

simonschmeisser:
### Description
As commented in https://github.com/ros-planning/moveit/pull/2320#issuecomment-746096098 I recently encountered some crashes when clearOctomap is being called while some planners still run. Even though that is a weird thing to do, it still shouldn't crash.

Also #2320 uses `UPDATE_SCENE` which causes a full scene to be transmitted while `UPDATE_GEOMETRY` should be sufficient.

### Checklist

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -545,7 +545,7 @@ void PlanningSceneMonitor::clearOctomap()
     else
     {
       ROS_WARN_NAMED(LOGNAME, "Unable to clear octomap since no octomap monitor has been initialized");
-    }
+    }  // Lift the scoped lock before calling triggerSceneUpdateEvent to avoid deadlock
   }
 
   if (removed)
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -531,18 +531,25 @@ void PlanningSceneMonitor::newPlanningSceneCallback(const moveit_msgs::PlanningS
 
 void PlanningSceneMonitor::clearOctomap()
 {
-  if (scene_->getWorldNonConst()->removeObject(scene_->OCTOMAP_NS))
-    triggerSceneUpdateEvent(UPDATE_SCENE);
-  if (octomap_monitor_)
+  bool removed = false;
   {
-    octomap_monitor_->getOcTreePtr()->lockWrite();
-    octomap_monitor_->getOcTreePtr()->clear();
-    octomap_monitor_->getOcTreePtr()->unlockWrite();
+    boost::unique_lock<boost::shared_mutex> ulock(scene_update_mutex_);
+    removed = scene_->getWorldNonConst()->removeObject(scene_->OCTOMAP_NS);
+
+    if (octomap_monitor_)
+    {
+      octomap_monitor_->getOcTreePtr()->lockWrite();
+      octomap_monitor_->getOcTreePtr()->clear();
+      octomap_monitor_->getOcTreePtr()->unlockWrite();
+    }
+    else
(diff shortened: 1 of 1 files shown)


simonschmeisser:
@felixvd friendly ping as you were the last one to touch this code

v4hn (approved):
Looks good to me. Awaits a second review.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -531,18 +531,25 @@ void PlanningSceneMonitor::newPlanningSceneCallback(const moveit_msgs::PlanningS
...
+  bool removed = false;
   {
-    octomap_monitor_->getOcTreePtr()->lockWrite();
-    octomap_monitor_->getOcTreePtr()->clear();
-    octomap_monitor_->getOcTreePtr()->unlockWrite();
-  }
-  else
-  {
-    ROS_WARN_NAMED(LOGNAME, "Unable to clear octomap since no octomap monitor has been initialized");
+    boost::unique_lock<boost::shared_mutex> ulock(scene_update_mutex_);
```
felixvd, line 536: It wasn't hard to figure out how this works, but I had to think for a second because I don't use mutexes a lot. Do you want to add a comment to the effect of `// Lock scene using mutex before removing the octomap. Lock expires after the code block` for the less familiar? I'm on the fence.

felixvd (approved):
My bad or introducing the bug. LGTM. Thanks for the fix of UPDATE_GEOMETRY, too.

Will merge after you decide about the proposed comment. I'm good either way.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
simonschmeisser: I should probably add a comment about the extra brackets that are there so that the scoped lock gets lifted before notifying consumers (`triggerSceneUpdateEvent`) in order to avoid a deadlock ...

felixvd: I didn't know that the lock needs to be lifted before that call. How about adding `// Lift the scoped lock before calling triggerSceneUpdateEvent to avoid deadlock` at the end of the block?

I'm fine with not adding the other comment because I suppose it's not "[nonobvious to a reader who understands C++ well]](https://google.github.io/styleguide/cppguide.html#Implementation_Comment_Donts)".

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -531,18 +531,25 @@ void PlanningSceneMonitor::newPlanningSceneCallback(const moveit_msgs::PlanningS
...
+    if (octomap_monitor_)
+    {
+      octomap_monitor_->getOcTreePtr()->lockWrite();
+      octomap_monitor_->getOcTreePtr()->clear();
+      octomap_monitor_->getOcTreePtr()->unlockWrite();
+    }
+    else
+    {
+      ROS_WARN_NAMED(LOGNAME, "Unable to clear octomap since no octomap monitor has been initialized");
+    }
```
felixvd, line 548:
```suggestion
    }  // Lift the scoped lock before calling triggerSceneUpdateEvent to avoid deadlock
```

v4hn:
Thanks @simonschmeisser . Congratulations on getting pull request number #2500. :clown_face: 
We sure deal with a lot of patches as time passes...

felixvd:
I hope we don't have to wait until [#12345](https://bugs.python.org/issue12345) to merge [this](https://github.com/ros-planning/moveit_tutorials/pull/554)

simonschmeisser:
It took me a long time to actually find something to fix just so I could get this PR number ... :wink: 

Thanks for reviewing and merging ... I suspect there's going to be another one as we had crashes in the octree->clear part as well that should be locked by themselves

PR #2501, opened 2021-01-26
[Servo] Halt Servo command on Pose Tracking stop

AdamPettinger:
### Description

Small change to make Pose Tracking publish a 0 twist message for Servo when it is stopping. Before this, if you were using a velocity controller and stopped Pose Tracking during a movement, the last published command to Servo would be nonzero and the manipulator would still move, see attached.

It might be worth exposing some kind of `halt()` function in Servo itself, which I can do here if needed. 

Haven't been able to replicate with the Servo demonstrations (position controllers), so attached videos on hardware. In the clip, I use Pose Tracking to point the left arm camera at the right gripper, but call `stopMotion()` and cancel the pose tracking before it gets there. In the first, the arm holds the last velocity command sent until I E-stop. In the second gif, a 0-twist command is sent to Servo during Pose Tracking stopping.

pose_tracking_problem

pose_tracking_fix

### DIFF
--- a/moveit_ros/moveit_servo/src/pose_tracking.cpp
+++ b/moveit_ros/moveit_servo/src/pose_tracking.cpp
@@ -317,6 +317,15 @@ void PoseTracking::doPostMotionReset()
   cartesian_position_pids_[1].reset();
   cartesian_position_pids_[2].reset();
   cartesian_orientation_pids_[0].reset();
+
+  // Send a 0 command to Servo to halt arm motion
+  auto msg = moveit::util::make_shared_from_pool<geometry_msgs::TwistStamped>();
+  {
+    std::lock_guard<std::mutex> lock(target_pose_mtx_);
+    msg->header.frame_id = target_pose_.header.frame_id;
+  }
+  msg->header.stamp = ros::Time::now();
+  twist_stamped_pub_.publish(msg);
 }
 
 void PoseTracking::updatePIDConfig(const double x_proportional_gain, const double x_integral_gain,
--- a/moveit_ros/moveit_servo/src/pose_tracking.cpp
+++ b/moveit_ros/moveit_servo/src/pose_tracking.cpp
@@ -307,16 +307,9 @@ geometry_msgs::TwistStampedConstPtr PoseTracking::calculateTwistCommand()
   return msg;
 }
 
-void PoseTracking::doPostMotionReset()
+void PoseTracking::stopMotion()
 {
-  stop_requested_ = false;
-  angular_error_ = boost::none;
-
-  // Reset error integrals and previous errors of PID controllers
-  cartesian_position_pids_[0].reset();
-  cartesian_position_pids_[1].reset();
-  cartesian_position_pids_[2].reset();
-  cartesian_orientation_pids_[0].reset();
+  stop_requested_ = true;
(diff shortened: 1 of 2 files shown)


AndyZe:
I appreciate the bug fix!

There's already a member function called `stopMotion()`. How about we move your zero-publication inside `stopMotion()` like this:


```
  void stopMotion()
  {
    stop_requested_ = true;
    
    // Send a 0 command to Servo to halt arm motion
    auto msg = moveit::util::make_shared_from_pool<geometry_msgs::TwistStamped>();
    {
      std::lock_guard<std::mutex> lock(target_pose_mtx_);
      msg->header.frame_id = target_pose_.header.frame_id;
    }
    msg->header.stamp = ros::Time::now();
    twist_stamped_pub_.publish(msg);
  }
```

Then doPostMotionReset() would always call `stopMotion()`?

```
void PoseTracking::doPostMotionReset()
{
  stopMotion();
  angular_error_ = boost::none;

  // Reset error integrals and previous errors of PID controllers
  cartesian_position_pids_[0].reset();
  cartesian_position_pids_[1].reset();
  cartesian_position_pids_[2].reset();
  cartesian_orientation_pids_[0].reset();

  stop_requested_ = false;
}
```

AndyZe:
The only issue I see with ^ is it sets `stop_requested_ = true;` then `stop_requested_ = false;` right afterward. That's a little weird but I guess it won't cause any harm.

AdamPettinger:
Looks good to me, will do in a second.

What do you think about removing the mutex-protected bit
```
{
  std::lock_guard<std::mutex> lock(target_pose_mtx_);
  msg->header.frame_id = target_pose_.header.frame_id;
}
```
As 0 velocity is 0 velocity in any frame and Servo can handle an empty frame command

AndyZe:
I'd rather keep the mutex to be on the safe side, TBH.

AdamPettinger:
> I'd rather keep the mutex to be on the safe side, TBH.

Works for me. One thing I see about moving the halting message to `stopMotion()` is that an API call to `stopMotion()` will result in that function and the 0-twist message being published twice (in the call, and then triggered later in `doPostMotionReset()`. 

Is that a problem?

AndyZe:
i'm OK with it

AndyZe (approved):

moveit_ros/moveit_servo/src/pose_tracking.cpp
```diff
@@ -307,8 +307,23 @@ geometry_msgs::TwistStampedConstPtr PoseTracking::calculateTwistCommand()
   return msg;
 }
 
+void PoseTracking::stopMotion()
+{
+  stop_requested_ = true;
+
+  // Send a 0 command to Servo to halt arm motion
+  auto msg = moveit::util::make_shared_from_pool<geometry_msgs::TwistStamped>();
```
henningkayser, line 315: Is this really enough? It seems like this could go in after `stop_requested_` so that it has no effect.

AndyZe: It looks fine to me. `stop_requested_` stops the Pose Tracking loop. Publishing the zero twist stops the Servo loop.

PR #2502, opened 2021-01-26
WIP: Use the "group_name" for Bullet collision checks

AndyZe:
Parse links in the CollisionRequest group_name and disable other links during the collision check. We are hoping to use this to speed up collision checking. I've tested this extensively with the tutorial but will work on adding unit tests.


moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
```diff
@@ -187,8 +187,29 @@ void CollisionEnvBullet::checkRobotCollisionHelper(const CollisionRequest& req,
...
+    }
+  }
+
   manager_->contactTest(res, req, acm, false);
 
+  // Reset
+  if (!req.group_name.empty())
+  {
+    ;//manager_->setActiveCollisionObjects(active_);  // All links
+  }
```
AndyZe, line 212: One thing I'm struggling with is resetting after an iteration. I'd like to reset the active links like this, but doing so seems to null the joint_model_group specification on the next iteration --

```
for(auto link_name : active_)
      manager_->enableCollisionObject(link_name);
```

AndyZe: ah, fixed it in 240a9630d

simonschmeisser:
I'm not 120% sure but doesn't the group_name imply to include all leaves/sub branches/stuff attached to that group?

AndyZe:
good point, i'll test whether this works for attached objects

simonschmeisser:
I was actually also concerned about the gripper/end_effector which with fcl is included when you set `group_name` to that of your robot/`manipulator` even though it is not in said group but only an appendix to that group

AndyZe:
> I was actually also concerned about the gripper/end_effector which with fcl is included when you set `group_name` to that of your robot/`manipulator` even though it is not in said group but only an appendix to that group

To me, that seems like unexpected behavior on the FCL side. I'd prefer it only checked the specified group.

AndyZe:
I find that attached collision object detection does not work for this PR. But, it doesn't work for Bullet on master branch, either.

moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
```diff
@@ -187,8 +189,30 @@ void CollisionEnvBullet::checkRobotCollisionHelper(const CollisionRequest& req,
         cow->getName(), state.getAttachedBody(cow->getName())->getGlobalCollisionBodyTransforms()[0]);
   }
 
+  // Specify active collision links
+  if (!req.group_name.empty())
+  {
+    auto joint_model_group = robot_model_->getJointModelGroup(req.group_name);
+    std::vector<std::string> active_links = joint_model_group->getLinkModelNames();
```
simonschmeisser, line 196:
```suggestion
    const std::vector<std::string> active_links = joint_model_group->getUpdatedLinkModelsNames();
```
https://github.com/ros-planning/moveit/blob/74b3e30db2e8683ac17b339cc124675ae52a5114/moveit_core/collision_detection_fcl/src/collision_common.cpp#L934

https://github.com/ros-planning/moveit/blob/74b3e30db2e8683ac17b339cc124675ae52a5114/moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h#L233

simonschmeisser (changes requested):
I added a suggestion to follow the behavior implemented in fcl

moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
AndyZe: Well, it stops detecting collisions with the yellow cube when I apply that change    :/

henningkayser: That's weird as this almost follows the [enableGroup()](https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection_fcl/src/collision_common.cpp#L931) implementation used for [FCL](https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp#L294). How about using `getUpdatedLinkModelsSet()` and get the names from there? That's actually how it works for FCL, not sure if there is something odd going on with duplicate entries.

Issue #2236, opened 2020-08-06
robot_interaction: Endeffector can be attached to wrong group


MatthiasNieuwenhuisen:
Hello together!

### Description

If a robot has multiple (partially overlapping) groups defined in the SRDF and at least one group has an end_effector, this endeffector could be attached to wrong groups, too.

Example (using the panda_moveit_config from the tutorial):

The arm has a kinematic chain from panda_link0 to panda_link8 and a an end_effector attached to panda_link8:
```xml
<group name="panda_arm">
  <chain base_link="panda_link0" tip_link="panda_link8" />
</group>
<end_effector name="hand" parent_link="panda_link8" group="hand" parent_group="panda_arm" />
```

We extend the model of the arm by attaching another link to panda_link8,  named alternative_tip and define another group:
```xml
<group name="alternative">
  <chain base_link="panda_link0" tip_link="alternative_tip" />
</group>
```

We now have two groups, one ending in panda_link8, with attached end_effector, and a second one including panda_link8, but continuing.

When using the RViz Motion Planning Display, the interactive marker for the second chain is not at its tip, but at the tip of the first chain (panda_arm).

The issue is caused by https://github.com/ros-planning/moveit/blob/ddb7440890e40581925a97e13571490dc7e60fe8/moveit_ros/interface/robot_interaction/src/robot_interaction.cpp#L586

There, an end_effector is selected not only based on its parent_group, but also if any other group contains the end_effector's parent_link. To me it is unclear, if the disjunction serves a purpose, as a parent_group is explicitly defined and should be respected. It has has been introduced in commit: https://github.com/ros-planning/moveit/commit/74b37106429eae40becb0b1902f46f6e74b8fb39 .

The issue might affect non-interactive planning, too, which I haven't checked, yet.

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Source or Binary build? Both
* If binary, which release version? 
* If source, which branch? melodic-devel

### Steps to reproduce
Define a planning group that contains an intermediate link from another group with an attached endeffector (to the other group). Select this planning group in RViz to show its interactive marker.

### Expected behaviour
The interactive marker should be attached to the last link of the planning group.

![correct](https://user-images.githubusercontent.com/1028032/89537463-aeefc880-d7f9-11ea-9738-e6bf526a509d.png)

### Actual behaviour
The interactive marker is attached to the link with an attached endeffector, not considering the end_effectors defined parent group:

[...]

PR #2515, opened 2021-02-12
Check wether endeffector parent group is empty or matches the current…

MatthiasNieuwenhuisen:
Hi,

this PR addresses my issue #2236 when using the RViz Motion Planning panel.

Without this change an endeffector control might be attached to the tip of the wrong kinematic chain if the parent link of the other chain's defined endeffector is part of the selected planning chain. This is the case if the selected chain is one link shorter, such that the tip of the selected chain the parent link of the tip of the other chain, or if the selected chain is longer and contains the parent link of the tip of the other chain.

With this change endeffectors in the srdf are only considered if their parent_group attribute matches the selected planning chain or if the optional parent_group attribute is left empty.

A similar problem exists in the RobotModel, which I will address in a later PR if this gets merged.

Regards,
Matthias

### DIFF
--- a/moveit_ros/robot_interaction/src/robot_interaction.cpp
+++ b/moveit_ros/robot_interaction/src/robot_interaction.cpp
@@ -284,7 +284,7 @@ void RobotInteraction::decideActiveEndEffectors(const std::string& group, Intera
   if (smap.first)
   {
     for (const srdf::Model::EndEffector& eef : eefs)
-      if ((jmg->hasLinkModel(eef.parent_link_) && ( jmg->getName().empty() || jmg->getName() == eef.parent_group_ ) ) &&
+      if ((jmg->hasLinkModel(eef.parent_link_) && (jmg->getName().empty() || jmg->getName() == eef.parent_group_)) &&
           jmg->canSetStateFromIK(eef.parent_link_))
       {
         // We found an end-effector whose parent is the group.
--- a/moveit_ros/robot_interaction/src/robot_interaction.cpp
+++ b/moveit_ros/robot_interaction/src/robot_interaction.cpp
@@ -284,7 +284,7 @@ void RobotInteraction::decideActiveEndEffectors(const std::string& group, Intera
   if (smap.first)
   {
     for (const srdf::Model::EndEffector& eef : eefs)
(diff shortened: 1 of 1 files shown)


felixvd (approved):
I had some trouble following this, but I think I got the idea. Yes, if two overlapping planning groups exist (e.g. for different tips on an end effector), an end effector might be misassigned. This should fix it.

Most of my doubt comes from the fact that I am still unsure about best practices when defining end effectors in the SRDF, and [our tutorial](https://ros-planning.github.io/moveit_tutorials/doc/urdf_srdf/urdf_srdf_tutorial.html) is not overly helpful. I assume you created the second planning group to display the interactive marker at the tip of the panda's gripper. Is that the best way to place the marker? Should there be a setting in Rviz? I guess that's a discussion for another issue.

felixvd:
Fixes #2236

jschleicher:
The failing test seems unrelated. Re-opening to start another travis run.

moveit_ros/robot_interaction/src/robot_interaction.cpp
```diff
@@ -284,7 +284,7 @@ void RobotInteraction::decideActiveEndEffectors(const std::string& group, Intera
   if (smap.first)
   {
     for (const srdf::Model::EndEffector& eef : eefs)
-      if ((jmg->hasLinkModel(eef.parent_link_) || jmg->getName() == eef.parent_group_) &&
+      if ((jmg->hasLinkModel(eef.parent_link_) && (jmg->getName().empty() || jmg->getName() == eef.parent_group_)) &&
```
v4hn, line 287:
```suggestion
      if ((jmg->getName() == eef.parent_group_ || (eef.parent_group_.empty() && jmg->hasLinkModel(eef.parent_link_)) &&
```

v4hn (commented):
I don't understand your logic, so I propose an alternative that makes more sense to me (untested).
The motivation for the disjunction was most likely an idiom you find in other places of MoveIt as well: Try to be smart if the user provides conflicting or insufficient information.
In this case the information would be missing knowledge about the parent group of an end effector. So I would propose to use the logic only when the parent group was not specified by the user.

v4hn:
@MatthiasNieuwenhuisen any chance you can test whether my proposed patch also addresses your problem?

PR #2519, opened 2021-02-16
Servo: add missing include

v4hn:
The code uses `std::atomic<bool>` further below, but relies on other headers including it.
It's not always there though. Compiling failed for me on Lunar-Linux with ROS melodic.


JafarAbdi (approved):

AndyZe (approved):
Thanks!

Issue #2518, opened 2021-02-16
Could NOT find OpenMP (clang, building from source)


ValerioMa:
### Description
When compiling from source on Ubuntu 18.04 ROS melodic using clang OpenMP can not be found and this generate a compilation error

### Your environment
* ROS Distro: [Melodic]
* OS Version: Ubuntu 18.04
* Building from source
* If binary, which release version? release
* If source, which branch? melodic-devel

### Steps to reproduce
compile from source using clang compiler

### Backtrace or Console output

Console error message [here](https://gist.github.com/ValerioMa/ac77eb2cae62e09f460f6d0e2139f00e)

ValerioMa:
I found a relevant discussion [here](https://github.com/fospald/fibergen/issues/5).

A valid solution for my case was to add  these lines in the `moveit_ros_perception` CMake file :
```cmake
if(CMAKE_C_COMPILER_ID MATCHES "Clang")
    set(OpenMP_C "${CMAKE_C_COMPILER}")
    set(OpenMP_C_FLAGS "-fopenmp=libomp -Wno-unused-command-line-argument")
    set(OpenMP_C_LIB_NAMES "libomp" "libgomp" "libiomp5")
    set(OpenMP_libomp_LIBRARY ${OpenMP_C_LIB_NAMES})
    set(OpenMP_libgomp_LIBRARY ${OpenMP_C_LIB_NAMES})
    set(OpenMP_libiomp5_LIBRARY ${OpenMP_C_LIB_NAMES})
endif()
if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  set(OpenMP_CXX "${CMAKE_CXX_COMPILER}")
  set(OpenMP_CXX_FLAGS "-fopenmp=libomp -Wno-unused-command-line-argument")
  set(OpenMP_CXX_LIB_NAMES "libomp" "libgomp" "libiomp5")
  set(OpenMP_libomp_LIBRARY ${OpenMP_CXX_LIB_NAMES})
  set(OpenMP_libgomp_LIBRARY ${OpenMP_CXX_LIB_NAMES})
  set(OpenMP_libiomp5_LIBRARY ${OpenMP_CXX_LIB_NAMES})
endif()
```

PR #2520, opened 2021-02-16
Melodic devel

ValerioMa:
### Description

Fix clang compilation error in finding OpenMP reported in issue #2518

### DIFF
--- a/moveit_ros/perception/CMakeLists.txt
+++ b/moveit_ros/perception/CMakeLists.txt
@@ -51,8 +51,26 @@ find_package(catkin REQUIRED COMPONENTS
   moveit_ros_occupancy_map_monitor
 )
 
-find_package(Eigen3 REQUIRED)
+# # Find OpenMP https://stackoverflow.com/questions/46414660/macos-cmake-and-openmp
+if(CMAKE_C_COMPILER_ID MATCHES "Clang")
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
Sorry, but this isn't convincing to me. All these settings should be handled by `find_package(OpenMP)`. If there is a bug in `FindOpenMP.cmake`, it should be fixed there instead of providing a laborious workaround here.

How is this related to #2474? Are you working in a OSX environment?

ValerioMa:
I close this because it is an exact duplicate of #2474 . Sorry if I didn't find it before also the proposed change is taken from the same StackOverflow post [here](https://stackoverflow.com/questions/46414660/macos-cmake-and-openmp)

rhaschke:
So, you are working in a OSX env as well? Does the alternative solution suggested in https://github.com/ros-planning/moveit/pull/2474#issuecomment-760106489 work?

ValerioMa:
No I am using Clang in Ubuntu 18.04 as reported in the linked issue

rhaschke:
In this case, I consider this PR independent. Did you try to install `libomp-dev` as suggested [here](https://askubuntu.com/questions/900702/how-to-compiler-openmp-program-using-clang)?

ValerioMa:
`apt-get --reinstall libomp-dev` fixed my problem thanks for helping me!

PR #2521, opened 2021-02-16
It's not an error not to define a plugin

v4hn:
An info message is fine. But octomap plugins are explicitly optional,
so a setup that does not define a plugin (the default!) must not report errors.

This is basically a fixup for https://github.com/ros-planning/moveit/pull/1873

### DIFF
--- a/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
+++ b/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
@@ -181,7 +181,7 @@ void OccupancyMapMonitor::initialize()
     }
   }
   else
-    ROS_INFO_NAMED(LOGNAME, "No 3D sensor plugin defined for octomap updates");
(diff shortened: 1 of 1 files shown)


moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -181,7 +181,7 @@ void OccupancyMapMonitor::initialize()
     }
   }
   else
-    ROS_ERROR_NAMED(LOGNAME, "Failed to find 3D sensor plugin parameters for octomap generation");
+    ROS_INFO_NAMED(LOGNAME, "No 3D sensor plugin defined for octomap updates");
```
rhaschke, line 184:
```suggestion
    ROS_INFO_NAMED(LOGNAME, "No 3D sensor plugin(s) defined for octomap updates");
```

rhaschke (approved):
Agree. Minor suggestion to further improve the message.

PR #2528, opened 2021-02-22
Update doxygen comments for distance() and interpolate()

### LINKED ISSUE #2527: JointModel interpolate() -- time argument is a misnomer
The [interpolate()](http://docs.ros.org/en/noetic/api/moveit_core/html/revolute__joint__model_8cpp_source.html#l00191) function takes an argument called `t`:

`  void RevoluteJointModel::interpolate(const double* from, const double* to, const double t, double* state) const`

The [description](http://docs.ros.org/en/noetic/api/moveit_core/html/classmoveit_1_1core_1_1RevoluteJointModel.html#a8ced1a14641a067315971c6bfa284384) says:

`Computes the state that lies at time t in [0, 1] on the segment that connects from state to to state. The memory location of state is not required to be different from the memory of either from or to. `

So it seems `t` is really a fraction. I think this is confusing.

AndyZe:
Update documentation of `interpolate()` and `distance()` to clarify a few things:

- The `t` parameter of interpolate isn't a time, it's a fraction
- `distance()` just sums the joint difference, for all joints. It's not an L2 norm or SSE or anything like that. Verify [here](http://docs.ros.org/en/noetic/api/moveit_core/html/robot__model_8cpp_source.html#l01236).

Fixes #2527

### DIFF
--- a/moveit_core/robot_model/include/moveit/robot_model/robot_model.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/robot_model.h
@@ -338,7 +338,17 @@ public:
   }
   double getMaximumExtent(const JointBoundsVector& active_joint_bounds) const;
 
+  /** \brief Return the sum of joint distances between two states. Only considers active joints. */
   double distance(const double* state1, const double* state2) const;
+
+  /**
+   * Interpolate between "from" state, to "to" state. Mimic joints are correctly updated.
+   *
+   * @param from interpolate from this state
+   * @param to to this state
+   * @param t a fraction in the range [0 1]. If 1, the result matches "to" state exactly.
+   * @param state holds the result
+   */
   void interpolate(const double* from, const double* to, double t, double* state) const;
 
   /** \name Access to joint groups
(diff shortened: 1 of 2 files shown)


tylerjw (approved):
Thank you for doing this!  I read through these and they look good.  I'll merge this once it passes CI if no one else requests changes.

simonschmeisser (approved):
thanks especially for the corner-case info about active joints!

v4hn:
> `distance()` just sums the joint difference, for all joints. It&#39;s not an L2 norm or SSE or anything like that.

Just because of the way you formulate this: summing up element-wise differences is the standard [L1 norm](https://en.wikipedia.org/wiki/Taxicab_geometry) and I'm missing that term a bit in your patch. There was even a proposal to add weighting to the components, but it was not merged as far as I know.

AndyZe:
> `distance()` just sums the joint difference, for all joints. It&#39;s not an L2 norm or SSE or anything like that.
> Just because of the way you formulate this: summing up element-wise differences is the standard [L1 norm](https://en.wikipedia.org/wiki/Taxicab_geometry) and I'm missing that term a bit in your patch. There was even a proposal to add weighting to the components, but it was not merged as far as I know.

I'm porting this to MoveIt2 now and will mention L1 norm  :+1:

PR #2530, opened 2021-02-26
Check whether we can transform to the frame an object is defined in

LoyVanBeek:
### Description

Fixes https://github.com/ros-planning/moveit/issues/323. When adding a new object to the planning scene and MoveIt cannot (yet?) transform the frame_id a new `CollisionObject` is defined in, you'll get a message 
`[/move_group][ERROR] [...]: Unable to transform from frame 'non_existant_frame' to frame '/map'. Returning identity.`
and get the object inserted in the wrong place. Moreover, the service `ApplyPlanningScene` returns with `success = true` and afterwards the object's ID is also in the PlanningSceneInterface's known_objects. Lies! Deception!

### Checklist
- [ ] ~~Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)~~ Not needed I think, this is the intended behavior

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1694,6 +1694,12 @@ bool PlanningScene::processCollisionObjectMsg(const moveit_msgs::CollisionObject
     if (object.operation == moveit_msgs::CollisionObject::ADD && world_->hasObject(object.id))
       world_->removeObject(object.id);
 
+    if (!getTransforms().canTransform(object.header.frame_id))
+    {
+      ROS_ERROR_STREAM_NAMED("planning_scene", "Cannot transform to object's frame_id '" << object.header.frame_id << "'");
+      return false;
+    }
+
     const Eigen::Affine3d& t = getTransforms().getTransform(object.header.frame_id);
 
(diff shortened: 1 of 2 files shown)


v4hn:
Thank you for the patch @LoyVanBeek .
However, this change targets `kinetic-devel` (which is fine in itself) and the problem does not exist anymore in `master` and `melodic-devel`.

The corresponding patches were provided by @rhaschke in https://github.com/ros-planning/moveit/pull/1406 and https://github.com/ros-planning/moveit/pull/2113 .

If you want to see this fixed in `kinetic-devel`, please consider backporting the full scope of these commits to fix all associated issues. Because of other changes this is no trivial cherry-pick.

LoyVanBeek:
Aha, hadn't seen those yet unfortunately. In that case, I'll stop right here (I was building a testcase). Instead, I'll make a workaround in the caller of this service and leave that until we finally switch to noetic (soon I hope).

LoyVanBeek:
Aha, hadn't seen those yet unfortunately. In that case, I'll stop right here (I was building a testcase). Instead, I'll make a workaround in the caller of this service and leave that until we finally switch to noetic (soon I hope).

PR #2532, opened 2021-02-28
Add utility functions to Python PSI

felixvd:
I found a few very basic functions that I had added in the past, but apparently forgot to push upstream. The `ApplyPlanningScene` function seems to have been implemented already in `wrap_python_planning_scene_interface.cpp` but not exposed in the Python class, so I added that as well.

The interface seems not very useful without these functions and I'd like to use them in a different project where we don't build MoveIt from source, so it would be great if these found their way into the binaries.

This should be ported back to Melodic, possibly even Kinetic.

### Checklist

### DIFF
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -78,6 +78,12 @@ class PlanningSceneInterface(object):
             else:
                 self._pub_co.publish(collision_object)
 
+    def add_object(self, collision_object):
+        """
+        Add an object to the planning scene
+        """
+        self.__submit(collision_object, attach=False)
+
     def add_sphere(self, name, pose, radius=1):
         """
         Add a sphere to the planning scene
@@ -118,6 +124,12 @@ class PlanningSceneInterface(object):
         co.planes = [p]
         co.plane_poses = [pose.pose]
         self.__submit(co, attach=False)
+    
+    def attach_object(self, attached_collision_object):
+        """
+        Attach an object in the planning scene
+        """
+        self.__submit(attached_collision_object, attach=True)
 
     def attach_mesh(self, link, name, pose=None, filename='', size=(1, 1, 1), touch_links=[]):
         aco = AttachedCollisionObject()
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -235,6 +235,12 @@ class PlanningSceneInterface(object):
         """
         return self._psi.apply_collision_object(conversions.msg_to_string(collision_object_message))
     
+    def apply_planning_scene(self, planning_scene_message):
+        """
+        Applies the planning scene message.
+        """
+        return self._psi.apply_planning_scene(conversions.msg_to_string(planning_scene_message))
+
     @staticmethod
     def __make_existing(name):
         """
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -230,12 +230,6 @@ class PlanningSceneInterface(object):
             aobjs[key] = msg
         return aobjs
     
-    def apply_collision_object(self, collision_object_message):
-        """
-        Applies the collision object message to the scene.
-        """
-        return self._psi.apply_collision_object(conversions.msg_to_string(collision_object_message))
-    
     def apply_planning_scene(self, planning_scene_message):
         """
         Applies the planning scene message.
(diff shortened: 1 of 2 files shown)


moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -55,10 +55,13 @@
 
 
 class PlanningSceneInterface(object):
-    """ Simple interface to making updates to a planning scene """
+    """ 
+    Python interface for a planning scene.
```
simonschmeisser, line 59: Does this provide an interface to modifying **a** `PlanningScene` or to modifying **the** `PlanningScene` curated by the `PlanningSceneMonitor`? (Not sure if the first one even exists in Python)

felixvd: True, it's the latter. A Python interface for the `PlanningSceneInterface`, not the `PlanningScene`.

There are no Python bindings for the `PlanningScene`, although there is one for the `RobotState`. There has been discussion in #900 and I should probably raise the issue again.

simonschmeisser (approved):

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -213,6 +229,18 @@ def get_attached_objects(self, object_ids=[]):
...
+        """
+        Applies the collision object message to the scene.
+        """
+        return self._psi.apply_collision_object(conversions.msg_to_string(collision_object_message))
+    
+    def apply_planning_scene(self, planning_scene_message):
+        """
+        Applies the planning scene message.
+        """
+        return self._psi.apply_planning_scene(conversions.msg_to_string(planning_scene_message))
```
v4hn, line 243: I don't think we should add these methods.

[`__submit` will already use the service](https://github.com/ros-planning/moveit/pull/2532/files#diff-ea03d418c2c0d6b5328a50bd314ece97aa036b872aaa8d04f37c82c3490a0fb3R78) iff the PSI object is instantiated as [`synchronous`](https://github.com/ros-planning/moveit/pull/2532/files#diff-ea03d418c2c0d6b5328a50bd314ece97aa036b872aaa8d04f37c82c3490a0fb3R71).

There is no need or merit in depending on the C++ implementation here.

felixvd: `__submit` is not user-facing, so users of this class can't publish their own `CollisionObject`. 

I also don't see the problem with depending on the C++ implementation.

I can implement `apply_collision_object` and `apply_attached_collision_object` with a call to `__submit`, but a pure Python API doesn't seem like it will be maintained, so adding C++ bindings with familiar names seems better.

v4hn: > __submit is not user-facing, so users of this class can't publish their own CollisionObject.

That's true, but that's why you added the `add_object` method that *uses* `__submit` (just like almost every other method in this class)!
So you basically add a second way to do the same thing if you also expose the C++ `apply*` methods.

felixvd: Oh. Right. These commits are so old that I forgot that I did that. I just knew that they worked and we've been using them.

I'm not sure what I think about 1) the naming being different from C++ for no reason and 2) maintaining a pure Python API instead of bindings as described in #900, but I can leave out the C++ methods for now.

Are you good with `apply_collision_object` and `apply_attached_collision_object` calling `__submit`?

v4hn: > I'm not sure what I think about 1) the naming being different from C++ for no reason and

neither am I, but that's what we got. The idea back then was to avoid breaking the whole class, but still support the update service for all existing methods. The only proper way to have a 1-to-1 mapping with the C++ classes would be to create an entirely new package for it. It's an interesting idea, but quite some work too. :-)

> Are you good with apply_collision_object and apply_attached_collision_object calling __submit?

No, because if `synchronous == False`, these methods would use the topic interfaces then!
With the current design we should not have the add vs apply distinction we have in C++.

felixvd: Also true.

Alright, I took `applyCollisionObject` out. `applyPlanningScene` was there already though, it just wasn't exposed in Python. Surely that was an oversight? I don't see a way to publish a PlanningScene message synchronously otherwise.

Yes, more Python bindings would be great. Someone from #900 should start adding PlanningScene (not PSI) bindings for WMD.

felixvd:
Looks like this CI failure is unrelated:

> anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

v4hn:
> `applyPlanningScene` was there already though, it just wasn't exposed in Python. Surely that was an oversight? I don't see a way to publish a PlanningScene message synchronously otherwise.

True. To stay consistent with the current layout we should in theory decide based on `synchronous` whether the PlanningScene is sent via `/planning_scene` or apply it via the service.
However there is currently no such Publisher and I don't think it's overly important to have it.

v4hn (approved):

PR #2540, opened 2021-03-08
Sanitize CHOMP initialization method parameter

onionsflying:
Return false when interpolation method is invalid.

Edit by v4hn: Additionally verify parameter during planner initialization.

### DIFF
--- a/moveit_planners/chomp/chomp_motion_planner/src/chomp_parameters.cpp
+++ b/moveit_planners/chomp/chomp_motion_planner/src/chomp_parameters.cpp
@@ -80,4 +80,18 @@ void ChompParameters::setRecoveryParams(double learning_rate, double ridge_facto
   this->planning_time_limit_ = planning_time_limit;
   this->max_iterations_ = max_iterations;
 }
+
+const std::vector<std::string> ChompParameters::VALID_INITIALIZATION_METHODS{ "quintic-spline", "linear", "cubic",
+                                                                              "fillTrajectory" };
+
+bool ChompParameters::setTrajectoryInitializationMethod(std::string method)
+{
+  if (std::find(VALID_INITIALIZATION_METHODS.cbegin(), VALID_INITIALIZATION_METHODS.cend(), method) !=
+      VALID_INITIALIZATION_METHODS.end())
+  {
+    this->trajectory_initialization_method_ = std::move(method);
+    return true;
+  }
+  return false;
+}
 }  // namespace chomp
--- a/moveit_planners/chomp/chomp_optimizer_adapter/src/chomp_optimizer_adapter.cpp
+++ b/moveit_planners/chomp/chomp_optimizer_adapter/src/chomp_optimizer_adapter.cpp
@@ -150,12 +150,18 @@ public:
       ROS_INFO_STREAM(
           "Param use_stochastic_descent was not set. Using default value: " << params_.use_stochastic_descent_);
     }
-    if (!nh.getParam("trajectory_initialization_method", params_.trajectory_initialization_method_))
+    // default
+    params_.trajectory_initialization_method_ = std::string("fillTrajectory");
+    std::string trajectory_initialization_method;
+    if (!nh.getParam("trajectory_initialization_method", trajectory_initialization_method))
     {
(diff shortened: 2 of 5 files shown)


v4hn:
Thanks for your contribution @onionsflying .
I extended your one-line patch to validate parameters during setup already.

Please verify this works for you too.

onionsflying:
Thanks, it looks good to me. @v4hn

moveit_planners/chomp/chomp_optimizer_adapter/src/chomp_optimizer_adapter.cpp
```diff
@@ -150,12 +150,18 @@ class OptimizerAdapter : public planning_request_adapter::PlanningRequestAdapter
...
+    params_.trajectory_initialization_method_ = std::string("fillTrajectory");
+    std::string trajectory_initialization_method;
+    if (!nh.getParam("trajectory_initialization_method", trajectory_initialization_method))
     {
-      params_.trajectory_initialization_method_ = std::string("fillTrajectory");
-      ROS_INFO_STREAM("Param trajectory_initialization_method was not set. Using New value as: "
+      ROS_INFO_STREAM("Param trajectory_initialization_method was not set. Using value: "
                       << params_.trajectory_initialization_method_);
     }
+    else if (!params_.setTrajectoryInitializationMethod(trajectory_initialization_method))
```
simonschmeisser, line 161: can we have brackets here as well?

moveit_planners/chomp/chomp_motion_planner/src/chomp_parameters.cpp
```diff
@@ -80,4 +80,18 @@ void ChompParameters::setRecoveryParams(double learning_rate, double ridge_facto
...
   this->max_iterations_ = max_iterations;
 }
+
+const std::vector<std::string> ChompParameters::VALID_INITIALIZATION_METHODS{ "quintic-spline", "linear", "cubic",
+                                                                              "fillTrajectory" };
+
+bool ChompParameters::setTrajectoryInitializationMethod(std::string method)
+{
+  if (std::find(VALID_INITIALIZATION_METHODS.cbegin(), VALID_INITIALIZATION_METHODS.cend(), method) !=
+      VALID_INITIALIZATION_METHODS.end())
```
simonschmeisser, line 90:
```suggestion
  if (VALID_INITIALIZATION_METHODS.count(method) != 1)
```
could be used if you switch from `std::vector<std::string>` to `std::set<std::string>` (or `.contains(method)` starting with c++ 20 ...)

moveit_planners/chomp/chomp_motion_planner/src/chomp_parameters.cpp
```diff
@@ -80,4 +80,18 @@ void ChompParameters::setRecoveryParams(double learning_rate, double ridge_facto
...
+  {
+    this->trajectory_initialization_method_ = std::move(method);
```
simonschmeisser, line 92:
```suggestion
    this->trajectory_initialization_method_ = method;
```
if you pass a `const std::string& method` you can micro-optimize even more :grin:

simonschmeisser (approved):
Thanks for your PR! Don't feel annoyed by my nitpicking please

simonschmeisser:
Oh and one more thing: there's an edit link somewhere on this page and you could use it to set a more descriptive PR title (you could copy the commit message)

moveit_planners/chomp/chomp_motion_planner/src/chomp_parameters.cpp
v4hn: We (sadly) don't have C++20, or we could improve so much code with Ranges... :smile: 

I don't see why I should use a rb-tree for a list of 4 elements, so I will keep this as is.

moveit_planners/chomp/chomp_optimizer_adapter/src/chomp_optimizer_adapter.cpp
v4hn: done

henningkayser (approved):

PR #2541, opened 2021-03-08
Warn empty state, resolves #2517

### LINKED ISSUE #2517: Make error message in PositionIKRequest less confusing
### Description

As described in [this ROS Answers post](https://answers.ros.org/question/371571/found-empty-jointstate-message/), calling the `compute_ik` service with an empty `robot_state` field displays the error `Found empty JointState message` because the field is parsed in [_robotStateMsgToRobotStateHelper](http://docs.ros.org/en/api/moveit_core/html/conversions_8cpp_source.html#l00375) after passing through [here](https://docs.ros.org/en/kinetic/api/moveit_ros_move_group/html/kinematics__service__capability_8cpp_source.html#l00134). Something like "No hint for IK provided - starting from current robot state" should be displayed before the IK is called, imo.

I believe the error message normally appears when an empty message arrives on `/joint_states`, and that behavior should persist. But I can't find where exactly it happens (presumably in `planning_scene_monitor`/`current_state_monitor`), so I might just keep the error print in `utils` and add the extra information.

This would be a good first-timer issue since it only requires small changes in a few files.

### Steps to reproduce

Call the `compute_ik` service with an empty `robot_state` field in the `PositionIKRequest` message.

FabianSchuetze:
### Description
The PR tries to address #2517 

I added a warning upon calling [`computeIKService`](https://github.com/FabianSchuetze/moveit/blob/30a170f224c37fc1f28de1640b8da0b3afef6614/moveit_ros/move_group/src/default_capabilities/kinematics_service_capability.cpp#L145) with a [`GetPositionIK::Request` message](http://docs.ros.org/en/hydro/api/moveit_msgs/html/msg/PositionIKRequest.html) without explicitly setting the robot state. The PR warns if the passed `joint state` is empty since this element seems to be the crucial ingredients to me. The error message mentioned in the [ROS-Answers question](https://answers.ros.org/question/371571/found-empty-jointstate-message/) is caused by the following check in [_robotStateMsgToRobotStateHelper](https://github.com/FabianSchuetze/moveit/blob/30a170f224c37fc1f28de1640b8da0b3afef6614/moveit_core/robot_state/src/conversions.cpp#L373):
```c++
  if (!rs.is_diff && rs.joint_state.name.empty() && rs.multi_dof_joint_state.joint_names.empty())
  {
    ROS_ERROR_NAMED(LOGNAME, "Found empty JointState message");
    return false;
  }
```

We could also duplicate this check in the service call, but I am not sure if this duplication is valuable. 

I might have misunderstood the issue and therefore implemented the wrong patch. I am grateful for any hints or suggestions for how to improve the PR! 

### Checklist

### DIFF
--- a/moveit_ros/move_group/src/default_capabilities/kinematics_service_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/kinematics_service_capability.cpp
@@ -145,6 +145,10 @@ void MoveGroupKinematicsService::computeIK(moveit_msgs::PositionIKRequest& req, 
 bool MoveGroupKinematicsService::computeIKService(moveit_msgs::GetPositionIK::Request& req,
                                                   moveit_msgs::GetPositionIK::Response& res)
 {
+  if (req.ik_request.robot_state.joint_state.position.empty())
+  {
+    ROS_WARN("Empty joint_state position in GetPositionIK::Request msgs. Starting from the current robot state");
+  }
   context_->planning_scene_monitor_->updateFrameTransforms();
 
   // check if the planning scene needs to be kept locked; if so, call computeIK() in the scope of the lock


moveit_ros/move_group/src/default_capabilities/kinematics_service_capability.cpp
```diff
@@ -145,6 +145,10 @@ void MoveGroupKinematicsService::computeIK(moveit_msgs::PositionIKRequest& req,
 bool MoveGroupKinematicsService::computeIKService(moveit_msgs::GetPositionIK::Request& req,
                                                   moveit_msgs::GetPositionIK::Response& res)
 {
+  if (req.ik_request.robot_state.joint_state.position.empty())
+  {
+    ROS_WARN("Empty joint_state position in GetPositionIK::Request msgs. Starting from the current robot state");
```
simonschmeisser, line 150:
```suggestion
    ROS_WARN_NAMED(getName(), "Empty joint_state position in GetPositionIK::Request msgs. Starting from the current robot state");
```
This way messages can be filtered by category and verbosity be selectively reduced.

simonschmeisser (approved):

v4hn (changes requested):
I strictly prefer #2544 over this approach.

FabianSchuetze:
I think #2544 is indeed preferable! Thanks for pointing this out!

felixvd:
Closing because it seems we are in agreement

PR #2544, opened 2021-03-10
Fix missing isEmpty check in compute_ik service

### LINKED ISSUE #2517: Make error message in PositionIKRequest less confusing

v4hn:
Everything works as expected here, if you specify an empty RobotState it is a common pattern in MoveIt to use the current state instead.

The error message is a leftover artifact from #499.

Closes #2517.

### DIFF
--- a/moveit_ros/move_group/src/default_capabilities/kinematics_service_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/kinematics_service_capability.cpp
@@ -75,7 +75,10 @@ void MoveGroupKinematicsService::computeIK(moveit_msgs::PositionIKRequest& req, 
   const moveit::core::JointModelGroup* jmg = rs.getJointModelGroup(req.group_name);
   if (jmg)
   {
-    moveit::core::robotStateMsgToRobotState(req.robot_state, rs);
+    if (!moveit::core::isEmpty(req.robot_state))
+    {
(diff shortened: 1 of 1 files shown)


felixvd (approved):
This actually makes more sense than what I suggested. LGTM.

PR #2547, opened 2021-03-10
Python Bindings for some of moveit_core

PeterMitrano:
### Description

Add python bindings for PlanningScene, complete enough to make most of the PlanningScene tutorial work in python. See #900 for earlier discussion. I can make a PR with the new tutorial, but for that to make sense we'd need to figure out how/where to generate docs for these bindings. For now you can just run the tutorial here to test the bindings: https://github.com/UM-ARM-Lab/moveit_tutorials/blob/python-bindings/doc/planning_scene_python/src/planning_scene_tutorial.py

This hopefully sets the stage for an overhaul of moveit_commander as well.

### Checklist


moveit_core/CMakeLists.txt
```diff
@@ -226,6 +231,32 @@ add_subdirectory(collision_distance_field)
...
+  moveit_collision_detection
+  moveit_kinematic_constraints
+  moveit_planning_scene
+  moveit_python
+  moveit_robot_model
+  moveit_robot_state
+  moveit_transforms
+)
+
+if(pybind11_catkin_FOUND)
```
rhaschke, line 244: pybind11 was declared as required dependency above. No need for this if. But doesn't harm as well.

moveit_core/CMakeLists.txt
```diff
@@ -226,6 +231,32 @@ add_subdirectory(collision_distance_field)
...
+  add_subdirectory(python)
+
+  pybind_add_module(pymoveit
```
rhaschke, line 247: I'm not convinced yet about this structuring: Shouldn't we have different libs/modules for moveit_core, moveit_ros (MGI), etc? I'm missing the name `moveit_core` here.

moveit_core/collision_detection/src/pycollision_detection.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <moveit/collision_detection/world.h>
+#include <moveit/collision_detection/collision_matrix.h>
+#include <moveit/collision_detection/collision_common.h>
+#include <moveit/python/pybind_rosmsg_typecasters.h>
+
+namespace py = pybind11;
+
+using namespace collision_detection;
+
+auto SET_ENTRY1 = py::overload_cast<const std::string&, const std::string&, bool>(&AllowedCollisionMatrix::setEntry);
```
rhaschke, line 13: I prefer inlining this definition where required. It's only used once, isn't it?

moveit_core/collision_detection/src/pycollision_detection.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void def_collision_detect_bindings(py::module& m)
+{
+  m.doc() = "World represents the state of the collision objects, robot, and other environment information";
+  py::class_<World, WorldPtr>(m, "World").def(py::init<>());
+  py::class_<CollisionRequest>(m, "CollisionRequest")
+      .def(py::init<>())
+      .def_readwrite("contacts", &CollisionRequest::contacts)
```
rhaschke, line 21: For a future version of pybind11, you will need to define `Contact` first. Please reorder your `class_` definitions accordingly.

moveit_core/kinematic_constraints/src/pykinematic_constraint.cpp
```diff
@@ -0,0 +1,45 @@
...
+auto ADD1 = py::overload_cast<const moveit_msgs::Constraints&,  //
+                              const moveit::core::Transforms&>(&KinematicConstraintSet::add);
+
+auto DECIDE1 = py::overload_cast<const moveit::core::RobotState&, bool>(&KinematicConstraintSet::decide, py::const_);
+
+void def_kinematic_constraints_bindings(py::module& m)
+{
+  m.doc() = "Class for joint, position, visibility, and other constraints";
+  m.def("constructGoalConstraints",
+        CONSTRUCT_GOAL_CONSTRAINTS1,      //
```
rhaschke, line 28: Same here as above: Please embed the `py::overload_cast` directly. Otherwise, it doesn't become clear immediately, where the extra definitions belong to.

moveit_core/planning_scene/src/pyplanning_scene.cpp
```diff
@@ -0,0 +1,96 @@
...
+                                         const moveit_msgs::Constraints&,  //
+                                         const std::string&,               //
+                                         bool>(&PlanningScene::isStateValid, py::const_);
+auto IS_STATE_VALID5 = py::overload_cast<const moveit::core::RobotState&,                       //
+                                         const kinematic_constraints::KinematicConstraintSet&,  //
+                                         const std::string&,                                    //
+                                         bool>(&PlanningScene::isStateValid, py::const_);
+
+auto SET_CURRENT_STATE1 = py::overload_cast<const moveit_msgs::RobotState&>(&PlanningScene::setCurrentState);
+auto SET_CURRENT_STATE2 = py::overload_cast<const robot_state::RobotState&>(&PlanningScene::setCurrentState);
```
rhaschke, line 62: Do you really need both variants of methods (msg and C++ class) for this and all the above methods?
I think, the C++ class is converted to a msg anyway (via casters), isn't it?

moveit_core/planning_scene/src/pyplanning_scene.cpp
```diff
@@ -0,0 +1,96 @@
...
+      .def(py::init<const moveit::core::RobotModelConstPtr&, const collision_detection::WorldPtr&>(),
+           py::arg("robot_model"), py::arg("world") = collision_detection::WorldPtr(new collision_detection::World()))
+      .def("checkSelfCollision", CHECK_SELF_COLLISION1)
+      .def("checkSelfCollision", CHECK_SELF_COLLISION2)
+      .def("checkCollision", CHECK_COLLISION1)
+      .def("checkCollision", CHECK_COLLISION2)
+      .def("getCurrentStateNonConst", &PlanningScene::getCurrentStateNonConst)
+      .def("getCurrentState", &PlanningScene::getCurrentState)
+      .def("getAllowedCollisionMatrix", &PlanningScene::getAllowedCollisionMatrix)
+      .def("isStateConstrained", IS_STATE_CONSTRAINED1, py::arg("state"), py::arg("constr"), py::arg("verbose") = false)
```
rhaschke, line 78: If you really need all those method variants, here, I do see the readability improvement of using previously declared method overloads.

moveit_core/python/src/pymoveit.cpp
```diff
@@ -0,0 +1,55 @@
...
+  m.doc() = "moveit core";
+
+  auto collision_detection_m = m.def_submodule("collision_detection");
+  auto planning_scene_m = m.def_submodule("planning_scene");
+  auto robot_model_m = m.def_submodule("robot_model");
+  auto robot_state_m = m.def_submodule("robot_state");
+  auto transforms_m = m.def_submodule("transforms");
+  auto kinematic_constraints_m = m.def_submodule("kinematic_constraints");
+  auto urdf_m = m.def_submodule("urdf");
+  auto random_numbers_m = m.def_submodule("random_numbers");
```
rhaschke, line 33: I think, `urdf` and `random_numbers` should be part of their own module/lib to match the C++ lib structure.

moveit_core/python/src/pymoveit.cpp
```diff
@@ -0,0 +1,55 @@
...
+  def_planning_scene_bindings(planning_scene_m);
+  def_kinematic_constraints_bindings(kinematic_constraints_m);
+
+  urdf_m.def("parseURDFFile", &urdf::parseURDFFile);
+  py::class_<urdf::ModelInterface, urdf::ModelInterfaceSharedPtr>(urdf_m, "ModelInterface")
+      .def(py::init<>())
+      .def("getName", &urdf::ModelInterface::getName);
+
+  auto srdf_m = m.def_submodule("srdf", "srdf");
+  py::class_<srdf::Model>(srdf_m, "Model").def(py::init<>()).def("initFile", &srdf::Model::initFile);
```
rhaschke, line 54: These definitions aren't yet useful, are they? Is there any need to expose these classes?

moveit_core/robot_model/src/pyrobot_model.cpp
```diff
@@ -0,0 +1,41 @@
...
+  return robot_model;
+}
+
+void def_robot_model_bindings(py::module& m)
+{
+  m.doc() = "Definition of a kinematic model. Not thread safe, however multiple instances can be created.";
+  py::class_<RobotModel, RobotModelPtr>(m, "RobotModel")
+      .def(py::init<const urdf::ModelInterfaceSharedPtr&, const srdf::ModelConstSharedPtr&>(), py::arg("urdf_model"),
+           py::arg("srdf_model"))
+      .def("getURDF", &RobotModel::getURDF)
```
rhaschke, line 26: Why do you want to expose the URDF model, if you cannot access its details?

moveit_core/robot_state/src/pyrobot_state.cpp
```diff
@@ -0,0 +1,89 @@
...
+                                                    >(&RobotState::setJointGroupPositions);
+
+auto SATISFIES_BOUNDS1 = py::overload_cast<const JointModelGroup*, double>(&RobotState::satisfiesBounds, py::const_);
+
+auto ROBOT_STATE_MSG_TO_ROBOT_STATE1 = py::overload_cast<const Transforms&,               //
+                                                         const moveit_msgs::RobotState&,  //
+                                                         RobotState&,                     //
+                                                         bool                             //
+                                                         >(robotStateMsgToRobotState);
+auto ROBOT_STATE_MSG_TO_ROBOT_STATE2 = py::overload_cast<const moveit_msgs::RobotState&,  //
```
rhaschke, line 32: Couldn't this be handled by an appropriate type caster?

moveit_core/setup.py
```diff
@@ -0,0 +1,11 @@
+## ! DO NOT MANUALLY INVOKE THIS setup.py, USE CATKIN INSTEAD
+
+from distutils.core import setup
+from catkin_pkg.python_setup import generate_distutils_setup
+
+d = generate_distutils_setup(
+    packages=['moveit_core'],
+    package_dir={'': 'planning_scene'}
```
rhaschke, line 8: Why are you referring to `planning_scene` folder here?

moveit_core/transforms/src/pytransforms.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+namespace py = pybind11;
+using namespace moveit::core;
+
+void def_transforms_bindings(py::module& m)
+{
+  m.doc() = "Provides an implementation of a snapshot of a transform tree that can be easily queried for transforming "
+            "different quantities. Transforms are maintained as a list of transforms to a particular frame. All stored "
+            "transforms are considered fixed.";
+  py::class_<Transforms, TransformsPtr>(m, "Transforms").def(py::init<const std::string&>())
```
rhaschke, line 15: Not useful yet. Do you really need this?

rhaschke (commented):
Thanks a lot for this effort. I have a few inline remarks (see below) that apply across the whole PR:
- Add a copyright disclaimer to all files
- Avoid extra definitions of overload_casts, where it doesn't improve readability

moveit_core/CMakeLists.txt
PeterMitrano: I thought it better to make it optional at least to start, so I'm going to remove the "required" rather than the if. But I think either way is totally reasonable

moveit_core/CMakeLists.txt
PeterMitrano: I sort of agree, but I'm not sure. I want to go with whatever is the most obvious/expected thing for a user.

C++ users would expect to be able to translate includes, for example. `<moveit/core/kinematic_constraints/utils.h>`. how should this translate to an import statement? I'd say it should probably by `from moveit.core.kinematic_constraints import utils` ?
Then let's take move group, you might turn `<moveit/move_group/move_group_capability.h>` into `from moveit.move_group import move_group_capability`?

But there's also an argument for using the package name, i.e `moveit_core`. ROS python users would certainly expect that, so something like `from moveit_core.kinematic_constraints import utils` and `from moveit_ros.move_group import move_group_capability`.

Either approach is ok I think, but we do need to figure out how we're going to do documentation generation for these bindings.

moveit_core/collision_detection/src/pycollision_detection.cpp
PeterMitrano: I did this because there are many more overloads that presumably we should implement, I just didn't do all of them right now.

PeterMitrano: Also one thing I struggled a lot with is formatting these overload definitions, they're really long so putting them up here is actually nicer for formatting.

moveit_core/python/src/pymoveit.cpp
PeterMitrano: True, as is they are useless. I'm expecting they will be exposed further as people need them. I don't see much benefit in deleting it, although separating into separate PRs could be reasonable if your concern is diff size.

moveit_core/setup.py
PeterMitrano: simple mistake, good catch!

moveit_core/CMakeLists.txt
PeterMitrano: also fully acknowledging I did neither of these approaches here, what I did is definitely bad. What do folks think is the best structure?

moveit_core/robot_state/src/pyrobot_state.cpp
PeterMitrano: true, although actually I think we need to somehow explicitly prevent/delete that overload. It doesn't work, and in general we should prohibit calling function that takes a ros message by mutable reference, because the mutation won't make it back to the python caller.

PeterMitrano:
A few points we should discuss in general:
1. Which python version are we supporting? Here I've written with/for python 3.8 because that's the default on 20.04 and I think most noetic users use 3.8
2. How do we generate documentation?
3. What next? I'm happy to continue (slowly) implementing this stuff, are there any strong opinions on where to go next?

moveit_core/robot_model/src/pyrobot_model.cpp
PeterMitrano: same point about "yes it's useless, but why not"

moveit_core/transforms/src/pytransforms.cpp
PeterMitrano: same point about "yes it's useless, but why not"

moveit_core/kinematic_constraints/src/pykinematic_constraint.cpp
PeterMitrano: Sure I can do that. My reason for pulling them out was formatting. It becomes pretty unreadable with them defined inline

moveit_core/collision_detection/src/pycollision_detection.cpp
PeterMitrano: Ok I think I've address this

moveit_core/CMakeLists.txt
PeterMitrano: Another possibly important note here -- if we ever want to add pure python to the package (whatever it ends up being called), we might want to set that up now. Apparently there's a common convention for doing that: https://github.com/pybind/pybind11/issues/1004#issuecomment-322941844 which I think would work in this case as well.

rhaschke: I prefer the 2nd proposed naming scheme, i.e. `from moveit_core.kinematic_constraints import utils`.

> If we ever want to add pure python to the package, we might want to set that up now.

I agree with the referenced scheme. I'm using the same approach in MTC.

moveit_core/python/src/pymoveit.cpp
rhaschke: My main concern was that urdf, srdf, random_numbers are not parts of MoveIt, but have their own ROS packages.

moveit_core/robot_state/src/pyrobot_state.cpp
rhaschke: Yeah, that's another concern. Generally, of course, it is possible to modify passed arguments.

moveit_core/transforms/src/pytransforms.cpp
rhaschke: Reply to all instances: It suggests a functionality that isn't there (yet).

rhaschke:
> 1. Which python version are we supporting? Here I've written with/for python 3.8 because that's the default on 20.04 and I think most noetic users use 3.8

pybind11 is rather agnostic regarding the python version. So, why do you see a need to restrict to a specific version?
If really needed, I would target 20.04's python version as well.

> 2. How do we generate documentation?

Do you want to have a doxygen-like API doc?

moveit_core/CMakeLists.txt
PeterMitrano: great, than I will implement that and update. If anyone else has thoughts on this please chime in!

moveit_core/python/src/pymoveit.cpp
PeterMitrano: that's true. I think the initial impetus was to expose the constructor for RobotModel which takes a URDF and SRDF, and therefore we needed those classes exposed and usable, but then I decided to do that in C++ and write a little wrapper function that just takes the paths for the urdf and srdf filenames. Since the more common way will be to use the RobotModelLoader from the ros interface, which will eventually have python bindings, I don't think people will be constructing a RobotModel directly.

random_numbers is going to be needed by anyone who wants to control RNG when calling setToRandomPosition. Again maybe this isn't super common, but I felt like it was good to expose it anyways.

Of course the better approach would be to go write bindings in those packages, which we could certainly do if the maintainers there are interested.

Those are just my thoughts, of course it's up to you all, let me know if you'd like me to remove these all the not-actually-moveit bindings.

moveit_core/transforms/src/pytransforms.cpp
PeterMitrano: ok, I will remove them from this PR

moveit_core/robot_state/src/pyrobot_state.cpp
PeterMitrano: > > 1. Which python version are we supporting? Here I've written with/for python 3.8 because that's the default on 20.04 and I think most noetic users use 3.8
> 
> pybind11 is rather agnostic regarding the python version. So, why do you see a need to restrict to a specific version?
> If really needed, I would target 20.04's python version as well.
I guess I was thinking about this for any pure python code we include, which so far was just the tutorial I was working on. So actually for this PR we don't need to worry about this. I'll make a PR to the tutorials repo and there it may be relevant.

> 
> > 1. How do we generate documentation?
> 
> Do you want to have a doxygen-like API doc?

Sure, something like that. moveit_commander API docs, and those are fine. I just think we need some kind of online docs people can browse, and that we can link to in whatever tutorials may be written for using the python bindings.

PeterMitrano:
> > 1. Which python version are we supporting? Here I've written with/for python 3.8 because that's the default on 20.04 and I think most noetic users use 3.8
> 
> pybind11 is rather agnostic regarding the python version. So, why do you see a need to restrict to a specific version?
> If really needed, I would target 20.04's python version as well.
I guess I was thinking about this for any pure python code we include, which so far was just the tutorial I was working on. So actually for this PR we don't need to worry about this. I'll make a PR to the tutorials repo and there it may be relevant.

> 
> > 1. How do we generate documentation?
> 
> Do you want to have a doxygen-like API doc?

Sure, something like that. moveit_commander API docs, and those are fine. I just think we need some kind of online docs people can browse, and that we can link to in whatever tutorials may be written for using the python bindings.

rhaschke:
I think [moveit_commander's API docs](http://docs.ros.org/en/noetic/api/moveit_commander/html/) are generated by rosdoc from the python code. sphinx can also auto-generate docs from pybind11 bindings:
https://pybind11.readthedocs.io/en/stable/advanced/misc.html#generating-documentation-using-sphinx

moveit_core/CMakeLists.txt
PeterMitrano: I'm been implementing this and I have a more specific decision/question... In the exmples I gave, `from moveit_core.kinematic_constraints import utils` the final name `utils` corresponds to a C++ header file. However, I'm wondering if this is unnecessary structure that will make the python code verbose. So keeping with that example, the choice is between:
```
from moveit_core.kinematic_constraints import utils
utils.constructGoalConstraints(...)
from moveit_core.kinematic_constraints.utils import constructGoalConstraints # also works
```
VS
```
from moveit_core import kinematic_constraints
kinematic_constraints.constructGoalConstraints(...)
```


To give another example, let's take the collision_detection namespace. Which do you prefer?
```
from moveit_core.collision_detection import collision_common
from moveit_core.collision_detection import world
world.World(...)
collision_common.CollisionResult(...)
```
VS
```
from moveit_core import collision_detection
collision_detection.World(...)
collision_detection.CollisionResult(...)
```

in C++ obviously the name of the header file doesn't appear in the name of the type, only the include path. So if we want we could drop it, which would make the python docs look more like the C++ docs. If we go the first route, we'd have things like moveit_core.planning_scene.planning_scene.PlanningScene when we could have  moveit_core.planning_scene.PlanningScene. Another advantage is it makes the bindings code shorter/simpler if we drop the distinction between header files and keep it at the namespace level.

moveit_core/robot_state/src/pyrobot_state.cpp
PeterMitrano: Do you know if we can  make sure this is not accidentally allowed? I thought I could just change the enable_if statement, so that the specialization didn't match if it was a ref to non-const, but I didn't work. This is what I tried:
```
struct type_caster<T, std::enable_if_t<ros::message_traits::IsMessage<T>::value &&
                                                             (std::is_const<std::remove_reference<T>>::value || !std::is_reference<T>::value)>>
```
but this still allows me to call a function that takes a ros message by ref to non-const, which is what I want to avoid.

PeterMitrano:
> I think [moveit_commander's API docs](http://docs.ros.org/en/noetic/api/moveit_commander/html/) are generated by rosdoc from the python code. sphinx can also auto-generate docs from pybind11 bindings:
> https://pybind11.readthedocs.io/en/stable/advanced/misc.html#generating-documentation-using-sphinx

Thanks for the tip. I've got this working ok, but I'm not sure where this will go or how to PR it. It's going to involve putting a handful of config/text files + Makefile somewhere, and then there will be build folders where HTML is generated. So where should that go, and how would that get hosted somewhere? We can leave all this out for this PR, which is already big, but just asking for next steps.

rhaschke:
To (automatically) install API docs via `rosdoc_lite`, please consult the wiki http://wiki.ros.org/rosdoc_lite. Can you link the config files you used?

moveit_core/CMakeLists.txt
rhaschke: I prefer the second variant as well. Usually, in C++ we have one class per header file. Hence, there is no need to name the header file once again.

moveit_core/python/src/pymoveit.cpp
PeterMitrano: I've removed all the not-actually-moveit bindings

PeterMitrano:
> To (automatically) install API docs via `rosdoc_lite`, please consult the wiki http://wiki.ros.org/rosdoc_lite. Can you link the config files you used?

Here's what I've got:

https://github.com/UM-ARM-Lab/moveit/tree/docs/moveit_core/doc

If you run `make html` in this folder, it should generate everything into `_build` and `_autosummary` folders, perhaps a few others.

Thanks for the link, I had no idea ros how the rosdoc stuff worked. I added a rosdoc.yaml file listing doxygen and sphinx and it seems to work locally! This is on the doc branch I linked to, so check it out and let me know if I should include the docs configuration stuff in this PR.

PeterMitrano:
I want to add the copyright statements now, but I'm not sure what to put. Do I use my name? or PickNik? or something else?

```
/*********************************************************************
 * Software License Agreement (BSD License)
 *
 *  Copyright (c) 2021, Peter Mitrano
 *  All rights reserved.
 *
 *  Redistribution and use in source and binary forms, with or without
 *  modification, are permitted provided that the following conditions
 *  are met:
 *
 *   * Redistributions of source code must retain the above copyright
 *     notice, this list of conditions and the following disclaimer.
 *   * Redistributions in binary form must reproduce the above
 *     copyright notice, this list of conditions and the following
 *     disclaimer in the documentation and/or other materials provided
 *     with the distribution.
 *   * Neither the name of Peter Mitrano nor the names of its
 *     contributors may be used to endorse or promote products derived
 *     from this software without specific prior written permission.
 *
 *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
 *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
 *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
 *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
 *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
 *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
 *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
 *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
 *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
 *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 *  POSSIBILITY OF SUCH DAMAGE.

/* Author: Peter Mitrano */

```

rhaschke:
Of course, you should use your name. It's your work! Please use my name for the ROS msg type caster, as this is copied 1:1 from MTC, isn't it?

PeterMitrano:
Ok, and yes I will put your name on those files.

moveit_core/python/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+set(MOVEIT_LIB_NAME moveit_python)
```
rhaschke, line 1: In MTC I have used `moveit_python_tools` as lib name.

rhaschke (commented):
This looks quite good already. 

> I added a rosdoc.yaml file listing doxygen and sphinx and it seems to work locally! 

Great. I tried myself for MTC, but it doesn't yet work: the API summaries for my pybind11 modules are empty.

> Let me know if I should include the docs configuration stuff in this PR.

Yes, please. But, I think you don't need to have a Makefile for the docs. rosdoc_lite (which is run by the ROS buildfarm) should do the job.

moveit_core/python/CMakeLists.txt
PeterMitrano: yea I like that better as well, I'll change it

mikeferguson: I think you also want to avoid moveit_python as a name, because there is also a (released) ROS package called [moveit_python](https://github.com/mikeferguson/moveit_python) and the imports might get clouded for people who have that installed

rhaschke (approved):
One more nitpick. Otherwise lgtm.

rhaschke:
Note, that there is still a [clang-format issue reported from Travis](https://travis-ci.com/github/ros-planning/moveit/jobs/491620028).
Also, do you already install `pymoveit_core.so`?

moveit_core/src/moveit_core/__init__.py
```diff
@@ -0,0 +1,2 @@
+from pymoveit_core import *
+
```
rhaschke, line 2: You shouldn't include the library here, but the local modules:
```suggestion
from . import collision_detection, kinematic_constraints, planning_scene, robot_model, robot_state, transforms
```
Otherwise, local extensions (in python code) don't become available. You don't yet have them, but for the future...

rhaschke (changes requested):
I found a more severe issue: local (python) extensions don't become visible right now.

moveit_core/src/moveit_core/__init__.py
PeterMitrano: I'm not sure, I think the line you suggest adding does nothing? And in my local testing we need the line `from pymoveit_core import *`, because otherwise we don't get any of the symbols to `pymoveit_core`.  For example try running `from moveit_core import load_robot_model`. As is this works, but with what you propose it does not. If you have examples that do/don't work that might help me understand.

I also tested adding `foo.py` to `moveit_core/src/moveit_core` with a function `test`, I can write `from moveit_core import foo` and `foo.test()` works fine. I also tested this in the submodules.

rhaschke: For the examples you gave, my suggested line is not important, but for this it is:
```python
import pymoveit_core as core
core.robot_model.RobotModel()  # without, you couldn't access the contents of core here
```
The symbols from the (library) submodules are loaded via `<submodule>/__init__.py`:
```python
from pymoveit_core.<submodule> import *
# without the proposed change, you wouldn't be able to augment / override methods in a submodule like this:
def added_to_submodule():
    pass
```
Adding a new submodule `foo`, of course, works in any case.

PeterMitrano: My thought was no one need reference pymoveit_core directly. you can just write moveit_core and get both the C++ bindings and any python code we add. This works, and I think is the same as what you wrote

```
import moveit_core
moveit_core.robot_model.RobotModel
```

I'm not sure what you mean by override. Like this? this works:

```
from moveit_core.kinematic_constraints import *                                                                                                                 

 constructGoalConstraints # original method

def constructGoalConstraints():
    print("you're mine now")
```

rhaschke: I need to test again. I'm currently working on MTC python bindings and there, I needed the suggested changes...
I will postpone that for tomorrow.

rhaschke: I tested and now I remember the exact open issue: With your approach you will not be able to access augmented python code like this:
```python
import moveit_core as core
core.robot_model.added_to_submodule()
```
Sorry for pointing initially into the wrong direction. As I prepared the fixup for testing anyway, you can find a corresponding PR [here](https://github.com/UM-ARM-Lab/moveit/pull/1).

rhaschke:
I'm inclined to slightly restructure the package layout. What about placing everything into a moveit namespace and have below that packages for each `moveit_*` ROS package, but dropping the (redundant) `moveit_`? That is:
```python
from moveit.core import robot_model, robot_state
m = robot_model.RobotModel()
```

felixvd:
That looks perfectly fine to me. As long as the convention we choose is easy to follow and remember, I don't mind which one it ends up being. I am hoping to read and try this out on the weekend.

PeterMitrano:
> I'm inclined to slightly restructure the package layout. What about placing everything into a moveit namespace and have below that packages for each `moveit_*` ROS package, but dropping the (redundant) `moveit_`? That is:
> 
> ```python
> from moveit.core import robot_model, robot_state
> m = robot_model.RobotModel()
> ```
> 

I'm glad you bring this up. I started looking into the other packages, and yea the naming gets weird. I'm a bit confused by the directory structure of moveit, because folder and namespace names don't match the library names (why is the package called moveit_ros_planning_interface, but the namespace is moveit::planning_interface, but the path is moveit_ros/planning_interface?) ... so take what I say here with a lot of salt.

But extending your proposal to another package, I think that would mean

```
from moveit import planning_interface 
planning_interface.MoveGroupInterface
planning_interface.PlanningSceneInterface
```

That looks nice, but then it turns everything into one module, which means one `pymoveit.so` file and the call to `pybind_add_module` in the top level CMakeLists, I think? Should be fine, but there may be some surprises.

If we stick with the current approach, which I do think is ok, then we'd be writing
```
from moveit_ros_planning_interface import planning_interface 
planning_interface.MoveGroupInterface
planning_interface.PlanningSceneInterface
```

I think it's fine either way, but I'm not totally sure implementing what you propose is straightforward. I'm happy to prototype it and see though!

rhaschke:
> That looks nice, but then it turns everything into one module, which means one `pymoveit.so` file and the call to `pybind_add_module` in the top level CMakeLists, I think? Should be fine, but there may be some surprises.

No, you can have separate C++ extension libs for each moveit package. I drafted the restructuring in https://github.com/UM-ARM-Lab/moveit/pull/1/commits/4a75ab6c8e6b63e97efff9c7eb155fdf1c558373
The appeal of the proposed approach is to have a global, unifying `moveit` package.

moveit_core/src/moveit_core/__init__.py
PeterMitrano: I've updated it to use `moveit.core`, the install works, and it is compatible with other packages, so when we add bindings for moveit_ros_planning for example, it should become something like `from moveit.planning import robot_model_loader`

felixvd (commented):
Thanks for this! I tried to get this latest version to run in Melodic, but had issues. The commits to reproduce the errors are on [this branch](https://github.com/felixvd/moveit_tutorials/tree/python-bindings):

- random_numbers isn't a part of `moveit.core`
- If I comment out the rng lines, I get errors about the `robot_model` module not being loaded correctly, it seems:

```
File "/root/linked_ws_moveit/src/moveit_tutorials/doc/planning_scene_python/src/planning_scene_tutorial.py", line 34, in main
    kinematic_model = robot_model.load_robot_model(urdf_path, srdf_path)
AttributeError: 'module' object has no attribute 'load_robot_model'
```

I left minor comments for now. Some other thoughts:

- What is missing for Tests 5 and 11 in the tutorial, which display the contact points?
- Is there a reason to keep the [RobotCommander class](https://github.com/ros-planning/moveit/blob/master/moveit_commander/src/moveit_commander/robot.py#L42)? Can we replace it with the new `RobotState`?
- Do we accept that the Python functions are camelCase and not snake_case? I would say yes.

PeterMitrano:
I've updated that tutorial, it should work now. 

https://github.com/UM-ARM-Lab/moveit_tutorials/blob/python-bindings/doc/planning_scene_python/src/planning_scene_tutorial.py

> * What is missing for Tests 5 and 11 in the tutorial, which display the contact points?
For test 5 I just moved the printing to test 1, I had issues with the RNG not giving contacts, not really sure what was happening there but it didn't seem like bindings issue. 
Test 11 was removed because at the time I hadn't figured out how to make a binding for it. I could probably go back and sort that out.

> * Is there a reason to keep the [RobotCommander class](https://github.com/ros-planning/moveit/blob/master/moveit_commander/src/moveit_commander/robot.py#L42)? Can we replace it with the new `RobotState`?
For this PR I don't want to delete anything, and I have no idea if it's a complete replacement yet. Some people might be upset if we just remove robot commander, so perhaps we can re-write it as a simple wrapper around these new bindings?

> * Do we accept that the Python functions are camelCase and not snake_case? I would say yes.

Actually I hadn't thought about it. There's no reason AFAIK that we can't use snake_case, which is what PEP8 says. That said I don't think it matters much.

moveit_core/doc/index.rst
```diff
@@ -0,0 +1,12 @@
+moveit_core python
+=======================================
+
+.. note::
+    These docs are autogenerated. Whenever you see the py prefix, you can ignore it.
+    For example, you should write ``import moveit_core`` instead of ``import pymoveit_core``.
```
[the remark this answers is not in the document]
PeterMitrano: True, and actually this re-organization seems to have broken documentation generation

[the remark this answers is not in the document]
PeterMitrano: ok got docs working again, actually it's much better now. Still pretty unreadible though :shrug:

PeterMitrano:
I would appreciate help with the travis errors:

```
moveit_core: CMakeLists.txt: error: unconfigured build_depend on 'pybind11_catkin'
moveit_core: CMakeLists.txt(260): error: referenced target 'pymoveit_core' is not defined
```

Some googling tells me the first one might be changed by making pybind11_catkin an exec/run depend but that seems very wrong. Also, I have no idea how the second error is happening. If it's entering that if statement I don't see how it pymoveit_core could be undefined.

PeterMitrano:
I'd love to keep this moving forward. Other than the travis error, are there any futher requested changes?

PeterMitrano:
Ok I've figured out the travis errors.

We get the first one because I have this listed as an optional dependency, and apparently that doesn't count in the linter. Making it an `exec_depend` does fix the problem, but is conceptually wrong I think.

The second one is because the target pymoveit_core is defined in a macro and so the linter can't see it. We could add `--ignore undefined_target` for moveit_core, to suppress the problem. Not sure if there's a better way to fix it.

tylerjw:
To get this to pass the `Format` check you are going to need the tool `pre-commit`.  I've written instructions for how to use it here: https://github.com/ros-planning/moveit2/wiki/pre-commit

PeterMitrano:
ok will do -- where do I get the `.pre-commit-config.yaml` file for using that tool? is using that tool any different from just running clang-format?

tylerjw:
The test it is failing looks to be trailing whitespace and `black` (clang-format like formatter for python).  The .pre-commit-config.yml should just be in base of the repo (I'm assuming you rebassed on master and that's how the test is now running).  You can just install pre-commit on your machine like this:
```
pip3 install pre-commit
```

Then in the moveit repo just run:

```
pre-commit run -a
```

That should fix all the formatting errors that CI is finding.  Then just add those changes and commit them and the test will start passing.  Once you've done that run:

```
pre-commit install
```

Then in the future when you run `git commit` it'll automatically run (as it installs itself as a pre-commit git hook) and fix formatting errors.  If it reports errors it'll stop the commit action and you'll have to do an add and commit again.

PeterMitrano:
I'm clearly doing something wrong, it isn't working for me and I'm not sure how it could:

```

**Comment by PeterMitrano** [comment]
wow ok I figured it out. First of all yaml != yml, and second this is a recent change and I should be getting the file from master by rebasing. it works now

**Comment by tylerjw** [comment]
`end-of-file-fixer` and `black` is still failing, run `pre-commit run -a` to fix these errors.

**Comment by tylerjw** [comment]
Deleted my comment, looks like you got it figured out.  I'm sorry this was so much trouble.  The whole point of this new formatter system was to make it much easier to run the auto-formatters so people wouldn't comment on PRs about them and instead focus on reviewing the substance of the PR.  This was also because people constantly seemed confused about how to run clang-format manually.

**Comment by PeterMitrano** [comment]
oh I agree it will be easier! it's just because I started this PR before you made that change, and didn't realize it was a new thing and I had to rebase.

**Comment by PeterMitrano** [comment]
Still hoping for feedback on how to resolve the catkin linting errors -- here's my understanding:

We get the first one because I have this listed as an optional dependency, and apparently that doesn't count in the linter. We can (1) make it an exec_depend does fix the problem or (2) make these bindings required and not optional  ?

The second one is because the target pymoveit_core is defined in a macro and so the linter can't see it. We could add --ignore undefined_target for moveit_core, to suppress the problem. Not sure if there's a better way to fix it.

**Comment by tylerjw** [comment]
I'm sorry if this has been explained before but why would we want the python bindings to be optional?  Does it greatly increase the build time?

**Comment by PeterMitrano** [comment]
It hasn't been discussed, I was just trying to be cautious with introducing it

**Review by rhaschke — CHANGES_REQUESTED** [review]
I have added two more commits: https://github.com/UM-ARM-Lab/moveit/pull/2.
- `moveit` should be declared as a proper namespace package
   Note that this conflicts with `from . import core` to allow 
   ```python
   import moveit
   moveit.core.xxx
   ```
- Making the python bindings non-optional also silences catkin_lint warnings.

**Comment by PeterMitrano** [comment]
yay all the checks have passed! I think this might be ready to go?

**Comment by rhaschke** [comment]
@PeterMitrano, do you mind squash-merging this as a single commit? If not, the long history of commits should be cleaned up...
@felixvd, could you have a look and provide a 2nd review?

**Comment by PeterMitrano** [comment]
I believe github has an option to squash on merge? that way we keep the history in my branch, but avoid having it on master

**Comment by rhaschke** [comment]
Sure. I just wanted to know if you accept squashing into a single commit. Seems so.

**Comment by PeterMitrano** [comment]
oh yea I would highly prefer that

**Review by felixvd — APPROVED** [review]
Love it. Almost the only thing that's missing for me is `setFromIK`. Cannot wait to see this and the moveit_tutorials branch merged.

I saw that you have a branch with the octomap bindings. That's also a good next step. With both, these bindings will actually be really useful.

File: moveit_core/doc/conf.py [hunk 17]
```diff
@@ -0,0 +1,200 @@
...
+# (source start file, target name, title, author,
+#  dir menu entry, description, category)
+texinfo_documents = [
+    (
+        master_doc,
+        "moveit_core",
+        "moveit_core Documentation",
+        author,
+        "moveit_core",
+        "One line description of project.",
```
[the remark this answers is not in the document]
PeterMitrano: Thank you for your review! I was told earlier to put "moveit maintainer team" as the author, which makes sense. But yes I can update the description & category entries to something informative. I think all the target name and title are reasonable already, but I'm open to suggestions.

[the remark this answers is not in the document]
felixvd: Actually then it's fine, I just thought it was an oversight. I'm good with this.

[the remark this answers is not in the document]
PeterMitrano: Also that description and category don't show up in the docs, so yea I think we can leave this as is.

PR #2549, opened 2021-03-10
Add docker hub login to travis yaml

tylerjw:
### Description

Similar to this pull request: https://github.com/ros-planning/moveit2/pull/380

This uses secret environment variables (set within travis under the repo settings) to login to docker hub to increase our pull limit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ jobs:
 
 before_script:
   - test -n "$DOCKER_USERNAME" && docker login -u "$DOCKER_USERNAME" --password "$DOCKER_PASSWORD" 
-  - test -n "$DOCKER_USERNAME" || echo "Docker Hub credentials unavailable, pulling auonymously"
+  - test -n "$DOCKER_USERNAME" || echo "Docker Hub credentials unavailable, pulling anonymously"
   - git clone -q --depth=1 https://github.com/ros-planning/moveit_ci.git .moveit_ci
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,8 @@ jobs:
     - env: TEST=code-coverage
 
 before_script:
+  - test -n "$DOCKER_USERNAME" && docker login -u "$DOCKER_USERNAME" --password "$DOCKER_PASSWORD" 
+  - test -n "$DOCKER_USERNAME" || echo "Docker Hub credentials unavailable, pulling auonymously"
   - git clone -q --depth=1 https://github.com/ros-planning/moveit_ci.git .moveit_ci
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,8 +37,7 @@ jobs:
     - env: TEST=code-coverage
 
 before_script:
-  - test -n "$DOCKER_USERNAME" && docker login -u "$DOCKER_USERNAME" --password "$DOCKER_PASSWORD" 
-  - test -n "$DOCKER_USERNAME" || echo "Docker Hub credentials unavailable, pulling anonymously"
+  - if [ -n "$DOCKER_USERNAME" ] ; then docker login -u "$DOCKER_USERNAME" --password ; else echo "Docker Hub credentials unavailable, pulling anonymously" ; fi
   - git clone -q --depth=1 https://github.com/ros-planning/moveit_ci.git .moveit_ci
 
 script:


v4hn:
@tylerjw Thanks for looking into it!

I don't see anything setup the wrong way, but apparently the command is executed without the environment variables set. Any idea?

henningkayser:
In the [Travis documentation](https://docs.travis-ci.com/user/environment-variables#defining-variables-in-repository-settings) it says that "variables are not automatically available to forks" and that Travis does not "provide these values to untrusted builds, triggered by pull requests from another repository". Seems like Travis doesn't allow running the single-pro-user setup that we were planning, of course only because of security considerations. A quick google search produced [this article](https://www.algolia.com/blog/engineering/travis-encrypted-variables-external-contributions/) that describes a method for setting up a public "API key dealer" that verifies requests using the job id and IP range. Looks straight forward to me, I'm just not sure if this is really worth the hassle, actually secure enough (at least API keys can be set to provide limited access and disabled if needed) and if it violates any user terms with Dockerhub/Travis.

v4hn:
> In the Travis documentation it says that "variables are not automatically available to forks" and that Travis does not "provide these values to untrusted builds, triggered by pull requests from another repository".

That makes a lot of sense, but is also quite unfortunate.

The approach to setup such a key dealer seems fragile and complex. Also I'm not convinced docker actually would approve.
The best solution would be to get docker to allow any number of pulls from travis' IP range (or CircleCI if we migrate) in exchange for a subscription.
It seems like either our use is not at all an intended workflow or their planned workflow is broken if you can't reliably pull their images from CI anymore. :frowning: 

Is pulling from docker in CI jobs really such an uncommon use-case?

Alternatively, would it be feasible to store the images in cache and only pull them only sporadically?

tylerjw:
It looks like this is one of the issues that might just be solved by switching to CircleCI.  [This FAQ on CircleCI](https://support.circleci.com/hc/en-us/articles/360050623311-Docker-Hub-rate-limiting-FAQ) states that they have partnered with docker so their users don't experience this.  They do warn that this might continue forever and users should still try to login.  Their solution to do this is this:

```
if [ -n "$DOCKERHUB_USERNAME" ]; then
  docker login --username $DOCKERHUB_USERNAME --password $DOCKERHUB_PASSWORD
else
  echo "Docker Hub credentials unavailable, pulling anonymously"
fi
```

tylerjw:
Another feature of CircleCI is that it can pass secrets to builds done on forks.  This has security implications, however if we only use this dockerhub account for pulls and it has no special permissions with regard to the dockerhub org moveit and can't can't push I don't really see the harm in doing that.

.travis.yml
```diff
@@ -37,6 +37,8 @@ jobs:
     - env: TEST=code-coverage
 
 before_script:
+  - test -n "$DOCKER_USERNAME" && docker login -u "$DOCKER_USERNAME" --password "$DOCKER_PASSWORD" 
+  - test -n "$DOCKER_USERNAME" || echo "Docker Hub credentials unavailable, pulling auonymously"
```
rhaschke, line 41:
```suggestion
  - test -n "$DOCKER_USERNAME" || echo "Docker Hub credentials unavailable, pulling anonymously"
```

rhaschke (changes requested):

rhaschke (approved):

tylerjw:
I merged your change but it looks like this is failing in travis because the variable is undefined.  I probably need a `set +u` around it or something.  I'm not sure this is a good solution though.

.travis.yml
```diff
@@ -37,6 +37,8 @@ jobs:
     - env: TEST=code-coverage
 
 before_script:
+  - test -n "$DOCKER_USERNAME" && docker login -u "$DOCKER_USERNAME" --password "$DOCKER_PASSWORD" 
+  - test -n "$DOCKER_USERNAME" || echo "Docker Hub credentials unavailable, pulling anonymously"
```
rhaschke, line 41: If DOCKER_USERNAME is not defined, the `test` fails and thus makes Travis fail. Try this:
```suggestion
  - if [ -n "$DOCKER_USERNAME" ] ; then docker login -u "$DOCKER_USERNAME" --password ; else echo "Docker Hub credentials unavailable, pulling anonymously" ; fi
```

v4hn:
travis is gone.

PR #2550, opened 2021-03-11
Fixed flood of errors on startup for `mesh_filter`

JStech:
### Description

The mesh filter was producing a flood of errors when it started, which stopped after a subscriber connected to its output. This prevents these errors, and instead produces a trickle of debug messages. I cleaned up a few little things, too, while I was at it.

### Checklist

### DIFF
--- a/moveit_ros/perception/mesh_filter/src/transform_provider.cpp
+++ b/moveit_ros/perception/mesh_filter/src/transform_provider.cpp
@@ -39,7 +39,7 @@
 #include <tf2_ros/buffer.h>
 #include <tf2_eigen/tf2_eigen.h>
 
-TransformProvider::TransformProvider(unsigned long interval_us) : stop_(true), interval_us_(interval_us)
+TransformProvider::TransformProvider(double update_rate) : stop_(true), update_rate_(update_rate)
 {
   tf_buffer_ = std::make_shared<tf2_ros::Buffer>();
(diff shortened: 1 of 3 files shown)


simonschmeisser (approved):
Would there also be a way for it to not complain at all when it has not yet received data (as opposed to incorrect data)? (haven't looked at the underlying code)

JafarAbdi (approved):

JStech:
> Would there also be a way for it to not complain at all when it has not yet received data (as opposed to incorrect data)? (haven't looked at the underlying code)

I think you're describing exactly what I'm fixing. When it has not yet received a depth map, the `frame_id_` is the empty string.

moveit_ros/perception/mesh_filter/src/transform_provider.cpp
```diff
@@ -110,17 +110,24 @@ void TransformProvider::run()
   while (!stop_)
```
AndyZe, line 110: This should probably be:

`while (ros::ok() && !stop_)`

To avoid getting stuck in an infinite loop.

AndyZe (approved):
Approved provided you make the small change in the while loop

moveit_ros/perception/mesh_filter/src/transform_provider.cpp
JStech: Good catch--done

Issue #2486, opened 2021-01-13
Joint Limits not showing in the joint_limits.yaml


ly19264:
### Description

After creating the moveit package using moveit setup assistant, the joints_limits.yaml does not show the joint limits present in the URDF. Does moveit consider the joint limits if yes, then where does it store them, if no, do we manually have to add them to joints_limits.yaml

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04

Steps to reproduce
Creating a moveit package from moveit setup assistant

Expected behaviour
Joint Limits should be taken from the URDF and be displayed in the joint_limits.yaml

### Actual behaviour
The joints_limits.yaml only shows the velocity limit
joint_limits:
joint0:
has_velocity_limits: true
max_velocity: 100
has_acceleration_limits: false
max_acceleration: 0

simonschmeisser:
If you have position limits set for a joint in `joint_limits.yaml` then those will override the ones from URDF. If you don't specify them in `joint_limits.yaml` then the ones from URDF will be used.

So it appears to me that you would like to have the values from URDF copied to `joint_limits.yaml` for convenience? Please open a PR for it if that is what you intend. Don't hesitate asking questions for any problems you encounter!

The code is here: https://github.com/ros-planning/moveit/blob/664ae01803abf5e0b4649063102357262de9e05c/moveit_setup_assistant/src/tools/moveit_config_data.cpp#L1118

If you are super enthusiastic, you could of course also create a widget/page to adjust these

ly19264:
@simonschmeisser ,

Thanks for the help, a quick query ,

Does the moveit package stores the joint limits from the URDF or does it reads those limits from the URDF during execution ?

simonschmeisser:
The package you create by using MoveIt setup assistant contain launch files that load config files (including joint_limits.yaml) as well as the SRDF and URDF to the parameter server (rosparam). The URDF is typically part of a robot package that you can install (from source or deb). MoveIt (class `RobotModel`) first reads the joint limits from the URDF (from parameter /robot_description) and then overrides them with values from joint_limits.yaml (if there are any). Typically the URDF contains the hard limits from the manufacturer and joint_limits.yaml sets your local setup (if you don't need a certain rotation or have some cable that is not long enough etc.)

ly19264:
@simonschmeisser , Is there a way to fetch/read the joint limits from the moveit package during runtime. I would like to access those values to check the solution is in limit or not.

[...]

PR #2553, opened 2021-03-18
included position information in joint_limits.yaml

Sunny478:
### Description

Included the position limits from URDF to the joint_limits.yaml file in /config. Resolves issue #2486 

### Checklist

### DIFF
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -1171,6 +1171,14 @@ bool MoveItConfigData::outputJointLimitsYAML(const std::string& file_path)
     const moveit::core::VariableBounds& b = joint->getVariableBounds()[0];
 
     // Output property
+    emitter << YAML::Key << "has_position_limits";
+    emitter << YAML::Value << "false";
+
+    // Output property
+    emitter << YAML::Key << "max_position";
+    emitter << YAML::Value << std::min(fabs(b.max_position_), fabs(b.min_position_));
+    
+    // Output property
     emitter << YAML::Key << "has_velocity_limits";
     if (b.velocity_bounded_)
       emitter << YAML::Value << "true";


moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -1170,6 +1170,14 @@ bool MoveItConfigData::outputJointLimitsYAML(const std::string& file_path)
 
     const moveit::core::VariableBounds& b = joint->getVariableBounds()[0];
 
+    // Output property
+    emitter << YAML::Key << "has_position_limits";
+    emitter << YAML::Value << "false";
+
+    // Output property
+    emitter << YAML::Key << "max_position";
```
simonschmeisser, line 1178: what about the lower limit? Don't you want to include that as well?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -1170,6 +1170,14 @@ bool MoveItConfigData::outputJointLimitsYAML(const std::string& file_path)
 
     const moveit::core::VariableBounds& b = joint->getVariableBounds()[0];
 
+    // Output property
+    emitter << YAML::Key << "has_position_limits";
+    emitter << YAML::Value << "false";
```
simonschmeisser, line 1175: why `false`? I'm not even sure what happens when you set that while the URDF actually has a position limit

simonschmeisser (changes requested):

v4hn:
> +    emitter << YAML::Key << "has_position_limits";
> +    emitter << YAML::Value << "false";
> 
> why `false`? I'm not even sure what happens when you set that while the URDF actually has a position limit

Turns out the "has_position_limits" key does not even exist [in the current logic](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/robot_model_loader/src/robot_model_loader.cpp#L110) loading the parameters. It only exists [inside the JointModel](https://github.com/ros-planning/moveit/blob/master/moveit_core/robot_model/src/joint_model.cpp#L142)

Also, I never use these overrides because they are MoveIt-specific, but the actual drivers should know about them.
My naive interpretation of the `has_*_bounds` fields in `joint_limits.yaml` was always that they enable/disable overriding the URDF entries, but looking at the parsing code that is not true at all! Any entry directly applies to the model and thus setting a hypothetical `has_position_limits` key to false would potentially make the joint continuous (potentially because the joint might still not report limits `-pi/pi` yet).

So the only possible alternatives for this PR seem to be to (1) either specify all joint limits explicitly, thus overriding any changes someone does in the URDF later on, or (2) adding the joint limits as YAML comments so users might rely on them if they want to.
I strictly prefer the second option to avoid issues with users expecting their URDF values to be used.
In both cases we actually do need a new `has_position_limits` parameter to describe continuous joints.

simonschmeisser:
We use the `joint_limits.yaml` as a way to go from generic hardware/vendor joint limits that are distributed in the urdf/xacro in packages (debs) to the limits that apply to a particular instance/ a particular robot. How do you handle that case @v4hn ? Or do you have a branch for each of your robots?

We use (1) with our internal equivalent of MoveIt Setup Assistant

To keep this PR small we could also add default joint limits (+- pi) as a comment for continuous joints or not add them at all

@Sunny478 not sure why you closed this, you can simply commit and push further improvements to your branch and they will show up here. Also please don't be shy and ask questions if they occur

Sunny478:
Sorry for closing, deleted my local branch accidentally.

`or (2) adding the joint limits as YAML comments so users might rely on them if they want to.` 
Should all the joint limits (including the velocity and acceleration limits) be added as YAML comments?

`Also, I never use these overrides because they are MoveIt-specific, but the actual drivers should know about them.`
Did you mean that there exist other drivers that fetch the joint limits information from joint_limits.yaml and not the original URDF files?

v4hn:
> We use the joint_limits.yaml as a way to go from generic hardware/vendor joint limits that are distributed in the urdf/xacro in packages (debs) to the limits that apply to a particular instance/ a particular robot. How do you handle that case @v4hn ?

We have this particular problem only with UR robots and their descriptions [provide joint-limit parameters in the xacro macros used to define them](https://github.com/ros-industrial/universal_robot/blob/kinetic-devel/ur_description/urdf/ur5.urdf.xacro#L18). As I always want to add more geometry to each robot instance, I will only ever use the macro anyway in my own descriptions. If you use a robot description that does not have equivalent parameters, I would expect the maintainers to be willing to add them?

> To keep this PR small we could also add default joint limits (+- pi) as a comment for continuous joints or not add them at all

Sounds reasonable, though it's a hack and people might get the wrong idea and assume [-pi;pi] == continuous joint.
So maybe add a simple comment like `# continuous` instead.

> Should all the joint limits (including the velocity and acceleration limits also) be added as YAML comments ?

The current behavior adds them as statements, right? So by the same argument they should be comments as well.
ttbomk URDF cannot specify acceleration at all, only effort (which of course restricts acceleration implicitly). So `has_acceleration_limits: false` (the default) is pretty much equivalent to commenting them out.
I was unsure about it when I read your question, but yes, I do the dynamics limits should be commented too.

>  Did you mean that there exist other drivers that fetch the joint limits information from joint_limits.yaml and not the original URDF files?

I meant it the other way around. MoveIt will respect the limits in `joint_limits.yaml` - at least it should, though some time parameterization code can actually violate them afaik. But the lower-level controllers will not know about them at all, e.g., [ros_control only parses the URDF](https://github.com/ros-controls/ros_control/blob/noetic-devel/joint_limits_interface/include/joint_limits_interface/joint_limits_urdf.h). Now if MoveIt's limits are lower than the actual hardware limits (almost always) and trajectory execution on the hardware is delayed for some reason the low-level controller can legitimately violate MoveIt's limits even if the planned trajectory does not. The same is obviously true for any approach you might use in parallel to MoveIt, possibly visual servoing or trained controllers.

gavanderhoorn:
@v4hn wrote:
> But the lower-level controllers will not know about them at all, e.g., [ros_control only parses the URDF](https://github.com/ros-controls/ros_control/blob/noetic-devel/joint_limits_interface/include/joint_limits_interface/joint_limits_urdf.h).

this is actually not entirely true. See [this bit of code](https://github.com/PickNikRobotics/ros_control_boilerplate/blob/ee107dbf432e2ee735666c39024245f9d8436420/src/generic_hw_interface.cpp#L171-L183) in @davetcoleman's boilerplate, which makes use of [this function](https://github.com/ros-controls/ros_control/blob/64b1b5ec3b5c44764d08a8013795686aa797f5a2/joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h#L75) to parse limits from the parameter server ([docs](https://github.com/ros-controls/ros_control/blob/64b1b5ec3b5c44764d08a8013795686aa797f5a2/joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h#L41-L74)).

And the structure of those parameters is identical to MoveIt's (with the exception it also supports specifying jerk limits, and support for *minimum* velocity, etc).

v4hn:
Turns out, I never sent this reply!

Thanks, I did not know ros_control supports a superset of MoveIt's specification for that! That's great!
My main point remains though, you have to ensure the limits are respected by *every* layer of your pipeline, not just by MoveIt.
It would be much better to have all the limits specified in the robot description, but that's not supported, even in ROS2 I guess? *sigh*.

Adding "global" limits to the moveit configuration is somewhat unexpected at least to me.

Sunny478:
`much better to have all the limits specified in the robot description, but that's not supported, even in ROS2 `
Then how do we expect to solve #2486 ?

v4hn:
> > much better to have all the limits specified in the robot description, but that's not supported, even in ROS2 `
> 
> Then how do we expect to solve #2486 ?

That's not related. I just stated my own opinion on using the limits file for global setting.
MoveIt supports the file/parameters and we are not going to change that (anytime soon).
I still propose to add all the limits as comments in the joint_limits.yaml file to resolve your (valid) request.

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -54,6 +54,13 @@ namespace moveit_setup_assistant
 // File system
 namespace fs = boost::filesystem;
 
+std::string toString(double val)
+{
+    std::stringstream ss("");
+    ss << val;
+    return ss.str();
+}
+
```
JStech, line 63: Why not just use `std::to_string()`?

simonschmeisser: `std::to_string()` unfortunately is a misnomer, it should be called `std::to_localized_string()` as it will give me (`LANG=de_DE`) an output of `1,23` instead of the machine readable and expected `1.23` as it uses the c functions internally that obey environment variables.

`std::stringstream` is more likely not to use localized strings but for safety https://github.com/ros-planning/moveit/blob/master/moveit_core/utils/src/lexical_casts.cpp
includes setting the `imbue`: https://github.com/ros-planning/moveit/blob/3a61f184b82d01c75333292ca3e12ecd2765ecde/moveit_core/utils/src/lexical_casts.cpp#L51

What this should really be doing is to use the `moveit::core::to_string()` from `#include <moveit/utils/lexical_casts.h>`

Sunny478: Thanks for this, I was looking for an alternative to this function.

Sunny478:
I am unable to get my code to pass the clang-format. I have downloaded/installed the required packages (clang-format) for my text editor (I am using Atom) and I think it does work the correct way but somehow it is still unable to pass the format check on github. Could someone help me out here...

simonschmeisser:
you typically need the exact same version of clang-format (clang-format-10) in our case.

Please also check here: https://moveit.ros.org/documentation/contributing/code/   pre-commit sounds pretty nice and easy to setup

What I often do is to just have a look at the diff here https://github.com/ros-planning/moveit/pull/2553/checks?check_run_id=2363057956 and change that manually

akashjinandra:
I might be out of the loop here, but it seems like this is already able to be done with the

PR #2554, opened 2021-03-18
PlanExecution: fix leftover state after preempt

v4hn:
Clearly differentiate between preemption signal and state.

Alternative to #2403.

Fixes https://github.com/ros-planning/moveit_task_constructor/issues/217

### DIFF
--- a/moveit_ros/planning/plan_execution/src/plan_execution.cpp
+++ b/moveit_ros/planning/plan_execution/src/plan_execution.cpp
@@ -83,7 +83,6 @@ plan_execution::PlanExecution::PlanExecution(
 
   default_max_replan_attempts_ = 5;
 
-  preempt_requested_ = false;
   new_scene_update_ = false;
 
   // we want to be notified when new information is available
@@ -100,7 +99,7 @@ plan_execution::PlanExecution::~PlanExecution()
 
 void plan_execution::PlanExecution::stop()
 {
-  preempt_requested_ = true;
+  preempt_.request();
 }
 
 std::string plan_execution::PlanExecution::getErrorCodeString(const moveit_msgs::MoveItErrorCodes& error_code)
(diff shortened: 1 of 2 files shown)


moveit_ros/planning/plan_execution/src/plan_execution.cpp
```diff
@@ -161,7 +160,8 @@ void plan_execution::PlanExecution::planAndExecute(ExecutableMotionPlan& plan,
 void plan_execution::PlanExecution::planAndExecuteHelper(ExecutableMotionPlan& plan, const Options& opt)
 {
   // perform initial configuration steps & various checks
-  preempt_requested_ = false;
+  bool preempt_requested = false;
+  preempt_.check_and_clear();  // clear any previous preempt_ request
```
rhaschke, line 164: Previously, `preempt_requested_` was reset here. We should keep this behaviour, shouldn't we?

rhaschke (approved):
This is much cleaner than my solution. I have pushed some suggestions. Please feel free to discard particularly the last one if you don't like the proposed single-line format.

rhaschke (approved):
You are right: we still need to reset the flag when entering `executeAndMonitor()` directly.

moveit_ros/planning/plan_execution/src/plan_execution.cpp
v4hn: That's a standard signalling issue: Either you respond to signals that might have been sent even before the call started (my proposal to remove the line), or there is a negligible time between calling the method and `stop()` actually triggering a preemption. There's ways around it of course, but I don't see the need for that here at the moment.

I actually prefer to reset the signal, but `planAndExecute` has the same problem with my old patch.
So now I basically added your previous patch to fix that case too.

v4hn:
> Please feel free to discard particularly the last one if you don't like the proposed single-line format.

I originally wrote it that way too, but split the statements for ease of readability.

Thanks for the immediate review. :-)
I'll merge once CI succeeds.

PR #2564, opened 2021-03-25
msa: fix segfault

v4hn:
happened when a group was defined in the launch file without a `ns` attribute.


rhaschke (approved):

felixvd:
Should be backported as far as I can see.

PR #2575, opened 2021-03-31
Make setToIKSolverFrame() public

### LINKED ISSUE #2574: private `setToIKSolverFrame()` breaks workflow
https://github.com/ros-planning/moveit/pull/2461 makes `setToIKSolverFrame()` private. A public `setToIKSolverFrame()` is necessary (afaik) to effectively use other parts of the moveit API.
For example, my code uses the kinematics solver to get multiple IK solutions for a robot arm with redundant DOFs.
```
const kinematics::KinematicsBaseConstPtr &solver = jmg->getSolverInstance();
auto kinematic_state = std::make_shared<robot_state::RobotState>(model_);

kinematic_state->setToIKSolverFrame(solverTrobot, solver);

Eigen::Affine3d pose_in_solver_frame = solverTrobot * ee_target_pose_in_robot_frame;
... (conversion to target_poses) ...
solver->getPositionIK(target_poses, ...)
```
[Link to my code](https://github.com/UM-ARM-Lab/arm_robots/blob/3c6f42e622330cca31da5d722dcb855cc25b2c48/jacobian_follower/src/jacobian_follower/jacobian_follower.cpp#L170)
[Moveit docs](http://docs.ros.org/en/api/moveit_core/html/classkinematics_1_1KinematicsBase.html#ab874f2e103378473f93e1cd1d3e2e347)

It is necessary to transform the target pose(s) in the robot frame into the solver frame. With `setToIKSolverFrame` now private, I have to duplicate the `setToIKSolverFrame()` in my own code.

@rhaschke @AndyZe (were on the PR mentioned above)

Some possible resolutions are:
1) Make `setToIKSolverFrame()` public again
2) Duplicate `setToIKSolverFrame()` logic in my own code
3) Find a different pathway to get multiple IK solutions from a pose in robot (not solver) frame. 

I appreciate any time and comments w.r.t. this issue, thanks!

AndyZe:
Fixes #2574.
Partially reverts #2461.

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -1799,11 +1799,6 @@ public:
 
   std::string getStateTreeString(const std::string& prefix = "") const;
 
-private:
-  void allocMemory();
-  void initTransforms();
-  void copyFrom(const RobotState& other);
-
   /**
    * \brief Transform pose from the robot model's base frame to the reference frame of the IK solver
    * @param pose - the input to change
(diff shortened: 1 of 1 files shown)


rhaschke (changes requested):
You also made `void markDirtyJointTransforms(const JointModel* joint)` public.
Please, also remove inline from `setToIKSolverFrame`. The compiler cannot actually inline the function, because its definition is not available. Thus the linker will throw an error when the function is used externally.

PR #2579, opened 2021-04-04
Add planning_pipeline, planner_id parameters to plan() in MoveGroupInterface

felixvd:
This adds an overload for the `plan()` and `move()` functions to the `MoveGroupInterface` so one can plan with a different planner in a single line. I was actually thinking of adding `move_ptp` and `move_lin` functions for even more convenience, but @v4hn didn't seem sold, so I kept it to this general API. I don't like the low discoverability of the LIN/PTP functions (which any user with conventional robot programming experience expects to be there), but I think this version is something we can all agree on.

It's a simple implementation, but I don't think it needs to be more complicated right now. It shouldn't cause trouble unless there are multiple simultaneous planning requests. Passing the parameters all the way down to `constructMotionPlanRequest` seems like more hassle than it is worth.

You can test these changes with [these tutorials](https://github.com/ros-planning/moveit_tutorials/pull/617).

### Checklist


tylerjw (commented):
Since the planning_pipeline is configured via a parameter in a yaml file would it make sense to make it a dynamic parameter that could be configured through ros interfaces?

felixvd:
I don't think it would see much use, so I wouldn't bother. Personally I've never changed those parameters (except for STOMP), I suspect that >99% of people won't try to tune them, and those who do won't need to do it live, but use the benchmark package or similar.

By contrast, using PTP/LIN/CIRC motions and falling back to OMPL or STOMP when that fails is something I've wanted to do for ages, so this interface fills a real need, at least for me.

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -597,14 +598,21 @@ def go(self, joints=None, wait=True):
             except TypeError:
                 self.set_joint_value_target(joints)
         if wait:
-            return self._g.move()
+            if planning_pipeline and planner_id:
+                return self._g.move(planning_pipeline, planner_id)
+            else:
+                return self._g.move()
```
v4hn, line 604: `mgc.go(joints= ..., planning_pipeline='OMPL')` should try to use the default planner of the pipeline (OMPL) and not silently revert to use the default pipeline.

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -617,7 +625,12 @@ def plan(self, joints=None):
             except MoveItCommanderException:
                 self.set_joint_value_target(joints)
 
-        (error_code_msg, trajectory_msg, planning_time) = self._g.plan()
+        if planning_pipeline and planner_id:
+            (error_code_msg, trajectory_msg, planning_time) = self._g.plan(
+                planning_pipeline, planner_id
+            )
+        else:
+            (error_code_msg, trajectory_msg, planning_time) = self._g.plan()
```
v4hn, line 633: same as above.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -732,11 +738,24 @@ class MoveGroupInterface
...
+     target.
+      This call is always blocking (waits for the execution of the trajectory to complete) and requires an asynchronous
+     spinner to be started.
+     The planning pipeline and planner specified in \e pipeline_id and \e planner_id is used. */
+  MoveItErrorCode move(const std::string& pipeline_id, const std::string& planner_id);
+
   /** \brief Compute a motion plan that takes the group declared in the constructor from the current state to the
      specified
-      target. No execution is performed. The resulting plan is stored in \e plan*/
+      target. No execution is performed. The resulting plan is stored in \e plan */
```
v4hn, line 750: if you touch the lines, you might as well fix the broken newline...

moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp
```diff
@@ -439,11 +439,22 @@ class MoveGroupInterfaceWrapper : protected py_bindings_tools::ROScppInitializer
     return move() == MoveItErrorCode::SUCCESS;
   }
 
+  bool movePython2(std::string planning_pipeline, std::string planner_id)
```
v4hn, line 442: why Python2? This is not about the python version, so please use more descriptive names if you still need it after simplifying.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -827,6 +827,30 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
...
+  {
+    const std::string current_pipeline_id = planning_pipeline_id_;
+    const std::string current_planner_id = planner_id_;
+    setPlanningPipelineId(pipeline_id);
+    setPlannerId(planner_id);
+    MoveItErrorCode error_code = move(wait);
+    setPlanningPipelineId(current_pipeline_id);
+    setPlannerId(current_planner_id);
+    return error_code;
+  }
```
v4hn, line 852: This is unnecessarily thread-unsafe and the alternative is not any more complicated (as you seem to imply).

Just modify `MoveGroupInterfaceImpl::move` and `::plan` to take the additional parameters you need.
If you plan to add any more overrides in these calls in the future, maybe make [use an options struct](https://google.github.io/styleguide/cppguide.html#Function_Argument_Comments).
Defining **5** new wrapper methods when the MGI already uses pimpl does not make sense.

v4hn (commented):
great idea!

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -579,8 +579,9 @@ def set_max_acceleration_scaling_factor(self, value):
                 "Expected value in the range from 0 to 1 for scaling factor"
             )
 
-    def go(self, joints=None, wait=True):
-        """ Set the target of the group and then move the group to the specified target """
+    def go(self, joints=None, wait=True, planning_pipeline=None, planner_id=None):
+        """Set the target of the group and then move the group to the specified target.
+        planning_pipeline and planner_id specify the planner that is used for this motion."""
```
JStech, line 584: Could you also explain what the `wait` variable means?

moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -597,14 +598,21 @@ def go(self, joints=None, wait=True):
...
+                return self._g.async_move()
 
-    def plan(self, joints=None):
-        """Return a tuple of the motion planning results such as
+    def plan(self, joints=None, planning_pipeline=None, planner_id=None):
+        """Return a tuple of the motion planning results of the format:
         (success flag : boolean, trajectory message : RobotTrajectory,
-         planning time : float, error code : MoveitErrorCodes)"""
+         planning time : float, error code : MoveitErrorCodes)
+        planning_pipeline and planner_id specify the planner that is used for this plan."""
```
JStech, line 615: Could you please also explain where the default values come from when these values are None?

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -827,6 +827,30 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
     }
   }
 
+  MoveItErrorCode plan(Plan& plan_, const std::string& pipeline_id, const std::string& planner_id)
```
JStech, line 830: Since `plan` is a callable below, maybe use a more different variable name here instead of just appending an underscore? Like `output_plan`?

JStech (commented):
I read through the code and just saw a couple things.

moveit_commander/src/moveit_commander/move_group.py
felixvd: Done

moveit_commander/src/moveit_commander/move_group.py
felixvd: Sure

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
felixvd: Right? What was I thinking. Christ.

moveit_commander/src/moveit_commander/move_group.py
felixvd: There is no default per pipeline, so I went with a more verbose error message instead.

v4hn: That's not true. [The OMPL plugin even supports defaults per group](https://github.com/ros-planning/moveit/blob/master/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp#L436-L456) and sending an empty `planner_id` triggers the default. A similar functionality is not implemented for `pilz_industrial_motion_planner` at the moment, but could surely be added there.

I believe rejecting an empty `planner_id` should be the responsibility of the respective planner manager, not of the front end.
The other way around (an empty pipeline, but non-empty planner_id) also makes sense for me as a way to set the planner for the currently selected pipeline.
I would expect both cases to be supported instead of throwing errors. :question:

felixvd: Admittedly I didn't look into it too deeply, I just looked at the ROS parameters to check for a default.

But since there is a `planner_id_` setting in the `move_group_interface` which is filled in by default, the planning plugin wouldn't receive a request with an empty `planner_id` field anyway. If we want to send the MotionPlanRequest with an empty `planner_id` field, the `plan` and `move` functions need to have a different default value than an empty string. Is that better? See the newest commit for how that would look.

I would just impose the (imo minor) restriction "set both id and pipeline to use this interface", and wrap it in `move_lin`, `move_ptp`, `move_ompl` functions myself.

felixvd:
It seems like the tests in the failing test pipeline all end like this:

```
 ... shutting down processing monitor complete
[ROSTEST]rostest teardown /home/runner/work/moveit/moveit/.work/target_ws/src/moveit/moveit_ros/move_group/test/test_check_state_validity_in_empty_scene.test complete

]]&gt;</system-out><system-err>&lt;![CDATA[
&amp;#x001B;[31m[check_state_validity_in_empty_scene-4] process has died [pid 63669, exit code 1, cmd /home/runner/work/moveit/moveit/.work/target_ws/src/moveit/moveit_ros/move_group/test/test_check_state_validity_in_empty_scene.py --gtest_output=xml:/home/runner/work/moveit/moveit/.work/target_ws/build/moveit_ros_move_group/test_results/moveit_ros_move_group/rosunit-check_state_validity_in_empty_scene.xml __name:=check_state_validity_in_empty_scene __log:=/root/.ros/log/09c6690a-ae2b-11eb-a7f0-0242ac110002/check_state_validity_in_empty_scene-4.log].
log file: /root/.ros/log/09c6690a-ae2b-11eb-a7f0-0242ac110002/check_state_validity_in_empty_scene-4*.log&amp;#x001B;[0m

]]&gt;</system-err></testsuite>
-------------------------------------------------
```

Anyone got an idea? @tylerjw

moveit_commander/src/moveit_commander/move_group.py
v4hn: > If we want to send the MotionPlanRequest with an empty planner_id field, the plan and move functions need to have a different default value than an empty string. Is that better?

No you are right. In C++, this is terminally ugly. My initial comment only referred to the python API though.
For C++, you would usually have different method names `planWithPipeline(string pipeline, string planner="")` and `planWithPlanner(string planner)` instead.
If you think this is a lousy implementation as well, you can still enforce your set-pipeline-and-planner constraint in the C++ API... Either way we should get this merged.

v4hn:
quite a few things changed since this was proposed. Could you rebase to see whether this fixes your issues?

PR #2580, opened 2021-04-06
Make setToIKSolverFrame accessible again

v4hn:
This is public API because the solver's getPositionIK requires input poses
in the frame and the relevant transform is otherwise inaccessible.

Partially reverts #2461.

Supersedes #2575


felixvd (approved):
Looks fine to me and addresses @rhaschke 's comments.

rhaschke (approved):

PR #2594, opened 2021-04-09
Avoid joint jump when SuddenHalt() is called in velocity mode

nbbrooks:
### Description

[This is a backport of a MoveIt2 bug fix](https://github.com/ros-planning/moveit2/pull/410/files)

We found out the hard way, when the suddenHalt() function was called for a robot in velocity control mode, it would send a command vector of all-zero positions. That could cause a really big joint jump depending what type of ros2_controller is active. This fixes that.

Also (really minor fix) it sets trajectory accelerations to 0 to match the comment.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -834,9 +834,9 @@ void ServoCalcs::suddenHalt(trajectory_msgs::JointTrajectory& joint_trajectory)
   // being 0 seconds in the past, the smallest supported timestep is added as time from start to the trajectory point.
   point.time_from_start.fromNSec(1);
 
-  if (parameters_->publish_joint_positions)
+  if (parameters_.publish_joint_positions)
     point.positions.resize(num_joints_);
-  if (parameters_->publish_joint_velocities)
+  if (parameters_.publish_joint_velocities)
     point.velocities.resize(num_joints_);
 
   // Assert the following loop is safe to execute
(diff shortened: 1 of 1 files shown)


v4hn:
Hi Nathan,

thanks for backporting. This is clearly urgent and for me would even justify a quick followup release for noetic because it's a safety issue. @tylerjw 

The PR fails CI because of MoveIt/MoveIt2 divergence:
Apparently [this commit](https://github.com/ros-planning/moveit2/commit/db1f3227ea59af9aac3cd290a0ce4555eddc3718#diff-7355431db583f9f75319bb82dee0c9379da3a7a720e37b49f07fc9e80d6175a9R243) switched to pointers (the I have no idea how this can be a "Port from MoveIt" commit when MoveIt does not even have these changes...)
I guess you could just change to `.` member access to solve it.

AndyZe:
It's still pretty unlikely this would cause issues. Somebody would need to be using a controller that accepts optional positions & vels, but only have Servo configured to publish vels. Definitely unusual!

But yeah, we should get it merged.

tylerjw:
> I have no idea how this can be a "Port from MoveIt" commit when MoveIt does not even have these changes...

It is a ROS1 -> ROS2 port.  The code that it is modifying was a copy of the ros1 code and was previous to this PR explicitly ignored.  There are some new features also (listed in the text of the commit message) that we also added like some more tests and examples.

As to why this variable changed to a pointer type, (if I remember correctly) that has to do with changes needed to support the changes in the parameter system from ROS2.

nbbrooks:
Oops, fixed the pointer access issue.

tylerjw:
Just like in another PR @AndyZe was asking me about.  You need to fetch and rebase this on master to get it to pass CI.  The name of the `Format` check was changed to `Format / pre-commit` and so the required `Format` check will never run.

tl;dr rebase on master.

v4hn:
> tl;dr rebase on master.

The bigger problem was that `Format` was still a required CI job (which never ran).
I adjusted the github repo settings just now and made `pre-commit` the required job.

v4hn (approved):

PR #2596, opened 2021-04-09
Lock the octomap/octree while collision checking

simonschmeisser:
### Description

We get crashes rarely where a collision check in FCL is ongoing while we update the octomap from another thread. This is due to the fact that we copy the `PlanningScene` maintained by `PlanningSceneMonitor` but the octomap shape/object is actually just a pointer and still shared with the one updated by the `PlanningSceneMonitor`. So far no locking of the octomap happens when doing collision checks this way (or in general?). Note that the octomap updater plugins in general only lock the octomap but not the `PlanningScene`.

I'm not super happy with this PR yet but would like to get some opinions already if I'm heading off completely in the wrong direction. Also I'm somewhat wondering why this doesn't impact people with a "live stream" 3d camera (we only update after each pick, ie every 5s or so), maybe because we use clearOctomap and then repopulate it with the new PointCloud?

### Checklist
- [x] make it build (cyclic dependency as `occupancy_map_monitor::OccMapTree : public octomap::OcTree` lives in `moveit_ros`
- [x] make sure the second cast is always valid, can there be actually any `octomap::OcTree`  instead of `occupancy_map_monitor::OccMapTree`
- [ ] is that `const` + `mutable` acceptable? I think so but I might be biased ...

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1289,6 +1289,26 @@ bool PlanningScene::usePlanningSceneMsg(const moveit_msgs::PlanningScene& scene_
     return setPlanningSceneMsg(scene_msg);
 }
 
+collision_detection::OccMapTreePtr createOctomap(const octomap_msgs::Octomap& map)
+{
+  std::shared_ptr<collision_detection::OccMapTree> om =
+      std::make_shared<collision_detection::OccMapTree>(map.resolution);
+  if (map.binary)
+  {
+    octomap_msgs::readTree(om.get(), map);
+  }
+  else
+  {
+    std::stringstream datastream;
+    if (map.data.size() > 0)
+    {
+      datastream.write((const char*)&map.data[0], map.data.size());
+      om->readData(datastream);
+    }
+  }
+  return om;
+}
+
 void PlanningScene::processOctomapMsg(const octomap_msgs::Octomap& map)
 {
   // each octomap replaces any previous one
@@ -1303,7 +1323,7 @@ void PlanningScene::processOctomapMsg(const octomap_msgs::Octomap& map)
     return;
   }
 
-  std::shared_ptr<octomap::OcTree> om(static_cast<octomap::OcTree*>(octomap_msgs::msgToMap(map)));
+  std::shared_ptr<collision_detection::OccMapTree> om = createOctomap(map);
   if (!map.header.frame_id.empty())
   {
     const Eigen::Isometry3d& t = getFrameTransform(map.header.frame_id);
@@ -1341,7 +1361,8 @@ void PlanningScene::processOctomapMsg(const octomap_msgs::OctomapWithPose& map)
     return;
   }
 
-  std::shared_ptr<octomap::OcTree> om(static_cast<octomap::OcTree*>(octomap_msgs::msgToMap(map.octomap)));
+  std::shared_ptr<collision_detection::OccMapTree> om = createOctomap(map.octomap);
+
   const Eigen::Isometry3d& t = getFrameTransform(map.header.frame_id);
   Eigen::Isometry3d p;
   tf2::fromMsg(map.origin, p);
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -480,14 +480,7 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest& 
                                    collision_detection::CollisionResult& res,
                                    const moveit::core::RobotState& robot_state) const
 {
-  // check collision with the world using the padded version
-  getCollisionEnv()->checkRobotCollision(req, res, robot_state, getAllowedCollisionMatrix());
-
-  if (!res.collision || (req.contacts && res.contacts.size() < req.max_contacts))
-  {
-    // do self-collision checking with the unpadded version of the robot
-    getCollisionEnvUnpadded()->checkSelfCollision(req, res, robot_state, getAllowedCollisionMatrix());
-  }
+  checkCollision(req, res, robot_state, getAllowedCollisionMatrix());
 }
 
 void PlanningScene::checkSelfCollision(const collision_detection::CollisionRequest& req,
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -851,6 +851,12 @@ bool PlanningScene::getOctomapMsg(octomap_msgs::OctomapWithPose& octomap) const
     if (map->shapes_.size() == 1)
     {
       const shapes::OcTree* o = static_cast<const shapes::OcTree*>(map->shapes_[0].get());
+
+      // lock the octomap if there is any as it might be shared with other PlanningScenes
+      collision_detection::OccMapTree::ReadLock lock;
+      const collision_detection::OccMapTree* occ = static_cast<const collision_detection::OccMapTree*>(o->octree.get());
+      lock = occ->reading();
+
       octomap_msgs::fullMapToMsg(*o->octree, octomap.octomap);
       octomap.origin = tf2::toMsg(map->shape_poses_[0]);
       return true;
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -38,6 +38,7 @@
 #include <moveit/planning_scene/planning_scene.h>
 #include <moveit/collision_detection_fcl/collision_detector_allocator_fcl.h>
 #include <geometric_shapes/shape_operations.h>
+#include <moveit/collision_detection/occupancy_map.h>
 #include <moveit/collision_detection/collision_tools.h>
 #include <moveit/trajectory_processing/trajectory_tools.h>
 #include <moveit/robot_state/conversions.h>
@@ -503,6 +504,16 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest& 
                                    const moveit::core::RobotState& robot_state,
                                    const collision_detection::AllowedCollisionMatrix& acm) const
 {
+  // lock the octomap if there is any as it might be shared with other PlanningScenes
+  collision_detection::OccMapTree::ReadLock lock;
+  collision_detection::CollisionEnv::ObjectConstPtr map = world_->getObject(OCTOMAP_NS);
+  if (map && map->shapes_.size() == 1)
+  {
+    const shapes::OcTree* o = static_cast<const shapes::OcTree*>(map->shapes_[0].get());
+    const collision_detection::OccMapTree* occ = static_cast<const collision_detection::OccMapTree*>(o->octree.get());
+    lock = occ->reading();
+  }
+
   // check collision with the world using the padded version
   getCollisionEnv()->checkRobotCollision(req, res, robot_state, acm);
 
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -529,6 +529,16 @@ void PlanningScene::checkCollisionUnpadded(const collision_detection::CollisionR
                                            const moveit::core::RobotState& robot_state,
                                            const collision_detection::AllowedCollisionMatrix& acm) const
 {
+  // lock the octomap if there is any as it might be shared with other PlanningScenes
+  collision_detection::OccMapTree::ReadLock lock;
+  collision_detection::CollisionEnv::ObjectConstPtr map = world_->getObject(OCTOMAP_NS);
+  if (map && map->shapes_.size() == 1)
+  {
+    const shapes::OcTree* o = static_cast<const shapes::OcTree*>(map->shapes_[0].get());
+    const collision_detection::OccMapTree* occ = static_cast<const collision_detection::OccMapTree*>(o->octree.get());
+    lock = occ->reading();
+  }
+
   // check collision with the world using the unpadded version
   getCollisionEnvUnpadded()->checkRobotCollision(req, res, robot_state, acm);
 
--- a/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
+++ b/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
@@ -99,7 +99,7 @@ void OccupancyMapMonitor::initialize()
                                                      << "\" specified but no TF instance (buffer) specified. "
                                                         "No transforms will be applied to received data.");
 
-  tree_.reset(new OccMapTree(map_resolution_));
+  tree_.reset(new collision_detection::OccMapTree(map_resolution_));
   tree_const_ = tree_;
 
   XmlRpc::XmlRpcValue sensor_list;
--- a/moveit_core/collision_detection/include/moveit/collision_detection/occupancy_map.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/occupancy_map.h
@@ -42,7 +42,7 @@
 #include <boost/function.hpp>
 #include <memory>
 
-namespace occupancy_map_monitor
+namespace collision_detection
 {
 typedef octomap::OcTreeNode OccMapNode;
 
@@ -115,4 +115,4 @@ private:
 
 using OccMapTreePtr = std::shared_ptr<OccMapTree>;
 using OccMapTreeConstPtr = std::shared_ptr<const OccMapTree>;
-}  // namespace occupancy_map_monitor
+}  // namespace collision_detection
--- a/moveit_core/collision_detection/include/moveit/collision_detection/occupancy_map.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/occupancy_map.h
@@ -86,7 +86,7 @@ public:
   using ReadLock = boost::shared_lock<boost::shared_mutex>;
   using WriteLock = boost::unique_lock<boost::shared_mutex>;
 
-  ReadLock reading()
+  ReadLock reading() const
   {
     return ReadLock(tree_mutex_);
   }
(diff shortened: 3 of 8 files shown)


v4hn:
You're right, the locking there is necessary but not implemented.

I think your patch makes sense, but we should avoid dependency cycles.
Would it make sense to generalize the concept of a collision object that needs locking with an interface in a class in `moveit_core` (e.g., `MonitoredCollisionObject`) and have the OctoMap implement that concept? In that case a separate list for such objects in `PlanningScene` sounds like a good idea.

CI fails with 
```
  /home/runner/work/moveit/moveit/.work/target_ws/src/moveit/moveit_core/planning_scene/src/planning_scene.cpp:516:27: error: passing ‘const occupancy_map_monitor::OccMapTree’ as ‘this’ argument discards qualifiers [-fpermissive]
  
         lock = occ->reading();
```

simonschmeisser:
> You're right, the locking there is necessary but not implemented.
> 
> I think your patch makes sense, but we should avoid dependency cycles.
> Would it make sense to generalize the concept of a collision object that needs locking with an interface in a class in `moveit_core` (e.g., `MonitoredCollisionObject`) and have the OctoMap implement that concept? In that case a separate list for such objects in `PlanningScene` sounds like a good idea.

It could also be solved less invasive by moving `OccMapTree` to moveit_core, either as a separate module or in the planning_scene module

> 
> CI fails with
> 
> ```
>   /home/runner/work/moveit/moveit/.work/target_ws/src/moveit/moveit_core/planning_scene/src/planning_scene.cpp:516:27: error: passing ‘const occupancy_map_monitor::OccMapTree’ as ‘this’ argument discards qualifiers [-fpermissive]
>   
>          lock = occ->reading();
> ```

I didn't look too closely at why the CI fails but assumed it was due to the cyclic dependency which gives me linker errors locally ... this error did not happen for me locally and I'm not sure why it does? `mutable` should fix that?

v4hn:
> It could also be solved less invasive by moving OccMapTree to moveit_core, either as a separate module or in the planning_scene module

Sounds good to me. It's just tiny wrapper around octomap anyway.

> I didn't look too closely at why the CI fails but assumed it was due to the cyclic dependency which gives me linker errors locally 

Might be that it pulled in a system header in the wrong place. Cycles are evil :-)

simonschmeisser:
> > It could also be solved less invasive by moving OccMapTree to moveit_core, either as a separate module or in the planning_scene module
> 
> Sounds good to me. It's just tiny wrapper around octomap anyway.
> 
Do you prefer a separate new module or as part of planning_scene?

v4hn:
> Do you prefer a separate new module or as part of planning_scene?

Neither. The internal collision objects are defined [in `collision_detection`](https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection/include/moveit/collision_detection/world.h#L79), so I'd prefer a new header there.
Unless anyone has a better proposal of course. :-)

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -503,6 +504,19 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest&
                                    const moveit::core::RobotState& robot_state,
                                    const collision_detection::AllowedCollisionMatrix& acm) const
 {
+  // lock the octomap if there is any as it might be shared with other PlanningScenes
+  collision_detection::OccMapTree::ReadLock lock;
+  collision_detection::CollisionEnv::ObjectConstPtr map = world_->getObject(OCTOMAP_NS);
+  if (map)
+  {
+    if (map->shapes_.size() == 1)
+    {
```
v4hn, line 513:
```suggestion
  if (map && map->shapes_.size() == 1)
  {
```
`&&` is a [sequence point](https://stackoverflow.com/questions/4176328/what-are-sequence-points-and-how-do-they-affect-undefined-behavior), no need for the nested testing.

v4hn (approved):
Aside from the nit above, it looks good to me.

henningkayser (commented):
Looks good to me. Shouldn't this lock also be applied to the `checkCollision()` overload without acm and both `checkCollisionUnpadded()` functions?

simonschmeisser:
You're referring to this one? https://github.com/ros-planning/moveit/blob/4ff48ddc94ec5f50ea901ecb7c3c0b59ffb7739b/moveit_core/planning_scene/src/planning_scene.cpp#L479

Just thought the same and was maybe expecting this to be the case a bit too much so that I missed it ...

This one is also missing: https://github.com/ros-planning/moveit/blob/4ff48ddc94ec5f50ea901ecb7c3c0b59ffb7739b/moveit_core/planning_scene/src/planning_scene.cpp#L534

any further?

simonschmeisser:
https://github.com/ros-planning/moveit/blob/4ff48ddc94ec5f50ea901ecb7c3c0b59ffb7739b/moveit_core/planning_scene/src/planning_scene.cpp#L840

as well

henningkayser (approved):
Thanks, that should be all.

henningkayser (commented):
... or maybe not quite. This seems to almost be related to https://github.com/flexible-collision-library/fcl/issues/473 and https://github.com/ros-planning/moveit/pull/2129. I think we should expose a ReadLock helper function either from the PlanningScene or the OctomapMonitor so that collision checking can be made thread-safe even when not using the `checkCollision()` convenience functions. This is kind-of scope creep as the user could also just use the same locking code, but it would be a nice follow-up.

v4hn:
> This is kind-of scope creep. I think we should expose a ReadLock helper function either from the PlanningScene or the OctomapMonitor

Yes, I think it creeps :) . But the discussion is well justified because locking in MoveIt is not consistent.

 `OCTOMAP_NS` (and thus the concept of a single octomap) is only defined in the PlanningScene, so we shouldn't add a wrapper to CollisionEnv, which is where I would like to have it. We could add a wrapper to OctomapMonitor, but I'm not sure it would actually help. Is the Monitor usually available in places where you would need to lock? If you need to keep it in scope just for that, you might also just reference the OccTree directly (and call its locking methods).

I'm not happy with exposing the lock methods in the PlanningScene. We don't offer any other locking of collision objects in that class (and for other dynamically-tracked objects it would be just as useful!), so it is rather confusing to expose a locking method for one specific object that *might* be tracked by the class.
If we add an API here, I would prefer to generalize the concept of lockable/dynamic objects in the planning scene, so, e.g., vision-tracked meshes could be included as well.

Here's more food for thought though. The [PSM provides locking for the whole PlanningScene object](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L611), but it does not acquire the octomap lock. This seems very counter-intuitive to me, but would existing setups loose too much time to update the octomap if we lock it there already? The obvious downside of the current approach is that the octomap can change *during planning*, with unexpected consequences (e.g., valid edges in the planner cache becoming invalid).

If we want to lock it there already, we would need a [recursive shared lock](https://stackoverflow.com/questions/36619715/a-shared-recursive-mutex-in-standard-c) because the locks in the PlanningScene would still be needed.
Alternatively we could introduce full locking directly in the PlanningScene, so the PSM only needs to call it there and does not have to maintain its own locking in the helper classes.

lianghongzhuo:
After using this patch, something weird happens to me. 
I'm testing this with ubuntu 18, moveit master with this patch and a real robot.
for example `rosservice call /clear_octomap` will not clean octomap anymore.

lianghongzhuo:
How to reproduce:
1. Install moveit master with this patch
2. `roslaunch moveit_tutorials detect_and_add_cylinder_collision_object_demo.launch`
3. Add `MotionPlanning` in rviz
4. The rviz freez or crashes
Screenshot from 2021-04-24 08-14-24
Error msg might usefull:
```
[ERROR] [1619244812.320281243]: Exception caught executing main loop job: boost: mutex lock failed in pthread_mutex_lock: Invalid argument
[ INFO] [1619244813.277638773]: Ready to take commands for planning group hand.
[ INFO] [1619244813.277738985]: Looking around: no
[ INFO] [1619244813.277794851]: Replanning: no
[ERROR] [1619244813.309339603]: Exception caught executing main loop job: boost: mutex lock failed in pthread_mutex_lock: Invalid argument

```

simonschmeisser:
I can indeed see some strange behavior in the demo @lianghongzhuo mentioned. I can also sometimes trigger a dead-lock between `PlanningSceneMonitor::newPlanningSceneMessage` and `moveit_rviz_plugin::MotionPlanningDisplay::drawQueryGoalState`

Backtraces:
```
1  futex_wait_cancelable                                                                                                                                                                                                                                                                                                                                                           futex-internal.h               88  0x7fee494ccad3 
2  __pthread_cond_wait_common                                                                                                                                                                                                                                                                                                                                                      pthread_cond_wait.c            502 0x7fee494ccad3 
3  __pthread_cond_wait                                                                                                                                                                                                                                                                                                                                                             pthread_cond_wait.c            655 0x7fee494ccad3 
4  boost::condition_variable::wait                                                                                                                                                                                                                                                                                                                                                 condition_variable.hpp         81  0x7fee4b681c0d 
5  boost::shared_mutex::lock                                                                                                                                                                                                                                                                                                                                                       shared_mutex.hpp               294 0x7fee4b682cff 
6  boost::unique_lock<boost::shared_mutex>::lock                                                                                                                                                                                                                                                                                                                                   lock_types.hpp                 346 0x7fee4b682cff 
7  boost::unique_lock<boost::shared_mutex>::unique_lock                                                                                                                                                                                                                                                                                                                            lock_types.hpp                 124 0x7fedb0225bfe 
8  planning_scene_monitor::PlanningSceneMonitor::newPlanningSceneMessage                                                                                                                                                                                                                                                                                                           planning_scene_monitor.cpp     565 0x7fedb0225bfe 
9  boost::function1<void, boost::shared_ptr<moveit_msgs::PlanningScene_<std::allocator<void>> const> const&>::operator()                                                                                                                                                                                                                                                           function_template.hpp          759 0x7fedb023c9de 
10 boost::detail::function::void_function_obj_invoker1<boost::function<void (boost::shared_ptr<moveit_msgs::PlanningScene_<std::allocator<void>> const> const&)>, void, boost::shared_ptr<moveit_msgs::PlanningScene_<std::allocator<void>> const>>::invoke(boost::detail::function::function_buffer&, boost::shared_ptr<moveit_msgs::PlanningScene_<std::allocator<void>> const>) function_template.hpp          159 0x7fedb023c9de 
11 boost::function1<void, boost::shared_ptr<moveit_msgs::PlanningScene_<std::allocator<void>> const>>::operator()                                                                                                                                                                                                                                                                  function_template.hpp          759 0x7fedb025c236 
12 ros::SubscriptionCallbackHelperT<boost::shared_ptr<moveit_msgs::PlanningScene_<std::allocator<void>> const> const&, void>::call                                                                                                                                                                                                                                                 subscription_callback_helper.h 144 0x7fedb025c236 
13 ros::SubscriptionQueue::call()                                                                                                                                                                                                                                                                                                                                                                                     0x7fee480b2422 
14 ros::CallbackQueue::callOneCB(ros::CallbackQueue::TLS *)                                                                                                                                                                                                                                                                                                                                                           0x7fee4805cd99 
15 ros::CallbackQueue::callAvailable(ros::WallDuration)                                                                                                                                                                                                                                                                                                                                                               0x7fee4805eb0b 
16 ros::AsyncSpinnerImpl::threadFunc()                                                                                                                                                                                                                                                                                                                                                                                0x7fee480b5245 
17 ??                                                                                                                                                                                                                                                                                                                                                                                                                 0x7fee496efbcd 
18 start_thread                                                                                                                                                                                                                                                                                                                                                                    pthread_create.c               463 0x7fee494c66db 
19 clone                                                                                                                                                                                                                                                                                                                                                                           clone.S                        95  0x7fee4a0c471f 

```
vs:
```
1   __lll_lock_wait                                                  lowlevellock.S              135 0x7fee494d01fd 
2   __GI___pthread_mutex_lock                                        pthread_mutex_lock.c        80  0x7fee494c9025 
3   boost::posix::pthread_mutex_lock                                 mutex.hpp                   62  0x7fee4b681558 
4   boost::mutex::lock                                               mutex.hpp                   116 0x7fee4b681558 
5   boost::unique_lock<boost::mutex>::lock                           lock_types.hpp              346 0x7fee4b681558 
6   boost::unique_lock<boost::mutex>::unique_lock                    lock_types.hpp              124 0x7fedb23ccca6 
7   boost::shared_mutex::lock_shared                                 shared_mutex.hpp            188 0x7fedb23ccca6 
8   boost::shared_lock<boost::shared_mutex>::lock                    lock_types.hpp              645 0x7fedb23ccca6 
9   boost::shared_lock<boost::shared_mutex>::shared_lock             lock_types.hpp              520 0x7fedb23aeef1 
10  collision_detection::OccMapTree::reading                         occupancy_map.h             91  0x7fedb23aeef1 
11  planning_scene::PlanningScene::checkCollision                    planning_scene.cpp          507 0x7fedb23aeef1 
12  planning_scene::PlanningScene::getCollidingPairs                 planning_scene.cpp          569 0x7fedb23af4b0 
13  planning_scene::PlanningScene::getCollidingLinks                 planning_scene.cpp          585 0x7fedb23b1954 
14  planning_scene::PlanningScene::getCollidingLinks                 planning_scene.h            535 0x7feda278fd7a 
15  moveit_rviz_plugin::MotionPlanningDisplay::drawQueryGoalState    motion_planning_display.cpp 754 0x7feda278fd7a 
16  moveit_rviz_plugin::MotionPlanningDisplay::changedQueryGoalState motion_planning_display.cpp 733 0x7feda2790455 
17  boost::function0<void>::operator()                               function_template.hpp       759 0x7fedb04aa2e7 
18  moveit_rviz_plugin::PlanningSceneDisplay::executeMainLoopJobs    planning_scene_display.cpp  255 0x7fedb04aa2e7 
19  moveit_rviz_plugin::PlanningSceneDisplay::update                 planning_scene_display.cpp  642 0x7fedb04aa699 
20  rviz::DisplayGroup::update                                       display_group.cpp           242 0x7fee4b6689e8 
... <Mehr>                                                                                                          

```


I also get some of these 

```
[ERROR] [1619468257.868416197]: Exception caught executing main loop job: boost: mutex lock failed in pthread_mutex_lock: Invalid argument
```

(the code of said tutorial looks like it could use some love as well ...)

Any ideas where those issues could come from?

lianghongzhuo:
@simonschmeisser 
Sorry, the reproduce code should use
`roslaunch moveit_tutorials obstacle_avoidance_demo.launch`
because the `detect_and_add_cylinder_collision_object_demo.launch` file did not contain tf2 transform publisher.
I didn't encouter the transform problem because I run `obstacle_avoidance_demo.luanch` fisrt, then the system contains the tf2 transform.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -479,14 +480,7 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest&
...
 {
-  // check collision with the world using the padded version
-  getCollisionEnv()->checkRobotCollision(req, res, robot_state, getAllowedCollisionMatrix());
-
-  if (!res.collision || (req.contacts && res.contacts.size() < req.max_contacts))
-  {
-    // do self-collision checking with the unpadded version of the robot
-    getCollisionEnvUnpadded()->checkSelfCollision(req, res, robot_state, getAllowedCollisionMatrix());
-  }
+  checkCollision(req, res, robot_state, getAllowedCollisionMatrix());
```
mamoll, line 483: It looks like this was done to eliminate duplicated code, right?

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -99,7 +99,7 @@ void OccupancyMapMonitor::initialize()
                                                      << "\" specified but no TF instance (buffer) specified. "
                                                         "No transforms will be applied to received data.");
 
-  tree_.reset(new OccMapTree(map_resolution_));
+  tree_.reset(new collision_detection::OccMapTree(map_resolution_));
```
mamoll, line 102: You need to change the namespace in line 44 instead.

mamoll (approved):
This is good idea and clearly necessary. I just have a couple small comments.

Why was occupancy_map.h moved to collision_detection?

moveit_core/planning_scene/src/planning_scene.cpp
simonschmeisser: Yes, because I proved during this PR that duplicate code can lead to bugs as I only locked it in one place ...

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
simonschmeisser: no? the wrapper `OccMapTree` was moved out of this namespace/module but occupancy_map_monitor is still there

simonschmeisser:
> Why was occupancy_map.h moved to collision_detection?

it would create a cyclic dependency between moveit_core and moveit_ros otherwise, see suggestions by @v4hn above 

@lianghongzhuo I think there might be a dead-lock with the combination of locking the Octomap and the shape cache that causes the issue you pointed to, I'll try to verify that and find a solution soon

lianghongzhuo:
Seems this pr still has some issues, see: https://github.com/ros-planning/moveit/issues/2666
Usually, first motion planning will work, but after running the motion planning several times, the moveit will be stuck.

rhaschke:
This is really bad news! Please, @simonschmeisser, @AndyZe, and @v4hn tackle this asap.
@lianghongzhuo: if you can provide a reproducible test case, that would be very helpful for sure.

lianghongzhuo:
I finally figured our some reproduce code:
First launch: 
with ocotomap:
`roslaunch moveit_tutorials obstacle_avoidance_demo.launch`
without ocotomap:
`roslaunch panda_moveit_config demo.launch rviz_tutorial:=true`
Then run below code:
```
from moveit_commander import MoveGroupCommander
commander = MoveGroupCommander("panda_arm")
for i in range(100):
    commander.set_start_state_to_current_state()
    j = commander.get_current_joint_values()
    j[4] += 0.2
    j[2] += 0.2
    commander.set_joint_value_target(j)
    commander.go(wait=True)
```
If there is no octomap involved, the code will run into an error and return, but with octomap involved, the whole code stuck and did not return.

PR #2597, opened 2021-04-09
GHA: Add pre-release job

rhaschke:
Fixes #2595.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 # This config uses industrial_ci (https://github.com/ros-industrial/industrial_ci.git).
 # For troubleshooting, see readme (https://github.com/ros-industrial/industrial_ci/blob/master/README.rst)
 
-name: BuildAndTest
+name: CI
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


tylerjw (approved):
looks like this should work, this will be really convenient for testing the pre-release tests.

tylerjw:
Maybe it would make sense to put this in a separate yaml file?  I don't care either way though.

rhaschke:
> Maybe it would make sense to put this in a separate yaml file?

I agree. Obviously, it's not possible to trigger a single job from a workflow.

rhaschke:
Weird, I was struggling with the correct naming of the pre-release job: It wasn't using the `matrix.distro` name.
Looks like this is only considered if the job isn't skipped. I enabled the job temporarily now to test the pre-release test.

rhaschke:
The pre-release test can now be triggered manually:
image

Here is an example run: https://github.com/ros-planning/moveit/actions/runs/735321782

v4hn:
Thank you so much @rhaschke !
This simplifies life a lot.
The only issue I see is that the logs for the test are so long that it's basically impossible to scroll to the actual error in the dynamically-loading web view. Downloading the raw logs works perfectly though.

Issue #2572, opened 2021-03-30
KDL kinematics plugin's forward kinematics


eferre89:
### Description

Up to my understanding, this is the current implementation of `getPositionFK` for the KDL kinematics plugin:

```cpp
bool valid = true;
for (unsigned int i = 0; i < poses.size(); i++)
{
  if (fk_solver_->JntToCart(jnt_pos_in, p_out) >= 0)
  {
    poses[i] = tf2::toMsg(p_out);
  }
  else
  {
    ROS_ERROR_NAMED("kdl", "Could not compute FK for %s", link_names[i].c_str());
    valid = false;
  }
}
return valid;
```

It seems to me that, at each iteration, `poses[i]` always takes the same value, regardless of `i` (regardless of the specific links in the `link_names` parameter). I think that the purpose of this for loop is to give each `pose[i]` the pose of the link we are computing FK for. 

Maybe the programmer intention was to provide an implementation that resembles the following snippet:

```cpp
bool valid = true;

for (unsigned int i = 0; i < poses.size(); i++)
{
    KDL::Frame p_out;

    if (fk_solver_->JntToCart(jnt_pos_in, p_out, link_indices[i]+1) >= 0)
    {
        poses[i] = tf2::toMsg(p_out);
    }
    else
    {
        ROS_ERROR_NAMED("nsp", "Could not compute FK for %s", link_names[i].c_str());
        valid = false;
    }
}

return valid;
```

where `link_indices` is computed from `link_names` and the joint model group.

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Binary build
* Maybe release version is not relevant, since I'm referring to source code

### Steps to reproduce
For example you can assign `link_names = getLinkNames()` (all links) and call `getPositionFK(link_names, joint_angles, poses)`.

### Expected behaviour
`poses` should contain forward kinematics for each link passed in `link_names`.

### Actual behaviour
`poses` contains forward kinematics for the last link repeated `link_names.size()` times.

### Backtrace or Console output

N/A

simonschmeisser:
This code indeed looks suspicious, please open a PR with your changes and ideally also a test like you described in "Steps to reproduce"

Personally I would rather use a RobotState, set the positions (joint values), call update and then get the FK positions from getGlobalLinkPose

rhaschke:
You are right, the code currently doesn't handle multiple link names, primarily because the KDL plugin doesn't support non-chain kinematic trees:
[...]

PR #2600, opened 2021-04-13
Bugfix/kdl forward kinematics with multiple frames

eferre89:
### Description

This PR allows computing FK with KDL for intermediate frames (frames that are not the tip of the chain). It closes issue #2572.

### Checklist


JStech (commented):
Thanks for the contribution. It's currently failing the formatting check--you can fix this by running `clang-format-10 -i` on it, or using `pre-commit run -a`.

rhaschke (changes requested):
I just noticed that I didn't add this comment as a review:

Usually kinematics plugins are tailored towards a _fixed_ subtree / subchain of the RobotModel. In this case, the mapping from link names to an internal id/index can be established once during initialization.
In your case, I suggest computing this map once in `initialize()`, not for each call to `getPositionFK()`.

PR #2602, opened 2021-04-14
Let users override fake execution type from demo.launch

v4hn:
This should be the main entry point for simulated robot systems
and we specifically want to support `last point` here for rostests.

This was an oversight in #1893 .


jschleicher (approved):
yep, absolutely. Thank you @v4hn !

PR #2609, opened 2021-04-16
Fix remove_attached_object API in Python PSI, add clear()

felixvd:
Setting the link name in `remove_attached_object` is not necessary. `planning_scene.cpp` treats the message as described.

I also added a `clear()` method because I could have used it today and saw that @mikeferguson also did in [moveit_python](http://docs.ros.org/en/api/moveit_python/html/classmoveit__python_1_1planning__scene__interface_1_1PlanningSceneInterface.html#aa36ea0b0171e79ddfea196aed4f9d8a4).

### DIFF
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -93,36 +93,26 @@ class PlanningSceneInterface(object):
                 self._pub_co.publish(collision_object)
 
     def add_object(self, collision_object):
-        """
-        Add an object to the planning scene
-        """
+        """ Add an object to the planning scene """
         self.__submit(collision_object, attach=False)
 
     def add_sphere(self, name, pose, radius=1):
-        """
-        Add a sphere to the planning scene
-        """
+        """ Add a sphere to the planning scene """
         co = self.__make_sphere(name, pose, radius)
         self.__submit(co, attach=False)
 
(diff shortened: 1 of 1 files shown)


moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -172,6 +160,11 @@ def attach_box(self, link, name, pose=None, size=(1, 1, 1), touch_links=[]):
             aco.touch_links = [link]
         self.__submit(aco, attach=True)
 
+    def clear(self):
+        """ Remove all objects from the planning scene """
+        self.remove_attached_object()
+        self.remove_world_object()
```
v4hn, line 166: It's a useful abstraction, but it would be nicer to expose this API as sending *a single* PlanningScene message.
It's just stupid overhead that can even show in Rviz (the attached objects might actually detach in RViz before being removed).
It looks like the asynchronous backend of the class does not even have a publisher to `planning_scene` at the moment...

v4hn (approved):
Looks good to me. CI failed for seemingly unrelated reasons, I restarted the job.

Please consider my inline comment though.

rhaschke (approved):
I can only support Michael's review. +1

felixvd:
Thanks for the reviews. I agree that there are things to be improved in the backend (a `CLEAR_ALL` command, a `DETACH` command separate from `REMOVE`...), but I prefer not to go that far down the rabbit hole right now.

PR #2610, opened 2021-04-16
Velocity limit error

christianlandgraf:
### Description

Integration of https://github.com/PilzDE/pilz_industrial_motion/pull/328: Do not check joint model groups which contain fixed joints only (e.g. an empty endeffector group).


v4hn (approved):
Makes sense.

felixvd (approved):
I agree, makes sense.

PR #2627, opened 2021-04-28
document solution in ROS_ERROR on failed self-filtering

v4hn:
Also move the log command to the base class (which even uses the same logname...)

I just spend some time on https://github.com/ros-planning/moveit_tutorials/issues/192
and found a solution eventually, but this could have been a no-brainer with more information...


AndyZe (approved):
:+1: for consolidating the error messages in one place and :+1: for suggesting a solution, too

PR #2629, opened 2021-04-28
Prepare 1.1.3 for Noetic

tylerjw:
### Description

@nbbrooks @v4hn here is the patch release of some of the latest features for noetic to make sure they get in with the next release.  Please merge this using the rebase or squash option once it is ready and make sure the commit message is just 1.1.3.

PR #2633, opened 2021-04-29
Fix prerelease test

rhaschke:
Attempts to fix https://github.com/ros-planning/moveit/pull/2629#issuecomment-829472966


tylerjw (approved):
lgtm

tylerjw:
It looks like it worked!  https://github.com/ubi-agni/moveit/runs/2469467512

PR #2635, opened 2021-04-30
[prerelease test] Fix shell syntax

jspricke:
Don't use shell expansion for package names and to remove a directory tree

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,7 +43,7 @@ jobs:
       - name: "Free up disk space"
         if: ${{ matrix.env.CCOV }}
         run: |
-          sudo apt-get -qq purge build-essential ghc*
+          sudo apt-get -qq purge build-essential "ghc*"
           sudo apt-get clean
           # cleanup docker images not used by us
           docker system prune -af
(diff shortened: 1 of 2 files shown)


jspricke:
Just stumbled over this while chatting with @v4hn about #2633..

rhaschke:
I agree that we want to avoid shell expansion for debian package names. However, I think that both, `ghc*` vs `"ghc*"` trigger shell expansion, don't they? Shouldn't we use `'ghc*'`?
Regarding `/usr/local`: We want to keep `/usr/local`, but only want to remove its contents.

jspricke:
> I agree that we want to avoid shell expansion for debian package names. However, I think that both, `ghc*` vs `"ghc*"` trigger shell expansion, don't they? Shouldn't we use `'ghc*'`?

Doesn't make a difference, both work.

> Regarding `/usr/local`: We want to keep `/usr/local`, but only want to remove its contents.

"a package must not rely on the presence or absence of files or directories in /usr/local for normal operation."
https://www.debian.org/doc/debian-policy/ch-opersys.html#site-specific-programs

Why would you need it?

v4hn:
```
❯ mkdir a
❯ cd a
❯ touch bc
❯ ls "b*"
ls: cannot access 'b*': No such file or directory
❯ ls b*
bc
```

> Regarding /usr/local: We want to keep /usr/local, but only want to remove its contents.

We were just discussing that. If you're that fastidious, we should actually only delete `/usr/local/*/*` because the subfolders are also [part of the FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html). But in the end it shouldn't matter either way and I'd also prefer to remove `/usr/local` altogether if it's empty.

rhaschke (approved):
Ok, convinced. Thanks for the fixup.

PR #2639, opened 2021-04-30
Print error before returning

lucbettaieb:
### Description

Print an error indicating that the planning pipelines are empty before returning.  It was previously just switched around in the code, so it returned before getting a chance as to complain.

### Checklist


v4hn (approved):

v4hn:
I fail to understand how that ever passed review when it was added...

Thanks for pointing it out!

lucbettaieb:
> I fail to understand how that ever passed review when it was added...
> 
> Thanks for pointing it out!

No problem!  I think I was hitting it so I figured I'd make the fix.

Issue #2649, opened 2021-05-06
auto_create_ikfast_moveit_plugin.sh fails due to sympy0.7.1 dependency not available from pip any more. 


ags-dy:
### Description

I have used auto_create_ikfast_moveit_plugin.sh script in this repo to generate ikfast plugins in the past.  When I tried to run the same code today it failed during the build of the docker images because it failed to pip install sympy 0.7.1.  I assume this is due to further depreciation of python2. 

### Your environment
* ROS Distro: [Noetic]
* OS Version: e.g. Ubuntu 20.04
* Binary build 1.1.2-1focal.20210424.005508

### Steps to reproduce
```bash
$rosrun moveit_kinematics auto_create_ikfast_moveit_plugin.sh --iktype Transform6D panda_arm.urdf panda panda_link0 
```

### Expected behaviour
The ikfast docker image should install sympy

### Actual behaviour
pip install of sympy fails

### Backtrace or Console output
```bash
$rosrun moveit_kinematics auto_create_ikfast_moveit_plugin.sh --iktype Transform6D panda_arm.urdf panda panda_link0 panda_link8 
Converting urdf to Collada
Building docker image
Sending build context to Docker daemon   2.56kB
Step 1/3 : FROM personalrobotics/ros-openrave
 ---> 3470beb7a703
Step 2/3 : RUN apt-key adv --keyserver 'hkp://keyserver.ubuntu.com:80' --recv-key C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654 &&     apt-key del 421C365BD9FF1F717815A3895523BAEEB01FA116 &&     apt-get update &&     apt-get install -y --no-install-recommends python-pip build-essential liblapack-dev ros-indigo-collada-urdf &&     apt-get clean && rm -rf /var/lib/apt/lists/*
 ---> Using cache
 ---> 691f6b13328c
Step 3/3 : RUN pip install sympy==0.7.1
 ---> Running in cd6fe1646158
Downloading/unpacking sympy==0.7.1
  Cannot fetch index base URL https://pypi.python.org/simple/
  Could not find any downloads that satisfy the requirement sympy==0.7.1
Cleaning up...
No distributions at all found for sympy==0.7.1
Storing debug log for failure in /root/.pip/pip.log
The command '/bin/sh -c pip install sympy==0.7.1' returned a non-zero code: 1
```

Thanks

ags-dy:
[...]

PR #2650, opened 2021-05-06
Install sympy0.7.1 from git as it is no long available from pypi

ags-dy:
### Description
auto_create_ikfast_moveit_plugin.sh fails due to sympy0.7.1 dependency not available from pip any more.  Fixed by installing directly from github

https://github.com/ros-planning/moveit/issues/2649

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,15 +19,16 @@ jobs:
           - IMAGE: master-ci
             CCOV: true
           - IMAGE: master-ci-shadow-fixed
-            CATKIN_LINT: true
-          - IMAGE: melodic-ci
             IKFAST_TEST: true
+            CATKIN_LINT: true
             CLANG_TIDY: true
+          - IMAGE: melodic-ci
+            CATKIN_LINT: true
     env:
       DOCKER_IMAGE: moveit/moveit:${{ matrix.env.IMAGE }}
       UPSTREAM_WORKSPACE: moveit.rosinstall
+      # Pull any updates to the upstream workspace (since after restoring it from cache)
       AFTER_SETUP_UPSTREAM_WORKSPACE: vcs pull $BASEDIR/upstream_ws/src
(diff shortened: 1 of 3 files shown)


rhaschke (approved):
Thanks for providing this fix. I just also noticed that our CI tests for ikfast were not running anymore since the transition to GHA. Hence, I pushed another commit to fix this.

rhaschke:
@tylerjw: I just noticed that the configuration of upstream, target, and downstream workspaces in ci.yaml is totally screwed!
We just define `UPSTREAM_WORKSPACE=moveit.rosinstall`, which pulls in everything (including moveit itself!) into the upstream workspace. Thus, we are building `moveit` twice - once in upstream and once in target workspace!
`moveit.rosinstall` is essentially used by user source builds because this is referenced in our [install instructions](https://moveit.ros.org/install/source/). 
We definitely need to split packages into upstream and downstream workspaces and leave moveit for the target workspace only! 
I'm tempted to separate concerns regarding user source builds (using `moveit.rosinstall`) and CI, which should use two additional files `./.github/workflows/[downstream|upstream].rosinstall`. However, often we need to keep these in sync...

tylerjw:
I think we should modify our docker images to use copy to move the target repo into the image and remove moveit and downstream from the moveit.rosinstall (and rename it to upstream.rosinstall) and then we should add a downstream.rosinstall with the downstream repos.

rhaschke:
@tylerjw, I don't get you comment: Why do you want to "move the target repo into the (docker) image"? The docker images (ci / shadow-fixed) are fixed and just provide the downstream dependencies. The target repo (moveit here) is bind-mounted into the container. No need to copy or move anything.

rhaschke:
@tylerjw, I prepared some commits to separate downstream/upstream work packages and running a test [here](https://github.com/ubi-agni/moveit/runs/2520297380?check_suite_focus=true).

PR #2652, opened 2021-05-06
Use private NodeHandle instead of child for PlanningPipeline topics

### LINKED ISSUE #2578: Planned paths are not displayed in MotionPlanning plugin by default anymore
With #2127 merged, the topic `/move_group/display_planned_path` that the Rviz plugin subscribes to does not receive plans to display anymore. The plans now arrive on topics that are namespaced to each planning plugin (e.g. `/move_group/planning_pipelines/ompl/display_planned_path`).

henningkayser:
Fixes #2578.

#2127 introduced a child NodeHandle for initializing the PlanningPipeline. Since the display topics were generated from that one, the MotionPlanning panel wasn't able to receive any display markers from the "move_group" namespace anymore. By advertising the topics from a private NodeHandle (`"~"`) all markers and display messages are again published from the node's namespace, which in case of MoveGroup is "move_group". -> All fine!

### DIFF
--- a/moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
+++ b/moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
@@ -67,25 +67,26 @@ public:
    * this topic (visualization_msgs::MarkerArray) */
   static const std::string MOTION_CONTACTS_TOPIC;
 
-  /** \brief Given a robot model (\e model), a node handle (\e nh), initialize the planning pipeline.
+  /** \brief Given a robot model (\e model), a node handle (\e pipeline_nh), initialize the planning pipeline.
       \param model The robot model for which this pipeline is initialized.
-      \param nh The ROS node handle that should be used for reading parameters needed for configuration
+      \param pipeline_nh The ROS node handle that should be used for reading parameters needed for configuration
       \param planning_plugin_param_name The name of the ROS parameter under which the name of the planning plugin is
      specified
       \param adapter_plugins_param_name The name of the ROS parameter under which the names of the request adapter
      plugins are specified (plugin names separated by space; order matters)
   */
-  PlanningPipeline(const moveit::core::RobotModelConstPtr& model, const ros::NodeHandle& nh = ros::NodeHandle("~"),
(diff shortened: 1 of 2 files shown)


moveit_ros/planning/planning_pipeline/include/moveit/planning_pipeline/planning_pipeline.h
```diff
@@ -168,7 +168,11 @@ class PlanningPipeline
 private:
   void configure();
 
-  ros::NodeHandle nh_;
+  // The NodeHandle to initialize the PlanningPipeline (parameters plugins) with
+  ros::NodeHandle pipeline_nh_;
```
henningkayser, line 172: Alternatively, we could also just use `nh_` (like before) and `root_nh_` or so.

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -49,17 +49,17 @@ const std::string planning_pipeline::PlanningPipeline::MOTION_PLAN_REQUEST_TOPIC
 const std::string planning_pipeline::PlanningPipeline::MOTION_CONTACTS_TOPIC = "display_contacts";
 
 planning_pipeline::PlanningPipeline::PlanningPipeline(const moveit::core::RobotModelConstPtr& model,
-                                                      const ros::NodeHandle& nh,
+                                                      const ros::NodeHandle& pipeline_nh,
```
jliukkonen, line 52: So with `pipeline_nh_` defaulting to `ros::NodeHandle("~")` and `private_nh_` defaulting to `ros::NodeHandle("~")`, what's the difference? What am I missing?-)

Should the default be different? _Could_ the default be different?

henningkayser: MoveGroup or better MoveItCpp constructs multiple planning pipelines with child node handles in order to separate pipeline and plugin parameters. That one shouldn't be used for the topics, though, and that's why there is now a separate node handle which always uses the private "move_group" namespace. The default is not used at all afaik, but technically you could initialize a pipeline with a both node handles pointing to the private namespace.

v4hn (approved):
Thanks. You are right it makes much more sense than the solution I proposed.
Especially when using Pipelines from other nodes the private handle makes more sense.

PR #2657, opened 2021-05-07
Allow selecting planning pipeline in MotionSequenceAction

felixvd:
I was trying out the `MotionPlanSequenceAction` today, and noticed that the planning pipeline cannot be switched from the initially loaded one. That makes it impossible to use e.g. LIN motions if OMPL is the default pipeline.

This PR uses the changes introduced in #2127 to allow switching the pipeline. The limitation is that only a single pipeline can be used per MotionSequence, but realistically no one will use this except for chaining `pilz` motions together, so I don't think this is a problem. I also don't think spending development time on it is worth it.

The PR also exposes the `constructMotionPlanRequest` function in the Python `MoveGroupInterface`, because defining a `MotionPlanSequenceRequest` without it is not feasible (unless you use a package like [the one linked in the tutorial](https://ros-planning.github.io/moveit_tutorials/doc/pilz_industrial_motion_planner/pilz_industrial_motion_planner.html#sequence-of-multiple-segments), but our intern could not get it to run, so we abandoned the idea).

### Checklist

### DIFF
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -44,6 +44,7 @@ from moveit_msgs.msg import (
     MoveItErrorCodes,
     TrajectoryConstraints,
     PlannerInterfaceDescription,
+    MotionPlanRequest,
 )
 from sensor_msgs.msg import JointState
 import rospy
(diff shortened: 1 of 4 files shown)


felixvd:
Rebase and friendly ping, since `MotionSequenceAction`s currently can't be used unless the `pilz_industrial_motion_planner` is the only planner that is loaded.

jschleicher (approved):
LGTM. Thank you for the follow-up!

PR #2660, opened 2021-05-11
Fixup for #2596

simonschmeisser:
I forgot changing one include in #2596 which apparently does not bother the moveit-ci but does bother our internal ci

the relevant include is redundant as it's also included by `occupancy_map_monitor.h`, maybe that was the reason for MoveIt-CI being contempt

the include in the header could most likely be replaced by a forward decl, not sure if that would be worth the effort?


simonschmeisser:
@tylerjw I don't get this, now it fails with the error I was intending to fix with this commit which did not fail before??? The build log does not point to the version in this PR but to the state before?

v4hn:
It also fails [in other pull-requests](https://github.com/ros-planning/moveit/pull/2659/checks?check_run_id=2554757920) right now.

It probably found the old file from a system installation before, but I'm not sure about the details.
As this is an obvious fixup, I will merge this without CI.

PR #2663, opened 2021-05-12
Refine CSM::haveCompleteState

v4hn:
I looked at these methods in the context of #2661 and was shocked they were in such a bad state.

Unifying the different overloads, I slightly changed the semantics of the method (for the better of course :cat: ).
The details are explained in fbecdeb .

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -18,6 +18,7 @@ API changes in MoveIt releases
 - `RobotTrajectory` provides a copy constructor that allows a shallow (default) and deep copy of waypoints
 - Replace the redundant namespaces `robot_state::` and `robot_model::` with the actual namespace `moveit::core::`. The additional namespaces will disappear in the future (ROS2).
 - Moved the library `moveit_cpp` to `moveit_ros_planning`. The classes are now in namespace `moveit_cpp`, access via `moveit::planning_interface` is deprecated.
+- The joint states of `passive` joints must be published in ROS and the CurrentStateMonitor will now wait for them as well. Their semantics dictate that they cannot be actively controlled, but they must be known to use the full robot state in collision checks. (https://github.com/ros-planning/moveit/pull/2663)
 
 ## ROS Melodic
 
--- a/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
+++ b/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
@@ -100,6 +100,13 @@ public:
   bool haveCompleteState() const;
 
   /** @brief Query whether we have joint state information for all DOFs in the kinematic model
+   *  @param oldest_allowed_update_time All joint information must be from this time or more current
+   *  @return False if we have no joint state information for one of the joints or if our state
+   *  information is more than \e age old*/
+  bool haveCompleteState(const ros::Time& oldest_allowed_update_time) const;
+
+  /** @brief Query whether we have joint state information for all DOFs in the kinematic model
+   *  @param age Joint information must be at most this old
    *  @return False if we have no joint state information for one of the joints or if our state
    *  information is more than \e age old
    */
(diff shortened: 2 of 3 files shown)


moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/current_state_monitor.h
```diff
@@ -99,6 +99,12 @@ class CurrentStateMonitor
    */
   bool haveCompleteState() const;
 
+  /** @brief Query whether we have joint state information for all DOFs in the kinematic model
+   *  @param min_update_time All joint information must be from this time or more current
+   *  @return False if we have no joint state information for one of the joints or if our state
+   *  information is more than \e age old*/
+  bool haveCompleteState(const ros::Time& min_update_time) const;
```
felixvd, line 106:
```suggestion
  bool haveCompleteState(const ros::Time& oldest_allowed_update_time) const;
```
Same for the other occurrences. More verbosity, more good.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -169,109 +171,79 @@ void planning_scene_monitor::CurrentStateMonitor::stopStateMonitor()
...
+    else
+      continue;
+
+    if (missing_joints)
+      missing_joints->push_back(joint->getName());
+    else
+      return false;
   }
-  return result;
+  return (missing_joints == nullptr) || missing_joints->empty();
```
felixvd, line 243: Does this need an enclosing bracket?

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
```diff
@@ -284,16 +256,17 @@ bool planning_scene_monitor::CurrentStateMonitor::waitForCurrentState(const ros:
     elapsed = ros::WallTime::now() - start;
     if (elapsed > timeout)
     {
-      ROS_INFO_STREAM("Didn't received robot state (joint angles) with recent timestamp within "
-                      << wait_time << " seconds.\n"
-                      << "Check clock synchronization if your are running ROS across multiple machines!");
+      ROS_INFO_STREAM_NAMED(LOGNAME,
+                            "Didn't received robot state (joint angles) with recent timestamp within "
```
felixvd, line 260:
```suggestion
                            "Didn't receive robot state (joint angles) with recent timestamp within "
```

felixvd (approved):
LGTM, thanks for the cleanup! Will rebase #2661 when this is merged.

moveit_ros/planning/planning_scene_monitor/src/current_state_monitor.cpp
v4hn: No, but I prefer them for clarity.

felixvd: I actually meant *another* closing bracket around all of it. But that works :)

felixvd:
I'm good with merging this if you're done

rhaschke (approved):
Nice cleanup. My only nitpick suggestion is to `inline` the new overloads of `haveCompleteState`.

rhaschke:
Aside from the newly introduced syntax error, this looks good to me. @v4hn, please merge yourself the way you like.

MIGRATION.md
```diff
@@ -18,6 +18,7 @@ API changes in MoveIt releases
 - `RobotTrajectory` provides a copy constructor that allows a shallow (default) and deep copy of waypoints
 - Replace the redundant namespaces `robot_state::` and `robot_model::` with the actual namespace `moveit::core::`. The additional namespaces will disappear in the future (ROS2).
 - Moved the library `moveit_cpp` to `moveit_ros_planning`. The classes are now in namespace `moveit_cpp`, access via `moveit::planning_interface` is deprecated.
+- The joint states of `passive` joints must be published in ROS and the CurrentStateMonitor will now wait for them as well. Their semantics dictate that they cannot be actively controlled, but they must be known to use the full robot state in collision checks. (https://github.com/ros-planning/moveit/pull/2663)
```
simonschmeisser, line 21: @v4hn 
I have two problems with this:
1. I don't know what I actually need to change after reading this
2. I'm not sure it's actually true (and certainly not desirable) since you only iterate over `getActiveJointModels` below

(We had long discussions with IPA folks if a gripper should publish `gripper_opening` or rather `finger_left` (ie 0.5*opening) but I think nobody ever came up with having to publish `finger_right` (a mimic joint) as well)

v4hn: > I don't know what I actually need to change after reading this

Doesn't "The joint states of `passive` joints must be published in ROS" say what needs to be done?
Please propose a better wording for this entry if you think more is needed.

Are you sure this affects what you think it does? Mimic joints are **not** passive joints in MoveIt's terminology. Passive joints have to be declared in the srdf via `<passive_joint>`.
The latter are actually also part of the group of "active" joint models (yes I know this naming doesn't make much sense, I didn't pick it).
Passive joint are actively moving joints we can't update via mimic joints or can't even control on our own.
Levers or flexible joints that swing downwards towards gravity might be examples.

I'm happy to be contradicted and learn how you use this joint tag differently.

> We had long discussions with IPA folks if a gripper should publish gripper_opening or rather finger_left (ie 0.5*opening) 

Either is possible as long as the joint actuation interfaces accept the same joint(s).
All other relevant joints in the gripper should be computable via mimic joints.
If both fingers can be actuated independently that would of course be a different story.
It would be nice if we could actually have such coupled joint groups modelled in a way where specifying either is enough to infer the other (also for non-linear translations).

simonschmeisser: Thanks! Now it makes sense ... somewhat ... I didn't know about the `passive` tag and category. I think people who use it will understand your migration notes. I misread this as mimic joints being required to be published and current with this PR

PR #2664, opened 2021-05-12
add warning if interpolation parameter is out of range [0, 1]

JStech:
### Description

Added warnings if the interpolation parameter `t` is outside the range [0, 1] or NaN. Would have helped in debugging something recently.

Maybe it would be good to also clamp the value to [0, 1] (and make a NaN a 0 or something)?

### Checklist

### DIFF
--- a/moveit_core/robot_model/include/moveit/robot_model/robot_model.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/robot_model.h
@@ -61,7 +61,7 @@ namespace core
 {
 MOVEIT_CLASS_FORWARD(RobotModel);  // Defines RobotModelPtr, ConstPtr, WeakPtr... etc
 
-static inline void checkInterpolationParamBounds(const char LOGNAME[], const double& t)
+static inline void checkInterpolationParamBounds(const char LOGNAME[], double t)
 {
   if (std::isnan(t) || std::isinf(t))
   {
--- a/moveit_core/robot_model/include/moveit/robot_model/robot_model.h
+++ b/moveit_core/robot_model/include/moveit/robot_model/robot_model.h
@@ -61,6 +61,15 @@ namespace core
 {
 MOVEIT_CLASS_FORWARD(RobotModel);  // Defines RobotModelPtr, ConstPtr, WeakPtr... etc
 
+static inline void checkInterpolationParamBounds(const char LOGNAME[], const double& t)
+{
+  if (std::isnan(t) || std::isinf(t))
+  {
+    throw Exception("Interpolation parameter is NaN or inf.");
+  }
+  ROS_WARN_STREAM_COND_NAMED(t < 0. || t > 1., LOGNAME, "Interpolation parameter is not in the range [0, 1]: " << t);
+}
+
 /** \brief Definition of a kinematic model. This class is not thread
     safe, however multiple instances can be created */
 class RobotModel
--- a/moveit_core/robot_model/src/robot_model.cpp
+++ b/moveit_core/robot_model/src/robot_model.cpp
@@ -49,7 +49,10 @@ namespace moveit
 {
 namespace core
 {
-const std::string LOGNAME = "robot_model";
+namespace
+{
+constexpr char LOGNAME[] = "robot_model";
+}  // namespace
 
 RobotModel::RobotModel(const urdf::ModelInterfaceSharedPtr& urdf_model, const srdf::ModelConstSharedPtr& srdf_model)
 {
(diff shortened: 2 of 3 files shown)


moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1290,6 +1290,8 @@ double RobotModel::distance(const double* state1, const double* state2) const
 
 void RobotModel::interpolate(const double* from, const double* to, double t, double* state) const
 {
+  ROS_WARN_STREAM_COND_NAMED(std::isnan(t) || t < 0. || t > 1., LOGNAME,
```
jliukkonen-pn, line 1293: Would it make sense to add `std::isinf` check too? Also, typo in word `parameter`.

jliukkonen-pn (commented):
Very nice improvement!

moveit_core/robot_model/src/robot_model.cpp
jliukkonen-pn: Actually, should this whole function just fail if the fraction `t` is not reasonable? Print error and return instead of giving a warning? After all, there was a crash when `t` was `NaN`. :thinking:

JStech: I believe infinite values will be caught by `t < 0. || t > 1.`. Actually, I think a NaN would also, but I'm not sure that's true across all platforms, etc. Thanks for the spelling catch.

JStech: I was on the fence about how strongly it should react to an incorrect value. Since it's a `void` return type, failure isn't really an option, just complaining and doing nothing. (Or, maybe, throwing an exception?)

felixvd: They should probably throw an exception. Interpolating from NaNs and infs sounds dangerous. What about an `assert`?

rhaschke: Values smaller than 0 and larger than 1 are perfectly fine: You just extrapolate beyond the point. However, a warning in this case is probably useful. `nan` and `inf` could throw.

rhaschke (commented):
Instead of repeating the same code snippet several times, I suggest providing an inline function for this sanity check.

v4hn:
Funny and frustrating how many comments such a minor patch can generate...

I support throwing an exception for `nan` and `inf` values,
though one could argue that the valid range is documented in doxygen and thus valid by contract.
(We are not writing Ada though, so there's no way to check contracts :)).
Printing a warning in this case and still running invalid computations is not an option for me.

I'm actually a fan of extrapolation-support here because it is supported for free and thus would allow all valid double values here,
but we also provide `RobotState::integrateVariableVelocity` for related use-cases.
So if you insist on restricting the use of the method to its documented specification it should throw for values outside [0;1] as well.

moveit_core/robot_model/src/robot_model.cpp
jliukkonen-pn: > Values smaller than 0 and larger than 1 are perfectly fine: You just extrapolate beyond the point. However, a warning in this case is probably useful. `nan` and `inf` could throw.

Thumbs up for this approach.

JStech:
> Instead of repeating the same code snippet several times, I suggest providing an inline function for this sanity check.

Where do you recommend I put such a function? Maybe `moveit_core/macros` somewhere?

rhaschke:
> Where do you recommend I put such a function? Maybe `moveit_core/macros` somewhere?

I would just put it into the same .cpp file. We don't need to expose this function, do we?

JStech:
> I would just put it into the same .cpp file. We don't need to expose this function, do we?

But I'm checking in both `robot_state.cpp` and `robot_model.cpp`--so define it in both? Another "local" option would be to put it in `robot_model.h`, which is included in both.

Also, one of the functions in `robot_state.cpp` then immediately calls the function in `robot_model.cpp`, so we don't really need the first check, although it'd probably be easier to interpret the error if it comes from the outer function.

I feel like all of this is a matter of style, and I don't see a strong reason to do it one way or another. Just tell me what will get approved :-P.

JStech:
I defined the check in `robot_model.h` and it's used in both RobotState and RobotModel

moveit_core/robot_model/include/moveit/robot_model/robot_model.h
```diff
@@ -61,6 +61,15 @@ namespace core
 {
 MOVEIT_CLASS_FORWARD(RobotModel);  // Defines RobotModelPtr, ConstPtr, WeakPtr... etc
 
+static inline void checkInterpolationParamBounds(const char LOGNAME[], const double& t)
```
rhaschke, line 64: Don't pass simple data types as double by reference:
https://stackoverflow.com/questions/7561690/why-do-we-not-pass-pod-by-reference-in-functions

rhaschke (commented):
One more nitpick. Otherwise, I approve.

moveit_core/robot_model/include/moveit/robot_model/robot_model.h
```diff
@@ -61,7 +61,7 @@ namespace core
 {
 MOVEIT_CLASS_FORWARD(RobotModel);  // Defines RobotModelPtr, ConstPtr, WeakPtr... etc
 
-static inline void checkInterpolationParamBounds(const char LOGNAME[], const double& t)
+static inline void checkInterpolationParamBounds(const char LOGNAME[], double t)
```
rhaschke, line 64: You could have kept `const`. But fine now, anyway.

rhaschke (approved):

PR #2665, opened 2021-05-12
New RobotState interpolation test

JStech:
### Description

I started writing these tests when I thought there was a bug in the interpolation code. There wasn't, I was just passing a bad parameter (hence #2664), but the tests are still worthwhile.

### Checklist

### DIFF
--- a/moveit_core/robot_state/test/robot_state_test.cpp
+++ b/moveit_core/robot_state/test/robot_state_test.cpp
@@ -658,12 +658,6 @@ TEST_F(OneRobot, testInterpolation)
         << "Interpolation of mimic joint mim_f failed.";
   }
 
-  std::cout << "Checking interpolation input verification. Two warnings (untested) and an exception are expected."
-            << std::endl;
-
-  state_a.interpolate(state_b, 1.1, interpolated_state, robot_model_->getJointModelGroup("base_from_base_to_e"));
-  state_a.interpolate(state_b, -0.1, interpolated_state, robot_model_->getJointModelGroup("base_from_base_to_e"));
-
   bool nan_exception = false;
   try
   {
@@ -671,7 +665,7 @@ TEST_F(OneRobot, testInterpolation)
   }
   catch (std::exception& e)
   {
-    std::cout << "Caught: " << e.what() << std::endl;
+    std::cout << "Caught expected exception: " << e.what() << std::endl;
     nan_exception = true;
   }
   EXPECT_TRUE(nan_exception) << "NaN interpolation parameter did not create expected exception.";
--- a/moveit_core/robot_state/test/robot_state_test.cpp
+++ b/moveit_core/robot_state/test/robot_state_test.cpp
@@ -636,6 +636,24 @@ TEST_F(OneRobot, testInterpolation)
     EXPECT_NEAR(1 - t, interpolated_state.getVariablePosition("joint_f"), 1e-9);
     EXPECT_NEAR(1.5 * (1 - t) + 0.1, interpolated_state.getVariablePosition("mim_f"), 1e-9);
   }
+
+  std::cout << "Checking interpolation input verification. Two warnings (untested) and an exception are expected."
+            << std::endl;
+
+  state_a.interpolate(state_b, 1.1, interpolated_state, robot_model_->getJointModelGroup("base_from_base_to_e"));
+  state_a.interpolate(state_b, -0.1, interpolated_state, robot_model_->getJointModelGroup("base_from_base_to_e"));
+
+  bool nan_exception = false;
+  try
+  {
+    state_a.interpolate(state_b, 1. / 0., interpolated_state, robot_model_->getJointModelGroup("base_from_base_to_e"));
+  }
+  catch (std::exception& e)
+  {
+    std::cout << "Caught: " << e.what() << std::endl;
+    nan_exception = true;
+  }
+  EXPECT_TRUE(nan_exception);
 }
 
 int main(int argc, char** argv)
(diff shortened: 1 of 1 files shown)


v4hn:
It's interesting to see new tests *decreasing* coverage... How did that happen?

Yay tests!

The only thing I would do differently is specifying `theta` values as fractions of pi/tau.

JStech:
> It's interesting to see new tests _decreasing_ coverage... How did that happen?

Apparently I write really bad tests.

I asked @tylerjw and he said, "We report codecov on a relwithdebinfo build because debug builds time out some of the tests.  That means that the codecov report has some error."

I'll update the angles.

JStech:
Also, I'm leaving this as a draft until #2664 is settled, and then I'll test that behavior too.

v4hn:
> We report codecov on a relwithdebinfo build because debug builds time out some of the tests. That means that the codecov report has some error.

If that's the reason and the tests you add actually are in that margin of error, I would vote to disable this code coverage report altogether or build an additional job as debug build for accurate information.
Since it was introduced the bot adds uninformative and very similar reports everywhere and if it's not even accurate enough to increase coverage when only new tests are added (whether or not they actually add newly covered lines) it's entirely useless...

tylerjw:
> whether or not they actually add newly covered lines

This might also be why the PR isn't increasing the coverage.  We don't report any lines in the test itself as part of coverage (only lines outside of the tests).

v4hn:
> This might also be why the PR isn't increasing the coverage.

That might be, but then it shouldn't *decrease* either. :)

> We don't report any lines in the test itself as part of coverage (only lines outside of the tests).

But I guess the tests themselves should also be excluded from 100% coverage, aren't they?
Sorry for stupid questions, I never looked into our coverage testing.

moveit_core/robot_state/test/robot_state_test.cpp
```diff
@@ -560,6 +560,84 @@ TEST_F(OneRobot, testPrintCurrentPositionWithJointLimits)
...
+
+  // Interpolate with itself
+  state_a.setToDefaultValues();
+  moveit::core::RobotState state_b(state_a);
+  moveit::core::RobotState interpolated_state(state_a);
+  for (size_t i = 0; i <= 10; ++i)
+  {
+    state_a.interpolate(state_b, static_cast<double>(i) / 10., interpolated_state,
+                        robot_model_->getJointModelGroup("base_from_base_to_e"));
+    EXPECT_TRUE(state_a.distance(state_b) < 1e-12);
```
tylerjw, line 575: Do you mind using EXPECT_NEAR here and defining the epsilon as a const at the top of the file?

Also, do you mind leaving a comment explaining what might have gone wrong if this fails to help anyone who does get this test to fail in the future.  You can do something like this:

```
EXPECT_NEAR(state_a.distance(state_b), 0, 1e-12) << "Reason this might fail";
```

moveit_core/robot_state/test/robot_state_test.cpp
```diff
@@ -560,6 +560,84 @@ TEST_F(OneRobot, testPrintCurrentPositionWithJointLimits)
...
+
+    for (const auto& link_name : robot_model_->getLinkModelNames())
+    {
+      EXPECT_FALSE(interpolated_state.getCollisionBodyTransform(link_name, 0).matrix().hasNaN());
```
tylerjw, line 579: Do you mind leaving a reason here too?

tylerjw (approved):
I generally approve of new tests, just a couple small comments.

moveit_core/robot_state/test/robot_state_test.cpp
JStech: Done. I didn't know that was possible with gtest--thanks for showing me

moveit_core/robot_state/test/robot_state_test.cpp
JStech: Done. I tried to add reasons everywhere, but some of them are probably redundant. Off the top of my head I can't imagine why some of these tests would fail. This particular check for NaNs was inspired by the original issue behind all this, which was because I was calling `interpolate` with a NaN interpolation parameter, which ended up littering the collision body transform with NaNs. Now, that particular error would be caught by the new check from #2664.

JStech:
I updated this to also check that `interpolate` throws an exception when the interpolation parameter is NaN. I also call it with 1.1 and -0.1, which generates warnings, although this behavior isn't actually tested, as it'd require setting up a `rosout` subscriber and verifying that the warnings are printed (which is doable, but it'd be some work--I'm open to pursuing it if the community thinks it's worthwhile).

JStech:
@tylerjw @v4hn any more comments? I think this is good to go

moveit_core/robot_state/test/robot_state_test.cpp
```diff
@@ -636,6 +636,24 @@ TEST_F(OneRobot, testInterpolation)
     EXPECT_NEAR(1 - t, interpolated_state.getVariablePosition("joint_f"), 1e-9);
     EXPECT_NEAR(1.5 * (1 - t) + 0.1, interpolated_state.getVariablePosition("mim_f"), 1e-9);
   }
+
+  std::cout << "Checking interpolation input verification. Two warnings (untested) and an exception are expected."
```
v4hn, line 661: well, "untested" really doesn't make any sense in a test. :-)
Also the exception is not visible on the command line (you catch it), so there's no need to mention it at all.

v4hn (approved):
Looks good after you addressed my comment on the `cout` message.

moveit_core/robot_state/test/robot_state_test.cpp
JStech: What I'm saying is that lines 664 and 665 _should_ produce warnings, but I didn't actually add the `/rosout_agg` listener to verify that they do. So it's code that's in a test but if it doesn't behave properly, the test will still pass (maybe not the best approach, now that I think about it).

As for the exception, I do catch it but I also print out the error message.

v4hn: Ah, sorry for the confusion. Your explanations make a lot of sense.
"untested" is a word you really don't want to see in tests. :-)

Testing extrapolation can make sense, but testing for the warnings does not seem too useful in my opinion.

Thanks for adding the clarification on the cout output as well.

I'll go ahead and merge.

PR #1790, opened 2019-11-28
add function to set cartesian speed of the end effector for a given trajectory

2scholz:
See #1556 for previous discussion.
I used the solution proposed by @kwh5484 as a basis.

To not duplicate any code I made `trajectory_processing::updateTrajectory` public and used it to adjust the velocities and accelerations of the waypoints according to the new time stamps.

PR #2674, opened 2021-05-19
Cartesian end effector speed setting (extension)

### LINKED ISSUE #1065: set_max_{velocity,acceleration}_scaling_factor not working in compute_cartesian_path
Why use 1.0 (max velocity) there?
https://github.com/ros-planning/moveit/blob/54d516a4e6fdc5db9dfcba4da9e916d807ea312a/moveit_ros/move_group/src/default_capabilities/cartesian_path_service_capability.cpp#L158

Thieso:
### Description

Further work on the PR #1790. 
Extended the mentioned PR by making the setting of cartesian end effector speed also usable for the function `computeCartesianPath`. 
Furthermore, I wrapped the `setMaxCartesianEndEffectorSpeed()` to be able to be used in the Python interface of Moveit. 

### Checklist

Fixes #1065


Thieso:
Also added an extension to moveit_msgs [here](https://github.com/ros-planning/moveit_msgs/pull/113) to enable the parameters to flow to the right place.

JStech (commented):
Thanks for picking this up. I made a few small suggestions. I wasn't sure how to run this to see it in action--could you explain how to run this?

moveit_core/trajectory_processing/src/cartesian_speed.cpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+    else
+    {
+      ROS_WARN_ONCE_NAMED("trajectory_processing.cartesian_speed",
+                          "Desired cartesian end-effector speed is not reached because of joint velocity constraints.");
+      time_diff[i] = old_time_diff;
+    }
+    // update the current_end_effector_state for next iteration
+    current_end_effector_state = next_end_effector_state;
+  }
```
schornakj, line 105: I'd suggest refactoring to keep more variables within the local scope of the `for` loop. The loop is basically iterating over pairs of waypoint indices, and since the state transforms depend on new values of `kinematic_state` that are gotten for each waypoint index, I think it would be cleaner to just determine new values of current and next based on the index i for the current waypoint.

Something like this could be one solution:
```suggestion
  double new_time_diff, old_time_diff;
  std::vector<double> time_diff(num_waypoints - 1, 0.0);

  for (size_t i = 0; i < num_waypoints - 1; i++)
  {
    Eigen::Isometry3d current_end_effector_state = trajectory.getWayPointPtr(i)->getGlobalLinkTransform(end_effector); 
    Eigen::Isometry3d next_end_effector_state = trajectory.getWayPointPtr(i + 1)->getGlobalLinkTransform(end_effector); 

    // get euclidean distance between the two waypoints
    double euclidean_distance =
        pow(pow(next_end_effector_state.translation()[0] - current_end_effector_state.translation()[0], 2) +
                pow(next_end_effector_state.translation()[1] - current_end_effector_state.translation()[1], 2) +
                pow(next_end_effector_state.translation()[2] - current_end_effector_state.translation()[2], 2),
            0.5);

    new_time_diff = (euclidean_distance / speed);
    old_time_diff = trajectory.getWayPointDurationFromPrevious(i + 1);

    // if constraints allow save the new time difference between waypoints
    if (new_time_diff > old_time_diff)
    {
      time_diff[i] = new_time_diff;
    }
    else
    {
      ROS_WARN_ONCE_NAMED("trajectory_processing.cartesian_speed",
                          "Desired cartesian end-effector speed is not reached because of joint velocity constraints.");
      time_diff[i] = old_time_diff;
    }
  }
```

moveit_core/trajectory_processing/test/test_cartesian_speed.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  for (size_t i = 0; i < num_waypoints - 1; i++)
+  {
+    kinematic_state = TRAJECTORY.getWayPointPtr(i + 1);
+    next_end_effector_state = kinematic_state->getGlobalLinkTransform("r_wrist_roll_link");
+    euclidean_distance +=
+        pow(pow(next_end_effector_state.translation()[0] - current_end_effector_state.translation()[0], 2) +
+                pow(next_end_effector_state.translation()[1] - current_end_effector_state.translation()[1], 2) +
+                pow(next_end_effector_state.translation()[2] - current_end_effector_state.translation()[2], 2),
+            0.5);
```
schornakj, line 132: Earlier comment about Eigen vector norm calculations also apply here.

Thieso:
> Thanks for picking this up. I made a few small suggestions. I wasn't sure how to run this to see it in action--could you explain how to run this?

It is actually pretty easy to run this. 
For example in Python when using cartesian paths you would normally call 
```
move_group.compute_cartesian_path(waypoints)
```
now if you want to restrict the cartesian velocity you can just call
```
move_group.set_max_cartesian_end_effector_speed(speed, "ee_link_name")
move_group.compute_cartesian_path(waypoints)
move_group.clear_max_cartesian_end_effector_speed()
```
Similarly you can use the function for point to point motions. 

Hope this helps and thanks for the suggestions.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -269,6 +269,15 @@ class MoveGroupInterface
       If the value is greater than 1, it is set to 1.0. */
   void setMaxAccelerationScalingFactor(double max_acceleration_scaling_factor);
 
+  /** \brief Set the maximum cartesian speed for the end effector link.
+      The unit of the speed is meters per second and needs to be greater than 0.
+      The desired cartesian speed can only be adhered to in parts of the trajectory
+      where the original cartesian speed was higher.*/
```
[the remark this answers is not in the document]
Thieso: Exactly correct. This is implemented this way so that the maximum joint velocities and acceleration are not exceeded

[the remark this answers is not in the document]
JStech: OK. Maybe it'd be clearer to say, "The desired speed is a maximum bound. Slower parts of the trajectory will be left unchanged." You could also rename the parameter `max_speed`.

moveit_core/trajectory_processing/test/test_cartesian_speed.cpp
```diff
@@ -0,0 +1,143 @@
...
+    current_end_effector_state = next_end_effector_state;
+  }
+  double avg_speed = euclidean_distance / TRAJECTORY.getWayPointDurationFromStart(num_waypoints);
+  ASSERT_DOUBLE_EQ(0.01, avg_speed);
+}
```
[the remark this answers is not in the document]
Thieso: Hi, 
I do not quite understand what you mean by hand calculated. If I would calculate the distance by hand I would use the exact same formula since it is the mathematically correct formula. 
So can you please specify a bit more what you mean by hand calculated? 
Thanks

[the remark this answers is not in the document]
gautz: Maybe the meaning is to hard-code a reference trajectory to test the overall output? For instance with 3-4 trajectory points? Would this make sense?

gautz:
I just successfully tested this with the move_group c++ interface and OMPL Constrained Planning. I can further test specific things if required.

moveit_core/trajectory_processing/test/test_cartesian_speed.cpp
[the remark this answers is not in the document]
schornakj: > Maybe the meaning is to hard-code a reference trajectory to test the overall output? For instance with 3-4 trajectory points? Would this make sense?

This is closer to what I meant. Comparing the output from the function against predetermined expected values is good practice.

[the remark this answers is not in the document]
Thieso: Ok sounds good. Then I will define some points in cartesian space and plan the cartesian trajectory and then it is easy to compare. 
The problem I am facing is that I don't know how to load the IK solver in the test script so that I can use plan in cartesian space. I am not familiar enough with the Moveit code base for that. Can you give me a hint where to find that information?

JStech:
\> It is actually pretty easy to run this.
> For example in Python when using cartesian paths you would normally call
> 
> ```
> move_group.compute_cartesian_path(waypoints)
> ```
> 
> now if you want to restrict the cartesian velocity you can just call
> 
> ```
> move_group.set_max_cartesian_end_effector_speed(speed, ee_link_name)
> move_group.compute_cartesian_path(waypoints)
> move_group.clear_max_cartesian_end_effector_speed()
> ```

Thanks--I was able to drop this right into the [Move Group Python interface tutorial](https://ros-planning.github.io/moveit_tutorials/doc/move_group_python_interface/move_group_python_interface_tutorial.html) and it worked great.

I'm not sure why pre-commit is hanging. I ran it locally and there are some formatting changes that it will require. You can install it via `pip` and then run `pre-commit run --all-files` in your worktree.

gautz:
Here is a small application example using 
``move_group.setMaxCartesianEndEffectorSpeed(cartesian_velocity_, welding_tcp_frame_);``
The length of the path is 0.145m.

1. ``cartesian_velocity_=0.05; // [m/s]``
cartesian_speed_0_05
2. ``cartesian_velocity_=0.5; // [m/s]``
catesian_speed_0_5

Thieso:
> I'm not sure why pre-commit is hanging. I ran it locally and there are some formatting changes that it will require. You can install it via `pip` and then run `pre-commit run --all-files` in your worktree.

Ok thanks for the hint. I ran the it and commited the changes. 

> Here is a small application example using
> `move_group.setMaxCartesianEndEffectorSpeed(cartesian_velocity_, welding_tcp_frame_);`
> The length of the path is 0.145m.

Thanks for testing it!

rhaschke:
> I'm not sure why pre-commit is hanging.

As this PR comes from a new contributor, GHA requires the explicit approval by a maintainer...

moveit_core/trajectory_processing/include/moveit/trajectory_processing/cartesian_speed.h
```diff
@@ -0,0 +1,47 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Benjamin Scholz */
+
+#pragma once
+
+#include <moveit/robot_trajectory/robot_trajectory.h>
+#include <moveit/robot_state/robot_state.h>
+#include <moveit/trajectory_processing/iterative_time_parameterization.h>
```
rhaschke, line 41: These two includes are not needed for this header. Please only include what is really needed.

rhaschke (commented):
I generally approve. Please consider the minor remark.
To get CI going, you could adapt https://github.com/ros-planning/moveit/blob/master/.github/workflows/upstream.rosinstall to use your moveit_msgs branch. Please clearly mark a corresponding commit as temporary!

moveit_core/trajectory_processing/include/moveit/trajectory_processing/cartesian_speed.h
v4hn: You actually don't need any proper moveit include here, just
```suggestion
#include <moveit/macros/class_forward.h>
#include <string>
namespace trajectory_processing {
  MOVEIT_CLASS_FORWARD(RobotTrajectory);;
}  // namespace trajectory_processing

File: moveit_core/trajectory_processing/src/cartesian_speed.cpp [hunk 6]
```diff
@@ -0,0 +1,106 @@
...
+  if (end_effector.empty())
+  {
+    std::vector<std::string> tips;
+    trajectory.getGroup()->getEndEffectorTips(tips);
+    if (tips.empty())
+    {
+      ROS_ERROR_STREAM_NAMED(LOGGER_NAME, "No end effector defined for group attached to trajectory.");
+      return false;
+    }
+    end_effector = tips[0];
```
v4hn, line 61: please add a matching `ROS_INFO` if more than one tip exists.

moveit_core/trajectory_processing/src/cartesian_speed.cpp
```diff
@@ -0,0 +1,106 @@
...
+  double euclidean_distance, new_time_diff, old_time_diff;
+  std::vector<double> time_diff(num_waypoints - 1, 0.0);
+
+  for (size_t i = 0; i < num_waypoints - 1; i++)
+  {
+    // get end effector state for current waypoint
+    Eigen::Isometry3d current_end_effector_state = trajectory.getWayPointPtr(i)->getGlobalLinkTransform(end_effector);
+
+    // get end effector state for next waypoint
+    Eigen::Isometry3d next_end_effector_state = trajectory.getWayPointPtr(i + 1)->getGlobalLinkTransform(end_effector);
```
v4hn, line 80: these should both be `const Eigen::Isometry3d&`

moveit_core/trajectory_processing/include/moveit/trajectory_processing/cartesian_speed.h
```diff
@@ -0,0 +1,47 @@
...
+
+namespace trajectory_processing
+{
+bool setMaxCartesianEndEffectorSpeed(robot_trajectory::RobotTrajectory& trajectory, const double speed,
+                                     std::string end_effector = "");
```
v4hn, line 46: Please provide an overload that takes `const LinkModel*` instead of the string and use this under the hood for the version that uses the string.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/cartesian_speed.h
```diff
@@ -0,0 +1,47 @@
...
+
+#pragma once
+
+#include <moveit/robot_trajectory/robot_trajectory.h>
+#include <moveit/robot_state/robot_state.h>
+#include <moveit/trajectory_processing/iterative_time_parameterization.h>
+
+namespace trajectory_processing
+{
+bool setMaxCartesianEndEffectorSpeed(robot_trajectory::RobotTrajectory& trajectory, const double speed,
```
v4hn, line 45: The method is not limited to end effectors, so I would propose to name it `setMaxCartesianLinkSpeed` instead.
This would of course imply renaming the message field, sorry for merging this before I did a full review here.
I overestimated @rhaschke's review :slightly_smiling_face: .

moveit_core/trajectory_processing/src/cartesian_speed.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    // if constraints allow save the new time difference between waypoints
+    if (new_time_diff > old_time_diff)
+    {
+      time_diff[i] = new_time_diff;
+    }
+    else
+    {
+      ROS_WARN_ONCE_NAMED(LOGGER_NAME,
+                          "Desired cartesian end-effector speed is not reached because of joint velocity constraints.");
```
v4hn, line 96: I'm not happy with this being a `ONCE` warning in a multi-purpose method.
Either this is a critical failure condition for the method (in which case it should not be called `Max`), or it is perfectly fine behavior.
In this case, we could report it to the user by returning a `double` value that represents the **slowest** Cartesian speed in the adapted trajectory.

moveit_core/trajectory_processing/test/test_cartesian_speed.cpp
```diff
@@ -0,0 +1,184 @@
...
+  state.setVariablePosition("panda_joint2", 0.15980591412916012);
+  state.setVariablePosition("panda_joint3", -0.000269206763021151);
+  state.setVariablePosition("panda_joint4", -1.4550637907602342);
+  state.setVariablePosition("panda_joint5", 0.0006907805230834268);
+  state.setVariablePosition("panda_joint6", 1.61442119426705);
+  state.setVariablePosition("panda_joint7", 0.7845267455355481);
+  trajectory.addSuffixWayPoint(state, 0.0);
+  WAYPOINT_DISTANCES.push_back(0.3);
+
+  // Second waypoint (+0.3 m in Y direction)
```
v4hn, line 93: Are these waypoints *incremental* or all w.r.t. the initial waypoint?
I think it might be useful to give the actual x/y/z values in the base frame here as well if the model ever changes....

moveit_core/trajectory_processing/include/moveit/trajectory_processing/cartesian_speed.h
Thieso: Actually, the `iterative_time_parametrization.h` is needed for the call to `updateTrajectory()` after adapting the trajectory times. but otherwise you are right. Thanks

moveit_core/trajectory_processing/src/cartesian_speed.cpp
```diff
@@ -0,0 +1,135 @@
...
+{
+  // In case the link name is not set, retrieve an end effector name from the
+  // joint model group specified in the robot trajectory
+  if (link_name.empty())
+  {
+    std::vector<std::string> tips;
+    trajectory.getGroup()->getEndEffectorTips(tips);
+    if (tips.empty())
+    {
+      ROS_ERROR_STREAM_NAMED(LOGGER_NAME, "No end effector defined for group attached to trajectory.");
```
felixvd, line 56:
```suggestion
      ROS_ERROR_STREAM_NAMED(LOGGER_NAME, "No end effector defined for group attached to trajectory, cannot set max cartesian link speed without argument.");
```

moveit_core/trajectory_processing/src/cartesian_speed.cpp
```diff
@@ -0,0 +1,135 @@
...
+  // Get link model based on given end link
+  if (!trajectory.getGroup()->hasLinkModel(link_name))
+    ROS_ERROR_STREAM_NAMED(LOGGER_NAME, "Link model was not specified in the robot trajectory");
+  const moveit::core::LinkModel* link_model = trajectory.getGroup()->getLinkModel(link_name);
+
+  // Call function for speed setting using the created link model
+  return setMaxCartesianLinkSpeed(trajectory, max_speed, link_model);
+}
+
+bool setMaxCartesianLinkSpeed(robot_trajectory::RobotTrajectory& trajectory, const double max_speed,
```
felixvd, line 74:
```suggestion
bool limitMaxCartesianLinkSpeed(robot_trajectory::RobotTrajectory& trajectory, const double max_speed,
```
This has the same name as the MGI function, but it actually does something else. What about this name, so the effect is clearer?

moveit_core/trajectory_processing/src/cartesian_speed.cpp
```diff
@@ -0,0 +1,135 @@
...
+    Eigen::Isometry3d current_link_state = trajectory.getWayPointPtr(i)->getGlobalLinkTransform(link_name);
+    Eigen::Isometry3d next_link_state = trajectory.getWayPointPtr(i + 1)->getGlobalLinkTransform(link_name);
+
+    // get euclidean distance between the two waypoints
+    euclidean_distance = (next_link_state.translation() - current_link_state.translation()).norm();
+
+    new_time_diff = (euclidean_distance / max_speed);
+    old_time_diff = trajectory.getWayPointDurationFromPrevious(i + 1);
+
+    // if constraints allow save the new time difference between waypoints
```
felixvd, line 109:
```suggestion
    // if constraints allow, save the new time difference between waypoints
```

moveit_core/trajectory_processing/test/test_cartesian_speed.cpp
felixvd: Imo this is fine as is.

moveit_core/trajectory_processing/src/cartesian_speed.cpp
```diff
@@ -0,0 +1,135 @@
...
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of Willow Garage nor the names of its
```
felixvd, line 18: If you both work at the same place, please replace Willow Garage with your employer. Otherwise, you can write something like "Neither the name of the authors nor the names of other".

Same for the other files.

felixvd (approved):
This is a great feature, and long overdue. Thanks a lot for the contribution. I only have some shallow comments, but it looks good to me overall.

Thieso:
Ok perfect, thanks for the help.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/cartesian_speed.h
rhaschke: The call to `updateTrajectory()` occurs in the `.cpp` file, not in the header file. Thus, the include is only needed in the `.cpp`.

moveit_ros/planning/planning_request_adapter_plugins/src/set_max_cartesian_link_speed.cpp
```diff
@@ -0,0 +1,85 @@
...
+/* Authors: Benjamin Scholz, Thies Oelerich, based off add_time_parameterization.cpp by Ioan Sucan */
+
+#include <moveit/planning_request_adapter/planning_request_adapter.h>
+#include <moveit/trajectory_processing/cartesian_speed.h>
+#include <class_loader/class_loader.hpp>
+#include <ros/console.h>
+
+namespace default_planner_request_adapters
+{
+class setMaxCartesianLinkSpeed : public planning_request_adapter::PlanningRequestAdapter
```
v4hn, line 46: I agree [with @gautz's comment](https://github.com/ros-planning/moveit_msgs/pull/122/files#r661510252)
The name must be capitalized.

moveit_ros/planning/planning_request_adapters_plugin_description.xml
```diff
@@ -47,4 +47,10 @@
     </description>
   </class>
 
+  <class name="default_planner_request_adapters/setMaxCartesianLinkSpeed" type="default_planner_request_adapters::setMaxCartesianLinkSpeed" base_class_type="planning_request_adapter::PlanningRequestAdapter">
```
v4hn, line 50: here accordingly in the name *and* the type

moveit_core/trajectory_processing/src/cartesian_speed.cpp
v4hn: I guess you resolved this because you [added the slowest speed into the ROS_WARN](https://github.com/ros-planning/moveit/pull/2674/files#diff-977a18c3ac8d4467d1fb062651186b55476f7ecb589a3ec2f3c665c6f6c9509dR128), but that's not what I meant.
The method is called `*limitMax*CartesianLinkSpeed` and by name uses would not expect a warning here at all!
The only solutions I see here are those I outlined above:

(1) Either remove the whole warning and accept it as normal method semantics, or 
(2) rename the method to `enforceCartesianLinkSpeed` or similar and make it a critical error in this method

I would prefer (1) by far, also because acceleration periods (esp. for high-density trajectories) **always** limit the Cartesian speed if the start/end waypoints have 0 velocities.

v4hn:
I rebased to get CI to run properly

v4hn:
CI failed

Thieso:
Right, sorry, forgot to change the function name in the test case. CI should work now.

Thieso:
Ok now the test case fails which is surprising since it works for me. Can you give me a hint why this might happen?

v4hn:
Briefly looking at this, I wonder why you would ask for an epsilon of 1e-10 here.
The offending test passes with an error of 0.13 mm/s.
Is a tolerance of 0.2mm/s (only below the specified target) acceptable for this test?

I'm unsure though why the test passes for you locally. I restarted the CI job to test repeatability.

Thieso:
Yes that seems acceptable to me since it test average speed which deviates due to acceleration to the maximum speed.

moveit_ros/planning/planning_request_adapter_plugins/src/set_max_cartesian_link_speed.cpp
```diff
@@ -0,0 +1,85 @@
...
+/* Authors: Benjamin Scholz, Thies Oelerich, based off add_time_parameterization.cpp by Ioan Sucan */
+
+#include <moveit/planning_request_adapter/planning_request_adapter.h>
+#include <moveit/trajectory_processing/cartesian_speed.h>
+#include <class_loader/class_loader.hpp>
+#include <ros/console.h>
+
+namespace default_planner_request_adapters
+{
+class SetMaxCartesianLinkSpeed : public planning_request_adapter::PlanningRequestAdapter
```
v4hn, line 46: [As proposed by @felixvd](https://github.com/ros-planning/moveit_resources/pull/88#discussion_r662130492), please change the name of the adapter to `LimitMaxCartesianLinkSpeed` to keep the names consistent.

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -269,6 +269,15 @@ class MoveGroupInterface
       If the value is greater than 1, it is set to 1.0. */
   void setMaxAccelerationScalingFactor(double max_acceleration_scaling_factor);
 
+  /** \brief Set the maximum cartesian speed for a given link.
+      The unit of the speed is meters per second and needs to be greater than 0.
+      The desired speed is a maximum bound. Slower parts of the trajectory will
+      be left unchanged. */
```
v4hn, line 275:
```suggestion
      be left unchanged. If no link_name is given, the endeffector tip will be assumed. */
```

moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
```diff
@@ -269,6 +269,15 @@ class MoveGroupInterface
       If the value is greater than 1, it is set to 1.0. */
   void setMaxAccelerationScalingFactor(double max_acceleration_scaling_factor);
 
+  /** \brief Set the maximum cartesian speed for a given link.
+      The unit of the speed is meters per second and needs to be greater than 0.
```
v4hn, line 273:
```suggestion
      The unit of the speed is meters per second and must be greater than 0.
```

v4hn (commented):
Looks like this will be the last iteration from my side. Looks good now! Thank you so much for investing time on this. :medal_sports:

Thieso:
Well now the test case fails miserably even though I did not change anything that should relate to the outcome of the test. So I assume there is something very wrong at some other point. Maybe the panda arm model that changes somehow? I am not sure how to debug this since locally everything works fine, just the CI fails. 
The command I use for testing ist `catkin run_tests --no-status --summarize moveit_core`.

v4hn:
> I am not sure how to debug this since locally everything works fine, just the CI fails.

Also in CI the test passed several times, but it seems to be flaky. We definitely can't merge this PR until the test passes again, otherwise we might just merge broken functionality. From the results it looks to me like `trajectory_processing::limitMaxCartesianLinkSpeed` fails to do anything at all for some reason.

moveit_core/trajectory_processing/src/cartesian_speed.cpp
```diff
@@ -0,0 +1,123 @@
...
+bool limitMaxCartesianLinkSpeed(robot_trajectory::RobotTrajectory& trajectory, const double max_speed,
+                                const moveit::core::LinkModel* link_model)
+{
+  if (max_speed <= 0.0)
+  {
+    ROS_ERROR_STREAM_NAMED(LOGGER_NAME, "Link speed must be greater than 0.");
+    return false;
+  }
+
+  std::string link_name = link_model->getName();
```
v4hn, line 85: you should not use this string here anymore. Instead, give the `LinkModel*` directly to the `getGlobalLinkTransform` method, there's an overload for it.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/iterative_time_parameterization.h
```diff
@@ -64,4 +64,6 @@ class IterativeParabolicTimeParameterization
   double findT1(const double d1, const double d2, double t1, const double t2, const double a_max) const;
   double findT2(const double d1, const double d2, const double t1, double t2, const double a_max) const;
 };
+
+void updateTrajectory(robot_trajectory::RobotTrajectory& rob_trajectory, const std::vector<double>& time_diff);
```
v4hn, line 68: could you please turn this method into a static method inside the `IterativeParabolicTimeParameterization` class above?
Sorry for these further incremental requests. I looked at the details again to maybe find the cause for the CI failures.

v4hn:
format failed.

Thieso:
I fixed the format, but probably that will not fix the test. Does anyone have any idea whats wrong?

Thieso:
Ok I have had a similar issue before I think. I added the EE name now to the function call. Can you please run CI again?

Thieso:
Hello, it would be great if we could finish this somehow. So could you please run CI again to see if it works now? Thank you

gautz:
I'd love to see this merged before the summer break too. I updated the corresponding PRs in MTC and moveit_resources too.

We should not forget to remove https://github.com/ros-planning/moveit/pull/2674/commits/d4c12e86b8eacd425b734d1ad69af3178e53c9d9 after https://github.com/ros-planning/moveit_msgs/pull/122 is merged

Thieso:
Ok seems to work now. The CI that fails does not seem to be related to my code. 
However, I am not sure why this does fix it. But I guess we can merge this for now??

Thieso:
Now CI fails just because of some compilation options which I do not think I can change right?

gautz:
In last meeting @henningkayser suggested further work on this PR ["This is useful, but we would need extra parametrization"](https://discourse.ros.org/t/moveit-manipulation-wg-meeting-july-22nd/21288/5) 

Knowing that @Thieso will probably not have time to continue working on this, I would advocate for merging the already useful functionality now and opening a new PR/issue to improve the parameterization.

With #2792 merged, can we re-run the CI and maybe merge this PR? @v4hn Optimal would be to first merge https://github.com/ros-planning/moveit_msgs/pull/122

v4hn:
replaced by #2856

PR #2682, opened 2021-05-22
collision_detection_fcl: Report link_names in correct order

xqms:
### Description

For distance queries, FCL may report results in different order than what was requested (e.g. `fcl::distance(o1, o2, ...)` might give an `fcl::DistanceResult` with `o2, o1`. The MoveIt wrapper does not check if this is the case, and thus may report nearest_points in order `o2,o1` with `link_name` in order `o1,o2`.

This PR fixes this issue by explicitly using the collision objects referenced in the `fcl::DistanceResult` object. It also adds a unit test which demonstrates the issue.

### Checklist
    I feel this is not applicable, since this is a pure bugfix.
    I think the risk that somebody depends on the current broken behavior is quite low.

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
@@ -304,7 +304,66 @@ TYPED_TEST_P(DistanceCheckPandaTest, DistanceSingle)
   }
 }
 
+// FCL < 0.6 completely fails the DistancePoints test, so we have two test
+// suites, one with and one without the test.
+template <class CollisionAllocatorType>
+class DistanceFullPandaTest : public DistanceCheckPandaTest<CollisionAllocatorType>
+{
+};
+
+TYPED_TEST_CASE_P(DistanceFullPandaTest);
+
+/** \brief Tests if nearest points are computed correctly. */
+TYPED_TEST_P(DistanceFullPandaTest, DistancePoints)
+{
+  // Adding the box right in front of the robot hand
+  shapes::Box* shape = new shapes::Box(0.1, 0.1, 0.1);
+  shapes::ShapeConstPtr shape_ptr(shape);
+
+  Eigen::Isometry3d pos{ Eigen::Isometry3d::Identity() };
+  pos.translation().x() = 0.43;
+  pos.translation().y() = 0;
+  pos.translation().z() = 0.55;
+  this->cenv_->getWorld()->addToObject("box", shape_ptr, pos);
+
+  collision_detection::DistanceRequest req;
+  req.acm = &*this->acm_;
+  req.type = collision_detection::DistanceRequestTypes::SINGLE;
+  req.enable_nearest_points = true;
+  req.max_contacts_per_body = 1;
+
+  collision_detection::DistanceResult res;
+  this->cenv_->distanceRobot(req, res, *this->robot_state_);
+
+  // Checks a particular point is inside the box
+  auto checkInBox = [&](const Eigen::Vector3d& p) {
+    Eigen::Vector3d inBox = pos.inverse() * p;
+
+    constexpr double EPS = 1e-5;
+    EXPECT_LE(std::abs(inBox.x()), shape->size[0] + EPS);
+    EXPECT_LE(std::abs(inBox.y()), shape->size[1] + EPS);
+    EXPECT_LE(std::abs(inBox.z()), shape->size[2] + EPS);
+  };
+
+  // Check that all points reported on "box" are actually on the box and not
+  // on the robot
+  for (auto& distance : res.distances)
+  {
+    for (auto& pair : distance.second)
+    {
+      if (pair.link_names[0] == "box")
+        checkInBox(pair.nearest_points[0]);
+      else if (pair.link_names[1] == "box")
+        checkInBox(pair.nearest_points[1]);
+      else
+        ADD_FAILURE() << "Unrecognized link names";
+    }
+  }
+}
+
 REGISTER_TYPED_TEST_CASE_P(CollisionDetectorPandaTest, InitOK, DefaultNotInCollision, LinksInCollision,
                            RobotWorldCollision_1, RobotWorldCollision_2, PaddingTest, DistanceSelf, DistanceWorld);
 
 REGISTER_TYPED_TEST_CASE_P(DistanceCheckPandaTest, DistanceSingle);
+
+REGISTER_TYPED_TEST_CASE_P(DistanceFullPandaTest, DistancePoints);
--- a/moveit_core/collision_detection_fcl/src/collision_common.cpp
+++ b/moveit_core/collision_detection_fcl/src/collision_common.cpp
@@ -540,6 +540,12 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
   {
     DistanceResultsData dist_result;
     dist_result.distance = fcl_result.min_distance;
+
+    // Careful here: Get the collision geometry data again, since FCL might
+    // swap o1 and o2 in the result.
+    const CollisionGeometryData* res_cd1 = static_cast<const CollisionGeometryData*>(fcl_result.o1->getUserData());
+    const CollisionGeometryData* res_cd2 = static_cast<const CollisionGeometryData*>(fcl_result.o2->getUserData());
+
 #if (MOVEIT_FCL_VERSION >= FCL_VERSION_CHECK(0, 6, 0))
     dist_result.nearest_points[0] = fcl_result.nearest_points[0];
     dist_result.nearest_points[1] = fcl_result.nearest_points[1];
@@ -547,10 +553,10 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
     dist_result.nearest_points[0] = Eigen::Vector3d(fcl_result.nearest_points[0].data.vs);
     dist_result.nearest_points[1] = Eigen::Vector3d(fcl_result.nearest_points[1].data.vs);
 #endif
-    dist_result.link_names[0] = cd1->getID();
-    dist_result.link_names[1] = cd2->getID();
-    dist_result.body_types[0] = cd1->type;
-    dist_result.body_types[1] = cd2->type;
+    dist_result.link_names[0] = res_cd1->getID();
+    dist_result.link_names[1] = res_cd2->getID();
+    dist_result.body_types[0] = res_cd1->type;
+    dist_result.body_types[1] = res_cd2->type;
     if (cdata->req->enable_nearest_points)
     {
       dist_result.normal = (dist_result.nearest_points[1] - dist_result.nearest_points[0]).normalized();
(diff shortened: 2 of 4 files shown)


moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
```diff
@@ -236,6 +236,43 @@ TYPED_TEST_P(CollisionDetectorPandaTest, DistanceSelf)
...
+
+  // Check that all points reported on "box" are actually on the box
+  for (auto& distance : res.distances)
+  {
+    for (auto& pair : distance.second)
+    {
+      if (pair.link_names[0] == "box")
+        EXPECT_NEAR(pair.nearest_points[0].x(), 0.38, 0.001);
+      else if (pair.link_names[1] == "box")
+        EXPECT_NEAR(pair.nearest_points[1].x(), 0.38, 0.001);
```
v4hn, line 269: Could you please check *all* dimensions and possibly write it in terms of `shape`/`pos` above?
It's bad practice to spell out numbers when they are derived from others and this test might easily break if geometry/locations change in the future.

v4hn (commented):
Thank you for the contribution and greetings to Bonn!

It's unexpected to see that FCL can change the order here and I wonder whether they introduced this issue in some later release.
I'm happy to consider the case in MoveIt, but I would also expect a bug report at the FCL repository. Did you create one there as well?

It would be great if you could improve the test case according to my comment below. Thank you lots for creating a test case to begin with. :+1:

xqms:
@v4hn thanks for the quick feedback and greetings back to Hamburg ;)

I'll improve the test case tomorrow or Monday and will look at the CI failure as well. I just wanted to write down what I found out about FCL itself:

Here is the code that applies the distance algorithm "in reverse" depending on the object type:
https://github.com/flexible-collision-library/fcl/blob/809dc40cad5a9d925214e1d1089f0e48193de413/include/fcl/narrowphase/distance-inl.h#L117-L138

and here is the same piece of code 9 years ago (it got moved around a bit, but is still very much the same):
https://github.com/flexible-collision-library/fcl/blob/0572525a7206915870ccff2e7b5d550eb7867ab0/src/distance.cpp#L81-L102

So I think it's safe to say that this has been the case for quite some time and is, I guess, intentionally designed that way. I'll open an issue with FCL about documenting this more explicitly. Changing the behavior on FCL's side is probably not a good idea.

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
```diff
@@ -236,6 +236,43 @@ TYPED_TEST_P(CollisionDetectorPandaTest, DistanceSelf)
...
+  collision_detection::DistanceRequest req;
+  req.acm = &*this->acm_;
+  req.type = collision_detection::DistanceRequestTypes::SINGLE;
+  req.enable_nearest_points = true;
+  req.max_contacts_per_body = 1;
+
+  collision_detection::DistanceResult res;
+  this->cenv_->distanceRobot(req, res, *this->robot_state_);
+
+  // Check that all points reported on "box" are actually on the box
```
AndyZe, line 261: If I understand correctly, the other option is that the reported points are on the hand? If that's right, please add comment:

`// The other option is that reported points are on the hand`

moveit_core/collision_detection_fcl/src/collision_common.cpp
```diff
@@ -586,12 +592,17 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
 #if (MOVEIT_FCL_VERSION >= FCL_VERSION_CHECK(0, 6, 0))
         dist_result.nearest_points[0] = contact.pos;
         dist_result.nearest_points[1] = contact.pos;
-        dist_result.normal = contact.normal;
+        Eigen::Vector3d normal = contact.normal;
 #else
         dist_result.nearest_points[0] = Eigen::Vector3d(contact.pos.data.vs);
         dist_result.nearest_points[1] = Eigen::Vector3d(contact.pos.data.vs);
-        dist_result.normal = Eigen::Vector3d(contact.normal.data.vs);
+        Eigen::Vector3d normal(contact.normal.data.vs);
```
AndyZe, line 599: It's unfortunate that there will be a small hit to performance here. We have an application where fast distance checks are important.

I guess it is worthwhile, though. Any other thoughts on the impact on performance?

AndyZe:
I'll give this a test today. Seems like a good idea, overall.

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
xqms: Done, thanks :)

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_panda.h
xqms: I added `... and not on the robot` which should make it more clear.

moveit_core/collision_detection_fcl/src/collision_common.cpp
xqms: This branch of code is only executed when `enable_signed_distance` is true, so the performance impact will only be visible for these cases.

I think we should definitely fix it, since the current code will report a normal in the wrong direction.

v4hn: I would not expect a performance drop at all (assuming compiler optimizations are enabled).
The allocated variable is only a helper because of the `#ifdef` and should be optimized away.

v4hn (approved):
Looks good to me now (unless CI objects).

Thank you for taking the time to clean up the test!
@AndyZe Please merge if your test succeeds.

moveit_core/collision_detection_fcl/src/collision_common.cpp
AndyZe: I'm always skeptical when somebody says, "just trust compiler optimizations."

Attached is some data from a BioIK control loop that does probably thousands of distance checks per second. On main branch, the longest duration is < `0.004s`. With this PR, the longest control period is` 0.0112s`.

I would like to merge the PR too but I don't like this.

[signed_distance_pr.txt](https://github.com/ros-planning/moveit/files/6539790/signed_distance_pr.txt)
[main.txt](https://github.com/ros-planning/moveit/files/6539791/main.txt)

xqms: Hm, that's a pretty significant increase... I'd guess the issue is not this particular line (it seems rather innocent and optimizable to me as well), but either the branch below (which inverts the normal) or the `res_cd1`/`res_cd2` stuff above, which definitely causes a memory read. But still, I'm quite surprised. The `distanceCallback` is called once per collision object pair, correct? My feeling was that anything we do here should not have a huge impact compared to the stuff FCL has to do internally on each `fcl::distance()` call.

Are you using just primitives or meshes as well?

Do you require the `nearest_points` / `normal` results in your application? Right now `enable_signed_distance` implies `nearest_points` and `normal` calculation, but I assume you don't actually use the normal (since it's wrong half of the time). Maybe an acceptable solution for your use case would be to add an additional `if (cdata->req->enable_nearest_points)`  around normal computation even in the signed distance case?

I could try to set up a loop that repeatedly evaluates distances and to get some insight using a profiler.

AndyZe: We're using meshes. Yeah, we don't care about the normal, just care about the distance.

I like this idea -- 

` Maybe an acceptable solution for your use case would be to add an additional if (cdata->req->enable_nearest_points) around normal computation even in the signed distance case?`

If you want to push that, I can test it easily.

xqms:
CI is failing since FCL < 0.6 (which is used on melodic) reports the nearest_points sometimes in the wrong frame (body coordinates):

https://github.com/flexible-collision-library/fcl/issues/171, https://github.com/flexible-collision-library/fcl/pull/288

In that case, the `nearest_points` and `normal` fields are completely wrong and there is no easy way to fix them on our side. Should we disable the test for FCL < 0.6? It's not easy to do, since the tests are templated on the collision solver type :thinking:

xqms:
> Should we disable the test for FCL < 0.6? It's not easy to do, since the tests are templated on the collision solver type

I went ahead and did that, so CI should pass now.

moveit_core/collision_detection_fcl/src/collision_common.cpp
v4hn: > I'm always skeptical when somebody says, "just trust compiler optimizations."

You better be! That's what benchmarks are for and it's why I used "expect".
Though you are the one who said he's not interested in GSoC projects for better Benchmarking. :-)
Thanks for updating the name by the way :medal_sports: 

I'm quite curious where this delay comes from, for educational reasons. But I also think the additional delay should not be that severe.

On optimizing the runtime: I might not be deep enough in this code at the moment, but wouldn't you just need one additional comparison `o1 == fcl_result.o1` to distinguish between the original case and the failure condition you found (inverted objects). We can't get around the negation and the pointer comparison, but that should be everything in the ideal case.

That being said, additionally guarding the lines that fill in nearest points/normal when using `enable_signed_distance` with `enable_nearest_points` makes some sense too, but it breaks method semantics in a bad way. :crying_cat_face: 
In case that's actually required to keep the rough current performance please update the logic to compute the distance only once (it's already computed with worth accuracy above) and add an entry to the [`MIGRATION.md`](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md)

v4hn:
> Should we disable the test for FCL < 0.6?

Thanks for implementing it. I would propose to go one step further though!
If it's documented in github issues that this request produces inconsistent results with FCL < 0.6 we should simply *reject the actual request* from user code in this case and `ROS_ERROR` tell them to upgrade their FCL version if they want to use it.

xqms:
Good idea! I implemented it as a throttled `ROS_ERROR` that allows the program to continue. That way, older programs that have `enable_nearest_points=true` or somehow are not affected by the points being switched don't break immediately.

moveit_core/collision_detection_fcl/src/collision_common.cpp
AndyZe: It looks like the average control loop period is within 10% now so I'd be happy merging this as-is.  :+1:

Master:  mean period = 0.0045s, worst case = 0.014s
This PR: mean period = 0.0052s, worst case = 0.013s

(Maybe I just screwed up the data collection before?)

AndyZe (approved):

moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp
```diff
@@ -49,6 +49,28 @@ namespace collision_detection
...
+{
+#if MOVEIT_FCL_VERSION < FCL_VERSION_CHECK(0, 6, 0)
+  if (req.enable_nearest_points)
+  {
+    // Known issues:
+    //   https://github.com/flexible-collision-library/fcl/issues/171,
+    //   https://github.com/flexible-collision-library/fcl/pull/288
+    ROS_ERROR_THROTTLE_NAMED(2.0, LOGNAME,
+                             "You requested a distance check with enable_nearest_points=true, "
+                             "but the FCL version MoveIt! was compiled against (%d.%d.%d) "
```
v4hn, line 65:
```suggestion
                             "but the FCL version MoveIt was compiled against (%d.%d.%d) "
```

v4hn:
Thanks @xqms . Please send more patches if you encounter other issues :1st_place_medal:

xqms:
Also thanks to you both @v4hn and @AndyZe for the fast review & feedback!

@v4hn Just wanted to point out that we *did* change the behavior for `enable_signed_distance=true` and `enable_nearest_points=false` now. Do you think we need to add something to `MIGRATION.md`?

To follow up on the performance discussion: I couldn't measure a difference between master (before the merge) and this PR. Here's the benchmark I used: https://gist.github.com/xqms/0500ed88ac9836f6bbe0022298635c8e
I consistently get 10ms per `distanceSelf()` call, nearly all of which is spent inside FCL according to the profiler. @AndyZe maybe your use case is different?

PR #2683, opened 2021-05-22
Revert "Lock the octomap/octree while collision checking (#2596)"

### LINKED ISSUE #2666: motion planning stuck after several calls.
### Description

Overview of your issue here.

### Your environment
* ROS Distro: [Melodic]
* OS Version: Ubuntu 18.04
* Source or Binary build?
* If source, master

### Steps to reproduce
```
self.hand_commander = MoveGroupCommander("hand")
self.hand_commander.set_start_state_to_current_state()
self.hand_commander.set_joint_value_target(hand_joint_positions)
self.hand_commander.go(wait=True)
```

### Expected behaviour
Go to the target joint position

### Actual behaviour
Code stuck at `self.hand_commander.go(wait=True)`
No error msg

~~My guess is this behaviour is related to https://github.com/ros-planning/moveit/pull/2663 which introduce a feature that `The joint states of `passive` joints must be published in ROS and the CurrentStateMonitor will now wait for them as well. ` If so, please provide a possible fix for a robot such as shadow hand.~~

v4hn:
This reverts commit 0935d681b0141108e9695db30107184996963be7.

It introduced a deadlock in the standard MoveGroup planning pipeline.
Octomap is already locked in `PlanningSceneMonitor::lockSceneRead()` [before planning in the capability](https://github.com/ros-planning/moveit/blob/master/moveit_ros/move_group/src/default_capabilities/move_action_capability.cpp#L222).
So the recursive locking calls (even in a different thread because OMPL spawns [a separate goal sampling thread](https://github.com/ros-planning/moveit/blob/master/moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp#L90)) can result in a deadlock if a writer lock is requested in the meanwhile.

@simonschmeisser To solve the issue you describe in  #2596, you will have to lock the `PlanningSceneMonitor` that maintains the `octomap_monitor_`. The only potential alternative I see is to deep-copy the octomap in `decoupleParent()` if you use that method.
According to its documentation, this should actually always happen, but it's not implemented there.
If you rely on collision-checking of a diff scene without locking the parent scene, you are definitely in undefined-behavior land.

Fixes #2666 .

Sorry for the inconvenience and the delay in addressing this urgent issue.
@rhaschke I believe we need another fixup release after this commit. Sorry for the trouble...

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -38,7 +38,6 @@
 #include <moveit/planning_scene/planning_scene.h>
 #include <moveit/collision_detection_fcl/collision_detector_allocator_fcl.h>
 #include <geometric_shapes/shape_operations.h>
-#include <moveit/collision_detection/occupancy_map.h>
 #include <moveit/collision_detection/collision_tools.h>
 #include <moveit/trajectory_processing/trajectory_tools.h>
 #include <moveit/robot_state/conversions.h>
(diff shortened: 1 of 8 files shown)


rhaschke (approved):
I will prepare a new release immediately.

simonschmeisser:
Sorry for the trouble, I didn't know that this implementation (or generally all variants that prioritize writing over reading) is not save for recursive use. Or maybe I overlooked that it already locked in `PlanningSceneMonitor::lockSceneRead()`. It could of course be removed there in favor of a more encapsulated locking in `PlanningScene` but that would require some more checking and discussions

I will evaluate how long it takes to actually duplicate the Octomap in `decoupleParent()`

PR #2684, opened 2021-05-23
refactors for OccMapTree in PlanningScene

v4hn:
This includes the useful components from #2596 .

The `PlanningScene` should create `OCTOMAP_NS` as the derived `OccMapTree` through all interfaces.
To get there we still have to move the header to `moveit_core/collision_detection`, so that part remains as well.

The third commit was pure cleanup.

@simonschmeisser please review, I did not even test-compile over here.

**EDIT**: The original PR description stated:
> The static cast is assumed to succeed in multiple places.

This is actually not true without the associated (invalid and reverted)  patches by @simonschmeisser as our code base performs all checks on whether an `occupancy_map_monitor` exists. Still, it is a good idea to ensure this is true, especially for code that creates diffs over monitored planning scenes and thus can't easily check whether an occupancy_map_monitor actually exists.

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -36,6 +36,7 @@
 
 #include <boost/algorithm/string.hpp>
 #include <moveit/planning_scene/planning_scene.h>
+#include <moveit/collision_detection/occupancy_map.h>
 #include <moveit/collision_detection_fcl/collision_detector_allocator_fcl.h>
 #include <geometric_shapes/shape_operations.h>
 #include <moveit/collision_detection/collision_tools.h>
@@ -1262,6 +1263,26 @@ bool PlanningScene::usePlanningSceneMsg(const moveit_msgs::PlanningScene& scene_
     return setPlanningSceneMsg(scene_msg);
 }
 
+collision_detection::OccMapTreePtr createOctomap(const octomap_msgs::Octomap& map)
+{
+  std::shared_ptr<collision_detection::OccMapTree> om =
+      std::make_shared<collision_detection::OccMapTree>(map.resolution);
+  if (map.binary)
+  {
+    octomap_msgs::readTree(om.get(), map);
+  }
+  else
+  {
+    std::stringstream datastream;
+    if (map.data.size() > 0)
+    {
+      datastream.write((const char*)&map.data[0], map.data.size());
+      om->readData(datastream);
+    }
+  }
+  return om;
+}
+
 void PlanningScene::processOctomapMsg(const octomap_msgs::Octomap& map)
 {
   // each octomap replaces any previous one
@@ -1276,7 +1297,7 @@ void PlanningScene::processOctomapMsg(const octomap_msgs::Octomap& map)
     return;
   }
 
-  std::shared_ptr<octomap::OcTree> om(static_cast<octomap::OcTree*>(octomap_msgs::msgToMap(map)));
+  std::shared_ptr<collision_detection::OccMapTree> om = createOctomap(map);
   if (!map.header.frame_id.empty())
   {
     const Eigen::Isometry3d& t = getFrameTransform(map.header.frame_id);
@@ -1314,7 +1335,8 @@ void PlanningScene::processOctomapMsg(const octomap_msgs::OctomapWithPose& map)
     return;
   }
 
-  std::shared_ptr<octomap::OcTree> om(static_cast<octomap::OcTree*>(octomap_msgs::msgToMap(map.octomap)));
+  std::shared_ptr<collision_detection::OccMapTree> om = createOctomap(map.octomap);
+
   const Eigen::Isometry3d& t = getFrameTransform(map.header.frame_id);
   Eigen::Isometry3d p;
   tf2::fromMsg(map.origin, p);
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -479,14 +479,7 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest& 
                                    collision_detection::CollisionResult& res,
                                    const moveit::core::RobotState& robot_state) const
 {
-  // check collision with the world using the padded version
-  getCollisionEnv()->checkRobotCollision(req, res, robot_state, getAllowedCollisionMatrix());
-
-  if (!res.collision || (req.contacts && res.contacts.size() < req.max_contacts))
-  {
-    // do self-collision checking with the unpadded version of the robot
-    getCollisionEnvUnpadded()->checkSelfCollision(req, res, robot_state, getAllowedCollisionMatrix());
-  }
+  checkCollision(req, res, robot_state, getAllowedCollisionMatrix());
 }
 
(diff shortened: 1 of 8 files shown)


rhaschke:
@v4hn, could you please rebase this onto master. I cancelled the started release (and force-pushed master to 95f9e243def0b2cc83b2bc48664c0a62d670dbc7), because the prerelease test failed unexpectedly: https://github.com/ros-planning/moveit/runs/2649204700?check_suite_focus=true.

Could you please have a look why this error happens:
```
2021-05-23T08:52:06.7875734Z Scanning dependencies of target moveit_ros_occupancy_map_monitor
2021-05-23T08:52:06.8084670Z [ 20%] Building CXX object CMakeFiles/moveit_ros_occupancy_map_monitor.dir/src/occupancy_map_monitor.cpp.o
2021-05-23T08:52:08.6880010Z /tmp/ws/src/moveit/moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp:40:10: fatal error: moveit/collision_detection/occupancy_map.h: No such file or directory
2021-05-23T08:52:08.6886937Z    40 | #include <moveit/collision_detection/occupancy_map.h>
2021-05-23T08:52:08.6903078Z       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

rhaschke:
I am looking into the new ikfast test issue...

v4hn:
> Could you please have a look why this error happens:

I believe it's an issue with our overlayed CI setup that this was not caught by CI. So many entwined issues at the moment. -.- 
This was basically my incompetence because 5baa760127277c8e8cb11704b3a92a6ee08345a4 was a fixup and had to be reverted as well. I pushed it directly. If anything else goes wrong today I will not be available anymore, sorry. Too many personal obligations. I will check in about 9 hours again though. :-/

rhaschke:
Thanks. I triggered a pre-release test again.

simonschmeisser (approved):
Looks good to me

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1262,6 +1263,26 @@ bool PlanningScene::usePlanningSceneMsg(const moveit_msgs::PlanningScene& scene_
...
+    octomap_msgs::readTree(om.get(), map);
+  }
+  else
+  {
+    std::stringstream datastream;
+    if (map.data.size() > 0)
+    {
+      datastream.write((const char*)&map.data[0], map.data.size());
+      om->readData(datastream);
+    }
```
tylerjw, line 1281: @JStech I don't know enough about octomaps to know what this is doing.  Do you have any feedback for this?

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -479,14 +480,7 @@ void PlanningScene::checkCollision(const collision_detection::CollisionRequest&
...
 {
-  // check collision with the world using the padded version
-  getCollisionEnv()->checkRobotCollision(req, res, robot_state, getAllowedCollisionMatrix());
-
-  if (!res.collision || (req.contacts && res.contacts.size() < req.max_contacts))
-  {
-    // do self-collision checking with the unpadded version of the robot
-    getCollisionEnvUnpadded()->checkSelfCollision(req, res, robot_state, getAllowedCollisionMatrix());
-  }
+  checkCollision(req, res, robot_state, getAllowedCollisionMatrix());
```
AndyZe, line 483: Nice to see this consolidation! I have to say, I don't understand most of the rest of this PR  (why the namespaces changed and so on).

moveit_core/planning_scene/src/planning_scene.cpp
v4hn: @simonschmeisser copied the logic from [`octomap_msgs`](https://github.com/OctoMap/octomap_msgs/blob/melodic-devel/include/octomap_msgs/conversions.h#L110). We can't use the upstream methods because they [instantiate a `octomap::OcTree` directly](https://github.com/OctoMap/octomap_msgs/blob/melodic-devel/include/octomap_msgs/conversions.h#L95) and we want the lockable wrapper `OccMapTree` instead.

JStech: Yeah, I'm no expert on Octomap internals, but this looks like it's just copying the serialized tree from the message into the `om` object.

v4hn:
It's somewhat frustrating to have three maintainers look at that with minor responses but without a review.

moveit_core/planning_scene/src/planning_scene.cpp
jliukkonen: It is possible to exit out of this function without having any useful data in the OccMapTree. The octomap message could have empty data array. Would it make sense to have a warning here if the data is empty and we end up adding basically nothing to the collision scene?

jliukkonen (approved):

moveit_core/planning_scene/src/planning_scene.cpp
v4hn: Good point. Both methods that use this helper have a guard for `!msg.data.empty()`, so I believe the case can't happen at the moment. Pondering this for a while I would prefer to allow the use-case though. `octomap_msgs::readTree` explicitly allows for this and this method is basically just a wrapper for it.

simonschmeisser: I cannot follow, why shouldn't an octomap potentially be empty?

v4hn:
Because in the PlanningScene it should be removed instead in these cases. That's actually happening for all full octomap updates that contain no data.

Of course it's not enforced by the octomap updater if the map is incrementally cleared, but that case is not relevant here either way.

moveit_ros/occupancy_map_monitor/src/occupancy_map_monitor.cpp
```diff
@@ -99,7 +99,7 @@ void OccupancyMapMonitor::initialize()
                                                      << "\" specified but no TF instance (buffer) specified. "
                                                         "No transforms will be applied to received data.");
 
-  tree_.reset(new OccMapTree(map_resolution_));
+  tree_.reset(new collision_detection::OccMapTree(map_resolution_));
```
tylerjw, line 102: please use make_shared instead of reset(new

```suggestion
  tree_ = std::make_shared<collision_detection::OccMapTree>(map_resolution_);
```

moveit_core/planning_scene/src/planning_scene.cpp
davetcoleman: I think this helper function is a great candidate for adding a unit test for.

davetcoleman (commented):
Great cleanup and refactor!!

tylerjw (approved):
I finally tested this locally.  I'm sorry it took me this long to give this a review.  The comment I left about make_shared can be changed in a separate PR as you weren't really changing the behavior of that line.

moveit_core/planning_scene/src/planning_scene.cpp
tylerjw: Please address this in a follow-on PR if you get a chance.  I merged this PR as it's a useful fix and has been sitting long enough.

v4hn:
Thanks, I just didn't find the time to get back to this yet.

PR #2687, opened 2021-05-24
Fix compiler warnings and fail on warnings in future

rhaschke:


.github/workflows/ci.yaml
```diff
@@ -25,6 +25,7 @@ jobs:
           - IMAGE: melodic-ci
             CATKIN_LINT: true
     env:
+      CXXFLAGS: "-Werror -Wall -Wextra -Wwrite-strings -Wunreachable-code -Wpointer-arith -Wredundant-decls -Wno-deprecated-copy"
```
v4hn, line 28: Wouldn't we want `-pedantic -std=c++14` as well?
Also I would have expected issues with upstream library headers...
Are all potentially-problematic ones included as `-isystem` now?

v4hn (approved):

.github/workflows/ci.yaml
rhaschke: We never used `-pedantic` in the past and as our target compilers are gcc + clang, I personally don't mind allowing non-standard compiler extensions.

v4hn:
> We never used `-pedantic` in the past

I believe so, but we also never compiled with `-Werror`.
I see `-pedantic` simply as an extension from "warning-free" to "warning-free according to specified standard" here.

> and as our target compilers are gcc + clang,

I agree, but some users provide fixes for MS Visual Studio as well (see upcoming maintainer meeting).
(Not that MSVS would be standard conform...)

> I personally don't mind allowing non-standard compiler extensions.

We [explicitly disable extension support in our compile instructions](https://github.com/ros-planning/moveit/blob/master/moveit_core/CMakeLists.txt#L7-L8)
and I see no reason not to actually enforce this on CI (unless huge chunks would start to fail).

rhaschke:
> We also never compiled with `-Werror`.

We didn't do this explicitly but implicitly: Travis CI had a WARNINGS_OK=false variable, which made the build fail on any warnings (not only compiler warnings).

> I see no reason not to actually enforce this on CI

We can give it a try, yes.

PR #2691, opened 2021-05-25
Make CI pedantic

v4hn:
We rely on C++14 and should enforce standard-compliance (not just warning-free builds).

Potential followup to #2687

Let's see if anything/what breaks.


v4hn:
Turns out the inconsistent semicolons where the only issue we had w.r.t. `-pedantic` builds.

I'd propose to merge this now.

rhaschke (approved):
Looks good to me. Thanks for the cleanup!

PR #2692, opened 2021-05-26
Misspelled MoveIt

v4hn:
Absurdly this check exists for some of the auxiliary repositories,
but was not present in this the main repository yet.

Slightly modified because we use a number of `MoveitCommander` style
class names that should not be touched.


v4hn:
The additional pre-commit check succeeds. Merging.

PR #2695, opened 2021-05-27
Fix setup assistant template (execution_type, moveit_resources_prbt_moveit_config)

mintar:
* Fix package path for Pilz pipeline: Use the launch file from the generated package, not from moveit_resources_prbt_moveit_config

* Add missing execution_type arg. This is necessary to avoid the following error:

        RLException: unused args [execution_type] for include of [ [...]moveit_controller_manager.launch.xml]

### DIFF
--- a/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/moveit_controller_manager.launch.xml
+++ b/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/moveit_controller_manager.launch.xml
@@ -1,7 +1,7 @@
 <launch>
 
   <!-- this argument is not used here, only necessary so that this launch file has the same args as fake_moveit_controller_manager.launch -->
-  <arg name="execution_type" default="interpolate" />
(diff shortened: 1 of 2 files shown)


rhaschke (approved):

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/moveit_controller_manager.launch.xml
```diff
@@ -1,5 +1,8 @@
 <launch>
 
+  <!-- this argument is not used here, only necessary so that this launch file has the same args as fake_moveit_controller_manager.launch -->
+  <arg name="execution_type" default="interpolate" />
```
rhaschke, line 4: What about making the fact that this is unused even more explicit:
```suggestion
  <arg name="execution_type" default="unused" />
```

mintar:
Thanks, I've added your suggested change.

Issue #2653, opened 2021-05-06
Failing Pilz unit test due to circular dependency


rhaschke:
https://github.com/ubi-agni/moveit/runs/2521124797
```
Full test results for 'build/pilz_industrial_motion_planner/test_results/pilz_industrial_motion_planner/rostest-test_integrationtest_get_solver_tip_frame.xml'
-------------------------------------------------
<?xml version="1.0" encoding="utf-8"?>
<testsuite errors="0" failures="1" name="unittest.suite.TestSuite" tests="1" time="1.418"><testcase classname="rostest.runner.RosTest" name="testintegrationtest_get_solver_tip_frame" time="1.4178"><failure type="AssertionError">test [integrationtest_get_solver_tip_frame] did not generate test results
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/opt/ros/melodic/lib/python2.7/dist-packages/rostest/runner.py", line 164, in fn
    self.assert_(os.path.isfile(test_file), "test [%s] did not generate test results"%test_name)
  File "/usr/lib/python2.7/unittest/case.py", line 422, in assertTrue
    raise self.failureException(msg)
</failure></testcase><system-out>&lt;![CDATA[
[ROSTEST]setup[/home/runner/work/moveit/moveit/.work/target_ws/src/moveit/moveit_planners/pilz_industrial_motion_planner/test/integrationtest_get_solver_tip_frame.test] run_id[49b9cdea-ae9e-11eb-bc73-0242ac110002] starting
&amp;#x001B;[1mstarted roslaunch server http://9cd72f10c291:39871/&amp;#x001B;[0m

SUMMARY
========

PARAMETERS
 * /robot_description: &lt;?xml version="1....
 * /robot_description_kinematics/manipulator/kinematics_solver: prbt_manipulator/...
 * /robot_description_kinematics/manipulator/kinematics_solver_search_resolution: 0.005
 * /robot_description_kinematics/manipulator/kinematics_solver_timeout: 0.005
 * /robot_description_kinematics/manipulator/link_prefix: prbt_
 * /robot_description_planning/cartesian_limits/max_rot_vel: 1.57
 * /robot_description_planning/cartesian_limits/max_trans_acc: 2.25
 * /robot_description_planning/cartesian_limits/max_trans_dec: -5
 * /robot_description_planning/cartesian_limits/max_trans_vel: 1
[...]

PR #2697, opened 2021-05-28
Migrate PRBT ikfast plugin from moveit_resources

v4hn:
As announced in https://github.com/ros-planning/moveit/issues/2653#issuecomment-850399443 .
Together with https://github.com/ros-planning/moveit_resources/pull/79 this is supposed to provide a straight-forward solution for #2653 .

Moving the plugin here circumnavigates problems with running tests in the moveit repository that are based on it. Before, this plugin was part of moveit_resources, but introduced a compile time dependency on moveit_core.
Because a repository's tests should run without having to compile downstream code (a sane requirement on buildfarm/prerelease tests), we move the plugin here instead.


moveit_kinematics/prbt_manipulator_ikfast_plugin/CHANGELOG.rst
```diff
@@ -0,0 +1,11 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package moveit_resources_prbt_ikfast_manipulator_plugin
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+0.8.0 (2021-04-28)
+------------------
+
+0.7.2 (2021-03-26)
```
rhaschke, line 8: The version info here doesn't match to versions of MoveIt anymore.

rhaschke (changes requested):
As said before, I do not support moving a _specific_ IKfast planner into the `moveit_kinematics` package, which should provide `generic` interfaces only.
Additionally, as pointed out in https://github.com/ros-planning/moveit_resources/pull/79#discussion_r642044458, moving the IK plugin into `moveit_kinematics`, will not resolve the issue that `moveit_resources` depends on `moveit_kinematics`.

v4hn:
> As said before, I do not support moving a specific IKfast planner into the moveit_kinematics package, which should provide generic interfaces only.

We could put it in a separate ROS package in the repository if that's your main concern.
The only important thing to get this to work is that the code is *somewhere* in this repository and it somewhat made sense to me to associate it with `moveit_kinematics`.

v4hn:
Closing as we decided to take [a different approach](https://github.com/ros-planning/moveit/issues/2653#issuecomment-856000201).
Still, someone has to actually do all the work for this...

PR #2700, opened 2021-05-31
Remove false octomap from catkin_packages LIBRARIES entries

0Nel:
### Description

Octomap is a dependency of moveit_core, but in the catkin_package description it is listed under LIBRARIES, which suggests it is a library that originates from moveit_core.

This PR is opened in order to fix this false entry, pointed out by @v4hn on another PR https://github.com/ros-planning/moveit/pull/2671

      Conceptually this line is probably wrong, because LIBRARIES should only list "the exported libraries from the project".
      In theory the entry OCTOMAP in DEPENDS below should do the same the right way.

      The corresponding line in moveit_core that you refer to is also 9 years old apparently and I guess nobody ever looked into it. 

Tested on a clean noetic build.

### Checklist

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -152,7 +152,6 @@ catkin_package(
     moveit_dynamics_solver
     moveit_utils
     moveit_test_utils
-    ${OCTOMAP_LIBRARIES}
   CATKIN_DEPENDS
     eigen_stl_containers
     geometric_shapes


tylerjw (approved):
Good catch!

AndyZe (approved):

v4hn:
Thanks.

0Nel:
Thanks for the quick review and merge!

Issue #2702, opened 2021-05-31
CHOMP doesn't read parameters from proper namespace


mintar:
### Description

Parameters read by `move_group` (`~` is `/move_group`):

```
# 1. Params that should be in a namespace, but aren't:
- ~/collision_clearence
- ~/collision_threshold
- ~/enable_failure_recovery
- ~/joint_update_limit
- ~/learning_rate
- ~/max_iterations
- ~/max_iterations_after_collision_free
- ~/max_recovery_attempts
- ~/obstacle_cost_weight
- ~/planning_time_limit
- ~/pseudo_inverse_ridge_factor
- ~/ridge_factor
- ~/smoothness_cost_acceleration
- ~/smoothness_cost_jerk
- ~/smoothness_cost_velocity
- ~/smoothness_cost_weight
- ~/trajectory_initialization_method
- ~/use_pseudo_inverse
- ~/use_stochastic_descent

# 2. Params that are in a namespace, but shouldn't (or MSA templates should be adjusted)
- ~/planning_pipelines/chomp/jiggle_fraction
- ~/planning_pipelines/ompl/jiggle_fraction

# 3. Params that are attempted to be read, but are not in the config file (that's fine, just uses defaults then)
- ~/planning_pipelines/chomp/capabilities
- ~/planning_pipelines/chomp/default_workspace_bounds
- ~/planning_pipelines/chomp/max_sampling_attempts
- ~/planning_pipelines/chomp/start_state_max_dt

# 4. Params where config file and code agree (yay!)
- ~/planning_pipelines/chomp/planning_plugin
- ~/planning_pipelines/chomp/request_adapters
- ~/planning_pipelines/chomp/start_state_max_bounds_error
```

Parameters from moveit_config (written by setup assistant):

```
# 1. 
- ~/planning_pipelines/chomp/collision_clearence
- ~/planning_pipelines/chomp/collision_threshold
- ~/planning_pipelines/chomp/enable_failure_recovery
- ~/planning_pipelines/chomp/joint_update_limit
- ~/planning_pipelines/chomp/learning_rate
- ~/planning_pipelines/chomp/max_iterations
- ~/planning_pipelines/chomp/max_iterations_after_collision_free
- ~/planning_pipelines/chomp/max_recovery_attempts
- ~/planning_pipelines/chomp/obstacle_cost_weight
- ~/planning_pipelines/chomp/planning_time_limit
- ~/planning_pipelines/chomp/pseudo_inverse_ridge_factor
[...]

PR #2703, opened 2021-06-01
MSA: Fix template (max_safe_path_cost)

mintar:
Also fix a typo in a debug message.

### DIFF
--- a/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
+++ b/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
@@ -88,7 +88,7 @@
     <env name="DISPLAY" value="$(optenv DISPLAY :0)" />
 
     <param name="allow_trajectory_execution" value="$(arg allow_trajectory_execution)"/>
-    <param name="max_safe_path_cost" value="$(arg max_safe_path_cost)"/>
+    <param name="sense_for_plan/max_safe_path_cost" value="$(arg max_safe_path_cost)"/>
     <param name="jiggle_fraction" value="$(arg jiggle_fraction)" />
     <param name="default_planning_pipeline" value="$(arg pipeline)" />
     <param name="capabilities" value="$(arg capabilities)" />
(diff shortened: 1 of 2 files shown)


moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
```diff
@@ -16,7 +16,7 @@
   <arg name="allow_trajectory_execution" default="true"/>
   <arg name="fake_execution" default="false"/>
   <arg name="execution_type" default="interpolate"/> <!-- set to 'last point' to skip intermediate trajectory in fake execution -->
-  <arg name="max_safe_path_cost" default="1"/>
+  <arg name="sense_for_plan/max_safe_path_cost" default="1"/>
```
rhaschke, line 19: Indeed, the dynamic reconfigure server for this parameter is placed within the namespace `sense_for_plan`:
https://github.com/ros-planning/moveit/blob/abb83f178075b4d5bb655df2dc9dd3e400e258ca/moveit_ros/planning/plan_execution/src/plan_with_sensing.cpp#L52-L53

rhaschke (approved):
Thanks for this fix. 
Please, provide an explanation / motivation for such a configuration change in future. This will simplify our review process.

rhaschke (changes requested):
Actually, you fixed the name in the wrong location. You need to adapt the `param` name here:
https://github.com/ros-planning/moveit/blob/781a82c19a0d9acc7b22a2bcbab0ec0f7e5e27c4/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch#L91

instead of the `arg` name.

mintar:
Ooops, good catch. I copy-pasted the stuff from my own moveit config to the template and messed up. It's fixed now.

rhaschke (approved):

v4hn:
Thanks. PlanWithSensing is definitely not something many people use with MoveIt these days. :cat2:

mintar:
> Thanks. PlanWithSensing is definitely not something many people use with MoveIt these days.

I'm not using it either, I just noticed this while debugging #2702.

Issue #1330, opened 2019-01-29
can't get padding to work!


skasewa:
### Description

I'm just trying to add padding to my robot model so that it reports collisions around some margin of the robot when doing a collision check, but I just cannot seem to figure out how to increase the padding on the robot model.

From what I can tell, no matter what I do, the padding remains 0.0

Following other comments here, I have added a padding.yaml and put in various numbers for the parameters (including very large numbers) but they have no effect whatsoever.

```
default_robot_padding: 1.0
default_robot_scale: 1.0
default_object_padding: 1.0
default_attached_padding: 1.0
```

I am not the first one to report this behaviour (I have seen other issues reporting similar things from 2014).

Could someone help me with:
- a working example where the padding value actually changes the behaviour of collision checking and planning
- a full working example with PlanningSceneMonitor instead of PlanningScene (given that apparently one isn't supposed to use PlanningScene, but there are no samples/tutorials provided that do not use PlanningScene)

I'm trying to figure out where the padding values eventually affect the robot model. It's quite tricky to trace through what goes on once a checkCollision call is made. It summons CollosionRobot through `active_collision_->getCollisionRobot();` in line 203 of `planning_scene.h ` but I cannot seem to figure out what active_collision_ is pointing to, to trace it further back up the chain. My sense is that the getCollisionRobot() does not pad the robot model, or whatever padding values it accesses are one of the several default 0.0 values.

I'm surprised this isn't better investigated or documented, but maybe I have just not been able to find suitable sources. Any directions addressing my concerns would be deeply appreciated; I have been struggling to get this to work for over a day now.

* ROS Distro: Kinetic
* OS Version:  Ubuntu 16.04
* Binary

### Steps to reproduce
* Load any robot model with some padding
* Add any collision object in the scene (not currently colliding with the robot)
* Run collisionCheck (It will report false)
* Repeat with increasing pad values
* Notice that no matter what the pad values, the collisionCheck is always false

### Expected behaviour
* If the object collides with the padded robot, collisionCheck should report it
* It might be that I am just implementing something wrong (e.g by using padding.yaml + rosparam in launch file), so I would be grateful if you could upload a sample which works.

### Actual behaviour
* Repeat with increasing pad values
* Notice that no matter what the pad values, the collisionCheck is always false

### Backtrace or Console output
Not provided (I may carve out a minimum example to put up later, but I have already spent too long on this)

simonschmeisser:
[...]

PR #2706, opened 2021-06-02
PSM: Don't read padding parameters from private namespace

### LINKED ISSUE #2699: PSM ros params wrongly needs namespace prefix
Quoting @mintar from [here](https://github.com/ros-planning/moveit/issues/1330#issuecomment-851349546):

> Just a further note for everyone else who stumbles upon this: I'm using a prefix (mobipick) for my robot. It used to work exactly how they described:
> 
> mobipick/move_group/robot_description_planning/default_attached_padding: 0.0
> 
> In the current ROS Noetic version of MoveIt (1.1.5), there's a second mobipick necessary to make it work:
> 
> mobipick/move_group/mobipick/robot_description_planning/default_attached_padding: 0.0

Looking into it a bit this is probably a bug introduced in Henning's efforts to have `move_group` use `MoveItCpp` for support of multiple planning pipelines.
Specifically, [this commit](https://github.com/ros-planning/moveit/pull/2127/commits/8df40df56ad58a6358834129b09b3a8b2c35a02c) replaces `move_group`'s own PlanningSceneMonitor with the one initialized by MoveItCpp.

Please have a look @henningkayser .

mintar:
Before this commit, the padding parameters were read from the private namespace of the move_group node (unlike all the other PSM parameters).

Fixes #2699.

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -1422,17 +1422,33 @@ void PlanningSceneMonitor::configureDefaultPadding()
     return;
   }
 
-  // Ensure no leading slash creates a bad param server address
-  const std::string robot_description =
+  // print depreciation warning if necessary
+  // TODO: remove this warning after 06/2022
+  const std::string old_robot_description =
       (robot_description_[0] == '/') ? robot_description_.substr(1) : robot_description_;
+  if (nh_.resolveName(old_robot_description) != nh_.resolveName(robot_description_))
+  {
+    if (nh_.hasParam(old_robot_description + "_planning/default_robot_padding") ||
+        nh_.hasParam(old_robot_description + "_planning/default_robot_scale") ||
+        nh_.hasParam(old_robot_description + "_planning/default_object_padding") ||
+        nh_.hasParam(old_robot_description + "_planning/default_attached_padding") ||
+        nh_.hasParam(old_robot_description + "_planning/default_robot_link_padding") ||
+        nh_.hasParam(old_robot_description + "_planning/default_robot_link_scale"))
+    {
+      ROS_WARN_NAMED(LOGNAME, "The path for the padding parameters has changed!");
+      ROS_WARN_NAMED(LOGNAME, "Old parameter path: '%s'", nh_.resolveName(old_robot_description + "_planning/").c_str());
+      ROS_WARN_NAMED(LOGNAME, "New parameter path: '%s'", nh_.resolveName(robot_description_ + "_planning/").c_str());
+      ROS_WARN_NAMED(LOGNAME, "Ignoring old parameters. Please update your moveit config!");
+    }
+  }
 
-  nh_.param(robot_description + "_planning/default_robot_padding", default_robot_padd_, 0.0);
-  nh_.param(robot_description + "_planning/default_robot_scale", default_robot_scale_, 1.0);
-  nh_.param(robot_description + "_planning/default_object_padding", default_object_padd_, 0.0);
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
`robot_description` is the same as `robot_description_` except having a leading slash removed.
I guess that in most cases, they are identical. Or do we store an _absolute_ param path in `robot_description_`?

mintar:
> do we store an absolute param path in robot_description_?

Yes. See #1330 .

mintar:
I just remembered that I need to fix a small thing tomorrow, don't merge yet.

mintar:
Ok, the thing I just changed is to make sure that the depreciation warning is only emitted when the `/robot_description` parameter is an absolute path. Otherwise, `old_robot_description` and `robot_description_` are identical, so it makes no sense printing that warning.

Back to your question: 

>  Or do we store an absolute param path in robot_description_?

This is actually more convoluted than I thought. Here goes:

On my robot, I'm starting `move_group` in the namespace `/mobipick`. I'm going to use this in the rest of my comment because it's the more general case, but it works the same without namespaces if you replace `/mobipick` with the empty string in the rest of my comment.

The PSM gets `robot_description_` from the `RobotModelLoader`, here:

https://github.com/ros-planning/moveit/blob/dba2095a0b4a4a2352c8991a779bafe11abeefc9/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L189

... which in turn gets it from the `RDFLoader`, here:

https://github.com/ros-planning/moveit/blob/dba2095a0b4a4a2352c8991a779bafe11abeefc9/moveit_ros/planning/robot_model_loader/include/moveit/robot_model_loader/robot_model_loader.h#L95

... which uses `searchParam("robot_description", ...)` to get the parameter:

https://github.com/ros-planning/moveit/blob/3361b2d1b6b2feabc2d3e93c75653f5a00e87fa4/moveit_ros/planning/rdf_loader/src/rdf_loader.cpp#L67

The parameter layout on the parameter server looks like this:

```
/mobipick/robot_description: <URDF>
/mobipick/move_group/...: some more parameters
```

Since `searchParam` cannot find `/mobipick/move_group/robot_description`, it searches upward and finds ` /mobipick/robot_description` (fully resolved, i.e., with a leading `"/"`).

So from now on, `robot_description_ == " /mobipick/robot_description"` in the PSM.

This is now used by the PSM to load some parameters:

https://github.com/ros-planning/moveit/blob/dba2095a0b4a4a2352c8991a779bafe11abeefc9/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L247

... and some more, except for the padding parameters, where it drops the leading `"/"`:

https://github.com/ros-planning/moveit/blob/dba2095a0b4a4a2352c8991a779bafe11abeefc9/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L1426-L1429

This results in the following parameters being read:

```
/mobipick/robot_description_planning/cartesian_limits
/mobipick/robot_description_planning/default_collision_operations
/mobipick/robot_description_planning/joint_limits
/mobipick/robot_description_planning/shape_transform_cache_lookup_wait_time
/mobipick/move_group/mobipick/robot_description_planning/default_attached_padding
/mobipick/move_group/mobipick/robot_description_planning/...
```

This PR changes this so the following parameters are read instead:

```
/mobipick/robot_description_planning/cartesian_limits
/mobipick/robot_description_planning/default_collision_operations
/mobipick/robot_description_planning/joint_limits
/mobipick/robot_description_planning/shape_transform_cache_lookup_wait_time
/mobipick/robot_description_planning/default_attached_padding
/mobipick/robot_description_planning/...
```

v4hn:
As Robert noticed after I merged #2703 you targeted `noetic-devel` but we merge things into `master` and `noetic-devel` is still rebased every now and then.
I retargeted your two related requests.

v4hn (approved):

PR #2718, opened 2021-06-16
Fix predefined poses benchmark example

captain-yoshi:
This fixes the `CombinePredefinedPosesBenchmark` a custom benchmark example derived from the BenchmarkExecutor class.
- Change the benchmark node type to the custom benchmark node. 
- The predefined poses named _ready_ and _extended_ exists only in the panda_arm group.

Will have to investigate further this error message probably triggered [here](https://github.com/ros-planning/moveit/blob/master/moveit_ros/benchmarks/src/simple_benchmarks/CombinePredefinedPosesBenchmark.cpp#L103):
``` shell
[ERROR] [1623807614.550173309]: Found empty JointState message
```
The benchmark works regardless of the error.

### DIFF
--- a/moveit_ros/benchmarks/examples/demo_panda_predefined_poses.yaml
+++ b/moveit_ros/benchmarks/examples/demo_panda_predefined_poses.yaml
@@ -16,7 +16,7 @@ benchmark_config:
         group: panda_arm      # Required
         timeout: 10.0
         output_directory: /tmp/moveit_benchmarks/
-        predefined_poses_group: panda_arm_hand # Group where the predefined poses are specified
+        predefined_poses_group: panda_arm # Group where the predefined poses are specified
         predefined_poses: # List of named targets
             - ready
             - extended
--- a/moveit_ros/benchmarks/examples/demo_panda_predefined_poses.launch
+++ b/moveit_ros/benchmarks/examples/demo_panda_predefined_poses.launch
@@ -26,7 +26,7 @@
   </group>
 
   <!-- Launch benchmark node -->
-  <node name="moveit_run_benchmark" pkg="moveit_ros_benchmarks" type="moveit_run_benchmark" output="screen" required="true">
+  <node name="moveit_run_benchmark" pkg="moveit_ros_benchmarks" type="moveit_combine_predefined_poses_benchmark" output="screen" required="true">
     <rosparam command="load" file="$(arg bench_opts)"/>
   </node>
 </launch>


v4hn:
> [ERROR] [1623807614.550173309]: Found empty JointState message

This error is a leftover artifact found in multiple places in MoveIt over time.
Could you please add a check for empty robot state before applying the message [as done here](https://github.com/ros-planning/moveit/blob/master/moveit_ros/move_group/src/default_capabilities/kinematics_service_capability.cpp#L78-L81)?

moveit_ros/benchmarks/examples/demo_panda_predefined_poses.yaml
```diff
@@ -16,7 +16,7 @@ benchmark_config:
         group: panda_arm      # Required
         timeout: 10.0
         output_directory: /tmp/moveit_benchmarks/
-        predefined_poses_group: panda_arm_hand # Group where the predefined poses are specified
+        predefined_poses_group: panda_arm # Group where the predefined poses are specified
```
henningkayser, line 19: odd, the `predefined_poses_group` was set to this on purpose, but I didn't find any related changes skimming the config packages.

henningkayser (approved):
Changes look reasonable to me. I agree that the empty joint_state message is an annoyance, but probably unrelated to the benchmark package per se. It's still worth looking into it.

moveit_ros/benchmarks/src/simple_benchmarks/CombinePredefinedPosesBenchmark.cpp
```diff
@@ -59,11 +59,6 @@ class CombinePredefinedPosesBenchmark : public BenchmarkExecutor
     // Load planning scene
     if (!psm_)
       psm_.reset(new planning_scene_monitor::PlanningSceneMonitor("robot_description"));
-    if (!psm_->newPlanningSceneMessage(scene_msg))
-    {
-      ROS_ERROR_NAMED(LOGNAME, "Failed to load planning scene");
-      return false;
-    }
```
captain-yoshi, line 66: I think the `newPlanningSceneMessage` tries to update the scene with an empty `scene_msg` thus the empty JointState error. This benchmark does not load a planing scene from ROS warehouse.

So the actual scene is loaded/created from the `psm_.reset` function.

Bottom line I think the error message was valid. It works as expected without spawning the JointState error.

captain-yoshi:
The error message triggers from [here](https://github.com/ros-planning/moveit/blob/master/moveit_ros/benchmarks/src/simple_benchmarks/CombinePredefinedPosesBenchmark.cpp#L62).

The empty joint_state message should be fixed in the `PlanningSceneMonitor`. I will remove the second commit. Passing an empty JointState to update the scene should not throw an error message for our use case.

v4hn:
It's a mystery to me right now why the previous values were set there...

Issue #2677, opened 2021-05-20
Strange MoveIt commander behavior with ROS Noetic with UR robots


dudasdavid:
### Description

I'm using ROS Noetic and moveit commander (move group python interface) with pose targets (using set_pose_target). In most of the time MoveIt is not able to plan any path (see below), and when it does it's a very strange solution. With my same application I had no issues with ROS Melodic, but the issue can be reproduced easily on ROS Noetic using the simulation or the real robot.

```console
[ INFO] [1621454575.663979800, 95.245000000]: Planning attempt 1 of at most 1
[ INFO] [1621454575.666822400, 95.248000000]: Planner configuration 'manipulator' will use planner 'geometric::RRTConnect'. Additional configuration parameters will be set when the planner is constructed.
[ INFO] [1621454575.667156900, 95.248000000]: manipulator/manipulator: Starting planning with 1 states already in datastructure
[ERROR] [1621454580.672333000, 99.602000000]: manipulator/manipulator: Unable to sample any valid states for goal tree
[ INFO] [1621454580.672491800, 99.602000000]: manipulator/manipulator: Created 1 states (1 start + 0 goal)
[ INFO] [1621454580.672859000, 99.602000000]: No solution found after 5.005736 seconds
[ INFO] [1621454580.683008900, 99.607000000]: Unable to solve the planning problem
[ INFO] [1621454580.685616400, 99.609000000]: Received event 'stop'
```
This issue was originally reported at [universal_robot](https://github.com/ros-industrial/universal_robot/issues/572) and it was suggested to report it here instead.

First I thought it's an issue with my own application above moveit commander, so I created a very simple vanila test environtment, but the issue still exists without any modification on the original packages! Here are the steps to reproduce:

1. git clone -b calibration_devel-staging https://github.com/fmauch/universal_robot
2. Change arm_controller to pos_joint_traj_controller in ur3e_moveit_planning_execution.launch
3. Launch the following launchfiles:
- roslaunch ur_gazebo ur3e_bringup.launch limited:=true
- roslaunch ur3e_moveit_config ur3e_moveit_planning_execution.launch sim:=true limited:=true
4. Run [my example code that can be found here](https://gist.github.com/dudasdavid/174816564d0fd51b59aefc41819a1194).

### Your environment
* ROS Distro: Noetic
* OS Version: Ubuntu 20.04
[...]

PR #2722, opened 2021-06-16
Refactor moveit_servo::LowPassFilter to be assignable

griswaldbrooks:
### Description
`moveit_servo::LowPassFilter` is not currently assignable. At times, this makes it difficult to use with a `std::vector` without pointer semantics. There isn't an inherit reason for the type to be non-assignable and the members are not exposed in the public interface https://abseil.io/tips/177

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/include/moveit_servo/low_pass_filter.h
+++ b/moveit_ros/moveit_servo/include/moveit_servo/low_pass_filter.h
@@ -63,7 +63,7 @@ private:
   double previous_measurements_[FILTER_LENGTH];
   double previous_filtered_measurement_;
   // Scale and feedback term are calculated from supplied filter coefficient
-  const double scale_term_;
-  const double feedback_term_;
+  double scale_term_;
+  double feedback_term_;
 };
 }  // namespace moveit_servo


griswaldbrooks:
@tylerjw adding tests

moveit_ros/moveit_servo/include/moveit_servo/low_pass_filter.h
```diff
@@ -63,7 +63,7 @@ class LowPassFilter
   double previous_measurements_[FILTER_LENGTH];
   double previous_filtered_measurement_;
   // Scale and feedback term are calculated from supplied filter coefficient
-  const double scale_term_;
-  const double feedback_term_;
+  double scale_term_;
```
AndyZe, line 66: Unrelated to your changes, but ...

Would you mind deleting the TODO at Line 49? I don't see it happening anytime soon, because this filter is very lightweight and well-suited to the application.

AndyZe (approved):
Looks good! I'm familiar with why this is necessary.

moveit_ros/moveit_servo/include/moveit_servo/low_pass_filter.h
gbrooks2-pn: Sure. I'll do that when I add the test

tylerjw (approved):
This looks good, thank you.

gbrooks2-pn:
@v4hn there were still items on this PR that needed to be addressed. Should I have done something to block accidental mergers?

griswaldbrooks:
@v4hn there were still items on this PR that needed to be addressed. Should I have done something to block accidental mergers?

AndyZe:
Just open a new PR. But in the future, you could mark it with WIP or DRAFT while you're making changes.

v4hn:
@griswaldbrooks yes, you did not mark it as WIP, it clearly solves a minor issue already and was approved twice. So there was no reason not to merge it as is from my perspective.

I'm unsure what you mean by "items that needed to be addressed". You mentioned missing tests, but I'm not convinced a test for a valid assignment operator in one of a lot of datastructures in MoveIt even makes much sense. It might be more useful to integrate a scanner that can point out `const` members in datastructures that break assignment operations unless we mark them as deliberate.

griswaldbrooks:
@v4hn @AndyZe  Got it. In the future I will mark it WIP, and I'll submit a followup.

@v4hn I'm not sure I would use that diminutive as that issue (and others like it) can be at the root of other undesirable downstream decisions.

> You mentioned missing tests

Yes, I believe that tests both prevent regressions but also describe intent, and there is a simple gtest I can write which then describes this intent.

> It might be more useful to integrate a scanner
I'm not seeing it in clang-tidy as an option? Presumably because there are enough cases where a non-copyable objects are valid design decisions (as in non-value types), but then if it were then we could `//NOLINT` them to document intent

v4hn:
> I'll submit a followup.

:+1: 

> I'm not sure I would use that diminutive as that issue (and others like it) can be at the root of other undesirable downstream decisions.

Sorry, I did not mean to sound dismissive. I do consider it an issue, which is why I merged your patch.
It's just that the fix is obvious and clean, in contrast to unresolved issues like [this](https://github.com/ros-planning/moveit/issues/2677).

> Yes, I believe that tests both prevent regressions but also describe intent, and there is a simple gtest I can write which then describes this intent.

Sure, do that. I was just pointing out that we would probably want that for the better part of the 363 class definitions (source `grep -r '^class[^;]*$' --include *.h`) we have in this repository and it's tedious and repetitive to add these tests for all of them.
Personally I see movability & assignability as the default requirements of a class in MoveIt and we should probably explicitly add comments to those few classes that fail one of these requirements intentionally.

> I'm not seeing it in clang-tidy as an option

Indeed, though I do think there should be interest in it, so maybe someone should stipulate/write such an option...
Disabled assignment operators are one thing, but `const` members [also interfere with move semantics](https://www.reddit.com/r/cpp/comments/8wbeom/coding_guideline_avoid_const_member_variables/).

gbrooks2-pn:
> we have in this repository and it's tedious and repetitive to add these tests for all of them.

Maybe, but don't they need tests anyway?

> Personally I see movability & assignability as the default requirements of a class in MoveIt

Agreed, though I don't know much about MoveIt.

>  also interfere with move semantics

Yup, because for non-heap allocated types, move and copy are the same iirc.

v4hn:
> but don't they need tests anyway?

You're invited to write more tests for MoveIt in general! :cat2:
We are horribly short on them almost everywhere.

> >  also interfere with move semantics
> 
> Yup, because for non-heap allocated types, move and copy are the same iirc.

The link I referenced actually points out that (default) move constructors for classes with const members will invoke the copy-constructor for them instead (nothing else it could do), hiding allocations and silently getting exceptions involved.

PR #2726, opened 2021-06-19
Fix reversed check in switchControllers

nbbrooks:
### Description

While parsing through this code I saw a reversed sign on this check. This could cause two edge cases

Case 1: empty start_controllers and non-empty stop_controllers
This should cause the SwitchController to be made, but does not. Right now it is impossible for TrajectoryExecutionManager to pass in this condition though (if stop_controllers is non-empty, then by definition start_controllers will also be non-empty).

Case 2: empty start_controllers and empty stop_controllers
This should not cause the SwitchController to be made, but does. Right now TrajectoryExecutionManager does not invoke switchControllers if both lists are empty though.

TLDR This won't change anything behavior wise but should still be fixed.

### Checklist


v4hn (approved):
Yes!

PR #2730, opened 2021-06-22
Port #2721 (fixed padding collision attached objects) to Noetic

AndyZe:


v4hn:
`noetic-devel` is still fast-forwarded on the `master` branch. I just updated the `noetic-devel` branch instead.

PR #2732, opened 2021-06-22
Avoid push_back within getAttachedBodyObjects

AndyZe:
This is a pretty small optimization but I don't like to see `push_back()` called within a function that should be performant. This should be faster if there are multiple attached collision bodies.


v4hn (approved):
might be that fewer elements are needed than reserved (if `createCollisionGeometry` returns `nullptr`), but yes, reserving is much better. :+1:

PR #2737, opened 2021-06-24
Fix trajectory panel

rhaschke:
This fixes a display bug in the rviz `TrajectoryPanel`, @v4hn mentioned today.
The second commit is some related, but independent code cleanup.


v4hn (approved):
:1st_place_medal: You beat me to it, I just wanted to write a patch for this exact bug.

I should have mentioned the other 8 issues on my list as well. :smile: 
My devious plan was/is to create patches for them whenever I have some spare time.

Thanks! :+1:

rhaschke:
Go ahead a list them as issues ;-)
If they are related to the rviz plugins, I probably can fix them quite easily.

Issue #2739, opened 2021-06-25
Servo server JointTrajectory output incorrect when `use_gazebo` is true


turniphead:
### Description

Servo server trajectory output is incorrect when `use_gazebo` is set to true

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Source or Binary build? binary
* If binary, which release version? apt list info: `ros-melodic-moveit-servo/bionic,now 1.0.8-1bionic.20210601.174208 amd64 [installed]`
* If source, which branch?

### Steps to reproduce
Launch servo server, configure `use_gazebo: true` and `command_out_type: trajectory_msgs/JointTrajectory`. 

### Expected behaviour
Output Trajectory should be useable with a ros controller

### Actual behaviour
I get this error from the controller: `Trajectory message contains waypoints that are not strictly increasing in time.`. Specifically, the second point of the trajectory is completely empty.

### Backtrace or Console output

**look at the second point of the following trajectory**
trajectory message output from servo server: https://gist.github.com/turniphead/046a5ec9282820ac4013aab71cc1e5a7

console output: https://gist.github.com/turniphead/fe2d375848cddcea69ab2e02ec091bae

luisrayas3:
Suggested fix: https://github.com/ros-planning/moveit/pull/2740

JafarAbdi:
Fıxed by https://github.com/ros-planning/moveit/pull/2740

PR #2740, opened 2021-06-25
Fix an off-by-one error in servo_calcs.cpp

luisrayas3:
When both:

use_gazebo: true
command_out_type: trajectory_msgs/JointTrajectory

are set in the parameters a trajectory with redundant points is
published by servo_server. The 2nd point in this trajectory is always
empty (i.e. 0 timestamp and empty position/velocity/acceleration
arrays). This turns out to be because it is erroneously never filled in.
The timestamps need to be shifted relative to the point's index but all
of the points should still be filled in.

### Description

In the loop in `ServoCalcs::insertRedundantPointsIntoTrajectory`, be sure to populate index 1 of the outgoing array.

Corresponding issue: https://github.com/ros-planning/moveit/issues/2739

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -597,15 +597,15 @@ bool ServoCalcs::convertDeltasToOutgoingCmd(trajectory_msgs::JointTrajectory& jo
 
 // Spam several redundant points into the trajectory. The first few may be skipped if the
 // time stamp is in the past when it reaches the client. Needed for gazebo simulation.
-// Start from 2 because the first point's timestamp is already 1*parameters_.publish_period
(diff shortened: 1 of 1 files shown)


v4hn (approved):
This cleanup was apparently already implemented [in moveit2](https://github.com/ros-planning/moveit2/pull/374).
I wonder whether there are other similar issues that should be ported here as well...

@JafarAbdi Could you review/merge this please or port your moveit2 patch instead?

JafarAbdi (approved):

JafarAbdi:
> This cleanup was apparently already implemented [in moveit2](https://github.com/ros-planning/moveit2/pull/374).
> I wonder whether there are other similar issues that should be ported here as well...
> 
> @JafarAbdi Could you review/merge this please or port your moveit2 patch instead?

Sorry, that's my bad, next time I'll make sure bug fixes in moveit2 get back-ported to moveit

Issue #2741, opened 2021-06-26
A list of minor issues


v4hn:
[Robert encouraged me](https://github.com/ros-planning/moveit/pull/2737#issuecomment-868504325) to post my list of minor issues on github and I don't see why I should keep it to myself. ;)

These are some minor issues I found while making a short screencast of the RViz GUI.
No guarantees on some of those as I use a somewhat wild mix of dependency versions that might induce some of them.
If you want to discuss any of them please create a new issue. Otherwise feel free to write patches to your heart's content :flying_saucer: 

- Planning an RViz trajectory with the PTP planner through an obstacle does not visibly indicate a failure (though plan validation fails) in the gui
- collision markers (red spheres published on a separate topic) do not vanish until new collisions happen
  It might make sense to publish an empty message to the topic on successful planning?
- `roslaunch moveit_resources_panda_moveit_config demo.launch`, add a `MotionPlanningDisplay`, save the config, quit, relaunch
  -> the Motion Planning panel is disabled, though the display is enabled. toggling the display makes the panel appear
- in demo mode the fake joint publisher can end up publishing no current robot states to rviz anymore during and after executing a trajectory - disabling+enabling the display shows the new robot state though! I have no idea how to reproduce this.
- pilz planner needs PTP as default (in RViz)
  There is currently no default there and users are required to selected one because planning fails without specifying one!
  Instead let's just default to PTP here
- RViz configs in MSA templates and the resources still use `XYOrbit` view
  This is totally unusable and `Orbit` view with the center at the center of the used workspace of the robot should be preferred.
- pilz should only complain about Multi-DOF joints when such a joint should actually be *planned for* (only shows before https://github.com/ros-planning/moveit_resources/pull/86)
- If you scroll through a planned trajectory with the slider (autoplay paused), and afterwards move the marker for the goal state again the robot state showing the trajectory waypoint will still be there and is relatively hard to get rid of. Moving the slider to the end/and or resuming play in the trajectory panel should probably disable the state to cleanup the screen when the user moved on.

Already done/PR created:

- geometric::TrajOpt is not implemented but in panda_moveit_config https://github.com/ros-planning/moveit_resources/pull/85
- on replay trajectory panel says "Play" even when the button stops #2737

rhaschke:
Why is the `XYOrbit` view "totally unusable"? I think that's the default in rviz.

v4hn:
> Why is the `XYOrbit` view "totally unusable"? I think that's the default in rviz.

[...]

PR #2744, opened 2021-06-28
Fix joints tab

rhaschke:
Fix issues of https://github.com/ros-planning/moveit/issues/2741 regarding the joints tab:
- switch order of `manipulation` and `joints` tabs
- correctly handle mimic joints (don't allow editing)


v4hn (approved):
Thanks for looking into some issues on my list!
I wanted to test this locally before merging.
I also added more new joint tab issues to the meta issue if you want to look at them. ;)

PR #2745, opened 2021-06-28
support arbitrary realtime-factors in trajectory visualization

v4hn:
The fact that trajectory visualization ignored the time parameterization by default
and that the only way to change this is the special value "REALTIME" always bothered me.

I changed the default to a realtime factor of 3x to get a nice preview of the trajectory
when using the "Plan&Execute" button.

https://user-images.githubusercontent.com/680358/123707186-9cdfa400-d869-11eb-9e94-95935f15d163.mp4

### DIFF
--- a/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
+++ b/moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
@@ -90,14 +90,17 @@ TrajectoryVisualization::TrajectoryVisualization(rviz::Property* widget, rviz::D
   robot_path_alpha_property_->setMin(0.0);
   robot_path_alpha_property_->setMax(1.0);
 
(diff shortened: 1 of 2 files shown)


moveit_ros/visualization/rviz_plugin_render_tools/src/trajectory_visualization.cpp
```diff
@@ -337,24 +340,49 @@ void TrajectoryVisualization::interruptCurrentDisplay()
...
-    {
-      t = boost::lexical_cast<float>(tm);
-    }
-    catch (const boost::bad_lexical_cast& ex)
-    {
-      state_display_time_property_->setStdString("0.05 s");
-    }
-    return t;
+    state_display_time_property_->setStdString(DEFAULT_TIME_STRING);
+    return DEFAULT_TIME_VALUE;
```
rhaschke, line 362: It is somewhat weird that you call a setter within this getter method, but I cannot propose a nicer solution w/o duplicating code.

rhaschke (approved):

PR #2747, opened 2021-06-30
Remove faulty gtest include

v4hn:
See https://github.com/ros-planning/moveit2/pull/526 for MoveIt2 patch.

https://github.com/ros-planning/moveit/pull/2471 was not sufficiently reviewed.

@henningkayser


rhaschke (approved):
Looks good. `gtest.h` should only be included in testing source.

PR #2748, opened 2021-06-30
use $(dirname) in MoveIt Setup Assistant

v4hn:
Available since lunar. This simplifies the descriptions in moveit_resources
quite a bit because their package name does not match the include path
and we have to adapt the names everywhere after edits.

Note that two candidates for this replacement still remain -
they define the file path to the rviz config in an arg value.
Due to [a bug](https://github.com/ros/ros_comm/issues/1487) these two can not be replaced at this moment.

@gavanderhoorn Could you please review this patch?


gavanderhoorn (approved):
Looks ok to me.

I assume the main motivation is:

> This simplifies the descriptions in moveit_resources quite a bit because their package name does not match the include path and we have to adapt the names everywhere after edits.

as there isn't anything too 'complex' about the expressions here in the MSA templates.

v4hn:
> I assume the main motivation is [...]

That's why I came up with the patch, yes. Also I believe it's the better approach for including launch files in the same folder anyway, but that alone would not justify the template changes.

Thanks for the review.

PR #2751, opened 2021-07-06
minor RobotTrajectory patches

v4hn:
Please look at the individual commits for details.

### DIFF
--- a/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
+++ b/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
@@ -51,8 +51,15 @@ MOVEIT_CLASS_FORWARD(RobotTrajectory);  // Defines RobotTrajectoryPtr, ConstPtr,
 class RobotTrajectory
 {
 public:
+  /** @brief construct a trajectory */
+  explicit RobotTrajectory(const moveit::core::RobotModelConstPtr& robot_model);
+
+  /** @brief construct a trajectory for the named JointModelGroup */
   RobotTrajectory(const moveit::core::RobotModelConstPtr& robot_model, const std::string& group);
 
+  /** @brief construct a trajectory for the JointModelGroup
+   *  If group is nullptr this is equivalent to the first constructor
+   */
   RobotTrajectory(const moveit::core::RobotModelConstPtr& robot_model, const moveit::core::JointModelGroup* group);
 
   /** Assignment operator, performing a shallow copy, i.e. copying waypoints by pointer */
--- a/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
+++ b/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
@@ -41,6 +41,7 @@
 #include <moveit_msgs/RobotTrajectory.h>
 #include <moveit_msgs/RobotState.h>
 #include <deque>
+#include <memory>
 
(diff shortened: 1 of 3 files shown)


moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
```diff
@@ -75,7 +77,7 @@ class RobotTrajectoryTestFixture : public testing::Test
     double duration_from_previous = 0.1;
     std::size_t waypoint_count = 5;
     for (std::size_t ix = 0; ix < waypoint_count; ++ix)
-      trajectory->addSuffixWayPoint(robot_state_, duration_from_previous);
+      trajectory->addSuffixWayPoint(*robot_state_, duration_from_previous);
```
felixvd, line 80: Why did this API change, even though the function itself did not seem to?

felixvd (commented):
1. The chain setter support is cool, but I've seen it nowhere else in the codebase.
2. I can't comment on the `make_shared`, but the rest LGTM.

v4hn:
On Wed, Jul 07, 2021 at 03:11:44AM -0700, Felix von Drigalski wrote:
> 1. The chain setter support is cool, but I've seen it nowhere else in the codebase.

Yes, we could use it in many other places I agree :-)

> 2. I can't comment on the `make_shared`, but the rest LGTM.

Aside from the fact that it doesn't build right now. ;)
I'll look into it.

> -      trajectory->addSuffixWayPoint(robot_state_, duration_from_previous);
> +      trajectory->addSuffixWayPoint(*robot_state_, duration_from_previous);
> 
> Why did this API change, even though the function itself did not seem to?

There are two interfaces in the API, one taking a pointer and one taking an actual RobotState.
The latter will *copy* the robot state, but the former will only add the pointer.
I wanted to have a more proper trajectory for these tests where not all the waypoints refer to the same robot state.

moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
rhaschke: I stumbled over this question as well. Here is the answer:
Actually both API exists: taking a `RobotState&` and `RobotStatePtr`. While the former creates a new state copy, the latter uses the pointer as is. I guess, @v4hn wanted to ensure that a copy is created here.

rhaschke (approved):
Looks good to me. @v4hn, please decide for yourself how you want to merge this (squash vs. merge commit).

moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
```diff
@@ -51,8 +52,15 @@ MOVEIT_CLASS_FORWARD(RobotTrajectory);  // Defines RobotTrajectoryPtr, ConstPtr,
...
 {
 public:
+  /** @brief construct a trajectory */
+  explicit RobotTrajectory(const moveit::core::RobotModelConstPtr& robot_model);
+
+  /** @brief construct a trajectory for the named JointModelGroup */
   RobotTrajectory(const moveit::core::RobotModelConstPtr& robot_model, const std::string& group);
 
+  /** @brief construct a trajectory for the JointModelGroup
+   *  If group is nullptr this is equivalent to the first constructor
```
simonschmeisser, line 62: Sorry, late to the party, could you please explain (in another PR) in the comments what it actually means for a `RobotTrajectory` to have a group set (or what happens when you use the default constructor)

:+1: for starting with the comments!

PR #2622, opened 2021-04-26
Add planning_pipeline_id setting to Python MGI

felixvd:
This does what the title says.

PR #2753, opened 2021-07-07
Fix get_planning_pipeline_id in Python MGI

### LINKED ISSUE #2752: Extra argument in get_planning_pipeline_id
### Description

There is an extra argument in get_planning_pipeline_id, which should be argument free.

### Your environment
* ROS Distro: Noetic
* If binary, which release version? 1.1.5

### Steps to reproduce
https://answers.ros.org/question/381866/how-to-use-get_planning_pipeline_id/?comment=381929#post-id-381929

felixvd:
Fixes #2752 , which was introduced with the new feature in #2622. I should have made `get_planning_pipeline_id` without an argument.

It also adds some debugging prints, which should be helpful.


v4hn (approved):
dumb bug indeed...

Issue #2676, opened 2021-05-20
MoveGroupCommander + pilz industrial motion planner not using the updated allowed collision matrix


robertjbush:
### Description

Asked by @felixvd to raise an issue ticket ([see this ROS answers question](https://answers.ros.org/question/378602/pilz-industrial-motion-planner-not-using-the-updated-allowed-collision-matrix/))

### My environment
* ROS Noetic in a docker container with Ubuntu 20.04
* Moveit! version 1.1.2
* ros-industrial-client with motoman driver
* motoman gp8 robot
* binary build

### Problem
* Using MoveGroup commander (Python API) and Pilz Industrial Motion Planner, changes to the allowed collision matrix (ACM) are published to the planning scene but do not propagate to the planning interface.

### Action Taken So Far
* Tested with an identical stack except with OMPL planner in the pipeline. Behaviour is not reproducible, changes propagate to the planning interface (perhaps through pre/post-processing plugins?).
* Tested with planning directly in rviz. Behaviour is not reproducible, changes propagate to the planning interface.

### Steps to reproduce

My base application is a docker container and that is detailed below.
```
ARG ROS_DISTRO=noetic
ARG BASE_IMAGE=ros:${ROS_DISTRO}-ros-base
ARG UID=0
ARG GID=0
ARG HOME=/root

FROM ${BASE_IMAGE} AS base

RUN apt-get update && apt-get upgrade -y && \
  apt-get install -y --no-install-recommends \
  clang-format \
  clang-tidy \
  cppcheck \
  git-core \
  libgl1-mesa-dri \
  libgl1-mesa-glx \
  libmuparser-dev \
  libopenblas-dev \
  libxml2-utils \
  nodejs \
  npm \
  python3-pip \
  python3-psutil \
  python3-vcstool \
  python-is-python3 \
  ros-${ROS_DISTRO}-controller-interface \
  ros-${ROS_DISTRO}-controller-manager \
  ros-${ROS_DISTRO}-control-toolbox \
  ros-${ROS_DISTRO}-diagnostic-updater \
  ros-${ROS_DISTRO}-eigen-conversions \
  ros-${ROS_DISTRO}-filters \
  ros-${ROS_DISTRO}-hardware-interface \
  ros-${ROS_DISTRO}-joint-limits-interface \
  ros-${ROS_DISTRO}-joint-state-controller \
  ros-${ROS_DISTRO}-joint-state-publisher \
  ros-${ROS_DISTRO}-moveit-core \
[...]

PR #2754, opened 2021-07-07
Bugfix/fix pilz collision detection moveit pr

aa-tom:
### Description

The pilz industrial motion planner does not account for updates to the allowed collision matrix when determining collision-free paths due to it creating a new `PlanningScene`, which obtains this information from the SRDF. This is covered in #2676 

This PR passes the most recent `PlanningScene` object obtained via MoveIt down to the relevant collision detection code, allowing changes to the collision matrix to propagate, thus allowing pilz to find paths which would otherwise be deemed unsuitable. 

This work includes:

- Modifying function signatures to accept the new `PlanningScene`
- Modifying docstrings to include the new `PlanningScene`
- Modifying function implementations to pass the new `PlanningScene` down to the next relevant function
- Using the new `PlanningScene` rather than the default one when performing collision detection [here](https://github.com/additiveautomations/aa-pilz/blob/27ec4894b6f3a17bb9a593e2b93e0d01568cad59/moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp#L576)
- Some functions now call functions accepting the `PlanningScene` argument, but do not have access to one themselves. In these instances, the code now builds a new `PlanningScene` using the robot state to pass down to the function in question [here](https://github.com/additiveautomations/aa-pilz/blob/27ec4894b6f3a17bb9a593e2b93e0d01568cad59/moveit_planners/pilz_industrial_motion_planner/test/test_utils.cpp#L1044) and [here](https://github.com/additiveautomations/aa-pilz/blob/27ec4894b6f3a17bb9a593e2b93e0d01568cad59/moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp#L91) (I'm unsure whether this is the best approach, or if we should pass the `PlanningScene` from MoveIt down to these functions too. Advice here would be much appreciated)

The video below demonstrates successful pilz path finding for a non-collision-free path where the particular collisions have been allowed via the collision matrix (one which failed in the example video given in #2676 ):

https://user-images.githubusercontent.com/86778266/124609016-280efa00-de67-11eb-8b5e-3f097447940b.mp4

### Checklist
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/) (Not relevant)
- [x] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes (Not relevant)
- [x] Include a screenshot if changing a GUI (Not relevant)


moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
```diff
@@ -85,7 +86,10 @@ bool pilz_industrial_motion_planner::TrajectoryBlenderTransitionWindow::blend(
   }
   trajectory_msgs::JointTrajectory blend_joint_trajectory;
   moveit_msgs::MoveItErrorCodes error_code;
-  if (!generateJointTrajectory(req.first_trajectory->getFirstWayPointPtr()->getRobotModel(),
+
+  const moveit::core::RobotModelConstPtr& robot_model = req.first_trajectory->getFirstWayPointPtr()->getRobotModel();
+  const planning_scene::PlanningSceneConstPtr scene(new planning_scene::PlanningScene(robot_model));
```
rhaschke, line 91: Here a new (empty) scene is created from scratch. Isn't that exactly what you wanted to avoid?

moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp
```diff
@@ -40,7 +40,8 @@
 #include <tf2_kdl/tf2_kdl.h>
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 
-bool pilz_industrial_motion_planner::computePoseIK(const moveit::core::RobotModelConstPtr& robot_model,
+bool pilz_industrial_motion_planner::computePoseIK(const planning_scene::PlanningSceneConstPtr& scene,
+                                                   const moveit::core::RobotModelConstPtr& robot_model,
```
rhaschke, line 44: Can't you get rid off the RobotModel argument? The PlanningScene is linked to a RobotModel.
This comment applies to several similar methods below!

moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp
```diff
@@ -571,8 +573,7 @@ bool pilz_industrial_motion_planner::isStateColliding(const bool test_for_self_c
   collision_detection::CollisionRequest collision_req;
   collision_req.group_name = group->getName();
   collision_detection::CollisionResult collision_res;
-  planning_scene::PlanningScene(robot_model).checkSelfCollision(collision_req, collision_res, *rstate);
-
```
rhaschke, line 575: Ouch. Thanks for spotting this issue!

rhaschke (changes requested):
Thanks for providing this fix! Overall this looks reasonable to me.
There is still a PlanningScene generated from scratch - see below.

A more general comment: Please don't include merge commits in PRs. Instead, rebase your PR branch onto latest master if needed.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp
aa-tom: Sure thing :+1:

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
aa-tom: It is... If you have an idea of which route of function calls could be used to pass the scene down here that'd be super useful, otherwise I can have a dig through the code

moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp
aa-tom: Credit to @cambel for identifying the root cause :smiley:

moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp
v4hn: You are right, you probably have to pass it down quite along way. Nonetheless It's definitely required here or trajectory blending can yield invalid trajectories that were fine before.
I expect the `PlanComponentsBuilder` will always operate on a single PlanningScene (with varying robot states), so what about replacing `PlanComponentsBuilder::setRobotModel` by `PlanComponentsBuilder::setPlanningScene`?
I did not try it, but it seems reasonable at a glance.

v4hn:
For [the second fresh scene](https://github.com/additiveautomations/aa-pilz/blob/27ec4894b6f3a17bb9a593e2b93e0d01568cad59/moveit_planners/pilz_industrial_motion_planner/test/test_utils.cpp#L1044) I would propose to replace the first parameter (`RobotModel`) of `testutils::generateTrajFromBlendTestData` by a `PlanningScene` and choose at call side whether it's ok for the test to use a fresh scene.

rhaschke:
@aa-tom, note that you missed to adapt API of the pilz unit tests. See CI results for details.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp
```diff
@@ -571,8 +573,7 @@ bool pilz_industrial_motion_planner::isStateColliding(const bool test_for_self_c
   collision_detection::CollisionRequest collision_req;
   collision_req.group_name = group->getName();
   collision_detection::CollisionResult collision_res;
```
cambel, line 575: Could you add the verbose argument to this collision request `req.verbose = true`? otherwise, the user gets a `No IK solution found` which is true, but the real reason is that a collision was detected at some point. That should print the links that were found to be in a collision.

rhaschke:
@aa-tom, when do you have time to finish the remaining issues of this open PR?

Please, remember not to use merge commits for PR branches! Please rebase instead onto master:
```
git rebase master
git push --force-with-lease
```

aa-tom:
> @aa-tom, when do you have time to finish the remaining issues of this open PR?
> 
> Please, remember not to use merge commits for PR branches! Please rebase instead onto master:
> 
> ```
> git rebase master
> git push --force-with-lease
> ```

Sorry for leaving this one on the wayside! I'm pretty busy this week, but should be able to get some time next week to address the changes. Will rebase too.

felixvd:
Friendly ping @aa-tom

robertjbush:
@felixvd Tom's been working on another bug in the ROS-I package Noether. I spoke to him and this is his next priority, apologises for the delay. According to the calendar your next release is due to start preparing 3rd September, is that correct?

v4hn:
> According to the calendar your next release is due to start preparing 3rd September, is that correct?

Sadly the calendar is more of a very rough guideline and was not adhered to in recent times.
More likely, the next release will happen once there is enough interest in it.

aa-tom:
Reopened [here](https://github.com/ros-planning/moveit/pull/2803) with the suggested changes (just need to update the unit tests_

PR #2755, opened 2021-07-07
Add planning_pipeline_id to MotionSequence service

felixvd:
In #2657 I added code to select the planning pipeline and planner to the `getMotionSequence` action. This is required to use the action and service when both OMPL and the Pilz planning pipelines are loaded, because `getMotionSequence` needs to use Pilz, but selects the default pipeline instead (OMPL).

I recently noticed that it should also be added to the service, which this PR does. It works, so if you need this feature you can use it. But it breaks some tests, and they seem to be in functionality (`Sequence`) that I don't have the time to dig into right now. Maybe @ct2034 or @jschleicher want to have a look.


jschleicher (approved):
👍

jschleicher:
The first test case checks for an empty request, which segfaulted with your unchecked list access ;-)

felixvd:
Very nice! Thanks for fixing my oversight :)

PR #2761, opened 2021-07-09
controller manager: enclose controller name in quotes

gavanderhoorn:
So empty controller names are easier to spot.

Some MoveIt configurations have the controller `name` set to `""`. This is valid (although annoying), but the formatting of the log lines makes it difficult to spot this.

Without the quotes:

```
Controller  failed with error INVALID_GOAL: ...
```

With the quotes:

```
Controller '' failed with error INVALID_GOAL: ...
```

### DIFF
--- a/moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
+++ b/moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
@@ -199,13 +199,13 @@ void FollowJointTrajectoryControllerHandle::controllerDoneCallback(
 {
   // Output custom error message for FollowJointTrajectoryResult if necessary
   if (!result)
-    ROS_WARN_STREAM_NAMED(LOGNAME, "Controller " << name_ << " done, no result returned");
+    ROS_WARN_STREAM_NAMED(LOGNAME, "Controller '" << name_ << "' done, no result returned");
   else if (result->error_code == control_msgs::FollowJointTrajectoryResult::SUCCESSFUL)
-    ROS_INFO_STREAM_NAMED(LOGNAME, "Controller " << name_ << " successfully finished");
+    ROS_INFO_STREAM_NAMED(LOGNAME, "Controller '" << name_ << "' successfully finished");
   else
-    ROS_WARN_STREAM_NAMED(LOGNAME, "Controller " << name_ << " failed with error "
-                                                 << errorCodeToMessage(result->error_code) << ": "
-                                                 << result->error_string);
+    ROS_WARN_STREAM_NAMED(LOGNAME, "Controller '" << name_ << "' failed with error "
+                                                  << errorCodeToMessage(result->error_code) << ": "
+                                                  << result->error_string);
   finishControllerExecution(state);
 }
 


gavanderhoorn:
Seriously. I add 2 single-quotes and the format check is suddenly failing?

moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
```diff
@@ -199,13 +199,13 @@ void FollowJointTrajectoryControllerHandle::controllerDoneCallback(
...
   else if (result->error_code == control_msgs::FollowJointTrajectoryResult::SUCCESSFUL)
-    ROS_INFO_STREAM_NAMED(LOGNAME, "Controller " << name_ << " successfully finished");
+    ROS_INFO_STREAM_NAMED(LOGNAME, "Controller '" << name_ << "' successfully finished");
   else
-    ROS_WARN_STREAM_NAMED(LOGNAME, "Controller " << name_ << " failed with error "
-                                                 << errorCodeToMessage(result->error_code) << ": "
-                                                 << result->error_string);
+    ROS_WARN_STREAM_NAMED(LOGNAME, "Controller '" << name_ << "' failed with error "
+                                                  << errorCodeToMessage(result->error_code) << ": "
+                                                  << result->error_string);
```
simonschmeisser, line 208:
```suggestion
                                                  << result->error_string.empty() ? state.getText() : result->error_string);
```
maybe? maybe not?

gavanderhoorn: I already fixed it. The formatting job shows the suggested formatting. The `<<` needed to be moved to the right by 1 space.

gavanderhoorn:
Windows CI is broken afaik.

moveit_plugins/moveit_simple_controller_manager/src/follow_joint_trajectory_controller_handle.cpp
simonschmeisser: this wasn't about the formatting but about displaying the generic action error string in case the other one is empty

gavanderhoorn: `/friday_afternoon`.

Sorry.

gavanderhoorn: Should probably be added in a separate PR though.

simonschmeisser: Hehe, let's not micro-PR things? It's the same topic after all

v4hn: > Hehe, let's not micro-PR things? It's the same topic after all

In theory I agree with @simonschmeisser, but I understand why we should not delay PRs because of other issues in the same code lines. Please create a quick follow-up PR for your improvement Simon! :medal_sports:

v4hn (approved):

v4hn:
> ctrlr mngr: enclose ctrlr name in quotes (#2761)

You lost some vowels there ;)

Issue #2757, opened 2021-07-08
use std::make_shared where applicable


v4hn:
MoveIt is sprinkled with `...SharedPtr::reset(new ...)` code and `SharedPtr(new ...)` constructors.
It is slightly (though in most cases not noticeably) faster and cleaner to use `std::shared_ptr<...>()` instead.

Anyone is welcome to help clean this up and create a pull-request along the lines of #2756 for wider parts of the code base. :muscle:

pvanlaar:
I like to help! @v4hn, could you assign me? regards Patrick

jaskiratsingh2000:
Hey  I am quite beginner and super interested to begin my contribution towards this repo and this issue seems interesting to me. Can you please let me know that where exactly do I have to make changes or add functionalities?
Any examples corresponding to this would be highly appreciable. Thanks!

jaskiratsingh2000:
Is this still open?

rhaschke:
As you can see a few lines above, this issue was tackled in #2762. However, because that PR is not yet merged (due to some open issues), this issue is not yet closed yet as well.

jaskiratsingh2000:
So @rhaschke Do you think that I should work on this?

jaskiratsingh2000:
@rhaschke Can you please assign this to me?

rhaschke:
I don't think there is anything you can do. I'm looking into the open issues right now.

jaskiratsingh2000:
@rhaschke Yes, please redirect me to super beginner friendly issue as I am supper interested to work under this repo. So I am looking forward to hearing from you

pvanlaar:
The PR is merged, so this issue can be closed.

rhaschke:
Closed via #2762.

PR #2762, opened 2021-07-09
Replaced ptr::reset( new Class) with make_shared or unique

pvanlaar:
### Description

As described in #2757, cleaned up the `SmartPointer::reset(new ...)`calls. They were replaced with std::make_shared or std::make_unique. Not using new or delete is a C++ best practice ([C++ Core guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-newdelete)). 

### Scope
To keep the PR's relatively concise, I only changed the Ptr::reset( occurrences that could be changed within a few lines of code. There are for example other Ptr::reset() calls that take the result of factory functions producing raw pointers. Cleaning that up would requiring changing the functions and I considered them out of the scope of this PR. If someone has an opinion about the scope, please let me know.

Edit: Removed catkin tools issue, because the CI build for ubuntu passes.

### Checklist
~~- [] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)~~ No functional changes
~~- [] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes~~ No API changes
~~- [ ] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html)~~ No functional changes


moveit_planners/ompl/ompl_interface/src/ompl_planner_manager.cpp
```diff
@@ -295,8 +295,6 @@ class OMPLPlannerManager : public planning_interface::PlannerManager
       pub_valid_states_ = nh_.advertise<moveit_msgs::DisplayRobotState>("ompl_planner_valid_states", 5);
       pub_valid_traj_ = nh_.advertise<moveit_msgs::DisplayTrajectory>("ompl_planner_valid_trajectories", 5);
       display_random_valid_states_ = true;
-      //    pub_valid_states_thread_.reset(new boost::thread(boost::bind(&OMPLPlannerManager::displayRandomValidStates,
-      //    this)));
```
v4hn, line 299: Please do not remove this comment. If anything, we should remove the whole commented out blocks instead and that would belong into a separate PR.

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.1.3)
 project(pilz_industrial_motion_planner)
 
 ## Add support for C++11, supported in ROS Kinetic and newer
-add_definitions(-std=c++11)
+add_definitions(-std=c++14)
```
v4hn, line 5: This is unrelated and should not be in the same PR/commit. Otherwise it's a good change though.

moveit_planners/sbpl/ros/sbpl_interface_ros/src/sbpl_meta_plugin.cpp
```diff
@@ -50,8 +50,7 @@ class SBPLMetaPlanner : public planning_interface::Planner
   void init(const planning_models::RobotModelConstPtr& model)
   {
     ros::NodeHandle nh;
-    // display_bfs_publisher_ = nh.advertise<visualization_msgs::Marker>("planning_components_visualization", 10, true);
```
v4hn, line 53: Why would you remove that line? Same as with the one in OMPL we should cleanup the context for this in a different PR if we remove commented-out code.

moveit_ros/perception/mesh_filter/include/moveit/mesh_filter/depth_self_filter_nodelet.h
```diff
@@ -105,10 +105,10 @@ class DepthSelfFiltering : public nodelet::Nodelet
...
   TransformProvider transform_provider_;
 
-  cv_bridge::CvImagePtr filtered_depth_ptr_;
-  cv_bridge::CvImagePtr filtered_label_ptr_;
-  cv_bridge::CvImagePtr model_depth_ptr_;
-  cv_bridge::CvImagePtr model_label_ptr_;
+  std::shared_ptr<cv_bridge::CvImage> filtered_depth_ptr_;
+  std::shared_ptr<cv_bridge::CvImage> filtered_label_ptr_;
+  std::shared_ptr<cv_bridge::CvImage> model_depth_ptr_;
+  std::shared_ptr<cv_bridge::CvImage> model_label_ptr_;
```
v4hn, line 111: This moves from `boost` to `std`. Our own updaters do not set this from any API that relies on the boost pointers, so I'm ok with the change in `master`.

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -103,21 +103,20 @@ rdf_loader::RDFLoader::RDFLoader(const std::string& urdf_string, const std::stri
   moveit::tools::Profiler::ScopedStart prof_start;
   moveit::tools::Profiler::ScopedBlock prof_block("RDFLoader(string)");
 
-  urdf::Model* umodel = new urdf::Model();
```
v4hn, line 106: yay, that one actually fixes potential memory leaks from what I can see :bouquet:

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -1333,7 +1333,7 @@ class MoveGroupInterface::MoveGroupInterfaceImpl
 
   // pose goal;
   // for each link we have a set of possible goal locations;
-  std::map<std::string, std::vector<geometry_msgs::PoseStamped> > pose_targets_;
+  std::map<std::string, std::vector<geometry_msgs::PoseStamped>> pose_targets_;
```
v4hn, line 1336: These changes in various places are unrelated, but might be required by clang-tidy?
I don't have a strong opinion on whether it's ok to have them in the same PR though.

v4hn (approved):
Thank you! I should create such issues more often if there are actually users out there who are willing to contribute such nice cleanups! :1st_place_medal: 

In general I approve. Please reply/address my inline comments though.

I will leave the second review to @tylerjw as he pushed for these changes most in the past :sunglasses:

moveit_core/collision_distance_field/src/collision_env_distance_field.cpp
```diff
@@ -1088,7 +1088,7 @@ PosedBodySphereDecompositionPtr
 CollisionEnvDistanceField::getPosedLinkBodySphereDecomposition(const moveit::core::LinkModel* ls, unsigned int ind) const
 {
   PosedBodySphereDecompositionPtr ret;
-  ret.reset(new PosedBodySphereDecomposition(link_body_decomposition_vector_[ind]));
+  ret = std::make_shared<PosedBodySphereDecomposition>(link_body_decomposition_vector_[ind]);
```
tylerjw, line 1091: Small nit can you change vector operator[] to at()?  Feel free to ignore this and we can do it in a separate change.

tylerjw (approved):
Generally approve, fix the small things @v4hn asked for and then let's merge this.

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
pvanlaar: This is necessary to use std::make_unique as that function was added later (c++ 14) to the standard than std::make_shared (c++ 11) for some reason. Do you agree that this can be changed here?

moveit_planners/ompl/ompl_interface/src/ompl_planner_manager.cpp
pvanlaar: Here I react to all comments on the removal of commented out code. I thought to remove these commented lines as they are dead and not subject to CI checks. Commented out code is an [official code smel](https://moderatemisbehaviour.github.io/clean-code-smells-and-heuristics/comments/c5-commented-out-code.html). When I encountered these comments I simply deleted them, as I thought that searching for them later is difficult and cumbersome. I will change all comments back and add them to a separate PR.

moveit_ros/perception/mesh_filter/include/moveit/mesh_filter/depth_self_filter_nodelet.h
pvanlaar: Thank you for clarifying the rationale behind allowing the change. Could you elaborate a bit more, because I don't fully understand it. 

- What are the updaters? Are they clean up scripts? 
- Do you mean with the second half, that these pointers are not exposed publically, so they can be changed with this PR? Otherwise, it has to be done with a deprecation?
- Do you mean that this change should only be in `master` not in `noetic-devel`?

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
pvanlaar: Apologies, These changes were done by running clang-format. They were required to be allowed to merge. I don't know how it possible that running clang-format on files I did not alter, changes the file in some places.

moveit_core/collision_distance_field/src/collision_env_distance_field.cpp
pvanlaar: Good catch. I will change it.

pvanlaar:
Thank you both for the quick reviews and the nice words. One question about the failing Windows build. Do we have to wait on that fix to be allowed to merge?

rhaschke:
The windows build should be fixed. Please, just push your fixups and CI will be re-triggered.

pvanlaar:
Great! Thanks for the quick reply.

moveit_planners/ompl/ompl_interface/src/ompl_planner_manager.cpp
v4hn: > Commented out code is an official code smell

I agree! However, in these parts it's actually worse than commented out code.
Some of these commented out functions would actually be useful from my perspective but do not apply anymore because of API changes over time.
In other cases, also in this particular one, the actual functionality is commented out, but not the whole related structure.
The DynamicReconfigure option still exists!

moveit_planners/pilz_industrial_motion_planner/CMakeLists.txt
v4hn: I see. Just leave it in then. But I would appreciate if you could update the comment above the line too. ;)

In general it's a good change because our whole code base requires C++14 and the pilz planner was added later but not adapted.

moveit_ros/perception/mesh_filter/include/moveit/mesh_filter/depth_self_filter_nodelet.h
v4hn: - We do not differentiate between `master` and `noetic-devel`.
- Ignore the parts about updaters, I thought this change was in a different module when I wrote the comment.
- These are private members of an implementation class, so there is no harm in changing them.

jliukkonen (approved):
Cool, thanks for cleaning up the codebase!

pvanlaar:
With the latest rebase that includes the latest commits on `noetic-devel` branch, I hope to address the Windows build failure. The PR check clang-tidy did also fail. I ran clang-tidy locally as described in the [docs](https://moveit.ros.org/documentation/contributing/code/), but it fixed many things that were not altered in this PR. I included these fixes in a separate commit. Does anyone know why this happens? I had a similar experience with clang-format.

rhaschke:
> I ran clang-tidy locally as described in the [docs](https://moveit.ros.org/documentation/contributing/code/), but it fixed many things that were not altered in this PR. I included these fixes in a separate commit. Does anyone know why this happens?

Do you use the same versions of clang-tidy / clang-format as in CI?

pvanlaar:
I ran version 10.0.0 of both clang-format and clang-tidy locally, which I understood to be the same versions as the CI build. Correct me if I am wrong.  Exactly the same commands as the [Style guidelines](https://moveit.ros.org/documentation/contributing/code/) were used.

jliukkonen:
>C:\Users\runneradmin\micromamba\envs\test\Library\include\boost/bind.hpp(41): note: The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.

While I can't tell you why the Windows build is the only one giving that error, the solution seems to be simple. There's some discussion here: [https://stackoverflow.com/a/65813178](https://stackoverflow.com/a/65813178)

Basically you would do this:
```
boost::bind(&MoveGroupSequenceAction::executeSequenceCallback, this, boost::placeholders::_1)
```
instead of this:
```
boost::bind(&MoveGroupSequenceAction::executeSequenceCallback, this, _1)
```
And then you would also change the include to point to `#include <boost/bind/bind.hpp>` instead of `#include <boost/bind.hpp>`. That fix doesn't seem unreasonable to apply to the codebase although it will end up being probably 50+ changes.

rhaschke:
The boost placeholders are just warnings (probably because the Windows build uses a more recent boost version).
The actual error is this one:
`C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.16.27023\include\type_traits(1271): error C2338: You've instantiated std::aligned_storage<Len, Align> with an extended alignment (in other words, Align > alignof(max_align_t)). Before VS 2017 15.8, the member type would non-conformingly have an alignment of only alignof(max_align_t). VS 2017 15.8 was fixed to handle this correctly, but the fix inherently changes layout and breaks binary compatibility (*only* for uses of aligned_storage with extended alignments). Please define either (1) _ENABLE_EXTENDED_ALIGNED_STORAGE to acknowledge that you understand this message and that you actually want a type with an extended alignment, or (2) _DISABLE_EXTENDED_ALIGNED_STORAGE to silence this message and get the old non-conformant behavior.`

rhaschke:
Weird, that there are still clang-tidy issues...

jliukkonen:
> The boost placeholders are just warnings (probably because the Windows build uses a more recent boost version).
> The actual error is this one:
> `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.16.27023\include\type_traits(1271): error C2338: You've instantiated std::aligned_storage<Len, Align> with an extended alignment (in other words, Align > alignof(max_align_t)). Before VS 2017 15.8, the member type would non-conformingly have an alignment of only alignof(max_align_t). VS 2017 15.8 was fixed to handle this correctly, but the fix inherently changes layout and breaks binary compatibility (*only* for uses of aligned_storage with extended alignments). Please define either (1) _ENABLE_EXTENDED_ALIGNED_STORAGE to acknowledge that you understand this message and that you actually want a type with an extended alignment, or (2) _DISABLE_EXTENDED_ALIGNED_STORAGE to silence this message and get the old non-conformant behavior.`

Ah yes, apologies for my misleading comment @pvanlaar. However, if you do end up fixing the Boost related warnings too, that would be a nice improvement. As for this error, I don't know which one of those solutions would be the correct one.

pvanlaar:
Thank you both for the replies. Don't worry @jliukkonen, I appreciate the help. Issue #2766 would solve those boost::bind warnings, so we won't have to create a separate issue for that. I will have a look at the current merge fails.

rhaschke:
@Tobias-Fischer and/or @wolfv: Could you have a look at https://github.com/ros-planning/moveit/pull/2762#issuecomment-880936888 and suggest a solution?
My gut feeling is to add the flag `_ENABLE_EXTENDED_ALIGNED_STORAGE` via build options in the CI script.

wolfv:
Thanks for the ping. I think the _ENABLE_EXTENDED_ALIGNED_STORAGE is probably what we want, longterm. 
I cannot really judge if this will have implications on the ABI of MoveIt. 
Ideally this flag would be applied somewhere higher level (e.g. catkin) so that all of ROS uses the same compiler option.

I think the flag is probably most cleanly added from CMake, as we're doing here:

https://github.com/RoboStack/ros-noetic/blob/cbd8a2011de90923a154bca7533d0e1578269065/patch/ros-noetic-geometric-shapes.win.patch#L23-L27

pvanlaar:
Thank you for the assistance @rhaschke. Nice solution to do this with the help of clang-tidy.

PR #2765, opened 2021-07-10
CI: Switch to Noetic builds by default

rhaschke:
So far, we only used ROS Melodic builds for CI (based on the `master-ci` docker container).
This PR attempts to switch to `noetic-ci` by default and using only a single `master-ci` build for Melodic.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,13 +17,13 @@ jobs:
       fail-fast: false
       matrix:
         env:
-          - IMAGE: master-ci
+          - IMAGE: noetic-ci
             CCOV: true
-          - IMAGE: master-ci-shadow-fixed
+          - IMAGE: noetic-ci-shadow-fixed
             IKFAST_TEST: true
             CATKIN_LINT: true
             CLANG_TIDY: true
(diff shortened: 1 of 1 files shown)


rhaschke:
I have no idea why the builds fail on the noetic-ci docker container. The errors are rather strange: looks like plugins are not found!
Maybe, we again have an ABI conflict?

rhaschke:
Looks like colcon is fundamentally broken when running catkin tests: no plugins are found!
Enforcing `catkin tools` as the builder for CI resolves most issues, but there is:
```
Errors     << moveit_tutorials:make /home/runner/work/moveit/moveit/.work/downstream_ws/logs/moveit_tutorials/build.make.000.log
/usr/bin/ld: CMakeFiles/visualizing_collisions_tutorial.dir/src/visualizing_collisions_tutorial.cpp.o: undefined reference to symbol 'ccd_vec3_origin'
/usr/bin/ld: /lib/x86_64-linux-gnu/libccd.so.2: error adding symbols: DSO missing from command line
```
I guess, `libccd` should be pulled in via `libfcl`. As we indeed have different versions of `libfcl` in Melodic and Noetic, this might be a real open issue...

PR #2769, opened 2021-07-15
Fix MSA templates

rhaschke:
Major change is to create a static_transform_publisher for each and every virtual joint type as suggested in https://github.com/ros-planning/moveit_resources/issues/89#issuecomment-879791874.
Additionally, some minor formatting changes.

### DIFF
--- a/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo.launch
+++ b/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo.launch
@@ -55,7 +55,7 @@
 
   <!-- Run Rviz and load the default config to see the state of the move_group node -->
   <include file="$(dirname)/moveit_rviz.launch" if="$(arg use_rviz)">
-    <arg name="rviz_config" value="$(find [GENERATED_PACKAGE_NAME])/launch/moveit.rviz"/>
+    <arg name="rviz_config" value="$(dirname)/moveit.rviz"/>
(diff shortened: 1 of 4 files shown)


felixvd (approved):
LGTM

moveit_setup_assistant/templates/moveit_config_pkg_template/launch/demo.launch
```diff
@@ -55,7 +55,7 @@
 
   <!-- Run Rviz and load the default config to see the state of the move_group node -->
   <include file="$(dirname)/moveit_rviz.launch" if="$(arg use_rviz)">
-    <arg name="rviz_config" value="$(find [GENERATED_PACKAGE_NAME])/launch/moveit.rviz"/>
+    <arg name="rviz_config" value="$(dirname)/moveit.rviz"/>
```
v4hn, line 58: you are right, it actually works due to our file layout - even though the use is buggy and does not set a good example I guess.. .  https://github.com/ros-planning/moveit/pull/2748#issue-681145812

rhaschke: You are right. I ran into this bug recently as well. I will revert this change.

felixvd: TMYK

PR #2775, opened 2021-07-17
MGI::setStartState: Only fetch current state when new state is diff

### LINKED ISSUE #2774: PlanningSceneInterface::setStartState(moveit_msgs::RobotState) crashes if there is no "current" robot state.
### Description

Hello,

I'm using Moveit in a situation without a robot, simulated or otherwise; I'd like to plan paths between hypothetical start-and-end states. I only have the URDF/SRDF model, so the notion of the "current" robot position is undefined.

In the implementation of `PlanningSceneInterface::setStartState(moveit_msgs::RobotState)`, note the line `impl_->getCurrentState(rs);` (https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp#L1611)

This means that, even if the `moveit_msgs::RobotState` fully defines the current robot state, this method call fails with an error like this: `[ERROR] [1626532077.440248038]: Failed to fetch current robot state`.

### Your environment
* ROS Distro: Noetic
* Binary build. installed through APT.

### Steps to reproduce

Simply call the method referred to using a moveit_msgs::RobotState, with every joint setting defined, but without a source of current robot joint settings.

The cause seems pretty obvious from the source code.

### Expected behaviour

The provided state is used as the start state in planning.

### Actual behaviour

The aforementioned error is produced, because the method assumes that there is a source of "current" robot states.

`[ERROR] [1626532077.440248038]: Failed to fetch current robot state`

rhaschke:
Fixes #2774.

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
@@ -1611,7 +1611,10 @@ void MoveGroupInterface::stop()
 void MoveGroupInterface::setStartState(const moveit_msgs::RobotState& start_state)
 {
   moveit::core::RobotStatePtr rs;
-  impl_->getCurrentState(rs);
+  if (start_state.is_diff)
+    impl_->getCurrentState(rs);
+  else
+    rs = std::make_shared<moveit::core::RobotState>(getRobotModel());
   moveit::core::robotStateMsgToRobotState(start_state, *rs);
   setStartState(*rs);
 }


v4hn (approved):

v4hn:
Obvious fix. Thanks for pointing it out @werner291
Your use-case is quite non-standard though and I would not be surprised if you found other similar issues.

werner291:
> Your use-case is quite non-standard though and I would not be surprised if you found other similar issues.

It is? I was informed that MoveIt was the de-facto standard in terms of motion planning and benchmarking a new planner, and that I would probably get some questions at review time if I didn't use it.

v4hn:
> It is? I was informed that MoveIt was the de-facto standard in terms of motion planning and benchmarking a new planner, and that I would probably get some questions at review time (when submitting something to a conference) if I didn't use it.

You would get some from me I'm sure :smile:
It's very useful for users to have an open source MoveIt implementation of your planner available!

What I meant is that the full MoveIt framework (which you can control through the `MoveGroupInterface`) usually connects to a simulated or real robot.
For pure benchmarking, [`moveit_ros_benchmarks`](https://github.com/ros-planning/moveit/tree/master/moveit_ros/benchmarks), [`robowflex`](https://github.com/KavrakiLab/robowflex) and others already provide wrappers so that you do not have to call these user-level control methods.

werner291:
Oooh, shiny! I'll be sure to go take a look. Especially seeing that robowflex is from Kavrakilab.

Gosh, there sure are a *lot* of layers to all of these systems... Of course, I'm speaking with the benefit of hindsight, obviously.

v4hn:
> Especially seeing that robowflex is from Kavrakilab.

That being said, I'll point out that the only maintainer is a PhD student there. He does a great job though. :-)

> Gosh, there sure are a *lot* of layers to all of these systems... Of course, I'm speaking with the benefit of hindsight, obviously.

Tell me about it. You don't know half of it :-)

PR #2781, opened 2021-07-20
Trajectory: Improve docstrings

v4hn:
Addresses https://github.com/ros-planning/moveit/pull/2751/files#r667319953

Please review @simonschmeisser

### DIFF
--- a/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
+++ b/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h
@@ -52,14 +52,20 @@ MOVEIT_CLASS_FORWARD(RobotTrajectory);  // Defines RobotTrajectoryPtr, ConstPtr,
 class RobotTrajectory
 {
 public:
-  /** @brief construct a trajectory */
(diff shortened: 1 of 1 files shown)


simonschmeisser:
So this mainly affects what will be included when "exported" in a RobotTrajectoryMsg right? It seems not to be used in `PlanningScene::isPathValid`?

v4hn:
> So this mainly affects what will be included when "exported" in a RobotTrajectoryMsg right? It seems not to be used in `PlanningScene::isPathValid`?

It's also used at least in the `unwind` methods. And should probably be used in others...

And sadly you are right about isPathValid. :/ But I would say this is actually a shortcoming in the interface. Changing the semantics of the group argument default value `""` is probably a bad idea, but we could at least add another overload to the set that uses the group of the RobotTrajectory. Could you create a pull-request for that? ;)

I intentionally kept the description here broad. Is that ok with you?

simonschmeisser (approved):

simonschmeisser:
Ok, I hoped to get even more insight from more wordy doc-strings but since the "group" is not used too consistently I think this is as good as it gets

PR #2783, opened 2021-07-21
Fix bullet plugin library path name

captain-yoshi:
### Description
Fix wrong plugin library path for the Bullet collision detector.

```shell
Exception while loading Bullet: Could not find library corresponding to plugin Bullet. Make sure the plugin description XML file has the correct name of the library and that the library actually exists.

```

### DIFF
--- a/moveit_core/collision_detector_bullet_description.xml
+++ b/moveit_core/collision_detector_bullet_description.xml
@@ -1,4 +1,4 @@
-<library path="lib/libcollision_detector_bt_plugin">
+<library path="lib/libcollision_detector_bullet_plugin">
   <class name="Bullet" type="collision_detection::CollisionDetectorBtPluginLoader"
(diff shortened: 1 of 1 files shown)


rhaschke (approved):

rhaschke:
Don't we have a unit test using the bullet collision???

captain-yoshi:
Yes there is but, to my knowledge, there are no test using bullet through the `collision_detection::CollisionPlugin`. I think it's the same for FCL.

This [PR from reach](https://github.com/ros-industrial/reach/pull/16/files) added a test for the same issue. It could give ideas on how to create a test for Bullet and FCL through the plugin. Maybe MoveIt has already tests that uses a plugin (Kinematics, planning).

rhaschke:
I suggest adapting the existing tests to load the collision library via the plugin mechanism.

Issue #2777, opened 2021-07-18
Picked collision objects lose their color after placed


corot:
### Description

Picked collision objects lose their color after placed. Also the type field gets emptied. I suppose somewhere in the code we transform the co from [CollisionObject](http://docs.ros.org/en/melodic/api/moveit_msgs/html/msg/CollisionObject.html) to [AttachedCollisionObject](http://docs.ros.org/en/melodic/api/moveit_msgs/html/msg/AttachedCollisionObject.html) w/o copying the color and the type, but I cannot find where.

https://user-images.githubusercontent.com/322610/126053927-124f630b-3f92-4147-b7fd-4ef8c5895c71.mp4

### Your environment
* ROS Distro: Melodic|
* OS Version: e.g. Ubuntu 18.04
* Source build with latest main

### Steps to reproduce
- Add collision objects with color to planning scene
- Pick and place anyone
- Upon picked, color changes to the default attached CO color (purple)
- Upon placed, color changes to the default CO color (green)

### Expected behaviour
Color remains after place

### Actual behaviour
Color reverts to default

corot:
For the color, we lose it on [this line](https://github.com/ros-planning/moveit/blob/master/moveit_core/planning_scene/src/planning_scene.cpp#L1810). Commenting it makes the color reappear after placing (though the attached object becomes White, not sure why). Kind of make sense... we remove the obj from the scene, as it becomes attached. But cannot see why we need to remove its color; if the object is not restored, it won't be used, but doesn't bother neither. @v4hn , u wrote these lines. What do u think?

https://user-images.githubusercontent.com/322610/126186196-fdc0c347-bfa6-4250-8ca1-e684a9c60132.mp4

v4hn:
> But cannot see why we need to remove its color; if the object is not restored, it won't be used, but doesn't bother neither.

It would block memory, which is bad design and can be a problem when adding/removing objects over longer time frames.
Removing an object is semantically different from attaching it though and the meta information should only be removed in the former case.
Could you provide a patch and/or test for this issue?

felixvd:
For the record, I've always thought that a separate `DETACH` keyword would make a lot more sense than using REMOVE for detaching an object from the robot.

v4hn:
> For the record, I've always thought that a separate `DETACH` keyword would make a lot more sense than using REMOVE for detaching an object from the robot.

[...]

PR #2785, opened 2021-07-26
Don't discard object color and type when attached, so we can reuse them once detached

corot:
### Description

Tentative fix for #2777

### Checklist

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1807,8 +1807,12 @@ bool PlanningScene::processCollisionObjectRemove(const moveit_msgs::CollisionObj
   else
   {
     world_->removeObject(object.id);
-    removeObjectColor(object.id);
-    removeObjectType(object.id);
+    if (!getCurrentState().hasAttachedBody(object.id))
+    {
+      // Don't discard object color and type when an attached object is just being detached
+      removeObjectColor(object.id);
+      removeObjectType(object.id);
+    }
   }
   return true;
 }


moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1807,8 +1811,12 @@ bool PlanningScene::processCollisionObjectRemove(const moveit_msgs::CollisionObj
   else
   {
     world_->removeObject(object.id);
-    removeObjectColor(object.id);
-    removeObjectType(object.id);
+    if (!robot_state_->hasAttachedBody(object.id))
+    {
+      // Don't discard object color and type when attached, so we can reuse them once detached
```
felixvd, line 1816:
```suggestion
      // Don't discard object color and type when an attached is just being detached
```

felixvd: For the record, I don't mind the lookup in this function at all.

felixvd (commented):
Looks like this approach might work without creating new dedicated functions for detaching/removing.

At the moment, the collision world stores the color and type of the attached object, although it would make more sense for that to be retrieved from the `AttachedBody` when it is detached. But it's not a big deal and the split between robot and world is convoluted no matter what.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -448,8 +448,12 @@ void PlanningScene::pushDiffs(const PlanningScenePtr& scene)
       if (it.second == collision_detection::World::DESTROY)
       {
         scene->world_->removeObject(it.first);
-        scene->removeObjectColor(it.first);
-        scene->removeObjectType(it.first);
+        if (!robot_state_->hasAttachedBody(it.first))
```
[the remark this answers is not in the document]
corot: Yep, we use `DESTROY`.  But looks like the other change is enough to restore the color of detached objects.
I can undo this one, but not sure what will be the implications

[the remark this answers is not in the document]
felixvd: The other change doesn't "restore" color, it just doesn't delete it from the world upon detaching. The color was left in the world during attaching.

I would prefer if the color was removed from the world during attaching, and added to the world using the info in the `AttachedBody` during detaching. But if you don't have the cycles to add that, this will work, too.

corot (commented):
Ok,  I remove the change on `pushDiffs`, so we only check for attached on `processCollisionObjectRemove`
Hope this is now good for merge;  I change the PR to ready-for-review

moveit_core/planning_scene/src/planning_scene.cpp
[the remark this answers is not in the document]
felixvd: I'd just add a comment to the effect of `// Workaround: Remember color in world while object is attached to robot, instead of recovering the info from AttachedBody`

felixvd:
I would be happy to merge this if the CI failures are taken care of

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1807,8 +1807,12 @@ bool PlanningScene::processCollisionObjectRemove(const moveit_msgs::CollisionObj
...
   {
     world_->removeObject(object.id);
-    removeObjectColor(object.id);
-    removeObjectType(object.id);
+    if (!robot_state_->hasAttachedBody(object.id))
+    {
+      // Don't discard object color and type when an attached object is just being detached
+      removeObjectColor(object.id);
+      removeObjectType(object.id);
+    }
```
v4hn, line 1815: I don't understand how the patch solves any problem.
`processCollisionObjectRemove` is only triggered for incoming `CollisionObject` messages, but from my perspective the correct way to attach a `CollisionObject` is by sending a single `AttachedCollisionObject` message. [This code](https://github.com/ros-planning/moveit/blob/master/moveit_core/planning_scene/src/planning_scene.cpp#L1559-L1570) removes the object from the world then but does not touch the color/type data - and that's what you want.

So I would argue the actual bug in your use-case is that you send a `REMOVE` message at all.
Still in these lines there is an associated bug, because color and type are removed even when `world_->removeObject(object.id) == false` (which is the case if the object already got attached).

v4hn (changes requested):
please also rebase to fix (part of) the failing CI tests.

Not mandatory to merge this, but a test for the behavior would be nice. :santa:

corot:
> please also rebase to fix (part of) the failing CI tests.

Rebased, but tests are failing,,, I'm rather puzzled, as I cannot see the connection with the changes on this PR!
```
[Testcase: testmove_group_interface_cpp_test] ... ERROR!
ERROR: max time [120.0s] allotted for test [move_group_interface_cpp_test] of type [moveit_ros_planning_interface/move_group_interface_cpp_test]
    testMethod()
  File "/opt/ros/melodic/lib/python2.7/dist-packages/rostest/runner.py", line 148, in fn
    self.test_parent.run_test(test)
  File "/opt/ros/melodic/lib/python2.7/dist-packages/rostest/rostest_parent.py", line 132, in run_test
    return self.runner.run_test(test)
  File "/opt/ros/melodic/lib/python2.7/dist-packages/roslaunch/launch.py", line 689, in run_test
    (test.time_limit, test.test_name, test.package, test.type))
--------------------------------------------------------------------------------

[ROSTEST]-----------------------------------------------------------------------


SUMMARY
 * RESULT: FAIL
 * TESTS: 0
 * ERRORS: 1
 * FAILURES: 0

```

v4hn:
somehow rostest failed for multiple independent tests. i restarted CI because I suspect infrastructure issues.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1807,8 +1807,12 @@ bool PlanningScene::processCollisionObjectRemove(const moveit_msgs::CollisionObj
   else
   {
     world_->removeObject(object.id);
-    removeObjectColor(object.id);
-    removeObjectType(object.id);
+    if (!robot_state_->hasAttachedBody(object.id))
```
v4hn, line 1810: This should fix the segfault in CI. Not every scene has a robot_state_ itself.

```suggestion
    if (!getCurrentState()->hasAttachedBody(object.id))
```

My other point remains though, I don't think this this patch makes sense.

moveit_core/planning_scene/src/planning_scene.cpp
corot: yes, right, u attach an obj by sending a `AttachedCollisionObject` with ADD or APPEND action*, but when that happens, we remove the attached `CollisionObject` from the planning scene [here](https://github.com/ros-planning/moveit/blob/f1f23180c41f3b4f9f2194b8f8d6bb407b668b82/moveit_core/planning_scene/src/planning_scene.cpp#L1559). This effectively removes the color (and type) information about the object. Once detached, we add back to the scene [here](https://github.com/ros-planning/moveit/blob/master/moveit_core/planning_scene/src/planning_scene.cpp#L1652), but by then we don't have color or type anymore.
So to my understanding the patch is more or less correct, if not particularly tidy. I can think on better solutions, but all involving much bigger changes. e.g. add the color info to the CollisionObject msg, or handle attach/detach as a different operation than add/remove.

_* offtopic:  what's the difference between APPEND and ADD?_

v4hn: > we remove the attached CollisionObject from the planning scene here

the line you refer to removes the object from the *world inside the planning scene*
afaics the lines in your patch are not called from there.

corot: mmm...  :thinking:    I think they are... indirectly;  w/o going deep into the code, `world_->removeObject` notifies observers that we want to `DESTROY` the obj, and the observers will pub a `CollisionObject` with `REMOVE` action.
Otherwise I misinterpreted something

corot: > APPEND can add shapes to a collision object. ADD ads new objects

Thanks for the clarification!

moveit_core/planning_scene/src/planning_scene.cpp
corot: right!!!   thanks!

v4hn:
Superseded by https://github.com/ros-planning/moveit/pull/2814 .

PR #2791, opened 2021-07-29
Fix bug where setting planning scene publish frequency is ignored

nbbrooks:
### Description

This fixes a bug where PlanningSceneMonitor::setPlanningScenePublishingFrequency does not update the rate used by the internal loop logic. This also increases the default frequency to 30 FPS, which I think is a better default value when you want to visualize movement of the robot with attached collision objects.

### Checklist

### DIFF
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -52,7 +52,7 @@
 
 namespace
 {
-constexpr double DEFAULT_PLANNING_SCENE_PUBLISH_FREQUENCY = 2.0; // Hz
+constexpr double DEFAULT_PLANNING_SCENE_PUBLISH_FREQUENCY = 30.0; // Hz
 }  // namespace
 
 namespace planning_scene_monitor
--- a/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
+++ b/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
@@ -50,6 +50,11 @@
 
 #include <memory>
 
+namespace
+{
+constexpr double DEFAULT_PLANNING_SCENE_PUBLISH_FREQUENCY = 2.0; // Hz
+}  // namespace
+
 namespace planning_scene_monitor
 {
 using namespace moveit_ros_planning;
@@ -137,7 +142,7 @@ PlanningSceneMonitor::PlanningSceneMonitor(const robot_model_loader::RobotModelL
 PlanningSceneMonitor::PlanningSceneMonitor(const planning_scene::PlanningScenePtr& scene,
                                            const robot_model_loader::RobotModelLoaderPtr& rm_loader,
                                            const std::shared_ptr<tf2_ros::Buffer>& tf_buffer, const std::string& name)
-  : monitor_name_(name), nh_("~"), tf_buffer_(tf_buffer), rm_loader_(rm_loader)
+  : monitor_name_(name), nh_("~"), tf_buffer_(tf_buffer), rm_loader_(rm_loader), publish_planning_scene_frequency_(DEFAULT_PLANNING_SCENE_PUBLISH_FREQUENCY), publish_planning_scene_frequency_updated_(false)
 {
   root_nh_.setCallbackQueue(&queue_);
   nh_.setCallbackQueue(&queue_);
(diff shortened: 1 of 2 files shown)


moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -348,12 +352,12 @@ void PlanningSceneMonitor::scenePublishingThread()
     ROS_DEBUG_NAMED(LOGNAME, "Published the full planning scene: '%s'", msg.name.c_str());
   }
 
+  ros::Rate rate(publish_planning_scene_frequency_);
   do
   {
     moveit_msgs::PlanningScene msg;
     bool publish_msg = false;
     bool is_full = false;
-    ros::Rate rate(publish_planning_scene_frequency_);
```
simonschmeisser, line 356: isn't this recreated every iteration and thus reacts to rate changes?

nbbrooks: Good point, I think my initial bug was I was not adjust the PSM rate that my widget was visualizing. I think there are still some useful improvements in this PR though (using atomic). I will update the description though.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -413,6 +417,11 @@ void PlanningSceneMonitor::scenePublishingThread()
       planning_scene_publisher_.publish(msg);
       if (is_full)
         ROS_DEBUG_NAMED(LOGNAME, "Published full planning scene: '%s'", msg.name.c_str());
+      if(publish_planning_scene_frequency_updated_) {
+        ROS_DEBUG_STREAM_NAMED(LOGNAME, "Updating planning scene publish frequency to " << publish_planning_scene_frequency_);
+        rate = ros::Rate(publish_planning_scene_frequency_);
+        publish_planning_scene_frequency_updated_ = false;
+      }
       rate.sleep();
```
v4hn, line 425: Please sleep *before* updating the rate to avoid a single longer cycle on change.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -348,12 +352,12 @@ void PlanningSceneMonitor::scenePublishingThread()
     ROS_DEBUG_NAMED(LOGNAME, "Published the full planning scene: '%s'", msg.name.c_str());
   }
 
+  ros::Rate rate(publish_planning_scene_frequency_);
```
v4hn, line 355: :+1:

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -50,6 +50,11 @@
 
 #include <memory>
 
+namespace
+{
+constexpr double DEFAULT_PLANNING_SCENE_PUBLISH_FREQUENCY = 30.0; // Hz
+}  // namespace
+
```
v4hn, line 57: Hm, I remember we had long debates about this in some other thread before (@felixvd and @simonschmeisser were involved I think?)
What was the consensus here?

30Hz strikes me as hell of a lot when octomaps or many objects might be involved...

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -413,6 +417,11 @@ void PlanningSceneMonitor::scenePublishingThread()
       planning_scene_publisher_.publish(msg);
       if (is_full)
         ROS_DEBUG_NAMED(LOGNAME, "Published full planning scene: '%s'", msg.name.c_str());
+      if(publish_planning_scene_frequency_updated_) {
```
v4hn, line 420: Why would you use two variables here instead of a simple atomic double?
Do you think it's necessary to log the Rate "non-"change if the frequency did not change?

Also, `if(flag){ ...; flag= false; }` breaks the use of an atomic. I would have expected [`compare_exchange_strong`](https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange) or something similar for an atomic.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
simonschmeisser: yes, I proposed that before. It will only create issues if the octomap or objects change frequently as well. Which could be solved with input rate limiting in case of the octomap. I then tried to bargain for a few more FPS but I was told we're not on a bazaar here :-P

Just yesterday I stumbled over a video from IronOx and marveled at the tasty fresh green lettuce when I suddenly saw a PlanningSceneDisplay updating at 4fps and cried a little ;-)

v4hn: I agree that we should increase that value, but it seems @felixvd found some issues with system load in the Motion Planning Rviz Display so it's might be a bit more involved than just stepping up the number.

v4hn:
Please rebase this patch when you address my feedback, to fix the independent failure of the clang-tidy job.

PR #2790, opened 2021-07-28
[fix] RobotState Constructor segfault

tylerjw:
### Description

This is a backport of a bug I fixed in moveit2.

It came from this PR: https://github.com/ros-planning/moveit2/pull/562/files

I am backporting the bug fix separate from that full PR because I believe it should be trivial to merge whereas that whole PR might require more discussion before we backport that pattern from moveit2 to moveit.  That doesn't mean that we can't backport the bugs we fix while applying that pattern even if we don't want to backport the pattern yet.

PR #2792, opened 2021-07-29
fix clang-tidy CI job

v4hn:
https://github.com/ros-planning/moveit/pull/2790 and https://github.com/ros-planning/moveit/pull/2674 currently show that the clang-tidy job broke, likely due to 55886f770b0f9ebe1f2297b92869e8ca97cfc60a .

Recovering the clang-tidy job might fail for a variety of reasons, so I will mark this as WIP.
Anyone, feel free to fix and merge this if you have time.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,6 +44,8 @@ jobs:
       CLANG_TIDY_BASE_REF: "${{ github.base_ref || github.ref }}"
       BEFORE_CLANG_TIDY_CHECKS: (cd $TARGET_REPO_PATH; clang-tidy --list-checks)
       BUILDER: ${{ matrix.env.BUILDER || 'catkin_tools' }}
+      CC: ${{ matrix.env.CLANG_TIDY && 'clang' }}
+      CXX: ${{ matrix.env.CLANG_TIDY && 'clang++' }}
 
     name: "${{ matrix.env.IMAGE }}${{ matrix.env.CATKIN_LINT && ' + catkin_lint' || ''}}${{ matrix.env.CCOV && ' + ccov' || ''}}${{ matrix.env.IKFAST_TEST && ' + ikfast' || ''}}${{ matrix.env.CLANG_TIDY && ' + clang-tidy' || '' }}"
     runs-on: ubuntu-latest
--- a/moveit_core/collision_detection_fcl/CMakeLists.txt
+++ b/moveit_core/collision_detection_fcl/CMakeLists.txt
@@ -5,6 +5,13 @@ add_library(${MOVEIT_LIB_NAME}
   src/collision_env_fcl.cpp
 )
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+  # FIXME(v4hn): clang 10-12 produce broken code for the only std::map::insert call in the method
+  # however, the method calls in here are hotspots in the whole pipeline and disabling optimizations
+  # is no sufficient solution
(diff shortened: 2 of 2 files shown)


v4hn:
does anyone see a connection between this pr and a bullet collision test failing with bad_alloc or has seen this failure before?

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -1573,8 +1573,10 @@ bool RobotState::setFromIK(const JointModelGroup* jmg, const EigenSTL::vector_Is
     std::string error_msg;
     if (!solver->supportsGroup(jmg, &error_msg))
     {
+      // skirt around clang-diagnostic-potentially-evaluated-expression
+      const kinematics::KinematicsBase& solver_ref = *solver;
       ROS_ERROR_NAMED(LOGNAME, "Kinematics solver %s does not support joint group %s.  Error: %s",
-                      typeid(*solver).name(), jmg->getName().c_str(), error_msg.c_str());
+                      typeid(solver_ref).name(), jmg->getName().c_str(), error_msg.c_str());
```
henningkayser, line 1579: Won't this yield the name of the base class instead of the actual solver plugin?

henningkayser (commented):
I don't see anything related to bullet. Is this repeatable? Also, FCL seems to fail for the same reason now.

moveit_core/robot_state/src/robot_state.cpp
v4hn: Why would it? typeid is a runtime method and the passed parameter is still the same.

v4hn:
> I don't see anything related to bullet. Is this repeatable? Also, FCL seems to fail for the same reason now.

> build/moveit_core/test_results/moveit_core/gtest-test_fcl_collision_detection_panda.xml: 18 tests, 0 errors, 2 failures, 0 skipped

Sorry, it is related to FCL indeed. I misread the logs before.
I can't reproduce this issue with my local system and clang 11.0.1, so this needs to be debugged in the CI environment.
@tylerjw and @rhaschke have way more experience in this, sorry :confused:

v4hn:
As a general update:

- The problem relates to clang and is not present with gcc
- The test breaks with clang 10,11,12 in Release builds
- compiling in Debug mode fixes the problem

This is getting really annoying.

v4hn:
I just distilled my insights from late-night debugging [into a `-O0` to a single (though highly crucial) source file](https://github.com/ros-planning/moveit/pull/2792/commits/5ad1228d32b5de36166b44a99d5c2e937118a1e5#diff-1d925a3cba705f6bb50dd725fd91f12cfd696c7cecf5c1758379dd87379049b2R8-R14).
Everyone is welcome to pick it up from here and find the concrete optimization pass that introduces the problem or find a way to work around it in source code.

If CI succeeds now I would still propose to merge it for now with an urgent followup TODO to unblock CI.

v4hn:
This succeeds as soon as https://github.com/ros-planning/moveit_visual_tools/pull/96 and https://github.com/ros-planning/moveit_visual_tools/pull/96 are merged, so I'll remove the WIP label already.

tylerjw:
> This succeeds as soon as [ros-planning/moveit_visual_tools#96](https://github.com/ros-planning/moveit_visual_tools/pull/96) and [ros-planning/moveit_visual_tools#96](https://github.com/ros-planning/moveit_visual_tools/pull/96) are merged, so I'll remove the WIP label already.

Did you mean to link the same PR twice or is there another PR that is also needed?

v4hn:
> Did you mean to link the same PR twice or is there another PR that is also needed?

updated above, sorry for the broken c&p.

tylerjw:
I don't like the fix of changing the optimization level but I don't have time or any better ideas at the moment.  I'll merge this and create an issue for fixing it better.

tylerjw (approved):

PR #2793, opened 2021-07-29
Generate version.h with git branch and commit hash

captain-yoshi:
### Description

This PR stores the git current branch and commit hash into the version header file. This is needed for benchmarking different commits so we can aggregate the results. This was taken from this [blog](http://xit0.org/2013/04/cmake-use-git-branch-and-commit-details-in-project/) who lead to this [git repo](https://github.com/pmirshad/cmake-with-git-metadata).

### Pitfalls
You have to re-cmake to update the hash definition. There is another [solution](https://stackoverflow.com/a/4318642) that will force a re-configure of cmake before a build every time the HEAD commit changes.

### Discussion
There may be better ways to find the `.git` folder.

We could remove the else block and define the git information as an empty string before the execute_process function.

### Result
Generated `version.h` file:
```cpp
#ifndef MOVEIT_VERSION_
#define MOVEIT_VERSION_
 
#define MOVEIT_VERSION "1.1.5-Alpha"
 
#define MOVEIT_VERSION_MAJOR 1
#define MOVEIT_VERSION_MINOR 1
#define MOVEIT_VERSION_PATCH 5
#define MOVEIT_VERSION_EXTRA "Alpha"
 
#define MOVEIT_GIT_BRANCH "git-commit-hash"
#define MOVEIT_GIT_COMMIT_HASH "f733739ec99bfa72f534d2b65c693dc0cfe696a3"
 
#endif
```


simonschmeisser:
With our internal jenkins CI we need to use the following as it otherwise shows the branch as "head"
```
if (DEFINED ENV{BRANCH_NAME})
    set(GIT_BRANCH $ENV{BRANCH_NAME})
else ()
    execute_process(
        COMMAND
            git rev-parse --abbrev-ref HEAD
        WORKING_DIRECTORY
            ${CMAKE_CURRENT_SOURCE_DIR}
        RESULT_VARIABLE
            GIT_BRANCH_RESULT
        OUTPUT_VARIABLE
            GIT_BRANCH
        OUTPUT_STRIP_TRAILING_WHITESPACE
    )
endif ()
```
don't know if that's need/works with github ci

v4hn:
> You have to re-cmake to update the hash definition

That's a pretty harsh requirement and would lead to incorrect results for user workspaces (including mine) the whole time!
If somehow possible I would like to avoid that pitfall.

> There is another solution that will force a re-configure of cmake before a build every time the HEAD commit changes

What does that mean for catkin workspaces? it's also a hassle if catkin_make/catkin build has to configure the whole workspace each time. If that should be the case (I would expect that), maybe there's a way to generate the hash entry at build-time instead of using `configure_file`?

Also, we probably want to add the hash as output to [moveit_version](https://github.com/ros-planning/moveit/blob/master/moveit_core/version/version.cpp#L42)

rhaschke:
I fixed the open issues:
- using a custom command (and target) to regenerate version.h on every build
- fetching any branch or tag name to fix https://github.com/ros-planning/moveit/pull/2793#issuecomment-889409855

rhaschke (approved):

Issue #2717, opened 2021-06-11
GSoC: Creation of a Benchmark Suite for Evaluating MoveIt Performance


captain-yoshi:
**mentors: @felixvd @mamoll @v4hn**

This issue will bundle all relevant information about the GSoC project. A Discord channel is also available to discuss on this project. I will try to add pertinent information from Discord to this issue.

## Motivation and goals
This GSoC project goal is to benchmark complete tests throughout the entire MoveIt pipeline and eventually the MoveIt Task Constructor (MTC). This will be done through a benchmark suite. Unit tests will also be benchmarked.
- Compare the impact of changes we consider integrating into MoveIt.
- Find regressions between MoveIt versions.
- Isolated tests or benchmarks don't necessarily reflect the entire MoveIt pipeline.
- Some algorithms do well in a particular environment but fall short in other contexts.

#### Phase 1: Exploration and benchmarks (June)
- Create multiple tests with a few scenes (empty scene, solid primitives, meshes, cluttered) with some joint/pose goals, path constraints, FCL and Bullet, planners (OMPL, CHOMP, STOMP) using the `moveit_ros_benchmarks` package.
- Pick and place benchmark using MTC.
- Analyze tests and decide whether we want to use only self contained tests or use and expand the moveit_ros_benchmarks when benchmarking planners throughout the entire MoveIt stack.
- Add an Amazon Pick and Place Challenge (APC) scenario and do benchmarks on it.

#### Phase 2: Creation of the benchmark suite
- Implement the benchmark suite to run tests.
- Aggregate and plot results through Python.
- Add more tests and scenes.

## Documentation
- [Benchmark Tutorial](https://ros-planning.github.io/moveit_tutorials/doc/benchmarking/benchmarking_tutorial.html)
- [Benchmark Visualizer](https://planners-benchmarking.readthedocs.io/en/latest/user_guide/5_visualizing_the_results.html)
- [Planner Arena](http://plannerarena.org/)
## Links
- [Proposal](https://docs.google.com/document/d/10DHBipL6kuHAcFy03jLOLiPuMkW4cKVAcSG4gw6ixBU/edit?usp=sharing)
- GSoC log
- [Discord](https://discord.gg/KzEf7Ysy) #gsoc2021

captain-yoshi:
### Meeting notes 09/06/2021

#### Previous week summary
- Reading + tutorials.
[...]

PR #2794, opened 2021-07-29
Draft: Support timed locking in PlanningSceneMonitor

henningkayser:
This replaces the shared_mutex of the PSM with a shared_timed_mutex
that allows specifying a timeout for attempting to lock the mutex
for read and write using the LockedPlanningSceneRO and
LockedPlanningSceneRW classes. In case of a failed attempt, a
TryLockTimeoutException is thrown so that the calling thread is
unblocked in case of concurrency issues.
Additionally, a new scoped lock wrapper class TimedLock allows
tracking lock sources (=caller names), start time, waiting time and
the duration that the lock persists. An optional DiagnosticStatus
publisher can be enabled for debugging these lock events at runtime.

The scope-less locking functions (lockSceneRead(), etc...) are deprecated
since they are error-prone and don't work well with the diagnostics feature.

My motivation for this was a lengthy debugging session for getting rid of spurious deadlocks and performance issues.
Especially, in scenarios where the planning scene monitor is being used in a time-constrained loop (servo collision checking, online sampling for controller commands) it's critical to understand where time is being spent and what threads are being blocked.
I'm not convinced that the DiagnosticStatus publisher should be merged as is (moveit_profiler would do a similar job), but the timed locking feature seems really useful and safer to me.

Left TODOs:
- [ ] Handle locking of octomap_monitor
- [ ] Keep track of lock owners and print names with warning/error/exception
- [ ] Move boilerplate code into separate header


moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -699,13 +787,23 @@ class LockedPlanningSceneRO
...
 public:
-  LockedPlanningSceneRW(const PlanningSceneMonitorPtr& planning_scene_monitor)
-    : LockedPlanningSceneRO(planning_scene_monitor, false)
+  LockedPlanningSceneRW(const PlanningSceneMonitorPtr& planning_scene_monitor,
+                        std::chrono::milliseconds timeout = std::chrono::milliseconds::zero())
+    : LockedPlanningScene(planning_scene_monitor)
   {
+    if (planning_scene_monitor_)
+    {
+      ulock_ = std::make_shared<UniqueTimedLock>(
```
henningkayser, line 804: Open for debate: The shared pointers here copy the old behavior. I'm thinking of not allowing this since it doesn't make a lot of sense to share write access between different entities. I'd suggest making `LockedPlanningSceneRW` non-copyable. `LockedPlanningSceneRO` should probably also not be using a shared pointer, acquiring another shared lock seems way more reasonable to me.

moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp
```diff
@@ -1062,6 +1075,55 @@ bool PlanningSceneMonitor::getShapeTransformCache(const std::string& target_fram
   return true;
 }
 
+UniqueTimedLock PlanningSceneMonitor::acquireUniqueLock(const std::string& source, std::chrono::milliseconds timeout)
+{
+  // TODO(henningkayser): lock/unlock octomap
```
henningkayser, line 1080: The octomap could be locked here and then unlocked with the destructor. However, this would increase the time the octree is being locked since some callbacks haven't done this before.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
```diff
@@ -717,5 +813,8 @@ class LockedPlanningSceneRW : public LockedPlanningSceneRO
   {
     return planning_scene_monitor_->getPlanningScene();
   }
+
+private:
+  std::shared_ptr<UniqueTimedLock> ulock_;
```
henningkayser, line 818: Open for debate: The shared pointers here copy the old behavior. I'm thinking of not allowing this since it doesn't make a lot of sense to share write access between different entities. I'd suggest making LockedPlanningSceneRW non-copyable. LockedPlanningSceneRO should probably also not be using a shared pointer, acquiring another shared lock seems way more reasonable to me.

moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h
henningkayser: moved to https://github.com/ros-planning/moveit/pull/2794#pullrequestreview-718589042

AndyZe:
Just a high-level comment. I wonder what the overhead of a `shared_timed_mutex` is compared to a regular `shared_mutex`. [Answers here](https://stackoverflow.com/a/56159440) suggest it can be significant. Should benchmark.

henningkayser:
> Just a high-level comment. I wonder what the overhead of a `shared_timed_mutex` is compared to a regular `shared_mutex`. [Answers here](https://stackoverflow.com/a/56159440) suggest it can be significant. Should benchmark.

Good point! I'd be surprised if there was a measurable difference between `shared_timed_mutex` (without timeouts) and `shared_mutex` since they both implement the same reader/writer behavior. The wrapper class `TimedLock` however could actually have performance impacts, we should make sure that this is not the case.

felixvd:
> Just a high-level comment. I wonder what the overhead of a `shared_timed_mutex` is compared to a regular `shared_mutex`. [Answers here](https://stackoverflow.com/a/56159440) suggest it can be significant. Should benchmark.

Do you have an idea how to benchmark this? If yes, we could integrate it in https://github.com/ros-planning/moveit/issues/2717

AndyZe:
@captain-yoshi and @felixvd here's a script that is good for benchmarking repeated, high-frequency reads of the planning scene. (https://github.com/ros-planning/moveit/pull/2698#issuecomment-853091796)

To properly test this PR, some writes to the planning scene should also be included (to mimic normal operation).

AndyZe:
I'm also curious how `std::mutex` would perform, compared to `shared_mutex`.

v4hn:
From your descriptions it looks like there are some things left to do to cleanup this patch.

I only skimmed through the changes for now. My main concern is backward compatibility. Is this patch supposed to be API compatible except for deprecating the non-RAII locking API (to which I definitely agree)?

jliukkonen (commented):
The idea of having more insight into what's going on is really good. However, the diagnostics come with extra cost and I'm not sure having them included should be the default in this case.

PR #2795, opened 2021-07-31
Format includes with clang-format

tylerjw:
### Description

@v4hn this was much easier in moveit than in moveit2... there were only 2 files I found with missing includes.

The gist of this change is that it formats the include lines so we can stop doing it manually.  I don't like manually re-organizing include lines and think clang-format should do this for me.  I hope you like this change as much as I do.

Here is the moveit2 PR this is a port of: https://github.com/ros-planning/moveit2/pull/583


v4hn:
We really need to get https://github.com/ros-planning/moveit/pull/2792 fixed so we can proceed with many other PRs....

tylerjw:
@v4hn I updated this with your suggested change in the PR to moveit2.

rhaschke (approved):
I like this very much! Didn't know you can define include groups.

tylerjw:
> I like this very much! Didn't know you can define include groups.

I learned this by poking around in some google code.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/bullet_utils.h
```diff
@@ -35,14 +35,16 @@
 #pragma once
 
 #include <btBulletCollisionCommon.h>
```
v4hn, line 37: This registers as a false-positive because bullet seems to use includes without prefix. Not much we can do about it I think, or do we want to add exceptions for specific header names?

moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
```diff
@@ -34,13 +34,15 @@
 
 /* Author: Jens Petit */
 
-#include <moveit/collision_detection_bullet/collision_env_bullet.h>
-#include <moveit/collision_detection_bullet/collision_detector_allocator_bullet.h>
-#include <moveit/collision_detection_bullet/bullet_integration/ros_bullet_utils.h>
-#include <moveit/collision_detection_bullet/bullet_integration/contact_checker_common.h>
 #include <boost/bind.hpp>
+
 #include <bullet/btBulletCollisionCommon.h>
```
v4hn, line 39: or, we just update all bullet includes to use the `bullet/` prefix. Afaik the bullet tutorials use includes without prefix though.

moveit_core/collision_distance_field/src/collision_env_distance_field.cpp
```diff
@@ -34,17 +34,20 @@
...
 
-#include <moveit/robot_model/robot_model.h>
-#include <ros/console.h>
+#include <memory>
+#include <utility>
+
+#include <boost/bind.hpp>
+
 #include <ros/assert.h>
+#include <ros/console.h>
```
v4hn, line 43: I would expect to see all `ros/` includes before any other ros package.

v4hn (commented):
This is a pain to review (I managed a quarter before I gave up) and I'm sure we will notice unintuitive includes because of this patch in the future...

tylerjw:
> This is a pain to review (I managed a quarter before I gave up) and I'm sure we will notice unintuitive includes because of this patch in the future...

More generally what I wanted was some sort of automated sorting of these.  I might be weird in that I don't really care if they create some odd orders, just that there is some order.  I want consistency in this and to stop getting review comments asking me to manually sort these.

If anything I think we should adopt a simpler set of rules than these that are essentially:

includes without / (mostly system includes)
includes with / in them
includes that start with the word moveit
includes that use local search "

v4hn:
> If anything I think we should adopt a simpler set of rules than these that are essentially:

then we might as well go for simple alphabetical order altogether. include order should not matter aside from aesthetics. all the things I've pointed out so far are issues I would be inclined to change if I'd encounter them in source code.

> to stop getting review comments asking me to manually sort these. 

If that's your only reason for trying to enforce a syntactic order on (semantic) includes I would say people should stop criticizing it unless there is no obvious order in the patch whatsoever.

If people look at the autosorted includes later and have to adjust the clang-format file to get them in a meaningful order that's just what I would call "verschlimmbessern" in German...

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/bullet_utils.h
rhaschke: What do you mean by false-positive here?

rhaschke:
> then we might as well go for simple alphabetical order altogether. include order should not matter aside from aesthetics. all the things I've pointed out so far are issues I would be inclined to change if I'd encounter them in source code.

I don't get your criticism yet. Aren't you one of the guys who prefer sorting e.g. for package dependencies as well?
This PR just helps to avoid redundant includes and provides a nice grouping of similar includes.
I hate simple alphabetical sorting because it destroys and semantic grouping which was introduced manually. I think this PR is a very nice compromise. 

> If people look at the autosorted includes later and have to adjust the clang-format file to get them in a meaningful order that's just what I would call "verschlimmbessern" in German...

I agree that we should come up with a rule set to push the includes into a meaningful order _before_ merging this PR. 
If you don't like the current rules, please list what you want to be changed. From your specific comments, I cannot yet infer any desired changes.

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/bullet_utils.h
v4hn: this includes bullet headers, but it's grouped with standard headers because there is no prefix used in the include path

v4hn:
that statement was mainly a reply to

>  I don't really care if they create some odd orders, just that there is some order. I want consistency in this and to stop getting review comments asking me to manually sort these.

generally I think the approach is worthwhile, but it's a hassle to look through all the changes and judge them.
I gave multiple rounds of feedback already.
in the last review I mainly noticed the bullet includes we either want to allow exceptions for or modify to use the `bullet/` prefix.

Also I would expect another group for `ros/` includes.

there is probably more to discuss in the rest of the patch...

moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/bullet_utils.h
rhaschke: Got it. However, before the bullet header was not grouped with anything else as well...

moveit_core/robot_model/include/moveit/robot_model/robot_model.h
```diff
@@ -37,21 +37,24 @@
...
 #include <moveit/robot_model/fixed_joint_model.h>
 #include <moveit/robot_model/floating_joint_model.h>
+#include <moveit/robot_model/joint_model_group.h>
 #include <moveit/robot_model/planar_joint_model.h>
-#include <moveit/robot_model/revolute_joint_model.h>
 #include <moveit/robot_model/prismatic_joint_model.h>
-
-#include <Eigen/Geometry>
-#include <iostream>
+#include <moveit/robot_model/revolute_joint_model.h>
```
tylerjw, line 57: One of the challenging things about this PR is that comments above blocks of includes essentially breaks the re-rodering.  Here it re-ordered above and below the `// joint types` line as separate blocks.  I've gone through and removed a bunch of these comments but here is one I forgot.

PR #2803, opened 2021-08-10
Bugfix/fix pilz collision detection

aa-tom:
### Description

The pilz industrial motion planner does not account for updates to the allowed collision matrix when determining collision-free paths due to it creating a new `PlanningScene`, which obtains this information from the SRDF. This is covered in #2676 

This PR passes the most recent `PlanningScene` object obtained via MoveIt down to the relevant collision detection code, allowing changes to the collision matrix to propagate, thus allowing pilz to find paths which would otherwise be deemed unsuitable. 

This work includes:

- Modifying function signatures to accept the new `PlanningScene`
- Modifying docstrings to include the new `PlanningScene`
- Modifying function implementations to pass the new `PlanningScene` down to the next relevant function
- Using the new `PlanningScene` rather than the default one when performing collision detection [here](https://github.com/additiveautomations/aa-pilz/blob/27ec4894b6f3a17bb9a593e2b93e0d01568cad59/moveit_planners/pilz_industrial_motion_planner/src/trajectory_functions.cpp#L576)
- Updating unit tests in line with the new function signatures

The video below demonstrates successful pilz path finding for a non-collision-free path where the particular collisions have been allowed via the collision matrix (one which failed in the example video given in #2676 ):

https://user-images.githubusercontent.com/86778266/124609016-280efa00-de67-11eb-8b5e-3f097447940b.mp4

### Checklist
- [x] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/) (Not relevant)
- [x] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes (Not relevant)
- [x] Include a screenshot if changing a GUI (Not relevant)


aa-tom:
Note: Unit tests need updating as function signatures have changed

aa-tom:
Hey @jschleicher. I believe the clang-format failures are fixed now. Could you approve the workflows when you have time?

v4hn:
Thank you for this effort! :+1: 

> Some functions now call functions accepting the PlanningScene argument, but do not have access to one themselves. In these instances, the code now builds a new PlanningScene using the robot state to pass down to the function in question here and here (I'm unsure whether this is the best approach, or if we should pass the PlanningScene from MoveIt down to these functions too. Advice here would be much appreciated)

[`testutils::generateTrajFromBlendTestData`](https://github.com/additiveautomations/aa-pilz/blob/27ec4894b6f3a17bb9a593e2b93e0d01568cad59/moveit_planners/pilz_industrial_motion_planner/test/test_utils.cpp#L1044) is currently not used in the code base at all from what I can see. Thus I would leave it untouched. If someone needs it for something involving a custom planning scene, we can always change it later.

[The `blend` method](https://github.com/additiveautomations/aa-pilz/blob/27ec4894b6f3a17bb9a593e2b93e0d01568cad59/moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp#L91) on the other hand seems critical to me because when blending between multiple trajectories we need to ensure the new trajectory is valid w.r.t. the current planning scene. To this end I would propose to add a `PlanningSceneConstPtr` to the [`TrajectoryBlendRequest`](https://github.com/additiveautomations/aa-pilz/blob/27ec4894b6f3a17bb9a593e2b93e0d01568cad59/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_blend_request.h#L43) and only create a new scene if this is a `nullptr`. Also all places where `blend` is called will need checking on whether you can/should pass in a scene there.

aa-tom:
@v4hn Glad to help, and thanks for the suggestions! That part of the description is actually outdated now as I was lazy and copied it from the old PR :sweat_smile: We're passing the PlanningScene down to [blend and all instances where it's called now](https://github.com/ros-planning/moveit/pull/2803/commits/fc6b6a70a2b96235ad5811cc19f08dba26954f2d). Props to @cambel for making the change in my fork first :smile:

v4hn:
> To this end I would propose to add a PlanningSceneConstPtr to the TrajectoryBlendRequest and only create a new scene if this is a nullptr.

I would still prefer that approach over adding another parameter to the methods. without it the TrajectoryBlendRequest does not actually specify the full request. That looks like bad design to me...

jschleicher (approved):
Thank you very much for fixing the planning scene! That looks great. I've updated the UML diagram.
The only nitpick: An additional test case failing without your fix and passing now would be great!

jschleicher:
@v4hn I don't get your point about the [blend](https://github.com/ros-planning/moveit/blob/6e746be47bf177daa0938b048d64b8c6e3d71dbc/moveit_planners/pilz_industrial_motion_planner/src/trajectory_blender_transition_window.cpp#L92) function. Do you still see an issue there? Could you explain further?

v4hn (approved):
I believe there is no technical issue here anymore and the PR seems to solve the basic problem throughout the API. I raised a software design point. As said before [the `blend` method](https://github.com/ros-planning/moveit/pull/2803/files#diff-9024d90adda63f780a7b5bc0014c0d20e5cdcf4c03f5a9b11bd9421b80163239R43-R47) originally had a `Request`/`Response` API, but this patch adds a third argument for the `PlanningSceneConstPtr`. The pointer could be part of the `Request` field though to keep the whole blend request in one datastructure (as is usually the goal when you design a `Request`/`Response` API).
If you are ok with this design flaw in the current patch, feel free to merge right awaty.
Otherwise someone should modify the patch.
I'll approve to leave the choice to you.

rhaschke:
Thanks for your work and your patience with this long-overdue PR.

PR #2808, opened 2021-08-12
[WIP] Issue 2801

### LINKED ISSUE #2801: JointConstraintSampler "tolerance" value widens the sampling range, leading to not-quite-unit quaternions in floating joints
I have a robot that includes a floating joint, and I'd like to plan from one exact configuration to another exact configuration.

However, the goal samplers appear to produce quaternions that are not quite unit length; they'll instead produce values like `6.97303e-17 -3.27952e-17 -7.03514e-17 1`. I've seen larger values too, such as in the `e-16` range (that's enough to break approximate equality comparisons).

This breaks some code further down the line, see this related issue: https://github.com/KavrakiLab/robowflex/issues/239

I believe that this has to do with these lines: https://github.com/ros-planning/moveit/blob/master/moveit_core/constraint_samplers/src/default_constraint_samplers.cpp#L89

How this feeds into the issue: When sampling joint constraints, MoveIt effectively treats every degree of freedom as a separate joint, and adds in a tiny extra "tolerance" value, on the order of `e-16`. This tolerance is actually added to the bounds when sampling.

Hence, if a joint with a range of `[0,0]` (such as a component of a unit quaternion that must be 0), this range effectively become `[-tolerance,tolerance]`.

For example, a JointConstraintSampler that's supposed to produce exactly the quaternion `(0,0,0,1)` will instead produce something like `(6.97303e-17, -3.27952e-17, -7.03514e-17, 1)`, which is does not have unit norm, and is therefore an invalid rotation.

I understand wanting the joint tolerance when looking for goal satisfaction due to rounding errors, but why is it used in the sampling range?

werner291:
### Description

This is a WIP pull request that aims to eventually fix #2801; please do not merge yet.

Specifically, this commit changes the implementation of `moveit_msgs::Constraints constructGoalConstraints(const moveit::core::RobotState& state, const moveit::core::JointModelGroup* jmg, double tolerance_below, double tolerance_above)` which originally assumed that every joint could simply be represented with a JointConstraint, which breaks floating joints, leading to https://github.com/KavrakiLab/robowflex/issues/239

### Checklist

### DIFF
--- a/moveit_core/kinematic_constraints/src/utils.cpp
+++ b/moveit_core/kinematic_constraints/src/utils.cpp
@@ -130,6 +130,61 @@ std::size_t countIndividualConstraints(const moveit_msgs::Constraints& constr)
 moveit_msgs::Constraints constructGoalConstraints(const moveit::core::RobotState& state,
                                                   const moveit::core::JointModelGroup* jmg, double tolerance)
 {
+  moveit_msgs::Constraints goal;
+
+  for (const JointModel* jm : jmg->getJointModels()) {
+
+    std::vector<double> vals;
+    state.copyJointGroupPositions(jmg, vals);
+    goal.joint_constraints.resize(vals.size());
+
+    switch (jm->getType()) {
+
+      case JointModel::UNKNOWN:
+        ROS_ERROR("Unable to construct goal constraints for unknown joint type.");
+        break;
+      case JointModel::REVOLUTE:
+      case JointModel::PRISMATIC:
+        case JointModel::PLANAR:
+      {
+
+        for (std::size_t i = 0; i < vals.size(); ++i) {
+          moveit_msgs::JointConstraint jc;
+          jc.joint_name = jmg->getVariableNames()[i];
+          jc.position = vals[i];
+          jc.tolerance_above = tolerance;
+          jc.tolerance_below = tolerance;
+          jc.weight = 1.0;
+          goal.joint_constraints.push_back(jc);
+        }
+      }
+        break;
+      case JointModel::FLOATING:
+      {
+        auto fjm = static_cast<const FloatingJointModel*>(jm);
+        geometry_msgs::PoseStamped floating_joint_pose;
+        floating_joint_pose.header.frame_id = ""; // TODO What's this?
+        floating_joint_pose.pose.position.x = vals[0];
+        floating_joint_pose.pose.position.y = vals[1];
+        floating_joint_pose.pose.position.z = vals[2];
+        floating_joint_pose.pose.orientation.x = vals[3];
+        floating_joint_pose.pose.orientation.y = vals[4];
+        floating_joint_pose.pose.orientation.z = vals[5];
+        floating_joint_pose.pose.orientation.w = vals[6];
+
+        auto constraints = constructGoalConstraints(fjm->getName(), floating_joint_pose, tolerance, tolerance);
+        goal.position_constraints.push_back(constraints.position_constraints[0]);
+        goal.orientation_constraints.push_back(constraints.orientation_constraints[0]);
+
+      }
+        break;
+
+      case JointModel::FIXED:
+        // No point in constraining a fixed joint.
+        break;
+    }
+  }
+
   return constructGoalConstraints(state, jmg, tolerance, tolerance);
 }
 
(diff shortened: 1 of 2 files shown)


werner291:
I expect that this is going to break some things that relied on the outcome of this method being only `JointConstraint`s.

moveit_core/kinematic_constraints/src/utils.cpp
```diff
@@ -130,6 +130,61 @@ std::size_t countIndividualConstraints(const moveit_msgs::Constraints& constr)
...
+    goal.joint_constraints.resize(vals.size());
+
+    switch (jm->getType()) {
+
+      case JointModel::UNKNOWN:
+        ROS_ERROR("Unable to construct goal constraints for unknown joint type.");
+        break;
+      case JointModel::REVOLUTE:
+      case JointModel::PRISMATIC:
+        case JointModel::PLANAR:
```
werner291: This isn't going to work with the cyclical nature of angle values, is it?

moveit_core/kinematic_constraints/include/moveit/kinematic_constraints/utils.h
```diff
@@ -88,8 +88,10 @@ moveit_msgs::Constraints constructGoalConstraints(const moveit::core::RobotState
 
 /**
  * \brief Generates a constraint message intended to be used as a goal
- * constraint for a joint group.  The full constraint will contain a
- * vector of type \ref JointConstraint, one for each DOF in the group.
+ * constraint for a joint group. The full constraint will contain a
+ * vector of type \ref JointConstraint, one for each DOF in the group,
+ * with one exception: floating joint settings will be represented
+ * by a \ref PositionConstraint and an \ref OrientationConstraint.
```
v4hn, line 94: what about planar joints? (hopefully the same as floating joints)

MIGRATION.md
```diff
@@ -1,4 +1,4 @@
-# Migration Notes
+#Migration Notes
```
v4hn, line 1: ?

MIGRATION.md
```diff
@@ -22,6 +22,7 @@ API changes in MoveIt releases
 - The joint states of `passive` joints must be published in ROS and the CurrentStateMonitor will now wait for them as well. Their semantics dictate that they cannot be actively controlled, but they must be known to use the full robot state in collision checks. (https://github.com/ros-planning/moveit/pull/2663)
 - Removed deprecated header `moveit/macros/deprecation.h`. Use `[[deprecated]]` instead.
 - All uses of `MOVEIT_CLASS_FORWARD` et. al. must now be followed by a semicolon for consistency (and to get -pedantic builds to pass for the codebase).
+- The `constructGoalConstraints` method now uses `PositionConstraint` and `OritentationConstraint` to represent floating joint. Note: this makes for a spherical tolerance region by default, instead of the original rectangular one.
```
v4hn, line 25: `OritentationConstraint` -> `OrientationConstraint`

moveit_ros/planning/rdf_loader/src/rdf_loader.cpp
```diff
@@ -106,13 +106,13 @@ rdf_loader::RDFLoader::RDFLoader(const std::string& urdf_string, const std::stri
   auto umodel = std::make_unique<urdf::Model>();
   if (umodel->initString(urdf_string))
   {
-    srdf_ = std::make_shared<srdf::Model>();
-    if (!srdf_->initString(*urdf_, srdf_string))
+    auto smodel = std::make_shared<srdf::Model>();
+    if (!smodel->initString(*umodel, srdf_string))
```
v4hn, line 110: please rebase to include this fix. It was merged in `master` a while ago.

v4hn (commented):
Generally I like the patch a lot! Could you clean this up please?

moveit_core/kinematic_constraints/include/moveit/kinematic_constraints/utils.h
werner291: Added a comment. Yes, planar joints are just floating joints with an extra constraint.

werner291:
Ok... This should to it. Had some difficulties getting this PR quite aligned with `master`, but I think this is correct?

v4hn (dismissed):
Just found this again. Looks good to go! Thanks for digging through these internals!

PR #2813, opened 2021-08-13
more fixes for the clang-tidy job

v4hn:
These two issues appeared in https://github.com/ros-planning/moveit/pull/2795 .
As clang-tidy only inspects files that changed, they did not show up before.

@tylerjw


tylerjw (approved):
good find, using type trates is_same here is really nice.

PR #2814, opened 2021-08-15
Preserve metadata when detaching objects

v4hn:
@corot [proposed to fix object color handling](https://github.com/ros-planning/moveit/pull/2785), but the handling itself does not seem to be broken the way they expect.

Instead, when trying to remove an attached object from the world, color/type was removed even when the object was not actually removed from the world (attached objects are not in the world, so their meta-data should stay around).

Additionally here is a test to verify handling of color and type works as expected when objects get attached/detached.

@corot The bugfix should help you find the culprit in your use case. Please have a look at the test as well.

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1806,7 +1806,9 @@ bool PlanningScene::processCollisionObjectRemove(const moveit_msgs::CollisionObj
   }
   else
   {
-    world_->removeObject(object.id);
+    if (!world_->removeObject(object.id))
+      return false;
+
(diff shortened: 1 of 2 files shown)


corot (approved):
Yup, much better than my PR, and equally preserves the color
Thanks!

felixvd:
Right. When we attach the object, we remove it from the world without removing the metadata. This avoids deleting the metadata when detaching the object with the REMOVE keyword.

I still think it would be much easier to understand the code with a DETACH keyword, and/or if the metadata was copied from the `AttachedBody` to the world, but there are bigger fish to fry.

felixvd (approved):
LGTM!

PR #2816, opened 2021-08-16
PS: add new tests for object pose handling

v4hn:
with the transition to the new CollisionObject::pose field.

One test currently fails.
Could you please address this together with the other issues I raised in
https://github.com/ros-planning/moveit/pull/2037#pullrequestreview-633565661 ?


felixvd:
I already see some tests failing, but maybe you could have a look anyway. I think these commits might cover the gist of it.

v4hn:
CI fails with timeouts, so some test runs into a deadlock with your change.

ANogin:
> If the object consists of a single shape, the object pose should be set to that shape's pose, and the primitive pose to identity. Otherwise, there is a regression for scenes that were built in a scene-graph-like manner by referring to previously placed objects.
> 
> It also looks nicer when you visualize the object pose in TF.

I came here because this change is causing a huge headache for me with MoveIt2. In my use case, the pose is not "empty", it is correctly set to identity, as the URDF was intentionally structured to have the origin of the robot link and attached object coincide, to make the whole thing simpler to set up. The issue is than two-fold:
- This changes the location of the attached object origin to some "random" place (the mesh origin - the URDF is created by exporting from Fusion using https://github.com/jaredgonzales/fusion360descriptor/ and Fusion has all meshes originate at global origin of the design, regardless of the local origin of the specific bodies). 
- This fails to change the pose of the subframes, so the subframe locations end up equally wrong.

This means that using the attached object as a frame id or tip name in IK/motion planning results in things failing in completely unpredictable and seemingly incomprehensible ways (I just happened to have the scene returned by `move_group`'s `get_planning_scene` logged and happened to see that the mesh origin was 0, which clued me to where to look in source code, otherwise I would have likely spent hours trying to understand why things work correctly for some attached objects, and completely bogus for others).

I see the following options:
1. If swapping, update the subframe poses so that they are unchanged.
2. Do not swap when subframes are present.
3. Never swap, and fix whatever incorrectly relies on the swapping behavior.
4. Add an extra boolean field `yes_this_is_the_real_pose_even_if_its_an_identity_and_do_not_mess_with_it` to the message type (obviously silly, but IMHO not more so than the current implementation).

IMHO option 3 is the right way, but I do not have bandwidth or understanding to implement it. I plan to submit a PR for option 2 (as this is the simplest fix for my use case - although I would expect this cause more people to bang their heads against the wall for weird failures in MoveIt2...)

Issue #2804, opened 2021-08-10
moveit_ros_planning fails to build with DCMAKE_BUILD_TYPE=Debug


werner291:
### Description

When building a ROS workspace with `catkin build --cmake-args -DCMAKE_BUILD_TYPE=Debug`, the build fails with a linker error on `moveit_ros_planning`.

I was hoping to debug MoveIt more easily, since I'll probably be making some modifications to it for a project I'm working on.

### Your environment
* ROS Distro: Noetic
* OS Version: Ubuntu 20.04 LTS

Fresh catkin workspace, with `moveit`, `moveit_visual_tools` and `moveit_resources` cloned, all from default branch (master, probably).

### Steps to reproduce

* Create a directory containing an src subdir
* Clone the three aforementioned repositories into it.
* Run `catkin build --cmake-args -DCMAKE_BUILD_TYPE=Debug`

### Expected behaviour

Buildl compiletes successfully, resulting in MoveIt with debug symbols.

### Actual behaviour

`moveit_ros_planning` compiles, but fails linking with the following error:

```
Errors     << moveit_ros_planning:make /home/werner/catkin_test/logs/moveit_ros_planning/build.make.000.log                                                                                                                                                                   
/usr/bin/ld: CMakeFiles/moveit_compare_collision_checking_speed_fcl_bullet.dir/src/compare_collision_speed_checking_fcl_bullet.cpp.o: undefined reference to symbol 'ccd_vec3_origin'
/usr/bin/ld: /lib/x86_64-linux-gnu/libccd.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [planning_components_tools/CMakeFiles/moveit_compare_collision_checking_speed_fcl_bullet.dir/build.make:247: /home/werner/catkin_test/devel/.private/moveit_ros_planning/lib/moveit_ros_planning/moveit_compare_collision_checking_speed_fcl_bullet] Error 1
make[1]: *** [CMakeFiles/Makefile2:3192: planning_components_tools/CMakeFiles/moveit_compare_collision_checking_speed_fcl_bullet.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make: *** [Makefile:141: all] Error 2
cd /home/werner/catkin_test/build/moveit_ros_planning; catkin build --get-env moveit_ros_planning | catkin env -si  /usr/bin/make --jobserver-auth=3,4; cd -
```


### Backtrace or Console output

```
[...]

PR #2819, opened 2021-08-17
[moveit_core] Fix export of FCL dependency

jspricke:
Regression of 93c3f63
Closes: #2804


jspricke:
This is a little hacky but according to someone who does not want to be named "should work", alternatively we could export the target but `EXPORTED_TARGETS` in catkin did not work or use `CFG_EXTRAS`...

moveit_core/CMakeLists.txt
```diff
@@ -44,8 +44,16 @@ endif()
 
 find_package(fcl QUIET)
 if (fcl_FOUND)
-  set(LIBFCL_LIBRARIES ${FCL_LIBRARIES})
-  set(LIBFCL_INCLUDE_DIRS ${FCL_INCLUDE_DIRS})
+  # convert cmake target to catkin variables for catkin_package DEPENDS
```
gavanderhoorn, line 47: There not really *catkin* variables, are they?

jspricke: true.

moveit_core/CMakeLists.txt
```diff
@@ -44,8 +44,16 @@ endif()
 
 find_package(fcl QUIET)
 if (fcl_FOUND)
-  set(LIBFCL_LIBRARIES ${FCL_LIBRARIES})
-  set(LIBFCL_INCLUDE_DIRS ${FCL_INCLUDE_DIRS})
+  # convert cmake target to variables for catkin_package DEPENDS
+  if(TARGET ${FCL_LIBRARIES})
```
rhaschke, line 48: This will break if `FCL_LIBRARIES` becomes a list of targets, right?

jspricke: why would that happen? Either it is a target or it is a list, or?

rhaschke: If they decide in future versions of FCL to export multiple targets...

jspricke: I guess we can fix that in the future, or in other words let's fix this with moveit2 and ament ;).

rhaschke: Ok, let's be optimistic that they don't change that soon ;-)

rhaschke (approved):

v4hn:
Thanks for finding the culprit. half-exporting the target explains the behavior... I remember we previously used `CFG_EXTRAS` for Boost components somewhere, but I can't find it anymore.

I don't particularly like the solution because other libraries might follow in the future, but it's good enough for now. Thank you!

Issue #2809, opened 2021-08-12
Fix clang-tidy and use optimizations


tylerjw:
### Description

https://github.com/ros-planning/moveit/pull/2792  Changes the optimization level of one package to fix a clang-tidy issue.  This seems gross to me and we should be able to figure out another way of fixing this.

v4hn:
it's worse than gross because the file is at the core of collision checking with FCL. So the slowdown is probably quite big.

The issue is not with clang-tidy though. 
It's that building `libmoveit_collision_detection_fcl.so` with clang produces invalid code and we recently added a test that points this out.

PR #2820, opened 2021-08-17
Fix issue #2809 (broken test with clang)

rhaschke:
The culprit of #2792 was essentially an optimization I proposed for #2698 to avoid copying strings to generate a map key.
Looks like this optimization leads to a stack-use-after-scope issue (found with asan), both in clang and gcc.
Reverting it, fixes the issue.

However, I'm not sure why this code doesn't work - the string references should exist:
```c++
#include <map>
#include <string>
#include <iostream>

int main(int argc, char const* argv[])
{
  using KeyT = std::pair<std::string, std::string>;
  std::map<KeyT, int> map;
  map.insert(std::make_pair(std::make_pair("foo", "bar"), 1));
  map.insert(std::make_pair(std::make_pair("bar", "foo"), 2));

  const std::string foo = "foo";
  const std::string bar = "bar";

  const std::pair<const std::string&, const std::string&> key =
      std::make_pair<const std::string&, const std::string&>(foo, bar);

  /*** at this point accessing elements of key results in stack-use-after-scope ***/
  std::cerr << key.first << ", " << key.second << std::endl;

  auto it = map.find(key);
  std::cerr << it->second << std::endl;
  return 0;
}
```

### DIFF
--- a/moveit_core/collision_detection_fcl/src/collision_common.cpp
+++ b/moveit_core/collision_detection_fcl/src/collision_common.cpp
@@ -495,9 +495,9 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
 
   double dist_threshold = cdata->req->distance_threshold;
 
-  const std::pair<const std::string&, const std::string&> pc = cd1->getID() < cd2->getID() ?
-                                                                   std::make_pair(cd1->getID(), cd2->getID()) :
-                                                                   std::make_pair(cd2->getID(), cd1->getID());
+  const std::pair<const std::string, const std::string> pc = cd1->getID() < cd2->getID() ?
+                                                                 std::make_pair(cd1->getID(), cd2->getID()) :
+                                                                 std::make_pair(cd2->getID(), cd1->getID());
 
   DistanceMap::iterator it = cdata->res->distances.find(pc);
 
--- a/moveit_core/collision_detection_fcl/src/collision_common.cpp
+++ b/moveit_core/collision_detection_fcl/src/collision_common.cpp
@@ -495,9 +495,9 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
 
   double dist_threshold = cdata->req->distance_threshold;
 
-  const std::pair<const std::string, const std::string> pc = cd1->getID() < cd2->getID() ?
-                                                                 std::make_pair(cd1->getID(), cd2->getID()) :
-                                                                 std::make_pair(cd2->getID(), cd1->getID());
+  const std::pair<const std::string, const std::string> pc12(cd1->getID(), cd2->getID());
+  const std::pair<const std::string, const std::string> pc21(cd2->getID(), cd1->getID());
+  const std::pair<const std::string, const std::string>& pc = cd1->getID() < cd2->getID() ? pc12 : pc21;
 
   DistanceMap::iterator it = cdata->res->distances.find(pc);
 
--- a/moveit_core/collision_detection_fcl/src/collision_common.cpp
+++ b/moveit_core/collision_detection_fcl/src/collision_common.cpp
@@ -495,9 +495,9 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
 
   double dist_threshold = cdata->req->distance_threshold;
 
-  const std::pair<const std::string, const std::string> pc12(cd1->getID(), cd2->getID());
-  const std::pair<const std::string, const std::string> pc21(cd2->getID(), cd1->getID());
-  const std::pair<const std::string, const std::string>& pc = cd1->getID() < cd2->getID() ? pc12 : pc21;
+  const std::pair<const std::string&, const std::string&> pc =
+      cd1->getID() < cd2->getID() ? std::make_pair(std::cref(cd1->getID()), std::cref(cd2->getID())) :
+                                    std::make_pair(std::cref(cd2->getID()), std::cref(cd1->getID()));
 
   DistanceMap::iterator it = cdata->res->distances.find(pc);
 
--- a/moveit_core/collision_detection_fcl/CMakeLists.txt
+++ b/moveit_core/collision_detection_fcl/CMakeLists.txt
@@ -5,12 +5,6 @@ add_library(${MOVEIT_LIB_NAME}
   src/collision_env_fcl.cpp
 )
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
-if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-  # FIXME(v4hn): clang 10-12 produce broken code for the only std::map::insert call in the method
-  # however, the method calls in here are hotspots in the whole pipeline and disabling optimizations
-  # is no sufficient solution
-  set_source_files_properties(src/collision_common.cpp PROPERTIES COMPILE_FLAGS -O0)
-endif()
 
 
 target_link_libraries(${MOVEIT_LIB_NAME} moveit_collision_detection ${catkin_LIBRARIES} ${urdfdom_LIBRARIES} ${urdfdom_headers_LIBRARIES} ${LIBFCL_LIBRARIES} ${Boost_LIBRARIES})


rhaschke:
I found a trick to retain the original optimization using string references.

tylerjw (approved):
Approve because it gets rid of the building with `-O0`.  Generally though isn't copying a short string going to be one of the most optimized features of any modern compiler and implementation of the stl.  I as a rule generally avoid ever doing `std::anything<Anything &>::anything` because by using references with STL types you can create pessimizations by avoiding elision optimizations the compiler could otherwise do.

tylerjw:
I copy-pasted the example code into godbolt and I'm not sure what doesn't work about it: https://godbolt.org/z/5azE4Exc5

griswaldbrooks:
I likewise am confused. There is a single scope in the pasted example.

rhaschke:
> Generally though isn't copying a short string going to be one of the most optimized features of any modern compiler and implementation of the STL?

Here, it's not only copying the strings but also allocating new memory for them. That's one of the most costly operations one can do.

griswaldbrooks:
@rhaschke I'm confused. Why would the pair heap allocate if the string was previously stack allocated, which I believe was @tylerjw 's conjecture?

tylerjw:
> Here, it's not only copying the strings but also allocating new memory for them. That's one of the most costly operations one can do.

It isn't always expensive to copy a string: https://tc-imba.github.io/posts/cpp-sso
Plus if it is a local variable only there is a good chance the compile optimizes the copy out: https://abseil.io/tips/117

moveit_core/collision_detection_fcl/src/collision_common.cpp
```diff
@@ -495,9 +495,9 @@ bool distanceCallback(fcl::CollisionObjectd* o1, fcl::CollisionObjectd* o2, void
 
   double dist_threshold = cdata->req->distance_threshold;
 
-  const std::pair<const std::string&, const std::string&> pc = cd1->getID() < cd2->getID() ?
```
griswaldbrooks, line 498: Probably a matter of style, but I believe you could have equivalently removed the references and preserved the ternary.

tylerjw:
Here is another blog post that explains SSO it in a helpful way: https://akrzemi1.wordpress.com/2014/04/14/common-optimizations/

griswaldbrooks:
@rhaschke looking at the code/PR, my guess is asan isn't happy/able to peer through the multiple levels of pointer indirection. `getId()` had three possible sources for the reference, two of which are themselves references, all of which are pointers

https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection_fcl/include/moveit/collision_detection_fcl/collision_common.h#L86-L98

and here, `cd` is also a pointer. `pc` only holding references then conceivably could have it's references go out of scope before being passed to `distances.find`. Or at least that's my working theory.

griswaldbrooks:
Or maybe some version of https://docs.microsoft.com/en-us/cpp/sanitizers/error-stack-use-after-scope?view=msvc-160#example-3---destructor-ordering-with-locals

griswaldbrooks:
I also get `SUMMARY: AddressSanitizer: stack-use-after-scope (/lib/x86_64-linux-gnu/libasan.so.5+0x6a1ac) 
` when using `g++ -fsanitize=address -std=c++17 main.cpp`

griswaldbrooks:
@rhaschke you know what's fun?
```
#include <string>
#include <iostream>

struct MyPair {
    const std::string& first;
    const std::string& second;
};

int main(int argc, char const* argv[])
{
  const std::string foo = "foo";
  const std::string bar = "bar";

  const MyPair key{foo, bar};

  /*** at this point accessing elements of key results in stack-use-after-scope ***/

  std::cout << key.first << ", " << key.second << std::endl;


  return 0;
}
```
https://godbolt.org/z/cE5PbvsfM

does not produce any error with address sanitizer.

griswaldbrooks:
Also, this code is fine
```
#include <string>
#include <iostream>
#include <utility>

int main(int argc, char const* argv[])
{
    const std::string foo = "foo";
    const std::string bar = "bar";

    const std::pair<const std::string&, const std::string&> key{foo, bar};

    std::cout << key.first << ", " << key.second << std::endl;
    return 0;
}
```

griswaldbrooks:
Forgot to mention this still generates the error
```
#include <string>
#include <iostream>
#include <utility>

int main(int argc, char const* argv[])
{
    const std::string foo = "foo";
    const std::string bar = "bar";

    const std::pair<const std::string&, const std::string&> key = std::make_pair(foo, bar);

    std::cout << key.first << ", " << key.second << std::endl;
    return 0;
}
```

griswaldbrooks:
Also
```
#include <string>
#include <iostream>
#include <utility>

int main(int argc, char const* argv[])
{
    const std::string foo = "foo";
    const std::string bar = "bar";

    const std::pair<const std::string&, const std::string&> key = std::pair<const std::string&, const std::string&>(foo, bar);

    std::cout << key.first << ", " << key.second << std::endl;
    return 0;
}
```
is fine. Seems to be something with `std::make_pair`

griswaldbrooks:
```
const std::pair<const std::string&, const std::string&> key = std::make_pair(std::ref(foo), std::ref(bar));
```
is good too

JafarAbdi:
@griswaldbrooks If I understand it correctly, it's because of std::make_pair always returning the decayed type [see](https://en.cppreference.com/w/cpp/utility/pair/make_pair), which call the copy-constructor for `pc` causing its members `first` & `second` to reference temporaries (the copies of `foo` & `bar`)

JafarAbdi:
I think we should go with using `std::cref` since the behavior is well documented in the standard [see](https://en.cppreference.com/w/cpp/utility/pair/make_pair)

```
const auto pc = cd1->getID() < cd2->getID() ? std::make_pair(std::cref(cd1->getID()), std::cref(cd2->getID())) :
                                              std::make_pair(std::cref(cd2->getID()), std::cref(cd1->getID()));
```

JafarAbdi (approved):

v4hn:
@tylerjw wrote
> It isn't always expensive to copy a string: https://tc-imba.github.io/posts/cpp-sso

Very true, but I don't want to rely on implicit optimizations and end up in situations where there's a significant runtime difference depending on whether my collision object is called `armature_handle` or `armature_hex_cap` at runtime. The latter is not "short" in libstdc++.

moveit_core/collision_detection_fcl/src/collision_common.cpp
rhaschke: Keeping the (const) references ensure optimization.

rhaschke:
@JafarAbdi, thanks for pointing out `std::cref`. Indeed that resolves the issue. I also profiled the code. 
@tylerjw, even with SSO, there is significant overhead explicitly copying the strings (note that most of the time is actually spent in `std::map::find`)
```
ref !find: 84
with cref: 1399715
with copy: 2036093
```
I will push an update to this PR using `std::cref` later.

```c++
#include <map>
#include <string>
#include <iostream>
#include <functional>
#include <chrono>

int main(int argc, char const* argv[])
{
  using KeyT = std::pair<std::string, std::string>;
  std::map<KeyT, int> map;
  map.insert(std::make_pair(std::make_pair("foo", "bar"), 1));
  map.insert(std::make_pair(std::make_pair("bar", "foo"), 2));

  const std::string foo = "foo";
  const std::string bar = "bar";

  auto start = std::chrono::steady_clock::now();
  for (int i = 0; i < 10000; ++i)
  {
    const std::pair<const std::string&, const std::string&> key1(foo, bar);
    const std::pair<const std::string&, const std::string&> key2(bar, foo);
    const std::pair<const std::string&, const std::string&>& key = foo < bar ? key1 : key2;
  }
  auto elapsed = std::chrono::steady_clock::now() - start;
  std::cerr << "ref !find: " << elapsed.count() << std::endl;

  start = std::chrono::steady_clock::now();
  for (int i = 0; i < 10000; ++i)
  {
    const std::pair<const std::string&, const std::string&> key1(foo, bar);
    const std::pair<const std::string&, const std::string&> key2(bar, foo);
    const std::pair<const std::string&, const std::string&>& key =
        foo < bar ? std::make_pair(std::cref(foo), std::cref(bar)) : std::make_pair(std::cref(bar), std::cref(foo));

    auto it = map.find(key);
  }
  elapsed = std::chrono::steady_clock::now() - start;
  std::cerr << "with cref: " << elapsed.count() << std::endl;

  start = std::chrono::steady_clock::now();
  for (int i = 0; i < 10000; ++i)
  {
    const std::pair<const std::string, const std::string> key1(foo, bar);
    const std::pair<const std::string, const std::string> key2(bar, foo);
    const std::pair<const std::string, const std::string>& key = foo < bar ? key1 : key2;

    auto it = map.find(key);
  }
  elapsed = std::chrono::steady_clock::now() - start;
  std::cerr << "with copy: " << elapsed.count() << std::endl;
  return 0;
}
```

griswaldbrooks:
@rhaschke very cool. Throwing it into quickbench shows a 2.5x speedup https://quick-bench.com/q/CtWmHGp4nHOglaietXVNfi8r7cY

@JafarAbdi do you have another reference about `std::decay` and the temporaries from `make_pair`? It wasn't obvious to me from the `std::make_pair` documentation.

moveit_core/collision_detection_fcl/src/collision_common.cpp
griswaldbrooks: Yup yup. Was more of a comment before the `std::cref` became the choice thing to do.

griswaldbrooks:
@JafarAbdi nvm, I got it
https://stackoverflow.com/questions/57860088/do-i-need-to-use-ref-with-make-pair-to-return-a-reference
https://stackoverflow.com/questions/25732386/what-is-stddecay-and-when-it-should-be-used

PR #2821, opened 2021-08-17
CI: Enable -Werror only for TARGET_WORKSPACE

rhaschke:
Warnings in downstream or upstream workspaces (i.e. external packages) shouldn't make CI fail due to `-Werror`.
Addresses https://github.com/ros-planning/moveit_visual_tools/pull/96#issuecomment-897069559.

[This CI build](https://github.com/ubi-agni/moveit/runs/3354072333?check_suite_focus=true) illustrates that warnings in external packages don't harm the CI job anymore.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
           - IMAGE: master-ci # ROS Melodic with all dependencies required for master
             CATKIN_LINT: true
     env:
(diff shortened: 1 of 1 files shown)


v4hn:
I agree that upstream should not use `-Werror`. But as I said before downstream warnings can appear because of bad changes in the framework API. A prominent example in the ros2 API came up [here](https://github.com/ros-planning/moveit2/pull/590#discussion_r682685839).

As we have full control over all repos in CI's downstream, I would prefer to keep it enabled there. After all, all current warnings there are resolved.

v4hn (approved):
Looks good if CI is happy with it :+1:

PR #2822, opened 2021-08-18
Remove python3-osrf-pycommon from Dockerfile

mintar:
This was a workaround for https://github.com/catkin/catkin_tools/issues/594. It is no longer necessary since python3-catkin-tools 0.7 added the proper dependency.


rhaschke (approved):
Great, that they finally released a new version!

Issue #2773, opened 2021-07-16
Pilz planner always allow collision between attached bodies and robot/world


cambel:
### Description

When planning with Pilz planners, the attached bodies are not considered for collision checking.

### Your environment
Ubuntu 18.04, Melodic, source build

### Steps to reproduce
Attach a body to the robot, define a target pose where ONLY the attach body is in collision with some other object, plan with any Pilz planner.

### Expected behaviour
Plan must fail.

### Actual behaviour
Plan succeed.

### Backtrace or Console output

Here is a video comparing planning with OMPL vs Pilz and the corresponding [logs](https://gist.github.com/cambel/5a7a01ec3ccc6e4a399d04a095237e30) 

https://user-images.githubusercontent.com/3798796/125898070-4ce1f7b1-c9aa-44af-8cc1-fbfb41cf53a4.mp4

cambel:
This is on top of #2676, after applying the changes in PR #2754, the collision check between the robot and the world with respect to the AllowCollisionMatrix works okay. However, the attached bodies seem to be ignored.

How does the planning scene knows that it should check the attached bodies when requesting a collisionCheck? I am checking this issue but I don't really get why the planning pipeline is ignoring this collision checks. Here in the planning_pipeline, there is a recheck of the plan generated by the planner. And as you can see in the logs above (with debug activated), the plan seems to be valid: `Planned path was found to be valid when rechecked`
https://github.com/ros-planning/moveit/blob/e8b54db3ae61a1314e49656ad84795d4cfc2e523/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp#L260-L336

cambel:
Thanks to @felixvd we were able to pin-point the problem. The Pilz planner were creating a `start_state` based only on the `robot_model` stored on instantiation of the planner, thus the attached bodies were not present. The solution is to use the current state from planning scene instead.
Since this issue is related to an open PR (#2754), so we leave the proposed solution [here](https://github.com/cambel/moveit/commit/b97e32085e7bf774772737c0ed3c06aba4730008) for now until the open PR is closed, or this changes can be merge there.

rhaschke:
Should be fixed via #2803.

cambel:
@rhaschke, no, this issue is not solved yet with #2803 but now I can PR the solution we mentioned in my previous comment.

PR #2824, opened 2021-08-18
Consider attached bodies in Pilz planner #2773

cambel:
### Description

See to #2773.

The Pilz planners create a `start_state` based only on the `robot_model` stored on instantiation of the planner, thus any new attached bodies are not present during planning. The solution here is to use the current state from the planning scene instead.

**Note**: the planning on the Pilz planner side still succeed (and maybe it shouldn't), however, with this PR's changes the extra recheck on the planning pipeline catch the invalid plan here: https://github.com/ros-planning/moveit/blob/e8b54db3ae61a1314e49656ad84795d4cfc2e523/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp#L260-L336 
 
### Checklist


moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator.cpp
```diff
@@ -214,23 +214,25 @@ void TrajectoryGenerator::validateRequest(const planning_interface::MotionPlanRe
   checkGoalConstraints(req.goal_constraints, req.start_state.joint_state.name, req.group_name);
 }
 
-void TrajectoryGenerator::convertToRobotTrajectory(const trajectory_msgs::JointTrajectory& joint_trajectory,
+void TrajectoryGenerator::convertToRobotTrajectory(const planning_scene::PlanningSceneConstPtr& scene,
```
rhaschke, line 217: What about passing in the start state only? This makes the API more self-explanatory and more specific.

rhaschke (commented):
I generally approve, but I suggest to pass only the minimally required info, namely the start state (instead of the full planning scene).

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator.cpp
```diff
@@ -214,23 +214,25 @@ void TrajectoryGenerator::validateRequest(const planning_interface::MotionPlanRe
...
 }
 
-void TrajectoryGenerator::convertToRobotTrajectory(const trajectory_msgs::JointTrajectory& joint_trajectory,
+void TrajectoryGenerator::convertToRobotTrajectory(const planning_scene::PlanningSceneConstPtr& scene, 
+                                                   const trajectory_msgs::JointTrajectory& joint_trajectory,
                                                    const moveit_msgs::RobotState& start_state,
                                                    robot_trajectory::RobotTrajectory& robot_trajectory) const
 {
-  moveit::core::RobotState start_rs(robot_model_);
+  moveit::core::RobotState start_rs(scene->getCurrentState());
```
jschleicher, line 222: What't the purpose of this change? `robotStateMsgToRobotState` overwrites start_rs two lines below, so this line only default-inializes a RobotState?

jschleicher (commented):
A question about the general approach: In https://github.com/ros-planning/moveit/issues/2773#issuecomment-881250168 you conclude, that the start state is taken from the robot model instantiated in the constructor.
Would it be feasible, to update the `TrajectoryGenerator::robot_model_` member variable from the scene on every new planning request (dropping the const-ness)? Or maybe drop the `robot_model_` member completely and use `scene->getRobotModel ()` throughout the TrajectoryGenerator class since we now have the scene available everywhere?
Could you add a testcase failing before and passing after your fix?

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator.cpp
cambel: Sure.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator.cpp
cambel: Now that you mention it. You are right, actually, it does not make much sense, even with the default-initialized `RobotState` using the `robot_model_` should work but the problem was the third parameter of the `robotStateMsgToRobotState`, it is not copying the attached collision objects.

cambel:
Based on @jschleicher comment, I change the code to a minimal change.

> maybe drop the robot_model_ member completely and use scene->getRobotModel () throughout the TrajectoryGenerator class since we now have the scene available everywhere?

I agree these kinds of errors may be avoided by using the scene instead of a fixed robot model.

moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator.cpp
v4hn: > robotStateMsgToRobotState overwrites start_rs two lines below, so this line only default-inializes a RobotState?

I disagree! If the RobotState message does not specify all joints (it is allowed to do that, esp. when `is_diff == true`), all other joint values should be taken from the passed scene and not be set to default! This was another undiscovered issue in this code. Please add this change back to your PR :-)

v4hn:
> I agree these kinds of errors may be avoided by using the scene instead of a fixed robot model.

This error would still have been there. Without looking through the code layers in the pilz planner, the core of this is most likely that MoveIt's [`MotionPlanRequest`](https://github.com/ros-planning/moveit_msgs/blob/master/msg/MotionPlanRequest.msg#L7-L10) allows to add a diff to the planning scene's robot state and sadly all planners have to handle this diff field individually for historical reasons.

Nonetheless :+1: to remove `robot_model_` and get it from the scene where required. This might become relevant in the ongoing debates to allow for dynamic robot models.

cambel:
Thanks for your comments. For now, I reverted to passing down the start state from the current planning scene, but only passing down the `RobotState` as suggested by @rhaschke

rhaschke (approved):
I further simplified the API. Thanks for your work @cambel.

jschleicher (approved):
Looks good to me now. Are you working on a testcase? Or should we merge as-is?

PR #2826, opened 2021-08-19
[noetic] Fix ccache

vatanaksoytezer:
Our ccache is broken in CI, thus our ccache action should be updated to match Github's new ccache action v2 as per https://github.com/ros-planning/moveit2/pull/619


rhaschke:
Merged via fast-forward of master

PR #2598, opened 2021-04-12
 Lock the Bullet collision environment, for thread safety

v4hn:
Replaces #2481 and adds minor refinements.

Please review and merge @AndyZe. You disabled maintainer-pushes to your fork, so I had to create a new pull-request.

PR #2830, opened 2021-08-20
Fix bullet constructor not updating world objects

yurirocha15:
### Description

Bullet Collision Env's "copy constructor" does not update the manager with the copied world objects (by calling the notify function).  This leads to an empty collision environment when planning a trajectory.

This is the constructor called in https://github.com/ros-planning/moveit/blob/master/moveit_core/planning_scene/src/planning_scene.cpp#L393-L397

which is used by the planning_scene_monitor to keep the move_group scene updated. 

### Checklist

### DIFF
--- a/moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
+++ b/moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
@@ -89,6 +89,9 @@ CollisionEnvBullet::CollisionEnvBullet(const CollisionEnvBullet& other, const Wo
   {
     addLinkAsCollisionObject(link.second);
   }
+
+  // get notifications any objects already in the new world
+  getWorld()->notifyObserverAllObjects(observer_handle_, World::CREATE);
 }
 
 CollisionEnvBullet::~CollisionEnvBullet()
--- a/moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
+++ b/moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
@@ -80,8 +80,6 @@ CollisionEnvBullet::CollisionEnvBullet(const moveit::core::RobotModelConstPtr& m
 CollisionEnvBullet::CollisionEnvBullet(const CollisionEnvBullet& other, const WorldPtr& world)
   : CollisionEnv(other, world)
 {
-  // TODO(j-petit): Verify this constructor
-
   // request notifications about changes to new world
   observer_handle_ = getWorld()->addObserver(boost::bind(&CollisionEnvBullet::notifyObjectChange, this, _1, _2));
 
--- a/moveit_core/planning_scene/CMakeLists.txt
+++ b/moveit_core/planning_scene/CMakeLists.txt
@@ -14,6 +14,7 @@ target_link_libraries(${MOVEIT_LIB_NAME}
   moveit_exceptions
   moveit_transforms
   moveit_collision_detection_fcl
+  moveit_collision_detection_bullet
   moveit_kinematic_constraints
   moveit_robot_trajectory
   moveit_trajectory_processing
--- a/moveit_core/planning_scene/CMakeLists.txt
+++ b/moveit_core/planning_scene/CMakeLists.txt
@@ -8,17 +8,12 @@ generate_export_header(${MOVEIT_LIB_NAME} EXPORT_FILE_NAME ${CATKIN_DEVEL_PREFIX
 target_include_directories(${MOVEIT_LIB_NAME} PRIVATE $<BUILD_INTERFACE:${CATKIN_DEVEL_PREFIX}/include>)  # for this library
 list(APPEND ${PROJECT_NAME}_INCLUDE_DIRS ${CATKIN_DEVEL_PREFIX}/include)  # for use by other libraries in moveit_core
 
-if(BULLET_ENABLE)
-    set(BULLET_LIB moveit_collision_detection_bullet)
-endif()
-
 target_link_libraries(${MOVEIT_LIB_NAME}
   moveit_robot_model
   moveit_robot_state
   moveit_exceptions
   moveit_transforms
   moveit_collision_detection_fcl
-  ${BULLET_LIB}
   moveit_kinematic_constraints
   moveit_robot_trajectory
   moveit_trajectory_processing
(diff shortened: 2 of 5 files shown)


moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
```diff
@@ -89,6 +89,9 @@ CollisionEnvBullet::CollisionEnvBullet(const CollisionEnvBullet& other, const Wo
   {
```
v4hn, line 89: If you actually use this constructor and it works for you with this PR, please remove the `TODO` to verify it above.

moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
```diff
@@ -89,6 +89,9 @@ CollisionEnvBullet::CollisionEnvBullet(const CollisionEnvBullet& other, const Wo
...
     addLinkAsCollisionObject(link.second);
   }
+
+  // get notifications any objects already in the new world
+  getWorld()->notifyObserverAllObjects(observer_handle_, World::CREATE);
```
v4hn, line 94: That's quite involved. I'd have to invest quite a bit of time to see whether this is actually correct.
A test/tests to illustrate the different behavior would be appreciated here.

The bullet plugin was based on the FCL checker.
Can you explain why [the same constructor for FCL works without this additional line?](https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp#L152-L163)

yurirocha15: I believe FCL updates the world objects here https://github.com/ros-planning/moveit/blob/master/moveit_core/collision_detection_fcl/src/collision_env_fcl.cpp#L157-L159 which is why it does not call the notifier.

I will work on some tests and ping you after I get them to work.

yurirocha15:
@v4hn I added the tests on the planning_scene folder as I believe they are related to the child scene collision environment not being correctly updated (when using Bullet). I created the same test for FCL as well. I verified that the test fail before this fix and pass after it.

As I had to add the bullet dependency to the tests, I also replicated the FCL multithreaded test (as #2598 made Bullet thread-safe).

moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
yurirocha15: Removed in 5495a9c

yurirocha15:
RoboStack's Windows build cannot find bullet:
```shell
-- Checking for module 'bullet'
--   No package 'bullet' found
-- Could NOT find Bullet: Found unsuitable version "", but required is at least "2.87" (found )
-- Version of Bullet too old or not available: disabling Bullet collision detection plugin. Try using Ubuntu 18.04 or later.
```
[Because of this BULLET_INC not being set](https://github.com/ros-planning/moveit/blob/master/moveit_core/CMakeLists.txt#L36-L43)  which leads to a build error:
```shell
fatal error C1083: Cannot open include file: 'moveit/collision_detection_bullet/collision_detector_allocator_bullet.h': No such file or directory
```

Any suggestions on how to proceed from here?

moveit_core/planning_scene/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ target_link_libraries(${MOVEIT_LIB_NAME}
   moveit_exceptions
   moveit_transforms
   moveit_collision_detection_fcl
+  moveit_collision_detection_bullet
```
rhaschke, line 17: This lib shouldn't be added explicitly here as it is a plugin!

moveit_core/planning_scene/test/test_multi_threaded.cpp
```diff
@@ -44,6 +44,8 @@
 #include <moveit/collision_detection/collision_common.h>
 #include <moveit/collision_detection/collision_env.h>
 #include <moveit/collision_detection/collision_detector_allocator.h>
+#include <moveit/collision_detection_bullet/collision_detector_allocator_bullet.h>
+#include <moveit/collision_detection_bullet/collision_env_bullet.h>
```
rhaschke, line 48: These includes are not available when Bullet is not available. You need to conditionally include them!
My suggestion is to add add_definitions(-DBULLET_ENABLE) [here](https://github.com/ros-planning/moveit/blob/4d6f805b8375104fd5dac4791032a9c11353fffe/moveit_core/CMakeLists.txt#L36-L40) and in c++ code use:
```cpp
#ifdef BULLET_ENABLE
...
#endif
```

rhaschke (commented):
The Bullet code should be disregarded if Bullet was not found.

moveit_core/planning_scene/test/test_multi_threaded.cpp
yurirocha15: Thank you for the suggestion! Added in 2433753

moveit_core/planning_scene/CMakeLists.txt
yurirocha15: If I remove it, the tests fail to compile as they cannot find the `collision_detection::CollisionDetectorAllocatorBullet` definition. I thought it was ok to link bullet here as there is a link for fcl already.

yurirocha15: Should I also add a condition here to only link the bullet plugin if the code was built with Bullet?

rhaschke: Ideally, you would load the plugin via the plugin mechanism instead of instantiating it directly ;-)

yurirocha15: Got it! Added in c28b9ae

rhaschke:
I will have a look into the remaining CI failure. Seems to be related to clang being used as the compiler...

yurirocha15:
> I will have a look into the remaining CI failure. Seems to be related to clang being used as the compiler...

I ran industrial_ci locally and was able to fix it. I believe it was related to resetting the class_loader ptr and then resetting the planning_scene ptrs. After explicitly setting the destruction order the segmentation fault disappeared (at least when running locally), Hopefully, it will pass now.

rhaschke (changes requested):
To avoid code duplication in tests, I provided https://github.com/makinarocks/moveit/pull/2 with more simplifications. However, this relies on https://github.com/ros-planning/moveit/pull/2834.
Otherwise, I approve.

yurirocha15:
@rhaschke as #2834 was merged, I cherry-picked the commits from https://github.com/makinarocks/moveit/pull/2 and rebased the branch on top of master.

rhaschke:
I pushed another commit to https://github.com/makinarocks/moveit/pull/2 to fix the `gtest` compatibility issues: https://github.com/makinarocks/moveit/pull/2/commits/893e63f2a1b1769343532d353f5c5b84c1de742a

rhaschke (approved):

Issue #2831, opened 2021-08-22
git version same as last release? 


v4hn:
why is [moveit's current version in the repo] (https://github.com/ros-planning/moveit/blob/1859ec2d248074c983367735f3034f3172de07b2/moveit/package.xml#L4)  1.1.5, when we  already released that version? didn't we agree some time ago to bump the versions right after each release?

v4hn:
[Here](https://github.com/ros-planning/moveit/issues/2036)'s the related issue of the original discussion.

I found this because #2037 introduced some API changes I wanted to account for in my code base.

rhaschke:
I vaguely remember... Looks like, we have followed this rule only _once_ in the past: after releasing version 1.0.4.
I guess the main reason that we didn't follow the rule in practice, is that the required procedure is much more complicated.
Maybe we should do it only if explicitly requested, like now? @tylerjw, what do you think?

rhaschke:
I added a corresponding PR: #2832.

v4hn:
Thanks! 

I wouldn't call one `sed` command + commit "much more complicated", but it's an additional step to take care of that is not enforced by the release tools... :confused: 
Also, the next person releasing should remember not to bump the version again.
Although that might actually be a valid practice and would result in odd numbers for releases and even ones for unreleased states.

Reopening the issue to finish the discussion on how to handle this in the future (or possibly how to help the release team remember it?)

rhaschke:
I think that's an item for the next maintainer meeting...

PR #2832, opened 2021-08-22
Bump version: 1.1.6

rhaschke:
Bump version number as requested in #2831.

### DIFF
--- a/moveit/package.xml
+++ b/moveit/package.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package format="2">
   <name>moveit</name>
-  <version>1.1.5</version>
+  <version>1.1.6</version>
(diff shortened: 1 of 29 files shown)


v4hn (approved):
Thanks, I was about to file the same PR.
I was just pondering whether to bump the trajopt `package.xml` as well, it seems @tylerjw and you don't agree on whether to bump it :)

rhaschke:
I didn't explicitly handle `trajopt` differently. Digging into it, I noticed that this package is currently `CATKING_IGNORE`d. Hence, it's not considered for version bumps automatically.

v4hn:
Yes, I don't believe it's in a useful state, but I never found the time to look into this GSoC outcome.
I guess it's just a matter of whether you use the bloom tools or sed to bump the package.xml entries.

PR #2836, opened 2021-08-24
Keep ob::PlannerTerminationCondition on heap

vinnamkim:
### Description
There is currently an invalid memory reference bug when constructing `ompl::base::plannerOrTerminationCondition` from `constructPlannerTerminationCondition`.

https://github.com/ompl/ompl/blob/9c3a20faaddfcd7f58fce235495f043ebee3e735/src/ompl/base/src/PlannerTerminationCondition.cpp#L197-L204

`ompl::base::plannerOrTerminationCondition` expects `ompl::base::PlannerTerminationCondition` alive in the memory while it is alive.
However, `constructPlannerTerminationCondition` allocates `ompl::base::PlannerTerminationCondition`s in the stack memory, then they are automatically destructed when returning `ompl::base::plannerOrTerminationCondition`.
This PR prevents this invalid memory reference bug by allocating  `ompl::base::PlannerTerminationCondition` on the heap memory for the lifetime of `ompl::base::plannerOrTerminationCondition`.

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/model_based_planning_context.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/model_based_planning_context.h
@@ -46,6 +46,14 @@
 #include <ompl/tools/multiplan/ParallelPlan.h>
 #include <ompl/base/StateStorage.h>
 
+namespace ompl
+{
+namespace base
+{
+class IterationTerminationCondition;
+}
+}  // namespace ompl
+
 namespace ompl_interface
 {
 namespace ob = ompl::base;
@@ -392,6 +400,8 @@ protected:
   std::vector<kinematic_constraints::KinematicConstraintSetPtr> goal_constraints_;
 
   const ob::PlannerTerminationCondition* ptc_;
+  std::vector<std::shared_ptr<ob::PlannerTerminationCondition>> ptcs_;
+  std::shared_ptr<ob::IterationTerminationCondition> it_ptc_;
   std::mutex ptc_lock_;
 
   /// the time spent computing the last plan
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -505,9 +505,13 @@ ompl_interface::ModelBasedPlanningContext::constructPlannerTerminationCondition(
   else if (termination_and_params[0] == "Iteration")
   {
     if (termination_and_params.size() > 1)
-      return ob::plannerOrTerminationCondition(
-          ob::timedPlannerTerminationCondition(timeout - ompl::time::seconds(ompl::time::now() - start)),
-          ob::IterationTerminationCondition(std::stoul(termination_and_params[1])));
+    {
+      ptcs_.clear();
+      ptcs_.push_back(std::make_shared<ob::PlannerTerminationCondition>(
+          ob::timedPlannerTerminationCondition(timeout - ompl::time::seconds(ompl::time::now() - start))));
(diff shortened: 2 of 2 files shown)


v4hn:
> ompl::base::plannerOrTerminationCondition expects ompl::base::PlannerTerminationCondition alive in the memory while it is alive.

Please provide proof for that claim, the implementation (see [here](https://github.com/ompl/ompl/blob/main/src/ompl/base/src/PlannerTerminationCondition.cpp#L200)) seems to capture the objects [implicitly by copy](https://en.cppreference.com/w/cpp/language/lambda), so I don't see a problem? Did you experience a bug that was fixed by this change?

vinnamkim:
> > ompl::base::plannerOrTerminationCondition expects ompl::base::PlannerTerminationCondition alive in the memory while it is alive.
> 
> Please provide proof for that claim, the implementation (see [here](https://github.com/ompl/ompl/blob/main/src/ompl/base/src/PlannerTerminationCondition.cpp#L200)) seems to capture the objects [implicitly by copy](https://en.cppreference.com/w/cpp/language/lambda), so I don't see a problem? Did you experience a bug that was fixed by this change?

More precisely, the problem occurs when giving the derived classes of `ompl::base::PlannerTerminationConditions` to `ompl::base::plannerOrTerminationCondition` as inputs (Moreover, `ompl::base::IterationTerminationCondition` is not derived class of  `ompl::base::PlannerTerminationConditions`). This is because [implicitly by copy](https://en.cppreference.com/w/cpp/language/lambda) of the base class can't copy the member variables of the derived class.

First, I did some experimentation with the `termination_condition` option turned on and found something odd. After that I added the custom termination condition as follows.

```
class NumSolutionTerminationCondition : public PlannerTerminationCondition
{
public:
  NumSolutionTerminationCondition(ProblemDefinitionPtr& pdef, size_t solutionsWindow);
  void processNewSolution(const Cost& solutionCost);

private:
  unsigned int numSolutions_;
  unsigned int solutionsWindow_;
  std::mutex mutex_;

NumSolutionTerminationCondition::NumSolutionTerminationCondition(ProblemDefinitionPtr& pdef, size_t solutionsWindow)
  : numSolutions_(0u), solutionsWindow_(solutionsWindow), PlannerTerminationCondition(plannerNonTerminatingCondition())
{
  pdef->setIntermediateSolutionCallback([this](const Planner* /*planner*/, const std::vector<const State*>& /*states*/,
                                               const Cost cost) { this->processNewSolution(cost); });
  ROS_INFO("NumSolutionTerminationCondition : %u", solutionsWindow_);
}

void NumSolutionTerminationCondition::processNewSolution(const Cost& solutionCost)
{
  std::lock_guard<std::mutex> lock_guard(mutex_);

  numSolutions_++;

  ROS_INFO("processNewSolution [%u/%u]", numSolutions_, solutionsWindow_);

  if (numSolutions_ == solutionsWindow_)
    terminate();
}
```

It showed that it was pointing to the wrong memory address for it's member variable:
```
processNewSolution [3546546843/4651654654]
```

rhaschke:
I agree. The objects should be captured by reference in OMPL. I'm surprised that this worked at all.

vinnamkim:
@v4hn @rhaschke 

I guess that capture-by-reference would not be helpful for our usage. It is same to capture the pointer of the variable. But, our variable is in the stack memory and they gets lost after the function returns. I agree that this should be fixed in OMPL's side. That would be much clearer.

v4hn:
> I guess that capture-by-reference would not be helpful for our usage.
> It is same to capture the pointer of the variable.
> But, our variable is in the stack memory and they gets lost after the function returns.

Yes, if OMPL will fix their end, we can also merge your patch. :)

rhaschke:
I filed a PR at OMPL: https://github.com/ompl/ompl/issues/830. With the suggested changes, this PR becomes superfluous.

simonschmeisser:
I couldn't find the PR @rhaschke just an issue

PR #2838, opened 2021-08-25
supports for non-convex mesh in collision_detection_bullet

CUN-bjy:
### Features

- Add concave mesh mode to MoveIt bullet collision detector
  - only for WorldObject in PlanningScene.
- Add a Flag for world status(is static or not)

### Details

***Problem***

1. In collision_detection_bullet, all collisions of mesh-type are converted to convex-hull type.

     - When non-convex meshes are imported, collisions are detected in free space.

       - On the other hand, in FCL, primitive triangles are directly added to AABB tree for collision checking.
         https://github.com/ros-planning/moveit/blob/bef8f7cdde399287dc8b75b9a36b4da54704b69d/moveit_core/collision_detection_fcl/src/collision_common.cpp#L829-L847
       - these outcomes are not intuitive for beginners.
         (https://github.com/ros-planning/moveit/issues/742#issuecomment-358353284)

       

> *why we use convex-hull:*
>
> - using a convex shape is good enough for speed-up. 
> - for continuous collision detection, meshes have to be convex shape if they are dynamic.
>   (don't need to be a convex-shape for a static object)

**_Solution_**

1. Import original mesh shape(when the world objects are static)
   - Add **_a mode for importing primitive triangles from original meshes_** in collision_detection_bullet (like using FCL).
   - Add one more option for users:
     - (default) fast computation, but use convex-hull(casual mesh) for collision detection 
     - (option) slow computation, but use non-convex shape(sophisticated mesh) for collision detection
       - if world objects are "static" and the user wants to use non-convex mesh in the planning scene, the user can choose it.
   - how to make bullet recognize which object is needed to be imported as non-convex?
     1. **_chosen by world information_**
        - we can add some information to `World` class.
          - if `use_concave_mesh_` is true, we can import the original mesh(not converted to convex-hull)
     2. chosen by object information
        - we need to change urdf and its parser to contain this info (convex or not) 
          - https://discourse.ros.org/t/add-attribute-to-indicate-if-a-mesh-is-a-convex-hull/4479/5
          - https://gist.github.com/erwincoumans/6e98f5ef6186f883933a41e3e2c84104
          - https://ros-industrial-tesseract.readthedocs.io/en/latest/_source/tesseract_urdf_doc.html
        - this approach is the best, *but it seems difficult to be applied..* this is a long-term goal.
   - (with dynamic non-convex mesh, use convex decomposition; solution below.)

> *(another solution)*
>
> Use V-HACD and make a collision array for a link 
>
> - In [tesseract](https://github.com/ros-industrial-consortium/tesseract), vhacd module is embedded with bullet. but, in moveit, the module is not included so the processes are messy.
> - convex decomposition(such as vhacd) make the original shapes approximated. Thus, we couldn't use this when it is required to plan high-accuracy motion.

## Results
**BEFORE** applying this PR

**AFTER** applying this PR

**TEST**
(scenario)
1. check collision on robot state1
2. check collision on robot state2
3. check conituous collision on state1 to state2

## Checklist

@rhaschke @v4hn  I think this feature needs to be included in bullet, pls review and feedback pls.

### DIFF
--- a/moveit_core/collision_detection/src/world.cpp
+++ b/moveit_core/collision_detection/src/world.cpp
@@ -41,11 +41,11 @@
 
 namespace collision_detection
 {
-World::World()
+World::World() : use_detailed_mesh_(false)
 {
 }
 
-World::World(const World& other)
+World::World(const World& other) : use_detailed_mesh_(false)
 {
   objects_ = other.objects_;
 }
--- a/moveit_ros/planning/collision_plugin_loader/src/collision_plugin_loader.cpp
+++ b/moveit_ros/planning/collision_plugin_loader/src/collision_plugin_loader.cpp
@@ -37,6 +37,28 @@
 static const std::string LOGNAME = "collision_detection";
 namespace collision_detection
 {
+
+void getUseDetailedMeshParam(ros::NodeHandle& nh, const planning_scene::PlanningScenePtr& scene)
+{
+  std::string param_name;
+  bool use_detailed_mesh;
+
+  if (nh.searchParam("use_detailed_mesh", param_name))
+  {
+    nh.getParam(param_name, use_detailed_mesh);
+  }
+  else if (nh.hasParam("/move_group/use_detailed_mesh"))
+  {
+    nh.getParam("/move_group/use_detailed_mesh", use_detailed_mesh);
+  }
+  else
+  {
+    use_detailed_mesh = false;
+  }
+
+  scene->getWorldNonConst()->setUseDetailedMesh(use_detailed_mesh);
+}
+
 void CollisionPluginLoader::setupScene(ros::NodeHandle& nh, const planning_scene::PlanningScenePtr& scene)
 {
   if (!scene)
@@ -68,6 +90,8 @@ void CollisionPluginLoader::setupScene(ros::NodeHandle& nh, const planning_scene
     // This is not a valid name for a collision detector plugin
     return;
   }
+
+  getUseDetailedMeshParam(nh, scene);
 
   activate(collision_detector_name, scene, true);
   ROS_INFO_STREAM("Using collision detector:" << scene->getActiveCollisionDetectorName().c_str());
--- a/moveit_core/collision_detection/include/moveit/collision_detection/world.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/world.h
@@ -75,6 +75,20 @@ public:
 
   MOVEIT_STRUCT_FORWARD(Object);
 
+  /** \brief Specify whether detailed meshes or convex hulls (the default) will be used
+   * If doing continuous collision detection, only convex hull meshes should be used.
+   * However if the world objects are 'static', you can enable this.
+   */
+  void setUseDetailedMesh(const bool use_detailed_mesh)
+  {
+    use_detailed_mesh_ = use_detailed_mesh;
+  }
+
+  bool getUseDetailedMesh()
+  {
+    return use_detailed_mesh_;
+  }
+
   /** \brief A representation of an object */
   struct Object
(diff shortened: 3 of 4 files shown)


simonschmeisser:
bounding-mesh sounds interesting but apparently the convex decomposition uses GCAL which makes it GPL

moveit_ros/planning/collision_plugin_loader/src/collision_plugin_loader.cpp
```diff
@@ -89,6 +89,27 @@ class CollisionPluginLoader::CollisionPluginLoaderImpl
     return false;
   }
 
+  void loadWorldOption(ros::NodeHandle& nh, const planning_scene::PlanningScenePtr& scene)
```
AndyZe, line 92:
```suggestion
  void getUseDetailedMeshParam(ros::NodeHandle& nh, const planning_scene::PlanningScenePtr& scene)
```

moveit_ros/planning/collision_plugin_loader/src/collision_plugin_loader.cpp
```diff
@@ -139,6 +160,8 @@ void CollisionPluginLoader::setupScene(ros::NodeHandle& nh, const planning_scene
     return;
   }
 
+  loader_->loadWorldOption(nh, scene);
```
AndyZe, line 163:
```suggestion
  loader_->getUseDetailedMeshParam(nh, scene);
```

AndyZe (approved):

AndyZe:
The function name still doesn't satisfy me but it looks good besides that. Approved assuming that change is made.  :+1:

moveit_core/collision_detection/src/world.cpp
```diff
@@ -41,11 +41,11 @@
 
 namespace collision_detection
 {
-World::World()
+World::World() : use_detailed_mesh_(false)
```
simonschmeisser, line 44:
```suggestion
World::World() : use_detailed_mesh_(true)
```
I would expect this to be the default and using convex hulls as an optional speedup/hack. I would also recommend to rename it to `approximate_meshes_using_convex_hull_` (with default `false`)

simonschmeisser (changes requested):

moveit_core/collision_detection/src/world.cpp
AndyZe: Defaulting to non-convex meshes is going to cause problems because this same code is used by FCL. I think convex hulls must be the default.

CUN-bjy: I agree with @AndyZe. this is an option to use concave mesh. Basically, bullet uses convex-based collision checking. and using concave_mesh_force parameter is an option on the bullet, too.

gavanderhoorn:
@simonschmeisser wrote:
> bounding-mesh sounds interesting but apparently the convex decomposition uses GCAL which makes it GPL

I wasn't suggesting you integrate it here.

Just that users would pre-process their collision meshes.

moveit_core/collision_detection/src/world.cpp
simonschmeisser: @AndyZe how would bullet being precise (and slower) by default negatively impact FCL-based collision checking?

AndyZe: Well, I actually have come around to agree with you, Simon. Since FCL defaults to "detailed meshes", then this setting makes sense:

`use_detailed_mesh_(true)`

CUN-bjy: we need to discuss this with maintainers. @rhaschke @v4hn

PR #2839, opened 2021-08-25
Adds missing deps to the package.xml template

rickstaa:
This commit adds the 'joint_trajectory_controller' and 'gazebo_ros_control' dependencies to the
`moveit_setup_assistant` package.xml template. These dependencies are needed for the gazebo simulation.


rickstaa:
@gavanderhoorn That makes sense, I will also keep them out of the new Noetic panda_moveit_config branch.

Issue #683, opened 2017-11-12
self-collision padding not configurable


simonschmeisser:
In https://github.com/ros-planning/moveit_core/pull/168 there was a discussion about adding self-collision paddings between ie the actual robot and fixed parts of the scene. We ran into some cases where this would be really useful/necessary (UR5, yumi) due to imprecise movements. It wouldn't hurt with other robots as well as people would be more relaxed not seeing it go by very closely to ie a edge of a table. I guess this could also be achieved using oversized collision geometries but that feels a bit hackish.

See my unanswered question here 
https://answers.ros.org/question/274892/padding-between-robot-and-fixed-environment/ as well

So is there any interest in a PR for a separate padding category for self-collisions?

bmagyar:
Big thumbs up.

 I had this with humanoids a few years back. The solution was oversized
meshes which was a big hack. If I remember well arm_navigation used to
support this.

davetcoleman:
This would be a great contribution, but you need to be careful where that padding is applied throughout the code base. In some cases you do not want to use it. For example, the robot's current state might not be in collision unless you are using the padding. But if MoveIt! thinks the current state is in collision, it will not be able to plan (except it does have a "wiggle joints" algorithm that will try to find a nearby state that is not in collision).

v4hn:
> So is there any interest in a PR for a separate padding category for self-collisions?

YES!

sjhansen3:
Is there any update on this? I dont understand the following based on the several posts I have read:
 
1. Is there a way to add padding to the links of the robot?
2. Is there a way to add padding to scene objects?
3. Is there a way to add padding to anything with our without YAML files?

[(https://github.com/ros-planning/moveit_ros/issues/341)]
seems to indicate that it might be possible to add padding, but Its still unclear. I tried a few different name spaces to no avail.

Sorry for my confusion, but after reading every ROS answer on this topic I am thoughrouly confused :/

Bottom line: What features regarding padding do and do not exist currently?

Thanks!

jayalatn:
Is someone working on self collision padding atm or is it already added ?

v4hn:
> Is someone working on self collision padding atm or is it already added?

[peci1 spent a lot of time fixing issues with the underlying `geometric_shapes` package](https://github.com/ros-planning/geometric_shapes/pulls?q=is%3Apr+author%3Apeci1), but to the best of my knowledge nobody worked on better higher-level integration in MoveIt.

jayalatn:
[...]

PR #2840, opened 2021-08-25
Added option to use padded robot for self-collision checking

grejj:
### Description

There has been some discussion related to adding configurable padding options when doing self-collision checking #683. As of now, I believe configurable link padding values are only applied to collision checking of the robot with the environment / world, not for self-collision checks.

For my application, instead of providing separate configurable padding values each for self-collision checks and environment collision checks, I just implemented a parameter to enable/disable using the same configured padding values that are used environment collision checks for self-collision checks as well:

`/move_group/use_padding_self_collisions true/false`

```xml
<node name="move_group" pkg="moveit_ros_move_group" type="move_group" respawn="false" output="screen">
    <param name="allow_trajectory_execution" value="true"/>
    <param name="max_safe_path_cost" value="1"/>
    <param name="jiggle_fraction" value="0.05" />
    <param name="start_state_max_dt" value="0.5" />
    <param name="max_sampling_attempts" value="100" />
    <param name="capabilities" value=""/>
    <param name="use_padding_self_collisions" value="true"/>
    <param name="disable_capabilities" value="move_group/ClearOctomapService"/>
</node>
```

May be useful for others as well. This functionality defaults to false/off if left unspecified and the current behavior is used where no padding values are used for self-collision checks. However if set to true and enabled, the padding values used for environment collision checks are also used for self-collision checks:

padding.yaml
```yaml
default_robot_padding: 0.05
default_robot_scale: 1.0
default_robot_link_padding: {   "base_link" : 0.05,
                                                   ......
                                              }
```

**Your environment**

- ROS Distro: Melodic
- OS Version: e.g. Ubuntu 18.04
- Source or Binary build? Source
- If source, which branch? Master

### DIFF
--- a/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
+++ b/moveit_core/planning_scene/include/moveit/planning_scene/planning_scene.h
@@ -1047,5 +1047,11 @@ private:
 
   // a map of object types
   std::unique_ptr<ObjectTypeMap> object_types_;
+
+  // used to specify if padded version of robot should also be used for self-collisions - disabled by default 
+  bool use_padding_self_collisions_;
+  static const std::string USE_PADDING_SELF_COLLISIONS_PARAM_NAME;
+
+  ros::NodeHandle nh_;
 };
 }  // namespace planning_scene
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -54,6 +54,7 @@ namespace planning_scene
 {
 const std::string PlanningScene::OCTOMAP_NS = "<octomap>";
 const std::string PlanningScene::DEFAULT_SCENE_NAME = "(noname)";
+const std::string PlanningScene::USE_PADDING_SELF_COLLISIONS_PARAM_NAME = "use_padding_self_collisions";
 
 const std::string LOGNAME = "planning_scene";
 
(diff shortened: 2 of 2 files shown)


simonschmeisser:
Thanks for your contribution and sorry no one looked at it yet!

I would find it more logical to have all padding related information in the `padding.yaml`, would that also be an option for you?

grejj:
Are you saying that rather than having flag to specify if the same padding values should be used for self-collisions in the launch file, it is better to have separate padding values for the environment/world and for the robot/self-collisions all specified in the padding.yaml?

simonschmeisser:
I was just suggesting to move your new parameter to `padding.yaml` so that it's all in one place

PeihengLu:
Hi! Changing `nh_('~')` to `nh_('robot_description_planning')` will enable you to enable and disable it with a parameter in padding.yaml.
So padding.yaml would look like
```
default_robot_padding: 0.025
default_robot_scale: 1.0
default_robot_link_padding: { "link_1": 2, 
                              "link_6": 10
                            }
use_padding_self_collisions: true
```
Other than that it's perfect! Even though for whatever reasons this is not approved, this helped us a lot in our project. Thank you very much!

PeihengLu:
I can confirm that this works in ubuntu 20.04 with ros noetic and lastest version of moveit_core

georgiablanco:
Hi All, is there any update on this? We are very keen on having the ability to configure padding on self collisions.

PR #2850, opened 2021-08-27
Remove unnecessary subframes map copy

felixvd:
As noted in https://github.com/ros-planning/moveit/pull/2816, I left a quick-and-dirty solution in #2037 . This commit was supposed to fix that, but it has a bug.

It's probably simple, so I'm putting this up for someone to make me look stupid.


felixvd:
Thanks! I knew it was something simple :)

So the map copy still happens after all, but at least it is avoided when only the object pose is updated. Happy with this as is, feel free to merge.

rhaschke (approved):

PR #2852, opened 2021-08-27
Fix pose-not-set-bug

mechwiz:
### Description

In #2816 , there was a fix for backwards compatibility for specifying a pose for a single collision shape. In that fix, I think there may be a bug. There is a check [here](https://github.com/ros-planning/moveit/blob/master/moveit_core/planning_scene/src/planning_scene.cpp#L1721) to determine if an object's pose has been set:
`if (object_pose.isApprox(Eigen::Isometry3d::Identity()))`

I think this check is problematic because if the pose is not set, all pose fields are zero. Checking if the pose is Identity assumes the unit quaternion which means that the "w" component is equal to 1. The way the code is currently, even if the object pose is set to identity, the shape pose will be used instead (which may not be aligned the same way as the object pose intended). This PR fixes this by explicitly checking that all fields of the object pose are zero.

### Checklist

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1718,9 +1718,8 @@ bool PlanningScene::shapesAndPosesFromCollisionObjectMessage(const moveit_msgs::
 
   bool switch_object_pose_and_shape_pose = false;
   if (num_shapes == 1)
-    if (object.pose.position.x == 0 && object.pose.position.y == 0 &&
-        object.pose.position.z == 0 && object.pose.orientation.x == 0 &&
-        object.pose.orientation.y == 0 && object.pose.orientation.z == 0 &&
+    if (object.pose.position.x == 0 && object.pose.position.y == 0 && object.pose.position.z == 0 &&
+        object.pose.orientation.x == 0 && object.pose.orientation.y == 0 && object.pose.orientation.z == 0 &&
         object.pose.orientation.w == 0)
     {
       switch_object_pose_and_shape_pose = true;  // If the object pose is not set but the shape pose is,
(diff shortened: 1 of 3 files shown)


moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1718,9 +1715,11 @@ bool PlanningScene::shapesAndPosesFromCollisionObjectMessage(const moveit_msgs::
 
   bool switch_object_pose_and_shape_pose = false;
   if (num_shapes == 1)
-    if (object_pose.isApprox(Eigen::Isometry3d::Identity()))
+    if (moveit::core::isEmpty(object.pose))
```
felixvd, line 1718: I had just accepted that identity object poses will always be overwritten by shape poses, but it's true that you might want to e.g. shift your mesh without changing the object pose itself. Nice!

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -1556,10 +1556,7 @@ bool PlanningScene::processAttachedCollisionObjectMsg(const moveit_msgs::Attache
         const moveit::core::AttachedBody* ab = robot_state_->getAttachedBody(object.object.id);
 
         // Allow overriding the body's pose if provided, otherwise keep the old one
-        if (object.object.pose.position.x == 0 && object.object.pose.position.y == 0 &&
-            object.object.pose.position.z == 0 && object.object.pose.orientation.x == 0 &&
-            object.object.pose.orientation.y == 0 && object.object.pose.orientation.z == 0 &&
-            object.object.pose.orientation.w == 0)
+        if (moveit::core::isEmpty(object.object.pose))
```
felixvd, line 1559: This looks a lot prettier now.

felixvd (approved):
Great catch, thanks!

Issue #1980, opened 2020-03-26
Representation of Attached vs Non-Attached objects in planning_scene


tylerjw:
In the meeting today we talked about how objects are represented in the planning scene.  This relates to the issue I'm trying to find a solution for (https://github.com/ros-planning/moveit/issues/1974) with subframes.

If the planning scene represented every object in the planning scene in a tree structure where an attached one was wrt the link it was attached to and unattached ones were wrt world it would mahiske the code that calculates transforms to arbitrary objects clearer.

The current structure is based around optimizing allocating all the space needed for run-time fixed length arrays of transforms (robot_state) separately from run time dynamic lengths (collision objects).  One way I see of combining these in a nice way would be to handle all memory allocation for planning scenes in a memory arena (protobuf has this feature implemented itself: https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.arena).  Another thing it is called is a memory pool (boost library: https://theboostcpplibraries.com/boost.pool).  If we go this way I'd suggest we use the boost library.  This would have the added benefit of allowing users to tune the tradeoff between memory allocation and run-time performance for their application and how much memory they expect to need for each planning_scene and how many they expect to need in a given program.

If we transitioned to a memory pool for allocation of the internal state of the planning scene we could have a simpler, more intuitive internal structure of it that  would be easier to use.

@felixvd I think it was you that mentioned another library that had a nicer internal representation of the collision scene.  Is that something you could post a link to?

felixvd:
I don't know if the *internal* representation is nicer, since I only skimmed the code[1], but I liked [Tesseract](https://github.com/ros-industrial-consortium/tesseract)'s message format for an [object in the scene](https://github.com/ros-industrial-consortium/tesseract_ros/blob/master/tesseract_ros/tesseract_msgs/msg/Link.msg). It separates visual and collision geometry cleanly and uses the same "Geometry" object for each. Apparently it doesn't even differentiate between a robot link and a link in the scene, which is also nice.

The [geometry class](https://github.com/ros-industrial-consortium/tesseract/blob/master/tesseract/tesseract_geometry/include/tesseract_geometry/geometry.h) looks clean and I noticed that there's [code for a scene graph](https://github.com/ros-industrial-consortium/tesseract/blob/master/tesseract/tesseract_scene_graph/src/graph.cpp) that may be useful for [this project](https://github.com/ros-planning/moveit/issues/202).

[...]

PR #2854, opened 2021-08-28
Add visual geometry mesh URL

### LINKED ISSUE #2284: Add visual representation to CollisionObjects
Currently, `CollisionObject`s are purely collision shapes, which are usually primitive shapes for performance reasons. However, we don't want our scenes to look like Minecraft, so I want to add a visual representation to objects in the scene, so that we can display complex geometries like [this](https://user-images.githubusercontent.com/4535737/91532771-c37f3680-e949-11ea-8294-6c1ae6c48cc2.png) without calculating the collisions for every triangle. The robot and the environment objects already have this feature, but not the `CollisionObject`s that we add to the PlanningScene.

I would add the fields `visual_geometry_mesh` and `visual_geometry_pose` fields to the `CollisionObject` message, the `world` object and the `AttachedBody`. This is the simplest way and should be straight forward to implement. The only downside is that it might affect performance (visual mesh is kept in memory and copied with the `world`/`AttachedBody` objects) and that only one mesh can be used for visualization.

Regarding the restriction to a single mesh:  
I think this is fine. If we wanted to allow more visual geometries, this would be a much bigger job and break API, so I don't want to do it. It wouldn't be worth it. (outline: split [CollisionObject](http://docs.ros.org/melodic/api/moveit_msgs/html/msg/CollisionObject.html) message into visual and collision geometry ([like](https://github.com/ros-industrial-consortium/tesseract_ros/blob/master/tesseract_ros/tesseract_msgs/msg/VisualGeometry.msg) [this](https://github.com/ros-industrial-consortium/tesseract_ros/blob/master/tesseract_ros/tesseract_msgs/msg/CollisionGeometry.msg)), and in `world`/`AttachedBody`,  split `shapes_` and `shape_poses_` vectors into `visual` and `collision`).

Regarding performance:  
I *think* @rhaschke proposed (but I couldn't find where) to only pass in a mesh file location. That requires the mesh to be saved on disc and to be cached in `ros_visualization`. I would prefer not to implement it this way if it isn't necessary, since it might be premature optimization. Thoughts?

I looked at the visualization code, and using the `VisualVisible`/`CollisionVisible` flag to visualize the objects in `planning_scene_render.cpp` and `robot_state_visualization.cpp` won't be difficult. But I would like an opinion on the mesh vs mesh file consideration before I start. Thanks!

felixvd:
### Description

Remember [this lightning talk](https://www.youtube.com/watch?v=Q-RIQi5ru3o)?

Screenshot from 2021-08-28 13-12-23

I finally got around to implementing this. CollisionObjects can now have a visual geometry, and the pick-place demo now picks up a beer bottle from a table:

https://user-images.githubusercontent.com/4535737/131205939-a8fcb420-2104-4865-9227-cd85b3f92a58.mp4

The collision geometry is as simple as before. See the new GUI fields on the left:

Screenshot from 2021-08-28 13-28-09

There are some limitations about this PR:

- All meshes need to have the same scale, because they are scaled by the parameter in the Rviz display. This is because I initially thought that adding `visual_geometry_mesh_scaling_factor` to each `AttachedBody` would make the constructor too long, since it is already a little ridiculous. But in retrospect, I should have done it. Or added a `VisualGeometry` object that both `world.cpp` and `AttachedBody` objects use. But I'm low on time, so it is how it is for now.

- There are no tests, even though it's easy to lose the geometry during copying (thanks for finding the last bug @simonschmeisser ). I also won't have the time to write any ;(

I would merge this because the functionality is useful, but I would prefer the scale to be defined per object. I don't know when I can make this change though, so if anyone feels so inclined, please be my guest. It should be straightforward to follow the changes in this PR and add a per-object scaling factor.

Some other notes: 

- I split PlanningSceneRender into a visual and collision node, like the rviz Robot. I originally tried to avoid adding another scene node, but changed my mind after trying it. This is definitely the better way.

- I am increasingly convinced that the separation into robot and scene is (and creates) cruft and should go. The conversion from `world` objects to `AttachedBody` objects is probably where it is the most obvious. Unifying them would go a long way to de-cluttering the PlanningScene code base. I would refer to [Tesseract's messages](https://github.com/ros-industrial-consortium/tesseract_ros/tree/master/tesseract_msgs/msg) (and maybe environment) to improve on this. No need to reinvent the wheel. This has come up before (e.g. #1980 ) and relates to the scene graph discussions.

- The tutorial demo sadly doesn't use [this prettier table](https://sketchfab.com/3d-models/antique-table-f01054c29829452a9f87437de57a3f6f). Maybe @simonschmeisser will add color for the visual mesh and tackle this some day :) 

- @simonschmeisser I boldly added your name to a TODO in the code that you seemed most knowledgeable about.

Fixes #2284. Depends on a PR in moveit_msgs and is required for the PR to moveit_tutorials.

### Checklist
- [X] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)
- [X] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes
- [X] Include a screenshot if changing a GUI


moveit_core/collision_detection/include/moveit/collision_detection/world.h
```diff
@@ -91,6 +91,13 @@ class World
      *  This frame is returned when getTransform() is called with the object's name. */
     Eigen::Isometry3d pose_;
 
+    /** \brief (Optional) The URL of the mesh used for the visual representation of this object.
+     * Only used for display in Rviz, not for collision checking. */
+    std::string visual_geometry_mesh_url_;
+
+    /** \brief The pose of the visual geometry (mesh). */
+    Eigen::Isometry3d visual_geometry_pose_;
```
simonschmeisser, line 99:
```suggestion
    Eigen::Isometry3d visual_geometry_pose_ = Eigen::Isometry3d::Identity();
```
then you can skip that later

moveit_core/collision_detection/src/world.cpp
```diff
@@ -87,6 +87,8 @@ void World::addToObject(const std::string& object_id, const Eigen::Isometry3d& p
     obj = std::make_shared<Object>(object_id);
     action |= CREATE;
     obj->pose_ = pose;
+    obj->visual_geometry_mesh_url_ = "";
+    obj->visual_geometry_pose_ = Eigen::Isometry3d::Identity();
```
simonschmeisser, line 91:
```suggestion
```
we did that in the header already, see above

moveit_core/collision_detection/src/world.cpp
```diff
@@ -260,6 +262,8 @@ bool World::setObjectPose(const std::string& object_id, const Eigen::Isometry3d&
   {
     obj = std::make_shared<Object>(object_id);
     action = CREATE;
+    obj->visual_geometry_mesh_url_ = "";
+    obj->visual_geometry_pose_ = Eigen::Isometry3d::Identity();
```
simonschmeisser, line 266:
```suggestion
```
same here, strings are always empty when created

simonschmeisser (changes requested):
some small first remarks while looking at the code as reference for another PR

moveit_core/collision_detection/include/moveit/collision_detection/world.h
simonschmeisser: If this creates issues with `Eigen::Isometry3d` being forward declared (but I don't think it is ...) you could also set the value in the constructor and move the constructor to the cpp file (where it actually belongs ...)

simonschmeisser:
Could you please do a rebase to include #2855

simonschmeisser:
Requires https://github.com/felixvd/moveit_msgs/pull/1 (or https://github.com/ros-planning/moveit_msgs/pull/129 ) as well

Open Todos:

- [ ] Less noise
- [ ] There is some jumping around when attaching/detaching objects, something is strange with transformations
- [ ] Decide when to enable/disable `mesh_use_embedded_material`

simonschmeisser:
grafik

So colored or textured objects are still not that impressive but it works. Tinting is to strong but I don't understand how to influence that yet.

Let the reviews begin! I will squash some of the commits before merging.

felixvd:
> So colored or textured objects are still not that impressive

Speak for yourself!!! I think they look great. Thanks a lot for pursuing this PR.

ahmadabouelainein:
Has this feature been discarded or is still pending?

PR #2865, opened 2021-09-04
MPD: do not save/restore warehouse parameters

v4hn:
If we reload these values from the config, setting the ROS parameters is much less useful. The correct db should be set in the parameters instead of the RViz config!
At least the *type* of warehouse_ros plugin (mongo or sqlite) cannot even be selected in the display, so you will probably need to meddle with the parameters anyway if you want to connect to a different db.

search for parameters warehouse_host/port because they are usually set at the top level, but you might want to set them differently for different move_groups.

I noticed the values in rviz are not updated when following [the new `warehouse_ros_sqlite` tutorial](https://github.com/ros-planning/moveit_tutorials/pull/658), but you would very much expect them to be!

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
@@ -228,6 +228,16 @@ void MotionPlanningDisplay::onInitialize()
 
   rviz::WindowManagerInterface* window_context = context_->getWindowManager();
   frame_ = new MotionPlanningFrame(this, context_, window_context ? window_context->getParentWindow() : nullptr);
+
+  ros::NodeHandle move_group_nh(ros::names::append(getMoveGroupNS(), "move_group"));
+  std::string param_name;
+  std::string host_param;
+  int port;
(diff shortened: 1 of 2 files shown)


simonschmeisser:
this probably needs to be added to `Migrations`? Why did you merge without review ... I know it's annoying to wait that long without feedback ...

v4hn:
> I know it's annoying to wait that long without feedback ...

Yes, I know this from both sides. I just noticed it's still stuck when reviewing local branches. As nobody had hard feelings about this (and I wouldn't expect anyone to) I merged it as [trivial change](https://moveit.ros.org/documentation/contributing/maintainer_pr/).

> this probably needs to be added to Migrations?

I don't think this justifies an entry in `MIGRATIONS.md`. Firstly, we should keep the size of this list reasonably small for people to actually look through it. Secondly, I cannot think of a use-case where the behavior would deteriorate because of this change. If you setup a warehouse the correct parameter is exported in the parameter, so if anything, the patch improves the use-case to take out the default value from Rviz that might be invalid in your particular setup. And thirdly, it fixes a trivial issue with the new sqlite warehouse plugin.

PR #2868, opened 2021-09-08
Fix JointConstraintSampler

vinnamkim:
### Description
 - For continous revolute joints, `constraint_samplers::JointConstraintSampler::JointInfo`'s bounds should not be limited by `moveit::core::VariableBounds`.
 - For example, if the desired position is `pi` and the tolerance bound is `[-0.3, 0.3]`, `JointConstraintSampler` should samples between `[pi - 0.3, pi]` or `[-pi, -pi + 0.3]`.
 - However, the current implementation samples only from `[pi - 0.3, pi]`.

Signed-off-by: Vinnam Kim <vinnam.kim@makinarocks.ai>

### Checklist

### DIFF
--- a/moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
+++ b/moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
@@ -67,7 +67,7 @@ public:
    *
    */
   JointConstraintSampler(const planning_scene::PlanningSceneConstPtr& scene, const std::string& group_name)
-    : ConstraintSampler(scene, group_name)
+    : ConstraintSampler(scene, group_name), random_number_generator_(new random_numbers::RandomNumberGenerator())
   {
   }
   /**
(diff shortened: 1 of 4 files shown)


simonschmeisser:
Thanks a lot for your contribution and sorry that nobody has commented on it yet. Also providing a test case and adding determinism to the sampling is highly appreciated. Very professional!

Now to the content: I'm no expert on revolute joints but I would rather expect the current behavior than what you implement in your PR as  pi+0.3 is quite different to -pi+0.3 in terms of distance traveled, cables that might be attached etc. But then again, I'm no expert on most of this

vinnamkim:
> Thanks a lot for your contribution and sorry that nobody has commented on it yet. Also providing a test case and adding determinism to the sampling is highly appreciated. Very professional!
> 
> Now to the content: I'm no expert on revolute joints but I would rather expect the current behavior than what you implement in your PR as pi+0.3 is quite different to -pi+0.3 in terms of distance traveled, cables that might be attached etc. But then again, I'm no expert on most of this

I think that there is no problem on the distance of `RevoluteJoint` + `Continous`. It seems that there are signed shortest distance implementations for `RevoluteJoint` + `Continous`. What I've found are as follows.

1.
https://github.com/ros-planning/moveit/blob/daa87acbf7e9113b07ed8a5bcd69ed9104906cfe/moveit_core/robot_model/src/revolute_joint_model.cpp#L150-L159

2.
https://github.com/ros-planning/moveit/blob/daa87acbf7e9113b07ed8a5bcd69ed9104906cfe/moveit_core/kinematic_constraints/src/kinematic_constraint.cpp#L272-L283

PR #2871, opened 2021-09-09
Fixed Bullet collision checker not taking ACM defaults into account.

werner291:
### Description

The `acmCheck` method in `bullet_utils.h` directly looked up the relevant entry in the `AllowedCollisionMatrix`, ignoring any default values. This PR replaced the method call with `getAllowedCollision`, which appears to be the correct method (please review).

### Checklist

### DIFF
--- a/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/bullet_utils.h
+++ b/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/bullet_utils.h
@@ -65,7 +65,7 @@ inline bool acmCheck(const std::string& body_1, const std::string& body_2,
 
   if (acm != nullptr)
   {
-    if (acm->getEntry(body_1, body_2, allowed_type))
+    if (acm->getAllowedCollision(body_1, body_2, allowed_type))
     {
       if (allowed_type == collision_detection::AllowedCollision::Type::NEVER)
       {


werner291:
Looking deeper into it, `PlanningScene::setPlanningSceneDiffMsg` seems to ignore default entries as well.

https://github.com/ros-planning/moveit/blob/ffe4697696147f0e4db116a19f70cde7c5f5d074/moveit_core/planning_scene/src/planning_scene.cpp#L1230 

Is that intended?

werner291:
Update: my code still doesn't seem to work despite this fix. I'll keep working on it.

werner291:
Nope, it was actually just due to my code linking against an old version of MoveIt. After updating the fix works, and the default ACM entries are followed correctly!

More test cases definitely wouldn't be a bad idea, though.

Please review for merging?

v4hn (approved):

v4hn:
Trivial fix for a problem that should never have existed! Thanks @werner291 :medal_sports: 

> More test cases definitely wouldn't be a bad idea, though.

It would be great if you could add a test for this in a followup PR. Sadly, if users do not provide tests, nobody will ever write them. Without them, it's of course difficult to ensure your use-case does not break again at some point.

simonschmeisser:
Also this feature does not seem to be used anywhere in MoveIt so it's not that surprising its broken. What's your use-case if I may ask?

v4hn:
True, though of course the user can fill these fields in messages. I would not expect MoveIt "to use" this anywhere else, unless a moveit_resources model makes use of it for tests.
But then, I'm really surprised to see srdfdom/moveit_setup_assistant/robot_model_loader do not have a way to set these entries!
It would be perfectly reasonable to support `<disable_collisions link1="xyz" link2="*"/>`, e.g. for objects that are only added for visual effect (or reference) or must never collide with anything no matter what.

simonschmeisser:
Since `noetic` you can simply omit the `<collision>` part of the URDF for that. Or use markers. But I admit that we tend to have a visual sphere at the TCP and a collision sphere somewhere inside the gripper where it can be disabled with `<disable_collisions>` conveniently ...

werner291:
> Also this feature does not seem to be used anywhere in MoveIt so it's not that surprising its broken. What's your use-case if I may ask?

My use case is a quadcopter (Yes, with a floating joint! That's why I'm finding all the bugs...) with an arm which must navigate around a tree, whereby the leaves are not counted as hard "do not ever touch or else..." obstacles, but rather as a fuzzy obstacle that should generally be avoided, but can be traversed when needed, but only in certain ways. Hence, I wanted to see them in Rviz, and I may need to check collisions at some point, but I needed to disable collisions during the global path planning since otherwise the task becomes literally impossible.

v4hn:
Nice use case! If you're allowed, please share some videos on Discord or discourse.ros.org once you have something to show :-)

werner291:
There's a Discord server?

v4hn:
> There's a Discord server?

https://moveit.ros.org/support/ says:

> Join the [MoveIt Discord channel](https://discord.com/invite/RrySut8) to view or post questions

If you can think of a better way to advertise the server, please send a pull-request for the website :)

PR #2872, opened 2021-09-10
Fix #2811

### LINKED ISSUE #2811: ompl_interface::ConstrainedGoalSampler::sampleUsingConstraintSampler doesn't appear to sample?
Hello,

I was working on trying to implement a custom `constraint_samplers::ConstraintSampler`, but I noticed that only the `project` method gets called, never the `sample` method.

The state I get as input always appears to be the same (I think it's the same as the starting state?) when I print it out.

When reading through https://github.com/ros-planning/moveit/blob/master/moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp#L90, I also don't really see any source of randomness in there... it appears to just be projecting the same state over and over.

Is this the intended behavior?

werner291:
This is a PR for #2811

### Description

### Checklist

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp
@@ -129,7 +129,7 @@ bool ompl_interface::ConstrainedGoalSampler::sampleUsingConstraintSampler(const 
                     verbose);
       constraint_sampler_->setGroupStateValidityCallback(gsvcf);
 
-      if (constraint_sampler_->project(work_state_, planning_context_->getMaximumStateSamplingAttempts()))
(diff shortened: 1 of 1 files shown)


felixvd (approved):
This looks a lot clearer to me, and `project` should really not do the same thing as `sample`, but I never touch this code so I am just adding this as a vote of support.

v4hn:
Our tests like it and this fixes quite a crucial aspect of method semantics. I'll just take your approval as a +1 @felixvd and merge. :cat2: 

Thanks for the effort @werner291 !

rhaschke:
Please use more informative PR descriptions, e.g. here: `ConstrainedGoalSampler: actually sample()`
The description usually becomes the final commit message.

v4hn:
My bad, I should have adjusted the commit message of course.

PR #2876, opened 2021-09-15
PlanningSceneDisplay: always update the main scene node's pose

rhaschke:
This fixes https://github.com/ros-planning/moveit_task_constructor/issues/291.
We need to update the pose of the PlanningSceneDisplay's scene node in each update() cycle.
So far, it was only updated if the PlanningSceneMonitor received any update.

The following video illustrates the new (left) vs. the old=released behaviour (right):

https://user-images.githubusercontent.com/5376030/133408681-e76f08c1-79cc-49e1-beac-6e046b7f8bc0.mp4

Without this PR, the planning scene is only updated once when it gets initialized (or when joint updates are received, which is not the case here).

### DIFF
--- a/moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
+++ b/moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp
@@ -188,7 +188,7 @@ void PlanningSceneDisplay::onInitialize()
 {
   Display::onInitialize();
 
-  // the scene node that contains everything
+  // the scene node that contains everything and is located at the planning frame
   planning_scene_node_ = scene_node_->createChildSceneNode();
 
   if (robot_category_)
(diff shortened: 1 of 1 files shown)


rhaschke:
The right video also illustrates an issue @captain-yoshi has pointed out in https://github.com/ros-planning/moveit_resources/issues/89#issuecomment-918418221. As the left video shows, this issue is fixed in the master branch of MoveIt. Obviously, we should schedule a new release!

simonschmeisser:
I wonder  how are other changes to static transforms handled? Do they trigger a PlanningSceneUpdate?

wolfv:
Thanks @v4hn for the notification. We're looking into it.

rhaschke:
> I wonder how are other changes to static transforms handled? Do they trigger a PlanningSceneUpdate?

The PlanningSceneMonitor updates TF transforms as part of [`updateFrameTransforms()`](https://github.com/ros-planning/moveit/blob/b6a928846510f4e44e78fee1eadef4745c15215e/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L1357). The actual functionality is implemented here:
https://github.com/ros-planning/moveit/blob/b6a928846510f4e44e78fee1eadef4745c15215e/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L1320-L1335

Issue #2843, opened 2021-08-26
Add test for (#2824)


felixvd:
https://github.com/ros-planning/moveit/pull/2824 fixed a bug with the pilz industrial motion planner, where attached objects were not considered during plans. 

It would be very helpful if this bug was covered by a test, for example with a motion like in [this tutorial](https://ros-planning.github.io/moveit_tutorials/doc/move_group_interface/move_group_interface_tutorial.html):

![image](https://user-images.githubusercontent.com/4535737/130881794-323e673f-9f6b-4b09-b5c1-aeece179802b.png)

This is a good issue for someone new. You only need to add a test, e.g. in one of [these files](https://github.com/ros-planning/moveit/tree/master/moveit_ros/planning_interface/test) or a new one, which fails before the PR and succeeds after.

You can run your test with `catkin run_tests -iv` or `rostest your_package your_test_file` as described [here](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html). If you are a new contributor, feel free to ask for help in this thread.

saurabh-1-1:
hey felix i am new to ros learned about moveit through sense think and act podcast really intrested in doing some open source work can you guide me how can i help solve this issue.

felixvd:
Hi,

I would suggest to start by following the tutorials, asking on ROS Answers if you have trouble with anything, then move on to read and understand the code as much as possible, try to build it from source, then write some new tests and check if they succeed.

gquey:
Hello, is this issue still open?  I am looking to make my first contribution and this looks like a good place to start. Thanks

rhaschke:
As you can see from the latest linked PR, this issue was already fixed in #2878.

PR #2878, opened 2021-09-19
Add test case for #2824

cambel:
Test case for planning with collision objects that collide with the environment with Pilz planners (#2824). Related to #2843

### Checklist


moveit_ros/planning_interface/test/python_move_group_planning.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
```
v4hn, line 1: please add a license header.

moveit_ros/planning_interface/test/python_move_group_planning.py
```diff
@@ -0,0 +1,79 @@
...
+
```
v4hn, line 2: The file tests the pilz_planner, so I would prefer to keep it [with the other pilz tests](https://github.com/ros-planning/moveit/tree/master/moveit_planners/pilz_industrial_motion_planner/test), in contrast to Felix's original post.

moveit_ros/planning_interface/test/python_move_group_planning.py
```diff
@@ -0,0 +1,79 @@
...
+            touch_links=["panda_rightfinger", "panda_leftfinger"],
+        )
+
+        # Plan a motion where the attached object 'box2' collides with the obstacle 'box1'
+        target_pose = self.group.get_current_pose(tcp_link)
+        target_pose.pose.position.y += 0.1
+
+        # Set planner to be Pilz's Linear Planner
+        self.group.set_planning_pipeline_id("pilz_industrial_motion_planner")
+        self.group.set_planner_id("LIN")
```
v4hn, line 58: I would prefer to keep this test limited to `PTP` unless the code paths for checking collisions are strictly separated between the two planners. `LIN` relies on point-wise IK requests that add an unnecessary complexity here and might make the test flaky for unrelated reasons.

moveit_ros/planning_interface/test/python_move_group_planning.test
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
```
v4hn, line 1: Please add the test file to `CMakeLists.txt` via `add_rostest`.

v4hn (commented):
Thanks for the effort! Please have a look at the inline feedback.

moveit_ros/planning_interface/test/python_move_group_planning.py
cambel: I guess, in a sense they are separated. `PTP` only checks collision of the target pose if the target pose is a Cartesian pose, otherwise it does not check collisions at all. I think `LIN` checks collisions for intermediate points. 
Now, neither checks collision of attached objects, as the only collision checks done in the Pilz planners are `checkSelfCollision`. So, after all, just testing `PTP` is fine I suppose.

cambel:
Thanks for the comments. I have updated the PR.

v4hn:
CI failure:

> Test node [moveit_ros_planning_interface/python_move_group_planning.py] does not exist or is not executable

moveit_ros/planning_interface/test/python_move_group_planning.py
v4hn: >  PTP only checks collision of the target pose if the target pose is a Cartesian pose, otherwise it does not check collisions at all. I think LIN checks collisions for intermediate points.
Now, neither checks collision of attached objects, as the only collision checks done in the Pilz planners are checkSelfCollision

Wait, what? I was only peripherally involved with #2824, but I thought that's exactly what it was meant to solve?
Also the test case you propose here relies on world collision checking. 

Does all of this still rely on MoveIt's post-planning check of the entire trajectory? :open_mouth:

cambel: No, right now, it does rely only on MoveIt's post-planning check. The fix in #2824 was to make the post-planning work properly. 

Now, if that post-planning is only optional, then I guess it is okay to add the same validation of the entire trajectory **here**, where the internal planner has returned a solution but before returning to the planning_pipeline 

https://github.com/ros-planning/moveit/blob/8c499f5d3dea8a75c6219be765d5a824e34d9056/moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator.cpp#L305-L317

It would work for every type of planner LIN, PTP, and CIRC.

I think this approach is acceptable as Pilz planners basically assume the existence of one possible solution, so checking that solution before returning should be safe and enough.

moveit_planners/pilz_industrial_motion_planner/test/python_move_group_planning.py
```diff
@@ -0,0 +1,117 @@
...
+        # self.assertEqual(error_code.val, MoveItErrorCodes.INVALID_MOTION_PLAN)
+
+        # Set planner to be Pilz's Point-To-Point Planner
+        self.group.set_planning_pipeline_id("pilz_industrial_motion_planner")
+        self.group.set_planner_id("PTP")
+        self.group.set_pose_target(target_pose)
+        success, plan, time, error_code = self.group.plan()
+
+        # Planning should fail
+        self.assertEqual(error_code.val, MoveItErrorCodes.INVALID_MOTION_PLAN)
```
jschleicher, line 110: maybe also test the return value for the sake of completeness?
```suggestion
        self.assertFalse(success)
        self.assertEqual(error_code.val, MoveItErrorCodes.INVALID_MOTION_PLAN)
```

jschleicher (approved):
Thank you very much for the follow-up! See another inline nitpick.

Merging master to (hopefully) fix cross-platform CI.

moveit_ros/planning_interface/test/python_move_group_planning.py
jschleicher: Without looking at the code, the idea was, that IK solvers use the collision-check-callback to choose a suitable solution that is closest to the seed (previous point). It should take into account attached objects, of course (otherwise it should be fixed). In contrast PTP only does that check once for the goal, if it is given in cartesian space.
Relying on the pipeline for collision checking was mainly a decision to not duplicate computation effort and speed-up planning, if I remember correctly.

moveit_planners/pilz_industrial_motion_planner/test/python_move_group_planning.py
cambel: Sure

moveit_ros/planning_interface/test/python_move_group_planning.py
v4hn: makes sense. let's keep it that way. If users disable the optional post-planning check, they should be aware of the nature of results from their planners.

PR #2882, opened 2021-09-21
Small bug fix and speedup in TOTG

AndyZe:


moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -959,11 +959,15 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
-    for (size_t j = 0; j < num_joints; j++)
+    for (size_t j = 0; j < num_joints; ++j)
     {
       new_point[j] = waypoint->getVariablePosition(idx[j]);
-      if (p > 0 && std::abs(new_point[j] - points.back()[j]) > min_angle_change_)
+      // If any joint angle is different, it's a unique waypoint
+      if (p > 0 && fabs(new_point[j] - points.back()[j]) > min_angle_change_)
+      {
         diverse_point = true;
+        break;
```
AndyZe, line 969: short-circuit, for speed

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -959,11 +959,15 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
     Eigen::VectorXd new_point(num_joints);
     bool diverse_point = (p == 0);
 
-    for (size_t j = 0; j < num_joints; j++)
+    for (size_t j = 0; j < num_joints; ++j)
     {
       new_point[j] = waypoint->getVariablePosition(idx[j]);
-      if (p > 0 && std::abs(new_point[j] - points.back()[j]) > min_angle_change_)
+      // If any joint angle is different, it's a unique waypoint
+      if (p > 0 && fabs(new_point[j] - points.back()[j]) > min_angle_change_)
```
AndyZe, line 966: I think std::abs was converting the angles to integers. Ouch

gavanderhoorn: There's still [std::fabs(..)](https://en.cppreference.com/w/cpp/numeric/math/fabs).

griswaldbrooks: I don't think `std::abs` casts to int https://en.cppreference.com/w/cpp/numeric/math/fabs https://godbolt.org/z/vsPT7jqa8

AndyZe: @griswaldbrooks I was looking at that documentation, too. There is also this, which does implicitly convert to int.

[https://en.cppreference.com/w/cpp/numeric/math/abs](
https://en.cppreference.com/w/cpp/numeric/math/abs)

I'm not sure which version we use here. I can say, for sure, that this PR changes the behavior drastically. (Maintenance motions for our project fail when using this PR).

griswaldbrooks: It doesn't implicitly convert `double` to `int` because overload resolution specifically sees that `double std::abs(double n)` is a better match than `int std::abs(int n)`. Also, see the compiler explorer link. Did you test the behavior different with only `abs` vs `fabs` leaving your short cutting?

AndyZe: You're completely right, short-circuiting did not allow the other joint angles to be copied.

AndyZe: I still think we should keep this PR so we don't have to depend on the overload of `std::abs`, though

v4hn (approved):
Consistency makes sense.

Personally I prefer `std::abs` too, but

PR #2888, opened 2021-09-28
Improve launch files

rhaschke:
This PR
* renames launch files from *.launch.xml -> .launch
* clarifies that all planning pipelines are loaded into a namespace by doing it once in `planning_pipeline.launch`

### DIFF
--- a/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
+++ b/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
@@ -47,23 +47,22 @@
   <group ns="move_group/planning_pipelines">
 
     <!-- OMPL -->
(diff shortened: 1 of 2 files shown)


v4hn:
> clarifies that all planning pipelines are loaded into a namespace by doing it once in planning_pipeline.launch

;+1:

> renames launch files from *.launch.xml -> .launch

The old naming removes these files from `roslaunch` autocompletion (which only looks for the `.launch` suffix). With your change the completion list would be much longer for no reason. :confused:

rhaschke:
> > renames launch files from *.launch.xml -> .launch
> 
> The old naming removes these files from `roslaunch` autocompletion (which only looks for the `.launch` suffix). With your change the completion list would be much longer for no reason. confused

I see. I was wondering why you would like those files to be named .launch.xml. Thanks for the explanation. I will remove this commit.

PR #2892, opened 2021-10-03
MGI: add missing replan/look options to interface

v4hn:
This came up in https://answers.ros.org/question/387120/online-path-replanning-when-encountering-an-obstacle-moveit/.

- reordered methods because looking requires replanning
- there's no sense in wrapping methods in methods.
  just use pimpl-friend paradigm instead. Someone could
  rework all the other methods in the future.

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
+++ b/moveit_ros/planning_interface/move_group_interface/include/moveit/move_group_interface/move_group_interface.h
@@ -782,12 +782,21 @@ public:
   /** \brief Stop any trajectory execution, if one is active */
   void stop();
 
-  /** \brief Specify whether the robot is allowed to look around before moving if it determines it should (default is
-   * true) */
-  void allowLooking(bool flag);
-
   /** \brief Specify whether the robot is allowed to replan if it detects changes in the environment */
   void allowReplanning(bool flag);
+
+  /** \brief Maximum number of replanning attempts */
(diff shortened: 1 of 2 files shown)


v4hn:
@wolfv, @traversaro, @Tobias-Fischer
your windows build broke with `Access Denied`. please set up a notifier as proposed in #2881.

Tobias-Fischer:
It might be the same `boa` issue as we see in https://github.com/RoboStack/ros-noetic/pull/174

wolfv:
I opened a PR with a fix.

Some background:
The problem is that on Windows you can't really erase files that are in use. We are trying to downgrade openssl from 3.0.0 to 1.1.1j but `rosdep` is using Python which in turn opened the openssl dll's and thus we get an error in micromamba trying to remove these files that are in use. 

We should handle the removal of open files more gracefully in micromamba in the future. for the time being we can pin openssl 1.1.1 and we should not have any problems.

rhaschke (approved):

rhaschke:
I was just in the process of writing a more meaningful commit message. You didn't even mention the added functions.

v4hn:
> I was just in the process of writing a more meaningful commit message. You didn't even mention the added functions.

Ah, sorry. I was wondering why you did not merge right away. :sweat_smile: 

> You didn't even mention the added functions.

True, I thought they are characterized well enough through the title and I considered the implementation choices more important to document :cat2:

Issue #2891, opened 2021-09-30
Servo: Apply TwistStamped in frame of header


udopa:
**Is your feature request related to a problem? Please describe.**

I'm using servo for path following and want to set a flexible TCP which can be offset to the known robot TCP (i.e. end of the kinematic chain, `TCP_Robot`) in Translation and Rotation. I'll call this frame `TCP_Shifted` for better explanation. In normal planning applications this is possible, for example by using a subframe of a collision object as introduced in #1439 and following issues/PRs. It would be good if I can publish a `TwistStamped` message with a `frame_id` pointing to a subframe of a collision object or any known frame in the planning scene. The latter ones could be published by a StaticTransformBroadcaster.

The motion should be executed in the given frame `TCP_Shifted`. This implies that:
- Linear motions (translation) can be executed in any frame, as long as the motion is transformed correctly (v_y could become v_x with a rotation change of 90°). This does work already.
- Angular motions (rotation) have to be executed in the given frame because otherwise the center of rotation will be wrong, and a linear motion will be introduced because of the lever between the tcp of the movegroup and the given `TCP_Shifted`. See the image for a more hands-on explanation of the problem. Currently the motion will be transformed to the controlling frame `TCP_Robot` "correctly", i.e. w_y would become w_x similar to the linear motion case. Therefore the described problem exists. 
![image](https://user-images.githubusercontent.com/46706225/135416869-240e51a8-7c1a-4beb-a6f5-475f8374c0ce.png)
_Image: TwistStamped with `header.frame_id=TCP_Shifted`. I want the motion to be executed in "TCP_Shifted" (desired state), but it is executed in "TCP_Robot" (current state) creating a linear motion from an angular motion and therefore moving the desired TCP even when only rotating._

This issue is related to #2866. I changed to code so that I am able to use any frame known in the scene in my `TwistStamped` commands, but I am not able to change the controlled frame to be the `TCP_Shifted`. This is possible by changing the code in servo_calcs.cpp at [L492](https://github.com/ros-planning/moveit/blob/d1e1c3d27ab33978fdcc5d7f9494afaa76496adc/moveit_ros/moveit_servo/src/servo_calcs.cpp#L492):
```cpp
planning_scene_monitor::LockedPlanningSceneRO planning_scene(planning_scene_monitor_);
const auto tf_moveit_to_incoming_cmd_frame =
    current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
[...]

PR #2894, opened 2021-10-04
Servo: Enable usage of transforms published during runtime

udopa:
### Description

_Please explain the changes you made, including a reference to the related issue if applicable_
This is related to #2891 and enables sending TwistStamped to Servo with frame_ids that were not known during startup resulting in more flexibility. With this approach, the following additional frames can be used (and were tested):
1. Static TF Broadcaster with a frame attached to the endeffector link of the selected movegroup. It should however work for any published frame. This could be potentially dangerous when applying the Twist in a frame far away from the actual control frame. Maybe there is some checking needed in that case.
2. Subframe of Attached Collision Object.

@AndyZe Here is the PR as discussed 

The changes are in 3 places:
1. Servo.start()
2. Servo.calculateSingleIteration()
3. Servo.cartesianServoCalcs()

Using only getFrameTransform for getting the transform right did not work for me, this is why I still use the version with the inverse as shown here. 
```cpp
const auto tf_moveit_to_incoming_cmd_frame =
    current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
    planning_scene->getFrameTransform(cmd.header.frame_id);

translation_vector = tf_moveit_to_incoming_cmd_frame.linear() * translation_vector;
angular_vector = tf_moveit_to_incoming_cmd_frame.linear() * angular_vector;
```

The version with robot_state does not recognize the frames from the static tf broadcaster, which is why I did not use it in the end. 
```cpp
const auto tf_moveit_to_incoming_cmd_frame =
    current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
    robot_state->getframeTransform(cmd.header.frame_id);
```

Furthermore, I think this code could also be used instead of the transformations (`tf_moveit_to_robot_cmd_frame_` and `tf_moveit_to_ee_frame_`). 
As far as I can see it, these transformations are also calculated every single iteration, but the frame ids have to be defined manually before starting servo (in the servo config file).  They are also not used any further. Therefore I suppose that `tf_moveit_to_robot_cmd_frame_` and `tf_moveit_to_ee_frame_` can be completely replaced with a `tf_moveit_to_incoming_cmd_frame` and a default transformation (in case there is no frame_id declared in the cmd header). This also results in not needing to declare the frames in the config.  

**Maintainers**: What do you think about that?

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -194,10 +194,12 @@ void ServoCalcs::start()
   last_sent_command_ = initial_joint_trajectory;
 
   current_state_ = planning_scene_monitor_->getStateMonitor()->getCurrentState();
+  planning_scene_monitor_->requestPlanningSceneState();
+  planning_scene_monitor::LockedPlanningSceneRO planning_scene(planning_scene_monitor_);
   tf_moveit_to_ee_frame_ = current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
-                           current_state_->getGlobalLinkTransform(parameters_.ee_frame_name);
+                           planning_scene->getFrameTransform(parameters_.ee_frame_name);
   tf_moveit_to_robot_cmd_frame_ = current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
-                                  current_state_->getGlobalLinkTransform(parameters_.robot_link_command_frame);
+                                  planning_scene->getFrameTransform(parameters_.robot_link_command_frame);
 
   stop_requested_ = false;
   thread_ = std::thread([this] { mainCalcLoop(); });
@@ -294,17 +296,19 @@ void ServoCalcs::calculateSingleIteration()
   have_nonzero_twist_stamped_ = latest_nonzero_twist_stamped_;
   have_nonzero_joint_command_ = latest_nonzero_joint_cmd_;
 
+  planning_scene_monitor_->requestPlanningSceneState();
+  planning_scene_monitor::LockedPlanningSceneRO planning_scene(planning_scene_monitor_);
   // Get the transform from MoveIt planning frame to servoing command frame
   // Calculate this transform to ensure it is available via C++ API
   // We solve (planning_frame -> base -> robot_link_command_frame)
   // by computing (base->planning_frame)^-1 * (base->robot_link_command_frame)
   tf_moveit_to_robot_cmd_frame_ = current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
-                                  current_state_->getGlobalLinkTransform(parameters_.robot_link_command_frame);
+                                  planning_scene->getFrameTransform(parameters_.robot_link_command_frame);
 
   // Calculate the transform from MoveIt planning frame to End Effector frame
   // Calculate this transform to ensure it is available via C++ API
   tf_moveit_to_ee_frame_ = current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
-                           current_state_->getGlobalLinkTransform(parameters_.ee_frame_name);
+                           planning_scene->getFrameTransform(parameters_.ee_frame_name);
 
   have_nonzero_command_ = have_nonzero_twist_stamped_ || have_nonzero_joint_command_;
 
@@ -487,11 +491,13 @@ bool ServoCalcs::cartesianServoCalcs(geometry_msgs::TwistStamped& cmd,
     }
     else
     {
+      // Update scene to know all existing transforms
+      planning_scene_monitor::LockedPlanningSceneRO planning_scene(planning_scene_monitor_);
       // We solve (planning_frame -> base -> cmd.header.frame_id)
       // by computing (base->planning_frame)^-1 * (base->cmd.header.frame_id)
       const auto tf_moveit_to_incoming_cmd_frame =
           current_state_->getGlobalLinkTransform(parameters_.planning_frame).inverse() *
-          current_state_->getGlobalLinkTransform(cmd.header.frame_id);
+          planning_scene->getFrameTransform(cmd.header.frame_id);
 
       translation_vector = tf_moveit_to_incoming_cmd_frame.linear() * translation_vector;
       angular_vector = tf_moveit_to_incoming_cmd_frame.linear() * angular_vector;


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -294,17 +296,19 @@ void ServoCalcs::calculateSingleIteration()
   have_nonzero_twist_stamped_ = latest_nonzero_twist_stamped_;
   have_nonzero_joint_command_ = latest_nonzero_joint_cmd_;
 
+  planning_scene_monitor_->requestPlanningSceneState();
```
JafarAbdi, line 299: Why keep requesting the whole planning scene on each iteration .? see https://github.com/ros-planning/moveit2/pull/673/files for how we fixed this issue for moveit2

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -294,17 +296,19 @@ void ServoCalcs::calculateSingleIteration()
...
+  planning_scene_monitor::LockedPlanningSceneRO planning_scene(planning_scene_monitor_);
```
JafarAbdi, line 300: This will lock the planning scene until this function exits, is this the desired thing .? you may want to wrap it in a new scope

AndyZe: What Jafar means is something like this:

```
{
  planning_scene_monitor::LockedPlanningSceneRO planning_scene(planning_scene_monitor_);
  
  ... use planning_scene ...
}
```

udopa: Done.

moveit_ros/moveit_servo/src/servo_calcs.cpp
udopa: I'm don't know how to do that, can you elaborate a bit more?

moveit_ros/moveit_servo/src/servo_calcs.cpp
udopa: I changed all calls to the locked planning scene into a function [here](https://github.com/ros-planning/moveit/pull/2894/commits/96175854ce790d86628d7cf1735183dccb82a3ff). Happy to get feedback if this is a good way to do it/whats the preferred way.

moveit_ros/moveit_servo/include/moveit_servo/servo_calcs.h
```diff
@@ -222,6 +222,9 @@ class ServoCalcs
   /** \brief Service callback to reset Servo status, e.g. so the arm can move again after a collision */
   bool resetServoStatus(std_srvs::Empty::Request& req, std_srvs::Empty::Response& res);
 
+  /** \brief Function to get locked planning scene to avoid blocking*/
```
AndyZe, line 225:
```suggestion
  /** \brief Lock the planning scene, for thread safety*/
```

moveit_ros/moveit_servo/src/servo_calcs.cpp
AndyZe: I believe @JafarAbdi means, you only have to do this once, during initialization. Since it is already done at L197, you don't need to do it again here. (Please double-check that this works, though. I'm not 100% sure.)

```suggestion
```

AndyZe (commented):
Getting pretty close to merge-ready in my opinion. Thanks!

AndyZe:
To test this, I added a static tf publisher to my Servo launch file:

`<node pkg="tf" type="static_transform_publisher" name="test_tf_broadcaster" args="0 0 0 0 0 0 1 ee_link test_link 10" />`

And set `robot_link_command_frame: test_link`. This didn't work for me:

`ros.moveit_core.transforms: Unable to transform from frame 'test_link' to frame 'world'. Returning identity.`

Is that how you intended for it to be used?

udopa:
> To test this, I added a static tf publisher to my Servo launch file:
> 
> `<node pkg="tf" type="static_transform_publisher" name="test_tf_broadcaster" args="0 0 0 0 0 0 1 ee_link test_link 10" />`
> 
> And set `robot_link_command_frame: test_link`. This didn't work for me:
> 
> `ros.moveit_core.transforms: Unable to transform from frame 'test_link' to frame 'world'. Returning identity.`
> 
> Is that how you intended for it to be used?
> 
> Maybe it only works with subframes, not tf frames. That's still a nice improvement.

I tested both scenarios, with and without the call
```
planning_scene_monitor_->requestPlanningSceneState();
```
Without it I was **not** able to send commands in any attached/published frame. With it, it's possible to use a both. -> I would do the call in every iteration (or maybe every n-th iteration?). I guess this is why your test didn't work. 

## Testing & Intended way of use
I didn't change anything in the servo configuration file at all. This works because in case the frame is not one of the ones defined in the config file, the calculations will be done here [L492](https://github.com/udopa/moveit/blob/96175854ce790d86628d7cf1735183dccb82a3ff/moveit_ros/moveit_servo/src/servo_calcs.cpp#L492). 
```yaml
## MoveIt properties
move_group_name:  welding_endeffector  # Often 'manipulator' or 'arm'
planning_frame: base_link  # The MoveIt planning frame. Often 'base_link' or 'world'

## Other frames
ee_frame_name: wrist_3_link  # The name of the end effector link, used to return the EE pose
robot_link_command_frame:  base  # commands must be given in the frame of a robot link. Usually either the base or end effector
```
For using the new feature I published a frame using the static tf broadcaster from inside a ros node and then send the commands. I also made sure to first start all servo components and publish the frame afterwards, so that servo should not know the new frame at the beginning. So basically like you did.

moveit_ros/moveit_servo/src/servo_calcs.cpp
udopa: This doesn't work. See my other comment [here](https://github.com/ros-planning/moveit/pull/2894#issuecomment-934144401)

AndyZe:
I'm going to be out for a week. So don't think I'm ignoring this PR- I consider it pretty important.

gautz:
@AndyZe any further work required for this PR to be merged? Any hint why the CI is failing?

gautz:
btw would this #2918 possibly simplify / make this PR obsolete? @v4hn

AndyZe:
@udopa it seems like the MoveIt `getFrameTransform()` function does not work well with externally-published tf's. I made an [issue here](https://github.com/ros-planning/moveit/issues/2921) for MoveIt1 and somebody already linked me to a patch for [MoveIt2](https://github.com/ros-planning/moveit2/pull/654).

For me, the easiest path is to bring this functionality you've requested into MoveIt2. But if you want to follow through with this PR, I'll still review it.

I'm kind of curious how you said it was working for you and if I missed something when testing, though.

udopa:
@AndyZe Any news on this topic? Do you need some more information? Quick feedback would be good

Issue #2889, opened 2021-09-29
Document linear nature of Bullet CCD integration


werner291:
Hello,

I recently noticed that the CCD parts of the `collision_detection_bullet` seem to use a convex hull approximation. For reference: 
 - https://github.com/ros-planning/moveit/blob/049be73b8343d67ffde400fed9ed4c387c6ea9e5/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/bullet_utils.h#L294
 - https://github.com/ros-planning/moveit/blob/049be73b8343d67ffde400fed9ed4c387c6ea9e5/moveit_core/collision_detection_bullet/include/moveit/collision_detection_bullet/bullet_integration/bullet_utils.h#L312

This can be a rather crude approximation when the motion includes a lot of rotations. For instance, consider a robotic arm that lays nearly flat on the ground, then swings almost 180 degrees through the air to lay almost flat in the opposite direction. In this case, the convex hull approximation won't cover the vast majority of the swept volume.

I understand that CCD is quite a complicated topic, so this method might be reasonable depending on circumstances, but I think it would make sense to at least document the fact that this approximation is used such that a user can act accordingly.

simonschmeisser:
Sounds like maybe there should be some maximum distance and CCD only be used in between interpolated points of that maximum distance?

werner291:
So, what I ended up doing was something like this:

```
double BulletContinuousMotionValidator::estimateMaximumRotation(const moveit::core::RobotStatePtr &st1,
                                                                const moveit::core::RobotStatePtr &st2) {// Upper bound on the rotation of any part of the robot.
    assert(st1->getRobotModel() == st2->getRobotModel());

    double max_angle = 0.0;
    for (const auto &jm: st1->getRobotModel()->getJointModels()) {
        switch (jm->getType()) {
            case moveit::core::JointModel::UNKNOWN:
                ROS_ERROR("Cannot compute maximum rotation with unknown joint type.");
                break;
            case moveit::core::JointModel::PRISMATIC:
            case moveit::core::JointModel::FIXED:
                // These do not affect rotation
                break;

            case moveit::core::JointModel::PLANAR: {
                const double *variables1 = st1->getJointPositions(jm);
                const double *variables2 = st2->getJointPositions(jm);
                max_angle += std::abs(variables1[2] - variables2[2]);
            }
                break;

            case moveit::core::JointModel::FLOATING: {

[...]

PR #2897, opened 2021-10-06
[WIP] Issue 2889

werner291:
### Description

See the discussion in Issue #2889.

TL;DR: Bullet-based CCD is currently implemented by taking the pairwise convex hulls of each collision ibject in the robot model in the "before" and "after" states, and performing a collision test against that.

If the robot's movement is a pure translation (and assuming all collision objects are convex), this corresponds perfectly with the swept volume of the robot during this movement. However, if the movement includes rotation, especially with large swinging motions, this is a severe under-approximation.

What this PR contributes is the addition of sub-steps in the collision detection method depending on the amount of rotation in the motion. This has the advantage of allowing for very few (if any) sub-steps for linear or near-linear motions, but still providing a far better approximation if there is rotation in the movement.

I specifically made the following changes:
- Added `JointModel::distanceRotation` as a virtual method to get an idea of the rotation induced by a state change on a per-joint basis. Name is borrowed from `FloatingJointModel` for consistency; change should be non-breaking for that class.
- Added `RobotState::maximumRotation` which combines the per-joint rotations to compute an upper bound on every link's rotation.
- Made necessary changes to `CollisionEnvBullet::checkRobotCollisionHelperCCD` and added a parameter `max_rotation_per_step_` that controls the number of sub-steps.

PR is currently WIP since I'd like some feedback on whether this method works as I'd expect, and on how to best test it.

### Checklist

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -3,6 +3,7 @@
 API changes in MoveIt releases
 
 ## ROS Noetic
+- Renamed `addAttachedOjects` to `createAttachedCollisionObjects` in `CollisionEnvBullet` to reflect the non-mutating nature of the method.
 - `CollisionObject` messages are now defined with a `Pose`, and shapes and subframes are defined relative to the object's pose. This makes it easier to place objects with subframes and multiple shapes in the scene. This causes several changes:
     - `getFrameTransform()` now returns this pose instead of the first shape's pose.
     - The Rviz plugin's manipulation tab now uses the object's pose instead of the shape pose to evaluate if object's are in the region of interest.
--- a/moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
+++ b/moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
@@ -185,7 +185,7 @@ void CollisionEnvBullet::checkRobotCollisionHelper(const CollisionRequest& req, 
   }
 
   std::vector<collision_detection_bullet::CollisionObjectWrapperPtr> attached_cows;
-  addAttachedOjects(state, attached_cows);
+  createAttachedCollisionObjects(state, attached_cows);
   updateTransformsFromState(state, manager_);
 
   for (const collision_detection_bullet::CollisionObjectWrapperPtr& cow : attached_cows)
--- a/moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
+++ b/moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
@@ -50,8 +50,8 @@ const double MAX_DISTANCE_MARGIN = 99;
 constexpr char LOGNAME[] = "collision_detection.bullet";
 }  // namespace
 
-CollisionEnvBullet::CollisionEnvBullet(const moveit::core::RobotModelConstPtr& model, double padding, double scale)
-  : CollisionEnv(model, padding, scale)
+CollisionEnvBullet::CollisionEnvBullet(const moveit::core::RobotModelConstPtr& model, double padding, double scale, double max_rotation_per_step)
+: CollisionEnv(model, padding, scale), max_rotation_per_step_(max_rotation_per_step)
 {
   // request notifications about changes to new world
   observer_handle_ = getWorld()->addObserver(boost::bind(&CollisionEnvBullet::notifyObjectChange, this, _1, _2));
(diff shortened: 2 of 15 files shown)


werner291:
Sorry, I have no idea how I requested five different reviews.

werner291:
Question: Why does the `RobotState::distance`` need the `JointModelGroup` exactly? And why not in the collision environment? What is the intended idea here?

simonschmeisser:
> Sorry, I have no idea how I requested five different reviews.

By contributing to areas of code that have five different code owners registered ... don't worry that they will all respond though ;-)



> Question: Why does the `RobotState::distance` need the `JointModelGroup` exactly? And why not in the collision environment? What is the intended idea here?

It's the distance in joint space afaik, so you can limit that summation to only a certain subset of joints, e.g. only the gripper or only one arm

MIGRATION.md
```diff
@@ -3,6 +3,10 @@
 API changes in MoveIt releases
 
 ## ROS Noetic
+- Added method `maximumRotation` to `RobotState`.
```
v4hn, line 6: please leave out that line. only things relevant during migration are added here.

MIGRATION.md
```diff
@@ -3,6 +3,10 @@
...
+- The method `distanceRotation` is now a virtual method of `JointModel` and implemented for all joint models.
+- Modified `CollisionEnvBullet`. It now takes a parameter `max_rotation_per_step_`, and subdivides CCD requests into sub-steps to better approximate rotations
```
v4hn, line 8: This addition is API-compatible, so no need to mention it here either.

MIGRATION.md
```diff
@@ -3,6 +3,10 @@
...
+- Renamed `addAttachedOjects` to `createAttachedCollisionObjects` in `CollisionEnvBullet` to reflect the non-mutating nature of the method.
```
v4hn, line 9: I'm not sure this should be mentioned either. It's an API change, but it's not user-facing unless users hack in the deep internals for their own projects. Usually only `collision_detection` interfaces should be used.

moveit_core/collision_detection_bullet/src/collision_env_bullet.cpp
```diff
@@ -208,27 +209,65 @@ void CollisionEnvBullet::checkRobotCollisionHelperCCD(const CollisionRequest& re
...
+    {
+      manager_CCD_->setCastCollisionObjectsTransform(link, st1.getCollisionBodyTransform(link, 0),
+                                                     st2.getCollisionBodyTransform(link, 0));
+    }
+
+    // Perform the contact test
+    manager_CCD_->contactTest(res, req, acm, false);
+
+    // Stop the operation after the first collision.
+    if (res.collision)
```
v4hn, line 266: you should actually respect `req.max_contacts` here and keep going until you found enough contacts.
That will probably also fix your invalidated CI test because you miss collisions that only happen further into the trajectory.

To abort checking on collision, users can do checks for just one contact.

moveit_core/collision_detection_bullet/test/test_bullet_continuous_collision_checking.cpp
```diff
@@ -368,6 +368,75 @@ TEST(ContinuousCollisionUnit, BulletCastMeshVsBox)
...
+
+  // Should be in collision when upright
+  cenv_->checkRobotCollision(req, res, upright, *acm_);
+  ASSERT_TRUE(res.collision);
+  res.clear();
+
+  // Now the CCD collision check should show the same as the previous check, as the robot hits the box when moving.
+  cenv_->checkRobotCollision(req, res, before_swing, after_swing, *acm_);
+  ASSERT_TRUE(res.collision);
+}
```
v4hn, line 438: I love this check. Thank you so much!

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1502,6 +1502,9 @@ class RobotState
   /** \brief Return the sum of joint distances to "other" state. Only considers active joints. */
   double distance(const RobotState& other, const JointModelGroup* joint_group) const;
 
+  /** \brief Return an upper bound on the rotation angle (radians) of any one link of the robot between two states. */
+  double maximumRotation(const RobotState& other) const;
+
```
v4hn, line 1507: This interface requires a lot more collision checks in your approach than are actually needed, because you can't consider the rotation distances separately for the different collision bodies. The upper bound of rotation for the base links will usually be drastically lower than for the end effector links.

Do you think it might be worth the effort to compute the distances for all collision bodies individually and discretize their path based on these values instead? The obvious downside would be the additional complexity of constructing possibly even more robot states and selectively checking only a subset of bodies in each one.
I'm asking because I work with robots with 50+ links and bimanual motion, so checking one arm for collisions over and over again just because the other one moves a lot sounds like a lot of overhead. That being said, I personally do not use bullet checking at all at the moment and my setup is surely not the average use case.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -920,6 +920,18 @@ double RobotState::distance(const RobotState& other, const JointModelGroup* join
...
 }
 
+double RobotState::maximumRotation(const RobotState& other) const
+{
+  double d = 0.0;
+  const std::vector<const JointModel*>& jm = this->getRobotModel()->getActiveJointModels();
+  for (const JointModel* joint : jm)
+  {
+    const int idx = joint->getFirstVariableIndex();
+    d += joint->distanceRotation(position_ + idx, other.position_ + idx);
```
v4hn, line 930:
```suggestion 
    d += joint->distanceRotation(&position_[idx], &other.position_[idx]);
```

It might be just me, but I consider this much more readable.

PR #2898, opened 2021-10-06
Servo: sync position limit enforcement with MoveIt2

mechwiz:
### Description
I was working on a project with @parunapu and we noticed an interesting behavior when using moveit_servo streaming velocity commands where joint limits would be exceeded. Upon further investigation I noticed in the [enforcePositionLimits()](https://github.com/ros-planning/moveit/blob/master/moveit_ros/moveit_servo/src/servo_calcs.cpp#L805) function that the current state velocities would be checked along with the joint position limits such that we only "halt if we are past a joint margin and joint velocity is moving even further past" (as described in a comment earlier in the function). The problem is that the check is for velocities greater or below zero which means that when the velocities become zero on that cycle, then on the next iteration, enforcePositionLImits() will return that everything is normal and a small delta velocity will actually end up being commanded. This will lead to a continuous jerky cycle of starts and stops violating the joint limit that we don't want. A simple fix would be to only check position limits with the margins but that would not take into account whether we are moving away or not from the limit. Hence, this fix simply uses the velocities already calculated in the `internal_joint_state_` variable to check if the command we want to send is going even further than the limit or not.

@AndyZe, let me know if this change makes sense.

### Checklist

### DIFF
--- a/moveit_ros/moveit_servo/src/servo_calcs.cpp
+++ b/moveit_ros/moveit_servo/src/servo_calcs.cpp
@@ -779,7 +779,7 @@ void ServoCalcs::enforceVelLimits(Eigen::ArrayXd& delta_theta)
   delta_theta = velocity_scaling_factor * velocity * parameters_.publish_period;
 }
 
-bool ServoCalcs::enforcePositionLimits(sensor_msgs::msg::JointState& joint_state)
+bool ServoCalcs::enforcePositionLimits(sensor_msgs::JointState& joint_state)
 {
   bool halting = false;
 
@@ -807,9 +807,9 @@ bool ServoCalcs::enforcePositionLimits(sensor_msgs::msg::JointState& joint_state
         auto joint_idx = std::distance(joint_state.name.begin(), joint_itr);
 
         if ((joint_state.velocity.at(joint_idx) < 0 &&
-             (joint_angle < (limits[0].min_position + parameters_->joint_limit_margin))) ||
+             (joint_angle < (limits[0].min_position + parameters_.joint_limit_margin))) ||
             (joint_state.velocity.at(joint_idx) > 0 &&
-             (joint_angle > (limits[0].max_position - parameters_->joint_limit_margin))))
+             (joint_angle > (limits[0].max_position - parameters_.joint_limit_margin))))
         {
           ROS_WARN_STREAM_THROTTLE_NAMED(ROS_LOG_THROTTLE_PERIOD, LOGNAME,
                                          ros::this_node::getName() << " " << joint->getName()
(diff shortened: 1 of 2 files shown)


moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -787,11 +787,13 @@ bool ServoCalcs::enforcePositionLimits()
   {
     // Halt if we're past a joint margin and joint velocity is moving even farther past
     double joint_angle = 0;
+    std::size_t idx = 0;
     for (std::size_t c = 0; c < original_joint_state_.name.size(); ++c)
```
AndyZe, line 791: I think you could use `idx` in the for loop and completely get rid of `c`.
```suggestion
    for (idx = 0; idx < original_joint_state_.name.size(); ++idx)
```

mechwiz: That makes sense! i'll update that and test again

mechwiz: yep! tested again and works

moveit_ros/moveit_servo/src/servo_calcs.cpp
```diff
@@ -787,24 +787,25 @@ bool ServoCalcs::enforcePositionLimits()
...
     if (!current_state_->satisfiesPositionBounds(joint, -parameters_.joint_limit_margin))
     {
-      const std::vector<moveit_msgs::JointLimits> limits = joint->getVariableBoundsMsg();
+      const std::vector<moveit_msgs::JointLimits>& limits = joint->getVariableBoundsMsg();
 
       // Joint limits are not defined for some joints. Skip them.
       if (!limits.empty())
       {
-        if ((current_state_->getJointVelocities(joint)[0] < 0 &&
+        if ((internal_joint_state_.velocity[idx] < 0 &&
```
AndyZe, line 806: If I understand right, this PR uses `internal_joint_state_` because it takes the low-pass filtering of the signal into account. `current_state_` that was used previously comes from the planning scene monitor so it isn't aware of that filtering.

I'm gonna suggest a comment here then LGTM. I'll try to get one other person to review it before merging.

AndyZe:
```suggestion
        // Use `internal_joint_state_` rather than `current_state_` so that low-pass filtering is accounted for
        if ((internal_joint_state_.velocity[idx] < 0 &&
```

mechwiz: So I'm actually using the `internal_joint_state_` because that contains the velocities that we want to command (it's passed in to the `composeJointTrajMessage` function) and will fix that issue described in the description. Meaning, we look at the velocities we want to command to see if they will put the joint even further from the joint limit or not. Looking at the current velocities is problematic since it doesn't give us that insight which leads to the issue described in the description where if the velocities were already zero due to a "halt" from the previous iteration, positions were not enforced since a zero-velocity case was considered OK thereby causing the next command to be nonzero.

PR #2901, opened 2021-10-11
Off-by-one error in servo_calcs.cpp Melodic branch

janneyow:
### Description
Servo server trajectory output incorrect.

- The 2nd point in the trajectory (instead of third) should be taken and the timestamp should also be shifted to the correct point.

- Minor fix that wasn't back-ported to the Melodic version

- Corresponding pull request: #2740 

### Checklist


v4hn:
Replaced by #2908 .

@simonschmeisser It's usually simpler to just do the cherry-pick yourself in these cases... Thanks for pointing it out @janneyow .
The fix will be in the next melodic release.

PR #2902, opened 2021-10-11
Backport ruckig trajectory_processing plugin

simonschmeisser:
### Description

This is a backport of  https://github.com/ros-planning/moveit2/pull/571

"Unfortunately" ruckig requires C++ 17 for its headers so this changes the `moveit_default_planning_request_adapter_plugins` to compile with C++ 17. It's a private property so it does not propagate anywhere and especially does not lead to any dependent code or project requiring C++ 17

### Checklist


AndyZe:
Please be aware, this won't work great until 712 is included.

https://github.com/ros-planning/moveit2/pull/712

simonschmeisser:
should also include https://github.com/ros-planning/moveit2/pull/777

AndyZe:
Trying to test this, it looks like I'm not able to install the Ruckig binary. Any ideas?

`rosdep install --from-paths src --ignore-src -r -y`

`moveit_core: Cannot locate rosdep definition for [ruckig]`

simonschmeisser:
I needed to update the cache: `rosdep update` and then it worked:
```
rosdep where-defined ruckig
https://raw.githubusercontent.com/ros/rosdistro/master/melodic/distribution.yaml
```

prasuchit:
Hi,

Is this still being considered for backport? I'm currently using noetic and would like to try this out with moveit. Is there an estimated timeline for merging this?

simonschmeisser:
There are some more commits in the PRs linked in the discussion that need to be included. No estimate when I'll do it. You could open a PR with those commits against my branch if you want to speed things up

prasuchit:
Hi @AndyZe ,

I haven't raised a PR yet, it's on my fork created from @simonschmeisser 's version of the backport. This is the first time I'm adding to a package as large as moveit so I wanted to verify it first with you guys.

I can use noetic without any issues, just the moveit source compilation fails because the latest moveit repo uses some cmake options that aren't being recognized by my current version. I thought instead of getting distracted by a different issue, I can get your help to fix the problem at hand.

simonschmeisser:
I just rebased my backport to the current state of MoveIt 1 master branch. What you could do now is to use `git cherry-pick SOMEHASH` where somehash is the commit hash of further changes in MoveIt2 needed to get ruckig plugin up and running. `moveit_build_options()` was introduced in MoveIt2, it has nothing to do with the cmake version. It needs to be replaced by similar behaviour when we want to integrate it into MoveIt1. That's what's called back-porting, it sometimes requires some real work ...

simonschmeisser:
There are a few more commits and PRs that went into the ruckig plugin so maybe actually @AndyZe hands-on approach of copy-pasting and adapting the final resulting file makes more sense than trying to cherry pick all of the commits from https://github.com/ros-planning/moveit2/commits/main/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp individually

AndyZe:
I'm going to try to get this across the finish line now   :+1:

rhaschke:
Given that `ruckig` is obviously missing from Bionic/Melodic, I suggest building that plugin optionally only on Focal/Noetic.
The other builds fail due to a simple unused argument. I'm surprised that MoveIt2 checks passed for this issue!

simonschmeisser:
> Given that `ruckig` is obviously missing from Bionic/Melodic, I suggest building that plugin optionally only on Focal/Noetic.

What makes that `obvious` and why shouldn't it be released to ROS1 as well?

rhaschke:
> > Given that ruckig is obviously missing from Bionic/Melodic ...

> What makes that `obvious`?

Sorry, you are right! The error message from the [failing CI build](https://github.com/ros-planning/moveit/actions/runs/3698484059/jobs/6264752584) made me conclude that `ruckig` is missing in Melodic:
`catkin_package() DEPENDS on 'ruckig' but neither 'ruckig_INCLUDE_DIRS' nor 'ruckig_LIBRARIES' is defined.`

Looks like this is another issue. I will have a look...

rhaschke:
Everything is compiling now. But there is a test failure on Melodic. Somebody with more insight should handle this...

AndyZe:
OK, this should be 100% synched with MoveIt2 now, and I've just tested it.  :+1:

rhaschke (approved):

rhaschke:
After having merged this PR, [GHA fails now on ruckig tests](https://github.com/ros-planning/moveit/actions/runs/3747836136) - but only for one out of 3 builds?! Same for the [prerelease test](https://github.com/ros-planning/moveit/actions/runs/3748862811/jobs/6366682148).
I wasn't able yet to reproduce the issue locally. As this currently blocks the prepared release, I'm grateful for help, @AndyZe.

AndyZe:
OK, I'll look into it. Thanks

rhaschke:
It doesn't seem to be related to a `Release` build. I don't have any more ideas...
If we can't find the underlying reason, we could just disable the test...

AndyZe:
That's fine with me, for now. It's going to take awhile to look into this strange error, I think.

Do you want to make a PR?

PR #2903, opened 2021-10-12
dont add any '-W*' definitions

LeroyR:
### Description

These should go in CI configuration if wanted.


LeroyR:
> Could you please justify your PR a bit more?

Sure, i just assumed that the specific enabled warnings and handling (esp. Werror) should be part of the specific developer/ci configuration instead of the build system instead

- Both of these flags enable different things depending on the compiler/version used.
- Some are questionable, even seen here as unused-parameter gets disabled
- Developers may not expect the build to fail by adding some random warning flag (werror)
- May break build on different compilers (ex. MSVC needs /W4 or something) 
- prefer `target_compile_options` over global definitions
- We should leave which compile warnings are shown to the developer, we dont enforce formatting/tidy in the build but instead with CI **upstream**

Also this does break the build on our configurations somewhere in Eigen :D

PR #2910, opened 2021-10-18
WIP: Rework Python API

rhaschke:
This is an attempt to modernize the Python API from `boost::python` to `pybind11` and reduce boilerplate code for type conversions based on initial work in #2547. See individual commits for details.

- [x] Move `py_bindings_tools` from `moveit_ros_planning_interface` to `moveit_core` (and some cleanup)
- [ ] Feed `rospy.names.get_mappings()` into `ROScppInitializer` to use python remappings for C++ as well?
- [x] Check whether [`ros_msg_typecasters.h`](https://github.com/ubi-agni/moveit/blob/3fc37652f97e864e4d952e5bbd9b28cb1ec4c231/moveit_core/python/tools/include/moveit/py_bindings_tools/ros_msg_typecasters.h) requires PyBytes handling 
  as introduced by @gleichdick in #1870
- [x] Migrate existing wrappers to `pybind11` and its automatic type conversions. Remove [`serialize_msg.h`](https://github.com/ros-planning/moveit/blob/de54829d70e9aac7deec8c654a73d0ed41ba604d/moveit_core/python/tools/include/moveit/py_bindings_tools/serialize_msg.h).
- [x] Replace [`py_conversions.h`](https://github.com/ubi-agni/moveit/blob/de54829d70e9aac7deec8c654a73d0ed41ba604d/moveit_core/python/tools/include/moveit/py_bindings_tools/py_conversions.h) with [`pybind11/stl.h`](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html)
- [x] Replace [`gil_releaser.h`](https://github.com/ubi-agni/moveit/blob/de54829d70e9aac7deec8c654a73d0ed41ba604d/moveit_core/python/tools/include/moveit/py_bindings_tools/gil_releaser.h) with [`pybind11::gil_scoped_release`](https://pybind11.readthedocs.io/en/stable/advanced/misc.html#global-interpreter-lock-gil)
- [x] Replace `eigenpy` with [`pybind11/eigen.h`](https://pybind11.readthedocs.io/en/stable/advanced/cast/eigen.html)
- [ ] Fix fundamental catkin issue https://github.com/ros/catkin/pull/1153
- [ ] Check backwards compatibility of `moveit_commander`
- [ ] Host python docs somewhere
  - #2606
  - #2914

@gleichdick, @PeterMitrano: I would be happy if you could contribute to this work, filing PRs against my [PR branch](https://github.com/ubi-agni/moveit/tree/rework-python-api).
What do you think of the cleanup so far?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,7 +25,6 @@ jobs:
         env:
           - IMAGE: noetic-ci
             CCOV: true
-            BUILDER: catkin_tools_devel
           - IMAGE: noetic-ci-testing
             IKFAST_TEST: true
             CATKIN_LINT: true


gleichdick:
Always happy to help :smile: 

`ros_msg_typecasters.h` does not need `ByteString`, as the buffer does not cross the boundary between pybind11/boost::python and another python module. `pybind11::bytes` could be used to get rid of the plain CPython function calls, but that'd be just for readability.

If we use `ros_msg_typecasters.h` in planning_interface, the interface of the wrapper changes. Do we take in accout users which use these wrappers directly?

Side note: I tried to enable coverage for moveit_commander, but some problems arised: ros/ros_comm#2199

rhaschke:
> If we use `ros_msg_typecasters.h` in planning_interface, the interface of the wrapper changes. Do we take in account users which use these wrappers directly?

Using the old library wrappers directly isn't very useful because you would need to handle serialized byte strings. I was expecting nobody to do that and thus not provide backwards compatibility to this awkward API.
Of course, backward compatibility should be maintained for all the existing python API operating on python ROS messages.

moveit_ros/planning_interface/test/CMakeLists.txt
```diff
@@ -14,19 +14,7 @@ if (CATKIN_ENABLE_TESTING)
   target_link_libraries(subframes_test moveit_move_group_interface ${catkin_LIBRARIES})
 
   add_rostest(python_move_group.test)
-  add_rostest(python_move_group_ns.test)
```
gleichdick, line 17: Just to be sure: This test file still exists, so why don't we execute it?

rhaschke: Well spotted! This should be added again. (It was removed in between because I tried to handle both tests with the same .test file.)

rhaschke:
@ros-planning/moveit-maintainers, @gleichdick, @PeterMitrano: These python wrapper improvements are lying around (nearly ready) for over a year now. The (only) road blocker was and is a major design flaw in catkin's devel-space layout (https://github.com/ros/catkin/pull/1153), which renders _multiple_ namespace packages installing into the same devel-space location impossible. 
Specifically, `moveit.core` introduced in #2547, `moveit.planning_interface` renamed from `moveit_ros_planning_interface` in this PR, and `moveit.task_constructor` cannot be used together within the same _devel_-space workspace. Essentially only the first installed package (i.e. `moveit.core`) will be accessible (see [CI failure](https://github.com/ros-planning/moveit/actions/runs/3882363987/jobs/6623779906)).
Install-space workspaces don't have that issue. MTC within a devel-space workspace that is overlaying an install-space workspace comprising MoveIt (e.g. /opt/ros) works as well.

As we won't fix catkin anymore, I only see the following options:
- Drop this PR and focus on python for MoveIt2 only (not really an option)
- Drop the (nice) idea of namespace package and rename packages from `moveit.*` to `moveit_*`
- Ignore the issue and recommend using install spaces, particularly `symlinked` installs with `colcon` to keep the benefits of a _devel_ space. As releases are not affected, I think that is a viable option.

What do you think? @henningkayser: We should discuss that in the next maintainer meeting.

PeterMitrano:
Seems like #2 of changing the naming is by far the least work and honestly not that bad. Moveit 1 doesn't seem to be getting much core support anymore, and python for moveit 1 isn't high on the priorities (although I really want it). So seems like the second option is best?

As for  #2613 I think we're still blocked on docs not generating/being hosted somewhere? #2606

moveit_core/robot_model/src/pyrobot_model.cpp
```diff
@@ -38,26 +38,44 @@
...
+      .def("get_active_joint_model_names", &RobotModel::getActiveJointModelNames)
 
-  py::class_<JointModelGroup, JointModelGroupPtr>(m, "JointModelGroup")
-      .def("getLinkModelNames", &JointModelGroup::getLinkModelNames)
-      .def("getJointModelNames", &JointModelGroup::getJointModelNames)
-      //
+      .def("get_planning_frame", &RobotModel::getModelFrame)
+      .def("get_robot_root_link", &RobotModel::getRootLinkName)
+      .def("has_group", &RobotModel::hasJointModelGroup, py::arg("group"))
+      .def("get_robot_name", &RobotModel::getName)
```
PeterMitrano, line 78: I was rebasing #2613 on top of this and noticed you have this method bound twice with different names. Is that right? I think we should only have `get_name`

PeterMitrano:
I've done some more investigating on how to get `import moveit.planning_interface` to work despite the devel space limitations. There are one or two ways to make it work using some tricks, but they require centralizing the python bindings to some extent, instead of distributing them completely across each package. Specifically, we could:

1. Put all the python bindings in one package, this means the .cpp files defining the bindings as well as the __init__.py files and such. In my opinion, this is a very clean & simple solution, but it means we have to maintain python bindings in a separate place
2. Put some import magic in the __init__.py files in a centralized location. The actual binding definitions can live in each specific package, and each package will build a file like `pymoveit_core.so` or `pymoveit_planning_interface.so`, and we would have one package that lets you do `import moveit.[core|planning_interface]`

Thoughts?

PeterMitrano:
Alternatively, we can just not have in centralized imports, and stick with `import moveit_core` and `import moveit_planning_interface`

PR #2925, opened 2021-10-21
Fix `ClassLoader: SEVERE WARNING`

rhaschke:
Resetting the `MotionPlanningDisplay` in rviz (or disabling+enabling it) issues the following warning:
```
[ WARN] ros.rosconsole_bridge.class_loader.ClassLoader: SEVERE WARNING!!!
Attempting to unload /vol/sandbox/ros/devel/lib//libmoveit_kdl_kinematics_plugin.so
while objects created by this library still exist in the heap!
You should delete your objects before destroying the ClassLoader. The library will NOT be unloaded.
```
The reason is that we don't release all references to the `RobotModel` (and thus it's IK solvers), before destroying the corresponding `RobotModelLoader`. I reduced the number of pending pointers from 21 to 7.
For now, I cannot immediately see where the remaining 6 refs are hidden. Thus, anybody who likes to contribute is highly welcome!

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/include/moveit/motion_planning_rviz_plugin/motion_planning_display.h
@@ -196,6 +196,7 @@ protected:
     OUTSIDE_BOUNDS_LINK
   };
 
+  void clearRobotModel() override;
   void onRobotModelLoaded() override;
   void onNewPlanningSceneState() override;
   void onSceneMonitorReceivedUpdate(planning_scene_monitor::PlanningSceneMonitor::SceneUpdateType update_type) override;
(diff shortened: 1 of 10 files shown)


v4hn:
Shouldn't we at least merge (most of this) patch by now and create an issue for the remaining leaked pointers?

rhaschke:
IMO, this PR is not very criticial. Thus, I would leave it open - hoping for somebody to contribute...
Note, that there are several debug outputs in there currently, which should be removed before merging. If left open, these outputs will directly facilitate further debugging.

JafarAbdi:
@rhaschke This [PR](https://github.com/ubi-agni/moveit/pull/5) should remove the remaining usage

rhaschke:
Thanks a lot @JafarAbdi for tracking down the remaining references! If CI succeeded, this should be ready for merging.

JafarAbdi (approved):

PR #2937, opened 2021-11-01
Reduced the minimum max velocity joint limit from 0.01 to 0.001 for totg

grejj:
### Description

Currently the minimum value you can specify for the the max velocity outputted for a joint during time optimal trajectory parameterization is 0.01 rad/s or m/s. For joints that are unable to move that fast and whose velocity limits are less than 0.01, the planned path outputted is unachievable with trajectory velocities being higher than the maximum velocity of the joint. 

Reduced the value to 0.001.

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -932,17 +932,25 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
     max_velocity[j] = 1.0;
     if (bounds.velocity_bounded_)
     {
+      if (bounds.max_velocity_ < std::numeric_limits<double>::epsilon())
+      {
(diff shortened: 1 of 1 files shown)


AndyZe (approved):

AndyZe:
@grejj I think this is a good change but do we even need a minimum. It looks to me like throwing an error might be better like this:

```
  if (max_acceleration_[j] <= 0)
  {
    ROS_ERROR_NAMED(LOGNAME, "Max acceleration must be greater than zero.");
    return false;
}
```

v4hn:
Thanks for creating the PR after our brief discussion on Discord!
I agree with @AndyZe.
@grejj Could you modify your patch to make the parameterization fail when values less-equal 0 are encountered in velocity or acceleration bounds? This will also give users clear feedback when they use configurations with `has_*_limits: true` without specifying an actual limit.

AndyZe:
LGTM.

rhaschke (approved):

rhaschke:
@grejj, please apply the suggested formatting changes. Ideally, install https://pre-commit.com to catch these at commit time.

Issue #2930, opened 2021-10-22
Multi DOF-joint error


romzn:
Hi Pilz-Team,
you did a fantastic job with the industrial motion planner! 
Unfortunately i have a problem with the implementation to another Robot Manipulator.

### Steps to reproduce

I implemented the `pilz_command_planner` into another moveit-config (Franka Panda).
When i launched the demo with the following statement.

```bash
roslaunch panda_moveit_config demo.launch pipeline:=pilz_command_planner
```
an error inside your `MoveGroupSequenceAction` came up:

```bash
Loading 'pilz_trajectory_generation/MoveGroupSequenceAction'...
[ INFO] [1634816344.711287404]: initialize move group sequence action
[ INFO] [1634816344.731243551]: Reading limits from namespace /robot_description_planning
[ERROR] [1634816344.731610133]: Multi-DOF-Joints not supported. The robot won't move.
[ERROR] [1634816344.731635020]: Multi-DOF-Joints not supported. The robot won't move.
Loading 'pilz_trajectory_generation/MoveGroupSequenceService'...
[ INFO] [1634816344.796014051]: Reading limits from namespace /robot_description_planning
[ERROR] [1634816344.796358150]: Multi-DOF-Joints not supported. The robot won't move.
[ERROR] [1634816344.796385235]: Multi-DOF-Joints not supported. The robot won't move.
[ INFO] [1634816344.825634767]: 
```
The `robot_description_planning` parameter contains:

```bash
rosparam get /robot_description_planning
joint_limits:
  panda_finger_joint1:
    has_acceleration_limits: false
    has_velocity_limits: true
    max_acceleration: 0
    max_velocity: 0.1
  panda_finger_joint2:
    has_acceleration_limits: false
    has_velocity_limits: true
    max_acceleration: 0
    max_velocity: 0.1
  panda_joint1:
    has_acceleration_limits: true
    has_velocity_limits: true
    max_acceleration: 3.75
    max_velocity: 2.175
  panda_joint2:
    has_acceleration_limits: true
    has_velocity_limits: true
    max_acceleration: 1.875
    max_velocity: 2.175
  panda_joint3:
    has_acceleration_limits: true
    has_velocity_limits: true
    max_acceleration: 2.5
    max_velocity: 2.175
  panda_joint4:
    has_acceleration_limits: true
    has_velocity_limits: true
    max_acceleration: 3.125
    max_velocity: 2.175
  panda_joint5:
    has_acceleration_limits: true
[...]

PR #2940, opened 2021-11-02
Improve error messages

rhaschke:
Address https://github.com/ros-planning/moveit/issues/2930#issuecomment-951944125 and various other weak error messages in Pilz planner:
- Downgrade ERROR to WARN
- Report affected joint name
- Quote (possibly empty) planner id

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/src/joint_limits_aggregator.cpp
+++ b/moveit_planners/pilz_industrial_motion_planner/src/joint_limits_aggregator.cpp
@@ -105,7 +105,7 @@ void pilz_industrial_motion_planner::JointLimitsAggregator::updatePositionLimitF
   {
     // LCOV_EXCL_START
     case 0:
-      ROS_ERROR_STREAM("no bounds set for joint " << joint_model->getName());
+      ROS_WARN_STREAM("no bounds set for joint " << joint_model->getName());
       break;
     // LCOV_EXCL_STOP
     case 1:
@@ -115,7 +115,7 @@ void pilz_industrial_motion_planner::JointLimitsAggregator::updatePositionLimitF
       break;
     // LCOV_EXCL_START
     default:
-      ROS_ERROR_STREAM("Multi-DOF-Joints not supported. The robot won't move. " << joint_model->getName());
+      ROS_WARN_STREAM("Multi-DOF-Joint '" << joint_model->getName() << "' not supported.");
       joint_limit.has_position_limits = true;
       joint_limit.min_position = 0;
       joint_limit.max_position = 0;
@@ -134,7 +134,7 @@ void pilz_industrial_motion_planner::JointLimitsAggregator::updateVelocityLimitF
   {
     // LCOV_EXCL_START
     case 0:
-      ROS_ERROR_STREAM("no bounds set for joint " << joint_model->getName());
+      ROS_WARN_STREAM("no bounds set for joint " << joint_model->getName());
       break;
     // LCOV_EXCL_STOP
     case 1:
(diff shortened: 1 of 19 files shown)


moveit_planners/pilz_industrial_motion_planner/src/trajectory_generator_lin.cpp
```diff
@@ -52,7 +52,7 @@
 namespace pilz_industrial_motion_planner
 {
 TrajectoryGeneratorLIN::TrajectoryGeneratorLIN(const moveit::core::RobotModelConstPtr& robot_model,
-                                               const LimitsContainer& planner_limits)
+                                               const LimitsContainer& planner_limits, const std::string& /*group_name*/)
```
v4hn, line 55: Is it too much of a hassle to store the group from the constructor for `LIN` and `CIRC` as well?
It's very inconsistent to ignore the parameter in two out of three constructors.

rhaschke: The group is only needed in `TrajectoryGeneratorPTP` to find joint limits for that group only.
As a consequence of [Pilz' software design](https://github.com/ros-planning/moveit/pull/2940/commits/c90aacbeb045292e01855938e059c72e194dafc8#diff-7b18543ba76a984ef64270ec9893a6f52c042a9cd3c9a233de62e2f49ccb1de5R66), I had to introduce the parameter in all other constructors as well :disappointed: 

The group itself isn't stored in any of those classes.

jschleicher (approved):
Could you explain, how this PR relates to #2930 ? The suggestion there was to re-formulate the "Multi-DOF"-error message but I cannot find that in your PR.

Generally your changes look good to me, approving.

rhaschke:
> Could you explain, how this PR relates to #2930 ? The suggestion there was to re-formulate the "Multi-DOF"-error message but I cannot find that in your PR.

That's only in the very first commit:
- https://github.com/ros-planning/moveit/pull/2940/commits/c65260dbee4b81fd0405f2f84947520de3666a01#diff-2607b041b4c67ce007d4f3910ee3f546a9fe4c6e3dc5757e70a197b75ca7e2c5R118
- https://github.com/ros-planning/moveit/commit/c65260dbee4b81fd0405f2f84947520de3666a01#diff-2607b041b4c67ce007d4f3910ee3f546a9fe4c6e3dc5757e70a197b75ca7e2c5R146

moveit_planners/pilz_industrial_motion_planner/src/joint_limits_aggregator.cpp
```diff
@@ -115,7 +115,7 @@ void pilz_industrial_motion_planner::JointLimitsAggregator::updatePositionLimitF
       break;
     // LCOV_EXCL_START
     default:
-      ROS_ERROR_STREAM("Multi-DOF-Joints not supported. The robot won't move. " << joint_model->getName());
+      ROS_WARN_STREAM("Multi-DOF-Joint '" << joint_model->getName() << "' not supported.");
```
AndyZe, line 118: This misses something we talked about in #2930: there still is no warning about virtual joints not being supported.

```suggestion
      ROS_WARN_STREAM("Multi-DOF or virtual joint '" << joint_model->getName() << "' is not supported.");
```

rhaschke: This warning is not particularly about virtual joints, but about a (virtual) `floating` joint. Thus, I consider the proposed message sufficient.

AndyZe:
```suggestion
      ROS_WARN_STREAM("Floating joint '" << joint_model->getName() << "' is not supported.");
```

AndyZe: I'm insistent because `Multi-ODF` is a joint type. That makes the message confusing.

rhaschke: Multi-DOF isn't a joint type itself, but it comprises both, the `floating` and `planar` joint types. Thus, the message is perfectly valid. Your proposal is too narrow.

PR #2943, opened 2021-11-03
Use PTP as default planner for Pilz

rhaschke:
A PlannerManager should always provide a default planner.
This partially addresses item 6 of #2741 by falling back to `PTP` when no `planner_id` is specified.

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/src/pilz_industrial_motion_planner.cpp
+++ b/moveit_planners/pilz_industrial_motion_planner/src/pilz_industrial_motion_planner.cpp
@@ -123,19 +123,24 @@ CommandPlanner::getPlanningContext(const planning_scene::PlanningSceneConstPtr& 
                                    moveit_msgs::MoveItErrorCodes& error_code) const
 {
   ROS_DEBUG_STREAM("Loading PlanningContext for request\n<request>\n" << req << "\n</request>");
-
-  // Check that a loaded for this request exists
-  if (!canServiceRequest(req))
+  std::string planner_id = req.planner_id;
+  if (planner_id.empty() && context_loader_map_.find("PTP") != context_loader_map_.end())
   {
-    ROS_ERROR_STREAM("No ContextLoader for planner_id '" << req.planner_id.c_str() << "' found. Planning not possible.");
+    planner_id = "PTP";
+    ROS_INFO_STREAM("Using default planner: " + planner_id);
+  }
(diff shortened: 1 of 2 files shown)


moveit_planners/pilz_industrial_motion_planner/src/pilz_industrial_motion_planner.cpp
```diff
@@ -123,19 +123,24 @@ CommandPlanner::getPlanningContext(const planning_scene::PlanningSceneConstPtr&
...
 {
   ROS_DEBUG_STREAM("Loading PlanningContext for request\n<request>\n" << req << "\n</request>");
-
-  // Check that a loaded for this request exists
-  if (!canServiceRequest(req))
+  std::string planner_id = req.planner_id;
+  if (planner_id.empty() && context_loader_map_.find("PTP") != context_loader_map_.end())
+  {
+    planner_id = "PTP";
+    ROS_INFO_STREAM("Using default planner: " + planner_id);
```
jschleicher, line 130:
```suggestion
    ROS_WARN_STREAM("Empty planner_id in the request. Using default planner: " + planner_id);
```

jschleicher (commented):
Haven't tested it. If you think, an empty planner_id should get a default just go for it. I however want to mention, that the behaviour between PTP, LIN and CIRC is fundamentally different from OMPL: Where the RRT vs. other algorithms are all expected to generate *any* collision-free path the assumption in the joint-space vs. cartesian trajectory generator is to take the direct (joint or cartesian) way. If not filled, the other MotionPlanRequest paremters such as velocity_scaling are different to what the user expects as well so I at least suggest a warning to the user... On the other hand I understand the use-case to simply press "Plan-and-execute" in RViz and have sensible defaults.

moveit_planners/pilz_industrial_motion_planner/src/pilz_industrial_motion_planner.cpp
rhaschke: I am undecided to merge this PR. I can follow both arguments. Consider it as a proposal. 
Other maintainers might comment.
Alternatively / additionally, we should configure a default planner.

rhaschke:
Closing. [`PTP` is configured as the default planner for rviz now](https://github.com/ros-planning/moveit/commit/170e6d2091a9e3f4b65b57eaa64a470d18b62cea) (only applies after re-creating `moveit_config` with MSA).

PR #2944, opened 2021-11-03
Various fixes

rhaschke:
In preparation of the Noetic release, I tackled two more items from the list in #2741, namely no. 2 and 11.
I'm not satisfied though, with hiding the "nullspace exploration" label as it just flickers up and one cannot read it anymore.

https://user-images.githubusercontent.com/5376030/140199990-f4d3bcb0-e31b-4800-b502-cdfc573d9c24.mp4

### DIFF
--- a/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
+++ b/moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
@@ -264,6 +264,11 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
     ROS_DEBUG_STREAM("Motion planner reported a solution path with " << state_count << " states");
     if (check_solution_paths_)
     {
+      visualization_msgs::MarkerArray arr;
+      visualization_msgs::Marker m;
+      m.action = visualization_msgs::Marker::DELETEALL;
+      arr.markers.push_back(m);
+
       std::vector<std::size_t> index;
       if (!planning_scene->isPathValid(*res.trajectory_, req.path_constraints, req.group_name, false, &index))
       {
@@ -301,7 +306,6 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
                              << private_nh_.resolveName(MOTION_CONTACTS_TOPIC));
 
             // call validity checks in verbose mode for the problematic states
-            visualization_msgs::MarkerArray arr;
             for (std::size_t it : index)
             {
               // check validity with verbose on
@@ -325,8 +329,6 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
               }
             }
             ROS_ERROR_STREAM("Completed listing of explanations for invalid states.");
-            if (!arr.markers.empty())
-              contacts_publisher_.publish(arr);
           }
         }
         else
@@ -335,6 +337,7 @@ bool planning_pipeline::PlanningPipeline::generatePlan(const planning_scene::Pla
       }
       else
         ROS_DEBUG("Planned path was found to be valid when rechecked");
+      contacts_publisher_.publish(arr);
     }
   }
 
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
@@ -336,28 +336,20 @@ void MotionPlanningDisplay::updateBackgroundJobProgressBar()
 
   if (n == 0)
   {
-    p->setValue(p->maximum());
-    p->update();
     p->hide();
     p->setMaximum(0);
+    p->setValue(0);
   }
   else
   {
-    if (n == 1)
+    if (p->maximum() < n)  // increase max
     {
-      if (p->maximum() == 0)
-        p->setValue(0);
-      else
-        p->setValue(p->maximum() - 1);
+      p->setMaximum(n);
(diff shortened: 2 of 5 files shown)


rhaschke:
Regarding item no. 2, contact markers, they disappeared automatically after 60s.

v4hn:
I just noticed I did not review this yet.

>  I'm not satisfied though, with hiding the "nullspace exploration" label as it just flickers up and one cannot read it anymore.

This is somewhat confusing to hear from you, as you are usually the most proficient qt-developer here. Did you just add the condition in the wrong place or is there some deeper reason for this behavior? I don't want to merge this yet as the patch suggests it would work now...

rhaschke:
This is not an issue with Qt, but a semantic problem: Close to singularities, the number of nullspace dimensions will vary between non-zero (in singularities) and zero. This causes the label to flicker.
I propose to _always_ keep the label - as before.

v4hn (approved):
I see! Up to now I assumed this is mainly designed for redundant arms and did not consider singularities.

But the current implementation has that problem already, doesn't it? Away from the singularity the sliders will dynamically change as well and flicker in the GUI. I don't know how to adapt the GUI best to avoid this "surprise".

For me, this patch improves the interface and I would like to merge this (unless you object).

rhaschke:
You are right, the old implementation also flickers the sliders close to a singularity. However, at least the label is always visible, such that the user can understand what is flickering. If the label is flickering as well, that's not possible anymore.

v4hn:
> You are right, the old implementation also flickers the sliders close to a singularity.
> However, at least the label is always visible, such that the user can understand what is flickering.

The label jumps up and down though, which is far from good design. :-)

I guess we agree that neither solution is ideal. How about an alternative then. The main reason I consider this an issue is because there is no functionality - nor a clear indication on the requirements for functionality. My first intuition was that this will only enable for 7+dof arms and is thus useless for 6dof Systems. So how about adding a message when no slider is available saying something like "(available near singularities)"?

rhaschke:
> The label jumps up and down though, which is far from good design.

I don't see how you want to avoid the jumping: If the GUI changes layout (for semantic reasons), it obviously needs to be adjusted.

> So how about adding a message when no slider is available saying something like "(available near singularities)"?

Sure, I can add a message. Do you agree with a tooltip hint?

v4hn:
On Thu, Nov 18, 2021 at 02:09:35AM -0800, Robert Haschke wrote:
> > So how about adding a message when no slider is available saying something like "(available near singularities)"?
> 
> Sure, I can add a message. Do you agree with a tooltip hint?

No, I would strictly prefer a message in place of the slider that would otherwise be available.
If we keep a single slider around but disable it, a tooltip for it might be an alternative.

simonschmeisser:
Somewhat off-topic, but could we add the group name to the title?

Also when I click and hold the slider, the robot "scans" the null-space, ie J4 keeps increasing at 1Hz and J6 decreasing  (opw_kinematics)

Finally I think hiding the label as well (this PR) is better than showing only the label (master) but I would suggest experimenting with disabling the slider (and adding a tooltip)

v4hn:
> Somewhat off-topic, but could we add the group name to the title?

The current design on multiple tabs (including joints and stored states) is not particularly good in general and I don't think such minor cosmetic changes really help there...

> Also when I click and hold the slider, the robot "scans" the null-space, ie J4 keeps increasing at 1Hz and J6 decreasing  (opw_kinematics)

Yes. That's the intended use I guess? I would have something else in mind for that feature (how about a checkbox and afterwards let the user drag individual joints that can be modified in nullspace?). But I do not have time for such features and I believe the current implementation is more than I would expect in this tab already. :+1:

rhaschke:
> Also when I click and hold the slider, the robot "scans" the null-space

Indeed, that's the intended functionality. You can only jog the null-space in that fashion because there is no absolute coordinate system available for it.


> let the user drag individual joints that can be modified in the null-space

The null-space isn't aligned with individual joints! Thus, it doesn't make sense to scan the null-space with individual joints.
The current implementation provides a single slider for each basis vector of the (linear) null-space. Sure, if the null-space is multi-dimensional (i.e. there are several sliders), one could try to optimally align those basis vectors with specific joints. But, how do you decide which joints to align with?

rhaschke:
@v4hn, I implemented the suggested tooltip and dummy slider.

v4hn:
> The null-space isn't aligned with individual joints! Thus, it doesn't make sense to scan the null-space with individual joints.

Yes, but you could still project an individual joint motion request (vector in joint space for a single joint that is relevant in the current null-space) onto a vector in the linear null-space, which will then move other joints together with it. I don't see why that would not be feasible?
The current implementation is nice, it's just not very user-friendly when you want to reach a specific joint value in null space.

rhaschke:
So, you want to have an interface to get as close as possible to the given joint configuration.
That's indeed something completely different. The user should be aware, that it's probably not possible to actually reach the joint config. Thus, one would need to somehow visualize the actual and the desired joint config...

v4hn:
> So, you want to have an interface to get as close as possible to the given joint configuration.

That's not what I described, though it might also be useful with good graphical feedback on what values are feasible (non-trivial as you said).

I would be happy with the same kind of "move in this direction" slider for each involved joint instead of each null-space basis vector. The sliders would become redundant of course, but behavior much more intuitive. E.g., "move this joint a bit further left in null-space". I would also expect that cases with 2+ null-space dimensions would behave much better than they currently do.

Anyway, this was just brainstorming on my side for now and does not belong here to begin with... If you think anyone is interested in this feel free to turn it into an issue.

PR #2951, opened 2021-11-06
Melodic: Use moveit-resources@master

rhaschke:
To simplify maintenance of https://github.com/ros-planning/moveit_resources, I would like to use its master branch for both, MoveIt `melodic-devel` and `master`/`noetic-devel`. AFAIK, melodic-devel doesn't provide the `ResolveConstraintFrames` adapter, which is configured in `moveit_resources@master`. However, not having it available just issues an error when launching move_group, but it shouldn't harm.

Besides paving the way to use `moveit_resources@master` (and thus unifying + simplifying many launch files) this PR provides compatibility to the new way of defining (multiple) planning pipelines. However, only a single pipeline can be used, specified via the `~default_planning_pipeline` parameter.


rhaschke:
I had to merge clang-tidy fixes from #2971 to fx CI issues (from clang-tidy). This is ready for review, @v4hn.
This PR is ideally merged together with #2971 by fast-forwarding the melodic-devel branch from the command line (the final merge commit already has a nice commit message mentioning both PRs).

v4hn:
These changes are somewhat invasive and moving `melodic-devel` to `moveit_resources#master` will force us to stay compatible with melodic in the future - or revert to the separate branches again soon.
Note that we didn't review backports for melodic in a long time now and de-facto moved to backport-on-user-request for the branch. I don't see us adding backward compatibility for every new feature in master the way you added support for the `default_planning_pipeline` here.

rhaschke:
Being able to use the new planning pipeline config in MoveIt Melodic will be beneficial not only for moveit_resources, but also for other moveit_config projects. For this reason, I consider this PR as a minimal backport of this functionality.
I agree, that we might need to branch off a melodic-devel branch in moveit_resources in the future again. But for now, we could simplify maintenance by merging.

rhaschke:
Ping @v4hn, @henningkayser. I would really like to merge this.

moveit_ros/move_group/src/move_group_context.cpp
```diff
@@ -40,14 +40,26 @@
...
 move_group::MoveGroupContext::MoveGroupContext(
     const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor, bool allow_trajectory_execution,
     bool debug)
   : planning_scene_monitor_(planning_scene_monitor)
   , allow_trajectory_execution_(allow_trajectory_execution)
   , debug_(debug)
 {
-  planning_pipeline_.reset(new planning_pipeline::PlanningPipeline(planning_scene_monitor_->getRobotModel()));
+  planning_pipeline_.reset(
+      new planning_pipeline::PlanningPipeline(planning_scene_monitor_->getRobotModel(), getConfigNodeHandle()));
```
v4hn, line 63: This will change the namespace of `display_planned_path` and thus break rviz configs in melodic.

moveit_ros/move_group/src/move_group.cpp
```diff
@@ -128,15 +130,15 @@ class MoveGroupExe
 
     // add capabilities listed in ROS parameter
     std::string capability_plugins;
-    if (node_handle_.getParam("capabilities", capability_plugins))
+    if (nh.getParam("capabilities", capability_plugins))
```
v4hn, line 133: capabilities and disabled_capabilities do not use the `default_planning_pipeline` namespaces. This breaks specifying custom or disabling standard capabilities.

v4hn (commented):
I do like the cleanup you do here and it makes sense to add it to melodic.
But as I already expressed above I'm no fan of sharing the same branch for moveit_resources anymore and would prefer to see the relevant cleanup ported to `moveit_resources`'s `melodic-devel` branch. I guess I'm mostly unhappy with 032541cdade0fb29f5bd9b589227793ae315925b which you need to support new configs. It's easy to introduce simple regressions in such "backports", as the ones I pointed out there.

I am currently not involved with melodic configurations, so I am not likely to spend time on this.
So if you want to force it, feel free to merge it after fixing the issues I pointed out in this review.
If there is no further regression this can probably be useful to someone - though I would urge them to compile moveit noetic/master when they already have a config for it.

moveit_ros/move_group/src/move_group_context.cpp
```diff
@@ -40,14 +40,27 @@
...
 #include <moveit/plan_execution/plan_with_sensing.h>
 
-move_group::MoveGroupContext::MoveGroupContext(
-    const planning_scene_monitor::PlanningSceneMonitorPtr& planning_scene_monitor, bool allow_trajectory_execution,
-    bool debug)
+namespace move_group
+{
+ros::NodeHandle MoveGroupContext::getConfigNodeHandle()
+{
+  ros::NodeHandle nh = ros::NodeHandle("~");
```
henningkayser, line 47:
```suggestion
  ros::NodeHandle nh("~");
```

henningkayser: nit: I think it would be more elegant to initialize `default_pipeline` with "~" and query the parameter afterwards, so that `nh` is only initialized once with the resulting namespace.

moveit_ros/move_group/src/move_group_context.cpp
henningkayser: Also, should this be switched to `std::make_shared`?

henningkayser (commented):
Overall, the simplifications make a lot of sense. I just have a couple minor things to add besides @v4hn's comments.

moveit_ros/move_group/src/move_group_context.cpp
rhaschke: Switching to `std::make_shared` is unrelated to this PR topic, isn't it?

moveit_ros/move_group/src/move_group_context.cpp
rhaschke: Sorry, I don't yet understand your suggestion. `default_pipeline` is a ROS parameter denoting the desired pipeline, e.g. `ompl`. How would initializing it with `"~"` help? Maybe just draft your suggested change?

rhaschke:
Looks like I was too optimistic in providing a lightweight compatibility layer. Comparing parameter names more closely, I noticed several issues and abandoned the original idea. I need to check, which parts of the cleanup are still usable.

rhaschke:
I will give this another try...

moveit_ros/move_group/src/move_group.cpp
rhaschke: You are right. I was misled by the fact that the Pilz planner loads its specific capabilities from its own namespace.
Works now.

rhaschke:
I think I found a clean solution to support the new configuration scheme (9b622a7): essentially, we just need to ensure that planning pipeline parameters, as well as adapter parameters, are read from the corresponding namespace.
The corresponding node handles are now initialized via `planning_interface::getConfigNodeHandle()`.

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
```diff
@@ -55,11 +55,12 @@ planning_pipeline::PlanningPipeline::PlanningPipeline(const robot_model::RobotMo
   : nh_(nh), robot_model_(model)
 {
   std::string planner;
-  if (nh_.getParam(planner_plugin_param_name, planner))
+  ros::NodeHandle cnh = planning_interface::getConfigNodeHandle(nh);
```
henningkayser, line 58: I still don't really follow why `getConfigNodeHandle()` is being used here. If a default pipeline is configured, `getConfigNodeHandle` will always return a node handle in the namespace of the default pipeline. With this change, non-default pipelines wouldn't be loaded anymore. If anything, the "fixed" namespace should be passed with the node handle from the constructor.

rhaschke: `default_planning_pipeline` is a parameter introduced by your new config scheme via #2127. If this *and* the corresponding pipeline namespace `move_group/planning_pipelines/$default_pipeline` exist, `getConfigNodeHandle()` will return a corresponding NodeHandle, thus using the config from the new scheme. Otherwise, it just falls back to the originally passed (private) move_group namespace, thus resorting to the old/existing scheme.

> With this change, non-default pipelines wouldn't be loaded anymore.

The parameter `default_planning_pipeline` was newly introduced in #2127 and **Melodic** doesn't know about this parameter (yet). So what exactly do you mean by loading non-default pipelines?
So far, Melodic can only load a single pipeline, the one configured in the private move_group namespace.
This PR allows using the new config scheme with multiple namespaces below `move_group/planning_pipelines/*`, but only a single one can be used at a time. I explicitly didn't want to support switching.

> The "fixed" namespace should be passed with the node handle from the constructor

What do you mean by _fixed_ namespace?

henningkayser: > I explicitly didn't want to support switching.

Oh, I didn't get that. So you are intentionally only allowing a single default pipeline instance. I'm fine with that if we really don't need that feature in melodic.

> What do you mean by fixed namespace?

With the assumption that melodic should support multiple pipelines, I thought that the namespace would better be handled by the code calling the pipeline constructor. With "fixed" I meant the switching of the namespace inside `getConfigNodeHandle()`.

rhaschke: > Oh, I didn't get that.

I was afraid that you misconceived my intention... Good, that we are on the same page now.
Do you see other issues or do you approve?

henningkayser (approved):

moveit_ros/planning/planning_pipeline/src/planning_pipeline.cpp
henningkayser: Looks good to me, thanks!

PR #2952, opened 2021-11-07
GHA: auto-sync noetic-devel to master

rhaschke:
Added new job to automatically fast-forward push noetic-devel to master after a successful push of master on ros-planning. 
This will automate the process of keeping master and noetic-devel in sync.

Note, the job will not run on pull-requests, different branches than master being pushed, or on forked repos.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -158,4 +158,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Fast-forward noetic-devel to sync with master
-        run: git push --ff-only HEAD:noetic-devel
+        run: git push origin HEAD:noetic-devel
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,6 @@ on:
   push:
     branches:
       - master
-      - "[kmn]*-devel"
 
 jobs:
   default:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -9,7 +9,6 @@ on:
   push:
     branches:
       - master
-      - "[kmn]*-devel"
```
v4hn, line 12: I see why you would you disable the workflow for noetic, but why do you not want to run this on k/m pushes anymore?
Even if the pull-requests will be tested independently I would also test pushes.

rhaschke: This file _is part of_ the `master` branch. The kinetic-devel and melodic-devel branches have their own CI config.
Thus, the setting was non-sense from the very beginning. However, it allowed to easily compare/share the config between those branches :wink:

v4hn:
This action currently fails because of the protected branch. Is there some way to allow the action to push while everything else should still require the review?

> remote: error: GH006: Protected branch update failed for refs/heads/noetic-devel.        
> remote: error: Required status check "WIP" is expected. At least 1 approving review is required by reviewers with write access.        
> To https://github.com/ros-planning/moveit
> ! [remote rejected] master -> noetic-devel (protected branch hook declined)

rhaschke:
I have seen that :disappointed: The problem didn't show up in my tests, unfortunately.
I see two options:
- Disable required checks for noetic-devel branch (but keep them for master, of course).
- Setup a secret token to push as rhaschke, who has correct permissions (preferred).

rhaschke:
Looks like my suggested change [still doesn't work](https://github.com/ros-planning/moveit/runs/4140416981). The only way is to [relax required checks](https://github.com/semantic-release/github/issues/175#issuecomment-891108176).

PR #2957, opened 2021-11-11
Add API stress tests for TOTG, fix undefined behavior

henningkayser:
I think this test shouldn't fail, but it does for me. I have only found minor issues in the TOTG code (will commit those) but weren't able to successfully get this passing. Has anyone else seen something similar?

```
$ ( source /home/runner/work/moveit/moveit/.work/upstream_ws/install/setup.bash && cd /home/runner/work/moveit/moveit/.work/target_ws && catkin_test_results --verbose; )
4731 Full test results for 'build/moveit_core/test_results/moveit_core/gtest-test_time_optimal_trajectory_generation.xml'
4732 -------------------------------------------------
4733 <?xml version="1.0" encoding="UTF-8"?>
4734 <testsuites tests="5" failures="1" disabled="0" errors="0" timestamp="2021-11-11T17:36:45" time="1.847" name="AllTests">
4735   <testsuite name="time_optimal_trajectory_generation" tests="5" failures="1" disabled="0" errors="0" time="1.847">
4736    <testcase name="test1" status="run" time="0" classname="time_optimal_trajectory_generation" />
4737    <testcase name="test2" status="run" time="0.302" classname="time_optimal_trajectory_generation" />
4738    <testcase name="test3" status="run" time="1.481" classname="time_optimal_trajectory_generation" />
4739    <testcase name="testLargeAccel" status="run" time="0.001" classname="time_optimal_trajectory_generation" />
4740    <testcase name="testPluginAPI" status="run" time="0.062" classname="time_optimal_trajectory_generation">
4741      <failure message="/home/runner/work/moveit/moveit/.work/target_ws/src/moveit/moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp:270&#x0A;Value of: totg.computeTimeStamps(trajectory)&#x0A;  Actual: false&#x0A;Expected: true&#x0A;Failed to compute timestamps on a new TOTG instance in iteration 2" type=""><![CDATA[/home/runner/work/moveit/moveit/.work/target_ws/src/moveit/moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp:270
4742  Value of: totg.computeTimeStamps(trajectory)
4743  Actual: false
4744  Expected: true
4745  Failed to compute timestamps on a new TOTG instance in iteration 2]]></failure>
4746    </testcase>
4747   </testsuite>
4748 </testsuites>
4749
4750 -------------------------------------------------
4751 build/moveit_core/test_results/moveit_core/gtest-test_bullet_continuous_collision_checking.xml: 8 tests, 0 errors, 0 failures, 2 skipped
4752 build/moveit_core/test_results/moveit_core/gtest-test_time_optimal_trajectory_generation.xml: 10 tests, 0 errors, 2 failures, 0 skipped
4753 build/moveit_ros_planning_interface/test_results/moveit_ros_planning_interface/nosetests-serialize_msg.py.xml: 12 tests, 0 errors, 0 failures, 1 skipped
4754 Summary: 1614 tests, 0 errors, 2 failures, 3 skipped
```


AndyZe:
Quite possibly caused by the identical waypoint. Try it without the identical waypoint

  waypoint_state.setJointGroupPositions(group, { -0.5, -3.52, 1.35, -2.51, -0.88, 0.63, 0.0 });
  trajectory.addSuffixWayPoint(waypoint_state, 0.1);
  waypoint_state.setJointGroupPositions(group, { -0.5, -3.52, 1.35, -2.51, -0.88, 0.63, 0.0 });

henningkayser:
> Quite possibly caused by the identical waypoint. Try it without the identical waypoint
> 
> waypoint_state.setJointGroupPositions(group, { -0.5, -3.52, 1.35, -2.51, -0.88, 0.63, 0.0 }); trajectory.addSuffixWayPoint(waypoint_state, 0.1); waypoint_state.setJointGroupPositions(group, { -0.5, -3.52, 1.35, -2.51, -0.88, 0.63, 0.0 });

That's exactly what I'm trying to debug. The problem is that the results are not consistent. If this was about the identical waypoints, all the calls to `computeTimeStamps()` should fail. But they only fail after calling TOTG repeatedly on new instances.

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
```diff
@@ -227,6 +229,49 @@ TEST(time_optimal_trajectory_generation, testLargeAccel)
...
+    TimeOptimalTrajectoryGeneration totg;
+    ASSERT_TRUE(totg.computeTimeStamps(test_trajectory)) << "Failed to compute time stamps";
+    EXPECT_TRUE(totg.computeTimeStamps(test_trajectory))
+        << "Failed to compute time stamps on the already parameterized trajectory with the same TOTG instance";
+  }
+
+  for (size_t i = 0; i < 100; ++i)
+  {
+    TimeOptimalTrajectoryGeneration totg;
+    ASSERT_TRUE(totg.computeTimeStamps(trajectory))
```
henningkayser, line 270: For me, it always fails in the second iteration of this call

AndyZe: OK, I don't know why that would happen. But I'd suggest adding a bit of code to TOTG to remove identical waypoints, if it doesn't have it already.

There can be lots of weird solutions between identical waypoints, including:

- Remain stationary
- Make a circle, arriving back at the start point

The new Ruckig plugin removes identical waypoints here:  https://github.com/ros-planning/moveit2/blob/54c6e08ead8d5888198cb819cd933cb6229a3077/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp#L144

henningkayser: It does remove identical waypoints, though https://github.com/ros-planning/moveit/blob/master/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp#L959

AndyZe: Maybe it's the joints that do not change that cause the issue? For those joints, the waypoint is still identical, I guess.

Maybe you should skip TOTG for joints that do not change between waypoints.

Or maybe it's a red herring...

henningkayser: I think it's something else. If you look at the lines 258-265, you can see that generating the same trajectory repeatedly works without issues, even using the same TOTG instance. But only, if you use a new TOTG instance every time (for the same trajectory here), it fails consistently. To me this smells like some issue with static memory or so. The waypoints are supported (as proven by the test), the velocity computation is flaky.

henningkayser:
With the last change, the test now runs a hundred cycles on two deep copies of the original trajectory. One trajectory is being iterated over the same TOTG instance, the other one gets a fresh instance with each cycle. All pass, they produce the same trajectory profiles. After that, the original trajectory from which the first two have been copied from is being iterated again a hundred times over fresh TOTG instances, and fails consistently after the first or second attempt. It should really not fail since the same thing has just been proven to work with the deep copies. What's wrong here? Is the deep-copy broken in a weird way? The consistency of this happening also doesn't look like some undefined behavior.

AndyZe:
Does the new version of the tests with deep copies and hundreds of iterations pass if the waypoints are all unique?

I mean, completely unique. Every joint angle different for every waypoint?

henningkayser:
> Does the new version of the tests with deep copies and hundreds of iterations pass if the waypoints are all unique?
> 
> I mean, completely unique. Every joint angle different for every waypoint?

Yes, they pass. But that's not what I'm trying to debug here. I picked the waypoints on purpose like that because they expose a general flakiness that I've observed in the past. The flakiness occurs mostly when there are trajectories that include a back-and-forth motion (A->B->A). The test proofs that such trajectories are valid, and in general TOTG produces valid and even completely equal outcomes as proven by the first two deep copies. They must be equal since TOTG should be a completely deterministic algorithm. If it fails in some cases, maybe even under specific circumstances like the example here, it means that there is an undefined behavior somewhere in the algorithm or even in RobotTrajectory, RobotState or somewhere else (maybe even in the test). I ran asan, msan, ubsan, and lsan on TOTG only and didn't find any fixes for this. Next step is to extent that to related classes in moveit_core. The reason I'm concerned about this is that it exposes undefined behavior which may lead to undefined outcomes in general, even in seemingly valid trajectories. Side note: I was only able to reproduce this using the `TimeOptimalTrajectoryGeneration` API and not with the underlying helper classes.

AndyZe:
Adding these few lines (which really don't do anything) causes the test to pass.

```
    for (size_t idx = 0; idx < trajectory.getWayPointCount(); ++idx)
    {
      robot_state::RobotState original_state = trajectory.getWayPoint(idx);
    }
```

AndyZe:
Ah, I think it might be that a deep copy calls `clear()`, which clears the waypoints and `duration_from_previous_`. Then it re-inserts the waypoints but it does not re-insert `duration_from_previous_`. So all of these deep copies are missing duration info.

https://github.com/ros-planning/moveit/blob/4db626d9c3b5d6296b012188a4a0bfe4bddf6bce/moveit_core/robot_trajectory/src/robot_trajectory.cpp#L61

https://github.com/ros-planning/moveit/blob/4db626d9c3b5d6296b012188a4a0bfe4bddf6bce/moveit_core/robot_trajectory/include/moveit/robot_trajectory/robot_trajectory.h#L232

AndyZe:
Pushed a fix that works for me locally. Hope it works for you, too

AndyZe:
We both missed something pretty obvious: the first 2 loops break out of the loop if successful:

```
      if (totg_success)
        break;
```
        
 The third loop doesn't. But, there is still something else to fix.

AndyZe:
I suspect the vel/accel limits aren't being read successfully because the first 2 loops finish at 43 waypoints. The third loop finishes in 6 waypoints. They all end at the right positions, though.

To me, that indicates either durations aren't copied properly or vel/accel limits aren't copied properly.

AndyZe:
Wait, why does `first_trajectory_msg_start` have 45 waypoints? That is not right -- somehow it is storing the output from TOTG, not the pre-TOTG state.

AndyZe:
It looks like there are some improvements here. The test is only failing about 20% of the time instead of 30% before I started working on it, and I made the tests harder. Still frustrating that it's not 100% reliable.

I'll clean up my commit history in a bit.

AndyZe:
This passes 10/10 times locally for me, now.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -963,21 +963,29 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+    double duration = trajectory.getWayPointDurationFromPrevious(p);
     // The first point should always be kept
     bool diverse_point = (p == 0);
 
     for (size_t j = 0; j < num_joints; ++j)
     {
       new_point[j] = waypoint->getVariablePosition(idx[j]);
-      // If any joint angle is different, it's a unique waypoint
-      if (p > 0 && std::fabs(new_point[j] - points.back()[j]) > min_angle_change_)
+      // If any joint angle is different and duration is nonzero, it's a unique waypoint
```
henningkayser, line 973: @AndyZe can you explain how this is supposed to fix the test case? All waypoints have a nonzero timestamp so this change shouldn't affect the outcome. Also, TOTG should not consider existing timestamps as it's performing a complete resampling and recomputing of timestamps, it doesn't take existing times into account.

henningkayser:
> This passes 15/15 times locally for me, now.

So what was the bug? From the changes I don't understand what would have been the cause.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: OK, that was a mistake. Will revert.

Please test locally, though. It works well for myself and one other.

AndyZe:
I'm not sure there is/was a bug. I think the test conditions weren't correct originally and that was causing the failure in the new tests. Some of the loops had this, some did not:

```
      if (totg_success)
        break;
```

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
```diff
@@ -227,6 +229,98 @@ TEST(time_optimal_trajectory_generation, testLargeAccel)
...
+
+    // Iteratively recompute timestamps with new totg instances
+    for (size_t i = 0; i < totg_iterations; ++i)
+    {
+      TimeOptimalTrajectoryGeneration totg;
+      bool totg_success = totg.computeTimeStamps(test_trajectory);
+      EXPECT_TRUE(totg_success) << "Failed to compute time stamps with a new TOTG instance in iteration " << i;
+
+      if (totg_success)
+        break;
```
AndyZe, line 304: @henningkayser you force-pushed over a commit of mine that fixes this exit condition. Probably none of the loops should have the exit condition

AndyZe (changes requested):

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
henningkayser: Thanks, fixed

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: I think I added this comment about duration, but it should be removed.

`// If any joint angle is different, it's a unique waypoint`

AndyZe (approved):

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -962,22 +963,34 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
-    Eigen::VectorXd new_point(num_joints);
+    Eigen::VectorXd new_point = Eigen::VectorXd::Zero(num_joints);
+
     // The first point should always be kept
-    bool diverse_point = (p == 0);
+    if (p == 0)
+    {
+      points.push_back(new_point);
+      continue;
+    }
```
jspricke, line 973: this is wrong, it adds a new zero vector whereas the original code adds the first vector, I have internal code that is broken by it. I would propose to revert all this.

Issue #2958, opened 2021-11-12
Add path_constraints to moveit_cpp interface


Colin-Kohler:
There is no current way to set the path_constraints field of a MotionPlanRequest when using the moveit_cpp interface. Recently, this feature was added to moveit2: https://github.com/ros-planning/moveit2/pull/752. Should be simple enough to port those changes over to moveit with minor changes. I have a forked version of moveit, that does exactly this and will submit a PR to incorporate these changes.

JafarAbdi:
Fixed by https://github.com/ros-planning/moveit/pull/2959

PR #2959, opened 2021-11-12
MoveitCpp - added ability to set path constraints for PlanningComponent.

Colin-Kohler:
### Description

PR for issue [2958](https://github.com/ros-planning/moveit/issues/2958). Only minor difference from the corresponding moveit2 implementation is the change in the moveit_msg definitions.


v4hn (approved):
straight-forward and useful addition. I have no idea why nobody merged it right away :sunglasses:

PR #2961, opened 2021-11-15
Add waypoint duration to the trajectory deep copy unit test

AndyZe:
Previously, only joint positions were tested.

This should settle the discussion [here](https://github.com/ros-planning/moveit/pull/2957#issuecomment-967195498) (PR #2957). The duration deep copy was correct, it seems.

### DIFF
--- a/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
+++ b/moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
@@ -118,10 +118,11 @@ protected:
 
     // Modify the first waypoint duration
     double trajectory_first_duration_before_update = trajectory->getWayPointDurationFromPrevious(0);
-    trajectory->setWayPointDurationFromPrevious(0, trajectory_first_duration_before_update + 0.1);
+    double new_duration = trajectory_first_duration_before_update + 0.1;
+    trajectory->setWayPointDurationFromPrevious(0, new_duration);
 
     // Check that the trajectory's first duration was updated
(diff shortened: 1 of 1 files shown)


moveit_core/robot_trajectory/test/test_robot_trajectory.cpp
```diff
@@ -115,6 +115,13 @@ class RobotTrajectoryTestFixture : public testing::Test
     std::vector<double> trajectory_first_state_after_update;
     trajectory_first_waypoint_after_update->copyJointGroupPositions(arm_jmg_name_, trajectory_first_state_after_update);
     EXPECT_EQ(trajectory_first_state[0], trajectory_first_state_after_update[0]);
+
+    // Modify the first waypoint duration
+    double trajectory_first_duration_before_update = trajectory->getWayPointDurationFromPrevious(0);
+    trajectory->setWayPointDurationFromPrevious(0, trajectory_first_duration_before_update + 0.1);
+
+    // Check that the trajectory's first duration was updated
+    EXPECT_NE(trajectory->getWayPointDurationFromPrevious(0), trajectory_first_duration_before_update);
```
v4hn, line 124: Why would you compare with `NE` when you know the value it's actually supposed to be? :smiley:

AndyZe: I'm just copying what was done at L249, for example. Very well, I'll update this.

v4hn (approved):

PR #2962, opened 2021-11-15
TOTG: reject waypoints that have a duration of zero

AndyZe:
I haven't seen that this is an issue but might as well add a check for it.

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -971,19 +971,21 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
     {
       new_point[j] = waypoint->getVariablePosition(idx[j]);
       // If any joint angle is different and duration is nonzero, it's a unique waypoint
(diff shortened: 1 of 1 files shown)


moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -963,14 +963,15 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
   {
     moveit::core::RobotStatePtr waypoint = trajectory.getWayPointPtr(p);
     Eigen::VectorXd new_point(num_joints);
+    double duration = trajectory.getWayPointDurationFromPrevious(p);
     // The first point should always be kept
     bool diverse_point = (p == 0);
```
AndyZe, line 968: Note that the first waypoint will be kept regardless of its "DurationFromPrevious"

v4hn:
I'm not sure I understand what you are doing here. TOTG is *supposed* to yield a time parameterization, *including* the duration_from_previous values. Shouldn't it be perfectly fine to pass in an unparameterized trajectory with 0 values for the duration fields (i.e. a path)?

AndyZe:
@v4hn you are right about identical timestamps being OK. Will close this PR and combine with #2957.

PR #2963, opened 2021-11-16
fix(msa): fix urdf double load bug

rickstaa:
This commit fixes a bug that prevented users from loading a correct URDF file after they loaded an incorrect URDF file first. This was caused since the `urdf_string_` was never cleared (see gif).

urdf_load_bug

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -159,6 +159,10 @@ jobs:
       - uses: actions/checkout@v2
       - name: Fast-forward noetic-devel to sync with master
         run: |
+          # Configure push user+url
+          git config --global user.name 'GHA sync bot'
+          git config --global user.email 'rhaschke@users.noreply.github.com'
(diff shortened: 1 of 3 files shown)


rhaschke (commented):
Thanks for hinting at this problem. I suggest a solution more upstream in `RDFLoader::loadXacroFileToString()` though.
I'm sorry for the commit mess. I accidentally reverted the wrong commit first...

rhaschke (approved):

rickstaa:
@rhaschke My first intention was to fix it in the `RDFLoader::loadXacroFileToString()`. I however was cautious doing so since this breaks code that uses the `RDFLoader::loadXacroFileToString()` to concatenate xacro files. I however agree with you that fixing it there is better as people should not use the `RDFLoader::loadXacroFileToString()` function in that way.

> I'm sorry for the commit mess. I accidentally reverted the wrong commit first...

No problem, these things also happen to me. Do you want me to fix it or merge it in the current state?

rhaschke:
I will squash-merge as is if CI is happy.

PR #2964, opened 2021-11-16
fix(msa): fix modified files bug

rickstaa:
This pull request fixes an MSA bug that caused files in a loaded MoveIt config to incorrectly be classified as externally modified. This happened when the `Generate Package` button is clicked multiple times and we go back to other tabs.


rhaschke (approved):
LGTM.

rickstaa:
@rhaschke Thanks for merging this. Should we also apply this change to https://github.com/ros-planning/moveit2/blob/b426a1c6f957522f8261f3500ee11f82199cdef6/moveit_setup_assistant/src/tools/moveit_config_data.cpp#L185?

rhaschke:
I'm not sure what you are referring to? Forward-porting to MoveIt2? @henningkayser will take care of this...

rickstaa:
@rhaschke Ah great!

henningkayser:
> I'm not sure what you are referring to? Forward-porting to MoveIt2? @henningkayser will take care of this...

Yes, we're still working on catching up but this will be in MoveIt 2, eventually. However, MSA ist being redesigned currently, so I'm not sure how applicable the changes are. @DLu  any thoughts?

PR #2971, opened 2021-11-19
Apply clang-tidy fixes

rhaschke:
Apply most clang-tidy fixes. However, ignore `readability-identifier-naming` as this would modify API.


v4hn (approved):
looks good for all I can see.

Issue #2766, opened 2021-07-10
avoid `boost`/`std::bind`


v4hn:
Similar in spirit to #2757, here is another nagging detail scattered throughout the code base:

Scott Meyer's "Effective Modern C++" Item 34 as well as [TOTW 108](https://abseil.io/tips/108) state:

> Avoid `boost`/`std::bind` [in favor of lambdas where applicable]

MoveIt uses `bind` in various places and readability suffers from it.
We should update the definitions to use lambdas with properly named arguments instead!

tylerjw:
All the lambdas!

rhaschke:
There is clang-tidy modernize-avoid-bind: https://github.com/ros-planning/moveit/commit/29d2de102e47c5ab88207a624f91b2b6c8752ed6 :smile: I will prepare a PR if clang-tidy is done...

v4hn:
> There is clang-tidy modernize-avoid-bind

:open_mouth:  Cool!

rhaschke:
Unfortunately, it doesn't translate placeholders into arguments :disappointed:  I'm trying clang-tidy-12 now...

PR #2972, opened 2021-11-20
clang-tidy modernize-avoid-bind

rhaschke:
Address #2766 employing clang-tidy's `modernize-avoid-bind`. This has several shortcomings:
- `clang-tidy-10`, which is used by default in Ubuntu 20.04, doesn't turn placeholder arguments into lambda arguments :disappointed: 
- `clang-tidy-12` fixes this, but the generated argument names are PH1, PH2, etc. and their type is `auto&&`. Not very readible!
- `clang-tidy-12` doesn't yet consider return values of bound functions :disappointed:

@v4hn, should we really continue along that path?


v4hn:
> @v4hn, should we really continue along that path?

I know it's been a while, but if somehow possible I would like to add the clang-format flag, yes.
BUT of course someone needs to give the variables proper names and new contributors cannot just apply the suggestion from CI to fix their code.
The latter part is not a big deal for me because people should just realize that the names are not ok. The former part is a lot of work, but is basically waiting for someone to step up, lookup the function signatures and turn the lambdas into something readable and informative. :-)

PR #2976, opened 2021-11-25
Ignore Emacs modules auto-generated files

captain-yoshi:
### Description
Ignore auto-generated Emacs stuff. Useful for those who uses Emacs as their IDE.


rhaschke (commented):
I didn't know what a kind of mess people collected here! One can easily [git-ignore](https://git-scm.com/docs/gitignore) these files _globally_ in _all_ your git repos configuring a `core.excludesFile` in your `~/.gitconfig`:

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

IMO, there is no need to clutter a repo-specific `.gitignore` file with this :wink:

captain-yoshi:
Agreed ! Did not know you could configure it globally. Thx for the tip.

PR #2977, opened 2021-11-26
Add marker for subgroups even if no endeffector is defined for them

v4hn:
For single groups, the old logic fell back to add a marker
for the last link if IK is supported for it and no endeffector is defined.

That (quite reasonable) fallback did not yet work for subgroups though.

For a subgroup definition like this in the srdf:
```
<group name="hand_all">
  <group name="hand_mf_group" />
  <group name="hand_rf_group" />
  <group name="hand_th_group" />
</group>
```
Before:
no-markers

After:
markers

### DIFF
--- a/moveit_ros/robot_interaction/src/robot_interaction.cpp
+++ b/moveit_ros/robot_interaction/src/robot_interaction.cpp
@@ -280,56 +280,53 @@ void RobotInteraction::decideActiveEndEffectors(const std::string& group, Intera
   const std::pair<moveit::core::JointModelGroup::KinematicsSolver, moveit::core::JointModelGroup::KinematicsSolverMap>&
       smap = jmg->getGroupKinematics();
 
-  // if we have an IK solver for the selected group, we check if there are any end effectors attached to this group
-  if (smap.first)
-  {
+  auto addActiveEndEffectorsForSingleGroup = [&](const moveit::core::JointModelGroup* single_group) {
+    bool found_eef{ false };
     for (const srdf::Model::EndEffector& eef : eefs)
-      if ((jmg->hasLinkModel(eef.parent_link_) || jmg->getName() == eef.parent_group_) &&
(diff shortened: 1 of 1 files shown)


rhaschke (approved):

moveit_ros/robot_interaction/src/robot_interaction.cpp
```diff
@@ -280,56 +280,53 @@ void RobotInteraction::decideActiveEndEffectors(const std::string& group, Intera
   const std::pair<moveit::core::JointModelGroup::KinematicsSolver, moveit::core::JointModelGroup::KinematicsSolverMap>&
       smap = jmg->getGroupKinematics();
 
-  // if we have an IK solver for the selected group, we check if there are any end effectors attached to this group
-  if (smap.first)
-  {
+  auto addActiveEndEffectorsForSingleGroup = [&](const moveit::core::JointModelGroup* single_group) {
```
rhaschke, line 283: clang-tidy complains about this variable name as lambda are considered as variables, not functions :disappointed: 
I couldn't find an option to convince clang-tidy that lambda are actually functions.

v4hn: Yes, we've seen this issue before. It would be nice to add a new family to clang-tidy for lambdas.

I thought I'd file an issue for it, but It seems https://bugs.llvm.org/ is currently offline due to a github migration.

I already wrote the pushed fixup yesterday, but could not push because github was down.

They all hate me :expressionless: .

Anyway, the best no-NOLINT solution I could come up with would be to use [`LocalVariableIgnoredRegexp`](https://clang.llvm.org/extra/clang-tidy/checks/readability-identifier-naming.html#cmdoption-arg-localvariableignoredregexp) and enforce that closures follow a naming convention, e.g., `lambdaAddActiveEndEffectorsForSingleGroup`. But that would be an ugly and fragile workaround.

PR #2983, opened 2021-12-07
fix round_collada_numbers.py by opening files in binary mode

tbazina:
### Description

Minor change to fix round_collada_numbers.py script by opening files for reading and saving in binary mode. No reference issue.

### DIFF
--- a/moveit_kinematics/ikfast_kinematics_plugin/scripts/round_collada_numbers.py
+++ b/moveit_kinematics/ikfast_kinematics_plugin/scripts/round_collada_numbers.py
@@ -44,7 +44,6 @@ from __future__ import print_function
 from lxml import etree
 import shlex
 import sys
-import io
 
 
 def doRound(values, decimal_places):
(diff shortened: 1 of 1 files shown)


rhaschke:
Could you please describe the problem with the current approach? Collada files are text files (XML). So why open them in binary mode?

tbazina:
When running:
`rosrun moveit_kinematics round_collada_numbers.py <file_name>.dae <file_name>_rounded.dae 5`

the error traceback is:
```
Collada Number Rounder
Rounding numbers to 5 decimal places

  File "/opt/ros/noetic/lib/moveit_kinematics/round_collada_numbers.py", line 97, in <module>
    dom = etree.parse(io.BytesIO(xml))
TypeError: a bytes-like object is required, not 'str'
```
Per io documentation (https://docs.python.org/3/library/io.html#binary-i-o), "the easiest way to create a binary stream is with open() with 'b' in the mode string". After the fix, the script works fine.

rhaschke (approved):
Thanks for the explanation. I added another cleanup commit and tested on both, python2 and python3.

tbazina:
Great, thank you for the additional cleanup. Everything seems to be working now.

PR #2986, opened 2021-12-08
Hotfix: Add backwards compatibility for old scene serialization format

### LINKED ISSUE #2985: Scene format breakage between 1.1.5 and 1.1.6 in released distro (Noetic)
Hi, we are downstream users of the `.scene` format for describing collision scenes. Today, I've noticed that with the latest MoveIt from binary on Noetic, our unit tests and operational code that relied on scene loading started to fail (indicating a parsing error for the translation component).

It turns out that this is due to #2037 and documented there in the migration guide for Noetic:
> - Planning scene geometry text files (`.scene`) have changed format. Add a line `0 0 0 0 0 0 1` under each line with an asterisk to upgrade old files if required.

The issue is that #2037 was released into binaries last month as part of https://github.com/ros/rosdistro/pull/31147 (v1.1.6), following an earlier released version (v1.1.5) that used the old API.

One solution would be to update all `.scene` files following the migration guide in our code, with other end-users having to do the same (with the implication that we'd have to maintain separate `.scene` files for Melodic). However, since this targeted an already released distro, I wonder if adding backwards-compatibility (check number of lines for this object, if too few, use default object pose) was on option? Alas, I don't have cycles to contribute this myself.

Many thanks,
Wolfgang

wxmerkt:
This PR adds backwards compatibility for the old scene serialization format to `PlanningScene` following #2037, cf. the regression report for Noetic in #2985. It first checks which version of the syntax is provided (by checking the line following the start of a new object - previously the number of shapes, now the translational component of the pose) and deactivates the subframe-poses if using the old format, as well as maintains compatibility by resetting the pose to identity.

@felixvd @rhaschke Can you please review this PR, backport it to noetic-devel, and release? :-)

I've added unittests for the old format as well. Detail on changes/logic in the individual commits. 

Resolves #2985

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1000,14 +1000,57 @@ bool PlanningScene::loadGeometryFromStream(std::istream& in)
   return loadGeometryFromStream(in, Eigen::Isometry3d::Identity());  // Use no offset
 }
 
-bool PlanningScene::loadGeometryFromStream(std::istream& in, const Eigen::Isometry3d& offset)
+bool PlanningScene::loadGeometryFromStream(std::istream& in_istream, const Eigen::Isometry3d& offset)
 {
-  if (!in.good() || in.eof())
+  if (!in_istream.good() || in_istream.eof())
   {
     ROS_ERROR_NAMED(LOGNAME, "Bad input stream when loading scene geometry");
     return false;
   }
+
+  // We need to get the full string of the stream for version detection,
+  // so create a stringstream for processing. We pass the entire istream
+  // into this stringstream
+  std::stringstream in;
+  in << in_istream.rdbuf();
+
+  // Read scene name
   std::getline(in, name_);
+
+  // Identify scene format version for backwards compatibility of parser
+  const std::string in_str = in.str();
+  std::vector<std::string> in_lines;
+  boost::split(in_lines, in_str, boost::is_any_of("\n"));
+  bool uses_new_scene_format = true;
+  for (std::size_t i = 0; i < in_lines.size() - 1; ++i)
+  {
+    // An asterisk at the start of the line signifies a new object.
+    // An object can consist of multiple shapes.
+    if (in_lines[i][0] == '*')
+    {
+      // Detect the version of the serialization format based on the
+      // line succeeding the object identifier.
+
+      // Trim leading and trailing spaces (in-place)
+      boost::algorithm::trim(in_lines[i + 1]);
+
+      // Reliable way of detecting the format version is to check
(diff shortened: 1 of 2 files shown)


rhaschke (approved):
Thanks for this contribution. I slightly simplified the code. Otherwise looks good to me.
Well done providing unit tests!

wxmerkt:
Fantastic, thank you very much for the swift review and improvements @rhaschke :-)

simonschmeisser (approved):
I'd still highly welcome a PR for adding a sane format and I'm somewhat anxious about rebasing #2854 onto this ...

wxmerkt:
> I'd still highly welcome a PR for adding a sane format and I'm somewhat anxious about rebasing #2854 onto this ...

Thanks for pointing this out - taking a glance at #2854 it appears to change the serialization format as well - having the unit test should ensure that in the future backwards compatibility can be maintained more easily while still enabling new features to be added - the format has been released and people in the field are relying on it (if it requires changing a unit test to pass, it shouldn't be a patch release)

v4hn:
Looks good to me. Really sorry for breaking the format.
@simonschmeisser If we want to change the format again in the future it would definitely make sense to add a version number at the beginning of the file and patch older versions to skip the version tag. 
Or we just directly go for a different format that does not asciify floating values anymore. :grimacing: 
The current way to detect the file version works but is definitely overhead...

rhaschke:
I strongly support migrating the file format to yaml. This still means asciifying floats, though. Do you have a binary format in mind, @v4hn?

v4hn:
> migrating the file format to yaml

If you got a hammer, ... :smile: 

> Do you have a binary format in mind?

Not really, but do you foresee any problem with defining one ourselves?
- version tag at start of file
- define (little endian) byte order for binary numbers
- add number of elements in front of each list
- dump pretty much the same data we store now.

I don't see the need to provide backward compatibility for files generated with a more current version of MoveIt, and that's the main downside I see for this style in contrast to yaml.

If we can keep parsers for old versions around somehow, we might even use the binary serialization of the `moveit_msgs::PlanningScene` message, but that might be a rather stupid idea.

rhaschke:
I think, having a human-readable and -editable format is required. I wouldn't go for a binary format. Actually, I'm surprised that you _really_ had something like this in mind. We don't use binary formats anywhere else...
Actually, I was thinking about yaml-based message serialization.

v4hn:
The two of use agreed to disagree on this matter before. I simply prioritize accurate state representation of floats way more than the ability to edit files with a lot of internal structure in a text editor. There are of course merits to both approaches.

If you want to go for plain text approaches, an enhanced version of urdf might be an alternative to yaml, especially if we want to add tree structure to the planning scene at some point.

PR #2988, opened 2021-12-10
quietly use backward_cpp/ros if available

v4hn:
This is plain convenient and you always need it when you did not explicitly add it.

Follow @tylerjw's initiative to add it to MoveIt2: https://github.com/ros-planning/moveit2/pull/794

The downside (at least on my system) is that the cmake config for the package is not implemented correctly and yields this warning for each find_package call:
```
CMake Warning (dev) at /usr/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:426 (message):
  The package name passed to `find_package_handle_standard_args` (Backward)
  does not match the name of the calling package (backward_ros).  This can
  lead to problems in calling code that expects `find_package` result
  variables (e.g., `_FOUND`) to follow a certain pattern.
```

### DIFF
--- a/moveit_core/cmake/moveit.cmake
+++ b/moveit_core/cmake/moveit.cmake
@@ -5,6 +5,8 @@ macro(moveit_build_options)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   set(CMAKE_CXX_EXTENSIONS OFF)
 
+  find_package(backward_ros QUIET)
+
   if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
Looks like a useful package, indeed. It won't enter release builds though, thus no more information will be available for users to report in issues...

rhaschke (approved):
Looks like a useful package, indeed. It won't enter release builds though, thus no more information will be available for users to report in issues...

tylerjw:
I get these same cmake warnings in ROS2 locally some times: https://github.com/pal-robotics/backward_ros/issues/10

I have not debugged it further to figure out why or how to fix this.

PR #2990, opened 2021-12-11
Faster CI

tylerjw:
- Use lld for linker
- Increase ccache size to 10gb

### Description

This is a backport of a couple of changes I opened a PR to moveit2 with.  These should speed up the CI runs by a bit.  The ccache cache is compressed before uploading to github and this should only increase the debug one from ~700mb to ~1.2gb.  This increase resulted in a halving of the time it takes to build the debug ci run for moveit2.  Secondly, with moveit2 ld seems to hang sometimes when building the Pilz planner.  Switching to the lld linker resulted in another speed up in addition to fixing the hanging issue of ld.


simonschmeisser:
Thanks for the hint with lld, I really need to check this for our downstream code as well

What's the idea behind enabling -fPIC? This is automatically enabled by linking to Qt btw (not by linking itself of course but by the CMake target ... )

tylerjw:
fPIC was just me trying to fix the errors in the ci run.  I have no idea why these changes would cause ci to fail.

rhaschke (approved):
Looks like `lld` doesn't work yet on `Ubuntu Bionic` (the `master` docker image runs Bionic). Thus, I reverted the lld-related changes. Increasing the ccache size is a great improvement already! I only checked _normal_ build's cache sizes (~250 MB).
I couldn't imagine that a debug build would require 8GB!

PR #2991, opened 2021-12-12
Add object colors retrieval to planning scene interface

jgdo:
- Retrieve object colors map from PlanningSceneInterface. Previously this was not possible, although color information itself is contained in the scene message. 
- Also adjust python bindings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,9 @@ qtcreator-*
 # Vim
 *.swp
 
+# JetBrains IDEs (e.g. Pycharm or CLion)
+.idea/
+
 # Continous Integration
 .moveit_ci
 
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -42,6 +42,7 @@ from geometry_msgs.msg import Pose, Point
 from shape_msgs.msg import SolidPrimitive, Plane, Mesh, MeshTriangle
 from .exception import MoveItCommanderException
 from moveit_msgs.srv import ApplyPlanningScene, ApplyPlanningSceneRequest
+from std_msgs.msg import ColorRGBA
 
 try:
     from pyassimp import pyassimp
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -51,6 +51,9 @@ qtcreator-*
 # Vim
 *.swp
 
+# JetBrains IDEs (e.g. Pycharm or CLion)
+.idea/
+
```
rhaschke, line 56: Please consider a [_global_ `gitignore` file instead](https://github.com/ros-planning/moveit/pull/2976#pullrequestreview-816325571).

v4hn, line 56: @rhaschke argued against more entries here before, but I'm ok with adding jetbrains to the list when vscode and qtcreator are there already.

rhaschke (commented):
Generally looks good. Do you mind directly targeting the new pybind11-based API?
We have an open PR for this: https://github.com/ros-planning/moveit/pull/2910.
Merging here would create new conflicts...

v4hn (approved):
looks good to me.

.gitignore
rhaschke: I'm pointing people to a better approach. But I don't mind adding JetBrains here as well.

rhaschke:
You are right, I referenced the wrong PR.
There is an [open PR](https://github.com/ubi-agni/moveit/pull/4) to augment #2910.
Probably, I will only find some time over the Christmas break to look into these in more detail. Please file your PR on top of https://github.com/ubi-agni/moveit/pull/4. Thanks.

jgdo:
I have [rewritten the change to use pybind11](https://github.com/jgdo/moveit/commit/70d438b4d320ec10f94229817c2376a1e17dac93), but I'll wait until the [`PR you mentioned`](https://github.com/ubi-agni/moveit/pull/4) is done, otherwise the PR chain will be too big IMHO.

rhaschke:
@jgdo, I merged https://github.com/ubi-agni/moveit/pull/4. Could you please file a PR of your pybind11-based changes against #2910? Thanks. Closing here.

PR #2993, opened 2021-12-13
pin previous tinyxml2 for static lib

wolfv:
@rhaschke let's see if this fixes the builds for now. 

We could also create a real "lock file" using a tool like [`conda-lock`](https://github.com/conda-incubator/conda-lock).


rhaschke:
Thanks a lot, @wolfv!

PR #2995, opened 2021-12-13
Restore compatibility for `RobotState::attachBody` API

marip8:
### Description

The release of 1.1.6 breaks the `moveit::core::RobotState::attachBody` API between ROS Noetic and Melodic/Kinetic. I'm supporting builds of a project in Noetic and Melodic that use this function, and currently the Melodic build is failing due to this API change. This PR restores compatibility by adding deprecated function signatures for `attachBody` that match those in ROS Melodic/Kinetic.

Based on discussion in #2985 and #2986, it sounds like an updated version of MoveIt might be released soon to fix that breakage. If so, it would be great to wrap this change into that release as well.

### Checklist

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -1699,6 +1699,27 @@ public:
     attachBody(id, pose, shapes, shape_poses, touch_links_set, link_name, detach_posture, subframe_poses);
   }
 
+  [[deprecated("No offset pose for the attached shapes can be specified with this overload; using a default offset of "
+               "identity")]] void
+  attachBody(const std::string& id, const std::vector<shapes::ShapeConstPtr>& shapes,
+             const EigenSTL::vector_Isometry3d& shape_poses, const std::set<std::string>& touch_links,
+             const std::string& link_name,
+             const trajectory_msgs::JointTrajectory& detach_posture = trajectory_msgs::JointTrajectory())
+  {
+    attachBody(id, Eigen::Isometry3d::Identity(), shapes, shape_poses, touch_links, link_name, detach_posture);
+  }
+
+  [[deprecated("No offset pose for the attached shapes can be specified with this overload; using a default offset of "
+               "identity")]] void
+  attachBody(const std::string& id, const std::vector<shapes::ShapeConstPtr>& shapes,
+             const EigenSTL::vector_Isometry3d& shape_poses, const std::vector<std::string>& touch_links,
+             const std::string& link_name,
+             const trajectory_msgs::JointTrajectory& detach_posture = trajectory_msgs::JointTrajectory())
+  {
+    std::set<std::string> touch_links_set(touch_links.begin(), touch_links.end());
+    attachBody(id, Eigen::Isometry3d::Identity(), shapes, shape_poses, touch_links_set, link_name, detach_posture);
+  }
+
   /** \brief Get all bodies attached to the model corresponding to this state */
   void getAttachedBodies(std::vector<const AttachedBody*>& attached_bodies) const;
 


v4hn:
Thanks for creating this PR.

Is it too frustrating for you to use `#ifdef` [version checks](https://github.com/ros-planning/moveit/blob/master/moveit_core/version/version.h.in#L38-L43) for the calls?
Other maintainers might disagree, but I would prefer a proper mention in the [`MIGRATION.md`](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) over adding a deprecated method to `master`.
Specifically the handling of object poses changed as well, so you do not just specify an optional "offset pose". If you specify poses w.r.t. the objects, MoveIt will silently use identity instead of the expected pose with your method.

marip8:
> Is it too frustrating for you to use #ifdef version checks for the calls?

Not necessarily. I was previously unaware that MoveIt had such a macro. This is a good short term solution for me.

> Specifically the handling of object poses changed as well, so you do not just specify an optional "offset pose". If you specify poses w.r.t. the objects, MoveIt will silently use identity instead of the expected pose with your method.

I'm not sure I understand. Per [this line](https://github.com/ros-planning/moveit/blob/a4e023b50c0813fa54be271a97d928e0f1433ab4/moveit_core/robot_state/src/attached_body.cpp#L81) the `pose` argument just applies a fixed transform to all of the provided shape poses to move them relative to the attachment link. Since this argument did not exist prior to 1.1.6, it seems like providing it as an identity matrix in the deprecated function I created would result in the same behavior as before

> Other maintainers might disagree, but I would prefer a proper mention in the MIGRATION.md over adding a deprecated method to master.

I understand not wanting to bloat the code base with deprecated functions, but there is something to be said for easy compatibility between two very similar distributions. It's a bit of a pain to have to fix my application code and CI/CD pipelines, etc. because a stable library broke API after being released for a distro, especially when backwards compatibility is easy to provide, as in this case.

jrgnicho (approved):

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1699,6 +1699,27 @@ class RobotState
     attachBody(id, pose, shapes, shape_poses, touch_links_set, link_name, detach_posture, subframe_poses);
   }
 
+  [[deprecated("No offset pose for the attached shapes can be specified with this overload; using a default offset of "
+               "identity")]] void
```
rhaschke, line 1703:
```suggestion
  [[deprecated("Use the new method that allows passing a base pose for the attached body. Using identity.")]] void
```

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1699,6 +1699,27 @@ class RobotState
...
+  attachBody(const std::string& id, const std::vector<shapes::ShapeConstPtr>& shapes,
+             const EigenSTL::vector_Isometry3d& shape_poses, const std::set<std::string>& touch_links,
+             const std::string& link_name,
+             const trajectory_msgs::JointTrajectory& detach_posture = trajectory_msgs::JointTrajectory())
+  {
+    attachBody(id, Eigen::Isometry3d::Identity(), shapes, shape_poses, touch_links, link_name, detach_posture);
+  }
+
+  [[deprecated("No offset pose for the attached shapes can be specified with this overload; using a default offset of "
+               "identity")]] void
```
rhaschke, line 1713:
```suggestion
  [[deprecated("Use the new method that allows passing a base pose for the attached body. Using identity.")]] void
```

rhaschke (commented):
I'm generally fine to provide this simple fallback here. Please improve the deprecation warning, e.g. using the suggestion.

rhaschke:
> Specifically the handling of object poses changed as well, so you do not just specify an optional "offset pose". If you specify poses w.r.t. the objects, MoveIt will silently use identity instead of the expected pose with your method.

@marip8, the key point here is that MoveIt also changed what is _returned_ as the pose of the attached body: Before #2037, MoveIt returned the pose of the _first_ shape, but now it returns the base pose, i.e. identity in your case. Hence, old code still expecting the first' shapes pose will be silently broken.
For this reason, keeping the breaking change might be more beneficial, because it forces the developer to think about her code.
See #2997 for a simple version check macro.

marip8:
> the key point here is that MoveIt also changed what is returned as the pose of the attached body: Before #2037, MoveIt returned the pose of the first shape, but now it returns the base pose, i.e. identity in your case. Hence, old code still expecting the first' shapes pose will be silently broken.

I think I follow. The `attachBody` method does not return anything, but some other function (`getFrameTransform` I think) accesses the `AttachedBody` object created by `attachBody` and extracts the value of `pose`, which in the case of the deprecated function would be identity rather than the pose of the first shape. Is that right?

rhaschke:
Correct.

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
rhaschke: Given this incompatibility, I agree with @v4hn and am inclined to reject this PR.

marip8: That's fine with me. The version macro solution works well enough for my application

jrgnicho:
Looks like there's consensus on rejecting this PR due to the reasons mentioned above, mainly the difference in the returned poses by the `getFrameTransform` function.  So @rhaschke  should we close this PR then?

rhaschke:
As @marip8 agreed to use a preprocessor approach, I think, yes, we can close this PR.

Issue #2996, opened 2021-12-14
Provide version check macro


simonschmeisser:
Qt has a very convenient macro for version checks:
https://doc.qt.io/qt-5/qtglobal.html#QT_VERSION_CHECK

We should have that too

relates to #2995

v4hn:
I wrote [a related macro for MTC before](https://github.com/ros-planning/moveit_task_constructor/blob/master/core/include/moveit/task_constructor/moveit_compat.h#L43-L45).

PR #2997, opened 2021-12-14
Provide MOVEIT_VERSION_CHECK macro

rhaschke:
- Rename `MOVEIT_VERSION` -> `MOVEIT_VERSION_STR`
- `MOVEIT_VERSION` becomes a _numeric_ identifier
- Use like: `#if MOVEIT_VERSION >= MOVEIT_VERSION_CHECK(1, 0, 0)`

This should be backported to Melodic as well.
Implements #2996.


rhaschke:
> The only thing I worry about with this PR is the change in content of `MOVEIT_VERSION` but I doubt there are many external users of the string currently.

I agree. As we will backport this PR asap to Melodic as well, I don't fear any problems though. The old version string wasn't usable for version checks and thus - as you said - is probably not really used externally.

v4hn:
> > Rename MOVEIT_VERSION -> MOVEIT_VERSION_STR

> The old version string wasn't usable for version checks and thus - as you said - is probably not really used externally.

At least provide a `MIGRATIONS.md` entry if you insist to break user builds for no reason in the next melodic release.
I would have strictly preferred an approach that does not require the rename, e.g., the MTC `*_GE` approach, a different name for the variable (`MOVEIT_INT_VERSION`/`MOVEIT_VERSION_INT`), or something else.

PR #3000, opened 2021-12-15
pilz planner: report invalid joint

v4hn:
Improve usability. I kept getting out of range errors in a start state and needed to implement this code to figure out which joints triggered the problem...

### DIFF
--- a/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/joint_limits_container.h
+++ b/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/joint_limits_container.h
@@ -140,15 +140,6 @@ public:
    */
   bool verifyPositionLimit(const std::string& joint_name, const double& joint_position) const;
 
-  /**
-   * @brief verify position limits of multiple joints
-   * @param joint_names
-   * @param joint_positions
-   * @return
-   */
-  bool verifyPositionLimits(const std::vector<std::string>& joint_names,
-                            const std::vector<double>& joint_positions) const;
-
 private:
   /**
    * @brief update the most strict limit with given joint limit
--- a/moveit_planners/moveit_planners/package.xml
+++ b/moveit_planners/moveit_planners/package.xml
@@ -19,6 +19,7 @@
   <exec_depend>chomp_motion_planner</exec_depend>
   <exec_depend>moveit_planners_chomp</exec_depend>
   <exec_depend>moveit_planners_ompl</exec_depend>
+  <exec_depend>pilz_industrial_motion_planner</exec_depend>
 
   <export>
     <metapackage/>
--- a/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.h
+++ b/moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/trajectory_generator.h
@@ -223,7 +223,7 @@ private:
    *     - The start state velocity is below
    * TrajectoryGenerator::VELOCITY_TOLERANCE
    */
-  void checkStartState(const moveit_msgs::RobotState& start_state) const;
+  void checkStartState(const moveit_msgs::RobotState& start_state, const std::string& group) const;
 
   void checkGoalConstraints(const moveit_msgs::MotionPlanRequest::_goal_constraints_type& goal_constraints,
                             const std::vector<std::string>& expected_joint_names, const std::string& group_name) const;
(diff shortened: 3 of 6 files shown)


moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/joint_limits_container.h
```diff
@@ -144,10 +144,11 @@ class JointLimitsContainer
    * @brief verify position limits of multiple joints
    * @param joint_names
    * @param joint_positions
+   * @param first_failing_joint returns first joint that fails position limits (if any)
    * @return
    */
-  bool verifyPositionLimits(const std::vector<std::string>& joint_names,
-                            const std::vector<double>& joint_positions) const;
+  bool verifyPositionLimits(const std::vector<std::string>& joint_names, const std::vector<double>& joint_positions,
+                            std::string* first_failing_joint = nullptr) const;
```
rhaschke, line 151: I don't understand why you restrict the API to report only the first offending joint. If there are multiple joints, this will require a tedious iteration to fix all of them.
I suggest passing a reference to a `vector<string>` instead.  If your main motivation was API compatibility, you could directly fill the (optional) string with a comma-separated list of joint names.

rhaschke (commented):
Doesn't the Pilz planner use the FixStartState planning adapter?
Issues like these shouldn't be handled in each and every planner, right?
Also, there are methods available apart from the proprietary `JointLimitsContainer::verifyPositionLimits` to validate the (initial) robot state, aren't they?
Long story short: I don't like this kind of code duplication.

moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/joint_limits_container.h
v4hn: my main motivation was to avoid iterating over everything when the original code did not. but I can also change that. :shrug:

rhaschke: In normal operation, i.e. without joint violations, the code needs to iterate all joints anyway and the failure case isn't time-critical, is it?

v4hn:
> Doesn't the Pilz planner use the FixStartState planning adapter?

[No.](https://github.com/ros-planning/moveit/blob/master/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/pilz_industrial_motion_planner_planning_pipeline.launch.xml#L4) It would be problematic to add any of the `Fix*` adapters to the pipeline because it generates its own time parameterization and the Fix* plugins rely on the final time parameterization after running all other adapters.

> Issues like these shouldn't be handled in each and every planner, right

no, not in path planners (almost all). but Pilz generates trajectories, and our fix adapters do not support this.

> Also, there are methods available apart from the proprietary JointLimitsContainer::verifyPositionLimits to validate the (initial) robot state, aren't they?

feel free to simplify the whole planner codebase to use our RobotState class. I agree this addresses only the (for me) urgent issue at hand, but it does do that.

Mainly this became an issue for me because we don't have soft/hard limit support, but that is obviously more of a GSoC project

rhaschke:
> It would be problematic to add any of the `Fix*` adapters to the pipeline because it generates its own time parameterization and the Fix* plugins rely on the final time parameterization after running all other adapters.

I don't agree. There are adapters that are applied _before_ the actual planner (most/all `Fix*` plugins) and there are ones that are applied afterwards (e.g. time parameterization). I agree, that the time parameterization adapter shouldn't be applied for the Pilz Motion Generator. But the pre-planning adapters should be applicable. I don't see a reason why not.

v4hn:
> the pre-planning adapters should be applicable. I don't see a reason why not.

The `Fix*` adapters [re-add the original initial state at the beginning of the generated trajectory](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_bounds.cpp#L190)!
Looking up the implementation again, I found there is at least a configurable (and a heuristic value) for the new duration_from_previous entry.

But nonetheless the final trajectory is not well-parameterized after adding a (second) zero-velocity, zero-acceleration waypoint in the beginning and all pipelines I worked with in the past that used this feature run the intermediate result through the (final) time parameterization adapter that should rectify this.
Especially when you select a "planner" that guarantees a trapezoidal velocity profile, a sudden "jump" motion with very low execution time but static end-point is most likely unwanted.

rhaschke (approved):

v4hn:
Ladies and Gentleman (and everyone else).

Pull-Request number 3000 got merged in this repository :champagne:

PR #3002, opened 2021-12-15
Don't try to print dirty transforms

jspricke:
printTransform() calls ASSERT_ISOMETRY() which most likely fails for a
dirty transform. Give that the initial state after creating a Robotstate
is dirty this allows to print the state after creation.

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -2194,7 +2194,6 @@ void RobotState::printStateInfo(std::ostream& out) const
 
 void RobotState::printTransform(const Eigen::Isometry3d& transform, std::ostream& out) const
 {
-  ASSERT_ISOMETRY(transform)  // unsanitized input, could contain a non-isometry
   Eigen::Quaterniond q(transform.linear());
   out << "T.xyz = [" << transform.translation().x() << ", " << transform.translation().y() << ", "
       << transform.translation().z() << "], Q.xyzw = [" << q.x() << ", " << q.y() << ", " << q.z() << ", " << q.w()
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
I understand your reasoning. But I think, your use case of printing just initialized RobotStates is too narrow. The assertion might also fail for "clean" transforms due to bad initialization or numerical issues. In this case, we are particularly interested to print these transforms.
Thus, I suggest dropping the assertion from `printTransform()` instead.

v4hn:
@rhaschke Jochen pretty much implemented what I discussed with him in a video call. @jspricke please add a mention to me when you create patches we discussed.

> The assertion might also fail for "clean" transforms due to bad initialization or numerical issues. In this case, we are particularly interested to print these transforms.

Of course that can happen. I disagree with your assessment though. I would definitely **not** want to print values for transforms that are "almost" isometries, because the printed numbers would usually still have a sufficiently close norm. Instead I would want to know contract for `Isometry3d` to be a valid isometry was violated.

I see three options:

- enforce (clean) Eigen::Isometry3d objects to be valid isometries throughout MoveIt in Debug mode.
  - this is the current approach and what Jochen (and I) proposed to keep here
  - it will only assert if you compiled without `-DNDEBUG`
  - with `-DNDEBUG` all transforms will be printed without the check
- weaken the assumption for printing to see concrete values (what you propose)
  - Will print *all* transforms (possibly many invalid ones) instead of asserting at the first one.
  - your current patch will not check for isometries at all anymore. If you want to force this approach, there should be an additional call to `checkIsometry(tf)` with a printed message indicating check results.
- do *both*, assert without `-DNDEBUG` and check (and print negative results) with `-DNDEBUG`

rhaschke:
`printTransforms()` is used for debugging usually/only, right? As said initially, I would be particularly interested in printing a broken transform in that case. If you want to `assert` _additionally_ (to bail out asap), that's fine for me. This would result in your option 3.

jspricke:
fine with me. Closing as this is the current behavior.

PR #3005, opened 2021-12-20
Do not assert on printTransform with non-isometry

v4hn:
Fix for #3002 because the author there decided to ignore the problem.

Instead of the assert print a tag and the matrix.
Building a Quaternion from non-isometries is undefined behavior in Eigen, thus the split.

The new output looks like this when isometries are broken:
```
Joint transforms:
  virtual_joint [dirty]: [NON-ISOMETRY] [6.92595e-310, 0, 0, 0; 6.92595e-310, 0, 7.41098e-323, 4.94066e-324; 8.98317e-317, 0, 0, 8.98318e-317; 0, 0, 0, 1]
  panda_joint1 [dirty]: [NON-ISOMETRY] [0, 0, 0, -nan; 0, 0, 0, 0; 0, 0, 0, 0; 0, 0, 0, 1]
  panda_joint2 [dirty]: [NON-ISOMETRY] [0, 1.63318e-319, 0, nan; 9.13379e-320, 0, 0, 0; 1.63318e-319, 0, 0, 0; 0, 0, 0, 1]
```
and usually like that (unchanged):
```
Joint transforms:
  virtual_joint: T.xyz = [0, 0, 0], Q.xyzw = [0, 0, 0, 1]
  panda_joint1: T.xyz = [0, 0, 0], Q.xyzw = [0, 0, 0, 1]
  panda_joint2: T.xyz = [0, 0, 0], Q.xyzw = [0, 0, 0, 1]
```

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -2194,11 +2194,20 @@ void RobotState::printStateInfo(std::ostream& out) const
 
 void RobotState::printTransform(const Eigen::Isometry3d& transform, std::ostream& out) const
 {
-  ASSERT_ISOMETRY(transform)  // unsanitized input, could contain a non-isometry
-  Eigen::Quaterniond q(transform.linear());
-  out << "T.xyz = [" << transform.translation().x() << ", " << transform.translation().y() << ", "
-      << transform.translation().z() << "], Q.xyzw = [" << q.x() << ", " << q.y() << ", " << q.z() << ", " << q.w()
-      << "]" << std::endl;
+  if (checkIsometry(transform, CHECK_ISOMETRY_PRECISION, false))
+  {
+    Eigen::Quaterniond q(transform.linear());
+    out << "T.xyz = [" << transform.translation().x() << ", " << transform.translation().y() << ", "
+        << transform.translation().z() << "], Q.xyzw = [" << q.x() << ", " << q.y() << ", " << q.z() << ", " << q.w()
+        << "]";
+  }
+  else
+  {
+    out << "[NON-ISOMETRY] "
+        << transform.matrix().format(
+               Eigen::IOFormat(Eigen::StreamPrecision, Eigen::DontAlignCols, ", ", "; ", "", "", "[", "]"));
+  }
+  out << std::endl;
 }
 
 void RobotState::printTransforms(std::ostream& out) const


v4hn:
@jspricke @rhaschke

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -2194,11 +2194,20 @@ void RobotState::printStateInfo(std::ostream& out) const
...
-  Eigen::Quaterniond q(transform.linear());
-  out << "T.xyz = [" << transform.translation().x() << ", " << transform.translation().y() << ", "
-      << transform.translation().z() << "], Q.xyzw = [" << q.x() << ", " << q.y() << ", " << q.z() << ", " << q.w()
-      << "]" << std::endl;
+  if (checkIsometry(transform, CHECK_ISOMETRY_PRECISION, false))
+  {
+    Eigen::Quaterniond q(transform.linear());
+    out << "T.xyz = [" << transform.translation().x() << ", " << transform.translation().y() << ", "
+        << transform.translation().z() << "], Q.xyzw = [" << q.x() << ", " << q.y() << ", " << q.z() << ", " << q.w()
+        << "]";
```
rhaschke, line 2202: Should we take the chance and simplify the code even more?
```suggestion
    out << "T.xyz = [" << transform.translation().transpose() << "], Q.xyzw = [" << q.vec().transpose() << ", " << q.w() << "]";
```

rhaschke (approved):
In https://github.com/ros-planning/moveit/pull/3002#issuecomment-995593734 you insisted on early assertion. Thus, it comes at a surprise for me that you now dropped the assertion completely.

v4hn:
> In #3002 (comment) you insisted on early assertion. Thus, it comes at a surprise for me that you now dropped the assertion completely.

Mostly I insist on not accessing the data at all when it's marked as dirty. You disagreed.
As @jspricke pointed out in private conversation these are not compatible if we want to avoid different behavior (the assert) for the very valid case where not all transforms are initialized yet.
He decided to ignore the bug because of these unnecessary discussions
I don't care enough to bikeshed this ad absurdum, as long as the actual bugs - (1) that you don't get any indication in Release build on whether the isometries are valid and (2) that a simple print of a valid use case behaves different between build types - are fixed.

jspricke:
* Michael Görner ***@***.***> [2021-12-21 02:02]:
>As @jspricke pointed out in private conversation these are not compatible if we want to avoid different behavior (the assert) for the very valid case where not all transforms are initialized yet.
>He decided to ignore the bug because of these unnecessary discussions

No, I decided that both behavior are fine and it is a maintainer 
decision how the class should behave.

PR #3011, opened 2021-12-24
Removed raw pointer creation using new from robot state test

pvanlaar:
### Description

The method RobotState::attachBody(AttachBody*) expects an owning raw pointer. This can be seen from the comment above the method (see below).  This PR replaces that owning raw pointer with a `std::unique_ptr`. The old method is marked deprecated. This allows removing multiple `delete`'s in robot_state.cpp. Additionally two other `new` keywords were replaced in the __test__ code of robot_state.

~~~
  /** \brief Add an attached body to this state. Ownership of the
   * memory for the attached body is assumed by the state.
   *
   * This only adds the given body to this RobotState
   * instance.  It does not change anything about other
   * representations of the object elsewhere in the system.  So if the
   * body represents an object in a collision_detection::World (like
   * from a planning_scene::PlanningScene), you will likely need to remove the
   * corresponding object from that world to avoid having collisions
   * detected against it.
   *
   * \note This version of the function (taking an AttachedBody
   * pointer) does not copy the AttachedBody object, it just uses it
   * directly.  The AttachedBody object stores its position data
   * internally.  This means you should <b>never attach a single
   * AttachedBody instance to multiple RobotState instances</b>, or
   * the body positions will get corrupted.  You need to make a fresh
   * copy of the AttachedBody object for each RobotState you attach it
   * to.*/
~~~

### Checklist
- ~~[ ] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)~~
- ~~[] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html)~~ refactoring, no behaviour changes
- ~~[ ] Include a screenshot if changing a GUI~~
- ~~[ ] While waiting for someone to review your request, please help review [another open pull request](https://github.com/ros-planning/moveit/pulls) to support the maintainers~~

## Future steps
I would like to separate the AttachedBody management into a separate class. Then RobotState could call that class. This new class would be focussed around `the attached_body_map_` field, which should be moved to the new class. What do you think?

### DIFF
--- a/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/robot_state.h
@@ -1068,12 +1068,12 @@ public:
       @param timeout The timeout passed to the kinematics solver on each attempt
       @param constraint A state validity constraint to be required for IK solutions */
   bool setFromIK(const JointModelGroup* group, const EigenSTL::vector_Isometry3d& poses,
-                 const std::vector<std::string>& tips, const std::vector<std::vector<double> >& consistency_limits,
+                 const std::vector<std::string>& tips, const std::vector<std::vector<double>>& consistency_limits,
                  double timeout = 0.0, const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
                  const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions());
   [[deprecated("The attempts argument is not supported anymore.")]] bool
   setFromIK(const JointModelGroup* group, const EigenSTL::vector_Isometry3d& poses,
-            const std::vector<std::string>& tips, const std::vector<std::vector<double> >& consistency_limits,
+            const std::vector<std::string>& tips, const std::vector<std::vector<double>>& consistency_limits,
             unsigned int /* attempts */, double timeout = 0.0,
             const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
             const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions())
@@ -1091,12 +1091,12 @@ public:
       @param constraint A state validity constraint to be required for IK solutions */
   bool setFromIKSubgroups(const JointModelGroup* group, const EigenSTL::vector_Isometry3d& poses,
                           const std::vector<std::string>& tips,
-                          const std::vector<std::vector<double> >& consistency_limits, double timeout = 0.0,
+                          const std::vector<std::vector<double>>& consistency_limits, double timeout = 0.0,
                           const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
                           const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions());
   [[deprecated("The attempts argument is not supported anymore.")]] bool
   setFromIKSubgroups(const JointModelGroup* group, const EigenSTL::vector_Isometry3d& poses,
-                     const std::vector<std::string>& tips, const std::vector<std::vector<double> >& consistency_limits,
+                     const std::vector<std::string>& tips, const std::vector<std::vector<double>>& consistency_limits,
                      unsigned int /* attempts */, double timeout = 0.0,
                      const GroupStateValidityCallbackFn& constraint = GroupStateValidityCallbackFn(),
                      const kinematics::KinematicsQueryOptions& options = kinematics::KinematicsQueryOptions())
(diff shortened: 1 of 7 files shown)


pvanlaar:
@v4hn As we worked before on refactorings, would you mind reviewing this?

v4hn (approved):
Looks reasonable to me.
You deprecate the old method, but could you also add an entry to the `MIGRATION.md` file?

You explicitly avoid changing any of the get APIs that return raw pointers, that makes the change a lot less debatable. :+1:

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1647,7 +1659,9 @@ class RobotState
    * the body positions will get corrupted.  You need to make a fresh
    * copy of the AttachedBody object for each RobotState you attach it
    * to.*/
-  void attachBody(AttachedBody* attached_body);
+  [[deprecated("Deprecated use the overload requiring an unique_ptr instead")]] void
+  attachBody(AttachedBody* attached_body);
+  // TODO Remove this function
```
rhaschke, line 1664: Deprecated functions are scheduled for removal automatically.
```suggestion
```

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -978,25 +978,30 @@ bool RobotState::hasAttachedBody(const std::string& id) const
 
 const AttachedBody* RobotState::getAttachedBody(const std::string& id) const
 {
-  std::map<std::string, AttachedBody*>::const_iterator it = attached_body_map_.find(id);
+  const auto it = attached_body_map_.find(id);
```
rhaschke, line 981: Why not using a reference here?
```suggestion
  const auto& it = attached_body_map_.find(id);
```

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
```diff
@@ -1647,7 +1659,9 @@ class RobotState
    * the body positions will get corrupted.  You need to make a fresh
    * copy of the AttachedBody object for each RobotState you attach it
    * to.*/
-  void attachBody(AttachedBody* attached_body);
+  [[deprecated("Deprecated use the overload requiring an unique_ptr instead")]] void
```
simonschmeisser, line 1662:
```suggestion
  [[deprecated("Deprecated use the overload requiring a unique_ptr instead")]] void
```
https://preply.com/en/question/which-is-correct-a-unique-or-an-unique-49748  English is such a lovely language :wink:

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
simonschmeisser, line 1664:
```suggestion
```
either remove this or add some reasoning when to remove this (ie some version of MoveIt 2)

rhaschke (commented):
I like the idea of migrating to `std::unique_ptr` internally. However, I'm not convinced to change the public API of a single function to `unique_ptr`. All remaining functions are still using raw pointers!

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
pvanlaar: Removed, as @rhaschke mentioned that removal of deprecated functions is automatic.

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
pvanlaar: Nice catch! Changed it to @rhaschke's suggestion.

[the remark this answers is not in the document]
pvanlaar: Changed to your suggestion.

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
pvanlaar: Removed. Is the automatic removal done by a CI tool?

moveit_core/robot_state/src/robot_state.cpp
pvanlaar: Indeed, that is better. Changed it.

pvanlaar:
@v4hn Added an entry to the Migration.md file as you suggested.

pvanlaar:
@rhaschke 

> I like the idea of migrating to `std::unique_ptr` internally. However, I'm not convinced to change the public API of a single function to `unique_ptr`. All remaining functions are still using raw pointers!

What do you mean with remaining functions? The methods in RobotState operating on the AttachedBody collection? Other methods in RobotState or in other places in the code base? 

The other methods in RobotState are used to get AttachedBodies, which are then still owned by the RobotState. These methods return non-owning raw pointers, which is fine. I am sure you know this already, thus probably I am understanding you wrong. If you mean functions or methods in other places, we can create other issues to correct those. Or do you want to include those chnages in this issue/pull request?

moveit_core/robot_state/include/moveit/robot_state/robot_state.h
rhaschke: No. Automatic here just meant that deprecated functions are automatically considered by maintainers for removal in some future version. There is no need to explicitly mark it with an extra TODO.

moveit_core/robot_state/src/robot_state.cpp
rhaschke: Sorry, I wasn't reading correctly here. Iterators, of course, shouldn't use references. They are pointer-like data structures.

rhaschke:
> All remaining functions are still using raw pointers.

I was referring to the getter functions. Having thought about it again, I agree that indicating transfer of ownership via a `unique_ptr` is fine while keeping raw pointers for getters.

Issue #2107, opened 2020-05-25
Some tests take too long in debug mode


peci1:
As observed in #1964 by me and (reportedly) elsewhere by @v4hn, some tests take very long when run in debug mode.

I'm creating this issue to track the long tests. Any effort to make the tests debug-friendly will be welcome.

I've built moveit in debug mode and ran all the tests using `make run_tests` in the build directory.

This is the list of all tests that took more than 10 seconds on my computer. The list was built running:

    find test_results -name '*.xml' -exec grep -H "testsuite " {} \; | grep moveit | sed 's/^\([^:]*\):.*name="\([^"]*\)".*time="\([^"]*\)".*$/\1,\2,\3/' | sort -t, -k3 -gr | grep '[0-9]\{2,\}\.[0-9]*$'

Results file | Test Suite | Test time (sec)
-- | -- | --
test_results/moveit_ros_robot_interaction/gtest-locked_robot_state_test.xml | LockedRobotState | 3813,58
test_results/moveit_core/gtest-test_distance_field.xml | TestSignedPropagationDistanceField | 423,5
test_results/moveit_core/gtest-test_collision_distance_field.xml | DistanceFieldCollisionDetectionTester | 313,944
test_results/moveit_ros_planning_interface/rostest-test_robot_state_update.xml | testrobot_state_update | 153,6266
test_results/moveit_ros_planning_interface/rosunit-robot_state_update.xml | test | 149,607
test_results/moveit_planners_chomp/rostest-test_chomp_moveit.xml | testchomp_test | 95,4951
test_results/moveit_planners_chomp/rosunit-chomp_test.xml | CHOMPMoveitTest | 80,478
test_results/moveit_core/gtest-test_time_optimal_trajectory_generation.xml | time_optimal_trajectory_generation | 47,418
test_results/moveit_ros_planning_interface/rostest-test_move_group_interface_cpp_test.xml | testmove_group_interface_cpp_test | 41,7406
test_results/moveit_ros_planning_interface/rosunit-move_group_interface_cpp_test.xml | MoveGroupTestFixture | 40,758
test_results/moveit_ros_planning_interface/rostest-test_cleanup.xml | testcleanup_tests | 19,9096
test_results/moveit_ros_planning_interface/rostest-test_subframes_test.xml | testsubframes_test | 11,363
test_results/moveit_commander/rostest-test_python_moveit_commander_ns.xml | testpython_moveit_commander_ns | 10,0509

peci1:
[...]

PR #3012, opened 2021-12-28
Disable timing tests in DEBUG mode

rhaschke:
Debug builds are significantly slower and thus shouldn't be considered for timing-based tests.
They frequently fail, e.g. https://github.com/ros-planning/moveit/runs/4648164688?check_suite_focus=true

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_pr2.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_pr2.h
@@ -87,10 +87,10 @@ protected:
   std::string kinect_dae_resource_;
 };
 
-#ifdef NDEBUG  // Don't perform timing tests in Debug mode (but evaluate expression)
-#define EXPECT_TIME_LT(EXPR, VAL) (void)(EXPR)
-#else
+#ifdef NDEBUG
 #define EXPECT_TIME_LT(EXPR, VAL) EXPECT_LT(EXPR, VAL)
+#else  // Don't perform timing checks in Debug mode (but evaluate expression)
+#define EXPECT_TIME_LT(EXPR, VAL) (void)(EXPR)
 #endif
 
 TYPED_TEST_CASE_P(CollisionDetectorTest);
--- a/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_pr2.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_pr2.h
@@ -87,6 +87,12 @@ protected:
   std::string kinect_dae_resource_;
 };
 
+#ifdef NDEBUG  // Don't perform timing tests in Debug mode (but evaluate expression)
+#define EXPECT_TIME_LT(EXPR, VAL) (void)(EXPR)
+#else
+#define EXPECT_TIME_LT(EXPR, VAL) EXPECT_LT(EXPR, VAL)
+#endif
+
 TYPED_TEST_CASE_P(CollisionDetectorTest);
 
 TYPED_TEST_P(CollisionDetectorTest, InitOK)
@@ -361,7 +367,7 @@ TYPED_TEST_P(CollisionDetectorTest, DiffSceneTester)
   new_cenv->checkSelfCollision(req, res, robot_state);
   double second_check = (ros::WallTime::now() - before).toSec();
 
-  EXPECT_LT(fabs(first_check - second_check), .05);
+  EXPECT_TIME_LT(fabs(first_check - second_check), .05);
 
   std::vector<shapes::ShapeConstPtr> shapes;
   shapes.resize(1);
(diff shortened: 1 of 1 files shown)


JafarAbdi (approved):

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_pr2.h
```diff
@@ -87,6 +87,12 @@ class CollisionDetectorTest : public ::testing::Test
   std::string kinect_dae_resource_;
 };
 
+#ifdef NDEBUG  // Don't perform timing tests in Debug mode (but evaluate expression)
```
v4hn, line 90:
```suggestion
#ifndef NDEBUG  // Don't perform timing tests in Debug mode (but evaluate expression)
```
Or do I misunderstand this expression?

v4hn (commented):
That's quite pragmatic. I mostly agree that we should not test timing of debug builds.
I'm not sure plain disabling them (silently!) is a good idea though.

From my side I wanted to migrate all such tests into https://github.com/captain-yoshi/moveit_benchmark_suite because timing tests should simply not be hard tests at all.
But I can't afford to spend work on improving that project at this time. @captain-yoshi works on it on the side and keeps making various incremental improvements. (Thanks you David!)

Lastly, please consider https://github.com/ros-planning/moveit/issues/2107 . The main offender for runtime in debug mode is https://github.com/ros-planning/moveit/blob/master/moveit_ros/robot_interaction/test/locked_robot_state_test.cpp , which should probably also just be disabled due to overly long runtime if we go with your proposed approach.

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_pr2.h
rhaschke: You are right. In my [original commit](https://github.com/ros-planning/moveit/commit/3ebeb650e4cdf0988cde8ae2fd666a2719ea8183) it was the correct expression. Here, it is inverted.

rhaschke:
@JafarAbdi, thanks for pointing out #2107. However, here I'm fighting flaky tests due to timing, while #2107 complains about slow tests in general.

moveit_core/collision_detection/include/moveit/collision_detection/test_collision_common_pr2.h
rhaschke: You are right. In my [original commit](https://github.com/ros-planning/moveit/commit/3ebeb650e4cdf0988cde8ae2fd666a2719ea8183), it was the correct expression. Here, it is inverted.

v4hn:
**Somewhat off-topic for this thread, but anyway**

> Factoring out the test into a benchmark suite isn't appropriate IMO. It actually tests that collision checking doesn't crash and

I fully agree that all functional tests belong to the core tests and for improved efficiency all tests should also check more functional aspects than just "doesn't crash" (at least one of them doesn't O.o).

> doesn't take too long in release builds

I absolutely disagree with this aspect. The threshold is arbitrary and there is no sane threshold that could account for the machine running the test, system load, compiler, file system, moon phase, ... Tests can be arbitrarily slow on insane configurations, but should always pass if they compute the correct results. Thus I would prefer to leave *any* time measurement out of our gtests.

You are right that this is out of scope here, and it makes sense to keep it until we have something better.
If we had the engineering resources, I would love to have the moveit_benchmark_suite (MBS :)) become a standalone performance testing suite (which it already is to some level) that you can run with multiple configurations and versions of the software stacks to compare how changes impact any performance metric on a variety of scenarios. That would include pointing out values that differ significantly between different software stacks.
But none of these should be hard tests and they could be treated just like our codecov reports.

rhaschke:
I agree with you, Michael, that a benchmark suite as you describe it, would be great. However, I'm not aware of any framework to compare benchmarks of different snapshots in a similar fashion like codecov. Is there something?
If not, we should keep some key timing tests as part of our hard tests. I agree that they are highly unreliable, but they ensure that we don't break anything completely by accident.

v4hn:
> However, I'm not aware of any framework to compare benchmarks of different snapshots in a similar fashion like codecov. Is there something?

As I wrote above, it was/is my goal for the benchmark suite that was started in last year's GSoC project and [MBS aims to compare different software stacks](https://github.com/captain-yoshi/moveit_benchmark_suite/blob/master/.docs/regression.md). But I don't think it's in a state we could integrate into CI (or a custom runner for fixed ) right away. That being said I did not actually test the current state for some time...

> If not, we should keep some key timing tests as part of our hard tests. I agree that they are highly unreliable, but they ensure that we don't break anything completely by accident.

I agree, that's why I marked my remarks as offtopic. :) Let's hope we can remove them at some point once we have benchmarking setup (if we ever get there...)

PR #3013, opened 2021-12-28
MSA cleanup

rhaschke:
This PR factors out two simple commits from #2938:
- Pass xacro args to SRDF loading as well (fixes an issue reported by @rickstaa somewhere)
- Modernize some more loops (no idea why clang-tidy ignored them)

### DIFF
--- a/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/setup_assistant_widget.cpp
@@ -448,11 +448,9 @@ void SetupAssistantWidget::highlightGroup(const std::string& group_name)
       config_data_->getRobotModel()->getJointModelGroup(group_name);
   if (joint_model_group)
   {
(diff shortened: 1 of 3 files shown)


rickstaa:
@rhaschke Thanks for this pull request. I mentioned a problem with the `xacro` not being used when loading the SRDF file. I can not find the exact comment, but I remember I removed it because of the work needed to implement a fix. I will use the **EDIT:** or ~~striketrough~~ syntax to prevent loss of information. Amazing that you remembered it and implemented it!

DLu:
@rhaschke Do you have an example of a config where the same xacro args need to get passed to the URDF and SRDF? I'm working on porting this to ROS 2.

rhaschke:
[`panda_moveit_config`](https://github.com/ros-planning/panda_moveit_config.git) uses the argument `hand:=[true | false]`.

PR #3015, opened 2021-12-29
Backport #3009: Move MoveItErrorCode to moveit::core

rhaschke:
In contrast to #3009, I decided to not deprecate the old definitions here (but only redirect them).


v4hn (approved):
True, there's no reason not to backport this to melodic without the deprecation.

rhaschke:
> unless you add typos ;)

Haha :wink: There were dozens of conflicts when cherry-picking and I tend to use GHA for compiling instead of my own machine...

rhaschke:
I will merge despite the (unrelated) clang-tidy issues. Looks like the default clang compiler was updated and is pickier now...

PR #3022, opened 2022-01-06
Removed wrong transform in sphere decomposition

Martin-Oehler:
### Description

For optimization-based trajectory planning, collision bodies can be sampled with spheres with `collision_detection::determineCollisionSpheres`. The current implementation introduces a wrong relative transform, leading to a wrong positioning of these spheres. The error is only noticeable, when collision bodies have an offset from the respective link pose.

Before the fix:
moveit_sphere_decomposition_bugged

With the fix:
moveit_sphere_decomposition_fixed

### Checklist


rhaschke (approved):
Thanks. I guess this is a left-over from #2037, which introduced relative transforms.

Martin-Oehler:
Thanks for approving. However, the code change in #2037 is not related to this.

rhaschke:
Merged, due to obvious bug with a simple and proven fix.

PR #2960, opened 2021-11-15
Add MSA write gazebo urdf feature

rickstaa:
### Description

This pull request adds several improvements to the simulation tab of the MSA. The most significant change is that the Gazebo URDF shown in the simulation tab was not saved in the old version. As a result, when the provided URDF was incompatible with the Gazebo, the `gazebo.launch` file was not working. The new version now ensures that a gazebo URDF is saved as a MoveIT configuration file when the provided URDF is invalid. It further provides users with the option to overwrite the provided URDF and fixes a small bug:

- [fix(msa): fix redunant transmission tag create bug](https://github.com/ros-planning/moveit/pull/2960/commits/74f7bd3f4c7ad60af5f7ac8f62ee85743ff007ae)

I first tried 22284a59f6bc8c3d7a138681c261f04df009d286. However, in the end, I added 74f7bd3f4c7ad60af5f7ac8f62ee85743ff007ae since it does a better job at checking whether a urdf is Gazebo compatible. 

#### New configuration no sim

new_now_sim

The `gazebo_panda.urdf` is **NOT**  added.

#### New configuration + sim + compatible xacro

Here I use https://github.com/frankaemika/franka_ros/pull/199 for the xacro description. This model is Gazebo compatible. I however had to add an inertial tag to the [panda_link8_sc link](https://github.com/frankaemika/franka_ros/blob/63ecb779ae49d023b5ef93434a457bb599c553ed/franka_description/robots/arm.xacro#L271-L289) since my script has no knowledge of why this link has no inertia.

```diff
<link name="${prefix}link8_sc">
      <collision>
        <origin xyz="0.0424 0.0424 -0.0250" rpy="${pi} ${pi/2} ${pi/2}"/>
        <geometry>
          <cylinder radius="${0.03+safety_distance}"  length="0.01" />
        </geometry>
      </collision>
      <collision>
        <origin xyz="0.0424 0.0424 -0.02" rpy="0 0 0"/>
        <geometry>
          <sphere radius="${0.03+safety_distance}"  />
        </geometry>
      </collision>
      <collision>
        <origin xyz="0.0424 0.0424 -0.03" rpy="0 0 0"/>
        <geometry>
          <sphere radius="${0.03+safety_distance}"  />
        </geometry>
      </collision>
+      <inertial>
+        <mass value="0"/>
+       <inertia ixx="1.0" ixy="0.0" ixz="0.0" iyy="1.0" iyz="0.0" izz="1.0" />
+     </inertial>
    </link>
```

[...]

PR #3027, opened 2022-01-16
Make demo gazebo work

hideo67:
### Description

Hi. This is the first time submitting a PR for moveit. Please tell me if I am missing any procedures.

I modified SA so that demo_gazebo.launch will work out of the box for simple robots that do not require special treating at the Gazebo side (e.g. grippers with mimic joints need special treating).

The Gazebo side plugin to talk with ROS, "gazebo_ros_control" will be given a namespace instead of running in the root namespace.  The default for gazebo_ros_control is to use the URDF robot name as the namespace. I  followed this.  This will require some support in the config files, and maybe the current SA tried to avoid such support.  However, trying to make it run in "/" seems to me, make the config files harder to understand.

- 3 template files changed
- 1 template file added
- 1 generated config file added
- 2 source files and 1 header changed.

The changes are:
- The URDF generated in the simulation tab in SA will specify the above mentioned robot name namespace in robotNamespace tag (moveit_config_data.cpp 588). This will make the gazebo_ros_control plugin read ros params under that namespace, and expose actions, services, topics under that namespace too.
- When populating Gazebo with the robot model by the gazebo_ros/spawn_model node, the model name must be specified, but the current source gives a fixed "robot" here. It is changed to be the URDF robot name. (gazebo.launch 22)
- The controller manager inside gazebo_ros_control will look for controller manager parameters under the robot name namespace.  So the rosparam path must be changed accordingly. The namespace will be added by a "ns" parameter to rosparam when loading the controller definitions. Two sets of controllers need to be loaded this way- gazebo_controllers.yaml and ros_contollers.yaml. (gazebo.launch 25-28)
- The controller manager inside gazebo_ros_control must receive a spawn request to load those controllers. The controller_manager/spawner node must have the namespace specifed. (gazebo.launch 30-32)
- The MoveitConrollerManager plugin (not be confused with ContollerManager) that runs within MoveIt needs to know where the ROS controller actions are waiting to be called. The existing simple_moveit_controllers.yaml file has the same purpose but the action namespace must be changed to be prefixed by the gazebo_ros_control's namespace. So I chose to generate a new file gazebo_moveit_controllers.yaml based on the generator function for simple_moveit_controllers.yaml (moveit_config_data.cpp 957- 985)
- The JointState events that comes out of gazebo_ros_control will now be prefixed by the namespace. MoveIt will subscribe to /joint_states, so the events need to be mapped.  I chose to add a joint_state_publisher node for this purpose. (demo_gazebo.launch 16-19)
- For some reason, the current demo_gazebo.launch will not launch a robot_state_publisher. It is launched only when a fake MoveItControllerManager is selected.  I could not guess any motivation behind this design, AFAIK robot_state_publisher is always necessary.  So I took the node tag that launches robot_state_publisher outside of the conditional. (demo.launch 40-44)
- To configure MoveItControllerManager to work with Gazebo, a new file gazebo_moveit_controller_manager.launch.xml is added.  This is specified by an argument to demo.launch. (demo_gazebo.launch 24)

The fix has worked for a simple 6-DOF robot URDF that I have not published.

I tried to make it work with one of the tutorial example URDFs, but they are rather complicated. PANDA has a gripper with mimic joints.

### Checklist

### DIFF
--- a/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
+++ b/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
@@ -412,19 +412,6 @@ public:
                                   std::string& relative_filepath) const;
 
   /**
-   * Get the package name from package.xml.
-   *
-   * For cases where the name differs from the directory.
-   * (Example: panda_description in moveit_resources.)
-   *
-   * @param package_xml_path path to "package.xml"
-   * @param[out] package_name package name will be stored here.
-   * @return true when all went OK. false if parsing fails or tag not found.
-   */
-  bool extractPackageNameFromPackageXml(const std::string& package_xml_path,
-		  std::string& package_name) const;
-
-  /**
    * Resolve path to .setup_assistant file
    * @param path resolved path
    * @return true if the path could be resolved
--- a/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
+++ b/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
@@ -412,6 +412,19 @@ public:
                                   std::string& relative_filepath) const;
 
   /**
+   * Get the package name from package.xml.
+   * 
+   * For cases where the name differs from the directory.
+   * (Example: panda_description in moveit_resources.)
+   *
+   * @param package_xml_path path to "package.xml"
+   * @param[out] package_name package name will be stored here.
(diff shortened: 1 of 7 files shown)


rhaschke:
Related to #2960.

hideo67:
After submitting the PR, I found that the original code does work without this modification, as long as you use "/" as the namespace for gazebo_ros_control.

So I'd like to close this PR.  There are already a lot for the committers to review.

Issue #2812, opened 2021-08-13
Defining 'pipelines_names' parameter causes CHOMP pipelines to be defined


pvanlaar:
### Description

During the switch from MoveItCpp to Move Group Interface, I forgot to remove a config file, which was still loaded on startup. In this file the pipeline names were still loaded:

~~~yaml
planning_pipelines:
  pipeline_names:
    - ompl
~~~

This caused two pipelines to be loaded with CHOMP planners. This can be verified by calling the query service (`rosservice call /query_planner_interface`). This results in the following output:
~~~
planner_interfaces: 
  - 
    name: "CHOMP"
    pipeline_id: "ompl"
    planner_ids: 
      - CHOMP
  - 
    name: "CHOMP"
    pipeline_id: "pipeline_names"
    planner_ids: 
      - CHOMP
~~~

In this case, there were no planning pipelines defined after loading the config file, but if one is defined with the same name (for example "ompl"), this pipeline can be overwritten. This was weird behaviour because I only used OMPL and have not referenced CHOMP planners anywhere.

### Environment
* ROS Distro: Noetic
* OS Version: Ubuntu 20.04
* Source: detached at tag 1.1.5

### Steps to reproduce
In this [repo](https://github.com/pvanlaar/moveit_pipeline_config_issue) a minimum example can be found.
Required are the following packages (defined in the example package):
- moveit 1.1.5
- panda_moveit_config 

1. Launch the move_group node including the `planning_pipelines` configuration
2. Call the `/query_planner_interface` service to verify

### Expected behaviour
No pipelines should be defined and probably a warning should be reported. 

### Actual behaviour
One or more CHOMP planners are defined. These planners are not the default planner of MoveIt (OMPL's RTTConnect as far as I know) and do not support pose targets.

### Next steps
Do you agree with the analysis/solution? I am happy to provide a pull request if we can agree on a solution.
Is there a place for reference documentation? In this case, the (setup) differences between MoveGroup and MoveItCpp would be nice to have.

fi6:
Got the same problem, calling service responds as follow:
```
rosservice call /my_gen3/query_planner_interface                                                           yida@yida-bigai-pc  15:39:12
planner_interfaces: 
  - 
    name: "CHOMP"
    pipeline_id: "arm"
    planner_ids: 
      - CHOMP
  - 
    name: "CHOMP"
    pipeline_id: "capabilities"
    planner_ids: 
      - CHOMP
  - 
    name: "CHOMP"
[...]

PR #3036, opened 2022-01-25
Improve loading of planning pipelines

rhaschke:
Ensure that planning pipelines considered in `~/planning_pipelines/*` actually have a `planning_plugin` parameter defined.
Otherwise, issue an error message. Looks like that's a common error (e.g. #2812) when migrating from Melodic's config scheme.


henningkayser (approved):
Good improvements!

PR #3040, opened 2022-01-27
pilz: Add type for overloaded function specification

jspricke:
### Checklist


simonschmeisser:
Could you please motivate a bit why this verbosity is required?

jspricke:
>Could you please motivate a bit why this verbosity is required?

`setJointGroupPositions` has two definitions:

```
void setJointGroupPositions(const JointModelGroup* group, const std::vector<double>& gstate)
void setJointGroupPositions(const JointModelGroup* group, const Eigen::VectorXd& values);
```

With recent versions of Eigen and gcc (Debian unstable) the compiler 
can't decide which one to use and fails.

rhaschke (approved):
Probably, Eigen provides initialization from brace-initializer list as well now...

simonschmeisser:
Thanks for clarifying!

simonschmeisser (approved):

PR #3056, opened 2022-02-09
Added special case for sphere bodies in sphere decomposition

Martin-Oehler:
### Description

For optimization-based trajectory planning, robot collision bodies are approximated with spheres. This is generally done by computing the bounding cylinder of every body and inserting spheres along the cylinder. This is done even in the case of a sphere collision body, even though the sphere decomposition of a sphere is trivial.

I fix this problem by adding a special case for sphere bodies in the sphere decomposition function.

Before:
moveit_pr_sphere_decomposition_before

After:
moveit_pr_sphere_decomposition_after

### Checklist

### DIFF
--- a/moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
+++ b/moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
@@ -47,18 +47,25 @@ std::vector<collision_detection::CollisionSphere>
 collision_detection::determineCollisionSpheres(const bodies::Body* body, Eigen::Isometry3d& relative_transform)
 {
   std::vector<collision_detection::CollisionSphere> css;
-
-  bodies::BoundingCylinder cyl;
-  body->computeBoundingCylinder(cyl);
-  unsigned int num_points = ceil(cyl.length / (cyl.radius / 2.0));
-  double spacing = cyl.length / ((num_points * 1.0) - 1.0);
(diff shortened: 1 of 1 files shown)


simonschmeisser:
Given that you seem to know how to use the Distance Field based collision checker would you mind writing down some of your knowledge? It does not seem to be covered in the tutorials yet.
I'll try to do a code review later on

v4hn (approved):
Thank you for providing the patch. The reasoning makes sense and your code looks good.

This is hard to evaluate with no functionality setup on top of it.
I do trust you enough here to approve this as-is, but would it be too much to ask for a test, also to make sure nobody breaks this again in the future?

Martin-Oehler:
> This is hard to evaluate with no functionality setup on top of it.

Yes, I see the problem here. This code is also used by the CHOMP trajectory planner, but I am not familiar with it.

> I do trust you enough here to approve this as-is, but would it be too much to ask for a test, also to make sure nobody breaks this again in the future?

How would such a test look like? Something like a URDF with only one sphere collision body and a test that expects just one sphere sample to be returned?

v4hn:
> How would such a test look like? Something like a URDF with only one sphere collision body and a test that expects just one sphere sample to be returned?

That's what I had in mind, but you can simply pass in a sphere body instead of starting out with a URDF.

rhaschke (approved):

PR #3063, opened 2022-02-17
Fix Python versioned dependency

jspricke:
### Checklist

### DIFF
--- a/moveit_core/package.xml
+++ b/moveit_core/package.xml
@@ -24,8 +24,8 @@
   <depend>boost</depend>
   <depend>eigen</depend>
   <depend>bullet</depend>
-  <depend condition="$ROS_DISTRO == noetic">python3</depend>
-  <depend condition="$ROS_DISTRO != noetic">python</depend>
+  <depend condition="$ROS_PYTHON_VERSION == 3">python3</depend>
+  <depend condition="$ROS_PYTHON_VERSION == 2">python</depend>
   <depend>eigen_stl_containers</depend>
   <depend condition="$ROS_DISTRO != noetic">libfcl-dev</depend>
   <depend condition="$ROS_DISTRO == noetic">fcl</depend>


rhaschke (approved):

rhaschke (commented):
Why do you actually need this? Will there be another distro after noetic?
Are you using noetic with python2?

jspricke:
>Why do you actually need this? Will there be another distro after noetic?

https://github.com/ros-o :D

rhaschke:
Interesting? Who runs this effort? For a new ROS1 distro, wouldn't you need to update the other distro-conditioned dependencies as well?

jspricke:
>Who runs this effort?

@v4hn, me and a some others, you are very welcome to join the effort :).

>For a new ROS1 distro, wouldn't you need to update the other distro-conditioned dependencies as well?

I did a test build on Debian unstable and this was the only one I found 
breaking.

simonschmeisser:
Nice effort, where do you communicate?

v4hn:
The `ros-o` organization started out a few days ago (as you can see) because I tried to build a ROS workspace on a non-Ubuntu newer-than-20.04 Linux (which is my normal development machine) and had to scratch together and write patches for newer versions of system dependencies of standard ROS packages. Many issues were experienced by other users already and wrote patches, some even reported to upstream and were not heard (yet). Prominently, the ROS for Debian project, spearheaded by @jspricke, collected quite a lot already in their debian packages. Because these are distribution-specific I opted to collect general patches in a github organization instead.

Aside from that there is no plan/discussion (yet). The idea is not to counteract community-migration to ROS2, but simply to prevent immediate bit-rot of ROS One which includes a lot of valuable code bases that are not directly there in ROS2 (yet (tm)).

I would expect that at least some upstream repositories will be willing to support this effort in a new `obese-devel` branch themselves, and we will just remove the ros-o forks again if anyone does. Be aware that the `ros-o` effort might become rolling release with future releases of upstream dependencies / OS version releases.

Also please note that there is currently no buildfarm/CI/rosdistro yaml set up for any of this.
If anyone wants to contribute something along that line, please get in touch with @jspricke or myself.
Also anyone else from the community is very much invited to join the organization and contribute.

PR #3065, opened 2022-02-19
do *not* automatically load robot description

v4hn:
... from move_group.launch .

This was a clear bug introduced in 261a0c4c9041d6fd2df6e7cd7d1160321c047c8b .
MoveIt should not overwrite a previously uploaded robot description.
It should only provide it optionally in demo mode.

I just found this issue mentioned in https://github.com/ros-industrial/universal_robot/pull/538#issuecomment-930073835 ,
but nobody reported or addressed it back here since then. I guess @gavanderhoorn had too many other things to do or didn't take it to be an upstream issue. :)

### DIFF
--- a/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
+++ b/moveit_setup_assistant/templates/moveit_config_pkg_template/launch/move_group.launch
@@ -37,7 +37,7 @@
                 " />
   -->
 
-  <arg name="load_robot_description" default="true" />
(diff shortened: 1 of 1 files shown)


mikeferguson (approved):
I actually recently ran into this and also forgot to report it (with a URDF that depends on environment variables to select various features - causing the URDF on a robot to be missing the arm, which really confused me).

rhaschke (approved):
I don't care very much about this default.

Issue #3073, opened 2022-03-01
collisions_updater not generating a complete matrix


LoyVanBeek:
### Description

A colleague of mine noticed that when using `rosrun moveit_setup_assistant collisions_updater ...  --default --always --keep --verbose ...` the allowed collisions matrix would lose almost all of it's entries, even ones that should obviously be there. 

### Your environment
* ROS Distro: [Noetic]
* OS Version: e.g. Ubuntu 20.04
* Source build: `tag/1.1.8` or rather my debug branch: https://github.com/ros-planning/moveit/compare/1.1.8...LoyVanBeek:debug/collision_updater with some extra logging

### Steps to reproduce
0. `cd ~/moveit_ws/src`
1. `git clone git@github.com:ros-industrial/universal_robot.git`
2. `cd ur10_e_moveit_config`  # and check content of `config/ur10e.srdf`
3. `catkin build --this`
4. Delete all the `<disable_collisions ...>` tags
5. `rosrun moveit_setup_assistant collisions_updater --config-pkg ur10_e_moveit_config --default --always --keep --trials 100000`
6. `git diff -w` to ignore changes to whitespace
7. Observe the diff to the collision matrix. Each link appears only once as `link1`, instead of some multiple times as it were before.

### Expected behaviour
The collision matrix should be unchanged, give or take some formatting. 

### Actual behaviour
Only a few pairs remain, in this case
```xml
<disable_collisions link1="base_link" link2="shoulder_link" reason="Adjacent"/>
<disable_collisions link1="ee_link" link2="wrist_1_link" reason="Never"/>
<disable_collisions link1="shoulder_link" link2="wrist_2_link" reason="Never"/>
<disable_collisions link1="wrist_1_link" link2="wrist_3_link" reason="Never"/>
```
Notice that all the `link1`s are unique while this was not the case before. 

### Backtrace or Console output
Console output using my debug branch: 
https://gist.github.com/LoyVanBeek/d782e4fadf2fa1d88f0ab4aaa064e43e

Notice the occurrences of [`filtered.insert`](https://github.com/ros-planning/moveit/compare/1.1.8...LoyVanBeek:debug/collision_updater#diff-afd5f8e1ad86bce85a80043828f0b90044790924e39d7ba7527a2a4fb99c29a7R172). Only 4 of those while there are much more pairs that should be disabled. 
[...]

PR #3076, opened 2022-03-03
Fix collisions_updater set comparison

LoyVanBeek:
### Description

After debugging the issue in https://github.com/ros-planning/moveit/issues/3073 I found that the `CollisionPairLess` set comparison operator for `srdf::Model::CollisionPair`s somehow only checked uniqueness of `link1_` instead of both.

So, instead of making a set `{(a, b), (a, c), (a, d), (b, c), (b, d)}` it only allowed `{(a, b), (b, c)}`, which results in an incomplete allowed collision matrix. 

### Checklist
- [x] ~~Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/)~~
- [x] ~~Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes~~
- [x] ~~Include a screenshot if changing a GUI~~

### DIFF
--- a/moveit_setup_assistant/src/collisions_updater.cpp
+++ b/moveit_setup_assistant/src/collisions_updater.cpp
@@ -120,7 +120,12 @@ struct CollisionPairLess
 {
   bool operator()(const srdf::Model::CollisionPair& left, const srdf::Model::CollisionPair& right) const
   {
-    return left.link1_ < right.link1_ && left.link2_ < right.link2_;
+    // bool result = left.link1_ < right.link1_ && left.link2_ < right.link2_;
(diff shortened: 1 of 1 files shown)


rhaschke (approved):
Thanks for pointing out the issue in #3073 and even better, providing a fix here!
I approve but suggest avoiding the creation of pairs, which is much less efficient: https://godbolt.org/z/xT8xT9P5v

fmessmer (approved):
I can confirm this patch gives me an expected collision matrix again

thanks @LoyVanBeek 
thanks @rhaschke

PR #3078, opened 2022-03-04
Add ptr definitions for TimeParameterization classes

v4hn:
These were not there before the polymorphy change either,
but they become a bit more important with it.

@rhaschke

### DIFF
--- a/moveit/package.xml
+++ b/moveit/package.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package format="2">
   <name>moveit</name>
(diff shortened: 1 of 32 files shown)


rhaschke (approved):
You beat me pushing this commit. Thanks anyway!

rhaschke:
I will prepare a patch release over the weekend. Hence, please disregard/drop my additional commit before merging.
Background: there is a regression (https://github.com/ros-planning/moveit/pull/3076) that should be released asap.

PR #3090, opened 2022-03-14
protect RobotState in RobotStateDisplay with mutex

simonschmeisser:
### Description

Currently the `RobotStateDisplay` is not thread-safe and a new incoming message updating the `robot_state_` while it's updated for rendering can cause a crash

### DIFF
--- a/moveit_ros/robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
+++ b/moveit_ros/robot_interaction/include/moveit/robot_interaction/locked_robot_state.h
@@ -62,6 +62,7 @@ class LockedRobotState
 public:
   LockedRobotState(const moveit::core::RobotState& state);
   LockedRobotState(const moveit::core::RobotModelPtr& model);
+  LockedRobotState();
 
   virtual ~LockedRobotState();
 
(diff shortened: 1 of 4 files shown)


JafarAbdi (commented):
What do you think about using [LockedRobotState](https://github.com/ros-planning/moveit/blob/master/moveit_ros/robot_interaction/include/moveit/robot_interaction/locked_robot_state.h) from robot interaction? looks like it's suited for this use case

simonschmeisser:
Actually got it to work but I'm not quite convinced this is nicer than the plain old mutex ... if it shall be this variant, please squash on merge

simonschmeisser:
friendly ping @JafarAbdi

rhaschke:
> Currently the `RobotStateDisplay` is not thread-safe and a new incoming message updating the `robot_state_` while it's updated for rendering can cause a crash

Actually, I cannot confirm this statement. RobotStateDisplay is served by the default, global message queue, which in turn is processed synchronously within rviz' main update loop. Thus, it's not possible that messages are incoming while they are processed.
I guess your are not using standard rviz, but your own version, serving the global message queue from several threads/spinners? I remember that we discussed this issue over in rviz a while ago...

simonschmeisser:
That's right, I use rviz as a library with AsyncSpinner (as does `rqt_rviz` afaik) in my application.

rhaschke:
Closing as too specific for your use case.

PR #3093, opened 2022-03-15
MSA: extended ACM editing

rhaschke:
As a follow-up to #2938, this PR aims to allow for editing the extended ACM in MSA, i.e. defining default-collision links, re-enabling and disabling collision pairs.
- [x] Display ACM considering default-collision links and re-enabled pairs
- [ ] ~~Choose a better data representation?~~
  For now, the CollisionMatrixModel directly operates on the `SRDFWriter`'s members. While that's the native data structure holding the relevant info, accessing any info requires several `string` compares. In the past, the `reason` strings were mapped onto enums internally and a map (`LinkPairMap`) was used instead of vectors.
However, I didn't notice a dramatic slowdown with a robot of ca. 20 links. So, maybe it's not worth the effort...
- [x] Reenable editing, particularly allow adding/removing defaullt-collision links.

### DIFF
--- a/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
+++ b/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
@@ -37,11 +37,10 @@
 #pragma once
 
 #include <moveit/macros/class_forward.h>
-#include <moveit/planning_scene/planning_scene.h>                     // for getting kinematic model
-#include <moveit/setup_assistant/tools/compute_default_collisions.h>  // for LinkPairMap
-#include <yaml-cpp/yaml.h>                                            // outputing yaml config files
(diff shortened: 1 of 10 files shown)


v4hn:
> Choose a better data representation?

I just tried to edit the ACM of our PR2+Shadow hand with 72 links and it did not even respond to clicks on empty checkboxes in the matrix view. Either I did not wait long enough, or you have a new bug in there now?

Assuming the representation is fine with everything <100 links, I would not bother to clean up anything under the hood here.

rhaschke:
Editing isn't yet enabled again - and it was stated (via an empty checkbox) in the primary comment :wink:
As you answered my question, I can continue now without further refactoring.

rhaschke:
I finally found some time to finalize this. Editing is possible again and defaults can be changed via context menu in header region.

PR #3100, opened 2022-03-24
URDF to Collision Object and Collision Object Database Fix  #2668

### LINKED ISSUE #2668: Is there an standard mechanism to build a collision object from an URDF?
**Problem**
I would like to automatically load collision objects described by their URDF into a planning scene. These URDF have several links attached by fixed joints and I would like to publish their TF. I would like that the position of these frames are automatically updated when objects are attached to a robot.  This problem brings me to the following questions

1. Is there a standard way to convert URDF (which only contains fixed joints) to a collision object message?
2. Is there a standard mechanism to publish the frames of the collision objects? 

**Alternatives I have considered**
To solve 1 I'm currently implementing an URDF parser which takes the links and fixed joints from an URDF and build a collision_object message. There is a thread [here](https://answers.ros.org/question/281826/add-collison-object-from-a-urdf-file-in-moveit/), but I did not found more information about how to implement it.

To publish the object frames I would use a robot_state publisher for each piece and program a node that listen to the planning scene topic/service and publish the  base frame of each piece or all the frames of the collision object primitives. 

**Additional context**
I'm building an automatic assembly system where different components are defined as URDF with several links. Some of these links are sockets, and the assembly of two components is done by bringing their sockets frames to the same pose. I would like to update the positions of the pieces taken by the robot with moveit by attaching the collision objects to the robot.

rafaelrojasmiliani:
### Description
This pull request addresses #2668. It contains two contributions
1. a function that converts a urdf into a collision object message [declared here](https://github.com/rafaelrojasmiliani/moveit/blob/a479ed0e034892a893d15045c7e2678932ba6a32/moveit_core/planning_scene/include/moveit/planning_scene/collision_object.h#L53) and [defined here](https://github.com/rafaelrojasmiliani/moveit/blob/a479ed0e034892a893d15045c7e2678932ba6a32/moveit_core/planning_scene/src/urdf2collisionobject.cpp#L82).
2. A warehouse for a Collision object database [declared here](https://github.com/rafaelrojasmiliani/moveit/blob/rafaelrojas_urdf2collisionobject/moveit_ros/warehouse/warehouse/include/moveit/warehouse/collision_object.h) and [defined here](https://github.com/rafaelrojasmiliani/moveit/blob/rafaelrojas_urdf2collisionobject/moveit_ros/warehouse/warehouse/src/collision_object.cpp)

This PR has some differences with respect to Captain Yoshi's [urdf_to_scene](https://github.com/captain-yoshi/urdf_to_scene) which has not be incorporated into moveit yet. However it have the same effect. The differences are
1. This implementation is done in a single function not a class
2. This implementations handle more possible errors

### Checklist


moveit_core/planning_scene/src/urdf2collisionobject.cpp
```diff
@@ -0,0 +1,331 @@
...
+}
+
+moveit_msgs::CollisionObject urdf_to_collision_object(const urdf::Model& _urdf_model)
+{
+  KDL::Tree kdl_tree;
+  moveit_msgs::CollisionObject result;
+
+  moveit_msgs::CollisionObject result_empty;  // empty collision object should be const
+  // ----------------------------------
+  // -------  Part I ----------
```
rhaschke, line 89: Please split your huge function into small blocks, for example along the parts you already identified.

moveit_core/planning_scene/src/urdf2collisionobject.cpp
```diff
@@ -0,0 +1,331 @@
...
+      pose = link->visual->origin;
+    }
+    else
+    {
+      ROS_WARN("Link ill defined: Link %s has not visual or collision tag", link->name.c_str());
+    }
+
+    if (link->collision or link->visual)
+    {
+      switch (geometry->type)
```
rhaschke, line 151: Converting a link into a pair (primitive, pose) is a good candidate for a separate function as well.

moveit_core/planning_scene/src/urdf2collisionobject.cpp
```diff
@@ -0,0 +1,331 @@
...
+      result.primitives.push_back(it.second.first);
+    }
+    else
+    {
+      ROS_ERROR("cannot find a tranform from %s to %s\n", root_link_name.c_str(), it.first.c_str());
+    }
+  }
+  for (const auto& it : mesh_map)
+  {
+    if (tf_buffer.canTransform(root_link_name, it.first, ros::Time::now()))
```
rhaschke, line 291: This code block is redundant to the previous one -> function.

rhaschke (changes requested):
Sorry for the late reply on this and thanks very much for this valuable contribution.
I have a few remarks, though:
- Please create a separate library for this converter.
- You assume link transforms to be published, e.g. via robot_state_publisher.
  However, if that's not the case, instead of throwing an error and ignoring the link, you could use the (default) zero-joint transform.
- An executable to populate the warehouse from a ROS param (or file) holding a URDF description should be added. That would also serve as a nice example on how to use this feature.
- Please cleanup your file names: I suggest `*urdf_to_collision_object.*` and `collision_object_storage.*`

rafaelrojasmiliani:
Thanks a lot for the review!! Sorry for the delay, I will check list in a couple of months.

PR #3106, opened 2022-03-31
bind() them all (not)

### LINKED ISSUE #2766: avoid `boost`/`std::bind`

v4hn:
Fixes #2766. Replaces #2972 .

@Abishalini will hate me for this changeset for its ease of porting to moveit2... :cry: 
But it was overdue, I managed to do it over a few evenings and I believe it makes the code much more readable.

This PR should get some priority because patches are quite likely to generate conflicts with this due to its size.
I split all non-trivial changes into separate commits with one huge commit in the end addressing all "trivial" (up to identifier naming) cases. During this work I found and reworked a number of internal details, often related to pointer-use where object lifetime was known or APIs that could be reworked with the flexibility of lambdas.

Eventually I added the clang-tidy check that Robert found. It's important to point out that the provided patch will usually work, but ends up with horrible overhead per parameter/capture and no useful identifiers. So users cannot just apply the patch. But then again, nobody will contribute new code that still uses `bind()` anyway... Is what you would hope.

@tylerjw @rhaschke


rhaschke (approved):
Thanks a lot for all this cleanup work. I like all the changes.
However, in a few locations you also changed public API (changing pointer arguments to references to indicate that valid instances are expected.)
 I'm fine with these changes as well, but I'm surprised to see them proposed from you. Don't you argue against public API changes usually?

I guess you already noticed that several tests are broken now.

rhaschke:
I shortly looked into the first failing test and it is very strange: The ModelBasedStateSpaceFactoryPtr returned from `factory_selector`:
https://github.com/ros-planning/moveit/blob/779b7c8b019f70898d4de3189f9261c9697d9b9f/moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp#L316
is NULL although the actually called function returns a valid pointer. Hence, somewhere in the hierarchy of calling this `std::function`, the return value is screwed! Looks like a gcc or libstdc++ bug.

rhaschke:
> I see the same issue with clang 13 (did not get to upgrade to 14 yet), so I'm not sure it's a compiler issue. We probably need to find a workaround whether or not we can pinpoint an external bug. That part of the API is insane anyway. Why pass a function to a method just to call it first thing and ignore it after that...

Yes, probably it is a bug in libstdc++. At some point in the call hierarchy the returned shared_ptr reference is copied (which is wrong in first place) and then released.

> Speaking about API: All the header-specific changes I made address `private` (and in one case `protected`) methods. the mesh helper might be debatable (see my commit message), but I do not consider the rest to be public API.

Sorry, I was convinced to have seen a public declaration. Checking again, you are right, everything is private or protected.

rhaschke:
> That part of the API is insane anyway. Why pass a function to a method just to call it first thing and ignore it after that...

You are right, I didn't noticed this before. I suggested two approaches to fix the issue in https://github.com/v4hn/moveit/pull/3 - both changing protected API of OMPL::PlanningContextManager, which is fine I think. Given your comment, I prefer the last one.

v4hn:
Makes sense. I cherry-picked your proposed modification of the protected `getPlanningContext` method. :+1:

rhaschke:
Closing and reopening to trigger CI...

rhaschke:
I tried to reproduce the shared_ptr issue in a minimal example, but there it works: https://godbolt.org/z/T68jzMYW4.

rhaschke:
@v4hn, I think this is ready to merge after squash-merging the fixups.

v4hn:
> I think this is ready to merge after squash-merging the fixups.

Thank you for spending the time to find my superfluous ampersand that broke the patch... :medal_sports: 
I integrated the fixes and will merge.

PR #3112, opened 2022-04-07
Constraint samplers with seed

tahsinkose:
### Description

This PR adds the ability to specify a seed for constraint samplers through the parameter server, which would be useful in deterministic testing and benchmarking of these objects.

### Checklist
- [x] No API changes

### DIFF
--- a/moveit_core/constraint_samplers/CMakeLists.txt
+++ b/moveit_core/constraint_samplers/CMakeLists.txt
@@ -28,10 +28,11 @@ if(CATKIN_ENABLE_TESTING)
   find_package(orocos_kdl REQUIRED)
   find_package(angles REQUIRED)
   find_package(tf2_kdl REQUIRED)
-
+  find_package(rostest REQUIRED)
   include_directories(SYSTEM ${orocos_kdl_INCLUDE_DIRS} ${angles_INCLUDE_DIRS} ${tf2_kdl_INCLUDE_DIRS})
 
-  catkin_add_gmock(test_constraint_samplers
+  add_rostest_gtest(test_constraint_samplers
+    test/constraint_samplers.test
     test/test_constraint_samplers.cpp
     test/pr2_arm_kinematics_plugin.cpp
     test/pr2_arm_ik.cpp
--- a/moveit_core/constraint_samplers/CMakeLists.txt
+++ b/moveit_core/constraint_samplers/CMakeLists.txt
@@ -31,7 +31,7 @@ if(CATKIN_ENABLE_TESTING)
 
   include_directories(SYSTEM ${orocos_kdl_INCLUDE_DIRS} ${angles_INCLUDE_DIRS} ${tf2_kdl_INCLUDE_DIRS})
 
-  catkin_add_gtest(test_constraint_samplers
+  catkin_add_gmock(test_constraint_samplers
     test/test_constraint_samplers.cpp
     test/pr2_arm_kinematics_plugin.cpp
     test/pr2_arm_ik.cpp
@@ -39,6 +39,7 @@ if(CATKIN_ENABLE_TESTING)
 
   target_link_libraries(test_constraint_samplers
     moveit_test_utils
+    gmock_main
     ${MOVEIT_LIB_NAME}
     ${catkin_LIBRARIES}
     ${angles_LIBRARIES}
--- a/moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
+++ b/moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
@@ -72,7 +72,8 @@ public:
     int rng_seed;
     if (ros::param::get("~constraint_sampler_random_seed", rng_seed))
     {
-      ROS_WARN_STREAM_NAMED("constraint_samplers", "Creating random number generator with seed " << std::to_string(rng_seed));
+      ROS_WARN_STREAM_NAMED("constraint_samplers",
+                            "Creating random number generator with seed " << std::to_string(rng_seed));
       random_number_generator_ =
           std::make_unique<random_numbers::RandomNumberGenerator>(static_cast<uint32_t>(rng_seed));
     }
@@ -320,7 +321,8 @@ public:
     int rng_seed;
     if (ros::param::get("~constraint_sampler_random_seed", rng_seed))
     {
-      ROS_WARN_STREAM_NAMED("constraint_samplers", "Creating random number generator with seed " << std::to_string(rng_seed));
+      ROS_WARN_STREAM_NAMED("constraint_samplers",
+                            "Creating random number generator with seed " << std::to_string(rng_seed));
       random_number_generator_ =
           std::make_unique<random_numbers::RandomNumberGenerator>(static_cast<uint32_t>(rng_seed));
     }
@@ -537,11 +539,11 @@ protected:
   bool validate(moveit::core::RobotState& state) const;
 
   std::unique_ptr<random_numbers::RandomNumberGenerator>
-      random_number_generator_;   /**< \brief Random number generator used to sample */
-  IKSamplingPose sampling_pose_;                                  /**< \brief Holder for the pose used for sampling */
-  kinematics::KinematicsBaseConstPtr kb_;                         /**< \brief Holds the kinematics solver */
-  double ik_timeout_;                                             /**< \brief Holds the timeout associated with IK */
-  std::string ik_frame_;                                          /**< \brief Holds the base from of the IK solver */
+      random_number_generator_;           /**< \brief Random number generator used to sample */
+  IKSamplingPose sampling_pose_;          /**< \brief Holder for the pose used for sampling */
+  kinematics::KinematicsBaseConstPtr kb_; /**< \brief Holds the kinematics solver */
+  double ik_timeout_;                     /**< \brief Holds the timeout associated with IK */
+  std::string ik_frame_;                  /**< \brief Holds the base from of the IK solver */
   bool transform_ik_; /**< \brief True if the frame associated with the kinematic model is different than the base frame
                          of the IK solver */
   bool need_eef_to_ik_tip_transform_; /**< \brief True if the tip frame of the inverse kinematic is different than the
--- a/moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
+++ b/moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
@@ -69,6 +69,17 @@ public:
   JointConstraintSampler(const planning_scene::PlanningSceneConstPtr& scene, const std::string& group_name)
     : ConstraintSampler(scene, group_name)
   {
+    int rng_seed;
+    if (ros::param::get("~constraint_sampler_random_seed", rng_seed))
+    {
+      ROS_WARN_STREAM_NAMED("constraint_samplers", "Creating random number generator with seed " << std::to_string(rng_seed));
+      random_number_generator_ =
+          std::make_unique<random_numbers::RandomNumberGenerator>(static_cast<uint32_t>(rng_seed));
+    }
+    else
+    {
+      random_number_generator_ = std::make_unique<random_numbers::RandomNumberGenerator>();
+    }
   }
   /**
    * \brief Configures a joint constraint given a Constraints message.
@@ -191,7 +202,8 @@ protected:
 
   void clear() override;
 
-  random_numbers::RandomNumberGenerator random_number_generator_; /**< \brief Random number generator used to sample */
+  std::unique_ptr<random_numbers::RandomNumberGenerator>
+      random_number_generator_;   /**< \brief Random number generator used to sample */
   std::vector<JointInfo> bounds_; /**< \brief The bounds for any joint with bounds that are more restrictive than the
                                      joint limits */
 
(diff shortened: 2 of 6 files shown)


moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
```diff
@@ -69,6 +69,18 @@ class JointConstraintSampler : public ConstraintSampler
   JointConstraintSampler(const planning_scene::PlanningSceneConstPtr& scene, const std::string& group_name)
     : ConstraintSampler(scene, group_name)
   {
+    int rng_seed;
+    if (ros::param::get("~constraint_sampler_random_seed", rng_seed))
+    {
+      ROS_WARN_STREAM_NAMED("constraint_samplers",
```
rhaschke, line 75: Why did you decided for a warning here? This is an info at most, preferably a debug message, isn't it?
```suggestion
      ROS_INFO_STREAM_NAMED("constraint_samplers",
```

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
```diff
@@ -69,6 +69,18 @@ class JointConstraintSampler : public ConstraintSampler
...
+                            "Creating random number generator with seed " << std::to_string(rng_seed));
```
rhaschke, line 76: No need to convert the int to a string. The streaming operator supports int.
```suggestion
                            "Creating random number generator with seed " << rng_seed);
```

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
```diff
@@ -69,6 +69,18 @@ class JointConstraintSampler : public ConstraintSampler
...
+      random_number_generator_ =
+          std::make_unique<random_numbers::RandomNumberGenerator>(static_cast<uint32_t>(rng_seed));
```
rhaschke, line 78: Instead of casting here, please use `uint` as the primary data type for _rng_seed_.

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
```diff
@@ -69,6 +69,18 @@ class JointConstraintSampler : public ConstraintSampler
   JointConstraintSampler(const planning_scene::PlanningSceneConstPtr& scene, const std::string& group_name)
     : ConstraintSampler(scene, group_name)
   {
+    int rng_seed;
+    if (ros::param::get("~constraint_sampler_random_seed", rng_seed))
```
rhaschke, line 73: As this parameter is used by the **Joint**ConstraintSampler, use a more specific name?
```suggestion
    if (ros::param::get("~joint_constraint_sampler_random_seed", rng_seed))
```

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
```diff
@@ -191,7 +203,8 @@ class JointConstraintSampler : public ConstraintSampler
 
   void clear() override;
 
-  random_numbers::RandomNumberGenerator random_number_generator_; /**< \brief Random number generator used to sample */
+  std::unique_ptr<random_numbers::RandomNumberGenerator>
```
rhaschke, line 206: Why do you need to change API? Please revert this.

rhaschke (commented):
Thanks for this contribution. Please adjust your code for a minimal changeset. The comments provided for JointConstraintSampler apply to IKConstraintSampler as well.

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
tahsinkose: Well, I had a bias from a production view. Generally, debug messages are not enabled in normal runs, which may cause this to slip away from the eyes. Anyways a simple rosparam lookup would still give the information needed, thus reducing to debug 👍🏼

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
tahsinkose: Without this, `random_number_generator_` gets default constructed and the assignment operator of `random_numbers::RandomNumberGenerator` is deleted, hence I can't do the following:
```
random_number_generator_ =random_numbers::RandomNumberGenerator(rng_seed);
```
This way it enables us to construct it once. This API change is thus necessary.

tahsinkose:
@rhaschke I wanted this capability for both constraint samplers as `IKConstraintSampler` might need that too. Addressed your review, PTAL 🙏🏼

tahsinkose:
Added IKConstraintSampler test.

tahsinkose:
@rhaschke PTAL 🙏🏼

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
```diff
@@ -69,6 +69,16 @@ class JointConstraintSampler : public ConstraintSampler
...
+    int rng_seed;
+    if (ros::param::get("~joint_constraint_sampler_random_seed", rng_seed))
+    {
+      ROS_DEBUG_STREAM_NAMED("constraint_samplers", "Creating random number generator with seed " << rng_seed);
+      random_number_generator_ = std::make_unique<random_numbers::RandomNumberGenerator>(rng_seed);
+    }
+    else
+    {
+      random_number_generator_ = std::make_unique<random_numbers::RandomNumberGenerator>();
+    }
```
v4hn, line 81: Please move the constructor into `default_constraint_samplers.cpp`. There's no need for exposing ros function calls in the headers.
Also you don't need to change the API, because you can just set the random seed parameter in the constructor member initialization (before `{}`).

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
```diff
@@ -305,6 +316,16 @@ class IKConstraintSampler : public ConstraintSampler
...
+    int rng_seed;
+    if (ros::param::get("~ik_constraint_sampler_random_seed", rng_seed))
+    {
+      ROS_DEBUG_STREAM_NAMED("constraint_samplers", "Creating random number generator with seed " << rng_seed);
+      random_number_generator_ = std::make_unique<random_numbers::RandomNumberGenerator>(rng_seed);
+    }
+    else
+    {
+      random_number_generator_ = std::make_unique<random_numbers::RandomNumberGenerator>();
+    }
```
v4hn, line 328: see above

v4hn (commented):
looks good after addressing the open issues.
Determinism is sure important! (and sadly by design MoveIt cannot give a lot of that)

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
tahsinkose: I saw you downvoted my explanation at https://github.com/ros-planning/moveit/pull/3112/files#r845836293. Please provide the full code for list initialization. I don't know of a way to sneak `ros::param::get()` if-else branch into the member initialization before {}

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
tahsinkose: It would be nicer if you explain why you disagree @v4hn. I'm genuinely curious about how this can be done without API* change. 

*`random_number_generator_` is fully internal, so it is not an API. It will simply have zero effects from the users' perspective.

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
rhaschke: I pushed a solution: just move the parameter lookup and creation of the RNG into a separate function to be called from the constructor's member initialization :wink:

rhaschke (approved):

moveit_core/constraint_samplers/include/moveit/constraint_samplers/default_constraint_samplers.h
tahsinkose: OK, that is a way indeed. I was focused on doing this purely in the member initialization. Thanks for it 👍🏼

rhaschke (changes requested):
@tahsinkose, looks like your new tests are never finishing.
Could you please have another look and fix them?

tahsinkose:
@rhaschke I just checked. The parameter server manipulation inevitably led to the existence of ROS master, hence the test needed to be updated to be a rostest.

moveit_core/constraint_samplers/CMakeLists.txt
```diff
@@ -28,17 +28,19 @@ if(CATKIN_ENABLE_TESTING)
   find_package(orocos_kdl REQUIRED)
   find_package(angles REQUIRED)
   find_package(tf2_kdl REQUIRED)
-
+  find_package(rostest REQUIRED)
   include_directories(SYSTEM ${orocos_kdl_INCLUDE_DIRS} ${angles_INCLUDE_DIRS} ${tf2_kdl_INCLUDE_DIRS})
 
-  catkin_add_gtest(test_constraint_samplers
+  add_rostest_gtest(test_constraint_samplers
+    test/constraint_samplers.test
```
tahsinkose, line 35: This seems like an unorthodox addition to `moveit_core` as I didn't see any instances of `rostest` in the package. I guess all `rostest` tests are categorized strictly differently than unit tests from the perspective of MoveIt conventions, but that's just a hunch regarding the status quo. 

If you think this is a serious violation, then I need to move the new added tests to a convenient place in `moveit_ros`.

tahsinkose:
The failing checks seem to be related with https://github.blog/2022-04-12-git-security-vulnerability-announced/. Also see https://github.com/actions/checkout/issues/760

rhaschke:
Hi @tylerjw,
did you run into git issues like [these](https://github.com/ros-planning/moveit/runs/6066739682?check_suite_focus=true) for MoveIt2 as well? If so, how did you solve them?
```
setup_upstream_workspace
$ sudo apt-get -qq install -y --no-upgrade --no-install-recommends python3-vcstool | grep -E 'Setting up' 
Setting up python3-vcstool (0.3.0-1) ...
EEE
=== /home/runner/work/moveit/moveit/.work/upstream_ws/src/geometric_shapes (git) ===
Could not determine remotes: fatal: unsafe repository ('/home/runner/work/moveit/moveit/.work/upstream_ws/src/geometric_shapes' is owned by someone else)
```
The problem only arises if the (upstream) cache was restored.

tahsinkose:
Any update on this @rhaschke @tylerjw? I think this issue should be affecting all PRs right now..

v4hn:
@JafarAbdi

rhaschke:
Closing and reopening to trigger CI.

rhaschke:
@tahsinkose, could you please rebase this onto the latest master (which has fixed CI).
@v4hn, I believe I have seen you rebasing and force-pushing other user's PRs in the past. However, I cannot find a corresponding github interface. How did you achieve that?

v4hn:
if the user allowed edits by maintainers I believe you can simply force-push to their PR branch. I wouldn't know how else I would have done that.

rhaschke:
> If the user allowed edits by maintainers I believe you can simply force-push to their PR branch.

No, that doesn't work. So, @tahsinkose, please rebase yourself. Thanks.

moveit_core/constraint_samplers/CMakeLists.txt
rhaschke: @v4hn, could you review as well and comment on this point? IMO, `moveit_core` isn't ROS-agnostic for a while already, e.g. using ROS logging in a few places. However, ROS parameters are actually not yet used in `moveit_core`.

rhaschke (approved):
The tests pass now. Thanks.

moveit_core/constraint_samplers/CMakeLists.txt
v4hn: I fully agree. If we allow ros parameters here (and we seem to agree we want to do so) then we also need to allow tests for them, so both reading the param as well as adding the ros test are fine by me.

Maybe the ros2 sync will split it out into a separate framework structure, but for this repo I will just merge this as is now.

v4hn:
Thanks for your patience with this @tahsinkose :+1:

tahsinkose:
No worries @v4hn I got used to the elaborate review process of MoveIt maintainers 😝

v4hn:
I'm not sure that's a good thing. ;) But we are volunteers after all :stuck_out_tongue: 
Do you have any idea why [your new tests now fails upstream](https://github.com/ros-planning/moveit/runs/6348486255?check_suite_focus=true) right away?

tahsinkose:
> I'm not sure that's a good thing. ;) But we are volunteers after all stuck_out_tongue Do you have any idea why [your new tests now fails upstream](https://github.com/ros-planning/moveit/runs/6348486255?check_suite_focus=true) right away?

Maybe just bad luck. The tests passed in this PR, not sure why they would fail in upstream 🤔

rhaschke:
Looks like the test is flaky. @tahsinkose, could you please have another look and make the test more robust?

tahsinkose:
It seems, when the true randomness is restored with rosparam deletion, it starts to be flaky 😄 Adding a different seed for the 3rd sampler should solve this issue @rhaschke.

Issue #3082, opened 2022-03-07
Dual arm planning error with IKConstraintSampler


bi3ri:
### Description

![dual arm](https://github.com/bi3ri/dual_arm_demo/blob/main/dual_arm2.gif?raw=true)

I want to plan and move two UR robots (UR5e and UR16e) at the same time. I followed the [example](https://github.com/o2ac/o2ac-ur/tree/main/catkin_ws/src/o2ac_examples) provided by @felixvd.

I am able to set two Poses through the RVIZ GUI and plan and execute a motion with both robots simultaneously (see gif above), but if I use the Moveit cpp API ([like this](https://github.com/bi3ri/dual_arm_demo/blob/main/dual_arm_demo/src/moveit_demo.cpp)) I get the following error:

```shell
[...]
[ERROR] [1646681894.535879134, 851.070000000]: IKConstraintSampler received dirty robot state, but valid transforms are required. Failing.
[ERROR] [1646681894.535901334, 851.070000000]: IKConstraintSampler received dirty robot state, but valid transforms are required. Failing.
[ERROR] [1646681894.535920515, 851.070000000]: IKConstraintSampler received dirty robot state, but valid transforms are required. Failing.
[Thread 0x7fffb97fa700 (LWP 62796) exited]
[ERROR] [1646681894.538607503, 851.073000000]: rocket_and_groot/rocket_and_groot: Unable to sample any valid states for goal tree
[ INFO] [1646681894.538643433, 851.073000000]: rocket_and_groot/rocket_and_groot: Created 1 states (1 start + 0 goal)
[ INFO] [1646681894.538657673, 851.073000000]: No solution found after 0.080671 seconds
[ INFO] [1646681894.538684344, 851.073000000]: Unable to solve the planning problem
[ WARN] [1646681894.538734974, 851.073000000]: More than one constraint is set. If your move_group does not have multiple end effectors/arms, this is unusual. Are you using a move_group_interface and forgetting to call clearPoseTargets() or equivalent?
```

I can not find 

### Your environment
* ROS Distro: noetic
* OS Version: Pop OS 20.04
* Binary 1.1.8-1focal.20220226.080138

### Steps to reproduce
I created a small demo project you can find [here](https://github.com/bi3ri/dual_arm_demo.git).

```shell
mkdir -p ~/dual_arm_ws/src
cd ~/dual_arm_ws/src
git clone https://github.com/bi3ri/dual_arm_demo.git
git submodule update --init --recursive
[...]

PR #3119, opened 2022-04-25
Add and fix dual arm test

cambel:
### Description

I spent too many hours debugging this issue #3082 with not much progress.
My best guess as to the source of the issue is that the `Robot State` gets `dirty` (or just not updated as it used to), so the Sampler does not even bother trying to plan. I don't know why though.

Following @felixvd idea, I created a simple test case. I tested it on the `melodic-devel` branch and the master's branch commit [95ab9e](https://github.com/ros-planning/moveit/commit/95ab9eef58b552fe366f0513487a6f6db5a3e71e), and the test passes with no problem. Then in the `master/noetic-devel` branch, it fails. 

Now, I want to ask, is there an efficient way to use this test case to pinpoint the offending commit?

The test case depends on the [dual arm demo](https://github.com/cambel/dual_arm_demo) environment proposed by @bi3ri. By the way, I think this dual-arm demo is simple enough that it would be a good addition to `moveit_resources`

### DIFF
--- a/moveit_ros/planning_interface/test/dual_arm_robot_state_update.py
+++ b/moveit_ros/planning_interface/test/dual_arm_robot_state_update.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python
+
+import unittest
+import rospy
+import rostest
+import os
+
+import moveit_commander
+
+from moveit_msgs.msg import MoveItErrorCodes, RobotTrajectory
+
+
+class RobotStateUpdateTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(self):
+        self.dual_arm_group = moveit_commander.MoveGroupCommander("dual_arm")
+        self.panda_1_group = moveit_commander.MoveGroupCommander("panda_1")
+        self.panda_2_group = moveit_commander.MoveGroupCommander("panda_2")
+
+    @classmethod
+    def tearDown(self):
+        pass
+
+    def plan(self, target):
+        self.dual_arm_group.set_joint_value_target(target)
+        return self.dual_arm_group.plan()
+
+    def test(self):
+        panda_1_pose = self.panda_1_group.get_current_pose("panda_1_link8")
+        panda_1_pose.pose.position.z -= 0.05
+        panda_1_pose.pose.position.x += 0.05
+
+        panda_2_pose = self.panda_2_group.get_current_pose("panda_2_link8")
+        panda_2_pose.pose.position.z -= 0.05
+        panda_2_pose.pose.position.x -= 0.05
+        self.dual_arm_group.set_start_state_to_current_state()
+        self.dual_arm_group.set_pose_target(
+            panda_1_pose, end_effector_link="panda_1_link8"
+        )
+        self.dual_arm_group.set_pose_target(
+            panda_2_pose, end_effector_link="panda_2_link8"
+        )
+
+        result = self.dual_arm_group.plan()
+        if isinstance(result, RobotTrajectory):
+            self.assertTrue(result.joint_trajectory.joint_names)
+        else:
+            success, plan, planning_time, error = result
+            self.assertTrue(error.val == MoveItErrorCodes.SUCCESS)
+
+
+if __name__ == "__main__":
+    PKGNAME = "moveit_ros_planning_interface"
+    NODENAME = "moveit_test_robot_state_update"
+    rospy.init_node(NODENAME)
+    rostest.rosrun(PKGNAME, NODENAME, RobotStateUpdateTest)
+
+    # suppress cleanup segfault in ROS < Kinetic
+    os._exit(0)
--- a/moveit_core/constraint_samplers/src/union_constraint_sampler.cpp
+++ b/moveit_core/constraint_samplers/src/union_constraint_sampler.cpp
@@ -131,6 +131,7 @@ bool UnionConstraintSampler::sample(moveit::core::RobotState& state, const movei
 
   if (!samplers_.empty())
   {
+    state.updateLinkTransforms();
     if (!samplers_[0]->sample(state, reference_state, max_attempts))
       return false;
   }
@@ -141,7 +142,7 @@ bool UnionConstraintSampler::sample(moveit::core::RobotState& state, const movei
     // but requires a state with clean link transforms as input. This means that we need to clean the link
     // transforms between calls to ConstraintSampler::sample.
     state.updateLinkTransforms();
-    if (!samplers_[i]->sample(state, state, max_attempts))
+    if (!samplers_[i]->project(state, max_attempts))
       return false;
   }
   return true;
--- a/moveit_core/constraint_samplers/src/union_constraint_sampler.cpp
+++ b/moveit_core/constraint_samplers/src/union_constraint_sampler.cpp
@@ -128,24 +128,7 @@ bool UnionConstraintSampler::sample(moveit::core::RobotState& state, const movei
 {
   state = reference_state;
   state.setToRandomPositions(jmg_);
-
-  if (!samplers_.empty())
-  {
-    state.updateLinkTransforms();
-    if (!samplers_[0]->sample(state, reference_state, max_attempts))
-      return false;
-  }
-
-  for (std::size_t i = 1; i < samplers_.size(); ++i)
-  {
-    // ConstraintSampler::sample returns states with dirty link transforms (because it only writes values)
-    // but requires a state with clean link transforms as input. This means that we need to clean the link
-    // transforms between calls to ConstraintSampler::sample.
-    state.updateLinkTransforms();
-    if (!samplers_[i]->project(state, max_attempts))
-      return false;
-  }
-  return true;
+  return project(state, max_attempts);
 }
 
 bool UnionConstraintSampler::project(moveit::core::RobotState& state, unsigned int max_attempts)
--- a/moveit_ros/planning_interface/test/CMakeLists.txt
+++ b/moveit_ros/planning_interface/test/CMakeLists.txt
@@ -16,6 +16,7 @@ if (CATKIN_ENABLE_TESTING)
   add_rostest(python_move_group.test)
   add_rostest(python_move_group_ns.test)
   add_rostest(robot_state_update.test)
+  add_rostest(dual_arm_robot_state_update.test)
   add_rostest(cleanup.test)
 
   SET(HELPER_LIB moveit_planning_interface_test_serialize_msg_cpp_helper)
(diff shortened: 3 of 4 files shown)


rhaschke:
> Now, I want to ask, is there an efficient way to use this test case to pinpoint the offending commit?

Use [git bisect](https://git-scm.com/docs/git-bisect).

> The test case depends on the [dual arm demo](https://github.com/cambel/dual_arm_demo) environment proposed by @bi3ri. By the way, I think this dual-arm demo is simple enough that it would be a good addition to `moveit_resources`

Having a dual arm test is would be great. Could you assemble a corresponding config from existing moveit_resources' robots as well?

cambel:
@rhaschke Thank you!

What a journey, but here we are:
```shell
a5607832744a38d6932dfea1c98ad4978e739c19 is the first bad commit
commit a5607832744a38d6932dfea1c98ad4978e739c19
Author: werner291 <werner@wernerkroneman.nl>
Date:   Tue Sep 14 12:48:57 2021 +0200

    Fix #2811 (#2872)
    
    This is a PR for #2811

:040000 040000 8c0e5f69d3d2e7d2669c78db2507880e7d125441 37f61fd50260ba433412c10da0c7d56f7844725e M	moveit_planners
```

I confirmed that reverting that commit fixed the problem. But maybe @v4hn and @werner291 can help us understand why the change was needed and how to keep the change but fix the underlying issue.

Later I will also add a URs dual environment to the moveit_resources for the test case here. It will also be very useful for the PR #2810.

ravnicas:
@cambel Thank you alot, i also linked to this issues from moveit2, since i was fighting the same behaviour on galactic and rolling.

cambel:
@bi3ri,

I did a quick test with TRAC-IK but it seems that that issue persists.
I plan to check that starting next week, hopefully, it does not take as long to find the problem.

simonschmeisser:
Contrary to the discussion in #2872 / #2811 there seem to be some differences between calling `sample` and `project`. My understanding of constraints is unfortunately limited so far but differences I found by now are:

`UnionConstraintSampler` randomizes the input for the JointModelGroup before calling the individual samplers in `sample` but does not in `project`: https://github.com/ros-planning/moveit/blob/3cbcbadf17c861f7672a7d931b8fcfb515193791/moveit_core/constraint_samplers/src/union_constraint_sampler.cpp#L130

`IKConstraintSampler` passes in the current position as seed with `project` but not with `sample` (in the first iteration(?)): https://github.com/ros-planning/moveit/blob/3cbcbadf17c861f7672a7d931b8fcfb515193791/moveit_core/constraint_samplers/src/default_constraint_samplers.cpp#L629

Which sampler is in use get's decided here: https://github.com/ros-planning/moveit/blob/3cbcbadf17c861f7672a7d931b8fcfb515193791/moveit_core/constraint_samplers/src/constraint_sampler_manager.cpp#L56

or in prose here:

 https://github.com/ros-planning/moveit/blob/3cbcbadf17c861f7672a7d931b8fcfb515193791/moveit_core/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler_manager.h#L91

moveit_ros/planning_interface/test/dual_arm_robot_state_update.py
```diff
@@ -32,11 +33,13 @@ def test(self):
         rocket_pose.pose.position.x += 0.05
 
         groot_pose = self.groot_group.get_current_pose("groot_tool0")
-        rocket_pose.pose.position.z -= 0.05
-        rocket_pose.pose.position.x -= 0.05
+        groot_pose.pose.position.z -= 0.05
```
simonschmeisser, line 36: it's slightly misleading to "hide" non formatting related changes like this in a commit titled "Fix format"

also if you install and enable pre-commit, this should be fixed automatically

simonschmeisser (commented):
If you want you could add a temporary commit referencing your moveit_resources PR to make this pass but the main work will of course be to figure out why that troublesome commit has been made and if we should revert it or fix it in some other way

moveit_ros/planning_interface/test/dual_arm_robot_state_update.py
cambel: Yeah, sorry, I re-organized the commits, just two the one with the test and the one reverting the issue being addressed here.

cambel:
Friendly ping to @v4hn

v4hn:
looking at this briefly from my phone I would think the actual error is that [UnionConstraintSampler should not call sample for any but the first sampler here](https://github.com/ros-planning/moveit/blob/master/moveit_core/constraint_samplers/src/union_constraint_sampler.cpp#L144) and call project for them instead (sample once and project everything on that result). I did not test this though, so it might bring in other issues to change it.

cambel:
@v4hn thanks for checking the problem. I think that makes sense to me since the code is split into the first "sampling" and then a `for loop` for the rest of the samplers, instead of just having a single `for loop` as is done in the `project` method. 

It still didn't work without updating the link transforms before the first sampling. But after adding that it seems to work fine according to my test case.

rhaschke:
Obviously, the difference between `ConstraintSampler::sample()` and `::project()` is not yet well documented.
Naively, I would agree to @werner291's interpretation (https://github.com/ros-planning/moveit/issues/2811#issuecomment-916821814):

> `sample` generates a completely new robot state, writing it into the passed-in data-structure, 
> whereas `project` actually takes the passed-in state and makes it compliant with the constraints.

However, the code speaks a different language. Here, both `sample()` and `project()` sample a new state. The only difference being that `project()` takes the passed-in state as a seed for IK search, while `sample()` uses a new random state.

More precisely, `JointConstraintSampler` doesn't distinguish at all:
https://github.com/ros-planning/moveit/blob/dd0fb2a3a91f2dad7275c70efd52754a01517116/moveit_core/constraint_samplers/src/default_constraint_samplers.cpp#L194-L197

IKConstraintSampler [calls `sampleHelper()`](https://github.com/ros-planning/moveit/blob/dd0fb2a3a91f2dad7275c70efd52754a01517116/moveit_core/constraint_samplers/src/default_constraint_samplers.cpp#L637) with arguments `project=false|true`, which is just passed to `callIK()`:
https://github.com/ros-planning/moveit/blob/dd0fb2a3a91f2dad7275c70efd52754a01517116/moveit_core/constraint_samplers/src/default_constraint_samplers.cpp#L629
and used to sample a new seed state for the first attempt of `sample()`:
https://github.com/ros-planning/moveit/blob/dd0fb2a3a91f2dad7275c70efd52754a01517116/moveit_core/constraint_samplers/src/default_constraint_samplers.cpp#L657-L661

Thus, the actual difference between `sample()` and `project()` is that the former also samples a new joint state as seed. However, both actually sample a new state (in constraint space).
Given this interpretation, I agree with @v4hn that `UnionConstraintSampler::sample()` should actually call `project()` - even for the first sample as the JMG was randomized before.
Also, #2872 should be reverted as we do want to use the current state as a seed.
I will push corresponding commits.

v4hn:
> Also, #2872 should be reverted as we do want to use the current state as a seed.

you lost me there. why would you want to project the working sample in a method called "sample"?

also calling sample for the first sampler instead of project leaves the initial distribution of samples with the first sampler instead of aways assuming uniform sampling. I don't see a reason to throw away that flexibility for custom samplers .

rhaschke:
> > Also, #2872 should be reverted as we do want to use the current state as a seed.
> 
> Why would you want to project the working sample in a method called "sample"?

As I pointed out, `project()` also performs sampling (in all MoveIt constraint samplers). The only difference between `project` and `sample` is within the `IKConstraintSampler`, where `sample()` - in contrast to `project()` - tries to reach the _randomly sampled_ Cartesian pose from a _random_ joint config instead from the _current_ one.

> Calling `sample` for the first sampler instead of `project` leaves the initial distribution of samples with the first sampler.

Again, `project` still can (and obviously should) sample in the constraint space according to its custom distribution.

rhaschke:
> Werner pointed out that the seed state passed to the `project` method never changes.

If _his_ custom `project()` method didn't sample at all but just returned the same (valid start) state, this state will be passed all the time. However, this doesn't mean that the passed state never changes. It will actually be the latest "projected" state (which isn't ideal as well).

Your assumption/claim that `sample()` and `project()` aren't used independently in MoveIt is wrong:
- `ompl_interface::ConstrainedSampler` uses [`sample()`](https://github.com/ubi-agni/moveit/blob/81a4ff9dd627202af175bf74abcd28cde3723b35/moveit_planners/ompl/ompl_interface/src/detail/constrained_sampler.cpp#L68)
- `ompl_interface::ConstrainedGoalSampler` used [`project()`](https://github.com/ubi-agni/moveit/blob/fd64e84dc6267da44c51312f8d84faf810838a3a/moveit_planners/ompl/ompl_interface/src/detail/constrained_goal_sampler.cpp#L134) (this was changed to `sample()` in #2872)
- `ompl_interface::ValidConstrainedSampler` implements both, [`sample()`](https://github.com/ubi-agni/moveit/blob/89a6752133177557fd4419d9f784d0f9c737c5c2/moveit_planners/ompl/ompl_interface/src/detail/constrained_valid_state_sampler.cpp#L85) and [`project()`](https://github.com/ubi-agni/moveit/blob/89a6752133177557fd4419d9f784d0f9c737c5c2/moveit_planners/ompl/ompl_interface/src/detail/constrained_valid_state_sampler.cpp#L68), but isn't used in MoveIt

> Following that argument we should actually remove `project` and remove the random initialization from our `sample` calls...

I am open to rename `project()` to `sample()` and only use this former variant everywhere. `project` is a completely misleading name. As you pointed out, we should always bias towards the current/initial state, which is what `project()` does.
However this would change public plugin API... Do you really want to change this in ROS1?

v4hn:
thanks a lot for the test! I still want that suite integrated with CI :grin:
from your plots it seems either the hypothesis (that using `project` in the goal sampler is a useful bias for close-by targets to get a better ik result) is wrong or it does not apply for your benchmark for some reason.

rhaschke:
Thanks, @captain-yoshi! The differences are rather marginal. Did you calculate std deviation as well? What about the significance of the results?

rhaschke:
That's what I expected: The differences in means are way smaller than the observed variances. Thus, the observed differences in speed are not significant and both variants essentially perform similarly.

simonschmeisser:
Do we have a conclusion here? What would be the next steps?

rhaschke:
I suggest to merge this up to commit 6964263a0b9e1bb819fb84d17acd702f9b25cb0a. Reverting #2872 should be handled in a different PR. 
@cambel, could you please revert this PR to 6964263a0b9e1bb819fb84d17acd702f9b25cb0a? Thanks.

cambel:
@rhaschke okay

captain-yoshi:
@v4hn @rhaschke Had an error in my script which did not rebuild the moveit_ompl_interface for each commit. I also changed the position goal request to a pose goal request (But I think we would have also see a difference). Now we clan clearly see the effect of projection (1.1.4 - 1.1.5) versus sample (1.1.6 - 1.1.7) when checking the time but becomes more obvious for waypoints and length metrics.

**200 trials of RRTConnect versus Pose Constraint PC1 (50 mm cartesian offset Z axis wrt. to panda_link_8) versus Moveit Version 1.1.4 through 1.1.7**
multi_rrtconnect_pc1

v4hn:
that's a beautiful example for something that could have been posted automatically from a hypothetical benchmark CI job in the original pull request. :+1:
I know it's a lot of work though.

PR #3123, opened 2022-04-28
Applied plan button fix to plan execute button

bsygo:
Rviz crashed when switching the move group while a motion was being executed through the plan and execute button.
The same issue was already fixed for the execute button (#1198) and is now also fixed for the plan and execute button.


rhaschke (approved):
Thanks!

rhaschke (approved):

v4hn:
Congratulations on getting the fix in master @bsygo . Robert was faster than me with merging after CI finished ;)

PR #3124, opened 2022-04-28
Fix bug in applying planning scene diffs that have attached collision objects

JafarAbdi:
### Description

This fixes a bug we had while executing an MTC task where having two different planning scenes diffs that add a collision object to the scene, the first commit is just a test that trigger the bug, the second commit sets `robot_state.is_diff` to true when calling `getPlanningSceneDiffMsg` which prevent [conversions.cpp#L369-L372](https://github.com/ros-planning/moveit/blob/master/moveit_core/robot_state/src/conversions.cpp#L369-L372) from deleting the aco from the robot state when applying the diffs, but now when calling [PlanningScene::setCurrentState](https://github.com/ros-planning/moveit/blob/master/moveit_core/planning_scene/src/planning_scene.cpp#L1146) while detaching an object that object is staying in the scene so the last commit fixes it by explicitly removing the aco in the case of scene diff

This work is sponsored by RE2 Robotics

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -652,8 +652,8 @@ void PlanningScene::getPlanningSceneDiffMsg(moveit_msgs::PlanningScene& scene_ms
   else
   {
     scene_msg.robot_state = moveit_msgs::RobotState();
-    scene_msg.robot_state.is_diff = true;
   }
+  scene_msg.robot_state.is_diff = true;
 
   if (acm_)
     acm_->getMessage(scene_msg.allowed_collision_matrix);
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -711,6 +711,19 @@ void PlanningScene::getPlanningSceneDiffMsg(moveit_msgs::PlanningScene& scene_ms
     if (do_omap)
       getOctomapMsg(scene_msg.world.octomap);
   }
+
+  // This fixes an issue where detaching an object from the robot state diff is causing it to stay as attached collision
+  // object once applied back to the main scene, to get around this issue we make sure to delete it from robot state
+  for (const auto& collision_object : scene_msg.world.collision_objects)
+  {
+    if (parent_->getCurrentState().hasAttachedBody(collision_object.id))
+    {
+      moveit_msgs::AttachedCollisionObject aco;
+      aco.object.id = collision_object.id;
+      aco.object.operation = moveit_msgs::CollisionObject::REMOVE;
+      scene_msg.robot_state.attached_collision_objects.push_back(aco);
+    }
+  }
 }
 
 namespace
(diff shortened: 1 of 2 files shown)


AndyZe (commented):
I've tested this now with the Panda tutorial and clearing collision scene objects. It seems to work great! But, I still think you need to check if `parent_` is a nullptr.

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -711,6 +711,18 @@ void PlanningScene::getPlanningSceneDiffMsg(moveit_msgs::PlanningScene& scene_ms
     if (do_omap)
       getOctomapMsg(scene_msg.world.octomap);
   }
+
+  // Ensure any detached collision objects get detached from the parent planning scene, too
+  for (const auto& collision_object : scene_msg.world.collision_objects)
```
simonschmeisser, line 716: What happens if the object is completely gone? Ie no longer attached but also removed from world.collision_objects?

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -712,8 +712,7 @@ void PlanningScene::getPlanningSceneDiffMsg(moveit_msgs::PlanningScene& scene_ms
       getOctomapMsg(scene_msg.world.octomap);
   }
 
-  // This fixes an issue where detaching an object from the robot state diff is causing it to stay as attached collision
-  // object once applied back to the main scene, to get around this issue we make sure to delete it from robot state
+  // Ensure any detached collision objects get detached from the parent planning scene, too
```
JafarAbdi, line 715: Revert the change to this comment

moveit_core/planning_scene/src/planning_scene.cpp
JafarAbdi: Do you mean after getting the diff as a message? I think in that case it shouldn't change the previous behavior

moveit_core/planning_scene/test/test_planning_scene.cpp
```diff
@@ -345,6 +345,129 @@ TEST_P(CollisionDetectorTests, ClearDiff)
...
+
+// Returns collision objects names sorted alphabetically
+std::set<std::string> get_collision_objects_names(const planning_scene::PlanningScene& ps)
+{
+  std::vector<moveit_msgs::CollisionObject> collision_objects;
+  ps.getCollisionObjectMsgs(collision_objects);
+  std::set<std::string> collision_objects_names;
+  for (const auto& collision_object : collision_objects)
+    collision_objects_names.emplace(collision_object.id);
+  return collision_objects_names;
```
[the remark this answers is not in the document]
JafarAbdi: I just remembered after applying the suggestion that `collision_objects_names` is a set of the IDs of the collision objects rather than the collision objects themselves, so this wouldn't compile :D

moveit_core/planning_scene/src/planning_scene.cpp
```diff
@@ -711,6 +711,18 @@ void PlanningScene::getPlanningSceneDiffMsg(moveit_msgs::PlanningScene& scene_ms
     if (do_omap)
       getOctomapMsg(scene_msg.world.octomap);
   }
+
+  // Ensure any detached collision objects get detached from the parent planning scene, too
```
rhaschke, line 715: This comment is misleading: The parent planning scene is not (and should not get) modified.

rhaschke (commented):
Sorry for being late on this PR. I have some more suggestions and comments. Please see #3142.

moveit_core/planning_scene/src/planning_scene.cpp
JafarAbdi: Agreed, I had a to-do to update the comment, but it got resolved somehow

PR #3126, opened 2022-04-29
CI failure for master-ci due to broken vcs

rhaschke:
MoveIt GHA are broken for master-ci (running ROS Melodic), because `vcs` fails while importing the upstream workspace:
```
  === /home/runner/work/moveit/moveit/.work/upstream_ws/src/geometric_shapes (git) ===
  Could not determine ref type of version: 
  === /home/runner/work/moveit/moveit/.work/upstream_ws/src/moveit_msgs (git) ===
  Could not determine ref type of version: 
  === /home/runner/work/moveit/moveit/.work/upstream_ws/src/srdfdom (git) ===
  Could not determine ref type of version: 
```
The `--debug` output of `vcs` isn't very helpful either.
Performing exactly the same actions locally in the very same docker container (i.e. installing vcstool and importing), yields no problems! Any hints are very welcome, @dirk-thomas, @tylerjw.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,7 +31,7 @@ jobs:
       CLANG_TIDY_ARGS: --fix --fix-errors --format-style=file
       DOCKER_IMAGE: moveit/moveit:${{ matrix.env.IMAGE }}
       UPSTREAM_WORKSPACE: .github/workflows/upstream.rosinstall
-      TARGET_WORKSPACE: $TARGET_REPO_PATH github:ros-planning/moveit_resources#master
+      TARGET_WORKSPACE: $TARGET_REPO_PATH github:ros-planning/moveit_resources#heads/master
       DOWNSTREAM_WORKSPACE: .github/workflows/downstream.rosinstall
       # Fix for https://github.blog/2022-04-12-git-security-vulnerability-announced
       GIT_CEILING_DIRECTORIES: ${{ github.workspace }}/.work
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,8 @@ jobs:
       UPSTREAM_WORKSPACE: .github/workflows/upstream.rosinstall
       TARGET_WORKSPACE: $TARGET_REPO_PATH github:ros-planning/moveit_resources#master
       DOWNSTREAM_WORKSPACE: .github/workflows/downstream.rosinstall
+      # Fix for https://github.blog/2022-04-12-git-security-vulnerability-announced
+      GIT_CEILING_DIRECTORIES: ${{ github.workspace }}/.work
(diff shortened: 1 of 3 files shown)


rhaschke:
> I've seen this before when Github is acting up -- and consequently not actually sending any 'real' Git data back.

What exactly do you mean by this, @gavanderhoorn? Do you think git doesn't clone the repo when asked to do so by vcs?
I see this issue consistently on GHA over the last 2-3 weeks. The chances to see the same issue on all distributed GH servers are very small as well :wink:

gavanderhoorn:
> What exactly do you mean by this, @gavanderhoorn? Do you think git doesn't clone the repo when asked to do so by vcs? 

Github has returned corrupted/garbage data in the past.

Some tools can deal with that and present an intelligible error message.

Some don't and the error message doesn't really indicate what's actually wrong.

> I see this issue consistently on GHA over the last 2-3 weeks. The chances to see the same issue on all distributed GH servers are very small as well wink

I agree it's unlikely, but when Github changed the structure of the `.tar.gz`s it returned via its API for about 45% of the requests, we also assumed there was something wrong with the repos instead of with GH. This made `rosdep` fail in predictable, but unexpected ways.

That got 'magically' solved one day.

rhaschke:
Another observation is that the issue only arises for the `master-ci` docker image, i.e. noetic images work fine.

JafarAbdi:
I've been debugging this for a while, looks like it's related to `git ls-remote https://github.com/ros-planning/moveit_msgs.git master` not return anything in CI see https://github.com/JafarAbdi/moveit/commit/b4eb1c913184acf17fa6b94f365fefcc36dfdf6f (I can't reproduce it locally or using master-ci image, they both return `152f6f6ac1ec83e0028c806831a4fbca32008f5e        refs/heads/master`), vcs uses the output to determine the version type see https://github.com/dirk-thomas/vcstool/blob/master/vcstool/clients/git.py#L516

Prefixing the branches names with `heads/` seem to fix it, but I still don't understand why it's happening in the first place https://github.com/JafarAbdi/moveit/commit/546332db3b071a01fab410996cccefb335d990c6

rhaschke:
Thanks a lot for digging into this, @JafarAbdi. I will use heads/* as a workaround for now.

rhaschke:
I studied this issue a little more and found the culprit: 
- [With both, `BASEDIR` defined and git `safe.directory` not defined, git 2.17 (on Bionic) fails, while git 2.25 (on Focal) succeeds.](https://github.com/rhaschke/test/actions/runs/2320275539)
- [Not using `BASEDIR` resolves the problem.](https://github.com/rhaschke/test/actions/runs/2320339806)
- ... as well as [defining `git config --global --add safe.directory "*"`](https://github.com/rhaschke/test/actions/runs/2320274581)

Issue #439, opened 2017-01-31
Updating PlanningScene after clearOctomap


simonschmeisser:
### Description

We use the following code to clear the Octomap:

```
m_planningSceneMonitor->clearOctomap();

// TODO: octomap is not updated visually with planningSceneMonitor->clearOctomap().
//       This workaround fixes the problem, but often causes a crash
planning_scene_monitor::LockedPlanningSceneRW ps(m_planningSceneMonitor);
    m_planningSceneMonitor->triggerSceneUpdateEvent(planning_scene_monitor::PlanningSceneMonitor::UPDATE_SCENE);
```

but shouldn't this update be triggered internally here?
https://github.com/ros-planning/moveit/blob/kinetic-devel/moveit_ros/planning/planning_scene_monitor/src/planning_scene_monitor.cpp#L498

```
void planning_scene_monitor::PlanningSceneMonitor::clearOctomap()
{
  octomap_monitor_->getOcTreePtr()->lockWrite();
  octomap_monitor_->getOcTreePtr()->clear();
  octomap_monitor_->getOcTreePtr()->unlockWrite();
}
```

Also I'm not sure about the locking necessary

I will provide a PR once I have the time and some feedback here

### Your environment
* ROS Distro: Kinetic

v4hn:
Hi Simon,

Thanks for contributing back! Looks like you're cleaning up part of your code base at the moment? :)

`PlanningSceneMonitor::clearOctomap` was added by @hersh two years ago to forward this functionality to the move_group and provide a ROS service to call it in fad82fbf31b40789d0dfff4b4313b37793434d0a . Nobody touched it since.  I assume it was more important to get the functionality and less so to get it 100% correctly integrated.

On Tue, Jan 31, 2017 at 07:46:59AM -0800, Simon Schmeisser wrote:
> but shouldn't this update be triggered internally here?

Yes, I believe so. This is an update to the scene, so it should report the change. You're right.

> Also I'm not sure about the locking necessary

What makes you think this is not necessary? I did not look into the details here yet.

> I will provide a PR once I have the time and some feedback here

Cool! :-)

PR #3134, opened 2022-05-10
Improve handling of empty octomap messages and "clear octomap" actions

### LINKED ISSUE #3045: MotionPlanning RViz plugin have to be manually refreshed after clear_octomap
### Description

After calling the `/clear_octomap` service, the MotionPlanning RViz plugin have to be reloaded (uncheck and re-check the plugin, or remove/add), for the octomap to disappear from the Planning Scene.

The same happens if an **empty** octomap is loaded from a file via `/move_group/load_map` (the old octomap is still shown until the plugin is reloaded). The empty map is generated via `/move_group/save_map` on a fresh scene.

The same behaviour is present when calling the `/apply_planning_scene` service with an empty octomap:
(untested code, it is a minimal version of the actual code I use)
```
get_planning_scene = rospy.ServiceProxy('/get_planning_scene', GetPlanningScene)
components = PlanningSceneComponents()
components.components = sum([
    PlanningSceneComponents.WORLD_OBJECT_NAMES,
    PlanningSceneComponents.WORLD_OBJECT_GEOMETRY,
    PlanningSceneComponents.OCTOMAP
])
ps = get_planning_scene(components)
ps = ps.scene  # moveit_msgs/PlanningScene
ps.is_diff = True
ps.world.octomap = OctomapWithPose()
apply_planning_scene = rospy.ServiceProxy('/apply_planning_scene', ApplyPlanningScene)
apply_planning_scene(ps)
```

I am not sure if any of these 3 methods of clearing the octomap are related, but I thought I'd just as well report all the similar issues I've run into.

### Your environment
* ROS Distro: Noetic
* OS Version: Pop!OS
* Built from source, branch `master` (moveit repo is on commit e7b388d238159de020a9c31f5896112c1513bf6b)
* The behaviour seems to be the same on moveit 1.1.7 from apt on Pop!OS 20.04

### Steps to reproduce
- Start RViz and moveit
- Load/scan any octomap
- Call service `/clear_octomap`

### Expected behaviour
I would expect the octomap to disappear from the Planning Scene after the service is called.

### Actual behaviour
The octomap is still shown until I reload the MotionPlanning plugin.

FSund:
### Description

Currently, if a `PlanningScene` message is published on the `monitoried_planning_scene` topic with an empty `msg.world.octomap.octomap.data` field, the octomap is skipped altogether in if it is a `msg.is_diff` is True ([source](https://github.com/ros-planning/moveit/blob/master/moveit_core/planning_scene/src/planning_scene.cpp#L1292)). I think perhaps an empty `data` field is interpreted as the octomap field not being filled. But this makes it impossible to clear the octomap by sending an empty octomap message (for example via the "Clear octomap" button in the MotionPlanning panel in RViz).

This PR sets `octomap.id` to `"empty"` if the octomap is empty (this string can be anything except an empty string), and checks `octomap.id.empty()` instead of `octomap.data.empty()` to see if the octomap should be processed.

Fixes #3045.

### Checklist

### DIFF
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -1288,8 +1288,9 @@ bool PlanningScene::setPlanningSceneDiffMsg(const moveit_msgs::PlanningScene& sc
   for (const moveit_msgs::CollisionObject& collision_object : scene_msg.world.collision_objects)
     result &= processCollisionObjectMsg(collision_object);
 
-  // process octomap updates
-  processOctomapMsg(scene_msg.world.octomap);
+  // if an octomap was specified, replace the one we have with that one
+  if (!scene_msg.world.octomap.octomap.data.empty())
+    processOctomapMsg(scene_msg.world.octomap);
 
   return result;
 }
--- a/moveit_core/planning_scene/src/planning_scene.cpp
+++ b/moveit_core/planning_scene/src/planning_scene.cpp
@@ -842,6 +842,12 @@ bool PlanningScene::getOctomapMsg(octomap_msgs::OctomapWithPose& octomap) const
     ROS_ERROR_NAMED(LOGNAME, "Unexpected number of shapes in octomap collision object. Not including '%s' object",
                     OCTOMAP_NS.c_str());
   }
+  else
+  {
+    // indicate empty octomap, which is different from a not set octomap field
+    octomap.octomap.id = "empty";
+    return true;
+  }
   return false;
 }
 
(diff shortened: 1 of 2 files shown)


rhaschke (approved):
Thanks a lot for this fix and sorry that your PR was ignored for so long.
In prinicple I approve, but I suggest to set a non-empty id only if the octomap was actually cleared. Otherwise, a corresponding update will be always triggered.

rhaschke:
Fixes #3380, #3342, #3045, #439

PR #3137, opened 2022-05-16
move_group can optionally monitor dynamics

v4hn:
I noticed this was disabled when using the `TrajectoryMonitor` feature of the `MoveGroup` action which returns the actually executed trajectory.
With the additional flag the recorded trajectory would not include velocities because they were not copied internally.

### DIFF
--- a/moveit_ros/move_group/src/move_group.cpp
+++ b/moveit_ros/move_group/src/move_group.cpp
@@ -303,6 +303,11 @@ int main(int argc, char** argv)
 
     move_group::MoveGroupExe mge(moveit_cpp, default_planning_pipeline, debug);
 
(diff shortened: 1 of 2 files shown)


moveit_ros/move_group/src/move_group.cpp
```diff
@@ -299,6 +299,11 @@ int main(int argc, char** argv)
 
     move_group::MoveGroupExe mge(moveit_cpp, default_planning_pipeline, debug);
 
+    if(pnh.param<bool>("monitor_dynamics", false))
```
rhaschke, line 302: This new parameter should be documented somewhere.

rhaschke (approved):

PR #3151, opened 2022-05-26
Enable CForest and AnytimePathShortening planners

simonschmeisser:
This enables the two meta planners CForest and AnytimePathShortening from OMPL for use within MoveIt.

@mamoll @henningkayser does this actually make sense or should PathHybridization give better results and supersede these?

There are still various issues, see below:

### Checklist
- [ ] CForest fails the second time it's called with the following exception: `Exception thrown during CForest::solve: PathLengthDirectInfSampler: There must be at least 1 start and and 1 goal state when the informed sampler is created.`
- [ ] https://github.com/ompl/ompl/pull/905 to survive the exception which sometimes also occurs on first planning attempt
- [ ] Pass the number of attempts to CForest/Anytime instead of starting `attempts*num_threads` planners
- [ ] Find out which of the warnings and errors need to be fixed
- [ ] Check if it supports `CostCovergenceTerminationCondition`


henningkayser (commented):
I have no experience with CForest, but it should work. AnytimePathShortening is actually already listed at the top of this function.

simonschmeisser:
Sorry for the bad state of this PR! I cleaned it up a bit but given the remaining issues (see checklist) this should actually rather be a draft/WIP

PR #3168, opened 2022-07-04
Standardize logging with LOGNAME variable

cambel:
### Description

Updating the logging name variable `name_` to `LOGNAME` to keep consistency with other modules.

### Checklist

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
+++ b/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
@@ -301,7 +301,7 @@ private:
   void loadControllerParams();
 
   // Name of this class for logging
-  const std::string LOGNAME = "trajectory_execution_manager";
+  static const std::string LOGNAME = "trajectory_execution_manager";
 
   moveit::core::RobotModelConstPtr robot_model_;
   planning_scene_monitor::CurrentStateMonitorPtr csm_;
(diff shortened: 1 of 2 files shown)


simonschmeisser (approved):

moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
```diff
@@ -301,7 +301,7 @@ class TrajectoryExecutionManager
   void loadControllerParams();
 
   // Name of this class for logging
-  const std::string name_ = "trajectory_execution_manager";
+  const std::string LOGNAME = "trajectory_execution_manager";
```
rhaschke, line 304: An upper-case variable indicates a global or static one. Thus, this should be changed to `static` as well:
```suggestion
  static const std::string LOGNAME = "trajectory_execution_manager";
```

rhaschke (commented):
Minor nitpick. Otherwise looks good.

moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
cambel: sure

moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
```diff
@@ -301,7 +301,7 @@ class TrajectoryExecutionManager
   void loadControllerParams();
 
   // Name of this class for logging
-  const std::string name_ = "trajectory_execution_manager";
+  static const std::string LOGNAME = "trajectory_execution_manager";
```
simonschmeisser, line 304: Getting closer :laughing: Other classes define this as a non-member in the cpp file

cambel: Okay

Issue #3156, opened 2022-06-07
GSoC 2022: Simultaneous Trajectory Execution 


cambel:
Hi!

This issue is about the [GSoC project](https://summerofcode.withgoogle.com/proposals/details/txCkuiaO) that I am working on.

The main goal of this project is to complete the [Simultaneous Trajectory Execution PR](https://github.com/ros-planning/moveit/pull/2810)  started last year to the point that it can be merged with MoveIt's mainstream code.

The subgoals are:
1.  Document the logic of the new trajectory execution manager for the handling of simultaneous trajectories. Similarly, documenting the trajectory collision method implemented in the current version of the draft PR, which is an essential part of this project.
2.  Create a tutorial demonstrating the usage of the simultaneous trajectory execution with a dual-arm system.
3.  Optimize the trajectory collision method. Ideas for this will be discussed later in this issue, feel free to join the discussion.

Further details of the implementation of this project will be added to this issue. Feel free to propose or discuss ideas relevant to this project on this issue.

cambel:
Summary of changes introduced in [Draft PR](https://github.com/ros-planning/moveit/pull/2810):
- Change `SimpleActionServers` to (General) `ActionServers`: 
    - SimpleActionServers only accept one action (goal) at a time. If a new goal is received but a previous goal has not been completed, the old goal gets `preempt`, and then the new goal is processed. In most cases that means the old goal is stopped/canceled and the 'ActionClient' gets a `preempt` message as result.
    - So, this implies that the preempt callbacks are no longer used. All new goals are accepted and run simultaneously.
- The module `TrajectoryExecutionManager` now has two usages:
    - **[Standard]** `Single Trajectory` The execution manager accepts only one trajectory at a time. In a similar fashion as the SimpleActionServers, if a new trajectory is received but an old one is still in execution, the old trajectory is aborted and then the new trajectory is executed.
    - **{New}** `Multiple Trajectories` The execution manager accepts all new trajectories and stores them in a queue (`continuous_execution_queue`). Trajectories are executed simultaneously, in the order of arrival, as long as they are valid:
        - The new trajectory is collision-free with all active trajectories*
        - The controllers for the joints being actuated in the trajectory are not already been used.
-   A scheduling manager is added to the `TrajectoryExecutionManager` for `Multiple Trajectories` that defines the priority for trajectory execution:
    - In principle, trajectories are executed in order of arrival FIFO.
[...]

PR #3169, opened 2022-07-07
WIP: Simultaneous Trajectory Execution #3156

cambel:
## Main PR for GSoC2022 project Simultaneous Trajectory Execution (STE) #3156.

This PR only includes the changes to the `TrajectoryExecutionManager` (TEM) and can just be tested via `MoveItCpp`.

Usage with MoveGroup will be addressed in a separate PR.

A test environment can be found [here](https://github.com/cambel/moveit_simultaneous_motions_demo).

The module TrajectoryExecutionManager now has two usages:

- [Standard] **Single Trajectory** (`push` method): The execution manager accepts only one trajectory at a time. In a similar fashion as the SimpleActionServers, if a new trajectory is received but an old one is still in execution, the old trajectory is aborted and then the new trajectory is executed.

- [New] **Multiple Trajectories** (`pushAndExecuteSimultaneous` method): The execution manager accepts all new trajectories and stores them in a queue (continuous_execution_queue). Trajectories are executed simultaneously, in the order of arrival, as long as they are valid:
    - The new trajectory is collision-free* with all active trajectories
    - The controllers for the joints being actuated in the trajectory are not already been used.

* All new trajectories are checked before execution within the TEM to guarantee that they are collision-free with any active trajectories or the current planning scene.

A scheduling manager is added to the TrajectoryExecutionManager for Multiple Trajectories that defines the priority for trajectory execution:

- In principle, trajectories are executed in order of arrival FIFO.
- Trajectories that are not immediately valid are stored in a backlog queue.
- Trajectories from the backlog takes priority over any new trajectory from the continuous_execution_queue
- Trajectories from the backlog can expire if they are not valid for execution after some period of time (this should be a dynamically configurable parameter)
- The scheduling manager checks trajectories from the backlog and the continuous_execution_queue at a fixed rate (this should be a dynamically configurable parameter)

### Checklist

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/cfg/TrajectoryExecutionDynamicReconfigure.cfg
+++ b/moveit_ros/planning/trajectory_execution_manager/cfg/TrajectoryExecutionDynamicReconfigure.cfg
@@ -9,6 +9,6 @@ gen.add("allowed_goal_duration_margin", double_t, 3, "Allow more than the expect
 gen.add("execution_velocity_scaling", double_t, 4, "Multiplicative factor for execution speed", 1, 0.1, 10)
 gen.add("allowed_start_tolerance", double_t, 5, "Allowed joint-value tolerance for validation of trajectory's start point against current robot state", 0.01, 0);
 gen.add("wait_for_trajectory_completion", bool_t, 6, "Wait for trajectory completion. If set to false, do not wait for controllers to converge to last way point, before reporting success.", True)
-gen.add("continuous_execution_thread_rate", int_t, 7, "Rate at which new/pending trajectories are processed on the continuous execution threat.", 10, 1, 1000)
+gen.add("continuous_execution_thread_rate", int_t, 7, "Rate at which new/pending trajectories are processed on the continuous execution thread.", 10, 1, 1000)
 
 exit(gen.generate(PACKAGE, PACKAGE, "TrajectoryExecutionDynamicReconfigure"))
--- a/moveit_ros/planning/trajectory_execution_manager/cfg/TrajectoryExecutionDynamicReconfigure.cfg
+++ b/moveit_ros/planning/trajectory_execution_manager/cfg/TrajectoryExecutionDynamicReconfigure.cfg
@@ -9,5 +9,6 @@ gen.add("allowed_goal_duration_margin", double_t, 3, "Allow more than the expect
 gen.add("execution_velocity_scaling", double_t, 4, "Multiplicative factor for execution speed", 1, 0.1, 10)
 gen.add("allowed_start_tolerance", double_t, 5, "Allowed joint-value tolerance for validation of trajectory's start point against current robot state", 0.01, 0);
 gen.add("wait_for_trajectory_completion", bool_t, 6, "Wait for trajectory completion. If set to false, do not wait for controllers to converge to last way point, before reporting success.", True)
+gen.add("continuous_execution_thread_rate", int_t, 7, "Rate at which new/pending trajectories are processed on the continuous execution threat.", 10, 1, 1000)
 
 exit(gen.generate(PACKAGE, PACKAGE, "TrajectoryExecutionDynamicReconfigure"))
--- a/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
+++ b/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
@@ -41,6 +41,7 @@
 #include <moveit/robot_model/link_model.h>
 #include <moveit/robot_state/robot_state.h>
 #include <moveit/robot_state/attached_body.h>
+#include <moveit/planning_scene_monitor/planning_scene_monitor.h>
 #include <moveit/planning_scene_monitor/current_state_monitor.h>
 #include <moveit/planning_scene/planning_scene.h>
 #include <moveit_msgs/CollisionObject.h>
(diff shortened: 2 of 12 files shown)


moveit_ros/planning/trajectory_execution_manager/cfg/TrajectoryExecutionDynamicReconfigure.cfg
```diff
@@ -9,5 +9,6 @@ gen.add("allowed_goal_duration_margin", double_t, 3, "Allow more than the expect
 gen.add("execution_velocity_scaling", double_t, 4, "Multiplicative factor for execution speed", 1, 0.1, 10)
 gen.add("allowed_start_tolerance", double_t, 5, "Allowed joint-value tolerance for validation of trajectory's start point against current robot state", 0.01, 0);
 gen.add("wait_for_trajectory_completion", bool_t, 6, "Wait for trajectory completion. If set to false, do not wait for controllers to converge to last way point, before reporting success.", True)
+gen.add("continuous_execution_thread_rate", int_t, 7, "Rate at which new/pending trajectories are processed on the continuous execution threat.", 10, 1, 1000)
```
simonschmeisser, line 12:
```suggestion
gen.add("continuous_execution_thread_rate", int_t, 7, "Rate at which new/pending trajectories are processed on the continuous execution thread.", 10, 1, 1000)
```

moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
```diff
@@ -242,6 +282,9 @@ class TrajectoryExecutionManager
   /// Enable or disable waiting for trajectory completion
   void setWaitForTrajectoryCompletion(bool flag);
 
+  /// Rate at which new/pending trajectories are processed on the continuous execution threat
+  void setcontinuousExecutionThreadRate(int rate);
```
simonschmeisser, line 286:
```suggestion
  void setContinuousExecutionThreadRate(int rate);
```

moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
```diff
@@ -167,48 +184,71 @@ class TrajectoryExecutionManager
 
   /// Add a trajectory for immediate execution. Optionally specify a controller to use for the trajectory. If no
   /// controller is specified, a default is used. This call is non-blocking.
-  bool pushAndExecute(const moveit_msgs::RobotTrajectory& trajectory, const std::string& controller = "");
+  /// Optionally specify an expiration time for the trajectory, from reception to execution (default 60 secs).
+  /// Optionally specify callback that is called when the execution of the trajectory is completed.
+  bool pushAndExecuteSimultaneous(const moveit_msgs::RobotTrajectory& trajectory, const std::string& controller = "",
```
simonschmeisser, line 189:
```suggestion
  bool pushAndExecuteSimultaneously(const moveit_msgs::RobotTrajectory& trajectory, const std::string& controller = "",
```
any native speaker around to comment?

moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
cambel: Fixed

cambel:
Changes suggested in #2810 

- [x] 1. rate r(10): document and make dynamic configurable
- [x] 2. `TrajectoryExecutionContext` Can we change this to be a unique_ptr or shared_ptr?
- [x] 3. The timeout could be part of the TrajectoryExecutionContext struct, or be passed with the corresponding push functions
- [x] 4. Debug messages: time_from_start is used a couple of times in this block.    Using ```const auto& time_from_start = current_context->trajectory_parts_[0].joint_trajectory.points.back().time_from_start```would make this function more readable imo
- [ ] 5. Related to (4): Is this really a useful debug comment like this? It would help if the blocked resources (group, controller) were referenced here.
- [ ] 6. `continuousExecutionThread` Please extend this comment to describe how the backlogs are being processed and iterated in this loop. 
- [ ] 7. `continuousExecutionThread` This loop seems a little too convoluted to me. I'll take some more time to review this more in depth and see if this could be simplified somehow.

simonschmeisser (commented):
You pushed some changes in the mean time so I'll just press submit now and continue reviewing with the current state

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -286,54 +301,62 @@ bool TrajectoryExecutionManager::push(const moveit_msgs::RobotTrajectory& trajec
...
+    context.reset();
     last_execution_status_ = moveit_controller_manager::ExecutionStatus::ABORTED;
   }
 
   return false;
 }
 
-bool TrajectoryExecutionManager::pushAndExecute(const moveit_msgs::RobotTrajectory& trajectory,
-                                                const std::string& controller)
+bool TrajectoryExecutionManager::pushAndExecuteSimultaneously(const moveit_msgs::RobotTrajectory& trajectory,
```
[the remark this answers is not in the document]
cambel: I think nobody is using `pushAndExecute`, as far as I know, the current version is broken. So, maybe it would be a good opportunity to clean up some or most of the overload methods that may not be used in the new API.

PR #3175, opened 2022-07-14
[moveit_setup_assistant] Fixes build error: "cannot convert ‘const YAML::Node’ to ‘boo ’ in initialization"

AM4283:
### Description
I was having a build error on my host os: Arch Linux x86_64 when trying to build the moveit setup assistant, and submitted to the below repo to fix an issue
https://github.com/ros-noetic-arch/ros-noetic-moveit-setup-assistant/issues/3  (build log is also present on this issue)
Originally, I made a patch to fix this build issue in trying to build this component as part of an AUR package, and I was requested to further submit a pull request here.

### DIFF
--- a/moveit_setup_assistant/src/tools/moveit_config_data.cpp
+++ b/moveit_setup_assistant/src/tools/moveit_config_data.cpp
@@ -1256,7 +1256,7 @@ template <typename T>
 bool parse(const YAML::Node& node, const std::string& key, T& storage, const T& default_value = T())
 {
   const YAML::Node& n = node[key];
-  bool valid = n;
+  bool valid = n.as<bool>();
   storage = valid ? n.as<T>() : default_value;
   return valid;
 }


rhaschke (approved):
Looks like a trivial and reasonable change. Thanks!

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -1256,7 +1256,7 @@ template <typename T>
 bool parse(const YAML::Node& node, const std::string& key, T& storage, const T& default_value = T())
 {
   const YAML::Node& n = node[key];
-  bool valid = n;
+  bool valid = n.as<bool>();
```
simonschmeisser, line 1259: doesn't that break for a node containing `false` and a `default_value` of `true`

simonschmeisser (commented):
doesn't this change behavior?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
simonschmeisser:
```suggestion
  bool valid = n.isDefined();
```
but according to here https://yaml-cpp.docsforge.com/master/api/YAML/Node/operator%20bool/  operator bool exists. Why does it not compile on arch?

simonschmeisser: And why does it say `boo` instead of `bool`?

rhaschke: You are right! `IsDefined()` implements the correct semantics here. `as<bool>()` would attempt to read a boolean value.

> [`operator bool()`](https://yaml-cpp.docsforge.com/master/api/YAML/Node/operator%20bool/) exists. Why does it not compile on arch?
I fixed the commit.

The conversion operator is declared `explicit`. This wasn't the case in older versions used in Ubuntu.

PR #3197, opened 2022-08-28
Improve Cartesian interpolation

rhaschke:
So far, Cartesian interpolation only considered _link frames_ to move along a straight-line. This PR augments this to arbitrary frames by considering a configurable offset to the link frame. Thus, now the frame resulting from link * offset should move along a straight line towards the target. Obviously, this is only relevant for rotational motions - a fixed offset for linear motions doesn't change the resulting motion. For example, the new implementation [allows for circular motions around a non-link origin](https://github.com/ros-planning/moveit_task_constructor/pull/380).

### DIFF
--- a/moveit_core/robot_state/CMakeLists.txt
+++ b/moveit_core/robot_state/CMakeLists.txt
@@ -20,6 +20,8 @@ install(DIRECTORY include/ DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION})
 
 # Unit tests
 if(CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
+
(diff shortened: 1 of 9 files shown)


simonschmeisser (approved):
Looks good but I would appreciate a test for the link_offset parameter

rhaschke:
Sure. I will add a short test...

rhaschke:
Eventually, I added some tests, also for the original `computeCartesianPath()` functionality.

rhaschke:
Argh! I was using gtest 1.10 features not yet available in Bionic/Melodic.

simonschmeisser:
Thanks a lot for improving the tests as well!

PR #3211, opened 2022-09-09
version.h: automatically bump patch number for devel builds

rhaschke:
Follow up on #2793:
Automatically recognize a devel build by comparing package.xml's version with available git tags. If they don't match, assume a devel version, increase the patch number and append `-devel`, e.g. yielding 1.1.10-devel instead of 1.1.9.
Note that this adapted version number is only reported in `version.h` and thus via `moveit_version`.
Libraries still get the old version number listed in `package.xml`. This was a deliberative decision: During development, it is convenient to (re)build individual libs/packages only. These should still have the same library version number to be found by existing downstream packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ The [MoveIt Motion Planning Framework for ROS](http://moveit.ros.org). For the R
 - We develop latest features on `master`.
 - The `*-devel` branches correspond to released and stable versions of MoveIt for specific distributions of ROS. `noetic-devel` is synced to `master` currently.
 - Bug fixes occasionally get backported to these released versions of MoveIt.
+- To facilitate compile-time switching, the patch version of `MOVEIT_VERSION` of a development branch will be incremented by 1 w.r.t. the package.xml's version number.
 
 ## MoveIt Status
 
--- a/moveit_core/version/CMakeLists.txt
+++ b/moveit_core/version/CMakeLists.txt
@@ -1,18 +1,11 @@
 # Generate and install version.h
 
-if(NOT "$ENV{USER}" STREQUAL "buildfarm") # Don't define version postfix on buildfarm
-	set(MOVEIT_VERSION_EXTRA "devel" CACHE STRING "version string postfix")
-endif()
-
-message(STATUS " *** Building MoveIt ${${PROJECT_NAME}_VERSION} ${MOVEIT_VERSION_EXTRA} ***")
-
 # https://stackoverflow.com/questions/13920072/how-to-always-run-command-when-building-regardless-of-any-dependency
 add_custom_command(
 	OUTPUT ${VERSION_FILE_PATH}/moveit/version.h always_rebuild
 	COMMAND ${CMAKE_COMMAND}
 	-DVERSION_FILE_PATH="${VERSION_FILE_PATH}"
 	-DMOVEIT_VERSION="${${PROJECT_NAME}_VERSION}"
-	-DMOVEIT_VERSION_EXTRA="${MOVEIT_VERSION_EXTRA}"
 	-P ${CMAKE_CURRENT_SOURCE_DIR}/version.cmake
 	WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
 )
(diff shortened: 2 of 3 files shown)


AndyZe:
I don't feel entirely qualified to review this. Is there a way to see that it worked?

rhaschke:
> Is there a way to see that it worked? Is there any downside if it does not work as intended?

Just run cmake on different commits. [It will print the corresponding version number](https://github.com/ros-planning/moveit/pull/3211/files#diff-4b71e6fe5afda88a9396dce33a788284d22b2326e5d13428b000b119f79438a5R55).

AndyZe:
OK, I'm testing like this:

> `cd ../moveit/moveit_core`
> `cmake . | grep "Building MoveIt"`   --> no output
> Change the version tag in ./package.xml to something new:  `<version>1.1.27</version>`
> `cmake . | grep "Building MoveIt"`   --> still no output

Sorry, I must be doing something wrong.

rhaschke:
The output message is part of the `make` log (not `cmake`):
```
> grep -r "Building MoveIt" logs
logs/moveit_core/build.make.log:--  *** Building MoveIt 1.1.10-devel ***
```
or `catkin b --no-deps moveit_core -v | grep "Building MoveIt"`

The devel prefix (and auto-increment of patch number) will only be omitted if the current git commit will have the same version tag as stated in package.xml. Thus changing the version in package.xml will not illustrate the behavior. You need to declare a version tag for your git commit.

AndyZe:
Thanks Robert. I can at least see _some_ output now. I'm not sure it's the intended output:

- This branch without other changes:  `*** Building MoveIt 1.1.10-devel ***`
- Add a couple test commits but no tags:  `*** Building MoveIt 1.1.10-devel ***`  (still, shouldn't I see a change?)
- Add a test commit with tag 1.1.11: `*** Building MoveIt 1.1.10-devel ***`  (stil, shouldn't I see a change?)

AndyZe:
Actually I think that^ output makes sense. Please confirm, then I'm good to merge.

rhaschke:
Yes, that's the intended behavior. 
The key question to other maintainers was whether this behavior is considered reasonable, particularly @simonschmeisser and @v4hn.

AndyZe (approved):

rhaschke:
Note that the bump of the version number is only visible in `version.h`, i.e. to source code using that include.
The library SO versions are stuck to the "old" version number given in `package.xml`.

rhaschke:
@tylerjw, I'm also very curious about your opinion on this PR.

v4hn (approved):
Sorry to leave this without review for so long.

Years ago, I already proposed to increase the patch version right after each release in order to get similar behavior, but the release team back then sadly insisted it complicates releases further and adds confusion.. But I definitely still approve the idea still.
The argument to keep the old version in the package.xml due to the soname coupling makes sense to me, though it adds more structural complexity (bike-shedding here).

Two requirements from my side to merge this PR:
- Please document the behavior somewhere (maybe [in the README below the branches policy](https://github.com/ros-planning/moveit#branches-policy)?)
- The previously merged and the proposed patch both fall short in case the sources are not built in a git repository. That might sound nonsensical, but typically [release tarballs](https://github.com/ros-planning/moveit/releases/tag/1.1.5) do not include a git structure and you would still expect them to report the correct version after building
  - In this case we should assume an unmodified release source. If someone (e.g., external package maintainers) patched them, they will modify the version number themselves
  - It might be nicer to report a git commit hash/name "none" or similar instead of the empty string.

I believe with Andy and my positive review this can be merged after these points are addressed. Thank you for pushing this @rhaschke.

v4hn (approved):

simonschmeisser:
Maybe it's a red herring but I see a lot more rebuilds of dependent packages in my catkin workspace since this change (I assume, haven't verified).
It seems the `version.h` is touched on every `catkin build` now? Can we avoid this somehow?

rhaschke:
You are right: `version.h` is regenerated on every catkin build. However, cmake's `configure_file` should take care of not changing the file's timestamp if the content didn't change:
https://stackoverflow.com/questions/71776252/cmake-configure-file-does-not-touch-if-the-file-is-the-same

Could you please collect more evidence that this is not the case?

Issue #3118, opened 2022-04-25
The issue of moveit_setup_assistant gui in defining the robot pose- joint label, joint value and slide bar is too small


darrenjiayinxie:
### Description

Using the setup assistant for a robot arm, and when defining the robot pose, the joint label, joint value and slide bar become extremely small.  I fixed the bug by myself. Hope it would help others that facing the same issue.

### Your environment
* ROS Distro: Noetic
* OS Version: Ubuntu 20.04
* Source build - master branch 

### Steps to reproduce
Using the setup assistant for a robot arm, and when defining the robot pose, the joint label, joint value and slide bar become extremely small. 
 
### Expected behaviour
![fixed](https://user-images.githubusercontent.com/16159311/165002381-e51835c2-3a9e-41bf-9da0-174e421e7d87.png)

### Actual behaviour
The joint label, joint value and slide bar become extremely small. 
![bug](https://user-images.githubusercontent.com/16159311/165002271-fb5882cd-6b2f-485a-bd40-ce239389a2a7.png)

### Steps to fix the bug
Modify the source code '/moveit/moveit_setup_assistant/src/widgets/robot_poses_widget.cpp':
modify line 525 and 831
![code_1](https://user-images.githubusercontent.com/16159311/165002580-2ee5113b-4d52-46d0-8e6d-ceb044168004.png)

![code_2](https://user-images.githubusercontent.com/16159311/165002586-2966dd1a-786c-472b-941b-ec21563a0978.png)

rhaschke:
I cannot confirm this issue. Do you use a high-resolution screen?

darrenjiayinxie:
> That's what I expected. The actual problem is that the window sizes are absolute. They should be computed relative e.g. to the font size or to the main window.

I agree. I think this should be fixed.

PR #3218, opened 2022-09-15
issue #3118

ManuelGonzalezL:
### The issue #3118 was that the joint value and slide bar were too small when using a 4k display. I implemented a solution that worked before in my personal projects, but, for the nature of the issue, testing it was complicated.


rhaschke:
Thanks for this contribution. However, you just replaced the existing hard-coded values with another set for high-res screens.
Originally, I suggested to _automatically_ compute the size relative to the font or window size.
I implemented a corresponding fix in #3219. Closing here.

PR #3223, opened 2022-09-25
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,6 +10,9 @@ on:
     branches:
       - master
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   default:
     strategy:
--- a/.github/workflows/format.yaml
+++ b/.github/workflows/format.yaml
@@ -8,6 +8,9 @@ on:
   pull_request:
   push:
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   pre-commit:
     name: pre-commit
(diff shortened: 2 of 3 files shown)


JafarAbdi (approved):

rhaschke:
Unfortunately, [this breaks syncing of `master` and `noetic-devel` branch](https://github.com/ros-planning/moveit/actions/runs/3198531639/jobs/5223787861#step:3:19) (which requires push permissions).
@sashashura, could you please suggest, which scope needs to have write permission to fix that?

rhaschke:
> @sashashura, could you please suggest, which scope needs to have write permission to fix that?

Looks like [this commit](https://github.com/ros-planning/moveit/commit/4c10a9e47a57dac915eb2b6e6ae26f445f2f884f) already [fixed it](https://github.com/ros-planning/moveit/actions/runs/3203007786/jobs/5232990610).

rhaschke:
Looks like write permission to the repository is needed anyways.

PR #3228, opened 2022-09-30
Drop shortcut avoiding planning when all constraints are already met

### LINKED ISSUE #3222: Python interface not planning to attached collision object subframe
### Description

When trying to plan with an attached collision object’s subframe set as the end effector link the following error is thrown: 
ERROR: Link 'end_effector/tcp' not found in model 'fanuc'

### Environment
* ROS Distro: Noetic 
* OS Version: Ubuntu 20.04
* Binary build
* latest moveit version

### Steps to reproduce
I have created a minimal example in following [repo](https://github.com/vinceHuyghe/example) this has been tested with [moveit_resources](https://github.com/ros-planning/moveit_resources) fanuc_moveit_config demo.launch and reproduces the error.

### Actual behaviour
ERROR: Link 'end_effector/tcp' not found in model 'fanuc'

rhaschke:
When calling the PlanAndExcute action, there was a shortcut avoiding planning (and execution) at all when all goal constraints were already met. However, this check also succeeded when there was an error, e.g. resolving link names. As proper link resolution might require some planning request adapters to be executed, e.g. ResolveConstraintFrames, we should not skip planning.

Note, calling the PlanOnly action, didn't have this shortcut as well, resulting in inconsistent behavior.

I just decided to remove the shortcut check completely.

Fixes #3222

### DIFF
--- a/moveit_ros/move_group/src/default_capabilities/move_action_capability.cpp
+++ b/moveit_ros/move_group/src/default_capabilities/move_action_capability.cpp
@@ -103,23 +103,6 @@ void MoveGroupMoveAction::executeMoveCallbackPlanAndExecute(const moveit_msgs::M
   ROS_INFO_NAMED(getName(), "Combined planning and execution request received for MoveGroup action. "
                             "Forwarding to planning and execution pipeline.");
 
-  if (moveit::core::isEmpty(goal->planning_options.planning_scene_diff))
-  {
-    planning_scene_monitor::LockedPlanningSceneRO lscene(context_->planning_scene_monitor_);
-    const moveit::core::RobotState& current_state = lscene->getCurrentState();
-
-    // check to see if the desired constraints are already met
-    for (std::size_t i = 0; i < goal->request.goal_constraints.size(); ++i)
-      if (lscene->isStateConstrained(current_state,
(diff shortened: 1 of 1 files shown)


moveit_ros/move_group/src/default_capabilities/move_action_capability.cpp
```diff
@@ -107,17 +107,6 @@ void MoveGroupMoveAction::executeMoveCallbackPlanAndExecute(const moveit_msgs::M
   {
     planning_scene_monitor::LockedPlanningSceneRO lscene(context_->planning_scene_monitor_);
     const moveit::core::RobotState& current_state = lscene->getCurrentState();
-
-    // check to see if the desired constraints are already met
-    for (std::size_t i = 0; i < goal->request.goal_constraints.size(); ++i)
-      if (lscene->isStateConstrained(current_state,
```
JafarAbdi, line 113: I think this's a bug anyway since it looks for the first constraint that is satisfied, rather than all of them, right?

JafarAbdi (approved):

moveit_ros/move_group/src/default_capabilities/move_action_capability.cpp
rhaschke: Yes, you are right! I didn't check this...

PR #3243, opened 2022-10-17
Simultaneous trajectory execution

cambel:
# Main PR for GSoC2022 project Simultaneous Trajectory Execution (STE) #3156.

With this PR, MoveIt now allows simultaneous execution of trajectories, as long as, each trajectory uses a different set of controllers. For example, in a dual arm environment, each arm can execute a different set of trajectories without needing to wait for the other arm to finish moving or manually synchronizing the motion of both arm into a single trajectory. Optionally, a collision check is performed right before execution of new trajectories to prevent collisions with active trajectories.

## Example use cases:
1. Several trajectories being planned and executed through the `MoveIt Motion Planning Rviz` plugin. 
simultaneous-execution-rviz
2. Trajectories being planned and executed from different scripts (Rviz + Python script)
simultaneous-execution-python-rviz
3. Real application examples can be found here https://github.com/ros-planning/moveit/pull/2810

## Feature description
The simultaneous trajectory execution feature can be enabled or disabled through the dynamic reconfigure parameter `/move_group/trajectory_execution/enable_simultaneous_execution`.
Optionally, the extra layer of collision checking, done right before execution of trajectories, can be enabled through the dynamic reconfigure parameter `/move_group/trajectory_execution/enable_collision_checking`.

### The TrajectoryExecutionManager API remains the same for execution in **blocking-mode**:
- New trajectories can be pushed through `TrajectoryExecutionManager::push()` .
- Then all pushed trajectory can be executed through `TrajectoryExecutionManager::execute()`.
- All controller's execution can be stopped through `TrajectoryExecutionManager::stopExecution()`.

### For execution in *simultaneous-mode*:
- New trajectories pushed through `TrajectoryExecutionManager::push()`  will be automatically executed. If the trajectory is pushed and executed successfully, a TrajectoryID will be returned.
- The `TrajectoryExecutionManager::execute()` method will do nothing.
- A set of trajectories that need to be executed in a strict sequential order can be pushed to the overload method `TrajectoryExecutionManager::push(vector<Trajectory>)`.
- The execution of a particular trajectory can be stopped through `TrajectoryExecutionManager::stopExecution(TrajectoryID)`.

## Changes to the API
- The `TrajectoryExecutionManager::push()` method optionally accepts a callback that is called when the trajectory execution is completed or aborted.
- The `TrajectoryExecutionManager::push(vector<Trajectory>)` method optionally accepts an additional callback that is called after each trajectory "segment" is completed successfully.

## MoveItCPP 
In **simultaneous-mode**, when multiple trajectories are executed in non-blocking mode `MoveItCpp::execute(group_name, robot_trajectory, blocking=false)`, each one can be executed simultaneously, if they are valid.

## Move Group Action Servers
To allow simultaneous execution of trajectories through the `move group interfaces` the `SimpleActionServers` `MoveActionCapabilityServer` and `ExecuteTrajectoryActionServer` were change to `ActionServers`. **Note:** The `SimpleActionClient` seem to work fine with the `ActionServer`.

### Checklist


Amin173:
Hi,
I am really looking forward to seeing this feature in MoveiIt! I was trying the code from this pull request for some simulations and I realized the `avoid_collision` argument in the [`compute_cartesian_path()`](http://docs.ros.org/en/jade/api/moveit_commander/html/classmoveit__commander_1_1move__group_1_1MoveGroupCommander.html#ad7f6d93d73bf43268ba983afb0dc4f23) does not work. Basically, it is always `True`, even if I specify it to be `False`; so the motion planning will fail if their is a collision in the trajectory even if  I give `avoid_collision=False`. This is not the case when I use the moveit [`master`](https://github.com/ros-planning/moveit) branch.

cambel:
Hi @Amin173 

Thank you for checking this new feature.

This PR introduces a new layer of collision checking right before the execution of a trajectory which is set ON/OFF based on the Dynamic Reconfigure parameter `trajectory_execution/enable_collision_checking`. This parameter is set to `true` by default. This parameter is independent of the collision checking done during motion planning, so setting `avoid_collision=False`  won't disable it.

azalutsky:
I plan on making an attempt at porting this to ROS2 for my own selfish needs. Stay tuned. Thanks for your awesome hard work on this @cambel, will email you if I run into issues.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -922,63 +1196,45 @@ moveit_controller_manager::ExecutionStatus TrajectoryExecutionManager::executeAn
...
-    catch (std::exception& ex)
-    {
-      ROS_ERROR_NAMED(LOGNAME, "Caught %s when canceling execution.", ex.what());
-    }
-}
-
 void TrajectoryExecutionManager::stopExecution(bool auto_clear)
 {
+  // We mark stop_execution_ as true and wait for the runEventManager threat to cancel all active trajectories
+  ROS_INFO_NAMED(LOGNAME, "StopExecution()");
```
azalutsky, line 1202: Looks more like a debug message.

azalutsky (commented):
minor finds.

moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -238,14 +238,54 @@ bool MoveItCpp::execute(const std::string& group_name, const robot_trajectory::R
...
+
+  if (!trajectory_execution_manager_->push(robot_trajectory_msg, "", callback))
+    return false;
+
+  {
+    std::unique_lock<std::mutex> ulock(execution_complete_mutex_);
+    active_trajectories_.push_back(trajectory_id);
+  }
+
   if (blocking)
```
AndyZe, line 271: @cambel there is no case for `blocking==false`, right? We were thinking about removing this blocking flag completely in MoveIt2.

cambel: Hi @AndyZe
For this new feature, `blocking==false` is used to send multiple trajectories to be simultaneously executed without waiting for each one to finish before executing the next one, as in this demo: 

https://github.com/cambel/moveit_simultaneous_motions_demo/blob/2c9b7f6077e59cea73b544107b7bda7e6fbb8ca1/panda_dual_arm_demo/src/simple_path_moveitcpp.cpp#L168-L170

AndyZe: OK, thanks for the reply.

Shouldn't there be an `else()` here? Like this:

if (blocking)
{
  ...
}
else
{
  // Still need to execute the trajectory somehow, right?
}

cambel: Thank you, I was missing the case where simultaneous execution mode is OFF and `blocking==false`.

mcres:
Dear all,

Are there any plans to merge this PR? This feature being missing limits using MoveIt in any setup with 2+, independent robots. Those are _a lot_ of cases, and I think this would benefit a lot of users.

Also, can we assume that any feature developed for `moveit` will be ported to `moveit2` as well?

Marwan-Refaat:
Hey everyone!

Seconding @mcres here, this seems to be quite a popular PR and would benefit a lot of users. 

I am looking to port this functionality to MoveIt2 and was wondering why this PR never got merged?  Does the current implementation have any limitations that prevent a current merge into `moveit` or  a future port into `moveit2`? 👀

PR #3257, opened 2022-10-31
Convert OMPL status to MoveItErrorCode in the OMPL interface

AndyZe:
This is a backport of this MoveIt2 PR:  https://github.com/ros-planning/moveit2/pull/1606

### DIFF
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/model_based_planning_context.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/model_based_planning_context.h
@@ -372,7 +372,7 @@ protected:
   void unregisterTerminationCondition();
 
   /** \brief Convert OMPL PlannerStatus to moveit_msgs::msg::MoveItErrorCode */
-  int32_t logPlannerStatus(og::SimpleSetupPtr ompl_simple_setup);
+  int32_t errorCode(const ompl::base::PlannerStatus& status);
 
   ModelBasedPlanningContextSpecification spec_;
 
--- a/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/model_based_planning_context.h
+++ b/moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/model_based_planning_context.h
@@ -277,7 +277,7 @@ public:
      @param timeout The time to spend on solving
      @param count The number of runs to combine the paths of, in an attempt to generate better quality paths
   */
-  bool solve(double timeout, unsigned int count);
+  const moveit_msgs::MoveItErrorCodes solve(double timeout, unsigned int count);
 
   /* @brief Benchmark the planning problem. Return true on successful saving of benchmark results
      @param timeout The time to spend on solving
@@ -370,6 +370,9 @@ protected:
 
   void registerTerminationCondition(const ob::PlannerTerminationCondition& ptc);
   void unregisterTerminationCondition();
+
+  /** \brief Convert OMPL PlannerStatus to moveit_msgs::msg::MoveItErrorCode */
+  int32_t logPlannerStatus(og::SimpleSetupPtr ompl_simple_setup);
 
   ModelBasedPlanningContextSpecification spec_;
 
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -773,11 +773,9 @@ const moveit_msgs::MoveItErrorCodes ompl_interface::ModelBasedPlanningContext::s
     ROS_DEBUG_NAMED(LOGNAME, "%s: Solving the planning problem once...", name_.c_str());
     ob::PlannerTerminationCondition ptc = constructPlannerTerminationCondition(timeout, start);
     registerTerminationCondition(ptc);
-    result.val = ompl_simple_setup_->solve(ptc) == ompl::base::PlannerStatus::EXACT_SOLUTION;
+    result.val = errorCode(ompl_simple_setup_->solve(ptc));
     last_plan_time_ = ompl_simple_setup_->getLastPlanComputationTime();
     unregisterTerminationCondition();
-    // fill the result status code
-    result.val = errorCode(ompl_simple_setup_->getLastPlannerStatus());
   }
   else
   {
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -883,8 +883,8 @@ int32_t ompl_interface::ModelBasedPlanningContext::errorCode(const ompl::base::P
       result = moveit_msgs::MoveItErrorCodes::TIMED_OUT;
       break;
     case ompl::base::PlannerStatus::APPROXIMATE_SOLUTION:
-      ROS_WARN_NAMED(LOGNAME, "Solution is approximate");
-      result = moveit_msgs::MoveItErrorCodes::SUCCESS;
+      ROS_WARN_NAMED(LOGNAME, "Solution is approximate. This usually indicates a failure.");
+      result = moveit_msgs::MoveItErrorCodes::PLANNING_FAILED;
       break;
     case ompl::base::PlannerStatus::EXACT_SOLUTION:
       result = moveit_msgs::MoveItErrorCodes::SUCCESS;
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -678,14 +678,12 @@ void ompl_interface::ModelBasedPlanningContext::postSolve()
   int v = ompl_simple_setup_->getSpaceInformation()->getMotionValidator()->getValidMotionCount();
   int iv = ompl_simple_setup_->getSpaceInformation()->getMotionValidator()->getInvalidMotionCount();
   ROS_DEBUG_NAMED(LOGNAME, "There were %d valid motions and %d invalid motions.", v, iv);
-
-  if (ompl_simple_setup_->getProblemDefinition()->hasApproximateSolution())
-    ROS_WARN_NAMED(LOGNAME, "Computed solution is approximate");
 }
 
 bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::MotionPlanResponse& res)
 {
-  if (solve(request_.allowed_planning_time, request_.num_planning_attempts))
+  res.error_code_ = solve(request_.allowed_planning_time, request_.num_planning_attempts);
+  if (res.error_code_.val == moveit_msgs::MoveItErrorCodes::SUCCESS)
   {
     double ptime = getLastPlanTime();
     if (simplify_solutions_)
@@ -709,14 +707,14 @@ bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::Motion
   else
   {
     ROS_INFO_NAMED(LOGNAME, "Unable to solve the planning problem");
-    res.error_code_.val = moveit_msgs::MoveItErrorCodes::PLANNING_FAILED;
     return false;
   }
 }
 
 bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::MotionPlanDetailedResponse& res)
 {
-  if (solve(request_.allowed_planning_time, request_.num_planning_attempts))
+  moveit_msgs::MoveItErrorCodes moveit_result = solve(request_.allowed_planning_time, request_.num_planning_attempts);
+  if (moveit_result.val == moveit_msgs::MoveItErrorCodes::SUCCESS)
   {
     res.trajectory_.reserve(3);
 
@@ -750,9 +748,9 @@ bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::Motion
       getSolutionPath(*res.trajectory_.back());
     }
 
-    // fill the response
     ROS_DEBUG_NAMED(LOGNAME, "%s: Returning successful solution with %lu states", getName().c_str(),
                     getOMPLSimpleSetup()->getSolutionPath().getStateCount());
+    res.error_code_.val = moveit_result.val;
     return true;
   }
   else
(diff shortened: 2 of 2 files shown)


moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -709,14 +707,14 @@ bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::Motion
...
-    res.error_code_.val = moveit_msgs::MoveItErrorCodes::PLANNING_FAILED;
     return false;
   }
 }
 
 bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::MotionPlanDetailedResponse& res)
 {
-  if (solve(request_.allowed_planning_time, request_.num_planning_attempts))
+  moveit_msgs::MoveItErrorCodes moveit_result = solve(request_.allowed_planning_time, request_.num_planning_attempts);
+  if (moveit_result.val == moveit_msgs::MoveItErrorCodes::SUCCESS)
```
rhaschke, line 717: Why introducing a new variable here? Can't you use `res.error_code_`?
```suggestion
  res.error_code_.val = solve(request_.allowed_planning_time, request_.num_planning_attempts);
  if (res.error_code_.val == moveit_msgs::MoveItErrorCodes::SUCCESS)
```

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -750,9 +748,9 @@ bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::Motion
       getSolutionPath(*res.trajectory_.back());
     }
 
-    // fill the response
     ROS_DEBUG_NAMED(LOGNAME, "%s: Returning successful solution with %lu states", getName().c_str(),
                     getOMPLSimpleSetup()->getSolutionPath().getStateCount());
+    res.error_code_.val = moveit_result.val;
```
rhaschke, line 753: Then this becomes obsolete.
```suggestion
```

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -763,21 +761,24 @@ bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::Motion
...
-  bool result = false;
+  moveit_msgs::MoveItErrorCodes result;
+  result.val = moveit_msgs::MoveItErrorCodes::FAILURE;
   if (count <= 1 || multi_query_planning_enabled_)  // multi-query planners should always run in single instances
   {
     ROS_DEBUG_NAMED(LOGNAME, "%s: Solving the planning problem once...", name_.c_str());
     ob::PlannerTerminationCondition ptc = constructPlannerTerminationCondition(timeout, start);
     registerTerminationCondition(ptc);
-    result = ompl_simple_setup_->solve(ptc) == ompl::base::PlannerStatus::EXACT_SOLUTION;
+    result.val = ompl_simple_setup_->solve(ptc) == ompl::base::PlannerStatus::EXACT_SOLUTION;
```
rhaschke, line 777: This shouldn't compile?!

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -795,7 +796,10 @@ bool ompl_interface::ModelBasedPlanningContext::solve(double timeout, unsigned i
 
       ob::PlannerTerminationCondition ptc = constructPlannerTerminationCondition(timeout, start);
       registerTerminationCondition(ptc);
-      result = ompl_parallel_plan_.solve(ptc, 1, count, hybridize_) == ompl::base::PlannerStatus::EXACT_SOLUTION;
+      if (ompl_parallel_plan_.solve(ptc, 1, count, hybridize_) == ompl::base::PlannerStatus::EXACT_SOLUTION)
```
rhaschke, line 799: What about the else case?

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/model_based_planning_context.h
```diff
@@ -371,6 +371,9 @@ class ModelBasedPlanningContext : public planning_interface::PlanningContext
   void registerTerminationCondition(const ob::PlannerTerminationCondition& ptc);
   void unregisterTerminationCondition();
 
+  /** \brief Convert OMPL PlannerStatus to moveit_msgs::msg::MoveItErrorCode */
+  int32_t logPlannerStatus(og::SimpleSetupPtr ompl_simple_setup);
```
rhaschke, line 375: The name and signature are somewhat misleading, aren't they? What about:
```suggestion
  int32_t errorCode(const ompl::base::PlannerStatus& status);
```

rhaschke (commented):
Some comments:

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
AndyZe: This must have been from an earlier commit. It is different now:

```
      if (ompl_parallel_plan_.solve(ptc, 1, count, hybridize_) == ompl::base::PlannerStatus::EXACT_SOLUTION)
      {
        result.val = moveit_msgs::MoveItErrorCodes::SUCCESS;
      }
```

moveit_planners/ompl/ompl_interface/include/moveit/ompl_interface/model_based_planning_context.h
AndyZe: yep, that's more useful in the case of `ompl_parallel_plan_` too  :+1:

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
AndyZe: This has been refactored to:

`result.val = errorCode(ompl_parallel_plan_.solve(ptc, 1, count, hybridize_));`

So there is no need for an if/else anymore.

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -852,6 +857,54 @@ void ompl_interface::ModelBasedPlanningContext::unregisterTerminationCondition()
...
+      ROS_WARN_NAMED(LOGNAME, "Unrecognized goal type");
+      result = moveit_msgs::MoveItErrorCodes::UNRECOGNIZED_GOAL_TYPE;
+      break;
+    case ompl::base::PlannerStatus::TIMEOUT:
+      ROS_WARN_NAMED(LOGNAME, "Timed out");
+      result = moveit_msgs::MoveItErrorCodes::TIMED_OUT;
+      break;
+    case ompl::base::PlannerStatus::APPROXIMATE_SOLUTION:
+      ROS_WARN_NAMED(LOGNAME, "Solution is approximate");
+      result = moveit_msgs::MoveItErrorCodes::SUCCESS;
```
simonschmeisser, line 887:
```suggestion
      result = moveit_msgs::MoveItErrorCodes::PLANNING_FAILED;
```
"approximate" can be anything really, for RRTConnect it can refer to a "solution" that consists of only a couple of the steps near the goal

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -795,7 +795,7 @@ bool ompl_interface::ModelBasedPlanningContext::solve(double timeout, unsigned i
 
       ob::PlannerTerminationCondition ptc = constructPlannerTerminationCondition(timeout, start);
       registerTerminationCondition(ptc);
-      result = ompl_parallel_plan_.solve(ptc, 1, count, hybridize_) == ompl::base::PlannerStatus::EXACT_SOLUTION;
+      result.val = errorCode(ompl_parallel_plan_.solve(ptc, 1, count, hybridize_));
```
simonschmeisser, line 798: Well here you changed the results by converting `APPROXIMATE` to `true` instead of `false` as before

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
simonschmeisser: I looked it up again because I didn't remember the details. Officially `approximate` means "any trajectory that starts at the start state but does not reach the goal state". There was a bug in RRTConnect where the opposite would also be approximate but that was fixed in https://github.com/ompl/ompl/pull/739    ... which has not been released yet

simonschmeisser: Yes, especially with the bug in OMPL still present in release builds this could give you "solutions" that contain a huge jump towards the goal state. Once this is released we could reconsider if a solution that "gets you away from start" can be somehow useful in any case? Note that it would not necessarily be closer to the goal or the goal easier to reach from the new position ...

AndyZe: Done -- thanks

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
AndyZe: Should be addressed by the latest commit

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -763,21 +760,24 @@ bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::Motion
...
   {
     ROS_DEBUG_NAMED(LOGNAME, "%s: Solving the planning problem once...", name_.c_str());
     ob::PlannerTerminationCondition ptc = constructPlannerTerminationCondition(timeout, start);
     registerTerminationCondition(ptc);
-    result = ompl_simple_setup_->solve(ptc) == ompl::base::PlannerStatus::EXACT_SOLUTION;
+    result.val = ompl_simple_setup_->solve(ptc) == ompl::base::PlannerStatus::EXACT_SOLUTION;
     last_plan_time_ = ompl_simple_setup_->getLastPlanComputationTime();
     unregisterTerminationCondition();
+    // fill the result status code
+    result.val = errorCode(ompl_simple_setup_->getLastPlannerStatus());
```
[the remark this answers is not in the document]
AndyZe: You mean like this? It gives an error.

`result.val = ompl_simple_setup_->getLastPlannerStatus()`

`error: conversion from ‘ompl::base::PlannerStatus’ to ‘moveit_msgs::MoveItErrorCodes_<std::allocator<void> >::_val_type’ {aka ‘int’} is ambiguous`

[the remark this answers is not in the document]
simonschmeisser:
```suggestion
```
no, like this, somehow the suggestion button wasn't there.

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -763,21 +760,24 @@ bool ompl_interface::ModelBasedPlanningContext::solve(planning_interface::Motion
...
-  bool result = false;
+  moveit_msgs::MoveItErrorCodes result;
+  result.val = moveit_msgs::MoveItErrorCodes::FAILURE;
   if (count <= 1 || multi_query_planning_enabled_)  // multi-query planners should always run in single instances
   {
     ROS_DEBUG_NAMED(LOGNAME, "%s: Solving the planning problem once...", name_.c_str());
     ob::PlannerTerminationCondition ptc = constructPlannerTerminationCondition(timeout, start);
     registerTerminationCondition(ptc);
-    result = ompl_simple_setup_->solve(ptc) == ompl::base::PlannerStatus::EXACT_SOLUTION;
+    result.val = ompl_simple_setup_->solve(ptc) == ompl::base::PlannerStatus::EXACT_SOLUTION;
```
simonschmeisser, line 776:
```suggestion
    result.val = errorCode(ompl_simple_setup_->solve(ptc));
```

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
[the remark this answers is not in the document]
AndyZe: I see -- done!

simonschmeisser (approved):

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -815,7 +813,10 @@ bool ompl_interface::ModelBasedPlanningContext::solve(double timeout, unsigned i
           for (unsigned int i = 0; i < max_planning_threads_; ++i)
             ompl_parallel_plan_.addPlanner(ompl::tools::SelfConfig::getDefaultPlanner(ompl_simple_setup_->getGoal()));
         bool r = ompl_parallel_plan_.solve(ptc, 1, count, hybridize_) == ompl::base::PlannerStatus::EXACT_SOLUTION;
-        result = result && r;
+        // Was this latest call successful too?
+        result.val = (result.val == moveit_msgs::MoveItErrorCodes::SUCCESS && r) ?
+                         moveit_msgs::MoveItErrorCodes::SUCCESS :
+                         moveit_msgs::MoveItErrorCodes::FAILURE;
```
rhaschke, line 819: I agree that this correctly translates the original code's idea.
However, it becomes more difficult to comprehend. As the actual OMPL error code isn't used anymore (which should be fixed IMO), I suggest a simplification:
https://github.com/AndyZe/moveit/pull/3

I think there is also a logical error in the old code:
Running multiple planning attempts in parallel (or sequence as here) is done to increase the chance of planning success, isn't it? Thus, only return success if all attempts succeeded doesn't make sense to me.

rhaschke (changes requested):
There are open issues to resolve before merging:

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
rhaschke: I addressed the remaining issues here:
https://github.com/ubi-agni/moveit/compare/pr/AndyZe/3257...ubi-agni:moveit:fix-parallel-planning

I suggest merging this PR with the proposed changes (https://github.com/AndyZe/moveit/pull/3) and subsequently handle the code cleanup in a separate PR.

rhaschke (approved):

PR #3258, opened 2022-10-31
 Planning request adapters: short-circuit if failure, return code rather than bool

AndyZe:
This is a backport of this MoveIt2 PR: https://github.com/ros-planning/moveit2/pull/1605

### DIFF
--- a/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
+++ b/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
@@ -42,22 +42,30 @@ namespace planning_request_adapter
 {
 namespace
 {
-bool callPlannerInterfaceSolve(const planning_interface::PlannerManager& planner,
-                               const planning_scene::PlanningSceneConstPtr& planning_scene,
-                               const planning_interface::MotionPlanRequest& req,
-                               planning_interface::MotionPlanResponse& res)
+moveit::core::MoveItErrorCode callPlannerInterfaceSolve(const planning_interface::PlannerManager& planner,
+                                                        const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                                        const planning_interface::MotionPlanRequest& req,
+                                                        planning_interface::MotionPlanResponse& res)
 {
   planning_interface::PlanningContextPtr context = planner.getPlanningContext(planning_scene, req, res.error_code_);
   if (context)
-    return context->solve(res);
+  {
+    // TODO(andyz): consider returning a moveit::core::MoveItErrorCode from context->solve()
+    bool result = context->solve(res);
+    return result ? moveit::core::MoveItErrorCode::SUCCESS : moveit::core::MoveItErrorCode::FAILURE;
+  }
   else
-    return false;
+  {
+    return moveit::core::MoveItErrorCode::FAILURE;
+  }
 }
 
-bool callAdapter(const PlanningRequestAdapter& adapter, const PlanningRequestAdapter::PlannerFn& planner,
-                 const planning_scene::PlanningSceneConstPtr& planning_scene,
-                 const planning_interface::MotionPlanRequest& req, planning_interface::MotionPlanResponse& res,
-                 std::vector<std::size_t>& added_path_index)
+moveit::core::MoveItErrorCode callAdapter(const PlanningRequestAdapter& adapter,
+                                          const PlanningRequestAdapter::PlannerFn& planner,
+                                          const planning_scene::PlanningSceneConstPtr& planning_scene,
+                                          const planning_interface::MotionPlanRequest& req,
+                                          planning_interface::MotionPlanResponse& res,
+                                          std::vector<std::size_t>& added_path_index)
 {
   try
   {
@@ -75,11 +83,10 @@ bool callAdapter(const PlanningRequestAdapter& adapter, const PlanningRequestAda
 
 }  // namespace
 
-bool PlanningRequestAdapter::adaptAndPlan(const planning_interface::PlannerManagerPtr& planner,
-                                          const planning_scene::PlanningSceneConstPtr& planning_scene,
-                                          const planning_interface::MotionPlanRequest& req,
-                                          planning_interface::MotionPlanResponse& res,
-                                          std::vector<std::size_t>& added_path_index) const
+moveit::core::MoveItErrorCode PlanningRequestAdapter::adaptAndPlan(
+    const planning_interface::PlannerManagerPtr& planner, const planning_scene::PlanningSceneConstPtr& planning_scene,
+    const planning_interface::MotionPlanRequest& req, planning_interface::MotionPlanResponse& res,
+    std::vector<std::size_t>& added_path_index) const
 {
   return adaptAndPlan(
       [&planner](const planning_scene::PlanningSceneConstPtr& scene, const planning_interface::MotionPlanRequest& req,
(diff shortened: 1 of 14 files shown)


moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -138,24 +142,27 @@ bool PlanningRequestAdapterChain::adaptAndPlan(const planning_interface::Planner
       fn = [&adapter = *adapters_[i], fn, &added_path_index = added_path_index_each[i]](
                const planning_scene::PlanningSceneConstPtr& scene, const planning_interface::MotionPlanRequest& req,
                planning_interface::MotionPlanResponse& res) {
+        // Abort pipeline and return in case of failure
         return callAdapter(adapter, fn, scene, req, res, added_path_index);
```
AndyZe, line 146: I think there's an opportunity here to abort even earlier. Will look into it.

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_path_constraints.cpp
```diff
@@ -112,24 +115,19 @@ class FixStartStatePathConstraints : public planning_request_adapter::PlanningRe
...
+      if (!bool(solved1) || !bool(solved2))
       {
-        ROS_WARN("Unable to plan to path constraints. Running usual motion plan.");
-        bool result = planner(planning_scene, req, res);
-        res.planning_time_ += res2.planning_time_;
-        return result;
+        ROS_WARN("Unable to meet path constraints at the start.");
+        moveit::core::MoveItErrorCode moveit_code(moveit_msgs::MoveItErrorCodes::START_STATE_VIOLATES_PATH_CONSTRAINTS);
+        res.error_code_.val = moveit_msgs::MoveItErrorCodes::START_STATE_VIOLATES_PATH_CONSTRAINTS;
+        return moveit_code;
```
sjahr, line 126: Nit: Would that work as well?
```suggestion
        res.error_code_.val = moveit_msgs::MoveItErrorCodes::START_STATE_VIOLATES_PATH_CONSTRAINTS;
        return moveit::core::MoveItErrorCode(res.error_code_.val);
```

sjahr (approved):
Looks good to me. Feel free to ignore my small Nit comment if it does not work or you don't like the proposed change

moveit_ros/planning/planning_request_adapter_plugins/src/fix_start_state_path_constraints.cpp
AndyZe: Yep, nice simplification

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
```diff
@@ -42,22 +42,30 @@ namespace planning_request_adapter
...
+                                                        const planning_interface::MotionPlanRequest& req,
+                                                        planning_interface::MotionPlanResponse& res)
 {
   planning_interface::PlanningContextPtr context = planner.getPlanningContext(planning_scene, req, res.error_code_);
   if (context)
-    return context->solve(res);
+  {
+    // TODO(andyz): consider returning a moveit::core::MoveItErrorCode from context->solve()
+    bool result = context->solve(res);
+    return result ? moveit::core::MoveItErrorCode::SUCCESS : moveit::core::MoveItErrorCode::FAILURE;
```
rhaschke, line 55: As long as `context->solve()` doesn't return an actual error code, the distinction between `SUCCESS` and `FAILURE` is the same as `true` and `false`.

rhaschke (commented):
Honestly, I don't yet see an improvement in this PR, which was unfortunately already merged.
As far as I can see, this "just" replaces the bool return code with a MoveIt error code. But why?
The error code is actually passed in `MotionPlanResponse`. Hence, there is no need to bloat the API as done here.

The PR title suggests some performance improvement ("short-circuit on failure"). I couldn't spot this. Can you please point that out?

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
AndyZe: @rhaschke you are right that the short-circuiting was never implemented. I tried on this branch but failed so far:  https://github.com/AndyZe/moveit2/tree/andyz/adapter_short_circuit

The PR title should have been changed but this way matches with what was merged into MoveIt2. I'm sorry about that.

There is an improvement in the returned code, though, as described here: https://github.com/ros-planning/moveit2/pull/1605#issue-1399761923

Previously, if planning failed, a 0 was always returned. That's not very informative.

rhaschke: I have read the original PR description. 
But I couldn't find the source code line generating the output mentioned there, e.g.:
`[moveit_cpp_tutorial-4] [ERROR] [1665063576.588242808] [moveit.planning_request_adapter]: Adapter  result: -10`
Could you please point that out?

I think the same output result could be easily achieved by printing `res.error_code_` instead of the return value - without changing the code as much as here.

AndyZe: Yes, i'll make a test branch for you in a bit

AndyZe: Here are 2 branches for you. The first one logs adapter success codes prior to this PR. The second logs post-PR.

https://github.com/AndyZe/moveit/tree/andyz/adapter_logging1
https://github.com/AndyZe/moveit/tree/andyz/adapter_logging2

If you check out either branch, add a collision obstacle in RViz, then try to plan a motion, you should see the prints. This is what I get:

**1**
```
[ERROR] [1667508713.025458736]: panda_arm/panda_arm: Motion planning start tree could not be initialized!
[ WARN] [1667508713.025508832]: ParallelPlan::solve(): Unable to find solution by any of the threads in 0.000220 seconds
[ WARN] [1667508713.033321659]: Goal sampling thread never did any work.
[ INFO] [1667508713.033496990]: Unable to solve the planning problem
[ERROR] [1667508713.033546494]: Adapter result: 0
[ERROR] [1667508713.033600887]: Adapter result: 0
[ERROR] [1667508713.033633277]: Adapter result: 0
[ERROR] [1667508713.033645159]: Adapter result: 0
[ERROR] [1667508713.033709058]: Adapter result: 0
[ERROR] [1667508713.033737349]: Adapter result: 0
[ WARN] [1667508713.054276705]: Fail: ABORTED: Catastrophic failure
```

**2**
```
[ INFO] [1667510127.223233904]: Start state appears to be in collision with respect to group panda_arm
[ WARN] [1667510127.315109714]: Unable to find a valid state nearby the start state (using jiggle fraction of 0.050000 and 100 sampling attempts).
[ERROR] [1667510127.315232783]: Adapter result: val: -10

[ERROR] [1667510127.315271245]: Adapter result: val: -10

[ERROR] [1667510127.315306039]: Adapter result: val: -10

[ERROR] [1667510127.315342869]: Adapter result: val: -10

[ERROR] [1667510127.315376119]: Adapter result: val: -10
```

I hope you do find an easier way!

AndyZe: Also, it looks like the `master` branch has collision detection issues unrelated to this PR. Sigh

rhaschke:
You were right, @AndyZe, your PR had short-circuiting of some planning adapters on failure. But I missed these tiny changes in comparison to the bulk change of the return types.

I quickly implemented an alternative solution to your PR: https://github.com/ros-planning/moveit/pull/3262. Please have a look.
Maybe, there are other planning adapters that need attention.

This PR here was reverted on the master branch.

moveit_core/planning_request_adapter/src/planning_request_adapter.cpp
rhaschke: > it looks like the `master` branch has collision detection issues unrelated to this PR.

What exactly do you mean? I didn't notice any issues.

AndyZe: > What exactly do you mean? I didn't notice any issues.

It doesn't matter right now. Let's focus on this first.

AndyZe:
> This PR here was reverted on the master branch.

Robert, several maintainers reviewed that PR before it was merged, in MoveIt1 and MoveIt2. Tyler, Henning, and Sebastian. To single-handedly revert it was rude. You could have reverted it as part of #3262 so the reversion gets reviewed, at least.

I'll review your #3262 now and I'll try to keep an open mind despite your single-handed reversion.

rhaschke:
Sorry for reverting this w/o additional review. I did that intentionally _prior_ to filing #3262 to have its diff against the original master. Otherwise, the diff would have been cluttered by the reversions of the return-value changes again, thus masking the actual content of the PR. I hope you agree that my proposal is more appealing. I suggest doing the same in MoveIt2 as well and will file an appropriate PR.

PR #3261, opened 2022-11-03
Improve processing of multiple planning attempts

rhaschke:
This is a follow-up to #3257, cleaning up the code of parallel planning. This PR factors out common code into a lambda function and fixes the evaluation of the planning result (keeping the first successful plan).

### DIFF
--- a/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
+++ b/moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
@@ -767,72 +767,47 @@ const moveit_msgs::MoveItErrorCodes ompl_interface::ModelBasedPlanningContext::s
 
   moveit_msgs::MoveItErrorCodes result;
   result.val = moveit_msgs::MoveItErrorCodes::FAILURE;
+  ob::PlannerTerminationCondition ptc = constructPlannerTerminationCondition(timeout, start);
+  registerTerminationCondition(ptc);
   if (count <= 1 || multi_query_planning_enabled_)  // multi-query planners should always run in single instances
   {
     ROS_DEBUG_NAMED(LOGNAME, "%s: Solving the planning problem once...", name_.c_str());
-    ob::PlannerTerminationCondition ptc = constructPlannerTerminationCondition(timeout, start);
-    registerTerminationCondition(ptc);
     result.val = errorCode(ompl_simple_setup_->solve(ptc));
     last_plan_time_ = ompl_simple_setup_->getLastPlanComputationTime();
-    unregisterTerminationCondition();
   }
   else
   {
     ROS_DEBUG_NAMED(LOGNAME, "%s: Solving the planning problem %u times...", name_.c_str(), count);
     ompl_parallel_plan_.clearHybridizationPaths();
-    if (count <= max_planning_threads_)
-    {
+
+    auto plan_parallel = [this, &ptc](unsigned int num_planners) {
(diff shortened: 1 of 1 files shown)


moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -767,72 +767,46 @@ const moveit_msgs::MoveItErrorCodes ompl_interface::ModelBasedPlanningContext::s
...
-    unregisterTerminationCondition();
   }
   else
   {
     ROS_DEBUG_NAMED(LOGNAME, "%s: Solving the planning problem %u times...", name_.c_str(), count);
     ompl_parallel_plan_.clearHybridizationPaths();
-    if (count <= max_planning_threads_)
-    {
+
+    auto plan_parallel = [this, &ptc](unsigned int num_planners) {
```
AndyZe, line 783: nit

```suggestion
    auto plan_parallel = [this, &ptc](unsigned int num_planners)
    {
```

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
```diff
@@ -767,72 +767,46 @@ const moveit_msgs::MoveItErrorCodes ompl_interface::ModelBasedPlanningContext::s
...
-        else
-          for (int i = 0; i < n; ++i)
-            ompl_parallel_plan_.addPlanner(ompl::tools::SelfConfig::getDefaultPlanner(ompl_simple_setup_->getGoal()));
-        bool r = ompl_parallel_plan_.solve(ptc, 1, count, hybridize_) == ompl::base::PlannerStatus::EXACT_SOLUTION;
-        res = res && r;
-      }
-      last_plan_time_ = ompl::time::seconds(ompl::time::now() - start);
-      unregisterTerminationCondition();
-      result.val = res ? moveit_msgs::MoveItErrorCodes::SUCCESS : moveit_msgs::MoveItErrorCodes::FAILURE;
+      for (int i = 0; i < n && result.val != moveit_msgs::MoveItErrorCodes::SUCCESS && !ptc(); ++i)
```
AndyZe, line 802: This bit is a little questionable:

`result.val != moveit_msgs::MoveItErrorCodes::SUCCESS && !ptc();`

If a success gets reported then why do we care about `ptc()`?

AndyZe (commented):
I haven't tested it yet, these are just some thoughts from looking at the code.

If you have instructions on _how_ to test it, that would be nice. Does the [MoveIt1 quickstart tutorial ](https://ros-planning.github.io/moveit_tutorials/doc/quickstart_in_rviz/quickstart_in_rviz_tutorial.html) use parallel planning?

moveit_planners/ompl/ompl_interface/src/model_based_planning_context.cpp
rhaschke: Interestingly clang-format didn't complain about this!?

rhaschke: Actually clang-format would revert this suggestion!

rhaschke:
This is simply parallel planning for multiple planning attempts. I reworded the commit message to clarify this.
Thus, to test, just perform standard planning from rviz, which uses 10 planning attempts by default.

AndyZe (approved):
I've tested it now  :+1:

PR #3263, opened 2022-11-06
MotionPlanningDisplay: Fix RobotState inconsistency

### LINKED ISSUE #3207: Collision of newly attached object not taken into account for first plan
### Description

Collision of newly attached object not taken into account for first plan.

### Your environment
* ROS Distro: Melodic
* OS Version: Ubuntu 18.04
* Binary
* 1.0.10-Alpha

### Steps to reproduce
1. load a panda robot and launch RViz, similar to [the tutorial](http://docs.ros.org/en/melodic/api/moveit_tutorials/html/doc/move_group_python_interface/move_group_python_interface_tutorial.html)
2. add a box obstacle to the planning scene using the python API [`planning_scene_interface.add_box(...)`]
3. add a sphere not too far from the robot's gripper, and attach it to the robot's gripper [`planning_scene_interface.add_sphere(...)`, `planning_scene_interface.attach_box(...)`]
4. plan a path between two arm configurations such that the attached sphere should easily collide with the non-attached box

### Expected behaviour
Planning should take into account the newly created and attached sphere to avoid collisions.

### Actual behaviour
This results in a plan that doesn't take the sphere-box collision into account. However, after this initial plan is executed, the sphere-box collision is properly taken into account for following plans.
The symmetric issue also exists: a just detached-and-removed sphere is still taken into account for the next planning/execution step.

No issue there.

rhaschke:
Fixes #3207, i.e.
1. attached objects missing in planning requests triggered from rviz, because automatic query state updates were only performed if the state was visible.
2. disabling execution if the attached bodies of the current RobotState and don't match those of the plan's start state.
Unfortunately, the trajectory execution [service](http://docs.ros.org/en/api/moveit_msgs/html/srv/ExecuteKnownTrajectory.html) and [action](http://docs.ros.org/en/api/moveit_msgs/html/action/ExecuteTrajectory.html) do not receive the plan's start state (but only the [`RobotTrajectory`](http://docs.ros.org/en/api/moveit_msgs/html/msg/RobotTrajectory.html)). Thus we cannot reject an invalid plan at that level (which would be the correct one in my opinion).

### DIFF
--- a/moveit_commander/src/moveit_commander/planning_scene_interface.py
+++ b/moveit_commander/src/moveit_commander/planning_scene_interface.py
@@ -64,23 +64,24 @@ class PlanningSceneInterface(object):
     See wrap_python_planning_scene_interface.cpp for the wrapped methods.
     """
 
-    def __init__(self, ns="", synchronous=False, service_timeout=5.0):
+    def __init__(self, ns="", synchronous=True, service_timeout=5.0):
         self._psi = _moveit_planning_scene_interface.PlanningSceneInterface(ns)
-
-        self._pub_co = rospy.Publisher(
-            ns_join(ns, "collision_object"), CollisionObject, queue_size=100
-        )
-        self._pub_aco = rospy.Publisher(
-            ns_join(ns, "attached_collision_object"),
-            AttachedCollisionObject,
-            queue_size=100,
-        )
         self.__synchronous = synchronous
+
         if self.__synchronous:
             self._apply_planning_scene_diff = rospy.ServiceProxy(
                 ns_join(ns, "apply_planning_scene"), ApplyPlanningScene
             )
             self._apply_planning_scene_diff.wait_for_service(service_timeout)
+        else:
+            self._pub_co = rospy.Publisher(
+                ns_join(ns, "collision_object"), CollisionObject, queue_size=100
+            )
+            self._pub_aco = rospy.Publisher(
+                ns_join(ns, "attached_collision_object"),
+                AttachedCollisionObject,
+                queue_size=100,
+            )
 
     def __submit(self, collision_object, attach=False):
         if self.__synchronous:
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -2300,5 +2300,54 @@ std::ostream& operator<<(std::ostream& out, const RobotState& s)
   return out;
 }
 
+bool haveSameAttachedObjects(const RobotState& left, const RobotState& right, const std::string& prefix)
+{
+  std::vector<const moveit::core::AttachedBody*> left_attached;
+  std::vector<const moveit::core::AttachedBody*> right_attached;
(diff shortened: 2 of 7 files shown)


rhaschke:
Note, this issue exists since the very beginning (> 10 years). Thus, MoveIt2 is for sure also affected and this PR should be ported.

moveit_commander/src/moveit_commander/planning_scene_interface.py
```diff
@@ -64,23 +64,24 @@ class PlanningSceneInterface(object):
...
-            queue_size=100,
-        )
         self.__synchronous = synchronous
+
         if self.__synchronous:
             self._apply_planning_scene_diff = rospy.ServiceProxy(
                 ns_join(ns, "apply_planning_scene"), ApplyPlanningScene
             )
             self._apply_planning_scene_diff.wait_for_service(service_timeout)
+        else:
```
AndyZe, line 76: Please add a comment about why these publishers were moved.

moveit_core/robot_state/src/robot_state.cpp
```diff
@@ -2300,5 +2300,54 @@ std::ostream& operator<<(std::ostream& out, const RobotState& s)
...
+    {
+      ROS_DEBUG_STREAM(prefix << "different object shapes: " << left_object->getName());
+      return false;  // shapes not matching
+    }
+
+    auto left_it = left_object->getShapePosesInLinkFrame().cbegin();
+    auto left_end = left_object->getShapePosesInLinkFrame().cend();
+    auto right_it = right_object->getShapePosesInLinkFrame().cbegin();
+    for (; left_it != left_end; ++left_it, ++right_it)
+      if (!(left_it->matrix() - right_it->matrix()).isZero(1e-4))
```
AndyZe, line 2343: I think these lines are cool!

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
```diff
@@ -1235,15 +1235,17 @@ void MotionPlanningDisplay::updateStateExceptModified(moveit::core::RobotState&
 void MotionPlanningDisplay::updateQueryStates(const moveit::core::RobotState& current_state)
 {
   std::string group = planning_group_property_->getStdString();
+  if (group.empty())
+    return;
```
AndyZe, line 1239: May want to print a warning here

AndyZe (approved):

moveit_commander/src/moveit_commander/planning_scene_interface.py
rhaschke: The publishers are only needed in asynchronous mode.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp
rhaschke: That's not a function triggered by the user. Thus, I don't see the need for a warning here.

PR #3264, opened 2022-11-06
MotionPlanningDisplay: Fix PlanningScene editing

rhaschke:
While working on #3263, I noticed several issues with planning scene editing, which I attempt to address here:
- [x] Allow extended selection mode in object list (only single selection was allowed before)
    Note: interactive marker handles are shown for the _current_ item only
- [x] "Clear Scene" button should clear rviz' local scene only instead of syncing automatically to the move_group's scene
- [x] Removing single objects from the scene isn't reflected in the query states
- [ ] Attached objects cannot be moved yet. I suggest to show their pose relative to the link their are attached to.
- [ ] It is not clear how incoming planning scene updates (from `move_group` node via topic `/move_group/monitored_planning_scene` and from current state monitor via `/joint_states`) should be handled. This is the current status:
  - [joint state updates are only considered, if the corresponding planning group was never selected before](https://github.com/ros-planning/moveit/blob/fe0c182a57b79f970eaca26c479e96d857242505/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp#L1216-L1234).
    This partially makes sense, because we don't want a goal query state be overridden by actual robot movements.
    However, [currently, updates are disabled as soon as the planning group was selected](https://github.com/ros-planning/moveit/blob/fe0c182a57b79f970eaca26c479e96d857242505/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_display.cpp#L1033).
    I suggest to change this as follows:
    - Only add the current planning group to `modified_groups_` if the corresponding query state was _actually_ modified.
    - Remove the planning group from the list if `<current>` was selected (and query state thus reset)
    - Have separate `modified_groups_` variables for start and goal query state.
  - Updates to collision and attached objects are always considered. This might override changes made to rviz' scene locally.

### DIFF
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -500,6 +500,7 @@ void MotionPlanningFrame::addSceneObject()
   populateCollisionObjectsList();
 
   // Automatically select the inserted object so that its IM is displayed
+  ui_->collision_objects_list->clearSelection();
   setItemSelectionInList(shape_name, true, ui_->collision_objects_list);
 
   planning_display_->queueRenderSceneGeometry();
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -487,10 +487,12 @@ void MotionPlanningFrame::addSceneObject()
 
     // Actually add object to the plugin's PlanningScene
     ps->getWorldNonConst()->addToObject(shape_name, shape, Eigen::Isometry3d::Identity());
-    populateCollisionObjectsList(&ps);
   }
   setLocalSceneEdited();
   planning_display_->queueRenderSceneGeometry();
+
+  // Finally add object name to GUI list
+  auto item = addCollisionObjectToList(shape_name, ui_->collision_objects_list->count(), false);
 
   // Automatically select the inserted object so that its IM is displayed
   ui_->collision_objects_list->clearSelection();
--- a/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
+++ b/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
@@ -232,13 +232,6 @@ void MotionPlanningFrame::approximateIKChanged(int state)
   planning_display_->useApproximateIK(state == Qt::Checked);
 }
 
-void MotionPlanningFrame::setItemSelectionInList(const std::string& item_name, bool selection, QListWidget* list)
-{
-  QList<QListWidgetItem*> found_items = list->findItems(QString(item_name.c_str()), Qt::MatchExactly);
-  for (QListWidgetItem* found_item : found_items)
-    found_item->setSelected(selection);
-}
-
 void MotionPlanningFrame::allowExternalProgramCommunication(bool enable)
 {
   // This is needed to prevent UI event (resuming the options) triggered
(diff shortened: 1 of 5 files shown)


moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -215,22 +212,29 @@ void MotionPlanningFrame::sceneScaleEndChange()
...
 {
-  QList<QListWidgetItem*> sel = ui_->collision_objects_list->selectedItems();
-  if (sel.empty())
+  QList<QListWidgetItem*> selection = ui_->collision_objects_list->selectedItems();
+  if (selection.empty())
     return;
-  planning_scene_monitor::LockedPlanningSceneRW ps = planning_display_->getPlanningSceneRW();
-  if (ps)
+
+  if (planning_scene_monitor::LockedPlanningSceneRW ps = planning_display_->getPlanningSceneRW())
```
AndyZe, line 219: I don't think this `if()` does anything. It looks like [`getPlanningSceneRW()`](https://github.com/ros-planning/moveit/blob/1f4435190f7498d746b8066fb72c956eebe8eeef/moveit_ros/visualization/planning_scene_rviz_plugin/src/planning_scene_display.cpp#L300) and the [constructor of LockedPlanningSceneRW](https://github.com/ros-planning/moveit/blob/1f4435190f7498d746b8066fb72c956eebe8eeef/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L614) cannot fail.

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -215,22 +212,29 @@ void MotionPlanningFrame::sceneScaleEndChange()
...
   {
```
AndyZe, line 220:
```suggestion
  {
    planning_scene_monitor::LockedPlanningSceneRW ps = planning_display_->getPlanningSceneRW();
```

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
```diff
@@ -142,11 +142,8 @@ void MotionPlanningFrame::clearScene()
...
     ps->getWorldNonConst()->clearObjects();
```
AndyZe, line 143: How would I trigger this to test it? I don't see a service call nor a button called "Clear" in the Motion Planning GUI.

AndyZe: Ah, I found it. It's the tiny button with the gray tab/X.

AndyZe:
I'm sorry to say, there appears to be a deadlock if multiple collision objects are loaded in the GUI then "Clear planning scene" is clicked.

https://user-images.githubusercontent.com/11284393/206752695-8ed2344c-1a30-480a-810d-894a6f01b00b.mp4

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_objects.cpp
rhaschke: `planning_scene_monitor_` could be a `nullptr`, couldn't it?
https://github.com/ros-planning/moveit/blob/1f4435190f7498d746b8066fb72c956eebe8eeef/moveit_ros/planning/planning_scene_monitor/include/moveit/planning_scene_monitor/planning_scene_monitor.h#L625-L628

moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame.cpp
```diff
@@ -497,12 +497,11 @@ void MotionPlanningFrame::addSceneObject()
   }
   setLocalSceneEdited();
 
-  planning_display_->addMainLoopJob([this] { populateCollisionObjectsList(); });
+  populateCollisionObjectsList();
```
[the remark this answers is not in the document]
rhaschke: Because this Qt slot function already runs in the main loop.

rhaschke:
> there appears to be a deadlock if multiple scene objects are loaded in the GUI then "Clear planning scene" is clicked.

Thanks for thoroughly testing this! I need to check where this comes from... Maybe a recursive lock?

rhaschke:
Closing and re-opening to re-trigger CI.
@AndyZe, I fixed the deadlock you found. It was - as expected - a recursive lock of the same mutex. Along the way, I found and fixed various other issues. If this PR gets merged, it should definitely _not_ be squashed, to preserve the full detailed history.

PR #3269, opened 2022-11-15
kinematics: add required Boost component

jspricke:


v4hn (approved):
Approved assuming CI succeeds. I have the same patch lying around but didn't get around to push it yet. :)

PR #3276, opened 2022-11-22
IK solver should return false if constraint is not satisfied

AndyZe:
This fixes a bug in `RobotState::setFromIK()` where the optional `GroupStateValidityCallbackFn()` fails. I think false should be returned.

### DIFF
--- a/moveit_core/robot_state/src/robot_state.cpp
+++ b/moveit_core/robot_state/src/robot_state.cpp
@@ -1515,10 +1515,15 @@ bool ikCallbackFnAdapter(RobotState* state, const JointModelGroup* group,
   for (std::size_t i = 0; i < bij.size(); ++i)
     solution[bij[i]] = ik_sol[i];
   if (constraint(state, group, &solution[0]))
+  {
     error_code.val = moveit_msgs::MoveItErrorCodes::SUCCESS;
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
While it makes sense to return false in that case, the return value isn't actually used in any IK plugin, e.g. KDL:
https://github.com/ros-planning/moveit/blob/72a9042773cd991fc77e689ca5e6c39999b85df1/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp#L402-L404

AndyZe:
I'm using this with [bio_ik](https://github.com/TAMS-Group/bio_ik) and somehow, my change here seems to make a difference. Let me verify

rhaschke:
> I'm using this with [bio_ik](https://github.com/TAMS-Group/bio_ik) and somehow, my change here seems to make a difference. Let me verify

For sure, the return value doesn't make a difference, because it wasn't propagated (#3277). Actually, I expect a stack error if the return value would have been used.

rhaschke:
Closing as this function shouldn't return anything at all (https://github.com/ros-planning/moveit/pull/3277#issuecomment-1324325699)

PR #3277, opened 2022-11-22
Fix ik_callback_fn: propagate return value

rhaschke:
Reviewing #3276, I noticed that the return value of `ikCallbackFnAdapter()` wasn't propagated through the lambda.
This issue was introduced in #3106.

As the [return value of this `IKCallbackFn` is actually never used](https://github.com/ros-planning/moveit/pull/3276#pullrequestreview-1190740340), it would be even cleaner to drop the boolean result and just rely on the error_code. Having two failure notification mechanisms, obviously results in confusion. Any thoughts?

### DIFF
--- a/moveit_core/constraint_samplers/src/default_constraint_samplers.cpp
+++ b/moveit_core/constraint_samplers/src/default_constraint_samplers.cpp
@@ -576,7 +576,7 @@ bool IKConstraintSampler::sampleHelper(moveit::core::RobotState& state, const mo
     adapted_ik_validity_callback = [this, state_ptr = &state](const geometry_msgs::Pose& /*unused*/,
                                                               const std::vector<double>& joints,
                                                               moveit_msgs::MoveItErrorCodes& error_code) {
-      return samplingIkCallbackFnAdapter(state_ptr, jmg_, group_state_validity_callback_, joints, error_code);
+      samplingIkCallbackFnAdapter(state_ptr, jmg_, group_state_validity_callback_, joints, error_code);
     };
 
   for (unsigned int a = 0; a < max_attempts; ++a)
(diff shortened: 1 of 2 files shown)


AndyZe:
I support the idea of changing `bool ikCallbackFnAdapter` to `void ikCallbackFnAdapter`

AndyZe (approved):
This looks fine to me as it is. But #3276 should be merged, too.

Although I'd still prefer to see this function signature for `ikCallbackFnAdapter()`, to reduce confusion (as you said):

```
void ikCallbackFnAdapter(RobotState* state, const JointModelGroup* group,
                         const GroupStateValidityCallbackFn& constraint, const geometry_msgs::Pose& /*unused*/,
                         const std::vector<double>& ik_sol, moveit_msgs::MoveItErrorCodes& error_code)
```

(return void rather than bool)

rhaschke:
Actually, `IKCallbackFn` already didn't return a value:
https://github.com/ros-planning/moveit/blob/9cc7e8fb0d5b9ceb09d5ba68b524e7a4ab7ca02f/moveit_core/kinematics_base/include/moveit/kinematics_base/kinematics_base.h#L153-L154

Thus, it was just incorrectly used. Fixed via the last commit. #3276 is obsolete with this.

AndyZe (approved):

AndyZe:
I was wondering how that compiled when you changed the return type...

rhaschke:
> I was wondering how that compiled when you changed the return type...

[The return type is not part of the function signature.](https://stackoverflow.com/questions/290038/is-the-return-type-part-of-the-function-signature) So, the compiler doesn't care as long as the return value is not used.

v4hn (approved):
Sorry about the missed return, but yes. Yet another place with broken context. Thanks for repairing it @rhaschke !

PR #3280, opened 2022-11-23
Load URDF file from config folder or regenerate by xacro

BobbyCephy:
Update GAZEBO_URDF_LOAD_ATTRIBUTE to load URDF file from config folder or to regenerate by xacro

### Checklist

### DIFF
--- a/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
@@ -884,6 +884,13 @@ void ConfigurationFilesWidget::changeCheckedState(QListWidgetItem* item)
 
   // Enable/disable file
   gen_files_[index].generate_ = generate;
+
+  // Update GAZEBO_URDF_LOAD_ATTRIBUTE to load URDF file from config folder or to regenerate by xacro
+  if (gen_files_[index].write_on_changes == MoveItConfigData::SIMULATION)
+  {
+    config_data_->save_gazebo_urdf_ = generate;
+    loadGenFiles();
+  }
 }
 
 // ******************************************************************************************
--- a/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
@@ -886,7 +886,7 @@ void ConfigurationFilesWidget::changeCheckedState(QListWidgetItem* item)
   gen_files_[index].generate_ = generate;
 
   // Update GAZEBO_URDF_LOAD_ATTRIBUTE to load URDF file from config folder or to regenerate by xacro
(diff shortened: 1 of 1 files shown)


BobbyCephy:
**gazebo.launch** always loads the urdf from the config folder of a moveit configuration package, even if the urdf is not generated by the MoveIt Setup Assistant. In this case **save_gazebo_urdf_** has to be **false**, so the **GAZEBO_URDF_LOAD_ATTRIBUTE** is set to the **URDF_LOAD_ATTRIBUTE**, to generate the urdf by the xacro script dynamically and the MoveIt Setup Assistant does not have to be used on a changed xacro file, if desired.

moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
```diff
@@ -884,6 +884,13 @@ void ConfigurationFilesWidget::changeCheckedState(QListWidgetItem* item)
 
   // Enable/disable file
   gen_files_[index].generate_ = generate;
+
+  // Update GAZEBO_URDF_LOAD_ATTRIBUTE to load URDF file from config folder or to regenerate by xacro
+  if (gen_files_[index].write_on_changes == MoveItConfigData::SIMULATION)
```
rhaschke, line 889: `write_on_changes` is a bitmask. Hence change to this:
```suggestion
  if (gen_files_[index].write_on_changes & MoveItConfigData::SIMULATION)
```

moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
```diff
@@ -884,6 +884,13 @@ void ConfigurationFilesWidget::changeCheckedState(QListWidgetItem* item)
...
+  {
+    config_data_->save_gazebo_urdf_ = generate;
+    loadGenFiles();
```
rhaschke, line 892: I gues you reload the to-be-generated files because `save_gazebo_urdf_` flag has changed?

rhaschke (changes requested):
It looks like that `config_data_->save_gazebo_urdf_ == True` if and only if 
`config_data_->changes` has the `MoveItConfigData::SIMULATION` bit set. 
Thus, the former should be redundant. I implemented a corresponding change in https://github.com/ros-planning/moveit/pull/3281. Could you please test whether this resolves your issue?

moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
BobbyCephy: Exactly, but actually only the depending files have to be regenerated or generated first, when all flags are certain.

moveit_setup_assistant/src/widgets/configuration_files_widget.cpp
BobbyCephy: Thank you! I was not sure about the best way, to find the related file.

PR #3287, opened 2022-11-30
Add joints_allowed_start_tolerance parameter (#3266)

### LINKED ISSUE #3266: Give custom angular actuators for each joints

Our robot has linear actuators and angular actuators. We have an `allowed_start_tolerance` of 0.05, which is ok for angles, but a bit high for the translations. We pre-compute trajectories, so when there is a bug we are often saved by this value. However, we would like to lower it for the linear actuators, but we cannot without requiring a greater accuracy from our angular actuators.

We could set a parameter, a simple key-value map for each joint, that the trajectory execution manager would check before executing trajectory, and the `allowed_start_tolerance` would acts as a default if the joint doesn't have an assigned start tolerance.

**Describe alternatives you've considered**

More simply, we could split the `allowed_start_tolerance` between linear and angular actuators. This is not ideal but it might be a bit easier to implement.

**Additional question**

I am using MoveIt! one but we may also want it in MoveIt! 2. How should I do in this case? Create an issue in both projects? Create an issue in MoveIt! 2 and reference it on an issue here?

Hugal31:
Allow to specify a per joint allowed start tolerance for the TrajectoryExecutionManager.

### Description

Add the `/move_group/trajectory_execution/joints_allowed_start_tolerance`, a string/double map to specify an allowed start tolerance value for each joint. If the value is not set or set to 0 for a joint, it will fall back to the regular allowed_start_tolerance.

Closes #3266.

### Implementation details

Unlike allowed_start_tolerance, this doesn't uses the dynamic reconfigure server, because it is not possible to use maps with it.

### Checklist

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
+++ b/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
@@ -205,6 +205,12 @@ public:
   /// Set joint-value tolerance for validating trajectory's start point against current robot state
   void setAllowedStartTolerance(double tolerance);
 
+  /// Set the start tolerance for a specific joint. Set to 0 to use the default value allowed_start_tolerance instead.
+  void setAllowedJointStartTolerance(std::string const& joint_name, double tolerance);
+
+  /// Set the start tolerance for a a set of  joint.
+  void setAllowedJointsStartTolerance(std::map<std::string, double> jointsStartTolerance);
+
   /// Enable or disable waiting for trajectory completion
   void setWaitForTrajectoryCompletion(bool flag);
 
@@ -265,6 +271,9 @@ private:
 
   void loadControllerParams();
 
+  double getJointAllowedStartTolerance(std::string const& jointName) const;
+  void updateJointsAllowedStartToleranceEmpty();
+
   moveit::core::RobotModelConstPtr robot_model_;
   planning_scene_monitor::CurrentStateMonitorPtr csm_;
   ros::NodeHandle node_handle_;
@@ -310,6 +319,9 @@ private:
   std::map<std::string, double> controller_allowed_goal_duration_margin_;
 
   double allowed_start_tolerance_;  // joint tolerance for validate(): radians for revolute joints
+  // joint tolerance per joint, overrides allowed_start_tolerance_.
+  std::map<std::string, double> joints_allowed_start_tolerance_;
+  bool joints_allowed_start_tolerance_empty_;  // True if joints_allowed_start_tolerance_ is empty or contains only 0
   double execution_velocity_scaling_;
   bool wait_for_trajectory_completion_;
 };
--- a/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
+++ b/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
@@ -313,7 +313,7 @@ private:
   std::map<std::string, double> controller_allowed_goal_duration_margin_;
 
   double allowed_start_tolerance_;  // joint tolerance for validate(): radians for revolute joints
-  // joint tolerance per joint, overrides allowed_start_tolerance_.
+  // tolerance per joint, overrides global allowed_start_tolerance_.
   std::map<std::string, double> joints_allowed_start_tolerance_;
   double execution_velocity_scaling_;
   bool wait_for_trajectory_completion_;
--- a/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
+++ b/moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
@@ -205,12 +205,6 @@ public:
   /// Set joint-value tolerance for validating trajectory's start point against current robot state
   void setAllowedStartTolerance(double tolerance);
 
-  /// Set the start tolerance for a specific joint. Set to 0 to use the default value allowed_start_tolerance instead.
-  void setAllowedJointStartTolerance(std::string const& joint_name, double tolerance);
-
-  /// Set the start tolerance for a a set of  joint.
-  void setAllowedJointsStartTolerance(std::map<std::string, double> jointsStartTolerance);
-
   /// Enable or disable waiting for trajectory completion
   void setWaitForTrajectoryCompletion(bool flag);
 
@@ -272,12 +266,13 @@ private:
   void loadControllerParams();
 
   double getJointAllowedStartTolerance(std::string const& jointName) const;
-  void updateJointsAllowedStartToleranceEmpty();
+  void updateJointsAllowedStartTolerance();
 
   moveit::core::RobotModelConstPtr robot_model_;
   planning_scene_monitor::CurrentStateMonitorPtr csm_;
   ros::NodeHandle node_handle_;
   ros::NodeHandle root_node_handle_;
+  ros::NodeHandle trajectory_execution_node_handle_;
   ros::Subscriber event_topic_subscriber_;
 
   std::map<std::string, ControllerInformation> known_controllers_;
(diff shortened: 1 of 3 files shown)


moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
```diff
@@ -310,6 +319,9 @@ class TrajectoryExecutionManager
   std::map<std::string, double> controller_allowed_goal_duration_margin_;
 
   double allowed_start_tolerance_;  // joint tolerance for validate(): radians for revolute joints
+  // joint tolerance per joint, overrides allowed_start_tolerance_.
+  std::map<std::string, double> joints_allowed_start_tolerance_;
+  bool joints_allowed_start_tolerance_empty_;  // True if joints_allowed_start_tolerance_ is empty or contains only 0
```
rhaschke, line 324: There is no need to introduce this boolean flag. Just use `joints_allowed_start_tolerance_.empty()`.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -176,6 +180,10 @@ void TrajectoryExecutionManager::initialize()
 
   reconfigure_impl_ = new DynamicReconfigureImpl(this);
 
+  ros::NodeHandle private_node_handle("~/trajectory_execution");
+  private_node_handle.getParam("joints_allowed_start_tolerance", joints_allowed_start_tolerance_);
+  updateJointsAllowedStartToleranceEmpty();
```
rhaschke, line 185: As we can't modify those parameters via `dynamic reconfigure`, I suggest loading those values again, each time _before_ a trajectory shall be executed. Actually, you could/should drop the C++ API to set those parameters then, because they would get overwritten.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1547,4 +1570,20 @@ void TrajectoryExecutionManager::loadControllerParams()
...
+  return start_tolerance_it != joints_allowed_start_tolerance_.end() && start_tolerance_it->second > 0 ?
+             start_tolerance_it->second :
+             allowed_start_tolerance_;
+}
+
+void TrajectoryExecutionManager::updateJointsAllowedStartToleranceEmpty()
+{
+  joints_allowed_start_tolerance_empty_ =
+      std::all_of(joints_allowed_start_tolerance_.begin(), joints_allowed_start_tolerance_.end(),
+                  [](auto const& pair) { return pair.second <= 0; });
```
rhaschke, line 1586: I missed to commit this comment:
Just reject/remove invalid tolerance values once when setting/loading them instead of using this complicated mechanism.

Hugal31:
Hi,

Can this be reviewed again?

moveit_ros/planning/trajectory_execution_manager/include/moveit/trajectory_execution_manager/trajectory_execution_manager.h
```diff
@@ -265,10 +265,14 @@ class TrajectoryExecutionManager
...
   void loadControllerParams();
 
+  double getJointAllowedStartTolerance(std::string const& jointName) const;
+  void updateJointsAllowedStartTolerance();
+
   moveit::core::RobotModelConstPtr robot_model_;
   planning_scene_monitor::CurrentStateMonitorPtr csm_;
   ros::NodeHandle node_handle_;
   ros::NodeHandle root_node_handle_;
+  ros::NodeHandle trajectory_execution_node_handle_;
```
rhaschke, line 275: This extra node handle is only needed in `updateJointsAllowedStartTolerance()`.
As creating node handles is not costly, I suggest dropping this variable as a member.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1554,4 +1566,26 @@ void TrajectoryExecutionManager::loadControllerParams()
     }
   }
 }
+
+double TrajectoryExecutionManager::getJointAllowedStartTolerance(std::string const& jointName) const
+{
+  auto start_tolerance_it = joints_allowed_start_tolerance_.find(jointName);
+  return start_tolerance_it != joints_allowed_start_tolerance_.end() && start_tolerance_it->second > 0 ?
+             start_tolerance_it->second :
+             allowed_start_tolerance_;
```
rhaschke, line 1575: The value `allowed_start_tolerance_ = 0` had a special meaning: don't validate all joints. I suggest keeping this special meaning for individual joints: Zero should mean to not validate a particular joint. This requires some changes throughout this PR.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1554,4 +1566,26 @@ void TrajectoryExecutionManager::loadControllerParams()
...
+}
+
+void TrajectoryExecutionManager::updateJointsAllowedStartTolerance()
+{
+  trajectory_execution_node_handle_.getParam("joints_allowed_start_tolerance", joints_allowed_start_tolerance_);
+
+  for (auto it = joints_allowed_start_tolerance_.begin(); it != joints_allowed_start_tolerance_.end();)
+  {
+    if (it->second <= 0)
```
rhaschke, line 1584: Here zero is explicitly erased.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -1554,4 +1566,26 @@ void TrajectoryExecutionManager::loadControllerParams()
...
+{
+  auto start_tolerance_it = joints_allowed_start_tolerance_.find(jointName);
+  return start_tolerance_it != joints_allowed_start_tolerance_.end() && start_tolerance_it->second > 0 ?
+             start_tolerance_it->second :
+             allowed_start_tolerance_;
+}
+
+void TrajectoryExecutionManager::updateJointsAllowedStartTolerance()
+{
+  trajectory_execution_node_handle_.getParam("joints_allowed_start_tolerance", joints_allowed_start_tolerance_);
```
rhaschke, line 1580: `ros::NodeHandle trajectory_execution_node_handle_(node_handle_, "trajectory_execution")`

rhaschke (commented):
Thanks for this improvement. I have a few remarks though. See below.

moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
Hugal31: I find `node_handle_.param("trajectory_execution/joints_allowed_start_tolerance", ...)` less verbose.

rhaschke (approved):
I added 3 commits to address some open issues. Please check for the two suggestions below. Otherwise, I approve.

moveit_ros/planning/trajectory_execution_manager/test/test_execution_manager.cpp
```diff
@@ -60,15 +60,21 @@ class MoveItCppTest : public ::testing::Test
...
     traj2.joint_trajectory.joint_names.push_back("panda_joint2");
     traj2.joint_trajectory.points[0].positions.push_back(1.0);
+    traj2.joint_trajectory.points[1].positions.push_back(1.5);
     traj2.multi_dof_joint_trajectory.joint_names.push_back("panda_joint3");
-    traj2.multi_dof_joint_trajectory.points.resize(1);
+    traj2.multi_dof_joint_trajectory.points.resize(2);
     traj2.multi_dof_joint_trajectory.points[0].transforms.resize(1);
+    traj2.multi_dof_joint_trajectory.points[1].transforms.resize(1);
+
+    ros::param::del("~/trajectory_execution/joints_allowed_start_tolerance");
```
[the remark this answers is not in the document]
Hugal31: I want to be sure the tests don't start with a value from another test. For example if `RejectTooFarFromStart` were to run after `DoNotValidateJointStartToleranceZero` without reseting this value, it would fail.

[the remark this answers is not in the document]
Hugal31: `setAllowedStartTolerance` doesn't change `joints_allowed_start_tolerance`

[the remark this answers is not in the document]
rhaschke: OK. I didn't notice, that you delete the map.

PR #3300, opened 2022-12-21
Debug Ruckig tests

AndyZe:
In order to get this test passing reliably, increase the allowable trajectory duration.


rhaschke (approved):
Great. Seems to fix the test. Thanks!

PR #3312, opened 2023-01-23
fix ruckig dependency with catkin

### LINKED ISSUE #3310: cannot compile from source
### Description

Overview of your issue here.
Compile from source, ruckig is also from source master branch.
moveit_core compile success, moveit_ros_planning failed.

```
Errors     << moveit_ros_planning:make src/logs/moveit_ros_planning/build.make.010.log                                            
In file included from /src/moveit/moveit/moveit_ros/planning/planning_request_adapter_plugins/src/add_ruckig_traj_smoothing.cpp:37:
/src/moveit/moveit/moveit_core/trajectory_processing/include/moveit/trajectory_processing/ruckig_traj_smoothing.h:41:10: fatal error: ruckig/ruckig.hpp: No such file or directory
   41 | #include <ruckig/ruckig.hpp>
      |          ^~~~~~~~~~~~~~~~~~~
```

### Your environment
* ROS Distro: noetic
* OS Version: e.g. Ubuntu 22.04
* Source master branch

v4hn:
catkin was not made for cmake target dependencies.

We do the same with fcl right above, but when porting/merging ruckig nobody cared...

FTR: ruckig does not have any `INTERFACE_LINK_LIBRARIES` and adding them for good measures in case it changes sets the variable to `ruckig_INTERFACE_LINK_LIBRARIES-NOT-FOUND` (breaking the build).

Fixes #3310 .

### DIFF
--- a/moveit_core/CMakeLists.txt
+++ b/moveit_core/CMakeLists.txt
@@ -51,8 +51,14 @@ else()
 endif()
 
 find_package(octomap REQUIRED)
+
 find_package(ruckig REQUIRED)
+# work around catkin_package not fetching the interface includes from the target
+# to forward to downstream dependencies. The includes do not need to be added
+# in include_directories below because the target is correctly imported here.
+get_target_property(ruckig_INCLUDE_DIRS ruckig::ruckig INTERFACE_INCLUDE_DIRECTORIES)
 set(ruckig_LIBRARIES "ruckig::ruckig")
+
 find_package(urdfdom REQUIRED)
 find_package(urdfdom_headers REQUIRED)
 


simonschmeisser:
Shouldn't it be sufficient to add add `ruckig::ruckig` here https://github.com/ros-planning/moveit/blob/7e02e7683572a8beaa70c0501f6c6929ddecacfa/moveit_ros/planning/planning_request_adapter_plugins/CMakeLists.txt#L24

or to `CATKIN_LIBRARIES`

moveit_core/CMakeLists.txt
```diff
@@ -51,8 +51,12 @@ else()
 endif()
 
 find_package(octomap REQUIRED)
+
 find_package(ruckig REQUIRED)
-set(ruckig_LIBRARIES "ruckig::ruckig")
+get_target_property(ruckig_LOCATION ruckig::ruckig LOCATION)
+get_target_property(ruckig_INCLUDE_DIRS ruckig::ruckig INTERFACE_INCLUDE_DIRECTORIES)
```
rhaschke, line 57: I think this is the only line that needs to be added to correctly set include paths.
In Noetic the includes are just found via `/opt/ros/*/include`.

rhaschke (commented):
Why don't you propose these changes upstream in https://github.com/pantor/ruckig?
They provide a package.xml and thus should also provide a ROS-compatible cmake config.

gavanderhoorn:
> Why don't you propose these changes upstream in https://github.com/pantor/ruckig? They provide a package.xml and thus should also provide a ROS-compatible cmake config.

also perhaps because @vatanaksoytezer is helping release Ruckig.

v4hn:
> Why don't you propose these changes upstream in https://github.com/pantor/ruckig?

From my perspective there is no bug in their cmake config.
There is a systematic shortcoming in `catkin_package` though and if anything at all we need to add support for targets in there.
Feel free to do so. :dog2:

moveit_core/CMakeLists.txt
v4hn: Technically that makes it compile, but it doesn't resolve the `ruckig::ruckig` target which will still be added to the `DEPENDS`, resulting in potential errors in downstream packages.

That might still happen because ruckig requires c++17, but at least I propose to make it explicit here.

rhaschke: > Technically that makes it compile, but it doesn't resolve the `ruckig::ruckig` target which will still be added to the `DEPENDS`, resulting in potential errors in downstream packages.

cmake correctly resolves library targets to the corresponding library via [`catkin_replace_imported_library_targets`](https://github.com/ros/catkin/blob/bb2299bc878a638d5a81226c21cb918359e773fc/cmake/catkin_package.cmake)
However, you are right that includes are not yet resolved.

rhaschke:
> From my perspective there is no bug in their cmake config.

It doesn't work as is for ROS1... That's a bug in my opinion.

> There is a systematic shortcoming in `catkin_package` though and if anything at all we need to add support for targets in there.

I agree, that resolving that issue in `catkin_package` would be even nicer. However, all my PRs I have filed in the last year were not even commented on. I don't see any chance to merge something into that repo anymore.

v4hn:
That said, the proposed patch for handling the target is a known and established workaround to handle targets in catkin_package.
The alternative approach to add an extra cmake file is possible too, but I like it much less than the current approach.

moveit_core/CMakeLists.txt
v4hn: Wow, now that's a(n incomplete/broken) feature I did not know. :+1: 

The c++17 declaration is also imported in `moveit_core`, I guess.
In [`moveit_ros_planning` it is set by the library user](https://github.com/ros-planning/moveit/blob/6ac1b9fcaee8f676d2f2758805d5f4d76415aa1d/moveit_ros/planning/planning_request_adapter_plugins/CMakeLists.txt#L20) which should not be done :disappointed: .
I will move the version bump to a different PR though.

lianghongzhuo:
I can confirm that this pr fixed my issues. Thanks a lot!

moveit_core/CMakeLists.txt
```diff
@@ -51,8 +51,14 @@ else()
 endif()
 
 find_package(octomap REQUIRED)
+
 find_package(ruckig REQUIRED)
+# work around catkin_package not fetching the interface includes from the target
+# to forward to downstream dependencies. These includes do not have to be added
+# globally in this package because the target is correctly imported here.
```
rhaschke, line 58: I don't understand what you mean by "add globally" here. What about this simplification:
```suggestion
# Set ruckig_INCLUDE_DIRS and ruckig_LIBRARIES for downstream packages
```

rhaschke (approved):
Otherwise, I approve.

moveit_core/CMakeLists.txt
v4hn: I changed the string a bit. I wanted a note in the file explaining why `include_directories` below lists all other external packages, but not this one.

PR #3313, opened 2023-01-23
Switch master build to C++17

v4hn:
This came up again [looking at the ruckig integration](https://github.com/ros-planning/moveit/pull/3312#discussion_r1084075557) which sneaked in C++17 for a very specific header we export.

I don't see any particular reason why we shouldn't switch for the whole framework now. 
Any contrary views?

E.g., @gavanderhoorn ? :)


rhaschke (approved):
Fine with me.

AndyZe (approved):
LGTM. I verified CMAKE_CXX_STANDARD is not used anywhere else that might conflict

PR #3315, opened 2023-01-24
Fix doxygen

rhaschke:
moveit_tutorials is broken due to missing files, which should be generated by rosdoc/doxygen:
https://github.com/ros-planning/moveit_tutorials/actions/runs/3966977821/jobs/6798414067

Looking into the output of the corresponding [doc job](https://build.ros.org/job/Ndoc__moveit__ubuntu_focal_amd64/102/consoleFull), I found hundreds of doxygen warnings.
This PR addresses many of them (240 out of 340)! Then I gave up after 2 hours of fixing.
Anybody continuing is highly welcome!

The actual culprit was the `.scene` file format description, which I added recently in `planning_scene.h`: it was terminated with a misspelled `\endvarbatim` command.

To avoid such issues in the future, we should add a `rosdoc` job to CI and fail on corresponding warnings. Obviously, this would require fixing all existing warnings first...

### DIFF
--- a/moveit_core/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler_manager.h
+++ b/moveit_core/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler_manager.h
@@ -124,15 +124,14 @@ public:
    *subgroup IKSolvers, the function will attempt to generate a sampler from the various subgroup solvers.
    *   - It will attempt to determine which constraints act on the IK link for the sub-group IK solvers, and attempts to
    *create ConstraintSampler functions by recursively calling \ref selectDefaultSampler for the sub-group.
-   *   - If any samplers are valid, it adds them to a vector of type \ref ConstraintSamplerPtr.
+   *   - If any samplers are valid, it adds them to a vector of type ConstraintSamplerPtr.
    *   - Once it has iterated through each sub-group, if any samplers are valid, they are returned in a
    *UnionConstraintSampler, along with a JointConstraintSampler if one exists.
    * @param scene The planning scene that will be used to create the ConstraintSampler
    * @param group_name The group name for which to create a sampler
    * @param constr The set of constraints for which to create a sampler
    *
-   * @return A valid \ref ConstraintSamplerPtr if one could be allocated, and otherwise an empty \ref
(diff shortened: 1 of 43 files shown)


moveit_planners/pilz_industrial_motion_planner/include/pilz_industrial_motion_planner/joint_limits_validator.h
```diff
@@ -109,12 +105,7 @@ class ValidationException : public std::runtime_error
   }
 };
 
-/**
- * @class ValidationJointMissingException
- * @brief Thrown the limits for a joint are defined in the urdf but not on the
- * parameter server (loaded from yaml)
- *
- */
+/** Thrown the limits for a joint are defined in the urdf but not on the parameter server (loaded from yaml) */
```
AndyZe, line 108:
```suggestion
/** Thrown when the limits for a joint are defined in the urdf but not on the parameter server (loaded from yaml) */
```

moveit_core/constraint_samplers/include/moveit/constraint_samplers/constraint_sampler_manager.h
```diff
@@ -124,15 +124,14 @@ class ConstraintSamplerManager
...
+   *   - If any samplers are valid, it adds them to a vector of type ConstraintSamplerPtr.
    *   - Once it has iterated through each sub-group, if any samplers are valid, they are returned in a
    *UnionConstraintSampler, along with a JointConstraintSampler if one exists.
    * @param scene The planning scene that will be used to create the ConstraintSampler
    * @param group_name The group name for which to create a sampler
    * @param constr The set of constraints for which to create a sampler
    *
-   * @return A valid \ref ConstraintSamplerPtr if one could be allocated, and otherwise an empty \ref
-   *ConstraintSamplerPtr
+   * @return A valid ConstraintSamplerPtr if one could be allocated, and otherwise an empty ConstraintSamplerPtr.
```
AndyZe, line 134:
```suggestion
   * @return A valid ConstraintSamplerPtr if one could be allocated, otherwise an empty ConstraintSamplerPtr.
```

AndyZe (approved):

PR #3316, opened 2023-01-24
runRuckigInBatches: drop num_waypoints parameters

rhaschke:
While working on #3315, I stumbled over this weird API of `runRuckigInBatches()`.
@simonschmeisser, why does it require a `num_waypoints` parameter? This number can be determined via `trajectory.getWayPointCount()`, can't it?


AndyZe (approved):
This was probably my work and I have to agree, your way is simpler.  :+1:

PR #3317, opened 2023-01-25
Fix public member naming

rhaschke:
As suggested by @v4hn in https://github.com/ros-planning/moveit/pull/3244#issuecomment-1403420139: Don't use `_` suffix for public struct/class members.

This also changed the casing of static, but non-const variables to upper case:
- https://github.com/ros-planning/moveit/pull/3317/files#diff-198a52470b3364f92b6d1f1d5e4274bc27956f08d172b676845933a212a652ddL261-L263
- https://github.com/ros-planning/moveit/pull/3317/files#diff-782fd4e1495a3e622fcbd5e280e28e84d3e8a3bde755e52acc56033b042c2eeaL299-L304

However, I would prefer lowercase names for those and only use uppercase for static const variables/members ([`GlobalConstantCase`](https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html#cmdoption-arg-globalconstantcase) and [`StaticConstantCase`](https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html#cmdoption-arg-staticconstantcase)). What do you think, @v4hn?


v4hn:
Thanks for looking into it. :)
My goal was not necessarily to enforce the convention for all of MoveIt at the cost of API compatibility, but just to adjust the single item at hand if we touch it anyway.
But you are right in that we would want to make clang-tidy happy if we change it.
Maybe let's see the full patch once CI passed.

rhaschke:
Interestingly clang-tidy didn't fix the [code snippet](https://github.com/ros-planning/moveit/blob/a266089ebe4a054e875aafebd477a140386a285e/moveit_core/planning_interface/include/moveit/planning_interface/planning_response.h#L48-L57) originally giving raise to this PR?!

rhaschke:
I pushed some initial changes to manually remove the trailing underscore suffix from `struct MotionPlanResponse`.
However, this is used literally everywhere in the code and vscode did not manage to correctly rename all member instances. I don't have time to fix all those issues manually...
I don't see a value in introducing those API-breaking changes and thus suggest to close this. @v4hn, please comment.

v4hn:
> I don't see a value in introducing those API-breaking changes and thus suggest to close this.

Looking through the changes, many lines just improve internal variable names, these would be fine independent of the response struct. The response field **is** a big API break though I strictly prefer the patched names without underscore.
That's why I referred to the current state as "legacy" use in MoveIt and I'm happy to see it cleaned up in MoveIt2. :heart: 
Let's leave it at that.

rhaschke:
> Looking through the changes, many lines just improve internal variable names, these would be fine independent of the response struct.

Shall we apply this PR without the last commit (changing `MotionPlanResponse`) then?

rhaschke:
clang-tidy didn't attempt to fix the offender: https://github.com/ros-planning/moveit/pull/3317#issuecomment-1404342530
So, no need to introduce ignore sections.

PR #3326, opened 2023-01-31
Simplify servo config + Reusable launch scripts

rhaschke:
When trying to generalize moveit_servo's launch scripts to handle user-specific config files, I noted a weird mechanism to load these parameters: 
First, a private parameter `~parameter_ns` was declared and its value was actually used as a namespace to load all parameters into. Instead of this error-prone and indirect scheme, I suggest loading these parameters directly into the private namespace of the node. I don't see a need to share these parameters across different server instances.

If one really wants to load those parameters somewhere else first, one can simply start the node with the corresponding namespace as its name.

The original purpose of this PR was to add an argument `config` to specify a .yaml file with the desired parameters (instead of locking it to `ur_simulated_config.yaml`). In this fashion, the launch scripts become reusable.

### DIFF
--- a/moveit_ros/moveit_servo/launch/cpp_interface_example.launch
+++ b/moveit_ros/moveit_servo/launch/cpp_interface_example.launch
@@ -1,9 +1,0 @@
-<launch>
-  <!-- Launch an example that sends commands via C++ API. -->
-
-  <node name="servo_server" pkg="moveit_servo" type="cpp_interface_example" output="screen" >
-    <param name="parameter_ns" type="string" value="optional_parameter_namespace" />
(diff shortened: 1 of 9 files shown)


AndyZe (approved):
> If one really wants to load those parameters somewhere else first, one can simply start the node with the corresponding namespace as its name.

IIRC the original purpose of the `parameter_ns` was for dual-arm systems. Do you see any problem with this new way for dual-arm systems? I guess it would be fine.

rhaschke:
I guess for dual-arm systems you need to run two separate servers - one for each arm. Thus, loading the appropriate parameters into the private namespaces of the individual servers cleanly separates them.
By the way, I didn't find any documentation about the configuration mechanism you implemented initially.

PR #3334, opened 2023-02-06
[ROS-O] use system pybind11

v4hn:
Ubuntu focal supports 2.4.3, which hopefully suffices here.
(ROS noetic bundles 2.5.0).

Ubuntu jammy+ provide newer versions compatible with system python.
Debian testing has Python3.11 now which breaks with pybind11 2.5.

I added a conditional for melodic to rely on the package instead, let's see whether it works or needs some tweaks.


rhaschke:
The ROS package provides a new utility macro `pybind_add_module` that automatically configures the correct target location and links catkin libs:
https://github.com/ipab-slmc/pybind11_catkin/blob/2.5.0/cmake/pybind11_catkin.cmake.in#L18-L19

Doing that manually, we could switch to system pybind11 instead.

v4hn:
turns out @jspricke had a similar patch lying around for a year already. :-)
I (hope I) reverted the switch to Python3 from his patch.

It feels insane to continue support for python2 by now...

v4hn:
two days of frustrated attempts to please CI later: CI passed
@simonschmeisser @rhaschke

moveit_core/python/tools/CMakeLists.txt
```diff
@@ -1,13 +1,16 @@
...
   src/pybind_rosmsg_typecasters.cpp
 )
 add_dependencies(${MOVEIT_LIB_NAME} ${catkin_EXPORTED_TARGETS})
+target_include_directories(${MOVEIT_LIB_NAME} SYSTEM PRIVATE ${PYTHON_INCLUDE_DIRS} ${pybind11_INCLUDE_DIRS})
 target_link_libraries(${MOVEIT_LIB_NAME} ${catkin_LIBRARIES} ${PYTHON_LIBRARIES})
 set_target_properties(${MOVEIT_LIB_NAME} PROPERTIES VERSION "${${PROJECT_NAME}_VERSION}")
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+  # avoid "no matching function for call to 'operator delete'" - https://github.com/pybind/pybind11/issues/1604
+  target_compile_options(${MOVEIT_LIB_NAME} PUBLIC -fsized-deallocation)
+endif()
```
rhaschke, line 13: This issue was fixed in https://github.com/pybind/pybind11/commit/759221f5c56939f59d8f342a41f8e2d2cacbc8cf, i.e. version 2.5.0.

rhaschke (commented):
Switching to system pybind11 has several drawbacks:
- We loose the convenience macro `pybind_add_module`
- There is more code required to handle various corner cases.

Wouldn't it be possible to just release instead a newer version of pybind11 as pybind11_catkin?

simonschmeisser:
> Switching to system pybind11 has several drawbacks:
> 
>     * We loose the convenience macro `pybind_add_module`
> 
>     * There is more code required to handle various corner cases.

Would this be reduced if limiting it to post-noetic "releases"?
> 
> 
> Wouldn't it be possible to just release instead a newer version of pybind11 as pybind11_catkin?

We would either need to enable internet access in the sbuild chroot or switch pybind11_catkin to git submodules, probably the submodules way is preferable

rhaschke:
> Would this be reduced if limiting it to post-noetic "releases"?

Sure, everything >= 2.5.0 has the sized-deallocation issue resolved. Still, the convenience macro from pybind11_catkin is a nice to have.

> We would either need to enable internet access in the sbuild chroot or switch pybind11_catkin to git submodules.

The latter seems to be reasonable, doesn't it? @wxmerkt?

wxmerkt:
Hi @rhaschke, yes, happy to switch pybind11_catkin to use submodules - and also to release newer versions of pybind11 (pending getting access to the repository again). I'll get on this next week

moveit_core/python/tools/CMakeLists.txt
v4hn: I found it necessary during development of the patch. I agree though that this shouldn't be necessary. I reverted the patch now that everything passes. Let's see.

v4hn: Please see CI for the latest commit, the error persists without the clang override.

v4hn:
> We loose the convenience macro pybind_add_module

As far as I can see this is the *only* thing pybind11_catkin provides that we need.
That is an awful reason for me to roll a custom version of a system package (that will age independent of the system over time).

> There is more code required to handle various corner cases.

The main overhead is the distinction of whether we are in melodic or not.
This will go away soon anyway. [The cleanup in tools/](https://github.com/ros-planning/moveit/pull/3334/files#diff-6481f3a4243a0d2edab67ae02a6eb935d5db77591a26d7c3a730cfcb5458f0b0R7) should be added either way.

@wxmerkt pybind11_catkin should not bundle its own version of pybind11 at all but utilize the system library and provide the convenience cmake method. At this point there seems to be no need to bundle a more current version than the target system and anything you tag will just age again.

moveit_core/python/tools/CMakeLists.txt
rhaschke: Hence, I was asking to keep `pybind11_catkin`, which provides version 2.5.0 for Noetic right now.

rhaschke:
> At this point there seems to be no need to bundle a more current version than the target system

For Melodic and Noetic we _do_ need a more recent version than the system-provided one.

v4hn:
> For Melodic and Noetic we do need a more recent version than the system-provided one.

I agree about melodic, we use API that is not available in Ubuntu 18.04.
But what is your reason for noetic? Is there some runtime functionality in 2.5 over 2.4.3 that moveit needs?
For me, [a 1-line conditional in a build file](https://github.com/ros-planning/moveit/pull/3334/files#diff-6481f3a4243a0d2edab67ae02a6eb935d5db77591a26d7c3a730cfcb5458f0b0R10-R13) is no sufficient reason to roll a custom library when the system package is almost the same.

Additionally you argue with the one cmake macro for pybind11_catkin. But currently, using that one cmake macro pulls in the whole pinned pybind11 API instead of providing a macro on top of the system package.

Let me elaborate on my main reason to push for using system packages in development branches where possible (in any project): I strongly prefer having the upstream development branch compile with system libraries by default and overriding these libraries should be decisions by the party *building* the stack, where interesting or required (e.g. compiling MoveIt with OMPL in workspace). That way the development targets *any* current system by default and anyone can compile on newer systems against newer dependencies. Sure, upstream API might have changed, but that's the anticipated source of errors. There is no reason to compile moveit against pybind11 2.5 in Ubuntu jammy which provides 2.9. Eventually it breaks for no reason in current conda environments and debian testing, because python 3.11 [requires 2.10.3](https://github.com/ipab-slmc/pybind11_catkin/pull/21). These kind of patches are a big reason why package distributions have to apply their own custom patch sets to get packages into their targeted environments.

Pinning libraries is important on branches targeting specific older platforms, but not the main development branch.

That being said, [ROS has a compatible way to pin a system library to a different version in a specific ROS distribution](http://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty) and it uses neither ExternalProject, nor submodules. It involves overriding the system package with the new version in a specific ROS release. This is precisely for the reason that newer systems will not need that package anymore. If we really need pybind11 2.5 in noetic (as regressing to an older version is not nice for users), `pybind11_catkin` should be a thin wrapper around a third-party released package in noetic.

moveit_core/CMakeLists.txt
```diff
@@ -250,9 +259,9 @@ pybind_add_module(pymoveit_core
     planning_scene/src/pyplanning_scene.cpp
     kinematic_constraints/src/pykinematic_constraint.cpp
 )
-target_include_directories(pymoveit_core SYSTEM PRIVATE ${catkin_INCLUDE_DIRS})
-target_link_libraries(pymoveit_core PRIVATE ${pymoveit_libs} ${catkin_LIBRARIES})
-
+target_include_directories(pymoveit_core SYSTEM PRIVATE ${PYTHON_INCLUDE_DIRS} ${pybind11_INCLUDE_DIRS} ${catkin_INCLUDE_DIRS})
+target_link_libraries(pymoveit_core PRIVATE ${pymoveit_libs} ${PYTHON_LIBRARIES} ${catkin_LIBRARIES})
```
simonschmeisser, line 263:
```suggestion
target_link_libraries(pymoveit_core PRIVATE ${pymoveit_libs} pybind11::module ${catkin_LIBRARIES})
```
Would it work to use modern cmake targets here to reduce the diff? There are multiple exported targets to choose from

https://github.com/pybind/pybind11/blob/master/tools/pybind11Config.cmake.in#L37

moveit_core/CMakeLists.txt
```diff
@@ -230,6 +229,16 @@ add_subdirectory(collision_distance_field)
 add_subdirectory(kinematics_metrics)
 add_subdirectory(dynamics_solver)
 
+find_package(pybind11 2.4.3 QUIET)
```
simonschmeisser, line 232:
```suggestion
find_package(pybind11 2.5.0 QUIET)
```
to avoid (known and unknown) regressions

moveit_core/package.xml
```diff
@@ -39,7 +39,8 @@
   <depend>moveit_msgs</depend>
   <depend>octomap</depend>
   <depend>octomap_msgs</depend>
-  <depend>pybind11_catkin</depend>
+  <depend condition="$ROS_DISTRO == melodic">pybind11_catkin</depend>
+  <depend condition="$ROS_DISTRO != melodic">pybind11-dev</depend>
```
simonschmeisser, line 43:
```suggestion
  <depend condition="($ROS_DISTRO == melodic) or ($ROS_DISTRO == noetic)">pybind11-dev</depend>
  <depend condition="($ROS_DISTRO != melodic) or ($ROS_DISTRO != noetic)">pybind11-dev</depend>
```

I vote for sticking with pybind11_catkin in noetic

v4hn: Once `pybind11_catkin` is independent on top of a system `pybind11-dev` package this can be simplified again.

simonschmeisser:
> That being said, [ROS has a compatible way to pin a system library to a different version in a specific ROS distribution](http://wiki.ros.org/bloom/Tutorials/ReleaseThirdParty) and it uses neither ExternalProject, nor submodules. It involves overriding the system package with the new version in a specific ROS release. This is precisely for the reason that newer systems will not need that package anymore. If we really need pybind11 2.5 in noetic (as regressing to an older version is not nice for users), `pybind11_catkin` should be a thin wrapper around a third-party released package in noetic.

From private communication I now understand that this would mean:

- release `pybind11 2.5.0`  (or newer?) as `ros-noetic-pybind11` (and melodic)
- trim down `pybind11_catkin` to just the cmake macro and the dependency on `pybind11` which evaluates to `ros-noetic-pybind11` or system `pybind11-dev`

That sounds like a clean approach, I like it.

rhaschke:
> Is there some runtime functionality in 2.5 over 2.4.3 that MoveIt needs?

@v4hn, only version 2.5 has fixed `sized-deallocation`. 
@simonschmeisser, I like your proposal very much. @wxmerkt, what do you think? Do you need help to realize this?

wxmerkt:
For context, we needed at least 2.5.0 for its handling of Eigen (from what I recall returning / passing by reference) and hence released the same packaging as for Melodic into Noetic.

I have a vague recollection of there having been a discussion of vendoring a ros-noetic-pybind11 at the time but it being rejected based on a system package available on some target platforms (2.4.3 for focal) with the suggestion being to have the bleeding-edge packaged version in a different package name and different include directories etc. (hence the pybind11_catkin/pybind11 directory structure along with the cfg_extras cmake to make the paths behave as if it's a system-packaged version).

I've now updated (and thanks to Robert it's fixed and re-released) to version 2.10.3 for Noetic as requested above. A key reasoning behind pybind11_catkin was that it is isolated from the system package, i.e. does not conflict and allows to have a common, up to date version if desired by users. Users can decide to simply rely on the system-package instead or use pybind11_catkin if requiring more recent features.

I am uncertain whether I'd support a switch of how the package operates mid-way through a distribution's lifecycle in contrast to targeting the above-mentioned approach for a newer distro (that point being moot since pybind11 is [vendored in ROS2](https://github.com/ros2/pybind11_vendor) and there is no official ROS1 successor outside of the community-supported ROS-O). If preferred, happy to have a ros-o branch or logic switch in pybind11_catkin that uses the system package if recent enough (?). What are your thoughts?

v4hn:
Thanks for the reply @wxmerkt !

> or logic switch in pybind11_catkin that uses the system package if recent enough

I believe that would be the best solution going forward as Robert wants to keep using the additional functionality in pybind11_catkin (the cmake macro). Can I provide any further steps to get this implemented?
The minor downside of the approach is that you will either have to rely on `ExternalProject` still if the system version is older or you will include a submodule (arguably the cleaner approach because building can be done without network access) that will be pulled in on newer systems even when the build will ignore it.

> > @rhaschke Is there some runtime functionality in 2.5 over 2.4.3 that MoveIt needs?
> 
> only version 2.5 has fixed sized-deallocation.

From what I have seen so far that statement is actually wrong . It's not functionality, but [a simple bugfix](https://github.com/pybind/pybind11/commit/759221) that can be worked around with one line of cmake inside MoveIt. The fix applied in 2.5 actually (correctly) *disables* the overloads that are available on 20.04 also with clang if we compile with `-fsized-deallocation`.

> we needed at least 2.5.0 for its handling of Eigen (from what I recall returning / passing by reference)

Do you have a link for this discussion? I don't see it [in the changelog](https://pybind11.readthedocs.io/en/stable/changelog.html#v2-5-0-mar-31-2020) explicitly, but it would be a tangible reason not to use 2.4.3. (Who is "we"?)

> [...] with the suggestion being to have the bleeding-edge packaged version in a different package name

That idea was flawed:
1. pinned bleeding-edge ages (as it did here and it will obviously age again).
2. cmake version requirements fulfill exactly the use case to allow different versions of a library with the same name on the system.

> A key reasoning behind pybind11_catkin was that it is isolated from the system package, i.e. does not conflict and allows to have a common, up to date version if desired by users. Users can decide to simply rely on the system-package instead or use pybind11_catkin if requiring more recent features.

I believe you could have done the same without changing the name and no user repository would have to change build scripts to go back to upstream at some point.

> I've now updated (and thanks to Robert it's fixed and re-released) to version 2.10.3 for Noetic as requested above

Thank you. I don't see why you would still do that "mid-way through a distribution's lifecycle", but let's hope there are no critical backward-incompatible changes. At least I believe you have dropped python2 support from pybind11_catkin's master branch, thus breaking `rosinstall_generator --rosdistro melodic pybind11_catkin --upstream-development`, but I doubt anyone would still expect that to work by now.

simonschmeisser:
I took the liberty to push a simplified version, feel free to drop the commit again if you object @v4hn 

Let's see if that works or if we need to add a dummy `pybind11::module` target for melodic/noetic

v4hn:
@simonschmeisser I'm not sure it's much cleaner now and it looks like CI still breaks. 😉

From the discussion above I think the consensus is to rework pybind11_catkin to use the system package if it is at least as new as the pinned version.
With this patch I wanted to drop the whole package from MoveIt, but Robert wants to keep it.

I can only look into any of this from next week again. A working patch is definitely welcome. 😊

simonschmeisser:
I migrated my main notebook to 22.04 on Saturday so I had no way to test this on melodic/noetic yesterday unfortunately. I'll try to make sure the next commit actually compiles :blush: Sorry for the noise.

simonschmeisser:
I failed ... please feel free to remove my commits or improve on them if the idea of a dummy target suits you (for all possible variants of you)

PR #3345, opened 2023-02-17
Disentangle joint name methods and add group_state to commander

v4hn:
- improve documentation of all MGI::get* functions for joint names
- expose get_variables in MGI python wrapper
- add "group_state" command to commander
- rename MGI::getJointNames() to getVariableNames()

The API-incompatible method rename can be dropped if requested, but it took me a while to figure out which method is correct when implementing `group_state`, so I believe refactoring makes a lot of sense.

### DIFF
--- a/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
+++ b/moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
@@ -1643,7 +1643,7 @@ void MoveGroupInterface::setRandomTarget()
   impl_->setTargetType(JOINT);
 }
 
-const std::vector<std::string>& MoveGroupInterface::getJointNames() const
+const std::vector<std::string>& MoveGroupInterface::getVariableNames() const
 {
   return impl_->getJointModelGroup()->getVariableNames();
 }
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -35,6 +35,7 @@ API changes in MoveIt releases
   You might need to define additional end-effectors.
 - Removed `ConstraintSampler::project()` as there was no real difference to `sample()`.
 - Removed `TrajectoryExecutionManager::pushAndExecute()` and the code associated to it. The code was unused and broken.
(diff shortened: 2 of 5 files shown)


moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
```diff
@@ -1644,6 +1644,11 @@ void MoveGroupInterface::setRandomTarget()
 }
 
 const std::vector<std::string>& MoveGroupInterface::getJointNames() const
+{
+  return getVariableNames();
+}
```
rhaschke, line 1649: Please move this implementation to the header - for inlining (and documenting use of the new function).

rhaschke (commented):
Approving with one small nitpick.

moveit_ros/planning_interface/move_group_interface/src/move_group_interface.cpp
v4hn: I put it here on purpose to keep the symbol in the built library during deprecation and it's documented in the deprecation tag already, but I don't care too much and will change it.

rhaschke: I see, that's a good argument too. Decide yourself. More importantly: this PR seems to break on Melodic (`master-ci`)

v4hn: Yes, live without f-strings is hard. :) Should be fixed now.

rhaschke: This type of format string is available in python 2.7:
```
a=1
"a={}".format(a)
```
You just don't yet have `f"{a}"`. Are you sure that this was causing the CI failure? I didn't have a look myself, sorry.

v4hn: That's just how I addressed it. :)
Yes, the CI output for failing tests is horrible, but that was the underlying mistake.

PR #3347, opened 2023-02-20
Addition of the AITstar planner from OMPL>=1.5.0

alaflaquiere:
### Description

Addition of the AITstar planner from OMPL to MoveIt.

- The AITstar planner is now included in `planning_context_manager.cpp`.
- The default parameters of the AITstar planner have been defined in `moveit_config_data.cpp`, so that the planner can be automatically configured by the setup_assistant.

IMPORTANT NOTE: The AITstar planner only properly works post commit [562f3ac](https://github.com/ompl/ompl/commit/562f3ac562f2c3bdcaaab7cd81da4fe00982b598) in OMPL.

A related [moveit_tutorial PR#770](https://github.com/ros-planning/moveit_tutorials/pull/770) has been created.


moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -863,6 +863,20 @@ std::vector<OMPLPlannerDescription> MoveItConfigData::getOMPLPlanners() const
   spar_stwo.addParameter("max_failures", "5000", "maximum consecutive failure limit. default: 5000");
   planner_des.push_back(spar_stwo);
 
+  OMPLPlannerDescription aitstar("AITstar", "geometric");
+  aitstar.addParameter(
+      "use_k_nearest", "1",
+      "Set whether to use a k-nearest RGG connection model. If false, AITstar uses an r-disc model. Default: 1");
```
simonschmeisser, line 869:
```suggestion
      "whether to use a k-nearest RGG connection model (1) or an r-disc model (0). Default: 1");
```
other descriptions seem to skip the verb

moveit_setup_assistant/src/tools/moveit_config_data.cpp
```diff
@@ -863,6 +863,20 @@ std::vector<OMPLPlannerDescription> MoveItConfigData::getOMPLPlanners() const
...
+  aitstar.addParameter("rewire_factor", "1.5",
+                       "Set the rewire factor of the RGG. Valid values: [1.0:0.01:3.0]. Default: 1.5");
```
simonschmeisser, line 871:
```suggestion
                       "rewire factor of the RGG. Valid values: [1.0:0.01:3.0]. Default: 1.5");
```

simonschmeisser (commented):
Thanks for your continued work on this! Looks like an interesting planner.

I added some nitpicks about the descriptions being more verbose than for other planners but looks fine otherwise

simonschmeisser:
seems like this planner is not available in melodics version of OMPL so you need to add an ifdef to enable it only on OMPL 1.6 or wherever it was introduced

alaflaquiere:
Other "informed trees" planners have been added to OMPL alongside `AITstar`; namely: `ABITstar`, `BITstar`, `EIRMstar` and `EITstar` (see [here](https://github.com/ompl/ompl/tree/main/src/ompl/geometric/planners/informedtrees)).

**_I haven't tested them_**, but if that makes sense I could also inlude them in the PR.
What do you think?

moveit_setup_assistant/src/tools/moveit_config_data.cpp
marlinstrub: From my experience smaller rewire factors result in better performance. Can we set the default to `1.001`?

marlinstrub:
Thank you @alaflaquiere for your effort to bring AIT* to MoveIt.

simonschmeisser:
> Other "informed trees" planners have been added to OMPL alongside `AITstar`; namely: `ABITstar`, `BITstar`, `EIRMstar` and `EITstar` (see [here](https://github.com/ompl/ompl/tree/main/src/ompl/geometric/planners/informedtrees)).
> 
> **_I haven't tested them_**, but if that makes sense I could also include them in the PR. What do you think?

Sounds like a great idea. Tiny problem will be that (some of) those planners have not been released yet so it will be ifdef disabled on ci.

`EIRMstar` is a multi-query planner which you need to handle slightly differently so that planner instances are reused for several planning attempts.

alaflaquiere:
> Sounds like a great idea. Tiny problem will be that (some of) those planners have not been released yet so it will be ifdef disabled on ci.
> 
> `EIRMstar` is a multi-query planner which you need to handle slightly differently so that planner instances are reused for several planning attempts.

Understood.
I simply added `BITstar` and `ABITstar` in my latest commit. These have both been released since OMPL_1.5.0 and do not require special handling for multi-query.

I have _**quickly**_ tested them by running `setup_assistant.launch` and the resulting `demo.launch` for a panda robot. But I do not guarantee their stability (`BITstar` crashed at least once during my few tests).

yurirocha15:
My team and I have been trying to use the informed planners with MoveIt for the past few weeks as well, and we found some issues that I think should be addressed before the planners are added.

The main issue we found is that the planners seem to not work as expected when using the `PoseModelStateSpace` (some results below). I believe the reason is that all informed planners use the [PathLengthDirectInfSampler](https://github.com/ompl/ompl/blob/main/src/ompl/base/samplers/informed/PathLengthDirectInfSampler.h), which uses the [getDimension method to define the ProlateHyperspheroid's dimension.](https://github.com/ompl/ompl/blob/main/src/ompl/base/samplers/informed/src/PathLengthDirectInfSampler.cpp#L244) However, the `PoseModelStateSpace` uses the [base class getDimension](https://github.com/ros-planning/moveit/blob/master/moveit_planners/ompl/ompl_interface/src/parameterization/model_based_state_space.cpp#L143), which returns the number of parameters in the joint space. Our current solution was to create a custom sampler (based on the PathLenghtDirectInfSampler) and [add conditions that return the right values depending on the current state space.](https://github.com/yurirocha15/moveit/blob/feature/support-informed-trees/moveit_planners/ompl/ompl_interface/src/samplers/ellipsoidal_informed_sampler.cpp#L376) However, to use the custom sampler, we need to also add a [custom objective](https://github.com/yurirocha15/moveit/blob/feature/support-informed-trees/moveit_planners/ompl/ompl_interface/src/objective/path_length_optimization_objective_for_informed_planner.cpp#L52). This solution is not pretty, and I don't think if it can be added to MoveIt in its current state. 
One "simple fix" would be to disable the `PoseModelStateSpace` if the planner is an informed planner (in a similar way as it is done when the user sets the `enforce_joint_model_state_space` config to true).

Here are two examples using AITstar and `PoseModelStateSpace` before and after the fix mentioned above:
- Before fix
pose_model_fail
```shell
[ INFO] [1678086438.675146074]: Subspace Dimension: 7  # I added this print to check the dimension of the PHS
[ INFO] [1678086438.675166284]: arm/arm[AITstar]: Solving the given planning problem. The current best solution cost is inf
[ INFO] [1678086438.776103384]: arm/arm[AITstar] (3 iterations): Found a new exact solution of cost 0.6065. Sampled a total of 0 states, 0 of which were valid samples (0.0 %). Processed 1 edges, 1 of which were collision checked (100.0 %). The forward search tree has 2 vertices, 1 of which are start states. The reverse search tree has 9 vertices, 8 of which are goal states.
[ INFO] [1678086498.664701978]: arm/arm[AITstar] (5 iterations): Found an exact solution of cost 0.6065.
[ INFO] [1678086498.664742229]: arm/arm[AITstar] (5 iterations): Sampled a total of 101950974 states, 1 of which were valid samples (0.0 %). Processed 1 edges, 1 of which were collision checked (100.0 %). The forward search tree has 2 vertices. The reverse search tree has 9 vertices.
[ INFO] [1678086498.664769979]: Solution found in 60.000104 seconds
```

- After fix
pose_model_success
```shell
[ INFO] [1678091143.570496101]: Planner configuration 'arm[AITstar]' will use planner 'geometric::AITstar'. Additional configuration parameters will be set when the planner is constructed.
[ INFO] [1678091143.581692097]: arm/arm[AITstar]: Solving the given planning problem. The current best solution cost is inf
[ INFO] [1678091143.680828618]: arm/arm[AITstar] (60 iterations): Found a new exact solution of cost 22.8668. Sampled a total of 116 states, 100 of which were valid samples (86.2 %). Processed 17 edges, 6 of which were collision checked (35.3 %). The forward search tree has 7 vertices, 1 of which are start states. The reverse search tree has 71 vertices, 10 of which are goal states.
[ INFO] [1678091143.831915034]: arm/arm[AITstar] (151 iterations): Found a new exact solution of cost 19.1980. Sampled a total of 344 states, 300 of which were valid samples (87.2 %). Processed 36 edges, 16 of which were collision checked (44.4 %). The forward search tree has 17 vertices, 1 of which are start states. The reverse search tree has 63 vertices, 10 of which are goal states.
[ INFO] [1678091144.194967473]: arm/arm[AITstar] (1006 iterations): Found a new exact solution of cost 17.8864. Sampled a total of 2448 states, 2100 of which were valid samples (85.8 %). Processed 78 edges, 27 of which were collision checked (34.6 %). The forward search tree has 21 vertices, 1 of which are start states. The reverse search tree has 159 vertices, 10 of which are goal states.
[ INFO] [1678091144.419001833]: arm/arm[AITstar] (1638 iterations): Found a new exact solution of cost 17.5757. Sampled a total of 3604 states, 3100 of which were valid samples (86.0 %). Processed 88 edges, 33 of which were collision checked (37.5 %). The forward search tree has 23 vertices, 1 of which are start states. The reverse search tree has 182 vertices, 10 of which are goal states.
[ INFO] [1678091146.300200200]: arm/arm[AITstar] (6819 iterations): Found a new exact solution of cost 17.3724. Sampled a total of 11102 states, 9600 of which were valid samples (86.5 %). Processed 104 edges, 39 of which were collision checked (37.5 %). The forward search tree has 26 vertices, 1 of which are start states. The reverse search tree has 357 vertices, 10 of which are goal states.
[ INFO] [1678091148.171368705]: arm/arm[AITstar] (11290 iterations): Found a new exact solution of cost 16.6169. Sampled a total of 17310 states, 15000 of which were valid samples (86.7 %). Processed 118 edges, 46 of which were collision checked (39.0 %). The forward search tree has 30 vertices, 1 of which are start states. The reverse search tree has 350 vertices, 10 of which are goal states.
[ INFO] [1678091203.632641217]: arm/arm[AITstar] (84115 iterations): Found an exact solution of cost 16.6169.
[ INFO] [1678091203.633079904]: arm/arm[AITstar] (84115 iterations): Sampled a total of 76729 states, 66400 of which were valid samples (86.5 %). Processed 118 edges, 46 of which were collision checked (39.0 %). The forward search tree has 29 vertices. The reverse search tree has 10 vertices.
[ INFO] [1678091203.633095234]: Solution found in 60.061931 seconds
```

moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
```diff
@@ -473,11 +473,20 @@ ompl_interface::ModelBasedPlanningContextPtr ompl_interface::PlanningContextMana
...
   if (it != pc->second.config.end() && boost::lexical_cast<bool>(it->second))
     factory = getStateSpaceFactory(JointModelStateSpace::PARAMETERIZATION_TYPE);
+  else if (req.planner_id.find("AITstar") != std::string::npos)
+    factory = getStateSpaceFactory(JointModelStateSpace::PARAMETERIZATION_TYPE);
+  else if (req.planner_id.find("ABITstar") != std::string::npos)
+    factory = getStateSpaceFactory(JointModelStateSpace::PARAMETERIZATION_TYPE);
+  else if (req.planner_id.find("BITstar") != std::string::npos)
+    factory = getStateSpaceFactory(JointModelStateSpace::PARAMETERIZATION_TYPE);
   else
     factory = getStateSpaceFactory(pc->second.group, req);
```
alaflaquiere, line 491: I added the following checks to enforce JointModelStateSpace when using any of the three Informed planners.

I'm a pretty poor C++ coder and don't know much about the inner workings of MoveIt, so I would be happy if someone more proficient could double check. :pray:

yurirocha15: I believe the planner_id can be any name the user wants to use in the `ompl_planning.yaml` file (even though it is usually the planner name, sometimes one can have two configurations for the same plan and use custom planner_ids to differentiate them). I reckon the `type` parameter is the one you are looking for.

```yaml
AITstar:    <-- This is the planner_id, which the user can define as they want
    samples_per_batch: 1000
    use_graph_pruning: true
    use_k_nearest: true
    rewire_factor: 1.001
    find_approximate_solutions: false
    set_max_num_goals: 10
    type: geometric::AITstar  <-- This is the planner type
```

You probably would need to do something like
```cpp
  auto type_it = pc->second.config.find("type");
  std::string planner_type;
  if (type_it != pc->second.config.end())
    planner_type = type_it->second;
```
to get the planner type, and then make the following comparison:
```cpp
else if (planner_type == "geometric::AITstar")
```

Finally, you probably will need to check with the maintainers if they agree with this change, or if they have a better suggestion.

alaflaquiere: Good point! I modified accordingly.
Thanks for the suggestion.

alaflaquiere:
@simonschmeisser What do you think about the latest changes?

simonschmeisser (approved):
Thanks for your commitment @alaflaquiere !   Looks good to me now and all three planners worked in my tests.

Thanks also to @marlinstrub and @yurirocha15 for your feedback, this has been a nice team effort!

It would be great if you could also open a PR for the tutorial here: https://github.com/ros-planning/moveit_tutorials/blob/master/doc/ompl_interface/ompl_interface_tutorial.rst

@marlinstrub do the planners support the more advanced termination conditions like `CostConvergence`?

simonschmeisser:
@yurirocha15 it would be nice to have a PR for your way of getting `PoseModelStateSpace` to work, it's easiest to discuss on actual code

alaflaquiere:
> Thanks for your commitment @alaflaquiere ! Looks good to me now and all three planners worked in my tests.
> 
> Thanks also to @marlinstrub and @yurirocha15 for your feedback, this has been a nice team effort!
> 
> It would be great if you could also open a PR for the tutorial here: https://github.com/ros-planning/moveit_tutorials/blob/master/doc/ompl_interface/ompl_interface_tutorial.rst
> 
> @marlinstrub do the planners support the more advanced termination conditions like `CostConvergence`?

I created [the following PR](https://github.com/ros-planning/moveit_tutorials/pull/770) to add the three planners to the list of available OMPL planners.
I didn't see any other section that needed an update, but tell me if I missed something.

alaflaquiere:
> In terms of possible future work:
> 
> I set `optimization_objective MaximizeMinClearanceObjective` and got the following output:
> 
> ```
> [INFO] [1679745693.825974090]: Maximizing Minimum Clearance: No direct informed sampling scheme is defined, defaulting to rejection sampling.
> [WARN] [1679745693.826021699]: RejectionInfSampler: The optimization objective does not have a cost-to-go heuristic defined. Informed sampling will likely have little to no effect.
> ```
> 
> I have no idea if that actually works for any other planner, just curious if this could be supported in theory ...
> 
> Also looking forward to a PR for EIRMstar of course wink

Maybe @marlinstrub can answer that?

Regarding EIRMstar, I would suggest to do that in a future PR, as I really don't have much time to spare for that right now.

simonschmeisser:
This currently cannot be merged because the git history is quite "twisted". I would recommend copying the two changed files away, resetting the branch to master and then do one fresh commit with your changes.

simonschmeisser:
> I haven't tried `CostConvergence` specifically, but AIT* and EIT* should support any (reasonable) termination condition. Is there an OMPL planner that doesn't support `CostConvergence`?

https://github.com/ompl/ompl/issues/874 mentions that planners need to actively call the intermediate solution callback for the termination conditions to work and unfortunately only few do

alaflaquiere:
> This currently cannot be merged because the git history is quite "twisted". I would recommend copying the two changed files away, resetting the branch to master and then do one fresh commit with your changes.

Done. :+1:

dbdxnuliba:
@alaflaquiere
@simonschmeisser 
AIT* is an optimizing planner, and I test it in moveit https://github.com/alaflaquiere/moveit-fork which integrate ompl 1.6.0
We focus on the path length rather than the speed, in moveit we set group planning time 5s,in a p2p in joint space plan demo
image
I got several log info such as " Found a new exact solution of cost ", but the value of cost first decreases then suddenly increases and then decreases again,here is the log info :
image
image
image

**we can see from the log that, the cost value is:11.5320->10.0759->10.0755->7.2064-> 16.1385(here suddenly increases,and then decrease)->13.7333->13.1666->12.8491->10.3396->9.2996.**

The logs contain two separate planning queries, the first of which goes from around 11 -> 7 cost, the second goes from 16 -> 9. AIT* is planning without knowledge of the prior solution, and finds different solutions each time.

Question:
1)Why here  contains two separate planning queries,in one plan p2p demo,is it a bug in moveit or something else
Any helps will be pleasured, thanks

yurirocha15:
@dbdxnuliba I believe you have several threads running in parallel (moveit run a max of 4 threads at the same time, if I'm not mistaken). As each thread work independently, they can have different path costs (which would explain the cost going up and down). After the end condition is met, the planner should return the path with the smallest cost.

To confirm this hypothesis, you can run the planner with a single thread by using setNumPlanningAttempts (or Planning Attempts in rviz) to 1 and see if the cost goes strictly down.

dbdxnuliba:
in main.cpp
here is the code 
```
int main(int argc, char **argv)
{
    ros::init(argc, argv, "test_ompl_planners");
    ros::NodeHandle nh("test_ompl_planners");
    ros::AsyncSpinner spinner(4);
    spinner.start();

    static const std::string LOGNAME = "test_ompl_planners";

    ros::Duration(1.0).sleep();
   moveit::planning_interface::MoveGroupInterface group(group_id);
    moveit::planning_interface::PlanningSceneInterface planning_scene_interface;

    group.setPlanningPipelineId("ompl");
    // group.setPlannerId("RRTConnect");
    group.setPlannerId(planner_id);
    // group.setPlannerId("LBTRRT");
    
    // group.setMaxVelocityScalingFactor(0.2);
    // group.setMaxAccelerationScalingFactor(0.1);
    group.setMaxVelocityScalingFactor(0.1);
    group.setMaxAccelerationScalingFactor(0.1);
    group.setPlanningTime(ompl_planner_time);
    group.setGoalJointTolerance(0.02);
    
    group.setJointValueTarget(target_joints);

    moveit::planning_interface::MoveGroupInterface::Plan plan;
    std::chrono::time_point<std::chrono::system_clock> start, end;
    start = std::chrono::system_clock::now();
    moveit::core::MoveItErrorCode moveit_states = group.plan(plan);
    getTrajectoryLength(plan.trajectory_);
    bool success = (moveit_states == moveit::planning_interface::MoveItErrorCode::SUCCESS);
    ROS_INFO("Visualizing plan first_joint_values (joint space goal) %s", success ? "" : "FAILED");
    if (!success)
    {
        ROS_ERROR("%s  failed to plan to first_joint_values,moveit_states.val():%d\n",group_id.c_str(),moveit_states.val);
        group.stop();
        return false;
    }
    end = std::chrono::system_clock::now();
    std::chrono::duration<double> spent_s = end - start;
    ROS_WARN("debug plan spent_s:%f\n", spent_s);
    start = std::chrono::system_clock::now();
    moveit_states = group.move();
    success = (moveit_states == moveit::planning_interface::MoveItErrorCode::SUCCESS);
    if (!success)
    {
        ROS_ERROR(" %s failed to plan to first_joint_values,moveit_states.val():%d\n",group_id.c_str(),moveit_states.val);
        group.stop();
        return -1;
    }
    end = std::chrono::system_clock::now();
    spent_s = end - start;
    ROS_WARN("debug move spent_s:%f\n", spent_s);
    ROS_INFO("%s joint_values,moveit_states.val:%d,arrive at  first_joint_values successfully!\n",group_id.c_str(),moveit_states.val);
    return 1;
}
```

yurirocha15:
@dbdxnuliba the MoveGroupInterface::move function plans and execute a motion. So, as suspected, you are planning twice: first with the MoveGroupInterface::plan function, then with the MoveGroupInterface::move function.

If you want to execute your first plan, you should be instead using `group.execute(plan)`. You can see the documentation for those functions [here.](http://docs.ros.org/en/noetic/api/moveit_ros_planning_interface/html/classmoveit_1_1planning__interface_1_1MoveGroupInterface.html#:~:text=move%20()-,Plan%20and%20execute,-a%20trajectory%20that)

dbdxnuliba:
thanks a lot
@yurirocha15 

> @dbdxnuliba the MoveGroupInterface::move function plans and execute a motion. So, as suspected, you are planning twice: first with the MoveGroupInterface::plan function, then with the MoveGroupInterface::move function.
> 
> If you want to execute your first plan, you should be instead using `group.execute(plan)`. You can see the documentation for those functions [here.](http://docs.ros.org/en/noetic/api/moveit_ros_planning_interface/html/classmoveit_1_1planning__interface_1_1MoveGroupInterface.html#:~:text=move%20()-,Plan%20and%20execute,-a%20trajectory%20that)

dbdxnuliba:
and do you have plan to integrate EIT* to moveit @alaflaquiere

alaflaquiere:
Thank you for answering @yurirocha15 !

@dbdxnuliba I currently don't have enough free time to integrate EIT* too.
It's possible that I might as some point in the future, but I wouldn't count on it too much.

dbdxnuliba:
thanks,I'm looking forward to your future work

yurirocha15:
> > I haven't tried `CostConvergence` specifically, but AIT* and EIT* should support any (reasonable) termination condition. Is there an OMPL planner that doesn't support `CostConvergence`?
> 
> [ompl/ompl#874](https://github.com/ompl/ompl/issues/874) mentions that planners need to actively call the intermediate solution callback for the termination conditions to work and unfortunately only few do

@simonschmeisser @marlinstrub I created a [PR](https://github.com/ompl/ompl/pull/1024) adding support for the `CostConvergence` termination condition to AIT* and EIT*. 
BIT* and ABIT* already support it.

alaflaquiere:
I tried to check the release roadmap on ZenHub, but to no avail.
Does any of you have any rough estimate of how soon I could hope to see this PR integrated in a future release?

simonschmeisser:
There is no fixed schedule for noetic releases but I'll ask around if we can have one. Then once we have a release a Noetic sync will have to happen, you can somewhat estimate them looking at https://discourse.ros.org/c/release/noetic/66

PR #3348, opened 2023-02-20
[Ruckig traj smoothing] Add a test, termination condition bugfix

AndyZe:
### Description

Previously I didn't understand the difference between `ruckig::Finished` and `ruckig::Working` so I focused only on `ruckig::Finished`. There's not much documentation but here it is:  https://github.com/pantor/ruckig/blob/master/include/ruckig/result.hpp

This corrects that error and adds a comment. A new test is also added.

### DIFF
--- a/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
+++ b/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
@@ -79,12 +79,13 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
     return false;
   }
 
-  auto ruckig_result = runRuckigInBatches(trajectory, ruckig_input);
-  if (ruckig_result.has_value())
-  {
-    trajectory = ruckig_result.value();
-  }
-  return ruckig_result.has_value();  // Ruckig failed to smooth the trajectory
+  // auto ruckig_result = runRuckigInBatches(trajectory, ruckig_input);
+  // if (ruckig_result.has_value())
+  // {
+  //   trajectory = ruckig_result.value();
+  // }
+  // return ruckig_result.has_value();  // Ruckig failed to smooth the trajectory
+  return runRuckig(trajectory, ruckig_input);
 }
 
 bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajectory,
@@ -142,12 +143,13 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
     }
   }
 
-  auto ruckig_result = runRuckigInBatches(trajectory, ruckig_input);
-  if (ruckig_result.has_value())
-  {
-    trajectory = ruckig_result.value();
-  }
-  return ruckig_result.has_value();  // Ruckig failed to smooth the trajectory
+  // auto ruckig_result = runRuckigInBatches(trajectory, ruckig_input);
+  // if (ruckig_result.has_value())
+  // {
+  //   trajectory = ruckig_result.value();
+  // }
+  // return ruckig_result.has_value();  // Ruckig failed to smooth the trajectory
+  return runRuckig(trajectory, ruckig_input);
 }
 
 std::optional<robot_trajectory::RobotTrajectory>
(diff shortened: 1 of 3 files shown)


moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -284,8 +274,7 @@ bool RuckigSmoothing::runRuckig(robot_trajectory::RobotTrajectory& trajectory,
 
   // Initialize the smoother
   double timestep = trajectory.getAverageSegmentDuration();
-  std::unique_ptr<ruckig::Ruckig<ruckig::DynamicDOFs>> ruckig_ptr;
-  ruckig_ptr = std::make_unique<ruckig::Ruckig<ruckig::DynamicDOFs>>(num_dof, timestep);
+  ruckig::Ruckig<ruckig::DynamicDOFs> ruckig(num_dof, timestep);
```
AndyZe, line 277: There's no reason why this needed to be a pointer

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -79,12 +79,7 @@ bool RuckigSmoothing::applySmoothing(robot_trajectory::RobotTrajectory& trajecto
     return false;
   }
 
-  auto ruckig_result = runRuckigInBatches(trajectory, ruckig_input);
```
AndyZe, line 82: The "batches" feature is not working well. Let's put this on ice. See https://github.com/ros-planning/moveit2/issues/1960

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -304,16 +293,31 @@ bool RuckigSmoothing::runRuckig(robot_trajectory::RobotTrajectory& trajectory,
...
+      ruckig_result = ruckig.update(ruckig_input, ruckig_output);
+
+      // The difference between Result::Working and Result::Finished is that Finished can be reached in one
+      // Ruckig timestep (constructor parameter). Both are acceptable for trajectories.
+      // (The difference is only relevant for streaming mode.)
 
-      if ((waypoint_idx == num_waypoints - 2) && ruckig_result == ruckig::Result::Finished)
+      // If successful and at the last trajectory segment
+      if ((waypoint_idx == num_waypoints - 2) &&
+          (ruckig_result == ruckig::Result::Working || ruckig_result == ruckig::Result::Finished))
```
henningkayser, line 304: wouldn't this lead to trajectories being aborted early? Maybe I'm reading it wrong. I see why this would be a good mode for streamed commands though.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -339,9 +343,8 @@ bool RuckigSmoothing::runRuckig(robot_trajectory::RobotTrajectory& trajectory,
           }
           target_state->update();
         }
-        ruckig_ptr = std::make_unique<ruckig::Ruckig<ruckig::DynamicDOFs>>(num_dof, timestep);
         initializeRuckigState(*trajectory.getFirstWayPointPtr(), group, ruckig_input, ruckig_output);
-        // Begin the while() loop again
+        // Begin the for() loop again
```
henningkayser, line 347: For readability, it would make a lot of sense to break the code into separate functions to loop over. There are so many for loops in here.

henningkayser (commented):
Just some minor questions from my end. The performance improvement seems apparent, but I don't really follow why it exists

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: No, Result::Working means Ruckig did find a solution but it wasn't within the timestep that Ruckig was initialized with at [L277](https://github.com/ros-planning/moveit/blob/7bb2d2961fa092de3a6039c0624d156b167e0a62/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp#L277).

That's acceptable because we set the correct waypoint duration at L306.

BTW I wouldn't mind checking with @pantor to see if we can get some paid help with optimizing this at some point. I know there are some further improvements that can be made.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: Did a first pass at this with a 100-line refactoring

henningkayser: sweet, already helps a lot

henningkayser (approved):

AndyZe:
I grabbed some plots to verify this doesn't do anything crazy. Seems good from what I can tell. Not quite optimal because of the iterative nature of multiple waypoints, but close to it.

J1, vel limit=2.18 rad/s, accel limit=3.75 rad/s^2, jerk limit=100
j1

J3, vel limit=2.18, accel limit=2.5, jerk limit=100
j3

J4, vel limit=2.18, accel limit=3.13, jerk limit=100
j4

PR #3355, opened 2023-02-28
only collect links with geometry for restricted collision checking

v4hn:
Assuming no unexpected bug this slightly reduces the size of the sets. This patch is for cleanup and I don't expect a performance gain.

@rhaschke I looked at this code after your comment [here](https://github.com/ros-planning/moveit_task_constructor/pull/428#issue-1589707550).


rhaschke (approved):

v4hn:
> Looks like, this seemingly trivial change causes CI failures...

Yes, `moveit_planners/pilz_industrial_motion_planner/test/python_move_group_planning.py` fails and it turns out to be correct to do so. Links without geometry can still have objects attached [in which case the link without own geometry still has to be considered active](https://github.com/ros-planning/moveit/blob/326e9b5fe76bdf856f9c4a2be56ce7e136651ff2/moveit_core/collision_detection_fcl/src/collision_common.cpp#L71-L82).

Closing.

PR #3359, opened 2023-03-03
Backport: Differential Drive for Planar Joints from Moveit2

scchow:
### Description

This PR is a backport of https://github.com/ros-planning/moveit2/pull/390 that implements a differential drive model for planar joints.

After backporting JointProperties into srdfdom for ROS1 (merged in https://github.com/ros-planning/srdfdom/pull/109), we can now specify a planar joint to have a `motion_model` property, which can be set to `holonomic` or `diff_drive` with:
```
<virtual_joint name="position" type="planar" parent_frame="odom" child_link="base_link"/>
<joint_property joint_name="position" property_name="motion_model" value="diff_drive" />
<joint_property joint_name="position" property_name="min_translational_distance" value="0.01" />
```

This PR is a backport of differential drive kinematics from MoveIt2 that are used when the planar joint is set to `diff_drive`.

The difference between planning for holonomic and differential drive enabled by this PR can be seen below, with the differential drive taking a rotate -> drive -> rotate approach compared to holonomic bases that can go straight to the specified goal position.

#### Holonomic:
Holonomic Robot Planning

#### Differential Drive:
Differential Drive Robot Planning

### Open Issues with Differential Drive Planning in MoveIt

Unfortunately, there are still some open issues I've encountered in the rest of the motion planning pipeline that prevents differential drive bases from being plug and play with MoveIt1. These will likely have to be addressed with separate PRs outside of this backport.

These issues include:
1. **Plan Execution**: Hitting the "execute" button on either of the planning screens above only moves the arm and not the base. I suspect this is because MoveIt! does not seem to handle publishing of planar joint states by default, nor do the corresponding "fake controllers" know what to do with these multi-dof joints.
2. ~~**Limited Workspaces**:  The default workspace of the robot seems to be limited to the blue cube visualized below. Trying to move the base outside this workspace results in a failure to plan, despite it being just open space.~~ I figured it out. There are fields in the Rviz GUI for specifying the workspace under the `Context` tab of the `MotionPlan` rviz tab. I'm still wondering where the default (2, 2, 2) values are set.
3. **Missing Parent Links:** The `dynamics_solver` complains that the planning group with planar joint does not have a parent link. This error is triggered [here](https://github.com/ros-planning/moveit/blob/326e9b5fe76bdf856f9c4a2be56ce7e136651ff2/moveit_core/dynamics_solver/src/dynamics_solver.cpp#L91) because the virtual planar joint that connects the base of the robot to an external frame does not have a proper parent link. I am not sure what are the consequences of not having the dynamics solver instantiated, but this does cause some `[ERROR]` messages to pop up.
4. When `moveit_core` is built as `Debug`, if the goal position of the base is rotated using the Rviz gui, the `ASSERT_ISOMETRY` [here](https://github.com/ros-planning/moveit/blob/326e9b5fe76bdf856f9c4a2be56ce7e136651ff2/moveit_core/robot_model/src/planar_joint_model.cpp#L228) fails and crashes the program. Despite this issue, when we build the package as `Release`, no error is reported and things seem to work correctly even if the assertion condition is not met.

I am looking for some guidance in resolving these issues, given my relative unfamiliarity with the inner workings of this aspect of the planning pipeline. If anyone has any insights into how I may fix these issues, please let me know.

### Checklist

### DIFF
--- a/moveit_core/robot_model/src/robot_model.cpp
+++ b/moveit_core/robot_model/src/robot_model.cpp
@@ -978,26 +978,26 @@ JointModel* RobotModel::constructJointModel(const urdf::Joint* urdf_joint, const
       }
     }
 
+    // parse a joint property string as a double with error handling
+    // if the property was successfully parsed,
+    // stores the resulting property value in prop_value and returns true
+    // otherwise logs a ROS_ERROR and returns false
+    auto parse_property_double = [&joint_name](auto& prop_name, auto& prop_value_str, double& prop_value) -> bool {
+      try
+      {
+        prop_value = moveit::core::toDouble(prop_value_str);
+        return true;
+      }
+      catch (const std::runtime_error& e)
+      {
+        ROS_ERROR_STREAM_NAMED(LOGNAME, "Unable to parse property " << prop_name << " for joint " << joint_name
+                                                                    << " as double: '" << prop_value_str << "'");
+      }
+      return false;
+    };
+
     for (const auto& [property_name, property_value_str] : srdf_model.getJointProperties(joint_name))
     {
-      // parse a joint property string as a double with error handling
-      // if the property was successfully parsed,
-      // stores the resulting property value in prop_value and returns true
-      // otherwise logs a ROS_ERROR and returns false
-      auto parse_property_double = [&joint_name](auto& prop_name, auto& prop_value_str, double& prop_value) -> bool {
-        try
-        {
-          prop_value = moveit::core::toDouble(prop_value_str);
-          return true;
-        }
-        catch (const std::runtime_error& e)
-        {
-          ROS_ERROR_STREAM_NAMED(LOGNAME, "Unable to parse property " << prop_name << " for joint " << joint_name
-                                                                      << " as double: '" << prop_value_str << "'");
-        }
-        return false;
-      };
-
       if (property_name == "angular_distance_weight")
       {
         double angular_distance_weight;
--- a/moveit_core/robot_model/src/robot_model.cpp
+++ b/moveit_core/robot_model/src/robot_model.cpp
@@ -984,16 +984,16 @@ JointModel* RobotModel::constructJointModel(const urdf::Joint* urdf_joint, const
       // if the property was successfully parsed,
       // stores the resulting property value in prop_value and returns true
       // otherwise logs a ROS_ERROR and returns false
-      auto parse_property_double = [&property_name, &property_value_str, &joint_name](double& prop_value) -> bool {
+      auto parse_property_double = [&joint_name](auto& prop_name, auto& prop_value_str, double& prop_value) -> bool {
         try
         {
-          prop_value = moveit::core::toDouble(property_value_str);
+          prop_value = moveit::core::toDouble(prop_value_str);
           return true;
         }
         catch (const std::runtime_error& e)
         {
-          ROS_ERROR_STREAM_NAMED(LOGNAME, "Unable to parse property " << property_name << " for joint " << joint_name
-                                                                      << " as double: '" << property_value_str << "'");
+          ROS_ERROR_STREAM_NAMED(LOGNAME, "Unable to parse property " << prop_name << " for joint " << joint_name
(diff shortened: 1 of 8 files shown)


moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -968,35 +968,37 @@ JointModel* RobotModel::constructJointModel(const urdf::Joint* urdf_joint, const
   {
     new_joint_model->setDistanceFactor(new_joint_model->getStateSpaceDimension());
     const std::vector<srdf::Model::PassiveJoint>& pjoints = srdf_model.getPassiveJoints();
+    std::string joint_name = new_joint_model->getName();
```
rhaschke, line 971: This should use a const reference.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -968,35 +968,37 @@ JointModel* RobotModel::constructJointModel(const urdf::Joint* urdf_joint, const
...
+      std::string property_value = property.second;
+      if (property_name == "angular_distance_weight")
       {
         double angular_distance_weight;
         try
         {
           std::string::size_type sz;
-          angular_distance_weight = std::stod(property.value_, &sz);
-          if (sz != property.value_.size())
+          angular_distance_weight = std::stod(property_value, &sz);
```
rhaschke, line 990: This parsing of double values is repeated below. Better move into a utility function!

rhaschke (commented):
Thanks for backporting this. As with the SRDF backport, I'm not at all happy with the implementation chosen in MoveIt2. Do you mind improving it directly here?

moveit_core/robot_model/src/robot_model.cpp
simonschmeisser: `std::stod` is 'designed' for use with natural language and assumes localized decimal separators. That's the polite version :wink:
Please use the conversion helpers in `moveit::tools`

Can we have CI ban using these functions for good?

scchow: Thanks for the feedback! 

So far, I've refactored the code to use `moveit::core::toDouble` instead of `std::stod` [here](https://github.com/scchow/moveit/compare/dev/diff-drive...scchow:moveit:dev/robot-str-parse), which has signficantly simplified the double parsing.

I was wondering whether a separate utility function is still necessary, since it is now just calling `toDouble` with a catch for improper values.

rhaschke: This is already a good improvement. Nevertheless, the error handling still constitutes a significant amount of redundancy. Simply define a lambda function parsing a specific double property with error handling.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -968,14 +968,103 @@ JointModel* RobotModel::constructJointModel(const urdf::Joint* urdf_joint, const
...
       }
     }
+
+    for (const auto& [property_name, property_value_str] : srdf_model.getJointProperties(joint_name))
+    {
+      // parse a joint property string as a double with error handling
+      // if the property was successfully parsed,
+      // stores the resulting property value in prop_value and returns true
+      // otherwise logs a ROS_ERROR and returns false
+      auto parse_property_double = [&joint_name](auto& prop_name, auto& prop_value_str, double& prop_value) -> bool {
```
rhaschke, line 987: I suggest moving this lambda definition outside the for loop.

scchow:
Thank you for all the feedback! I've incorporated the comments from the reviews and rebased to the current master branch in the most recent commit.

rhaschke (approved):
Thanks for iterating on this.

rhaschke:
@scchow, do you mind filing a PR against MoveIt2 with the code improvements made here?

scchow:
@rhaschke Can do. I am going to loop back to updating JointProperties for srdfdom in ROS2 (https://github.com/ros-planning/srdfdom/issues/110), then submit a PR with the improvements for MoveIt2. 

Juggling some other work right now, so it may be a week before I can get my ROS2 container up and running.

PR #3360, opened 2023-03-05
Add a version of TOTG `computeTimeStamps()` for combining sub-trajectories

AndyZe:
### Description

If the user has planned several sub-paths or trajectories, they often want to combine them and re-parameterize so the motion is smooth.

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -1026,4 +1026,22 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
 
   return true;
 }
+
+bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
+    const std::vector<robot_trajectory::RobotTrajectory>& input_trajs, robot_trajectory::RobotTrajectory& output_traj,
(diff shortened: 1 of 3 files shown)


simonschmeisser:
It would be lovely to have a variant where you can specify max velocity per segment instead of for the whole combined trajectory.

AndyZe:
@simonschmeisser I think that would be tricky. TOTG is about blending the radii of circles around the waypoints, so I don't know what would happen if you suddenly switched the circle radius.

I'm sure there's a way but that would be way deeper than this PR.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -1026,4 +1026,22 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
 
   return true;
 }
+
+bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
```
henningkayser, line 1030: This function doesn't really look specific to TOTG at all. Why not just append the trajectories first and call the existing one?

AndyZe: That's a good point. Assuming people can find the API for that. I'll just close this.

PR #3362, opened 2023-03-07
Skip executing zero-duration trajectories

rhaschke:
zero-duration trajectories cause the default ROS joint trajectory controller to [fail](https://github.com/ros-planning/moveit_task_constructor/issues/433#issuecomment-1456178725): 
`Dropping all 1 trajectory point(s), as they occur before the current time. Last point is 0.000000s in the past.`

In contrast to https://github.com/ros-planning/moveit_task_constructor/pull/438, I looked for the _latest_ possibility to skip those trajectories. I explicitly decided not to prune in `RobotTrajectory` message generation already, but only just before execution.

### DIFF
--- a/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
+++ b/moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
@@ -819,8 +819,8 @@ bool TrajectoryExecutionManager::configure(TrajectoryExecutionContext& context,
 {
   if (trajectory.multi_dof_joint_trajectory.points.empty() &&
       (trajectory.joint_trajectory.points.empty() ||
-       trajectory.joint_trajectory.header.stamp + trajectory.joint_trajectory.points.back().time_from_start ==
-           ros::Time()))
+       // zero-duration trajectory (start-time stamp + overall duration == 0) causes controller issues
+       (trajectory.joint_trajectory.header.stamp + trajectory.joint_trajectory.points.back().time_from_start).isZero()))
   {
     // empty trajectories don't need to configure anything
     return true;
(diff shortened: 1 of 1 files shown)


moveit_ros/planning/trajectory_execution_manager/src/trajectory_execution_manager.cpp
```diff
@@ -817,7 +817,10 @@ bool TrajectoryExecutionManager::configure(TrajectoryExecutionContext& context,
                                            const moveit_msgs::RobotTrajectory& trajectory,
                                            const std::vector<std::string>& controllers)
 {
-  if (trajectory.multi_dof_joint_trajectory.points.empty() && trajectory.joint_trajectory.points.empty())
+  if (trajectory.multi_dof_joint_trajectory.points.empty() &&
+      (trajectory.joint_trajectory.points.empty() ||
+       trajectory.joint_trajectory.header.stamp + trajectory.joint_trajectory.points.back().time_from_start ==
+           ros::Time()))
```
gavanderhoorn, line 823: as there's now three conditions, one of which takes a couple of seconds to 'parse': perhaps add a comment explaining what this is checking for / guarding against?

gavanderhoorn (approved):
Looks OK to me.

Off-topic here perhaps, but should we add a (debug) log statement to indicate the empty trajectory is 'skipped'?

rhaschke:
> Should we add a (debug) log statement to indicate the empty trajectory is 'skipped'?

I don't think so. It is an empty trajectory in the end. So nothing to do anyway. The other scenarios were skipped silently as well.

simonschmeisser:
> > Should we add a (debug) log statement to indicate the empty trajectory is 'skipped'?
> 
> I don't think so. It is an empty trajectory in the end. So nothing to do anyway. The other scenarios were skipped silently as well.

if it were empty you wouldn't be able to access `points.back()` ... I have done hacky things with trajectories when starting to work with MoveIt and not setting `time_from_start` was most likely not the worst. :+1: for a debug print in this case

rhaschke:
Ok, debug message added.

simonschmeisser (approved):

PR #3369, opened 2023-03-10
Fix MoveItCpp issues

rhaschke:
- Fix `const` member accessors. They should return a `ConstPtr` instead of a `const Ptr&`!
- Fix SEVERE ClassLoader warning on destruction of `MoveItCpp`:
  - PSM was released before copy of its RobotModel -> removed extra RobotModel copy
  - `clearContents()` was broken:
    - resets in wrong order: psm_ should be last
    - trajectory_execution_manager_ was missing
    
    I suggest to omit `clearContents()` and just rely on the (correct) ordering of member variables.
    While this is not explicit, we ensure that we don't miss any newly added member variable.

### DIFF
--- a/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
+++ b/moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
@@ -143,13 +143,14 @@ public:
   const std::map<std::string, planning_pipeline::PlanningPipelinePtr>& getPlanningPipelines() const;
 
   /** \brief Get the stored instance of the planning scene monitor */
-  const planning_scene_monitor::PlanningSceneMonitorPtr& getPlanningSceneMonitor() const;
+  planning_scene_monitor::PlanningSceneMonitorConstPtr getPlanningSceneMonitor() const;
   planning_scene_monitor::PlanningSceneMonitorPtr getPlanningSceneMonitorNonConst();
 
-  const std::shared_ptr<tf2_ros::Buffer>& getTFBuffer() const;
+  std::shared_ptr<const tf2_ros::Buffer> getTFBuffer() const;
+  std::shared_ptr<tf2_ros::Buffer> getTFBuffer();
 
   /** \brief Get the stored instance of the trajectory execution manager */
-  const trajectory_execution_manager::TrajectoryExecutionManagerPtr& getTrajectoryExecutionManager() const;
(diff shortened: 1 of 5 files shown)


moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
```diff
@@ -143,13 +143,14 @@ class MoveItCpp
   const std::map<std::string, planning_pipeline::PlanningPipelinePtr>& getPlanningPipelines() const;
 
   /** \brief Get the stored instance of the planning scene monitor */
-  const planning_scene_monitor::PlanningSceneMonitorPtr& getPlanningSceneMonitor() const;
+  planning_scene_monitor::PlanningSceneMonitorConstPtr getPlanningSceneMonitor() const;
   planning_scene_monitor::PlanningSceneMonitorPtr getPlanningSceneMonitorNonConst();
 
-  const std::shared_ptr<tf2_ros::Buffer>& getTFBuffer() const;
+  std::shared_ptr<const tf2_ros::Buffer> getTFBuffer() const;
+  std::shared_ptr<tf2_ros::Buffer> getTFBuffer();
```
v4hn, line 150: I was never a fan of the `*NonConst` naming scheme and think this approach is much cleaner. :+1:

moveit_ros/planning/moveit_cpp/src/moveit_cpp.cpp
```diff
@@ -164,7 +162,7 @@ bool MoveItCpp::loadPlanningPipelines(const PlanningPipelineOptions& options)
 
 moveit::core::RobotModelConstPtr MoveItCpp::getRobotModel() const
 {
-  return robot_model_;
+  return planning_scene_monitor_->getRobotModel();
```
v4hn, line 165: With regards to a recent review of yours this should be a header-only implementation. No strong feelings though.

v4hn (approved):
Approved with nits.

In general: Ouch... Shows how many maintainers used this class until now and cared about the design. :crying_cat_face:

moveit_ros/planning/moveit_cpp/include/moveit/moveit_cpp/moveit_cpp.h
rhaschke: Me neither. The *NonConst scheme was introduced/continued by @henningkayser. I didn't attempt to change that.

Issue #3364, opened 2023-03-08
Memory leak from moveit mesh filter


luorui93:
### Description
We recently found a possible memory leak when using the mesh filter in moveit_ros_perception package.
We actually tracked down the possible bug to this line of code:

https://github.com/ros-planning/moveit/blob/326e9b5fe76bdf856f9c4a2be56ce7e136651ff2/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp#L126

For some reason, the `delete[] data` line was commented out. After uncommenting it, the memory leak was solved, and we saw no functionality setbacks. 

### Your environment
* ROS Distro: [Noetic]
* OS Version: e.g. Ubuntu 20.04
* Source build 
* master branch

### Steps to reproduce
Run the mesh filter with nodelet and as long as there is any subscriber to the output topic, the memory leak will happen.

berkealgul:
Hello, I opened pr for the fix you mentioned

PR #3371, opened 2023-03-11
Memory leak hot fix for mesh filter

berkealgul:
### Description

According [this issue](https://github.com/ros-planning/moveit/issues/3364) there is memory leak from mesh filter. I hotfixed the problem according to the issue

### DIFF
--- a/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
+++ b/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
@@ -116,9 +116,9 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
   const float* src = reinterpret_cast<const float*>(depth_msg->data.data());
   const size_t size = depth_msg->width * depth_msg->height;
   // scale data by factor 1000 and convert from float to unsigned short
-  auto data = std::make_unique<unsigned short[]>(size);
-  std::transform(src, src + size, data.get(), [](float value) { return value * 1000; });
-  mesh_filter_->filter(data.get(), GL_UNSIGNED_SHORT);
+  Eigen::Array<unsigned short, Eigen::Dynamic, 1> data =
+      (1000. * Eigen::Map<const Eigen::Array<float, Eigen::Dynamic, 1>>(src, size)).cast<unsigned short>();
+  mesh_filter_->filter(&data[0], GL_UNSIGNED_SHORT);
 
   if (pub_filtered_depth_image_.getNumSubscribers() > 0)
   {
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
Thanks for your initiative. While your solution fixes the memory leak, it defeats the original idea of the code:  `data` was declared `static` here to avoid frequent memory allocation and deallocation, which is rather costly.

rhaschke:
I just noticed that `data_size` was never adjusted, thus rendering this optimization idea pointless. Another issue with static storage is missing thread safety. Thus one should drop the static storage...

rhaschke:
Instead of optimizing memory usage, one might try to optimize the scaling operation using `Eigen`...

rhaschke:
Ideally, one would not need data copying and scaling at all and just pass the original float data...

rhaschke:
I don't have a depth camera available to test. Please @berkealgul and @luorui93 test and review, particularly https://github.com/ros-planning/moveit/pull/3371/commits/f64ceb51cd2eee5319176f4cba7a269f9604121f and https://github.com/ros-planning/moveit/pull/3371/commits/b7b062a58e29903718f64a02f868c93e3eb9a807.

luorui93:
> I don't have a depth camera available to test. Please @berkealgul and @luorui93 test and review, particularly [f64ceb5](https://github.com/ros-planning/moveit/commit/f64ceb51cd2eee5319176f4cba7a269f9604121f) and [b7b062a](https://github.com/ros-planning/moveit/commit/b7b062a58e29903718f64a02f868c93e3eb9a807).

Thanks! I'll test these changes and get back you asap.

luorui93 (approved):
Tested on my machine and no memory leak was found.

rhaschke:
@luorui93, thanks for testing. Did you test both listed commits? Is there any difference? I'm not sure, why/whether the scaling by factor 1000 was needed...

rhaschke:
I'm confused now: b7b062a doesn't comprise the scaling by factor 1000, but you report it works as expected.
On the other hand, you state omitting the scaling makes the filter fail... I guess that's only happening if you _also_ use the conversion from `float` to `uint`?

rhaschke:
@luorui93, please file a new issue regarding the failing depth filter for a real RealSense camera. Particularly, please provide a rosbag file and your MoveIt sensor config to reproduce the issue. Thanks.

luorui93:
> @luorui93, please file a new issue regarding the failing depth filter for a real RealSense camera. Particularly, please provide a rosbag file and your MoveIt sensor config to reproduce the issue. Thanks.

Sure, I actually managed to fix the issue so I'll also send a pull request too probably. However I didn't use any moveit sensor config file (actaully I only used the mesh filter as a separate tool without any other moveit components), would you mind telling me what's the config file you are referring to? It's possible that this "moveit sensor config" could solve the issue already.

rhaschke:
If you already identified and fixed the bug, that's even better. I was referring to this [yaml config](https://ros-planning.github.io/moveit_tutorials/doc/perception_pipeline/perception_pipeline_tutorial.html#yaml-configuration-file-point-cloud).

luorui93:
I see, then the config file should not affect whether the issue occurs or not. I'll post an issue along with a pull request.
.

PR #3375, opened 2023-03-16
Check for a Ruckig jerk limit parameter

AndyZe:
### Description

We want the ability to set custom joint jerk limits for the Ruckig smoother, but in MoveIt1, the RobotModel does not contain that info. My feeling is, there probably isn't a desire to create a large API-breaking PR in MoveIt1 for adding a jerk limit. Here's what adding jerk to the RobotModel would look like, from MoveIt2:  https://github.com/ros-planning/moveit2/pull/683

So, I made this small, simple PR to add a ROS jerk limit parameter.

### Test instructions

I've been testing like this:

1. Add Ruckig jerk params to `panda_moveit_config/launch/demo.launch`
  - `<param name="/ruckig/panda_joint1/jerk_limit" type="double" value="300.0" />`  (and so on)
2. Add the Ruckig smoothing adapter to the top of the list at ompl_planning_pipeline.launch.xml
  - default="default_planner_request_adapters/AddRuckigTrajectorySmoothing
3. Launch with:  `roslaunch moveit_resources_panda_moveit_config demo.launch rviz_tutorial:=true`

### DIFF
--- a/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
+++ b/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
@@ -253,8 +253,7 @@ bool RuckigSmoothing::getRobotModelBounds(const double max_velocity_scaling_fact
                        << " rad/s^2. You can define acceleration limits in the URDF or joint_limits.yaml.");
       ruckig_input.max_acceleration.at(i) = max_acceleration_scaling_factor * DEFAULT_MAX_ACCELERATION;
     }
-    // TODO(andyz): Add bounds.jerk_bounded_ for MoveIt1
-    // For now, we check for a jerk parameter
+    // Read jerk limits from parameters since bounds.jerk_bounded_ was never implemented for MoveIt1
     ros::NodeHandle nh;
     double jerk_limit = DEFAULT_MAX_JERK;
     std::string jerk_param = "ruckig/" + vars.at(i) + "/jerk_limit";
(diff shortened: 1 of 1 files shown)


simonschmeisser:
Is that original PR actually API-breaking? or just simply ABI-breaking? We don't guarantee either one actually and I would vote for the real deal

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -251,10 +251,20 @@ bool RuckigSmoothing::getRobotModelBounds(const double max_velocity_scaling_fact
...
     }
     // TODO(andyz): Add bounds.jerk_bounded_ for MoveIt1
-    ruckig_input.max_jerk.at(i) = DEFAULT_MAX_JERK;
-    ROS_WARN_STREAM_ONCE_NAMED(LOGNAME, "Joint jerk limits are not defined. Using the default "
-                                            << DEFAULT_MAX_JERK
-                                            << " rad/s^3. You can define jerk limits in joint_limits.yaml.");
+    // For now, we check for a jerk parameter
+    ros::NodeHandle nh;
+    double jerk_limit = DEFAULT_MAX_JERK;
+    if (nh.getParam("ruckig/" + vars.at(i) + "/jerk_limit", jerk_limit))
```
henningkayser, line 257: where in the moveit config are these being set?

AndyZe: I would think it's in a launch file. Wherever the user wants to put it.

henningkayser: this clearly needs to be documented. The warning below hints at the `joint_limits.yaml` btw, which doesn't contain this parameter at all. Please either leave a note about fixing this, or do so by providing necessary updates for the resources package, possibly tutorials

henningkayser (approved):

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: I'll fix the message below (thanks!) and update the tutorials and (maybe) moveit_resources.

AndyZe: here's the documentation PR:  https://github.com/ros-planning/moveit_tutorials/pull/764

I'm going to go ahead and test this one last time then merge it

PR #3376, opened 2023-03-17
Mitigate Ruckig overshoot

AndyZe:
### Description

When smoothing a trajectory, sometimes it's not possible to avoid overshoot. An example is:

current state is (position = 0, velocity = 0), target state is (position = 1, velocity = -1)

However, the overshoot can be reduced by extending the segment duration, thereby reducing the velocity at the goal state. Here's a graphical explanation.

ruckig_duration_extension

### DIFF
--- a/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
+++ b/moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
@@ -282,13 +282,15 @@ bool RuckigSmoothing::runRuckig(robot_trajectory::RobotTrajectory& trajectory,
   ruckig::Result ruckig_result;
   double duration_extension_factor = 1;
   bool smoothing_complete = false;
+  size_t waypoint_idx = 0;
   while ((duration_extension_factor <= MAX_DURATION_EXTENSION_FACTOR) && !smoothing_complete)
   {
-    for (size_t waypoint_idx = 0; waypoint_idx < num_waypoints - 1; ++waypoint_idx)
+    while (waypoint_idx < num_waypoints - 1)
     {
+      moveit::core::RobotStatePtr curr_waypoint = trajectory.getWayPointPtr(waypoint_idx);
       moveit::core::RobotStatePtr next_waypoint = trajectory.getWayPointPtr(waypoint_idx + 1);
 
-      getNextRuckigInput(trajectory.getWayPointPtr(waypoint_idx), next_waypoint, group, ruckig_input);
+      getNextRuckigInput(curr_waypoint, next_waypoint, group, ruckig_input);
 
       // Run Ruckig
       ruckig_result = ruckig.update(ruckig_input, ruckig_output);
@@ -318,17 +320,16 @@ bool RuckigSmoothing::runRuckig(robot_trajectory::RobotTrajectory& trajectory,
       if (ruckig_result != ruckig::Result::Working && ruckig_result != ruckig::Result::Finished)
       {
         duration_extension_factor *= DURATION_EXTENSION_FRACTION;
-        // Reset the trajectory
-        trajectory = robot_trajectory::RobotTrajectory(original_trajectory, true /* deep copy */);
 
         const std::vector<int>& move_group_idx = group->getVariableIndexList();
-        extendTrajectoryDuration(duration_extension_factor, num_waypoints, num_dof, move_group_idx, trajectory,
+        extendTrajectoryDuration(duration_extension_factor, waypoint_idx, num_dof, move_group_idx, trajectory,
                                  original_trajectory);
 
         initializeRuckigState(*trajectory.getFirstWayPointPtr(), group, ruckig_input, ruckig_output);
-        // Begin the for() loop again
+        // Continue the loop from failed segment, but with increased duration extension factor
         break;
       }
+      ++waypoint_idx;
     }
   }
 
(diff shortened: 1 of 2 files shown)


moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -291,6 +291,7 @@ bool RuckigSmoothing::runRuckig(robot_trajectory::RobotTrajectory& trajectory,
       getNextRuckigInput(trajectory.getWayPointPtr(waypoint_idx), next_waypoint, group, ruckig_input);
 
       // Run Ruckig
+      ruckig::Trajectory<ruckig::DynamicDOFs> ruckig_trajectory;
```
henningkayser, line 294:
```suggestion
      ruckig::Trajectory<ruckig::DynamicDOFs> ruckig_trajectory(6);
```

I think the constructor requires the dof as constructor argument

AndyZe: I got it to compile like so:

`ruckig::Trajectory<ruckig::DynamicDOFs, ruckig::StandardVector> ruckig_trajectory(ruckig::DynamicDOFs);`

Note sure if this will run well, yet.

henningkayser: DynamicDOFs resolves size_t=0 (https://github.com/pantor/ruckig/blob/682dbafd45e41c94ed9d6ce45bb60b252e7faf14/include/ruckig/utils.hpp#L15). But the constructor has an argument that passes the dofs to the member, pretty sure these should be the number of joints

AndyZe: thanks!

AndyZe: The question really is that number of joints is not known at compile time

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -291,41 +292,63 @@ bool RuckigSmoothing::runRuckig(robot_trajectory::RobotTrajectory& trajectory,
...
+      if (overshoots || (ruckig_result != ruckig::Result::Working && ruckig_result != ruckig::Result::Finished))
       {
         duration_extension_factor *= DURATION_EXTENSION_FRACTION;
         // Reset the trajectory
         trajectory = robot_trajectory::RobotTrajectory(original_trajectory, true /* deep copy */);
 
         const std::vector<int>& move_group_idx = group->getVariableIndexList();
-        extendTrajectoryDuration(duration_extension_factor, num_waypoints, num_dof, move_group_idx, trajectory,
-                                 original_trajectory);
+        extendTrajectoryDuration(duration_extension_factor, num_waypoints, num_dof, move_group_idx, original_trajectory,
```
AndyZe, line 348: This was a bug!

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -354,13 +377,12 @@ void RuckigSmoothing::extendTrajectoryDuration(const double duration_extension_f
...
-    trajectory.setWayPointDurationFromPrevious(
-        time_stretch_idx,
-        duration_extension_factor * original_trajectory.getWayPointDurationFromPrevious(time_stretch_idx));
+    double new_timestep =
+        duration_extension_factor * original_trajectory.getWayPointDurationFromPrevious(time_stretch_idx);
+    trajectory.setWayPointDurationFromPrevious(time_stretch_idx, new_timestep);
     // re-calculate waypoint velocity and acceleration
     auto target_state = trajectory.getWayPointPtr(time_stretch_idx);
     const auto prev_state = trajectory.getWayPointPtr(time_stretch_idx - 1);
-    double timestep = trajectory.getAverageSegmentDuration();
```
AndyZe, line 363: Another bug

rhaschke:
I don't get your initial example: Halving the velocity at the target state isn't what you want, do you? The target state velocity is an input parameter, isn't it?
The core goal of `ruckig` is to provide _time-optimal_ trajectories considering velocity, acceleration, and jerk limits. Why would you want to _extend_ the trajectory duration?

AndyZe:
Good question! Some people want the time-optimal solution which passes through the given waypoints. Others want to avoid overshoot. An example of a few groups I've worked with who want to avoid overshoot are:

- A window-washing robot
- Optics grinding

So that's why I intended to make the overshoot mitigation optional.

AndyZe:
Good question! Some people want the time-optimal solution which passes through the given waypoints. Others want to avoid overshoot. An example of a few groups I've worked with who want to avoid overshoot are:

- A window-washing robot
- Optics grinding

So that's why I intended to make the overshoot mitigation optional.

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
```diff
@@ -441,4 +439,31 @@ void RuckigSmoothing::getNextRuckigInput(const moveit::core::RobotStateConstPtr&
                    ruckig_input.max_acceleration.at(joint));
   }
 }
+
+bool RuckigSmoothing::checkOvershoot(ruckig::Trajectory<ruckig::DynamicDOFs, ruckig::StandardVector>& ruckig_trajectory,
+                                     const size_t num_dof, ruckig::InputParameter<ruckig::DynamicDOFs>& ruckig_input,
+                                     const double overshoot_threshold)
+{
+  // For every timestep
+  for (double time_from_start = 0; time_from_start < ruckig_trajectory.get_duration();
```
henningkayser, line 448: Is there a feasible way to not start from 0 every time?

henningkayser (approved):

moveit_core/trajectory_processing/src/ruckig_traj_smoothing.cpp
AndyZe: MoveIt2 has that feature already  :)   I guess I can backport it to MoveIt1

https://github.com/ros-planning/moveit2/pull/1990

Issue #3386, opened 2023-03-24
The mesh filter is not working with real Realsense camera data.


luorui93:
### Description

The current mesh filter works perfectly with simulated Kinect camera in Gazebo. However when applied to the real world depth image returned from the RealSense D415, the filter doesn't work anymore and the returned filtered image is filled with 0.

### Your environment
* ROS Distro: Noetic
* OS Version: Ubuntu 20.04
* Source Build
* Master branch

### Steps to reproduce
A sample data is recorded in a rosbag and attached below. The /filtered_realsense_depth contains the output from the mesh filter. The /realsense1/depth/image_rect_raw contains the original data from Realsense D415.

### Expected behaviour
The filtered depth image should only the depth data with the robot part filtered.

### Actual behaviour
The filtered depth image is blank with 0.

### Backtrace or Console output
[realsense.zip](https://github.com/ros-planning/moveit/files/11067093/realsense.zip)

PR #3387, opened 2023-03-24
handling different encodings of depth images correctly

luorui93:
### Description

This is to solve the issue  #3386. The cause of the issue is possibly due to incorrect handling of different encodings of a depth image. The original code assumed the depth image to be encoded in `32FC1` while the depth image returned by Realsense D415 is encoded in `16UC1`.  This issue already existed in the code before the change made to solve #3371.
By using the predefined filter base behavior for `GL_UNSIGNED_SHORT`, the depth image can be properly filtered.
However I don't have other RGBD cameras at hand so I can't know if there is any other encoding used by other cameras, thus I only considered `16UC1` case separately and used the same implementation for all other possible encodings. 

### Checklist

### DIFF
--- a/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
+++ b/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
@@ -126,7 +126,7 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
     mesh_filter_->filter(reinterpret_cast<const float*>(depth_msg->data.data()), GL_FLOAT);
   else
   {
-   ROS_WARN_STREAM_THROTTLE_NAMED(1.0, LOGNAME, "Unsupported encoding of depth image: " << depth_msg->encoding);
+    ROS_WARN_STREAM_THROTTLE_NAMED(1.0, LOGNAME, "Unsupported encoding of depth image: " << depth_msg->encoding);
     return;
   }
 
--- a/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
+++ b/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
@@ -45,6 +45,7 @@
 #include <cv_bridge/cv_bridge.h>
 
 namespace enc = sensor_msgs::image_encodings;
+static const std::string LOGNAME = "depth_self_filter_nodelet";
 
 mesh_filter::DepthSelfFiltering::~DepthSelfFiltering()
 {
@@ -116,8 +117,9 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
   // Handling of two possible encodings of a depth image: 16UC1 and 32FC1
   if (depth_msg->encoding == sensor_msgs::image_encodings::TYPE_16UC1)
   {
-    ROS_WARN_STREAM_THROTTLE(1.0, "The input depth image is encoded in the deprecated 16UC1 format, please consider "
-                                  "converting it to 32FC1 according to ROS REP-118!");
+    ROS_WARN_STREAM_ONCE_NAMED(LOGNAME,
+                               "The input depth image uses a 16UC1 encoding. Consider converting the publisher to "
+                               "generate the canonical 32FC1 encoding instead according to ROS REP-118.");
     mesh_filter_->filter(depth_msg->data.data(), GL_UNSIGNED_SHORT);
   }
   else if (depth_msg->encoding == sensor_msgs::image_encodings::TYPE_32FC1)
--- a/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
+++ b/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
@@ -115,7 +115,11 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
 
   // Handling of two possible encodings of a depth image: 16UC1 and 32FC1
   if (depth_msg->encoding == sensor_msgs::image_encodings::TYPE_16UC1)
+  {
+    ROS_WARN_STREAM_THROTTLE(1.0, "The input depth image is encoded in the deprecated 16UC1 format, please consider "
+                                  "converting it to 32FC1 according to ROS REP-118!");
     mesh_filter_->filter(depth_msg->data.data(), GL_UNSIGNED_SHORT);
+  }
   else if (depth_msg->encoding == sensor_msgs::image_encodings::TYPE_32FC1)
     mesh_filter_->filter(reinterpret_cast<const float*>(depth_msg->data.data()), GL_FLOAT);
   else
--- a/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
+++ b/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
@@ -113,8 +113,16 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
   params.setCameraParameters(info_msg->K[0], info_msg->K[4], info_msg->K[2], info_msg->K[5]);
   params.setImageSize(depth_msg->width, depth_msg->height);
 
-  const float* src = reinterpret_cast<const float*>(depth_msg->data.data());
-  mesh_filter_->filter(src, GL_FLOAT);
+  // Handling of two possible encodings of a depth image: 16UC1 and 32FC1
+  if (depth_msg->encoding == "16UC1")
+  {
+    mesh_filter_->filter(depth_msg->data.data(), GL_UNSIGNED_SHORT);
+  }
+  else
+  {
+    const float* src = reinterpret_cast<const float*>(depth_msg->data.data());
+    mesh_filter_->filter(src, GL_FLOAT);
+  }
 
   if (pub_filtered_depth_image_.getNumSubscribers() > 0)
   {
--- a/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
+++ b/moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
@@ -126,7 +126,7 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
     mesh_filter_->filter(reinterpret_cast<const float*>(depth_msg->data.data()), GL_FLOAT);
   else
   {
-    ROS_WARN_STREAM_THROTTLE(1.0, "Unsupported encoding of depth image: " << depth_msg->encoding);
+   ROS_WARN_STREAM_THROTTLE_NAMED(1.0, LOGNAME, "Unsupported encoding of depth image: " << depth_msg->encoding);
     return;
   }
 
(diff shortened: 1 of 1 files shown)


luorui93:
@rhaschke Please review the pull request, thanks.

rhaschke (approved):
I was expecting exactly this problem. Thanks for the fix!

gavanderhoorn:
I've ran into this problem myself a couple of times, but I believe the proposed fix is not actually the correct one.

ROS uses metres for distance measurements, and that's why the original only used/supported `GL_FLOAT`: depth images are supposed to be encoded with `32FC1`, which use floating point nrs to encode distance measurements.

The problem is the realsense drivers, which appear to use `16UC1` to encode millimetres, which goes against [REP 118: Depth Images](https://www.ros.org/reps/rep-0118.html).

To correct for this (ie: convert the offending depth image to one which adheres to the REP), you can use the [depth_image_proc/convert_metric](http://wiki.ros.org/depth_image_proc#depth_image_proc.2Fconvert_metric) nodelet. It'll take the `16UC1` encoded image and output a `32FC1` image (it can actually also do it the other way around, but that's not documented on the wiki). As it's a nodelet, it does incur some overhead, but mostly just the unit conversion, not the (de)serialisation penalty.

So my suggestion would be: document the fact MoveIt follows REP-118 and the only encoding accepted is `32FC1`. Printing the error proposed in this PR is also good.

I would not change the current implementation to accept `16UC1` encoded images.

gavanderhoorn (changes requested):
Marking as `request-changes` to avoid this getting (accidentally) merged.

rhaschke:
While I agree that the ROS default is to use float encoding and meters, I don't see any problem to offer `16UC1` as well particularly because it comes at no extra cost. @gavanderhoorn, you didn't list any concrete argument to forbid the `16UC1` format. 

Regarding the scaling by 1000: The [old code](https://github.com/ros-planning/moveit/pull/3371/files#diff-5ead9ca297a7f08875db08973bb53a07c5100ac52a6c0885906abfebc50720fcL122-L123) was doing the conversion exactly the other way around, i.e. `float` meters to `uint` millimeters!

gavanderhoorn:
> While I agree that the ROS default is to use float encoding and meters, I don't see any problem to offer `16UC1` as well particularly because it comes at no extra cost. @gavanderhoorn, you didn't list any concrete argument to forbid the `16UC1` format.

I believe adhering to convention and not supporting a deprecated encoding should be sufficient motivation.

From REP-118:

> Consumers of depth images are only required to support the canonical floating point representation.

It does say this as well:

> Alternatively, a device driver may publish depth images encoded as 16-bit unsigned integer, where each pixel is depth in millimeters.

but also:

> This differs from the standard units recommended in REP 103.

and as always: deviations from REPs should be considered undesirable and at the very least come with a very good motivation.

We should not argue about whether to perhaps conform to spec, the *burden of proof* is in the other direction.

rhaschke:
As you quote REP-118:

> Consumers of depth images are _only required_ to support the canonical floating point representation.

This doesn't forbid consumers to support other formats as well :wink:

gavanderhoorn:
True. I quoted it on purpose.

I still don't see a good rationale for why `16UC1` should be supported.

rhaschke:
> a device driver may publish depth images encoded as 16-bit unsigned integer, where each pixel is depth in millimeters.

Supporting that format comes at no extra cost on our side. So why not supporting it?

gavanderhoorn:
Because we'd be enabling something which goes against an accepted standard and (arguably) best practice.

The `16UC1` encoded depth image was basically only "allowed" in REP-118 because there was a driver at the time it was written which used it and which was used by a large number of people. In a way, it was added for bw-compatibility.

The fact intel chose to not conform to standards is very unfortunate, but it doesn't seem necessary to support this exception any more.

simonschmeisser:
So the realsense camera can be used in a conforming way as well? Maybe we could print out an error with a link how to configure it correctly then?

gavanderhoorn:
> So the realsense camera can be used in a conforming way as well? Maybe we could print out an error with a link how to configure it correctly then?

By using the nodelet I mentioned, yes.

Contrary to what my comments may imply, I'm not completely opposed to merging this PR.

I just wanted to make it *really* clear there is actually a standard for this, and `16UC1` goes against it and the issue is actually on the depth image producing side (in this case: the realsense driver).

rhaschke:
> The 16UC1 encoded depth image was basically only "allowed" in REP-118 because there was a driver at the time it was written, which used it and which was used by a large number of people.

Doesn't the same reasoning apply to the RealSense driver as well? I fully agree that Intel is to blame here. However, on our side, it doesn't cost any performance to support that format as well. So why not do it?

gavanderhoorn:
> So why not do it?

we're starting to repeat ourselves here.

I've already made it clear why I believe we shouldn't do it, you've made it clear why you believe we should.

I'll withdraw my review and we can let other maintainers contribute to the discussion.

rhaschke:
Surely, not supporting it on our side, may increase pressure on Intel to change their decision. However, I'm afraid they just refer to REP-118 and claim _"a device driver may publish depth images encoded as 16-bit unsigned integer"_.

luorui93:
Thanks for the insightful discussion @rhaschke, @gavanderhoorn! I think I can understand both sides' rationale. So maybe I would suggest print a warning in the `16UC1` case saying that this format is not recommended and not conform to the REP 118 and the depth format `32FC1` is recommended? In this way we can acknowledge the end-user such rule exists without losing the compatibility.

moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
```diff
@@ -115,7 +115,11 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
 
   // Handling of two possible encodings of a depth image: 16UC1 and 32FC1
   if (depth_msg->encoding == sensor_msgs::image_encodings::TYPE_16UC1)
+  {
+    ROS_WARN_STREAM_THROTTLE(1.0, "The input depth image is encoded in the deprecated 16UC1 format, please consider "
+                                  "converting it to 32FC1 according to ROS REP-118!");
```
rhaschke, line 120: I still think the `16UC1` format is fully covered by REP-118 and should be supported as it doesn't cost us any effort. If this warning is issued for that format at all, it should be issued only once, i.e. using `ROS_WARN_STREAM_ONCE`.

rhaschke (commented):
I don't want to merge this w/o approval from another maintainer. @simonschmeisser, @v4hn what is your take on this?

moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
```diff
@@ -113,8 +113,20 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
   params.setCameraParameters(info_msg->K[0], info_msg->K[4], info_msg->K[2], info_msg->K[5]);
   params.setImageSize(depth_msg->width, depth_msg->height);
 
-  const float* src = reinterpret_cast<const float*>(depth_msg->data.data());
-  mesh_filter_->filter(src, GL_FLOAT);
+  // Handling of two possible encodings of a depth image: 16UC1 and 32FC1
+  if (depth_msg->encoding == sensor_msgs::image_encodings::TYPE_16UC1)
+  {
+    ROS_WARN_STREAM_THROTTLE(1.0, "The input depth image is encoded in the deprecated 16UC1 format, please consider "
+                                  "converting it to 32FC1 according to ROS REP-118!");
```
v4hn, line 120:
```suggestion
    ROS_WARN_STREAM_ONCE_NAMED(LOGNAME, "The input depth image uses a 16UC1 encoding. Consider converting the publisher to generate the canonical 32FC1 encoding instead according to ROS REP-118.");
```

plus the additional
```
static const std::string LOGNAME = "depth_self_filter_nodelet";
```
at the top

v4hn (approved):
> and at the very least come with a very good motivation.

- it's a one line patch to support it and that line literally comes for free
- we already have code for both cases in the mesh filter's `filter` method
- a user complained that it does not work and even provided the patch
- There is no readily usable launch file which includes the nodelet filter in the realsense-ros repository and there will almost surely never be one. [The documented ROS one branch](https://github.com/ros/rosdistro/blob/68f7990d23cfb40025fe5bf075314758bf564d34/noetic/distribution.yaml#L8084) doesn't even exist in the upstream repository anymore (it probably got renamed to [`ros1-legacy`](https://github.com/IntelRealSense/realsense-ros/tree/ros1-legacy))
- rviz supports it and users rightly complain that if they can see the cloud in their viewer, MoveIt should be able to process it as well. (I know that argument is chicken-and-egg recursion, so don't count it in your list)

That is a sufficient set of reasons for me and I approve merging the patch almost as-is.
I'm happy to include the warning for the 16UC1 case as well, but it should definitely be `ROS_WARN_STREAM_ONCE` instead and I propose a different phrasing.

@luorui93 could you implement the proposed changes?

gavanderhoorn:
> That is a sufficient set of reasons for me and I approve merging the patch almost as-is.

those are reasons for you to accept the patch.

It's not a rationale for why someone should deviate from the REP.

But ok.

moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
luorui93: Thanks, I've added the lines.

moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
```diff
@@ -113,8 +114,21 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
...
+    ROS_WARN_STREAM_ONCE_NAMED(LOGNAME,
+                               "The input depth image uses a 16UC1 encoding. Consider converting the publisher to "
+                               "generate the canonical 32FC1 encoding instead according to ROS REP-118.");
+    mesh_filter_->filter(depth_msg->data.data(), GL_UNSIGNED_SHORT);
+  }
+  else if (depth_msg->encoding == sensor_msgs::image_encodings::TYPE_32FC1)
+    mesh_filter_->filter(reinterpret_cast<const float*>(depth_msg->data.data()), GL_FLOAT);
+  else
+  {
+    ROS_WARN_STREAM_THROTTLE(1.0, "Unsupported encoding of depth image: " << depth_msg->encoding);
```
v4hn, line 129:
```suggestion
   ROS_WARN_STREAM_THROTTLE_NAMED(1.0, LOGNAME, "Unsupported encoding of depth image: " << depth_msg->encoding);
```
You forbade maintainers from pushing to your pull-request branch, otherwise I would have pushed it myself before merge :-)

rhaschke (approved):

moveit_ros/perception/mesh_filter/src/depth_self_filter_nodelet.cpp
```diff
@@ -113,8 +114,21 @@ void mesh_filter::DepthSelfFiltering::filter(const sensor_msgs::ImageConstPtr& d
...
+    ROS_WARN_STREAM_ONCE_NAMED(LOGNAME,
+                               "The input depth image uses a 16UC1 encoding. Consider converting the publisher to "
+                               "generate the canonical 32FC1 encoding instead according to ROS REP-118.");
+    mesh_filter_->filter(depth_msg->data.data(), GL_UNSIGNED_SHORT);
+  }
+  else if (depth_msg->encoding == sensor_msgs::image_encodings::TYPE_32FC1)
+    mesh_filter_->filter(reinterpret_cast<const float*>(depth_msg->data.data()), GL_FLOAT);
+  else
+  {
+   ROS_WARN_STREAM_THROTTLE_NAMED(1.0, LOGNAME, "Unsupported encoding of depth image: " << depth_msg->encoding);
```
rhaschke, line 129: fix formatting:
```suggestion
    ROS_WARN_STREAM_THROTTLE_NAMED(1.0, LOGNAME, "Unsupported encoding of depth image: " << depth_msg->encoding);
```

PR #3388, opened 2023-03-25
Generalize RobotState::setFromIK()

rhaschke:
So far, `setFromIK` only accepted target (link) frames that were rigidly connected to a solver's tip frame. This, for example, excluded the fingertip of an actuated gripper, because that would be separated by an active joint from the arm's tooltip. However, as long as this joint is not part of the JMG, the corresponding transform can be considered as fixed as well.

This PR generalizes the functions `getRigidlyConnectedParentLinkModel()` in RobotState and RobotModel to receive an optional JMG pointer. If present, only (active) joints from that group are considered non-fixed. This PR also enables subframe support for `setFromIK` - simply by using `getRigidlyConnectedParentLinkModel()`, which already supported that.

There is one drawback of this approach: A repeated application of `setFromIK` with the same target frame and JMG (as in `computeCartesianPath()`), will repeat the search for the common fixed parent link. Additionally, the passed RobotState needs to be up-to-date. We could mitigate this by pulling the corresponding code into a separate function and calling it once in `computeCartesianPath()`.

### DIFF
--- a/moveit_core/robot_model/src/robot_model.cpp
+++ b/moveit_core/robot_model/src/robot_model.cpp
@@ -1176,14 +1176,29 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
   return nullptr;
 }
 
-const LinkModel* RobotModel::getRigidlyConnectedParentLinkModel(const LinkModel* link)
+const LinkModel* RobotModel::getRigidlyConnectedParentLinkModel(const LinkModel* link, const JointModelGroup* jmg)
 {
   if (!link)
     return link;
   const moveit::core::LinkModel* parent_link = link->getParentLinkModel();
   const moveit::core::JointModel* joint = link->getParentJointModel();
+  decltype(jmg->getJointModels().cbegin()) begin{}, end{};
+  if (jmg)
+  {
+    begin = jmg->getJointModels().cbegin();
+    end = jmg->getJointModels().cend();
+  }
 
-  while (parent_link && joint->getType() == moveit::core::JointModel::FIXED)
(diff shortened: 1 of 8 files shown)


rhaschke:
I cherry-picked #3143 and added a fix for the new, failing unit test.

moveit_core/robot_model/src/robot_model.cpp
```diff
@@ -1176,14 +1176,28 @@ LinkModel* RobotModel::getLinkModel(const std::string& name, bool* has_link)
...
+  auto is_fixed = [begin, end](const moveit::core::JointModel* joint) {
+    if (joint->getType() == moveit::core::JointModel::FIXED)
+      return true;
+    if (begin != end && std::find(begin, end, joint) == end)
+      return true;
+    return false;
+  };
 
-  while (parent_link && joint->getType() == moveit::core::JointModel::FIXED)
+  while (parent_link && is_fixed(joint))
```
simonschmeisser, line 1200: It took me some while to parse and understand above code, maybe split into two lambdas? It's a shame that `contains` is only C++ 20

rhaschke: I agree. Splitting into two lambdas (`is_fixed(joint) && in_jmg(joint)`) is not possible, because that would change the logic. But I renamed the lambda and added some comments to clarify semantics. Hope that helps.

simonschmeisser: it would be `parent_link && (is_fixed(joint) || ! in_jmg(joint))` or `parent_link && (is_fixed(joint) || ! jmg->hasJoint(joint.name)))`

rhaschke: I don't consider this complex expression easier to parse :wink:

PR #3392, opened 2023-04-03
MoveGroupCommander.execute can take JointTrajectory argument

v4hn:
ChatGPT thought this would be valid and I agree that this simpler call makes sense. :)

Additionally the argument was named wrongly as it implies that the MoveGroupInterface::Plan class has a corresponding ROS message one might pass as argument.

Small quality of life improvement.

### DIFF
--- a/moveit_commander/src/moveit_commander/move_group.py
+++ b/moveit_commander/src/moveit_commander/move_group.py
@@ -680,12 +680,15 @@ class MoveGroupCommander(object):
         path.deserialize(ser_path)
         return (path, fraction)
 
-    def execute(self, plan_msg, wait=True):
+    def execute(self, trajectory, wait=True):
         """Execute a previously planned path"""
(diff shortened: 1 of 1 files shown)


moveit_commander/src/moveit_commander/move_group.py
```diff
@@ -680,12 +680,15 @@ def compute_cartesian_path(
         path.deserialize(ser_path)
         return (path, fraction)
 
-    def execute(self, plan_msg, wait=True):
+    def execute(self, trajectory, wait=True):
         """Execute a previously planned path"""
+        if not hasattr(trajectory, "joint_trajectory"):
+            plan_msg = RobotTrajectory(joint_trajectory=trajectory)
```
mikeferguson, line 686: this should be "trajectory = ..."

rhaschke (approved):

PR #3393, opened 2023-04-03
CSM: do not jump back in time

v4hn:
- This addresses the use-case where multiple drivers publish joint
  state messages. Of course these do not give any guarantees on the order of
  incoming joint states and can lead to the CSM generating states
  with non-incremental time stamps, thus leading to TrajectoryMonitor recording
  trajectories with non-incremental waypoints.
  A different fix for this might rework the CSM altogether to allow configuring what policy should be employed to synchronize between the different joint publishers, but I believe this approach to use the latest timestamp of anything received is sufficient for the class.

- The downside of the simple fix in the first commit is that the monitor does not work well with
  looping rosbags anymore as the time resets. That's why I added a second commit with minor overhead to restore functionality (slightly more correct than before) when used with a rosbag that loops its time stamps.


rhaschke (approved):
LGTM.

PR #3396, opened 2023-04-06
Add JointModel::satisfiesAccelerationBounds()

AndyZe:
### Description

We already have versions for satisfiesPositionBounds() and satisfiesVelocityBounds(), why not Acceleration too?

### DIFF
--- a/moveit_core/robot_model/src/joint_model.cpp
+++ b/moveit_core/robot_model/src/joint_model.cpp
@@ -122,6 +122,16 @@ bool JointModel::satisfiesVelocityBounds(const double* values, const Bounds& oth
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


moveit_core/robot_model/src/joint_model.cpp
```diff
@@ -115,10 +115,32 @@ bool JointModel::enforceVelocityBounds(double* values, const Bounds& other_bound
 bool JointModel::satisfiesVelocityBounds(const double* values, const Bounds& other_bounds, double margin) const
 {
   for (std::size_t i = 0; i < other_bounds.size(); ++i)
+  {
+    if (!other_bounds[i].velocity_bounded_)
+    {
+      continue;
+    }
```
AndyZe, line 122: Not having this previously was a bug

rhaschke (approved):
LGTM

PR #3397, opened 2023-04-08
Address deprecation warnings in Debian bookworm

v4hn:
- drop implicit CC definition
- avoid deprecated std::random_shuffle
- Fix qt5 deprecation
- do not use deprecated QVariant::operator<

The first four should be rather uncontroversial.

@rhaschke, please have a look at the last one, I expect you don't want to merge this as-is for your generalized code?

### DIFF
--- a/moveit_core/collision_detection/include/moveit/collision_detection/collision_matrix.h
+++ b/moveit_core/collision_detection/include/moveit/collision_detection/collision_matrix.h
@@ -94,9 +94,6 @@ public:
   /** @brief Construct the structure from a message representation */
   AllowedCollisionMatrix(const moveit_msgs::AllowedCollisionMatrix& msg);
 
-  /** @brief Copy constructor */
-  AllowedCollisionMatrix(const AllowedCollisionMatrix& acm) = default;
-
   /** @brief Get the type of the allowed collision between two elements.
    *  Return true if the entry is included in the collision matrix. Return false if the entry is not found.
    *  @param name1 name of first element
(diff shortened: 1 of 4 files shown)


moveit_core/collision_detection/include/moveit/collision_detection/collision_matrix.h
```diff
@@ -94,9 +94,6 @@ class AllowedCollisionMatrix
   /** @brief Construct the structure from a message representation */
   AllowedCollisionMatrix(const moveit_msgs::AllowedCollisionMatrix& msg);
 
-  /** @brief Copy constructor */
-  AllowedCollisionMatrix(const AllowedCollisionMatrix& acm) = default;
-
```
rhaschke, line 99: Aren't you a fan of explicit code? Instead of removing this explicit definition of the copy constructor, I suggest to _also_ add an explicit definition of the copy assignment operator.

moveit_setup_assistant/src/tools/collision_linear_model.cpp
```diff
@@ -273,7 +273,16 @@ bool SortFilterProxyModel::lessThan(const QModelIndex& src_left, const QModelInd
...
-    bool smaller = (value_left < value_right);
+    bool smaller{};
+    switch (value_left.type())
+    {
+      case QVariant::Int:
+        smaller = value_left.toInt() < value_right.toInt();
+        break;
+      default:
+        smaller = value_left.toString() < value_right.toString();
+        break;
```
rhaschke, line 284: :+1: For the purpose here, `Int` and `String` are fully sufficient.

rhaschke (approved):
Approving. But I suggest using explicit definitions of copy constructor and assignment.

moveit_core/collision_detection/include/moveit/collision_detection/collision_matrix.h
v4hn: I considered that too, but to be complete we would also want to add the default move signatures as well and we rarely use the `= default` notation in this repository unless necessary.

PR #3409, opened 2023-04-23
Read goal tolerances from kinematics configuration

v3040928092:
### Description

MoveIt often fails to plan a valid path because goal joint position, goal position, and goal orientation tolerances are hard-coded as arbitrary/magic numbers. Other values in the same Move Group interface implementation are easily configurable.

These changes allow overriding the default (hard-coded) values of `0.1 millimeter` and `0.1 degree` in `kinematics.yaml` to relax the tolerances for debugging inverse kinematics or when using low-fidelity hardware such as potentiometers which jump by several degrees due to digital noise while reading arm joint positions.

The values made configurable by this pull request are:

* `moveit::planning_interface::MoveGroupInterface::MoveGroupInterfaceImpl::goal_joint_tolerance_`
  * Read from `goal_joint_tolerance` value in `kinematics.yaml` under the joint model group name
* `moveit::planning_interface::MoveGroupInterface::MoveGroupInterfaceImpl::goal_position_tolerance_`
  * Read from `goal_position_tolerance` value in `kinematics.yaml` under the joint model group name
* `moveit::planning_interface::MoveGroupInterface::MoveGroupInterfaceImpl::goal_orientation_tolerance_`
  * Read from `goal_orientation_tolerance` value in `kinematics.yaml` under the joint model group name

An example of overriding them in `kinematics.yaml` generated by MoveIt setup assistant:

```
arm
   goal_joint_tolerance: 0.1
   goal_position_tolerance: 0.1
   goal_orientation_tolerance: 0.1
```

This area of the code is not covered by tests, see https://github.com/ros-planning/moveit/blob/1249c2b183e557d502a50c135fa327f55f26fedf/moveit_ros/planning_interface/test/move_group_interface_cpp_test.cpp

The constraint tolerance defaults in MoveIt appear to be completely undocumented. The closest help topic is https://ros-planning.github.io/moveit_tutorials/doc/kinematics_configuration/kinematics_configuration_tutorial.html, which could be updated to talk about these three new settings.

### Checklist

### DIFF
--- a/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
+++ b/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
@@ -57,6 +57,9 @@ static const std::string MOVEIT_ROBOT_STATE = "moveit_robot_state";
 // Default kin solver values
 static const double DEFAULT_KIN_SOLVER_SEARCH_RESOLUTION = 0.005;
 static const double DEFAULT_KIN_SOLVER_TIMEOUT = 0.005;
+static const double DEFAULT_GOAL_JOINT_TOLERANCE = 1e-4;
+static const double DEFAULT_GOAL_POSITION_TOLERANCE = 1e-4;
+static const double DEFAULT_GOAL_ORIENTATION_TOLERANCE = 1e-3;
 
 // ******************************************************************************************
 // Structs
--- a/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
+++ b/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
@@ -57,9 +57,6 @@ static const std::string MOVEIT_ROBOT_STATE = "moveit_robot_state";
 // Default kin solver values
 static const double DEFAULT_KIN_SOLVER_SEARCH_RESOLUTION = 0.005;
 static const double DEFAULT_KIN_SOLVER_TIMEOUT = 0.005;
-static const double DEFAULT_GOAL_JOINT_TOLERANCE = 1e-4;
-static const double DEFAULT_GOAL_POSITION_TOLERANCE = 1e-4;
-static const double DEFAULT_GOAL_ORIENTATION_TOLERANCE = 1e-3;
 
 // ******************************************************************************************
 // Structs
--- a/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
+++ b/moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
@@ -73,6 +73,9 @@ struct GroupMetaData
   std::string kinematics_solver_;               // Name of kinematics plugin to use
   double kinematics_solver_search_resolution_;  // resolution to use with solver
   double kinematics_solver_timeout_;            // solver timeout
+  double goal_joint_tolerance_;                 // joint tolerance for goal constraints
+  double goal_position_tolerance_;              // position tolerance for goal constraints
+  double goal_orientation_tolerance_;           // orientation tolerance for goal constraints
   std::string kinematics_parameters_file_;      // file for additional kinematics parameters
   std::string default_planner_;                 // Name of the default planner to use
 };
--- a/moveit_setup_assistant/src/widgets/group_edit_widget.cpp
+++ b/moveit_setup_assistant/src/widgets/group_edit_widget.cpp
@@ -281,14 +281,18 @@ void GroupEditWidget::setSelected(const std::string& group_name)
   setLineEditValue(group_metadata.kinematics_solver_timeout_, DEFAULT_KIN_SOLVER_TIMEOUT, kinematics_timeout_field_);
 
   // Load goal joint tolerance
-  setSpinBoxValue(group_metadata.goal_joint_tolerance_, DEFAULT_GOAL_JOINT_TOLERANCE, goal_joint_tolerance_field_);
+  setSpinBoxValue(group_metadata.goal_joint_tolerance_,
+                  moveit::planning_interface::MoveGroupInterface::DEFAULT_GOAL_JOINT_TOLERANCE,
+                  goal_joint_tolerance_field_);
 
   // Load goal position tolerance
-  setSpinBoxValue(group_metadata.goal_position_tolerance_, DEFAULT_GOAL_POSITION_TOLERANCE,
+  setSpinBoxValue(group_metadata.goal_position_tolerance_,
+                  moveit::planning_interface::MoveGroupInterface::DEFAULT_GOAL_POSITION_TOLERANCE,
                   goal_position_tolerance_field_);
 
   // Load goal orientation tolerance
-  setSpinBoxValue(group_metadata.goal_orientation_tolerance_, DEFAULT_GOAL_ORIENTATION_TOLERANCE,
+  setSpinBoxValue(group_metadata.goal_orientation_tolerance_,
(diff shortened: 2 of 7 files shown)


rhaschke (approved):
I generally approve. Would be nice, if you could add the defaults to the kinematics.yaml template of MSA - to document that they are configurable now.

v3040928092:
All good points, thank you. I haven't been able to find the kinematics.yaml template for MoveIt Setup Assistant on first attempt (it's not in `moveit_setup_assistant/templates/moveit_config_pkg_template/config`), going to investigate the code to see where that's coming from.

rhaschke:
kinematics.yaml is generated by code: 
https://github.com/ros-planning/moveit/blob/a7684e7891416dc92214023b2a004213e49a3a86/moveit_setup_assistant/src/tools/moveit_config_data.cpp#L456

v3040928092:
Thanks! I have one more thing to look at. When MoveIt Setup Assistant modifies an existing MoveIt configuration and updates all the files, it also updates the kinematics configuration. So it has to load the existing kinematics.yaml and then re-write it from scratch using new values.

If I leave that code alone and only add these settings when creating a new kinematics configuration, then launching the MSA for existing project would just erase these values, if I understand correctly.

See `MoveItConfigData::inputKinematicsYAML`

Let's say I modify this function as well so that it both loads and (re) writes these values correctly and they are not lost. The data structure they are stored in is actually re-used in the MoveIt RViz plugin:

```
struct GroupMetaData
{
  std::string kinematics_solver_;               // Name of kinematics plugin to use
  double kinematics_solver_search_resolution_;  // resolution to use with solver
  double kinematics_solver_timeout_;            // solver timeout
  std::string kinematics_parameters_file_;      // file for additional kinematics parameters
  std::string default_planner_;                 // Name of the default planner to use
};
```
So if I modify this data structure to fix the above issue, then I have to either provide defaults in RViz plugin, or actually modify the UI and add text boxes to read these from directly on that planning UI panel.

All of a sudden I am messing with UI that probably thousands of people are using every day.

So the two choices I have are:

- Ignore that these new settings will be overwritten (basically wiped out) when running MSA for existing configuration
- Split the data structure into two: one that will be used by MoveIt RViz plugin, and one that will be used by MSA to load and write kinematics.yaml

Any thoughts?

v3040928092:
Everything seems to be working well:

- MoveIt Setup Assistant creates the kinematics config with these values
- MSA allows user to enter their own values in the UI
- MSA saves an existing kinematics config keeping the currently configured values
- MSA lets user update these values in the UI
- Goal tolerances are loaded by RViz and working: this lets me continue working on my project, yay!

I think it's ready to go?

moveit_setup_assistant/src/widgets/group_edit_widget.cpp
```diff
@@ -94,6 +94,21 @@ GroupEditWidget::GroupEditWidget(QWidget* parent, const MoveItConfigDataPtr& con
   kinematics_timeout_field_->setMaximumWidth(400);
   form_layout->addRow("Kin. Search Timeout (sec):", kinematics_timeout_field_);
 
+  // goal joint tolerance to use when planning with solver
+  goal_joint_tolerance_field_ = new QLineEdit(this);
```
rhaschke, line 98: Better use `QDoubleSpinBox` to constrain inputs.

moveit_setup_assistant/src/widgets/group_edit_widget.cpp
```diff
@@ -94,6 +94,21 @@ GroupEditWidget::GroupEditWidget(QWidget* parent, const MoveItConfigDataPtr& con
...
+  goal_joint_tolerance_field_->setMaximumWidth(400);
+  form_layout->addRow("Goal Joint Tolerance (m):", goal_joint_tolerance_field_);
```
rhaschke, line 100: Units are meters for prismatic and rad for rotational joints. Use `(m|rad)`?

moveit_setup_assistant/src/widgets/group_edit_widget.cpp
```diff
@@ -249,6 +264,33 @@ void GroupEditWidget::setSelected(const std::string& group_name)
...
   kinematics_timeout_field_->setText(QString::number(*timeout));
 
+  // Load goal joint tolerance
+  double* goal_joint_tolerance = &config_data_->group_meta_data_[group_name].goal_joint_tolerance_;
+  if (*goal_joint_tolerance == 0)
+  {
+    // Set default value
+    *goal_joint_tolerance = DEFAULT_GOAL_JOINT_TOLERANCE;
+  }
+  goal_joint_tolerance_field_->setText(QString::number(*goal_joint_tolerance));
```
rhaschke, line 274: I suggest replacing the pointer with a reference:
```diff
-double* goal_joint_tolerance = &config_data_->group_meta_data_[group_name].goal_joint_tolerance_;
+double& goal_joint_tolerance = config_data_->group_meta_data_[group_name].goal_joint_tolerance_; 
```

In general, try to avoid code redundancy, e.g. by introducing a lambda function:
```cpp
auto setValue = [](double& value, const double dflt, QDoubleSpinBox *widget) {
  if (value <= 0)
    value = dflt;
  widget->setValue(value);
};
setValue(config_data_->group_meta_data_[group_name].goal_joint_tolerance_, DEFAULT_GOAL_JOINT_TOLERANCE, goal_joint_tolerance_field_);
```

rhaschke (changes requested):
Thanks for the iterations. I have some minor comments to address.

v3040928092:
OK, I think it's ready. I regression tested editing existing MoveIt config, creating new MoveIt config, and RViz plugin loading the goal tolerances correctly.

spinners

You didn't specify how you wanted the spinners to behave, so I set them like this:

```
const int DECIMALS_TOLERANCE = 4;
const double MIN_TOLERANCE = 1e-5;
const double MAX_TOLERANCE = 1.0;
const double STEP_TOLERANCE = 1e-4;
```

moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
```diff
@@ -57,6 +57,9 @@ static const std::string MOVEIT_ROBOT_STATE = "moveit_robot_state";
 // Default kin solver values
 static const double DEFAULT_KIN_SOLVER_SEARCH_RESOLUTION = 0.005;
 static const double DEFAULT_KIN_SOLVER_TIMEOUT = 0.005;
+static const double DEFAULT_GOAL_JOINT_TOLERANCE = 1e-4;
+static const double DEFAULT_GOAL_POSITION_TOLERANCE = 1e-4;
+static const double DEFAULT_GOAL_ORIENTATION_TOLERANCE = 1e-3;
```
rhaschke, line 62: These are redundant to the `constexpr` declarations in `move_group_interface.h`

rhaschke (approved):
Looks good to me. Thanks for those improvements.

moveit_setup_assistant/include/moveit/setup_assistant/tools/moveit_config_data.h
v3040928092: It's not 100% clear to me which headers are appropriate to include in which modules, hopefully keeping them in `move_group_interface.h` is fine. Pushed a commit to address this.

valnovytskyy:
This is set to go into `noetic-devel`, I assume someone will cherry pick back to main after the merge and this is OK?

rhaschke:
> This is set to go into noetic-devel, I assume someone will cherry pick back to main after the merge and this is OK?

I changed the target branch to `master`. `noetic-devel` is automatically updated subsequently.

v3040928092:
I fixed the kin. parameters group and adjusted the button height to match the line edit height as a bonus 💅

params file

PR #3412, opened 2023-04-26
A version of TOTG computeTimeStamps() that accounts for torque limits

AndyZe:
### Description

This adds a new version of TOTG `computeTimeStamps()` which takes robot dynamics into consideration. The robot inertia model is pulled from the URDF.

Here's the algorithm in a nutshell:

```
  // 1. Call computeTimeStamps() to time-parameterize the trajectory with given vel/accel limits.
  // 2. Run forward dynamics to check if torque limits are violated at any waypoint.
  // 3. If a torque limit was violated, decrease the acceleration limit for that joint and go back to Step 1.
```

This will probably be interesting to @mechwiz @stephanie-eng @sjahr 

Sorry for the large line count. Most of that comes from syncing with MoveIt2 (the 2nd commit). You could ignore that one when reviewing, if you want -- focus mostly on `computeTimeStampsWithTorqueLimits()`


moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -175,10 +184,94 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
   TimeOptimalTrajectoryGeneration(const double path_tolerance = 0.1, const double resample_dt = 0.1,
                                   const double min_angle_change = 0.001);
 
+  // clang-format off
```
henningkayser, line 187: why disable clang-format? this shouldn't be necessary

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -45,6 +45,15 @@
 
 namespace trajectory_processing
 {
+enum LimitType
```
henningkayser, line 48: LimitType shouldn't be public if it's only being used by a single private function. Overall, the use of an enum-to-string map just to print a simple log message seems overkill to me. I would be less confusing if you would just simplify the original code with [std::clamp](https://en.cppreference.com/w/cpp/algorithm/clamp)

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -48,6 +49,11 @@ namespace trajectory_processing
 {
 const std::string LOGNAME = "trajectory_processing.time_optimal_trajectory_generation";
 constexpr double EPS = 0.000001;
+constexpr double DEFAULT_TIMESTEP = 1e-3;
+constexpr double DEFAULT_SCALING_FACTOR = 1.0;
+constexpr double DEFAULT_VELOCITY_LIMIT = 1.0;
```
henningkayser, line 54: why? I thought we are good with [requiring users to set correct limits](https://github.com/ros-planning/moveit2/pull/1794)?  Scaling factor and timestep already have default arguments provided in the header. Let's not add more magic numbers to the code

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
```diff
@@ -175,10 +184,94 @@ class TimeOptimalTrajectoryGeneration : public TimeParameterization
...
+  * \param[in,out] trajectory A path which needs time-parameterization. It's OK if this path has already been
+  * time-parameterized; this function will re-time-parameterize it.
+  * \param velocity_limits Joint names and velocity limits in rad/s
+  * \param acceleration_limits Joint names and acceleration limits in rad/s^2
+  * \param max_velocity_scaling_factor A factor in the range [0,1] which can slow down the trajectory.
+  * \param max_acceleration_scaling_factor A factor in the range [0,1] which can slow down the trajectory.
+  */
+  // clang-format on
+  bool computeTimeStamps(robot_trajectory::RobotTrajectory& trajectory,
+                         const std::unordered_map<std::string, double>& velocity_limits,
```
henningkayser, line 221: The API would be much clearer if it would just use the joint bounds type. This would also reduce a lot of the code

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -874,83 +880,342 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+}
+
+bool TimeOptimalTrajectoryGeneration::computeTimeStampsWithTorqueLimits(
+    robot_trajectory::RobotTrajectory& trajectory, const geometry_msgs::Vector3& gravity_vector,
+    const std::vector<double>& joint_torque_limits, double accel_limit_decrement_factor,
+    const std::unordered_map<std::string, double>& velocity_limits,
+    const std::unordered_map<std::string, double>& acceleration_limits, const double max_velocity_scaling_factor,
+    const double max_acceleration_scaling_factor) const
+{
+  // 1. Call computeTimeStamps() to time-parameterize the trajectory with given vel/accel limits.
```
henningkayser, line 1053: Is there any reason why this should be associated with TOTG? To me this looks like it could be a stand-alone trajectory processing feature.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -989,7 +1260,7 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
   }
 
   // Now actually call the algorithm
-  Trajectory parameterized(Path(points, path_tolerance_), max_velocity, max_acceleration, 0.001);
+  Trajectory parameterized(Path(points, path_tolerance_), max_velocity, max_acceleration, DEFAULT_TIMESTEP);
```
henningkayser, line 1263: now I see why you added the constant. Are we actually sure that 0.001 is a good number? What if `resample_dt_` is smaller? Maybe we should set this one as a factor of `resample_dt_` to ensure adequate accuracy.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -874,83 +880,342 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+          // centripetal acceleration caused by velocity of another joint. This should be uncommon on serial manipulators
+          // because their torque limits are high enough to withstand issues like that (or it just wouldn't work at all...)
+
+          // Reset
+          waypoint->copyJointGroupAccelerations(group->getName(), joint_accelerations);
+
+          // Check if decreasing acceleration of this joint actually decreases joint torque. Else, increase acceleration.
+          double previous_torque = joint_torques.at(joint_idx);
+          joint_accelerations.at(joint_idx) *= (1 + accel_limit_decrement_factor);
+          if (!dynamics_solver.getTorques(joint_positions, joint_velocities, joint_accelerations, link_wrenches,
```
henningkayser, line 1159: It's great to see the dynamic solver actually being used!

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -874,83 +880,342 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+    zero_wrench.force.z = 0;
+    zero_wrench.torque.x = 0;
+    zero_wrench.torque.y = 0;
+    zero_wrench.torque.z = 0;
+    // KDL (the dynamics solver) requires one wrench per link
+    std::vector<geometry_msgs::Wrench> vector_of_zero_wrenches(group->getLinkModels().size());
+    return vector_of_zero_wrenches;
+  }();
+
+  // Copy the waypoints so we can modify them while iterating
```
henningkayser, line 1102: we can modify the container type as well. For resetting, you can simply store a copy

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -874,83 +880,342 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+  const std::vector<geometry_msgs::Wrench> link_wrenches = [&group] {
+    geometry_msgs::Wrench zero_wrench;
+    zero_wrench.force.x = 0;
+    zero_wrench.force.y = 0;
+    zero_wrench.force.z = 0;
+    zero_wrench.torque.x = 0;
+    zero_wrench.torque.y = 0;
+    zero_wrench.torque.z = 0;
+    // KDL (the dynamics solver) requires one wrench per link
+    std::vector<geometry_msgs::Wrench> vector_of_zero_wrenches(group->getLinkModels().size());
```
henningkayser, line 1098:
```suggestion
    std::vector<geometry_msgs::Wrench> vector_of_zero_wrenches(group->getLinkModels().size(), zero_wrench);
```
looks like you wanted to do this?

moveit_core/trajectory_processing/include/moveit/trajectory_processing/time_optimal_trajectory_generation.h
AndyZe: Well, this is copied from MoveIt2. Not sure if it's worth changing here.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: Well, it is based on TOTG. See L1116.

But making it a standalone feature seems OK too. Possibly keep it in the `trajectory_processing` folder, call it `time_optimal_with_torque_limits`? Or `iterative_torque_limit_parameterization`?

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: DEFAULT_SCALING_FACTOR and DEFAULT_TIMESTEP are gone now per your other comments.

DEFAULT_VELOCITY_LIMIT and DEFAULT_ACCEL_LIMIT were in the code previously, they simply weren't constants at the top of the file.  https://github.com/ros-planning/moveit/blob/8635f82e4823b970479f360216bbe8292f79e8e4/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp#L927

I'll see how difficult this will be to remove in MoveIt1. IIRC it breaks some of the tests

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: I don't know what you mean. Create another, temporary RobotTrajectory like this?

`RobotTrajectory temp_trajectory(trajectory, true /* deep copy */);`

I had that originally but figured it would be less efficient b/c it copies the RobotModel, too

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: Do we want to break TOTG for the thousands of MoveIt users out there who don't have acceleration limits defined?

( ﾟｰﾟ)

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: done

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
henningkayser: I thought we already had that discussion, but anyway it's not really in scope of this PR so I won't block it over this.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
henningkayser: It copies a pointer which is negligible. Copying RobotStates might still have a performance impact, not a big deal though.

moveit_core/trajectory_processing/include/moveit/trajectory_processing/iterative_torque_limit_parameterization.h
```diff
@@ -0,0 +1,78 @@
...
+/* Author: Andy Zelenak */
+/* Description: Time-parameterize a trajectory with Time Optimal Trajectory Generation, then iterate until torque limits are obeyed. */
+
+#pragma once
+
+#include <moveit/trajectory_processing/time_optimal_trajectory_generation.h>
+
+namespace trajectory_processing
+{
+class IterativeTorqueLimitParameterization
```
henningkayser, line 43: This is great!

henningkayser (approved):

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -874,83 +877,193 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+    // For active joints only (skip mimic joints and other types)
+    const moveit::core::VariableBounds& bounds = rmodel.getVariableBounds(vars[active_joint_indices[idx]]);
+
+    // Limits need to be non-zero, otherwise we never exit
+    if (bounds.velocity_bounded_)
+    {
+      if (bounds.max_velocity_ <= 0.0)
+      {
+        ROS_ERROR_NAMED(LOGNAME, "Invalid max_velocity %f specified for '%s', must be greater than 0.0",
+                        bounds.max_velocity_, vars[idx].c_str());
```
rhaschke, line 906: Shouldn't this read: `vars[active_joint_indices[idx]]`?

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -874,83 +877,193 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+    {
+      max_acceleration[idx] = DEFAULT_ACCELERATION_LIMIT;
+    }
   }
-  else
+
+  return doTimeParameterizationCalculations(trajectory, max_velocity, max_acceleration);
+}
+
+bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
```
rhaschke, line 937: This method has strong overlap with the other computeTimeStamps method.
Better factor out common code into a separate helper function!

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -874,83 +877,193 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+  if (!group)
+  {
+    ROS_ERROR_NAMED(LOGNAME, "It looks like the planner did not set the group the plan was computed for");
+    return false;
+  }
+
+  if (hasMixedJointTypes(group))
+  {
+    ROS_WARN_NAMED(LOGNAME, "There is a combination of revolute and prismatic joints in the robot model. TOTG's "
+                            "`path_tolerance` will not function correctly.");
```
rhaschke, line 1060: Could you please elaborate on this? I guess you refer to different units/scaling required for revolute vs. prismatic joints, while path_tolerance only provides a single parameter?

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -1024,4 +1144,32 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+      std::any_of(joint_models.cbegin(), joint_models.cend(), [](const moveit::core::JointModel* joint_model) {
+        return joint_model->getType() == moveit::core::JointModel::JointType::REVOLUTE;
+      });
+
+  return have_prismatic && have_revolute;
+}
+
+double TimeOptimalTrajectoryGeneration::verifyScalingFactor(const double requested_scaling_factor) const
+{
+  double scaling_factor = std::clamp(requested_scaling_factor, -1.0, 1.0);
```
rhaschke, line 1167: Why do you allow for negative scaling factors?

rhaschke (commented):
Dear @AndyZe, I have a few important pose-merge comments. Please address them.

AndyZe:
Thanks Robert, I will do this in the next few days :+1:

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: Yeah, you've got the idea. Will tweak the comment.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: yep, done

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: I believe you're correct. Done.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
AndyZe: Completely agree with you but won't address now. It's not really related to this "torque-limited" feature.

BTW I'm no longer at PickNik. I'm working on ag robots with AppliedLogix now.

moveit_core/robot_model/include/moveit/robot_model/joint_model_group.h
```diff
@@ -581,6 +581,10 @@ class JointModelGroup
   bool isValidVelocityMove(const double* from_joint_pose, const double* to_joint_pose, std::size_t array_size,
                            double dt) const;
 
+  /** \brief Computes the indices of joint variables given a vector of joint names to look up */
+  bool computeJointVariableIndices(const std::vector<std::string>& joint_names,
+                                   std::vector<size_t>& joint_bijection) const;
+
```
rhaschke, line 587: This seems to be an exact copy of `computeIKIndexBijection` - except throwing a different error message!
Increasing code redundancy should be strongly avoided!

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
rhaschke: You seem to like programming by copy-and-paste! I'm trying to cleanup the redundancy and discovered some more issues in this PR, namely you are incorrectly indexing the joints. Could you please open your PR #3426 for additional commits?

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -874,83 +877,193 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
...
+  double velocity_scaling_factor = verifyScalingFactor(max_velocity_scaling_factor);
+  double acceleration_scaling_factor = verifyScalingFactor(max_acceleration_scaling_factor);
 
-  double acceleration_scaling_factor = 1.0;
-  if (max_acceleration_scaling_factor > 0.0 && max_acceleration_scaling_factor <= 1.0)
+  // Get the velocity and acceleration limits for all active joints
+  const moveit::core::RobotModel& rmodel = group->getParentModel();
+  const std::vector<std::string>& vars = group->getVariableNames();
+  std::vector<size_t> active_joint_indices;
+  if (!group->computeJointVariableIndices(group->getActiveJointModelNames(), active_joint_indices))
```
rhaschke, line 887: Hi @AndyZe,
In contrast to the previous code, you switched from _all_ variables to _active_ variables when looking up max velocities and accelerations. However, the TOTG algorithm still considers _all_ variables, doesn't it?
I expect this to segfault due to mismatching vector lengths for variables and limits! Did you test this with robot models comprising mimic joints?

rhaschke:
@AndyZe, we get complaints from users on the TOTG behavior since the latest release: https://github.com/ros-planning/moveit/issues/3450#issuecomment-1658530276

PR #3426, opened 2023-05-10
Minor cleanup and debugging in TOTG

AndyZe:
### Description

Address @rhaschke 's comments from here:  https://github.com/ros-planning/moveit/pull/3412#pullrequestreview-1420919625

PR #3427, opened 2023-05-11
Fixup TOTG

rhaschke:
This fixes several issues introduced in #3412. Replaces #3426.

Apart from a lot of code redundancy introduced in #3412, this PR also fixes a severe bug introduced there:
Because the TOTG algorithm considers all group variables (including mimic joints), the lookup of velocity and acceleration limits cannot be restricted to active joint variables only. Differently sized vectors might lead to segfaults!

To be honest, I am somewhat disappointed about the poor quality of PRs and reviews from approved maintainers in the last few months. Please improve!

### DIFF
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -903,7 +903,7 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(robot_trajectory::RobotT
       if (bounds.max_velocity_ <= 0.0)
       {
         ROS_ERROR_NAMED(LOGNAME, "Invalid max_velocity %f specified for '%s', must be greater than 0.0",
-                        bounds.max_velocity_, vars[idx].c_str());
+                        bounds.max_velocity_, vars[active_joint_indices[idx]].c_str());
         return false;
       }
       max_velocity[idx] =
@@ -1057,7 +1057,7 @@ bool TimeOptimalTrajectoryGeneration::doTimeParameterizationCalculations(robot_t
   if (hasMixedJointTypes(group))
   {
     ROS_WARN_NAMED(LOGNAME, "There is a combination of revolute and prismatic joints in the robot model. TOTG's "
-                            "`path_tolerance` will not function correctly.");
+                            "`path_tolerance` parameter will not function correctly.");
   }
 
   const unsigned num_points = trajectory.getWayPointCount();
@@ -1164,9 +1164,9 @@ bool TimeOptimalTrajectoryGeneration::hasMixedJointTypes(const moveit::core::Joi
 
 double TimeOptimalTrajectoryGeneration::verifyScalingFactor(const double requested_scaling_factor) const
 {
-  double scaling_factor = std::clamp(requested_scaling_factor, -1.0, 1.0);
+  double scaling_factor = std::clamp(requested_scaling_factor, 0.01, 1.0);
 
-  if (requested_scaling_factor > 1.0 || requested_scaling_factor < -1.0)
+  if (requested_scaling_factor > 1.0 || requested_scaling_factor < 0.01)
   {
     ROS_WARN_NAMED(LOGNAME, "Invalid max_scaling_factor specified, defaulting to %f instead.", scaling_factor);
   }
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -1064,6 +1064,7 @@ bool TimeOptimalTrajectoryGeneration::doTimeParameterizationCalculations(robot_t
   const unsigned num_points = trajectory.getWayPointCount();
   const std::vector<int>& idx = group->getVariableIndexList();
   const unsigned num_joints = group->getVariableCount();
+  assert(group->getVariableCount() == max_velocity.size());
 
   // Have to convert into Eigen data structs and remove repeated points
   //  (https://github.com/tobiaskunz/trajectories/issues/3)
--- a/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
+++ b/moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
@@ -921,8 +921,8 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
     else  // default limit
     {
       max_velocity[j] = DEFAULT_VELOCITY_LIMIT;
-      ROS_WARN_NAMED(LOGNAME, "No velocity limits defined for '%s'! Define them in URDF or joint_limits.yaml",
-                     name.c_str());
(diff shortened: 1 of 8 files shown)


rhaschke:
@AndyZe, I noticed that the unit test is rather slow: it takes more than a minute to finish the single time parameterization test on my side. I have seen that your code essentially iteratively adapts the acceleration bounds based on observed torques. Looks like many iterations are required to converge. Did you analyze the behavior? How can the performance be further improved?

rhaschke:
Looks like I tested with a DEBUG build. With a release build, the time is down to 7s, which is still some time but maybe acceptable.

AndyZe:
I know what the problem is:

https://github.com/ros-planning/moveit/blob/4aeccc712293577e64918c0bb185ef8c38eeed84/moveit_core/trajectory_processing/test/test_iterative_torque_limit_parameterization.cpp#LL110C3-L110C78

This should be a shorter distance. There are a ton of waypoints right now.

rhaschke:
Indeed, reducing the amount of joint motion brings the test duration down to 3ms. Is this ready to merge?

AndyZe:
> To be honest, I am somewhat disappointed about the poor quality of PRs and reviews from approved maintainers in the last few months. Please improve!

> Is this ready to merge?

lol. Reviewing now.

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -954,115 +881,86 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
...
     // ACCELERATION LIMIT
-    // Check if a custom limit was specified as an argument
-    bool set_acceleration_limit = false;
-    it = acceleration_limits.find(vars[idx]);
+    it = acceleration_limits.find(name);  // check for a custom limit
     if (it != acceleration_limits.end())
     {
-      max_acceleration[idx] = it->second * acceleration_scaling_factor;
-      set_acceleration_limit = true;
+      max_acceleration[j] = it->second * acceleration_scaling_factor;
```
AndyZe, line 931:
```suggestion
      if (it->second <= std::numeric_limits<double>::epsilon())
      {
        ROS_ERROR_NAMED(LOGNAME, "Invalid acceleration limit %f specified for '%s', must be greater than 0.0", it->second,
                        name.c_str());
        return false;
      }
      max_acceleration[j] = it->second * acceleration_scaling_factor;
```

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
```diff
@@ -954,115 +881,86 @@ bool TimeOptimalTrajectoryGeneration::computeTimeStamps(
...
-    if (!set_acceleration_limit)
+    else  // default limit
     {
-      ROS_ERROR_STREAM_NAMED(LOGNAME, "No acceleration limit was defined for joint "
-                                          << vars[idx].c_str()
-                                          << "! You have to define acceleration limits in the URDF or "
-                                             "joint_limits.yaml");
-      return false;
+      max_acceleration[j] = DEFAULT_ACCELERATION_LIMIT;
+      // ROS_WARN_NAMED(LOGNAME, "No acceleration limits defined for '%s'! Define them in joint_limits.yaml", name.c_str());
```
AndyZe, line 948: Unless you had a special reason for this...

```suggestion
      ROS_WARN_NAMED(LOGNAME, "No acceleration limits defined for '%s'! Define them in joint_limits.yaml", name.c_str());
```

moveit_core/trajectory_processing/test/test_iterative_torque_limit_parameterization.cpp
```diff
@@ -41,57 +41,24 @@ TEST(time_optimal_trajectory_generation, test_totg_with_torque_limits)
...
 
-  const std::string srdf =
-      "<?xml version=\"1.0\" ?>"
-      "<robot name=\"one_robot\">"
-      "<virtual_joint name=\"base_joint\" child_link=\"base_link\" parent_frame=\"odom_combined\" type=\"planar\"/>"
-      "<group name=\"single_dof_group\">"
-      "<joint name=\"joint_a\"/>"
-      "</group>"
-      "</robot>";
+  const std::string srdf = R"(<?xml version="1.0" ?>
```
AndyZe, line 55: Is this R necessary? Seems strange:

`R"(<?xml version="1.0" ?>`

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
```diff
@@ -410,6 +411,61 @@ TEST(time_optimal_trajectory_generation, testSingleDofDiscontinuity)
   }
 }
 
+TEST(time_optimal_trajectory_generation, testMimicJoint)
+{
+  // Request a trajectory. This will serve as the baseline.
+  // Then decrease the joint torque limits and re-parameterize. The trajectory duration should be shorter.
+
```
AndyZe, line 418:
```suggestion
```

moveit_core/trajectory_processing/test/test_iterative_torque_limit_parameterization.cpp
rhaschke: Read about [raw string literals](https://en.cppreference.com/w/cpp/language/string_literal) (since C++11).

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
rhaschke: I deliberatively commented that out, because we didn't warn about missing acceleration limits in the past.
However, I'm fine to include it as well.

AndyZe (approved):

moveit_core/trajectory_processing/src/time_optimal_trajectory_generation.cpp
rhaschke: I think we should throttle those warnings to `ONCE`. We don't need to repeat the same info over and over for all joints.

AndyZe: fine with me

moveit_core/trajectory_processing/test/test_time_optimal_trajectory_generation.cpp
```diff
@@ -410,6 +411,58 @@ TEST(time_optimal_trajectory_generation, testSingleDofDiscontinuity)
...
+      </robot>)";
+
+  const std::string srdf = R"(<?xml version="1.0" ?>
+      <robot name="one_robot">
+        <virtual_joint name="base_joint" child_link="base_link" parent_frame="odom_combined" type="planar"/>
+        <group name="group">
+          <joint name="joint_a"/>
+          <joint name="joint_b"/>
+        </group>
+      "</robot>)";
```
simonschmeisser, line 443: Looks like an extra quote here?

rhaschke: Well spotted. Wasn't noticed, because the extra quote goes as text in the XML and is simply ignored.

PR #3437, opened 2023-05-21
OMPL: Add EIT* and EIRM* planners

simonschmeisser:
This adds support for EIT* and EIRM* planners to the OMPL plugin. https://github.com/ompl/ompl/pull/844

They are not yet part of a release and unfortunately OMPL does not increment its version number post release so the versioned includes are off by one and should actually check for 1.7.0

### Checklist
- [ ] Wait for OMPL 1.7.0 release to have a proper include guard
- [ ] Needs support for roadmap loading/saving in EIRM*

```
error: no matching function for call to ‘ompl::geometric::EIRMstar::EIRMstar(const ompl::base::PlannerData&)’
```
@marlinstrub  @yurirocha15 @alaflaquiere


moveit_planners/ompl/ompl_interface/src/planning_context_manager.cpp
```diff
@@ -229,6 +233,14 @@ MultiQueryPlannerAllocator::allocatePersistentPlanner<ompl::geometric::LazyPRMst
 {
   return new og::LazyPRMstar(data);
 };
+#if OMPL_VERSION_VALUE >= 1006000
+template <>
+inline ompl::base::Planner*
+MultiQueryPlannerAllocator::allocatePersistentPlanner<ompl::geometric::EIRMstar>(const ob::PlannerData& data)
+{
+  return new og::EIRMstar(data);
+};
```
yurirocha15, line 242: EIRMstar does not seem to have a constructor that loads the PlannerData (https://github.com/ompl/ompl/blob/main/src/ompl/geometric/planners/informedtrees/EIRMstar.h#L74). I am not sure how it supports multi-query, as I haven't had time to read the paper, but I think this part of the code will probably fail to build.

simonschmeisser: There are two aspects, MultiQuery means once initialized previous runs will be considered in some way. PersistensPlanner means there are also ways to load and save PlannerData from disk.

I could just change this `data->getSpaceInformation()` but I was trying to inspire you or @marlinstrub to actually contribute the feature in OMPL :wink: for feature parity with other MultiQuery planners

rhaschke (approved):
I approve. Obviously, before merging, this requires a new release of OMPL.

rhaschke:
@simonschmeisser Could you please resolve the conflicts.

Issue #3436, opened 2023-05-21
Time derivative of a Jacobian


ivatavuk:
Hello, 
I was wondering if there is a way of computing the Jacobian time derivative integrated in MoveIt? 
To clarify what I'm talking about, if a task-space velocity is:

$$\begin{equation}
  \boldsymbol{v} = \boldsymbol{J}\dot{\boldsymbol{q}}
\end{equation}$$

Acceleration is then:

$$\begin{equation}
  \boldsymbol{a} = \dot{\boldsymbol{J}}\dot{\boldsymbol{q}} + \boldsymbol{J}\ddot{\boldsymbol{q}}
\end{equation}$$

where $\dot{\boldsymbol{J}}$ is the Jacobian time derivative.

I would like to discuss the possibility of adding such functionality under RobotState, where Jacobian computation is located. I see this functionality is available in kdl, [here](https://github.com/orocos/orocos_kinematics_dynamics/blob/master/orocos_kdl/src/chainjnttojacdotsolver.hpp).

I think this would be a useful feature to have, and would be glad to work on it and provide a pull request, but would first want to get some feedback from the MoveIt community:
  - Is there already such a feature integrated in MoveIt, and I'm missing it?
  - Is [robot_state](https://github.com/ros-planning/moveit/blob/noetic-devel/moveit_core/robot_state/src/robot_state.cpp) the appropriate place for it, in a form of a function similar to getJacobian? Would this feature be within the intended scope of MoveIt's RobotState?
  
 Thanks in advance!

simonschmeisser:
Welcome, what would be typical use cases for the time derivative of the jacobian?

How would you calculate it? Would you need to store it because it needs a trajectory to be calculated? `RobotState` is used in many places and should ideally not grow further in memory consumption

rhaschke:
RobotState is the right place to add this feature and your proposal to add a function `getJacobianDerivative` is perfect.
@simonschmeisser, we don't need to store this - we don't store the Jacobian either.
@ivatavuk, MoveIt includes an inverse [dynamics solver](https://github.com/ros-planning/moveit/blob/fa6331ca55f97a4615d4166c14a5e47d9bbfc7d9/moveit_core/dynamics_solver/include/moveit/dynamics_solver/dynamics_solver.h), but this doesn't expose acceleration-specific stuff.

ivatavuk:
Thank you for the feedback!

@simonschmeisser
[...]

PR #3439, opened 2023-05-27
Fix planar joint Jacobian bug

ivatavuk:
### Description

While working on [this](https://github.com/ros-planning/moveit/issues/3436) I think I found a bug in Jacobian calculation for planar joints. 

The fix is simple, in lines performing `joint_axis` calculation, instead of `joint_transform`, `joint_transform.linear()` should be used, like for the other joint types. 

I've written a simple test, I don't think it needs to be merged, but it demonstrates the bug so you can check it.
For the example in the test, current `getJacobian` implementation calculates the `joint_axis` of the first joint to be [1, 0, 0]^T when q[0] = 0, and [11, 0, 0]^T when q[0] = 10, which results in different `Jacobian` values.


rhaschke (approved):
Thanks for this fix.

PR #3449, opened 2023-06-14
Improvements to create_ikfast_moveit_plugin.py

rhaschke:


gavanderhoorn:
Changes look sane enough. Have not run-tested it.

One question: why e53e80a4e3f41f21269221e0f4de86a50730169c?

rhaschke:
> One question: why [e53e80a](https://github.com/ros-planning/moveit/commit/e53e80a4e3f41f21269221e0f4de86a50730169c)?

The script checks from `${ROBOT_NAME}.srdf` that the robot name matches and the planning group exists.
However, some moveit_config packages might use a different `.srdf` filename. For example, `panda_moveit_config` uses `panda.srdf.xacro`, which made the script fail adapting `kinematics.yaml`.

gavanderhoorn:
> For example, `panda_moveit_config` uses `panda.srdf.xacro`, which made the script fail adapting `kinematics.yaml`.

that seems sufficiently unorthodox to maybe not cater to it?

rhaschke:
I don't agree. Why shouldn't we support such moveit_configs as well? Specifically, using xacro for SRDF is probably a common use case.

gavanderhoorn:
> Specifically, using xacro for SRDF is probably a common use case.

you probably encounter different users.

In any case, I don't object to the change.

PR #3457, opened 2023-06-22
Fix segfault when planning with differential drive planar joints

scchow:
### Description

In my previous PR #3359, where I backported differential drive from MoveIt2, I noted in the "Open Issues with Differential Drive Planning in MoveIt" section that:

> When moveit_core is built as Debug, if the goal position of the base is rotated using the Rviz gui, the ASSERT_ISOMETRY [here](https://github.com/ros-planning/moveit/blob/326e9b5fe76bdf856f9c4a2be56ce7e136651ff2/moveit_core/robot_model/src/planar_joint_model.cpp#L228) fails and crashes the program. Despite this issue, when we build the package as Release, no error is reported and things seem to work correctly even if the assertion condition is not met.

In further testing, I also noticed that RRT* was segfaulting intermittently after generating 100 or so plans. It turns out these two issues are interconnected.

#### Root Cause

- In a rare case when two states passed into `interpolate` are very close to each other, the result of [`angles::normalize_angle()`](https://github.com/ros/angles/blob/54ad72b0ce473a2e1d98da447898bf2fcd36519f/angles/include/angles/angles.h#L84C1-L89) called by [`angles::shortest_angular_distance()`](https://github.com/ros/angles/blob/54ad72b0ce473a2e1d98da447898bf2fcd36519f/angles/include/angles/angles.h#L104-L107) which is called by [`computeTurnDriveTurnGeometry()`](https://github.com/scchow/moveit/blob/69ca441bc343c51544b19dde9c5a095481a3d906/moveit_core/robot_model/src/planar_joint_model.cpp#L177C11-L177C11) will round to `0.0` due to floating point arithmetic truncation when `M_PI` is first added to, then subtracted from the angle difference.
- The result of this is used in the `interpolate()` as the divisor for splitting the path into [turn, drive, turn phases](https://github.com/scchow/moveit/blob/69ca441bc343c51544b19dde9c5a095481a3d906/moveit_core/robot_model/src/planar_joint_model.cpp#L217-L219). If there is also no `x`, `y`, and `z` displacement, this results in a divide-by-zero operation (`0.0/0.0`), which C++ helpfully sets the result as `nan` without throwing any errors.
- This `nan` is then used in subsequent state calculations, causing both the `ASSERT_ISOMETRY` error noticed in my prior PR when the package is built as debug and eventually culminating in a segfault.

#### Fixes

To fix this issue, I added:
- a check to only normalize the angle when the angle difference is sufficiently large which avoids unnecessary floating point truncation.
- a check for if the states are the same, which will cause interpolate to just return the final state to avoid additional divide-by-zero errors.

### Implication on ROS 2 Implementation

As far as I know, this same error should also occur in moveit2. I am currently setting up a ROS2 container to test this and will submit a follow up PR if it turns out to be susceptible to the same issue.

### Remaining Issues
After solving this problem, I've encountered another issue that arises when compiling with the debug flag set. When visualizing the geometry of a robot with a differential drive base using Rviz, we run into a separate assertion error which is caused by this line in the [RViz Plugin](https://github.com/ros-planning/moveit/blame/69ca441bc343c51544b19dde9c5a095481a3d906/moveit_ros/visualization/motion_planning_rviz_plugin/src/motion_planning_frame_joints_widget.cpp#L336). The backtrace can be found [here](https://pastebin.com/TjVDax45). This appears to be a completely separate issue, and I will have to debug this separately. This issue does not seem to appear when things are compiled with release.

### Checklist

### DIFF
--- a/moveit_core/robot_model/src/planar_joint_model.cpp
+++ b/moveit_core/robot_model/src/planar_joint_model.cpp
@@ -174,7 +174,15 @@ void computeTurnDriveTurnGeometry(const double* from, const double* to, const do
     initial_turn = angle_backward_diff;
   }
   drive_angle = from[2] + initial_turn;
-  final_turn = angles::shortest_angular_distance(drive_angle, to[2]);
+  final_turn = to[2] - drive_angle;
+
+  // Check if normalization is needed, because
+  // if fabs(final_turn) << M_PI, the floating point arithmetic performed in `normalize_angle`
+  // will cause final_turn to be rounded to 0
+  if (final_turn < -boost::math::constants::pi<double>() || final_turn >= boost::math::constants::pi<double>())
+  {
+    final_turn = angles::normalize_angle(final_turn);
+  }
 }
 
 void PlanarJointModel::interpolate(const double* from, const double* to, const double t, double* state) const
@@ -208,17 +216,35 @@ void PlanarJointModel::interpolate(const double* from, const double* to, const d
     double dx, dy, initial_turn, drive_angle, final_turn;
     computeTurnDriveTurnGeometry(from, to, min_translational_distance_, dx, dy, initial_turn, drive_angle, final_turn);
 
+    // approximate cost (distance traveled) doing initial turn
     double initial_d = fabs(initial_turn) * angular_distance_weight_;
+    // approximate cost (distance traveled) doing straight drive
     double drive_d = hypot(dx, dy);
+    // approximate cost (distance traveled) doing final turn
     double final_d = fabs(final_turn) * angular_distance_weight_;
 
+    // total cost of executing manuever
     double total_d = initial_d + drive_d + final_d;
 
+    // If the difference between `from` and `to` is so low that it is within floating point arithmetic error
+    // or if `from == to`, the following operations will result in nan
+    // Just set the return state to target state and don't interpolate.
+    if (total_d == 0.0)
+    {
+      state[0] = to[0];
+      state[1] = to[1];
+      state[2] = to[2];
+      return;
+    }
+
+    // fraction of cost for each segment
     double initial_frac = initial_d / total_d;
     double drive_frac = drive_d / total_d;
     double final_frac = final_d / total_d;
 
     double percent;
+
+    // If the current time step is still in the initial rotation phase.
     if (t <= initial_frac)
     {
       percent = t / initial_frac;
@@ -226,6 +252,7 @@ void PlanarJointModel::interpolate(const double* from, const double* to, const d
       state[1] = from[1];
       state[2] = from[2] + initial_turn * percent;
     }
+    // If the current time step is doing the driving phase.
     else if (t <= initial_frac + drive_frac)
     {
       percent = (t - initial_frac) / drive_frac;
@@ -233,6 +260,7 @@ void PlanarJointModel::interpolate(const double* from, const double* to, const d
       state[1] = from[1] + dy * percent;
       state[2] = drive_angle;
     }
+    // If the current time step is in the final rotation phase.
     else
     {
       percent = (t - initial_frac - drive_frac) / final_frac;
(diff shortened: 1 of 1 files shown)


moveit_core/robot_model/src/planar_joint_model.cpp
```diff
@@ -208,31 +216,51 @@ void PlanarJointModel::interpolate(const double* from, const double* to, const d
...
+    // approximate cost (distance traveled) doing final turn
     double final_d = fabs(final_turn) * angular_distance_weight_;
 
+    // total cost of executing manuever
     double total_d = initial_d + drive_d + final_d;
 
+    // If the difference between `from` and `to` is so low that it is within floating point arithmetic error
+    // or if `from == to`, the following operations will result in nan
+    // Just set the return state to target state and don't interpolate.
+    if (total_d == 0.0)
```
rhaschke, line 232: The comment correctly states that you want to test for floating point accuracy like so:
```suggestion
    if (total_d < std::numeric_limits<float>::epsilon())
```

moveit_core/robot_model/src/planar_joint_model.cpp
```diff
@@ -174,7 +174,15 @@ void computeTurnDriveTurnGeometry(const double* from, const double* to, const do
...
-  final_turn = angles::shortest_angular_distance(drive_angle, to[2]);
+  final_turn = to[2] - drive_angle;
+
+  // Check if normalization is needed, because
+  // if fabs(final_turn) << M_PI, the floating point arithmetic performed in `normalize_angle`
+  // will cause final_turn to be rounded to 0
+  if (final_turn < -boost::math::constants::pi<double>() || final_turn >= boost::math::constants::pi<double>())
+  {
+    final_turn = angles::normalize_angle(final_turn);
+  }
```
rhaschke, line 185: As you correctly handle `final_turn == 0` below, why do you need to replace `angles::shortest_angular_distance`?

rhaschke (commented):
Thanks for tracking that down and proposing a fix. See my comments below.

moveit_core/robot_model/src/planar_joint_model.cpp
scchow: Similar note as above. With the above change, the divisions in the following lines should technically work even if the angle change is very small. 

However, if we are ok with saying we won't support values this small, we can definitely use `std::numeric_limits<float>::epsilon()` as a threshold.

moveit_core/robot_model/src/planar_joint_model.cpp
scchow: When using `angles::shortest_angular_distance(drive_angle, to[2])`, if `to[2] - drive_angle << M_PI`, the result is `final_turn = 0.0`. 

Whereas with this implementation, `final_turn` will actually be assigned the difference between `to[2]` and `drive_angle` regardless of the magnitude of the difference. 

The reason this occurs is because the [`normalize()` function](https://github.com/ros/angles/blob/54ad72b0ce473a2e1d98da447898bf2fcd36519f/angles/include/angles/angles.h#L84C1-L89) does an addition by `M_PI` and then subtraction by `M_PI`, thus `normalize(x) != x` if `x` is very small due to floating point arithmetic, as shown by [this script](https://onlinegdb.com/XfEgGWwEYJ).

Note that we are performing the exact same operations as [`angles::shortest_angular_distance`](https://github.com/ros/angles/blob/54ad72b0ce473a2e1d98da447898bf2fcd36519f/angles/include/angles/angles.h#L104-L107), except with an extra check to see if normalization is actually necessary, in order to avoid this loss of precision.

I was not sure whether this loss of precision was acceptable, so I wrote it this way to avoid truncation as much as possible. I can revert it to `shortest_angular_distance` if this error is acceptable.

scchow:
I've rebased to most recent commit on noetic-devel and implemented the suggested changes.

v4hn (approved):
makes sense to me.

PR #3464, opened 2023-06-28
Fix the added path index vector merge in PlanningRequestAdapterChain

### LINKED ISSUE #3463: PlanningRequestAdapterChain merge the added path indices in the wrong order
### Description

When multiple planning request adapters (PRA) add states to the trajectory, the indices of the outer PRA can be incremented if the inner PRA added a state with a lower index. If I understand this correctly, this is wrong because the inner PRA indices should be incremented and not the opposite.

### Your environment

* ROS Distro: Noetic
* Binary build 1.1.12-1focal.20230524.152908

### Steps to reproduce

* Have at least two PRAs.
* Have the inner one (the latter in the pipeline PRA list) add a state with a low index. 
* Have the outer one (the first in the pipeline PRA list) add a state with a higher index.

For example, given the following pipeline:

```
my_pra/A
my_pra/B
my_pra/C
```

* A runs `planner()`
  * B runs `planner()`
    * C runs `planner()`
      * The planner output a 4 state long trajectory
    * C appends a state to the trajecotry, adds `4` to its `added_path_index`. The trajectory is now 5 states long.
  * B prepends a state to the trajectory, adds `0` to its `added_path_index`. The trajectory is now 6 states long.
* A retrieves B's trajectory and appends a state. It adds `6` to its `added_path_index`. The trajectory is now 7 states long.

So far so good. Now the `PlanningRequestAdapterChain` [merges the different `added_path_index`](https://github.com/ros-planning/moveit/blob/69ca441bc343c51544b19dde9c5a095481a3d906/moveit_core/planning_request_adapter/src/planning_request_adapter.cpp#L153). The goal is to update the indices if other states were added before them.

### Expected behaviour

The index of the state added by C is updated to 5 to reflect that it's now the before-the-last state.
The final merged `added_path_index` is `0, 5, 6`.

### Actual behaviour

The indices are merged and incremented in the wrong order.
1. A index `6` is added.
2. B index `0` is added. Because it is lower than the 6, the 6 is incremented by 1.
3. C index `4` is added. Because it its lower than the 7, the 7 is incremented by 1.

The final merged `added_path_index` is `0, 4, 8`. `4` doesn't correspond the the state added by `C` and `8` isn't even a valid index.

Hugal31:
### Description

When merging the `added_path_index` for the different planning request adapters, the `PlanningRequestAdapterChain` was adding them in the wrong order, causing wrong incrementation of the indices.

Closes #3463.

### Checklist
- [X] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html)

### DIFF
--- a/moveit_core/planning_request_adapter/CMakeLists.txt
+++ b/moveit_core/planning_request_adapter/CMakeLists.txt
@@ -12,3 +12,11 @@ install(TARGETS ${MOVEIT_LIB_NAME}
   RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
 
 install(DIRECTORY include/ DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION})
+
+if(CATKIN_ENABLE_TESTING)
+    catkin_add_gtest(test_planning_request_adapter_chain test/test_planning_request_adapter_chain.cpp)
+    target_link_libraries(test_planning_request_adapter_chain
+        moveit_test_utils
+        moveit_planning_interface
+        ${MOVEIT_LIB_NAME})
+endif()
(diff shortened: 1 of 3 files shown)


rhaschke (approved):
Thanks for finding this bug and providing a fix!

Issue #3007, opened 2021-12-22
Don't ignore CollisionObject's pose on getKnownObjectNamesInROI


corot:
### Description

Method `PlanningSceneInterface::getKnownObjectNamesInROI` ignores CollisionObject's pose attribute. [Here](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp#L115) it checks both mesh_poses and primitive_poses without composing with root pose, breaching the [msg specification](https://docs.ros.org/en/noetic/api/moveit_msgs/html/msg/CollisionObject.html)

### Your environment
* ROS Distro: [Noetic]
* OS Version: Ubuntu 20.04
* Source build, latest master

### Steps to reproduce
No need; the bug is clear and easy to explain cause CollisionObject's pose is relatively recent.

### Expected behaviour
Both loops, [meshes](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp#L118) and [primitives](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp#L125)  should use root pose * mesh/primitive poses to check if within ROI (not simply mesh/primitive poses as now).
Also, isn't [this check](https://github.com/ros-planning/moveit/blob/master/moveit_ros/planning_interface/planning_scene_interface/src/planning_scene_interface.cpp#L115) already obsolete? that is, is it not allowed now to have both empty mesh and primitive poses array? (if the root object pose is already enough)

### Actual behaviour

We ignore CollisionObject's pose

rhaschke:
Closed via #3477

PR #3477, opened 2023-07-20
Fix: consider collision_object's pose when checking if it's inside the ROI

corot:
### Description

Fix for #3007. The PR consists on transforming mesh/primitive poses by root pose to when checking if the object is within ROI (not simply mesh/primitive poses as now). More details on the issue description. 

I noticed that the same problem exists on ROS 2

### Checklist


rhaschke (approved):

corot:
question:  looks like master and noetic-devel are synonymous;  should we drop one of them to make things clearer?

PR #3489, opened 2023-08-26
Addressing discrepancies in parameter definition between IKFast solvers and the plugin template

nothingstopsme:
### Description

Two cases have been identified where the discrepancies occur:

1. When an IKFast solver is of an IK type Translation\*AxisAngle4D or Translation\*AxisAngle\*Norm4D, the current implementation takes the roll, pitch, or yaw value from the pose frame's orientation directly as the angle parameter passed to the solver; however, that does not match the angle definition adopted in IKFast, according to the solver-generating [script](https://github.com/rdiankov/openrave/blob/master/python/ikfast.py#L3403) from line 3403 to 3406.

    A quick example can immediately highlight the difference:
    If the orientation of a pose frame is described by (roll = PI, pitch = 0, yaw = 0), then a direction vector (0, 0, 1) in the 
    manipulator's frame will be mapped to (0, 0, -1); in that case, a TranslationXAxisAngle4D solver will view it as PI/2 
    from the x-axis, while the current implementation of the plugin template calls the solver with an angle of 0.

2. There are several IK types (e.g. TranslationDirection5D) of which IKFast solvers require callers to pass in parameters calculated from the manipulator's direction vector defined in its own frame; nonetheless, that vector is currently fixed to (0, 0, 1) in the plugin template, and this assumption might become invalid if users have a different robot setup and manually generate their solver cpp with a customised wrapper XML specifying a different direction.

So this PR contains two commits aimed at addressing these two cases, respectively:

1. Modifying the plugin template so that it calculates those angles the same way as indicated in the [script](https://github.com/rdiankov/openrave/blob/master/python/ikfast.py#L3403)
2. Adding an input argument "--eef_direction" to support customising the direction vector when running "create_ikfast_moveit_plugin.py"

### Checklist


rhaschke (approved):
Thanks for these fixes, which look very reasonable to me. Could you provide some evidence that the behavior is now correct? Ideally, add some unit tests. Minimally, provide some images illustrating the behavior before and after your change.

nothingstopsme:
Hi, @rhaschke, thank you for the feedback.

I have added a unit test based on the current testing implementation for kinematics plugin.

Packages that are needed for this test can be generated by:
```bash
# Given the current working directory is at the root of MoveIt source tree
./moveit_kinematics/test/test_4dof/test_4dof.sh
```

Then this test can be launched together with other plugin tests via:
```bash
catkin run_tests --no-deps -i moveit_kinematics
```

Please note that the package-generating script "test_4dof.sh" needs to be run with a version of "create_ikfast_moveit_plugin.py" supporting the feature "--eef_drection" added in d9625f18b15d83845295592aa3c94d8498df63ab, but the IKFast plugin template itself can be either one (before or after this PR being applied). So another way to verify the changes in this PR is to run "test_4dof.sh" with the current IKFast plugin template (without modifications) and then perform the same unit test. The following test report will show that IKFast plugins created this way fail some of the test cases set up by the script.

rhaschke (approved):

rhaschke:
Thanks for this contribution, particularly adding a test as well.

PR #3499, opened 2023-09-25
more validation on urdf

v4hn:
I just helped a colleague debug these two issues and it's really surprising how few things go wrong when they are violated. E.g., if multiple joints (even of different types) are specified everything works just fine, except that you get constant red link in rviz with the log that "linkname" collides with "linkname".


rhaschke (approved):
Thanks for adding this check. However, I think the second check (for duplicate siblings) can be omitted: due to the recursive nature of the function, duplicates will be identified on a deeper call level.
Also, I suggest skipping duplicates (in addition to issuing an error message).
I added appropriate commits.

v4hn:
Thanks for the iteration and the test. Obviously the first patch was too roundabout.
After playing a bit more with your test, I noticed that urdfdom already enforces the unique link names, so the only case that needed testing here was whether a link can be reached from the root through more than one path.
I simplified the test further and added a more helpful error message.

rhaschke:
Fair enough. For completeness: these are the checks for uniqueness in urdfdom:
- for [links](https://github.com/ros/urdfdom/blob/6d3c5de088359523857befbe730dbdd5543e35ba/urdf_parser/src/model.cpp#L178)
- for [joints](https://github.com/ros/urdfdom/blob/6d3c5de088359523857befbe730dbdd5543e35ba/urdf_parser/src/model.cpp#L221)

I further improved the error message.

PR #3500, opened 2023-09-27
fix apt check in docker workflow

DavidPL1:
### Description

Apparently this check isn't working:

https://github.com/ros-planning/moveit/blob/f8241d8e9b6bbd1c238bf30606414da3ef018182/.github/workflows/docker.yaml#L30-L39

it seems that `echo "no_cache=$have_updates" >> "$GITHUB_OUTPUT"` is not working anymore, and `steps.apt.outputs.no_cache` is always `null`.

So the apt packages status of the image currently has no influence on whether it is rebuilt. 

To fix this issue, this PR, for all jobs using this step, changes the exit value, i.e., the outcome of the job's step, and uses the outcome variable for later checks.

### Checklist

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -35,11 +35,10 @@ jobs:
           image: ${{ env.IMAGE }}
           run: |
             apt-get update
-            have_updates=$(apt-get --simulate upgrade | grep -q "^0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.$" && echo false || echo true)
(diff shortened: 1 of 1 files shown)


rhaschke:
I pushed a fixing commit: e5c1d5a736ab8aeeb45bba5beb44b6b16db7c4a1
The problem was more fundamental: the underlying action was not supporting GHA environment files (see https://github.com/addnab/docker-run-action/pull/49 for details)

Issue #659, opened 2017-10-18
Found empty JointState message error per every execution of plans


130s:
### Description

Ref. #283: Similar error msg reported

When running [chomp tutorial](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/chomp_interface_tutorial.html), the robot on RViz moves but the following error printed. `sensor_msgs/JointState` seems to be returned. I'm just wondering.

```
[ERROR] [1508358890.347763034]: Found empty JointState message
```

```
$ rostopic info /joint_states 
Type: sensor_msgs/JointState

Publishers: 
 * /joint_state_publisher (http://130s-p50:35333/)

Subscribers: 
 * /robot_state_publisher (http://130s-p50:43231/)
 * /move_group (http://130s-p50:35237/)

$ rostopic hz /joint_states 
subscribed to [/joint_states]
average rate: 10.000
        min: 0.100s max: 0.100s std dev: 0.00007s window: 9
average rate: 10.000
        min: 0.100s max: 0.100s std dev: 0.00009s window: 19
average rate: 9.999
        min: 0.100s max: 0.100s std dev: 0.00011s window: 29
average rate: 10.000
        min: 0.100s max: 0.100s std dev: 0.00011s window: 39
average rate: 10.000
        min: 0.100s max: 0.100s std dev: 0.00011s window: 49
average rate: 10.000
        min: 0.100s max: 0.100s std dev: 0.00012s window: 59
```

Overview of your issue here.

### Your environment
* ROS Distro: [Indigo|Jade|Kinetic]
* OS Version: e.g. Ubuntu 16.04
* Source or Binary build?
* If binary, which release version?
* If source, which git commit or tag?

### Steps to reproduce
```
cd catkin_ws/src
git clone git@github.com:ros-planning/moveit.git
git clone git@github.com:ros-planning/moveit_resources.git
cd catkin_ws
rosdep install --from-paths src --ignore-src -r -y
catkin b
source devel/setup.bash
roslaunch moveit_resources demo_chomp.launch
```

### Expected behaviour
No error msg above

### Backtrace or Console output

```
$ roslaunch moveit_resources demo_chomp.launch
... logging to /home/n130s/.ros/log/bbfbc5f6-b443-11e7-8cb7-c85b766a6139/roslaunch-130s-p50-25425.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

[...]

PR #3504, opened 2023-10-02
Gracefully handle zero-size bodies in determineCollisionSpheres()

### LINKED ISSUE #3502: collision_distance_field causes memory exhaustion to size 0 geometry
### Description
A fix idea on collision_distance_field, to prevent memory exhaustion & crash.

### Your environment
* ROS Distro: [Melodic]
* OS Version: Ubuntu 18.04
* Source build
* If source, which branch: melodic_devel

### Steps to reproduce
In the robot urdf file, in one of existing `<collision>`, add an element with geometry of size 0,
ex) 
```
    <collision name="link1">
      <geometry>
        <box size="0 0 0"/>
      </geometry>
    </collision>
```
select CHOMP as planner, which uses collision_distance_field collision checker. 

### Expected behaviour
It should throw an error, if size=0 is illegal.

### Actual behaviour
A rapid consumption of memory occurs, and finally the process gets killed, with a message like below.
> [move_group-3] process has died [pid 502666, exit code -9, cmd /home/ubuntu/catkin_ws/devel/lib/moveit_ros_move_group/move_group --debug __name:=move_group __log:=/home/ubuntu/.ros/log/ea5069cc-11b0-11ee-ac45-0242ac110002/move_group-3.log].
log file: /home/ubuntu/.ros/log/ea5069cc-11b0-11ee-ac45-0242ac110002/move_group-3*.log

The trouble occurs at:
https://github.com/ros-planning/moveit/blob/1fbb2c4c555fe59cb316fea184760468aa702d66/moveit_core/collision_distance_field/src/collision_distance_field_types.cpp#L47
in lines
> L53 unsigned int num_points = ceil(cyl.length / (cyl.radius / 2.0));

and

> L57  for (unsigned int i = 1; i < num_points - 1; i++)

When size 0 box is passed, following calculation happened (may be environment-dependent) 
1) cyl.length / (cyl.radius / 2.0) = 0.0/0.0 = nan;
2) unsigned int num_points = ceil(nan) =0;
Then, in L57 loop limit becomes (num_points - 1)=0xffffffff , in the loop some memory consumption is made, that leads to exhaustion. 

### Fix idea
At least num_points =0 is illegal, and should be rejected. 
Whether it's better to throw exception or just ignore, I'm not sure.

### other information
Gazebo client also crashes with URDF with size 0 object, so maybe size 0 is widely illegal?

rhaschke:
Fixes #3502

### DIFF
--- a/moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
+++ b/moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
@@ -49,22 +49,29 @@ collision_detection::determineCollisionSpheres(const bodies::Body* body, Eigen::
   std::vector<collision_detection::CollisionSphere> css;
   if (body->getType() == shapes::ShapeType::SPHERE)
   {
-    collision_detection::CollisionSphere cs(body->getPose().translation(), body->getDimensions()[0]);
-    css.push_back(cs);
+    const auto radius = body->getDimensions()[0];
+    if (radius > 0.0)
+    {
+      collision_detection::CollisionSphere cs(body->getPose().translation(), radius);
+      css.push_back(cs);
+    }
   }
   else
   {
     bodies::BoundingCylinder cyl;
     body->computeBoundingCylinder(cyl);
-    unsigned int num_points = ceil(cyl.length / (cyl.radius / 2.0));
-    double spacing = cyl.length / ((num_points * 1.0) - 1.0);
-    relative_transform = cyl.pose;
-
-    for (unsigned int i = 1; i < num_points - 1; i++)
+    if (cyl.radius > 0.0 && cyl.length > 0.0)
     {
-      collision_detection::CollisionSphere cs(
-          relative_transform * Eigen::Vector3d(0, 0, (-cyl.length / 2.0) + i * spacing), cyl.radius);
-      css.push_back(cs);
+      unsigned int num_points = ceil(cyl.length / (cyl.radius / 2.0));
+      double spacing = cyl.length / ((num_points * 1.0) - 1.0);
+      relative_transform = cyl.pose;
+
+      for (unsigned int i = 1; i < num_points - 1; i++)
+      {
+        collision_detection::CollisionSphere cs(
+            relative_transform * Eigen::Vector3d(0, 0, (-cyl.length / 2.0) + i * spacing), cyl.radius);
+        css.push_back(cs);
+      }
     }
   }
 


v4hn (approved):
The patch should fix the mentioned problem, but there is another inconsistency still in:

- independent of whether the dimensions are empty, a non-zero [padding can still add volume to bounding cylinders](https://github.com/ros-planning/geometric_shapes/blob/3e4fede08f3f6dc48e8b1ff60f83d5e5c92a07d4/src/bodies.cpp#L601C20-L601C20) (default is 0.0).
- however, the special logic for spheres does not do that and I suggest to change it to
```
BoundingSphere bsphere;
body->computeBoundingSphere(bsphere);
collision_detection::CollisionSphere cs(bsphere.center, bsphere.radius);
```

rhaschke:
> * independent of whether the dimensions are empty, a non-zero [padding can still add volume to bounding cyl.s](https://github.com/ros-planning/geometric_shapes/blob/3e4fede08f3f6dc48e8b1ff60f83d5e5c92a07d4/src/bodies.cpp#L601C20-L601C20)

I'm afraid that I don't understand your point yet: Isn't the bounding cylinder computed from the effective body shape, i.e. including the padding anyway?

> * however, the special logic for spheres does not do that and I suggest to change it

What is special about spheres? Please directly apply the suggestions. I don't see how/where you would like to apply it.

toru-kuga:
@rhaschke  Thank you for the rapid fix!

I tried the update and confirmed that size 0 objects(sphere,cylinder,box) does not cause memory exhaustion,
and plan ends safely.
However the planning ends with another error (seems like https://github.com/ros-planning/moveit/issues/659 ) in my test,
so normal operation test is incomplete, which is probably a different issue from this.

rhaschke:
> However the planning ends with another error

Can you please provide the log error?

toru-kuga:
Log is:

> [ INFO] [1696475524.213156475, 140.539000000]: Planning request received for MoveGroup action. Forwarding to planning pipeline.
[ INFO] [1696475524.213194677, 140.539000000]: Using planning pipeline 'chomp'
[ INFO] [1696475524.459398270, 140.785000000]: CHOMP trajectory initialized using method: quintic-spline 
[ INFO] [1696475524.459462517, 140.785000000]: The following collision detectors are available in the planning scene.
[ INFO] [1696475524.459479380, 140.785000000]: HYBRID
[ INFO] [1696475524.459497176, 140.785000000]: Active collision detector is: HYBRID
[ INFO] [1696475524.606311735, 140.931000000]: First coll check took 0.146779181
[ INFO] [1696475524.623696842, 140.949000000]: iteration: 0
[ INFO] [1696475524.633145897, 140.958000000]: Chomp Got mesh to mesh safety at iter 0. Breaking out early.
[ INFO] [1696475524.666412870, 140.990000000]: Chomp path is collision free
[ INFO] [1696475524.666476809, 140.991000000]: Terminated after 12 iterations, using path from iteration 11
[ INFO] [1696475524.666498467, 140.991000000]: Optimization core finished in 0.049508 sec
[ INFO] [1696475524.666524362, 140.991000000]: Time per iteration 0.004128
[ INFO] [1696475524.666543996, 140.991000000]: Planned with Chomp Parameters (learning_rate, ridge_factor, planning_time_limit, max_iterations), attempt: # 1 
[ INFO] [1696475524.666567294, 140.991000000]: Learning rate: 0.010000 ridge factor: 0.010000 planning time limit: 10.000000 max_iterations 200 
[ERROR] [1696475524.666927508, 140.991000000]: Found empty JointState message

operation is : 
* launch a simulation model to gazebo and Rviz 
* In Rviz select chomp/CHOMP pipeline pair
* In Rviz set start=current, goal=current, and "Plan"
  
This error is probaply not (directly) related to this update, because it occurs even after I undo the update and rebuild moveit.

rhaschke:
@toru-kuga, I think the "empty JointState message" error stems from your specific URDF. I don't see that for `panda_moveig_config`.

v4hn:
`empty JointState` appears whenever a `RobotState` message is not marked as `is_diff`. This likely has a very different culprit.
The issue I pointed out is also independent of this concrete PR (which is why I didn't just push it).
Merging the patch to improve the current state.

I might file a follow-up PR with my concerns if I can actually show-case an issue in a test (the API for all of this is disastrous).

toru-kuga:
@rhaschke I tried this on our prototype including UR5 and other parts and joint (out of Moveit's control).
So I agree that the error comes from other part in URDF, maybe the joint.

PR #3506, opened 2023-10-06
consider distance field padding for spheres

v4hn:
Addresses a rather small inconsistency I pointed out [here](https://github.com/ros-planning/moveit/pull/3504#pullrequestreview-1657191555).

Two more helper patches are included to make the test nicer and the API slightly better.

### DIFF
--- a/moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
+++ b/moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
@@ -49,11 +49,11 @@ collision_detection::determineCollisionSpheres(const bodies::Body* body, Eigen::
   std::vector<collision_detection::CollisionSphere> css;
   if (body->getType() == shapes::ShapeType::SPHERE)
   {
-    const auto radius = body->getDimensions()[0];
-    if (radius > 0.0)
+    bodies::BoundingSphere bsphere;
+    body->computeBoundingSphere(bsphere);
+    if (bsphere.radius > 0.0)
     {
-      collision_detection::CollisionSphere cs(body->getPose().translation(), radius);
-      css.push_back(cs);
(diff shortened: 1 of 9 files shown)


moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
```diff
@@ -52,8 +52,9 @@ collision_detection::determineCollisionSpheres(const bodies::Body* body, Eigen::
     const auto radius = body->getDimensions()[0];
```
rhaschke, line 52: This radius doesn't consider padding and scaling. I guess that's what you wanted to consider?

moveit_core/utils/src/robot_model_test_utils.cpp
```diff
@@ -265,6 +265,17 @@ RobotModelBuilder& RobotModelBuilder::addVisualBox(const std::string& link_name,
   return *this;
 }
 
+RobotModelBuilder& RobotModelBuilder::addCollisionSphere(const std::string& link_name, double radius,
+                                                         geometry_msgs::Pose origin)
+{
+  urdf::CollisionSharedPtr coll(new urdf::Collision);
+  urdf::SphereSharedPtr geometry(new urdf::Sphere);
```
rhaschke, line 272: This should use std::make_shared

rhaschke (approved):
I will push fixes for the suggested changes.

moveit_core/collision_distance_field/src/collision_distance_field_types.cpp
v4hn: No need for scaling when it checks only for 0.0, but fair enough! :)

rhaschke (approved):

Issue #3515, opened 2023-10-26
FCLShapeCache does not cache attached collision objects between different robot states on the same trajectory


rr-mark:
### Description

tl;dr:

The attached collision objects at each waypoint along a RobotTrajectory are not aliased, and so the FCLShapeCache misses the cache for each new waypoint, and a huge amount of processing time is spent in the fcl collision detection algorithm looping over every vertex and index in every collision object for every step of the trajectory.

Longer context:

I am planning trajectories using pilz_industrial_motion_planner's plan_sequence_path service. I am using relatively large attached collision meshes (~100k faces). These trajectories take a very long time to plan (~1s per waypoint), and more importantly the move_group node leaks memory continuously (~10Mb per waypoint), which is not recovered when the plan is finished.

I have been profiling the move_group, to see if what I'm wanting to do is possible, and have found that almost all (>99%) of the processing time is spent in the `FCLGeometryConstPtr::createCollisionGeometry` method. I believe (although correct me if I'm wrong) that the vast majority of this time would be saved if the FCLShapeCache was hit, but instead it is missed every time. This is happening because the attached collision objects associated with each waypoint along the RobotTrajectory are different objects (even though they represent the same objects, just moved slightly).

### Your environment
* ROS Distro: Noetic
* OS Version: 20.04
* Source or Binary build? Both; first the binary build, and then the source code when debugging.
* If binary, which release version? 1.1.12
* If source, which branch? master: b4b1a27f605031c54e80a9b3465f7a3cabe6cc88

### Steps to reproduce
I can set this up if it would help, but I don't think it's that kind of bug.

### Expected behaviour
I would assume the attached collision objects along a single trajectory would be cached, so they didn't have to be processed for every step along the trajectory.

### Actual behaviour
Planning takes a long time, and the memory usage of the move_group nodes increases quickly, and does not go down again when planning has finished.

### Backtrace or Console output

N/A

rr-mark:
I am more than happy to put some time into raising a PR for this, if my reasoning about the caching is correct. Although I am unsure whether it would be better to change the behaviour of the cache or to change how the robot trajectory is generated so that the attached collision objects are aliased.

simonschmeisser:
[...]

PR #3518, opened 2023-10-30
Adds a MeshCache to _msgsToAttachedBody to prevent duplicate attached collision objects being stored in FCLShapeCache

rr-mark:
### Description

Fixes https://github.com/ros-planning/moveit/issues/3515

Previously, when planning a sequence motion, the attached collision objects would be converted to separate shapes::Mesh* for each waypoint along the path. Then each instance would be come a separate fcl::BVHModel instance in the FCLShapeCache, leading to large computational costs and memory usage when using large attached collision objects.

This PR adds caching of shapes constructed from meshes to _msgsToAttachedBody in conversions.cpp. This means that the shapes::Mesh* pointers passed into the FCLShapeCache are correctly aliased, and fcl::BVHModels do not need to be regenerated.

The mesh cache has an (approximate) maximum memory usage, and evicts cache entries using a last-used-first-out strategy when this memory usage is exceeded.

### Checklist
- [✓] **Required by CI**: Code is auto formatted using [clang-format](http://moveit.ros.org/documentation/contributing/code)
- [ ] Extend the tutorials / documentation [reference](http://moveit.ros.org/documentation/contributing/) N/A (I think?)
- [ ] Document API changes relevant to the user in the [MIGRATION.md](https://github.com/ros-planning/moveit/blob/master/MIGRATION.md) notes N/A (no API changes)
- [✓] Create tests, which fail without this PR [reference](https://ros-planning.github.io/moveit_tutorials/doc/tests/tests_tutorial.html)
- [ ] Include a screenshot if changing a GUI N/A (no GUI changes)

### DIFF
--- a/moveit_core/robot_state/CMakeLists.txt
+++ b/moveit_core/robot_state/CMakeLists.txt
@@ -43,4 +43,7 @@ if(CATKIN_ENABLE_TESTING)
 
   catkin_add_gtest(test_aabb test/test_aabb.cpp)
   target_link_libraries(test_aabb ${MOVEIT_LIB_NAME} moveit_test_utils)
+
+  catkin_add_gtest(test_mesh_cache test/test_mesh_cache.cpp)
+  target_link_libraries(test_mesh_cache ${MOVEIT_LIB_NAME} moveit_test_utils)
 endif()
--- a/moveit_core/robot_state/CMakeLists.txt
+++ b/moveit_core/robot_state/CMakeLists.txt
@@ -2,6 +2,7 @@ set(MOVEIT_LIB_NAME moveit_robot_state)
 
 add_library(${MOVEIT_LIB_NAME}
   src/attached_body.cpp
+  src/mesh_cache.cpp
   src/conversions.cpp
   src/robot_state.cpp
   src/cartesian_interpolator.cpp
--- a/moveit_core/robot_state/include/moveit/robot_state/mesh_cache.h
+++ b/moveit_core/robot_state/include/moveit/robot_state/mesh_cache.h
@@ -41,43 +41,81 @@ namespace moveit
 {
 namespace core
 {
+
+/// \brief The type used to hash meshes.
 using MeshHash = std::size_t;
+
+/// \brief The type used to hash times.
 using TimeHash = std::pair<int, int>;
 
+/// \brief Convert a shape_msgs::mesh to a std::string.
 std::string meshString(const shape_msgs::Mesh& mesh);
 
+/// \brief Convert a std::string to a MeshHash.
 MeshHash hashMeshString(const std::string& mesh_string);
 
+/// \brief Convert a shape_msgs::Mesh to a MeshHash.
 MeshHash hashMesh(const shape_msgs::Mesh& mesh);
 
+/// \brief Convert a ros::Time to a TimeHash.
 TimeHash hashTime(const ros::Time& time);
 
+/** \brief Control block used to store a shapes::ShapeConstPtr, and related metadata.
+
+    The metadata allows for exact comparisons with shape_msgs::Mesh types,
+    and for tracking the relative usage times of meshes.*/
 struct MeshCacheControlBlock
 {
+  /// \brief The input shape_msgs::Mesh as a std::string.
   std::string mesh_string_;
+  /// \brief The hash of `mesh_string_`.
   MeshHash mesh_hash_;
+  /// \brief The cached shape.
   shapes::ShapeConstPtr shape_;
+  /// \brief The last time this shape was used.
   ros::Time last_used_;
+  /// \brief The hash of `last_used_`.
   TimeHash last_used_hash_;
+  /// \brief The approximate memory use of this control block.
   std::size_t approximateMemoryUse() const;
 };
 
+/** \brief A cache of shapes::ShapeConstPtr shapes, constructed from shape_msgs::Mesh.
+
+    This cache is designed to allow shapes to be aliased, rather than being regenerated.
+    Automatically evicts members to maintain an (approximate) maximum memory usage.
+    Uses a least-recently-used cache eviction policy.*/
 class MeshCache
 {
 public:
+  /** \brief Cache constructor
+      \param min_size_to_cache The minimum size of a shape_msgs::Mesh to cache.
+      \param max_cache_size The (approximate) maximum memory usage of the cache.*/
   MeshCache(std::size_t min_size_to_cache = 0, std::size_t max_cache_size = std::numeric_limits<std::size_t>::max());
-  static MeshCache& threadLocalCache();
+  /** \brief Returns a reference to a thread-local static MeshCache. */
+  static MeshCache& threadLocalCache(std::size_t min_size_to_cache = 0,
+                                     std::size_t max_cache_size = std::numeric_limits<std::size_t>::max());
+  /** \brief Returns the shapes::ShapeConstrPtr corresponding to `mesh`.
+
+      The returned shape will come from the cache if present.
+      The generated shape will be added to the cache if it is not present.*/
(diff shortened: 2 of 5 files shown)


rr-mark:
> I really recommend you to install pre-commit, saves some round trips with the ci. https://moveit.ros.org/documentation/contributing/code/

Yeah, sorry, my editor had the wrong clang-format loaded and kept undoing my changes. I think it's fixed now.

simonschmeisser:
Very nice PR, thanks!

> Some open questions I have on this:
> 
>     * What values should be used for min_size_to_cache and max_cache size? The 1Mb and 1Gb values are just guesses. I feel like these should probably be user-configurable, although I don't know how that process works here.

Did you measure the overhead? Is there a size where it does not help to use the cache? I would suspect the turn-even << 1MB?
> 
>     * Is there a better way to hash a mesh than converting it to string and hashing the string?

Maybe calculating hashes directly on `mesh.triangles.data()` and `mesh.vertices.data()`? 
> 
>     * The MeshCacheControlBlock currently stores both the ShapeConstPtr and the std::string mesh, which makes it twice as large as it maybe needs to be. Any smart ideas on how to guarantee meshes are correct without storing both?

So this basically boils down to the question if hash collisions are likely (enough) to warrant the storage costs. Maybe storing something like number of vertices and/or edges would be sufficient?

rr-mark:
> > * What values should be used for min_size_to_cache and max_cache size? The 1Mb and 1Gb values are just guesses. I feel like these should probably be user-configurable, although I don't know how that process works here.
> 
> Did you measure the overhead? Is there a size where it does not help to use the cache? I would suspect the turn-even << 1MB?

I haven't measured any overhead, and I'm not sure if my use cases are representative of wider use. I think the use case for the minimum size is if someone has a few large collision objects and many small ones, then the idea is to not cache the small objects to avoid kicking the large objects out of the cache. But I guess that use case is relatively specific, so the min size should default to zero?

For the maximum cache size, I guess it depends on how much RAM the user's computer has.

> > * Is there a better way to hash a mesh than converting it to string and hashing the string?
> 
> Maybe calculating hashes directly on `mesh.triangles.data()` and `mesh.vertices.data()`?

Makes sense. I'll try that and see how it goes.

> > * The MeshCacheControlBlock currently stores both the ShapeConstPtr and the std::string mesh, which makes it twice as large as it maybe needs to be. Any smart ideas on how to guarantee meshes are correct without storing both?
> 
> So this basically boils down to the question if hash collisions are likely (enough) to warrant the storage costs. Maybe storing something like number of vertices and/or edges would be sufficient?

I was assuming hash collisions were very unlikely, but that the cache returning the wrong shape would be critically bad.

rr-mark:
> > > * What values should be used for min_size_to_cache and max_cache size? The 1Mb and 1Gb values are just guesses. I feel like these should probably be user-configurable, although I don't know how that process works here.
> > 
> > 
> > Did you measure the overhead? Is there a size where it does not help to use the cache? I would suspect the turn-even << 1MB?
> 
> I haven't measured any overhead, and I'm not sure if my use cases are representative of wider use. I think the use case for the minimum size is if someone has a few large collision objects and many small ones, then the idea is to not cache the small objects to avoid kicking the large objects out of the cache. But I guess that use case is relatively specific, so the min size should default to zero?
> 
> For the maximum cache size, I guess it depends on how much RAM the user's computer has.
> 
> > > * Is there a better way to hash a mesh than converting it to string and hashing the string?
> > 
> > 
> > Maybe calculating hashes directly on `mesh.triangles.data()` and `mesh.vertices.data()`?
> 
> Makes sense. I'll try that and see how it goes.
> 
> > > * The MeshCacheControlBlock currently stores both the ShapeConstPtr and the std::string mesh, which makes it twice as large as it maybe needs to be. Any smart ideas on how to guarantee meshes are correct without storing both?
> > 
> > 
> > So this basically boils down to the question if hash collisions are likely (enough) to warrant the storage costs. Maybe storing something like number of vertices and/or edges would be sufficient?
> 
> I was assuming hash collisions were very unlikely, but that the cache returning the wrong shape would be critically bad.

I've changed the default `min_size_to_cache` to 0, and added a TODO to consider exposing these parameters to the user (I don't think I know enough about how user-defined parameters in moveit work to do a good job of this myself).

I looked into changing the hash to use the raw message type rather than the string representation, but I think this is just going to add a lot of code complexity for little gain.

rr-mark:
It looks like there's a problem using this PR with OMPL:

When two attached collision objects in the scene have the same mesh (and so their shape::Mesh* pointers are aliased by the MeshCache in this PR), the OMPL planner hangs forever on `ompl_parallel_plan_.solve(...)`.

Before I investigate further, I'd like to know if either:
 - my approach in this PR is right, and this is a bug in OMPL (or the OMPL wrapper in Moveit), which I should investigate
 - OR the aliasing of meshes in this PR is not desirable behaviour, and I should try some other method of de-duplicating the entries in the FCLShapeCache
 - OR some other course of action would be better

rr-mark:
@rhaschke could I get your thoughts on this PR and the issue with OMPL when you have the time? I'm keen to get this ready to merge, but not sure how to proceed.

simonschmeisser:
This is a bit surprising as I couldn't spot any obvious new locks/mutexes and the cache is thread-local.

Could you provide a backtrace from when it hangs?

```
echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
gdb --pid=$(pidof moveit_core) -ex 'set logging on' -ex 'set pagination off' -ex 'thread apply all bt full' -ex detach -ex quit
```

rr-mark:
> Could you provide a backtrace from when it hangs?
Sure, I'll give that a go.

> This is a bit surprising as I couldn't spot any obvious new locks/mutexes and the cache is thread-local.
I don't think it's my code that's hanging, but rather than OMPL is hanging as a result of the aliased meshes.

rr-mark:
I am struggling to attach gdb to the move_group in our system due to our docker and dependency management setup. I have passed this over to @rr-tom-noble who knows more about these systems. We should have a backtrace for you soon.

While trying to attach gdb, I did notice that it is not a full hang. OMPL just goes a lot slower than it should. I'm wondering if it is some kind of cache thrashing on the OMPL end. Hopefully we'll know more shortly.

rr-mark:
[gdb.txt](https://github.com/ros-planning/moveit/files/13416783/gdb.txt)

rr-mark:
I wonder if this is the wrong solution to the problem, and if something a bit more targeted to FCL would be more appropriate?

If I can get FCL to hash the meshes in its cache, I could change that cache to identify identical meshes which don't share a pointer, and change that cache to look more like the cache I have implemented here.

That would hopefully solve the problem without so many possible downstream consequences, and not introduce a second cache where a second cache isn't really needed.

I'll probably look into doing this sometime next week.

simonschmeisser:
> When two attached collision objects in the scene have the same mesh (and so their shape::Mesh* pointers are aliased by the MeshCache in this PR), the OMPL planner hangs forever on `ompl_parallel_plan_.solve(...)`.

I missed this particular point (**two**) until now. The cache used in the FCL collision checker plugin (not actually part of FCL but MoveIt) assumes there's at most one attached and one non-attached instance of a certain mesh. Your use-case breaks this assumption.

https://github.com/ros-planning/moveit/blob/d8e0a72e9a69202521aad251a9fa8e2696d8934e/moveit_core/collision_detection_fcl/src/collision_common.cpp#L370

https://github.com/ros-planning/moveit/blob/d8e0a72e9a69202521aad251a9fa8e2696d8934e/moveit_core/collision_detection_fcl/src/collision_common.cpp#L695

Not sure how to fix/extend this actually ...

rr-mark:
Since this PR introduces downstream problems, and the performance issues appear to be fixed by https://github.com/moveit/moveit/pull/3590, I'm going to close it.

